PR #39, opened 2026-06-15
drop shapely as an intersection backend (#36)

### LINKED ISSUE #36: Why are we using shapely as a backend at all?
> def _resolve_backend(backend: str, grid) -> str:
    """Resolve ``"auto"`` to a concrete, grid-appropriate backend.

 >   Prefers exact S2 (``spherely.SpatialIndex``) when available. Without it,
    falls back per grid family: **mortie** for HEALPix (its native MOC order
    matches the grid), **shapely** for rectilinear (a global MOC order is far
    too coarse for fine projected tiles and over-commissions every granule to
    every shard)

This seems wrong to me-- right now we're checking if spherely with spatial index is present, and then are falling back to shapely if it isn't. We should be checking if spatial index is present in spherely-- and if not, we should still use spherely, just the brute method! Shapely really shouldn't be used at all-- there's too many issues with crossing merdians and near pole performance.

Am I missing something here? Is there a good reason why we would use shapely instead of spherely?

espg:
Closes #36

## What this changes

Removes shapely as the **intersection backend** in `src/zagg/catalog/shardmap.py`, per [@espg's Option A decision](https://github.com/englacial/zagg/issues/36#issuecomment-4703872859) ("simpler code to maintain"). shapely stays a dependency — it is still used for WKB decode (`catalog/sources.py`) and footprint geometry (`grids/`); only the WGS84 `STRtree` *intersection* path is gone.

### Approach

The fallback ladder is now **`SpatialIndex` → spherely-brute → loud error`**, all sphere-correct:

- `_resolve_backend("auto", grid)` returns `"spherely"` whenever spherely **imports** (regardless of `hasattr(spherely, "SpatialIndex")`). Only when spherely is genuinely absent does it fall to `"mortie"` for HEALPix grids; non-HEALPix grids resolve to `"spherely"` and then raise.
- `_intersect_spherely` uses `spherely.SpatialIndex` when present (fast tree query) and elementwise `spherely.intersects` otherwise — a brute `O(granules × shards)` scan with no tree prefilter. Sphere-correct, no fork needed. Degraded perf is accepted until `SpatialIndex` lands in spherely upstream ([@espg accepted this tradeoff](https://github.com/englacial/zagg/issues/36#issuecomment-4703872859)).
- `_intersect_shapely` and the `"shapely"` entry in `_BACKENDS` are **deleted**. The genuinely-absent-spherely case raises `ImportError` with an install pointer (`_SPHERELY_INSTALL_HINT`, pointing at the README spherely fork install).

```python
# _resolve_backend (auto):
if _spherely_available():
    return "spherely"
is_healpix = hasattr(grid, "parent_order") and hasattr(grid, "child_order")
return "mortie" if is_healpix else "spherely"   # spherely path raises if absent
```

`_spherely_available()` uses `importlib.import_module` (not `find_spec`) so it is robust whether spherely is installed or stubbed in `sys.modules`.

## Phases

- [x] **Phase 1** (`06c07b0`) — rewrite `_resolve_backend` + `_intersect_spherely` (SpatialIndex-or-brute), delete `_intersect_shapely` and the shapely `_BACKENDS` entry, add the loud raise-with-install-pointer; refresh module/`build` docstrings.
- [x] **Phase 2** (`900e5e2`) — tests for the brute path (a fake stock-spherely module forcing the no-`SpatialIndex` branch) and the absent-spherely `ImportError`; updated `_resolve_backend` expectations (no shapely fallback). Also fixed one new mypy `[assignment]` finding my phase-1 refactor introduced (renamed `polys` → `poly_arr`).
- [x] **Phase 3** (`5a97cb4`) — docs note on the brute-force perf caveat (`README.md`, `docs/api/catalog.md`, `docs/rectilinear.md`).

## How it was tested

- `pytest tests/test_shardmap.py` — **15 passed, 1 skipped**. New `TestBuildSpherelyBrute` exercises the real `_intersect_spherely` brute branch via a fake spherely (planar-bbox `intersects`, no `SpatialIndex`) and reproduces the old spatial-split assertions; `TestSpherelyAbsent` asserts the `ImportError` (explicit `backend="spherely"` and `auto` on a rect grid). The `SpatialIndex` fork test is skipped (fork not in the venv).
- Full suite — **212 passed, 1 skipped** (`uv run pytest -q`).
- `ruff check src tests --select=E,F,W,I --ignore=E501` — clean (the CI lint bot's flags).
- `pre-commit` codespell + mypy hooks run on the changed files: codespell clean; mypy clean on the test file and clean on `shardmap.py` **except one pre-existing** error (see below).

## Questions for review

1. **Cannot run the real spherely backends here** — spherely isn't installed in this env (it's the non-PyPI fork). The brute path and absent-spherely error are exercised with a lightweight fake spherely module (mirroring `create_polygon`/`intersects`) and `sys.modules` patching; the `SpatialIndex` fork path stays skip-gated as before. Worth a confirm against a real stock-spherely build (does it expose `intersects` with the `(array, scalar) → bool array` shape the brute path assumes, and import cleanly in the arm64 Lambda layer?).
2. **Pre-existing mypy error left untouched** (§4): `shardmap.py:292` `_BACKENDS[chosen](...)` → `Cannot call function of unknown type [operator]`. It is on `main` already, unrelated to this change — flagging rather than fixing.
3. **Pre-existing `ruff format` drift** (not this PR): `ruff format --check src` reports 13 files would be reformatted on a clean `main` too. CI lint (`lint.yml`) runs `ruff check` on the diff only — not `ruff format` — so this isn't a gate; left alone to avoid touching unrelated files.

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,18 +23,15 @@ comes from the **same pipeline config** the aggregator uses (`--config`), so the
 shard map can't be built against a different grid than the run.
 
 ```bash
-# Install the catalog extra (STAC fetch + shard-map build). The default
-# geometry backend is mortie (HEALPix MOC); non-HEALPix grids use spherely.
+# Install the catalog extra (STAC fetch + shard-map build). The geometry
+# backend defaults to `auto`: exact-S2 spherely if its fork is installed (used
+# for all grids), else mortie (HEALPix MOC); rectilinear grids require spherely.
 pip install 'zagg[catalog]'
 
-# spherely (exact-S2 intersection): the SpatialIndex build is a fork not on
-# PyPI (benbovy/spherely#118) — install it separately for the fast tree path
-# (pick the wheel for your python/platform from the release assets):
+# Optional: the exact-S2 spherely SpatialIndex backend is a fork not on PyPI
+# (benbovy/spherely#118) — install it separately (pick the wheel for your
+# python/platform from the release assets):
 # pip install "spherely @ https://github.com/espg/spherely/releases/download/v0.1.1-spatialindex/spherely-0.1.1+spatialindex-cp312-cp312-manylinux_2_28_x86_64.whl"
-# Without the fork, stock spherely still works via a slower brute-force path
-# (O(granules x shards), no tree prefilter) until SpatialIndex lands upstream.
-# shapely is no longer an intersection backend (it had antimeridian/near-pole
-# bugs, #36) — it is still used for WKB decode and footprint geometry.
 
 # ICESat-2 convenience — cycle number computes dates automatically:
 uv run python -m zagg.catalog --config atl06.yaml --short-name ATL06 --cycle 22 \
--- a/README.md
+++ b/README.md
@@ -23,14 +23,18 @@ comes from the **same pipeline config** the aggregator uses (`--config`), so the
 shard map can't be built against a different grid than the run.
 
 ```bash
-# Install the catalog extra (STAC fetch + shard-map build). The geometry
-# backend is mortie (HEALPix) / shapely (rectilinear) by default.
+# Install the catalog extra (STAC fetch + shard-map build). The default
+# geometry backend is mortie (HEALPix MOC); non-HEALPix grids use spherely.
 pip install 'zagg[catalog]'
 
-# Optional: the exact-S2 spherely SpatialIndex backend is a fork not on PyPI
-# (benbovy/spherely#118) — install it separately (pick the wheel for your
-# python/platform from the release assets):
+# spherely (exact-S2 intersection): the SpatialIndex build is a fork not on
+# PyPI (benbovy/spherely#118) — install it separately for the fast tree path
+# (pick the wheel for your python/platform from the release assets):
 # pip install "spherely @ https://github.com/espg/spherely/releases/download/v0.1.1-spatialindex/spherely-0.1.1+spatialindex-cp312-cp312-manylinux_2_28_x86_64.whl"
+# Without the fork, stock spherely still works via a slower brute-force path
+# (O(granules x shards), no tree prefilter) until SpatialIndex lands upstream.
+# shapely is no longer an intersection backend (it had antimeridian/near-pole
+# bugs, #36) — it is still used for WKB decode and footprint geometry.
 
 # ICESat-2 convenience — cycle number computes dates automatically:
 uv run python -m zagg.catalog --config atl06.yaml --short-name ATL06 --cycle 22 \
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -26,7 +26,7 @@ WKB decode (``sources.py``) and footprint geometry (``grids/``).
 """
 from __future__ import annotations
 
-import importlib.util
+import importlib
 import json
 import time
 from dataclasses import dataclass, field
@@ -69,10 +69,19 @@ def _resolve_backend(backend: str, grid) -> str:
     """
     if backend != "auto":
         return backend
-    if importlib.util.find_spec("spherely") is not None:
+    if _spherely_available():
         return "spherely"
     is_healpix = hasattr(grid, "parent_order") and hasattr(grid, "child_order")
     return "mortie" if is_healpix else "spherely"
+
+
+def _spherely_available() -> bool:
+    """True if ``import spherely`` succeeds (any build, fork or stock)."""
+    try:
+        importlib.import_module("spherely")
+    except ImportError:
+        return False
+    return True
 
 
 def _region_parts(region, metadata) -> list:
@@ -122,9 +131,9 @@ def _intersect_spherely(records, grid, all_shards) -> Dict[int, List[int]]:
             idx.append(i)
     if not polys:
         return {}
-    polys = np.asarray(polys)
+    poly_arr = np.asarray(polys)
     has_index = hasattr(spherely, "SpatialIndex")
-    tree = spherely.SpatialIndex(polys) if has_index else None
+    tree = spherely.SpatialIndex(poly_arr) if has_index else None
 
     out: Dict[int, List[int]] = {}
     for shard in all_shards:
@@ -136,7 +145,7 @@ def _intersect_spherely(records, grid, all_shards) -> Dict[int, List[int]]:
         if tree is not None:
             hits = tree.query(s_poly, predicate="intersects")
         else:
-            hits = np.flatnonzero(spherely.intersects(polys, s_poly))
+            hits = np.flatnonzero(spherely.intersects(poly_arr, s_poly))
         if len(hits) > 0:
             out[int(shard)] = [idx[int(h)] for h in hits]
     return out
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -11,17 +11,22 @@ endpoint at dispatch time via ``data_source.driver`` -- the map itself stays
 endpoint-neutral and never needs the Catalog at run time. It also records the
 grid ``signature()`` so a run can refuse a map built for a different grid.
 
-Geometry backends (all sphere-aware where it matters):
+Geometry backends (all sphere-correct):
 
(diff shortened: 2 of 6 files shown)


espg (commented):
Adversarial review of the shapely-backend drop. The core diff is correct and matches the #36 spec: the fallback ladder (SpatialIndex -> brute `spherely.intersects` -> loud ImportError), `_resolve_backend` branches (auto+spherely -> spherely; auto+HEALPix-no-spherely -> mortie; auto+non-HEALPix-no-spherely -> spherely which then raises), index-mapping via `idx`, the `_intersect_shapely`/`_BACKENDS["shapely"]` deletion, and the docs all line up. Tests genuinely exercise the real brute branch. Only two minor, non-blocking findings inline.

---

README.md
```diff
@@ -23,14 +23,18 @@ comes from the **same pipeline config** the aggregator uses (`--config`), so the
 shard map can't be built against a different grid than the run.
 
 ```bash
-# Install the catalog extra (STAC fetch + shard-map build). The geometry
-# backend is mortie (HEALPix) / shapely (rectilinear) by default.
+# Install the catalog extra (STAC fetch + shard-map build). The default
+# geometry backend is mortie (HEALPix MOC); non-HEALPix grids use spherely.
```
espg: 🤖 *from Claude (review)*

This line misstates the resolved default and contradicts the spec for #36 (and `docs/api/catalog.md`, which gets it right). Per `_resolve_backend`, `auto` resolves to **spherely whenever spherely imports — for all grids, including HEALPix**; mortie is only the fallback for HEALPix *when spherely is absent*. As written ("default geometry backend is mortie (HEALPix MOC); non-HEALPix grids use spherely") it reads as: mortie is the default for HEALPix and spherely is reserved for non-HEALPix. That's backwards from the actual resolution order. Suggest matching the catalog.md phrasing: default `auto` → spherely whenever importable, else mortie for HEALPix only.

---

src/zagg/catalog/shardmap.py
```diff
@@ -267,8 +266,10 @@ def build(
             ``signature``).
         region : list of (lats, lons), optional
             Coverage mask in WGS84. Defaults to the catalog bbox rectangle.
-        backend : {"auto", "spherely", "mortie", "shapely"}
-            Geometry backend. ``"auto"`` -> spherely if available, else mortie.
+        backend : {"auto", "spherely", "mortie"}
```
espg, line 269: 🤖 *from Claude (review)*

The backend set was correctly narrowed to `{"auto", "spherely", "mortie"}` here, but the removal is **incomplete**: `src/zagg/catalog/__init__.py:200` still lists `"shapely"` in the CLI argparse choices:

```python
parser.add_argument("--backend", default="auto",
                    choices=["auto", "spherely", "mortie", "shapely"])
```

So `--backend shapely` is still an accepted CLI flag, but it now dies with `ValueError: unknown backend: 'shapely'` at build time (passthrough in `_resolve_backend` → not in `_BACKENDS`). That's a broken, user-facing choice that should be dropped to fully satisfy #36 ("delete the shapely backend"). `__init__.py` isn't in this PR's diff — please add the one-line edit (and ideally a small test asserting `shapely` is no longer accepted) in this PR rather than leaving a dangling option.

---

tests/test_shardmap.py
```diff
@@ -66,33 +73,97 @@ def _granule_shards(sm):
...
+                and self.y0 <= other.y1 and other.y0 <= self.y1)
+
+
+def _fake_create_polygon(*, shell, oriented):  # noqa: ARG001 (mirror real sig)
+    lons = [pt[0] for pt in shell]
+    lats = [pt[1] for pt in shell]
+    return _FakePoly(lons, lats)
+
+
+def _fake_intersects(a, b):
```
espg, line 100: 🤖 *from Claude (review)*

Worth being explicit about what this fake does and does **not** verify. The fake reduces every polygon to a planar lon/lat AABB and `intersects` is a planar box-overlap test, so these tests confirm the brute branch's **plumbing** — `_intersect_spherely` building `poly_arr`, taking the `tree is None` branch, calling `spherely.intersects(poly_arr, s_poly)`, and mapping `np.flatnonzero(...)` hits back through `idx` — but they do **not** exercise S2/sphere-correctness (antimeridian/near-pole), which is the whole reason shapely was dropped (#36). The module docstring acknowledges the real S2 path is covered in the conda sidecar, so this is a documented limitation, not a defect — flagging so the reviewer knows the "brute path is correct" guarantee here rests on the sidecar `TestBuildSpherely`, not these.

The plumbing fidelity itself looks right: `spherely.intersects(array, scalar)` broadcasts elementwise in real spherely, matching `_fake_intersects`'s `np.atleast_1d` + per-element loop, and the array/scalar arg order is fine since `intersects` is symmetric. Index-mapping via `idx` is identical to the SpatialIndex branch. No bug here.

---

tests/test_shardmap.py
```diff
@@ -66,33 +73,97 @@ def _granule_shards(sm):
...
+def fake_spherely(monkeypatch):
+    """Install a brute-only fake spherely module (no SpatialIndex)."""
+    mod = types.ModuleType("spherely")
+    mod.create_polygon = _fake_create_polygon
+    mod.intersects = _fake_intersects
+    monkeypatch.setitem(sys.modules, "spherely", mod)
+    return mod
+
+
+class TestBuildSpherelyBrute:
```
espg, line 115: 🤖 *from Claude (review)*

Minor test-coverage gap (non-blocking): `_intersect_spherely`'s `if not polys: return {}` early-out (all granule footprints fail `_to_spherely_polygon`) is not exercised on the brute path. It's a cheap case to add with the fake (e.g. a catalog whose granules produce no valid polygons, asserting an empty shard map) and would lock in that the backend degrades to `{}` rather than erroring. Up to you whether it's worth it given the sidecar coverage.

---

espg (commented):
🤖 *from Claude (review)*

Core diff matches the #36 spec well — `_resolve_backend` branches are correct (auto+spherely→spherely for all grids, auto+HEALPix-no-spherely→mortie, auto+rect-no-spherely→spherely-then-raise), the brute path's `spherely.intersects(poly_arr, s_poly)` + `np.flatnonzero` → `idx` mapping is right, and the absent-spherely case raises loudly with an install pointer. One real gap to fix in this PR: the CLI `--backend` choices in `src/zagg/catalog/__init__.py:200` still list `"shapely"`, so it's an accepted-but-broken flag (resolves to a `ValueError`). Plus a README doc inaccuracy on the default-backend wording and two minor non-blocking test-coverage notes. Details inline.

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review's diff-scoped findings (`671aa25`):

- **Bug — CLI accepted a broken backend.** The `--backend` argparse `choices` in `src/zagg/catalog/__init__.py` still listed `"shapely"`, so `--backend shapely` parsed fine and then raised `ValueError: unknown backend: 'shapely'` in `build`. Dropped `"shapely"` from the choices; added `test_cli_rejects_shapely_backend` (argparse `SystemExit` at parse time).
- **README inaccuracy.** `README.md` said the backend is "mortie (HEALPix) / shapely (rectilinear) by default" — backwards. Corrected to match `_resolve_backend` and `docs/api/catalog.md`: `auto` → spherely whenever its fork imports (all grids), else mortie for HEALPix; rectilinear requires spherely.
- **Test gap.** Added `test_brute_empty_records_early_out` covering the `if not polys: return {}` early-out on the brute path.

The two non-blocking review notes (the fake-spherely tests verify plumbing, not true S2 sphere-correctness — that's the doc'd sidecar's job) stand as flagged for you.

Testing: `pytest` full suite **214 passed, 1 skipped**; `ruff check src tests` clean (CI's gate). Phases 1–3 + this fold complete the issue's scope; the only open confirm is Q1 (validating the real stock-spherely `intersects` shape against an actual build, which isn't installed in this env). Marking `waiting` for your merge / that confirm.

---

espg:
> The two non-blocking review notes (the fake-spherely tests verify plumbing, not true S2 sphere-correctness — that's the doc'd sidecar's job) stand as flagged for you.

open a new issue to track these

PR #41, opened 2026-06-15
Concurrency-aware Lambda orchestrator: pre-flight probe + FD-exhaustion guard

### LINKED ISSUE #28: runner.py lambda orchestrator: default max_workers exhausts client file descriptors, silently dropping cells
## Problem

`_run_lambda` (`src/zagg/runner.py:452`) fans out via `ThreadPoolExecutor(max_workers=max_workers)`, where each worker holds an open socket to the Lambda endpoint for a synchronous `invoke`. The lambda-backend default is `max_workers=1700` (and the CLI accepts arbitrary values).

On stock macOS / many Linux login shells the open-file limit (`ulimit -n`) defaults to **256**. Once concurrent workers exceed it, invocations fail with:

```
[Errno 24] Too many open files
Could not connect to the endpoint URL: ".../functions/process-shard/invocations"
```

This is a **client-side** failure — AWS never receives those cells, so they're dropped from the output while the run still "completes." In practice `--max-workers 900` lost ~430/1330 cells; even `--max-workers 500` silently caps at ~246 concurrent (≈ the 256 FD limit) if `ulimit` isn't raised.

## Why it's a footgun

- The default (`1700`) is far above any common `ulimit -n`, so the *default* invocation can fail on a fresh machine.
- The error reads like an AWS/network fault, not a local FD limit.
- Wall-clock is gated by the slowest single cell anyway, so high concurrency has little upside — the risky default buys nothing.

## Proposed fixes (some combination)

1. **Saner default** — clamp default `max_workers` to something safe on a 256 FD limit (e.g. 128–200), or derive from `resource.getrlimit(RLIMIT_NOFILE)` minus headroom.
2. **Bounded connection reuse** — share one boto3 client with `botocore.config.Config(max_pool_connections=...)` across threads so FD use is bounded regardless of worker count.
3. **Pre-flight** — read `ulimit -n` at startup; if `max_workers` exceeds it, raise the soft limit via `resource.setrlimit` (up to hard) or clamp + warn clearly.
4. **Clear error** — catch `OSError: [Errno 24]` and surface "raise ulimit -n or lower --max-workers" instead of the raw connection error.

## Workaround (documented)

`ulimit -n 8192` before running; keep `--max-workers ≤ min(ulimit -n − headroom, account concurrency)`. See `docs/deployment/lambda.md` → "Concurrency, workers, and file descriptors".

espg:
Closes #28

## What & approach

The Lambda backend fans out one synchronous `invoke` per cell across a `ThreadPoolExecutor`, each worker holding an open socket to the endpoint. Two limits could silently cap or drop that fan-out; this PR surfaces both **before** dispatch, per the settled plan in #28 ([Option A, standalone module](https://github.com/englacial/zagg/issues/28#issuecomment-4703992695)).

**New module `src/zagg/concurrency.py`** — carrier-agnostic probe helpers (take explicit boto3 clients, return plain numbers/dataclasses) so a future Step Functions task can import them without pulling in the in-process dispatch loop ([why standalone](https://github.com/englacial/zagg/issues/28#issuecomment-4703992695)):

- `fd_safe_max_workers()` — derives a worker ceiling from `resource.getrlimit(RLIMIT_NOFILE)` minus headroom (each worker holds one socket).
- `raise_for_fd_exhaustion(exc, max_workers)` — turns a raw `OSError [Errno 24]` into an actionable message (*"raise `ulimit -n` or lower `--max-workers`"*); non-errno-24 passes through untouched.
- `probe_concurrency(...)` — reads the account ceiling (`lambda:GetAccountSettings`) + current usage (CloudWatch `ConcurrentExecutions`, `Maximum` over a trailing window) and **reports** the function's reserved concurrency (`lambda:GetFunctionConcurrency`, informational — not the limiter). Padding = **5% of the ceiling, floored at ≥100 free slots** ([settled](https://github.com/englacial/zagg/issues/28#issuecomment-4703939896)). Every AWS call is behind a graceful `try/except` so missing IAM degrades the report rather than raising.
- `compute_available_workers(requested, ...)` — clamps to `min(requested, fd_ceiling, account_available)`; falls back to the FD ceiling alone when account concurrency is unreadable.

**Wiring in `runner.py` `_run_lambda`:**

- Pre-flight probe clamps `max_workers` before fan-out; `max_pool_connections` is sized to the clamped count so connections can't outrun the FD budget.
- The dispatch loop catches `OSError` from `future.result()` and routes it through `raise_for_fd_exhaustion` so any residual exhaustion is loud, not a silently dropped cell.
- `_log_concurrency_report` logs the account view + final worker count.

**Docs** (`docs/deployment/lambda.md`): added `ulimit -n 8192` to the example workflow, a new "Concurrency, workers, and file descriptors" section, and tightened the errno-24 troubleshooting note.

## Phases

- [x] **Phase 1** — `src/zagg/concurrency.py` probe logic (FD soft-limit → max_workers, boto3 capacity/concurrency reporting behind graceful try/except, errno-24 guidance helper) + unit tests (mocked boto3 + faked rlimit). `tests/test_concurrency.py`.
- [x] **Phase 2** — wire into `runner.py` lambda dispatch (bound default workers, surface FD exhaustion loudly) + wiring tests + docs. `tests/test_runner_concurrency.py`.

The broader roadmap in #28 (mid-run re-check, gated IAM `template.yaml` change) is deliberately **out of scope** here and left for follow-up — see Questions for review.

## How tested

All AWS is mocked — nothing touches live AWS. boto3 clients are `unittest.mock.MagicMock`; the FD soft limit is faked by monkeypatching `resource.getrlimit`.

- `tests/test_concurrency.py` (18 tests): FD ceiling derivation/floor, errno-24 message + chaining + non-24 no-op, padding math (pct vs floor), `available` floor, and graceful degradation when each of account-settings / CloudWatch / function-concurrency fails.
- `tests/test_runner_concurrency.py` (5 tests): probe clamps the `ThreadPoolExecutor` pool size; errno-24 in a future re-raised with guidance; `_log_concurrency_report` account/FD-fallback/reserved logging.
- Local: `ruff check src tests` clean; `pytest -v` → **230 passed, 1 skipped** (the 1 skip pre-existing). New module files are `ruff format`-clean and `mypy`-clean.

## Questions for review

1. **Scope.** This PR lands Phases 1–2 (the probe module + runner wiring + docs). The mid-run re-check and the gated `cloudwatch:GetMetricStatistics` IAM addition to `deployment/aws/template.yaml` are **not** included — the latter touches deploy infra (CLAUDE.md §1) and I'd rather confirm whether the deployed dispatch role already has `lambda:GetAccountSettings` / `cloudwatch:GetMetricStatistics` before opening that change. The probe degrades gracefully without them today. OK to track those as a follow-up, or do you want the IAM change in this PR?
2. **FD headroom** is a fixed 32 descriptors below the soft limit. Reasonable, or would you prefer a percentage?
3. **CloudWatch window** is `Maximum` over a trailing 5 min at 60s period. Confirmed `Maximum` per your note; the 5 min lookback is to ensure at least one datapoint despite the metric's ~1 min lag — acceptable?

---

### DIFF
--- a/docs/deployment/lambda.md
+++ b/docs/deployment/lambda.md
@@ -230,6 +230,12 @@ aws lambda update-function-code \
 ## Testing
 
 ```bash
+# Raise the open-file limit before fanning out: each concurrent worker holds
+# one socket to the Lambda endpoint, and the default soft limit (often 256)
+# would otherwise cap concurrency. See "Concurrency, workers, and file
+# descriptors" below.
+ulimit -n 8192
+
 # Build a shard map
 uv run python -m zagg.catalog --config atl06.yaml --short-name ATL06 --cycle 22 \
     --polygon antarctica.geojson
@@ -242,6 +248,31 @@ uv run python -m zagg --config atl06.yaml --catalog catalog.json \
 uv run python deployment/aws/invoke_lambda.py \
   --config atl06.yaml --catalog catalog.json --dry-run
 ```
+
+## Concurrency, workers, and file descriptors
+
+The Lambda backend fans out one synchronous `invoke` per cell across a thread
+pool, and each in-flight worker holds an open socket to the Lambda endpoint.
+Two limits bound how many can run at once, and the orchestrator checks both
+**before** dispatch so cells are never silently dropped:
+
+- **Open file descriptors (`ulimit -n`).** If concurrent workers exceed the
+  process's open-file soft limit (256 on stock macOS / many Linux shells),
+  invokes fail with `OSError: [Errno 24] Too many open files` — a client-side
+  failure AWS never sees. The runner derives a safe ceiling from the soft limit
+  and surfaces errno-24 with actionable guidance instead of a raw connection
+  error. Raise the limit before a large run: `ulimit -n 8192`.
+- **Account Lambda concurrency.** The runner reads the account
+  `ConcurrentExecutions` ceiling and current usage (CloudWatch) and clamps
+  workers to the available headroom (5% padding, floored at 100 free slots), so
+  a run can't saturate the account pool and throttle itself or other Lambda
+  activity. This degrades gracefully if the dispatch role lacks
+  `lambda:GetAccountSettings` / `cloudwatch:GetMetricStatistics` — it then
+  bounds workers by the FD limit alone.
+
+Keep `--max-workers ≤ min(ulimit -n − headroom, account concurrency)`. The
+orchestrator enforces this automatically; setting `ulimit -n` higher simply
+raises the FD ceiling it can use.
 
 ## Performance
 
@@ -272,4 +303,8 @@ uv run python deployment/aws/invoke_lambda.py \
     Check that the Lambda execution role has `s3:PutObject` permission for the output bucket.
 
 !!! warning "Too many open files"
-    Decrease max workers (e.g., `--max-workers 50`) or increase ulimit (`ulimit -n 10000`).
+    `[Errno 24] Too many open files` means concurrent workers exceeded the
+    open-file soft limit and cells would be dropped. Raise it (`ulimit -n 8192`)
+    or lower `--max-workers`. See "Concurrency, workers, and file descriptors"
+    above — the orchestrator now clamps workers to the FD and account-concurrency
+    limits automatically.
--- a/src/zagg/concurrency.py
+++ b/src/zagg/concurrency.py
@@ -1,0 +1,287 @@
+"""Pre-flight concurrency probe for the Lambda orchestrator.
+
+The Lambda backend fans out one synchronous ``invoke`` per cell across a
+``ThreadPoolExecutor``. Two independent limits can silently cap or drop that
+fan-out, and this module surfaces both *before* dispatch:
+
+* **Client file descriptors.** Each in-flight worker holds an open socket to
+  the Lambda endpoint. When concurrent workers exceed the process's open-file
+  soft limit (``ulimit -n``, often 256), invokes fail with
+  ``OSError: [Errno 24] Too many open files`` -- a client-side failure AWS
+  never sees, so those cells drop while the run still "completes".
+  :func:`fd_safe_max_workers` derives a worker ceiling from
+  ``RLIMIT_NOFILE``; :func:`raise_for_fd_exhaustion` turns a raw errno-24 into
+  an actionable message.
+
+* **Account Lambda concurrency.** Saturating the account-wide concurrent
+  execution pool throttles this run *and* any other Lambda activity in the
+  account. :func:`compute_available_workers` reads the account ceiling
+  (``lambda:GetAccountSettings``) and current usage (CloudWatch
+  ``ConcurrentExecutions``), pads it, and clamps the requested worker count.
+
+The boto3 helpers are deliberately carrier-agnostic -- they take explicit
+clients and return plain numbers/dataclasses -- so a future Step Functions
+task can import and call them without pulling in the in-process dispatch loop.
+IAM-dependent calls degrade gracefully: if the dispatch identity lacks
+``lambda:GetAccountSettings`` / ``cloudwatch:GetMetricStatistics`` /
+``lambda:GetFunctionConcurrency``, the probe falls back rather than failing.
+"""
+
+import logging
+import resource
+from dataclasses import dataclass
+from datetime import datetime, timedelta, timezone
+
+logger = logging.getLogger(__name__)
+
+# Open file descriptors held per worker beyond its endpoint socket (stdio,
+# the catalog handle, boto3 metadata sockets, ...). Subtracted from the soft
+# limit so the derived ceiling leaves room for the rest of the process.
+_FD_HEADROOM = 32
+
+# Account-concurrency padding: keep at least this percentage of the account
+# ceiling free (absorbs the ~1 min CloudWatch lag and other processes), with a
+# hard floor so small accounts still reserve a usable buffer. Settled on #28.
+_CONCURRENCY_PADDING_PCT = 0.05
+_CONCURRENCY_PADDING_FLOOR = 100
+
+# CloudWatch ConcurrentExecutions read window. The metric is emitted at 1 min
+# resolution; we take the Maximum over a short trailing window (settled on #28).
+_CW_PERIOD_S = 60
+_CW_LOOKBACK_S = 300
+
+
+def fd_safe_max_workers(headroom: int = _FD_HEADROOM) -> int:
+    """Largest worker count the open-file soft limit can safely sustain.
+
+    Each Lambda worker holds one open socket, so the number of concurrent
+    workers is bounded by ``RLIMIT_NOFILE`` (the ``ulimit -n`` soft limit)
+    minus headroom for the rest of the process.
+
+    Parameters
+    ----------
+    headroom : int
+        File descriptors to reserve for non-worker use (stdio, catalog,
+        boto3 metadata sockets). Default 32.
+
+    Returns
+    -------
+    int
+        ``soft_limit - headroom``, floored at 1.
+    """
+    soft, _hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    return max(1, soft - headroom)
+
+
+def raise_for_fd_exhaustion(exc: OSError, max_workers: int) -> None:
+    """Re-raise an errno-24 ``OSError`` with actionable guidance.
+
+    The raw error -- ``[Errno 24] Too many open files`` / "Could not connect
+    to the endpoint URL" -- reads like an AWS or network fault, not a local
+    file-descriptor limit. Catch it at dispatch and call this so the operator
+    knows to raise ``ulimit -n`` or lower ``--max-workers``. Non-errno-24
+    ``OSError`` is left untouched (returns without raising).
+
+    Parameters
+    ----------
+    exc : OSError
+        The caught error.
+    max_workers : int
+        The worker count in effect, for the message.
+
+    Raises
+    ------
+    OSError
+        A new errno-24 ``OSError`` chained to ``exc`` with guidance, when
+        ``exc`` is errno 24.
+    """
+    if exc.errno != 24:
+        return
+    soft, hard = resource.getrlimit(resource.RLIMIT_NOFILE)
+    raise OSError(
+        24,
+        f"Too many open files: {max_workers} Lambda workers exceeded the "
+        f"open-file soft limit (ulimit -n = {soft}, hard = {hard}). Cells "
+        f"would be silently dropped. Raise the limit (e.g. `ulimit -n 8192`) "
+        f"or lower --max-workers to <= {fd_safe_max_workers()}.",
+    ) from exc
+
+
+@dataclass
+class ConcurrencyReport:
+    """Pre-flight view of account Lambda concurrency headroom.
+
+    Attributes
+    ----------
+    account_limit : int or None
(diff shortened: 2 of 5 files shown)


espg (commented):
🤖 *from Claude (review)*

Adversarial review of the #28 concurrency probe. The probe/clamp math, graceful-degradation paths, and their tests are solid and well-covered. Two findings undercut element (e) of the settled plan ("surface FD exhaustion loudly, not silent cell drops"), plus a couple of minor items.

**1. CRITICAL — the FD guard is dead code in production (`src/zagg/runner.py:487-493`).**
The new guard catches `OSError` from `future.result()`:
```python
except OSError as e:
    raise_for_fd_exhaustion(e, max_workers)
    raise
```
But `_invoke_lambda_cell` wraps the entire `lambda_client.invoke(...)` in a blanket `except Exception as e: last_error = str(e)` (runner.py ~683) and **returns a normal result dict** (`status_code: None`) instead of propagating. So an errno-24 raised inside a real `.invoke()` is swallowed in the worker; `future.result()` returns that dict and never raises. The guard can only fire for an error raised *outside* `_invoke_lambda_cell` — which in practice never happens. The result is exactly the silent-drop behavior #28 set out to prevent: FD-exhausted cells become `status_code: None` error results, not a loud failure. `tests/test_runner_concurrency.py::test_errno_24_in_future_reraised_with_guidance` passes only because it monkeypatches `_invoke_lambda_cell` *itself* to raise, bypassing the real `except Exception`. Suggest: detect errno-24 inside `_invoke_lambda_cell`'s `except` (it's a non-retryable, run-fatal condition) and re-raise via `raise_for_fd_exhaustion` there, or have the worker re-raise FD errors so the dispatcher guard can see them — then add a test that exercises the *real* `_invoke_lambda_cell` swallow path.

**2. HIGH — errno-24 from boto likely isn't an `OSError` anyway (`src/zagg/runner.py:489`, `src/zagg/concurrency.py:141`).**
botocore wraps client-side socket failures in `botocore.exceptions.ConnectionError` / `EndpointConnectionError`, which subclass `BotoCoreError(Exception)` — **not** `OSError`. The "Could not connect to the endpoint URL" string the `raise_for_fd_exhaustion` docstring itself cites is a botocore `ConnectionError`. So even with finding 1 fixed, an `except OSError` (and `exc.errno != 24`) won't match the real-world FD-exhaustion failure; the raw socket `OSError` is buried as the `__cause__` of a botocore exception. Suggest catching the botocore connection exception (or inspecting the chained cause / message for errno 24 / "Too many open files") rather than relying on a bare `OSError` with `.errno == 24`.

**3. MEDIUM — stale worker count in the dispatch log (`src/zagg/runner.py:402`).**
`logger.info(f"Processing {len(cells)} ... (lambda, {max_workers} workers)")` runs *before* the clamp at line 435, so it prints the unclamped request (e.g. 1700) while the pool actually uses the clamped value. Either move that log after the clamp or drop the worker count from it (the new `_log_concurrency_report` already reports the effective count).

**4. LOW / nits.**
- `concurrency.py:163` magic literal `24` — fine, but `errno.EMFILE` would read clearer and match the docstring intent.
- `_log_concurrency_report` ordering (`runner.py:564`): when `account_limit is None` *and* `function_reserved` is set, it logs the reserved-concurrency INFO then the unreadable-account WARNING — slightly odd but harmless.
- `fd_safe_max_workers` is called a second time inside `raise_for_fd_exhaustion` (line 171) re-reading rlimit; minor, just noting the double read.

Not blocking, but findings 1+2 mean the headline "surfaces errno-24 with actionable guidance instead of a raw connection error" (docs + module docstring) is not actually achieved on the real dispatch path — worth resolving before this is called done. The account-concurrency clamp and FD *ceiling* (the proactive half) do work and are well tested.

---

espg:
🤖 *from Claude*

Folded in the self-review findings (commit `2f82db9`):

- **Finding 1 (critical) — FD guard was dead code.** The real `_invoke_lambda_cell` swallows all exceptions in its `except Exception` block and returns a `status_code: None` result, so an errno-24 never reached the `future.result()` guard — it became a silent dropped cell, exactly what #28 targets. Fixed by detecting FD exhaustion *inside* `_invoke_lambda_cell`'s except (it's run-fatal — every later cell hits it too) and re-raising via `raise_for_fd_exhaustion` before the swallow. The loop-level guard is kept as a backstop for exhaustion that surfaces at submit time. Added `tests/test_runner_concurrency.py::TestInvokeLambdaCellFdExhaustion` exercising the real swallow path (bare + botocore-wrapped re-raise, and that unrelated errors still return a result).
- **Finding 2 (high) — errno-24 isn't a top-level `OSError`.** botocore wraps client-side socket failures in `ConnectionError` ("Could not connect to the endpoint URL"), burying the real `OSError` in `__cause__`. Added `is_fd_exhaustion(exc)` which walks the `__cause__`/`__context__` chain (cycle-safe) for an `OSError` with `errno.EMFILE` or the "too many open files" message; `raise_for_fd_exhaustion` now takes `BaseException` and uses it. Tests cover bare EMFILE, cause-chain wrapping, message-only match, unrelated errors, and a self-referential chain.
- **Finding 3 (medium) — stale worker count in log.** The pre-dispatch "Processing N cells (lambda, M workers)" line printed the unclamped request. Dropped the count there; the post-clamp count is logged by `_log_concurrency_report`.
- **Finding 4 (nits)** — switched the magic `24` to `errno.EMFILE`; `fd_safe_max_workers(soft=...)` now accepts the already-read soft limit to avoid a redundant `getrlimit`.

Local: `ruff check src tests` clean, new files `ruff format`/`mypy`-clean, `pytest` → 239 passed, 1 skipped (pre-existing). The remaining mypy notes (`boto3`/`botocore` import-untyped in `runner.py`) are pre-existing on `main`, not from this diff.


---

espg (commented):
🤖 *from Claude (review)*

Second pass. The four prior findings are correctly fixed; one CI-blocking format issue and one low-severity double-wrap remain, plus two minor notes.

**Prior findings — verified fixed**
1. FD guard no longer dead: `raise_for_fd_exhaustion(e, max_workers)` is now the first statement in the `except` of `_invoke_lambda_cell` (runner.py:710), before `last_error = str(e)`, so it runs on every attempt before the swallow. ✅
2. botocore-wrapped errno-24: `is_fd_exhaustion` walks `__cause__`/`__context__` and matches both `OSError`/`EMFILE` and the message text; cycle-guarded with a `seen` id-set (concurrency.py:156-165). ✅
3. Stale worker count gone from the pre-clamp log (runner.py:404); clamped count logged in `_log_concurrency_report`. ✅
4. Magic `24` → `errno.EMFILE` throughout. ✅

---

**Findings**

**[CI-blocking] `ruff format --check` fails on tests/test_runner_concurrency.py**
`lint.yml` runs `ruff format --check`; this file is not formatted (the dict/`monkeypatch.setattr` calls and `ConcurrencyReport(...)` kwargs aren't expanded, and two inline `import` statements lack the blank line after). `ruff check` passes, but format does not — the draft will go red. Fix: `ruff format tests/test_runner_concurrency.py`. (concurrency.py and test_concurrency.py are already clean.)

**[Low] Double-wrap of the guidance OSError at the loop backstop**
runner.py:444-452 — when `_invoke_lambda_cell` already re-raised the guidance `OSError(EMFILE, ...)`, `future.result()` surfaces *that* OSError, and the backstop calls `raise_for_fd_exhaustion(e, max_workers)` on it again. `is_fd_exhaustion` returns True (it's EMFILE), so it wraps an already-wrapped error in a second guidance OSError chained to the first — duplicated guidance text in the chain. `errno` and the actionable message are preserved, so functionally fine, just noisy. Optional fix: skip re-wrapping if `e` is already our `OSError` with `errno == EMFILE` (e.g. `if not (isinstance(e, OSError) and e.errno == errno.EMFILE): raise_for_fd_exhaustion(e, max_workers)`), or accept the minor nesting. Not blocking.

**[Minor] `max_workers=None` default yields "None Lambda workers" in the message**
runner.py:637 / concurrency.py:197 — `_invoke_lambda_cell(..., max_workers=None)` is the default; if ever called without it, the guidance reads "None Lambda workers exceeded…". All production/test call sites pass it, so cosmetic only. Consider defaulting the message to a generic phrasing when `max_workers` is None.

**[Minor] Two `boto3.Session()` instances**
runner.py:434 (probe) and runner.py:451 (dispatch client) create separate sessions. Harmless, but a single `session` reused for both would be tidier and matches the "lightweight session" comment.

---

**Things I checked that are correct**
- Probe math: padding `max(floor, pct*limit)`, `available = max(1, limit - current - padding)`, FD fallback when `account_limit is None`, final `max(1, min(requested, fd, available))` — all correct and tested (test_concurrency.py covers floor-dominates, pct-dominates, over-subscribed→1, each AWS read degrading independently).
- `raise_for_fd_exhaustion` is a true no-op on non-FD errors, so retries for `TooManyRequestsException`/timeouts are unaffected.
- `test_unrelated_error_still_returns_result` exercises the *real* (non-monkeypatched) `_invoke_lambda_cell` swallow path with a mocked client — genuinely guards the regression.
- All AWS is mocked (`MagicMock` clients, `boto3.Session` stubbed, `resource.getrlimit` monkeypatched); nothing touches live AWS.
- No new mypy errors from the diff (untyped boto3 client params match existing repo style); `ruff check` clean.

One narrow false-positive note (not a finding): the message-text branch of `is_fd_exhaustion` could match a botocore exception whose stringified message coincidentally contains "too many open files" from the remote side. Lambda `FunctionError`s don't raise (they go through the `continue`/`last_error` path), so the exposure is tiny — fine as-is.

Net: address the ruff-format failure before the draft can go green; the double-wrap and the two minor notes are optional.

---

espg:
🤖 *from Claude*

Phases 1–2 complete; both fresh-context self-review passes folded in:

- **Pass 1 (`2f82db9`):** fixed the two findings that undercut the "surface FD exhaustion loudly" goal — (1) the FD guard was effectively dead code because `_invoke_lambda_cell` swallowed the `OSError` before the dispatcher could see it (now `raise_for_fd_exhaustion` runs first inside the worker's `except`); (2) botocore wraps client-side socket errors in `ConnectionError`, not `OSError`, so errno-24 wasn't matching — `is_fd_exhaustion` now walks `__cause__`/`__context__` (cycle-guarded) and matches `EMFILE`/message text. Also fixed the stale pre-clamp worker-count log and switched the magic `24` → `errno.EMFILE`. Added a test that drives the *real* (non-monkeypatched) swallow path.
- **Pass 2 (`8a6a71c`):** `ruff format` on the test file (the one CI-blocking item) + reuse a single `boto3.Session`. The double-wrap and `max_workers=None`-in-message notes were judged cosmetic and left.

**CI:** `ruff`, `test (3.12)`, `test (3.13)`, `build` all green; the dual-arch Lambda-layer builds were still running at stop (no new deps added, so no layer-size impact expected). Local: 230 passed, 1 skipped; all AWS mocked.

**Open for you (Question 1 in the body):** the mid-run re-check and the `cloudwatch:GetMetricStatistics` IAM grant in `deployment/aws/template.yaml` are out of scope here — the latter touches deploy infra (§1) and the probe degrades gracefully without it today. Track as a follow-up, or do you want the IAM change in this PR? Marking `waiting` on that + your merge.

---

espg:
> Open for you (Question 1 in the body): the mid-run re-check and the cloudwatch:GetMetricStatistics IAM grant in deployment/aws/template.yaml are out of scope here — the latter touches deploy infra (§1) and the probe degrades gracefully without it today. Track as a follow-up, or do you want the IAM change in this PR? Marking waiting on that + your merge.

track as a follow up

PR #42, opened 2026-06-16
rename parent_morton event field to shard_key (#24)

### LINKED ISSUE #24: Reconcile Catalog API
The entire catalog API needs a refresh. Right now we have a single `build_catalog` function that both fetches a subset from NASA CMR, and then also builds a shard map. This works for the morton backend because we specify the `order` level in `build_catalog`, but we we don't *have* an `order` parameter for arbitrary grids. It's also debatable if it actually does work for the morton case; it assumes that the morton order specified in the yml template matches what's provided to `build_catalog`, but this isn't actually checked or enforced.

What we really need is a separation of concerns for two related tasks:

  1. Cloning / fetching the CMR (or equivalent STAC endpoint).
  2. Building the shard map

Inputs for (1) is a query of what, when, and where -- which data product, over what time period, and over what region. The output of (1) is either on memory pyarrow (we'll need support for the geometry column, which is assumed to be either WKT or WKB), or, a 'stac catalog' on disk in the form of a geoparquet file.

Input for (2) is the output from (1), plus grid specifications... which could (but don't strictly have to) come from the yml template that we use to define the aggregator.

While (1) and (2) should be distinct functions (and likely classes) to keep the architecture clean, we can of course provide a convenience function that executes them together in sequence. Thinking about how a user would interact with this system, something to keep in mind is where this is going in future-- **multiple** data products that are aggregated to *compatible* data cubes / grids. Compatible is important here; we may have different resolutions in both the spatial and temporal sense, but the grids should be able to nest together cleanly-- i.e., be aligned and whole number factors for nesting. So in the future, a user would likely provide to a new function (3) a list of tuples containing pairs of `Catalog` and `ShardMap` objects, or, containing pairs of `Catalog` and yml aggregation templates. The job of function (3) would be to *validate* that tuples are compatible *across* the input list-- that is that they conform to the same spatial coverage (they overlap), and that they comprise of resolutions that can nest, and grids that are aligned. This would forbid *mixing* morton / healpix grids with arbitrary grids, and would require multiple arbitrary grids in a given list to have the same projection and resolutions defined that can nest cleanly.

The big question that remains open is *how* do we encapsulate grid specs, and is that encapsulation separate from defining the region? My sense is that the grid specs need to *be allowed* to change from aggregator to aggregator-- this fits well with *requiring* the yml templates to define grid specs. The reason for this is because mix resolution aggregations (which again, will be required to be compatible), will likely need to define different chunking schemes for the output arrays. It's even possible (in the *far* future), that different aggregators may even output to different types of files (i.e., parquet, or geotiff) at different locations... although this certainly isn't what we're designing for in this current beta. The converse approach is to *not* have grid specs in the aggregator templates-- or perhaps more specifically, to let them be either optional, or something that is inheritable. The thinking here is that the most *common* case is that we actually do want the same grid specs for multiple data products; so it would be nice to define the grid specs once, and then over-ride them as they vary. 

Regardless of what (3) looks like, right now we need to pin down a separation for spatiotemporal metadata fetch, and the processing of that metadata into a manifest that can be used to distribute work to workers.

espg:
Closes #24

## What this does

Hard-renames the grid-neutral shard identifier in the Lambda **event payload** from the HEALPix-flavored `parent_morton` to `shard_key`, and makes the HEALPix-specific `child_order` **optional for non-HEALPix grids** (still required/forwarded for HEALPix). No back-compat aliases — a clean break, per @espg.

### Checklist
- [x] Rename `parent_morton` → `shard_key` end-to-end (event field + the `ProcessingMetadata` contract + handler/runner logging & error responses)
- [x] Make `child_order` optional for non-HEALPix (mirror existing `parent_order` handling); required for HEALPix
- [x] Update the ATL templates *if needed* (determination below)
- [x] Update all other `parent_morton` references (docs, auth example)
- [x] Tests for the new event shape (HEALPix with `child_order`, rectilinear without; gate accepts `shard_key`, rejects missing / legacy field)

## §1 by-name authorization (deployment/aws/)

CLAUDE.md §1 forbids editing `deployment/aws/` unless an issue names it. @espg explicitly authorized editing `deployment/aws/lambda_handler.py` by name and asked that the handler and runner be kept in sync: https://github.com/englacial/zagg/issues/24#issuecomment-4713387130 (*"yes, edit the lambda_handler; I want the lambda_handler and runner kept in sync for this edit"* and *"hard rename"*).

## Approach

Kept `lambda_handler.py` and `runner.py` in lockstep on the wire contract:

- **`deployment/aws/lambda_handler.py`** — event field `parent_morton` → `shard_key` (docstring, structured logging, the required-params gate, dispatch into `process_shard`, and both error/result payloads). `child_order` dropped from the unconditional required list and instead required only when the resolved config's grid type is `healpix`:
  ```python
  grid_type = config.output.get("grid", {}).get("type", "healpix")
  if grid_type == "healpix" and "child_order" not in event:
      return {"statusCode": 400, ...}  # Missing required parameters: child_order
  ```
  `parent_order` stays in the base required list (the orchestrator always sends the key — `None` for non-HEALPix), and the dispatch now reads `event.get("parent_order")` (was a hard index), matching the setup path.

- **`src/zagg/runner.py`** — `agg()` now computes `child_order` only for HEALPix (`get_child_order(config) if grid_type == "healpix" else None`), so a rectilinear run no longer raises `output.grid.child_order is required`. `_invoke_lambda_cell` emits `shard_key` (was `parent_morton`) and only includes `child_order` in the event when it is set; its result dict key `morton` → `shard_key` (and the one consumer log line).

- **`src/zagg/schema.py` + `src/zagg/processing.py`** — the `ProcessingMetadata` TypedDict field and the `process_shard` metadata key are renamed `parent_morton` → `shard_key` (the `process_shard` docstring already advertised `shard_key`, so this resolves a latent mismatch the handler read across at `metadata["parent_morton"]`). The internal local/param in `process_shard`/`_read_group` is renamed too, since the function is already grid-agnostic (`shard_key` is its parameter). The deprecated, explicitly HEALPix-flavored `process_morton_cell(parent_morton=...)` keeps its `parent_morton` name intentionally.

- **`src/zagg/auth.py`, `docs/deployment/lambda.md`** — event-payload examples/tables updated; the lambda doc's "rename tracked in #24, not done here" note is rewritten to describe the landed grid-neutral schema.

## ATL templates: no change needed

@espg asked whether the ATL config templates need updating. **They do not:**
- `parent_morton` never appears in `src/zagg/configs/atl06.yaml` or `atl06_polar.yaml` — it's purely an *event* field, never a config field.
- The `child_order` in `atl06.yaml` is the **HEALPix grid-spec** field (`output.grid.child_order: 12`), which is legitimate and stays. `atl06_polar.yaml` is a **rectilinear** grid and already has no `child_order` — which is exactly the case the optional-`child_order` change now supports end-to-end.

What was *not* touched (deliberately, out of scope): `docs/design/architecture.md` (internal morton algorithm prose), `tests/conftest.py` (a local morton variable in a fixture), `docs/index.md` (a `process_morton_cell(parent_morton=...)` call — the kept legacy API), and the `morton` DataFrame coordinate column (a real grid coord, unrelated to the event field).

## How tested

- New `tests/test_lambda_handler.py` (loads the handler by path): missing `shard_key` → 400; legacy `parent_morton` no longer accepted → 400; HEALPix event missing `child_order` → 400; HEALPix event (with `child_order`) and rectilinear event (no `child_order`, `parent_order=None`) both dispatch and the shard key reaches `process_shard`.
- New `TestInvokeLambdaCellEvent` in `tests/test_runner.py`: the per-cell event built by `_invoke_lambda_cell` carries `shard_key` (no `parent_morton`); `child_order` present for HEALPix, omitted when `None`.
- Updated the `_invoke_lambda_cell` stub in `tests/test_runner_concurrency.py` to the new `shard_key` result key.
- Full suite: `uv run pytest -v` → **256 passed, 1 skipped** (pre-existing skip). `uv run ruff check src tests` → clean. mypy on the changed source files → **0 new errors** vs the `main` baseline (6 pre-existing, all missing-stub / unrelated).

## Questions for review

- **`ruff format --check` drift (pre-existing, not from this PR).** On `main` HEAD (`7cabbfc`), `ruff format --check` already reports `processing.py`, `runner.py`, `schema.py`, `lambda_handler.py`, and `test_runner.py` as "would reformat" — true with both the env's ruff and the CI-pinned `v0.14.10`. I verified this against a clean checkout *before* my edits, so it's standing drift, not something I introduced. Per the "don't fix unrelated pre-existing CI failures" rule I did **not** mass-reformat those files; my added lines follow the surrounding style and my new `test_lambda_handler.py` is already `ruff-format`-clean. Flagging in case you'd like a separate format-only sweep.
- **`child_order` still sent in setup-mode events.** I scoped the optional/omit behavior to the per-cell *process* event (`_invoke_lambda_cell`). The setup event still includes `child_order` (value `None` for non-HEALPix); the setup handler only reads it inside the `grid_type == "healpix"` branch, so `None` is harmless there. Say the word if you'd like setup made symmetric (omit-when-None) too.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TqvW9p9SYsoxHNNpRsNsgv)_

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -6,9 +6,9 @@ This is an AWS-specific wrapper around the cloud-agnostic processing module.
 Event payload (default / process mode):
 {
     "chunk_idx": int,
-    "parent_morton": int,
-    "parent_order": int,
-    "child_order": int,
+    "shard_key": int,           # grid-agnostic shard identifier
+    "parent_order": int,        # HEALPix only (omit for other grids)
+    "child_order": int,         # HEALPix only (omit for other grids)
     "granule_urls": [str, ...],
     "store_path": str,          # e.g. "s3://bucket/prefix.zarr"
     "s3_credentials": {         # creds for reading NSIDC source data
@@ -182,7 +182,7 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         json.dumps(
             {
                 "event_type": "lambda_invocation",
-                "parent_morton": event.get("parent_morton"),
+                "shard_key": event.get("shard_key"),
                 "granule_count": len(event.get("granule_urls", [])),
                 "child_order": event.get("child_order"),
                 "request_id": context.aws_request_id,
@@ -192,11 +192,13 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     )
 
     try:
-        # Validate required parameters
+        # Validate required parameters. ``child_order`` is HEALPix-specific and
+        # only required once the grid is known to be HEALPix (checked below);
+        # ``parent_order`` is forwarded by the orchestrator for every grid (None
+        # for non-HEALPix), so its key is always present.
         required_params = [
-            "parent_morton",
+            "shard_key",
             "parent_order",
-            "child_order",
             "granule_urls",
             "store_path",
             "s3_credentials",
@@ -228,13 +230,22 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         if config is None:
             from zagg.config import default_config
             config = default_config("atl06")
-        grid = from_config(config, parent_order=event["parent_order"])
+
+        # child_order is required for HEALPix runs (drives the leaf order); it is
+        # absent/unused for non-HEALPix grids.
+        grid_type = config.output.get("grid", {}).get("type", "healpix")
+        if grid_type == "healpix" and "child_order" not in event:
+            error_msg = "Missing required parameters: child_order"
+            logger.error(error_msg)
+            return {"statusCode": 400, "body": json.dumps({"error": error_msg})}
+
+        grid = from_config(config, parent_order=event.get("parent_order"))
 
         # Process the shard using cloud-agnostic function
         from zagg.processing import process_shard
         df_out, metadata = process_shard(
             grid,
-            event["parent_morton"],
+            event["shard_key"],
             event["granule_urls"],
             s3_credentials=s3_creds,
             config=config,
@@ -274,7 +285,7 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
             json.dumps(
                 {
                     "event_type": "processing_complete",
-                    "parent_morton": metadata["parent_morton"],
+                    "shard_key": metadata["shard_key"],
                     "cells_with_data": metadata["cells_with_data"],
                     "total_obs": metadata["total_obs"],
                     "duration_s": metadata["duration_s"],
@@ -302,7 +313,7 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
             "body": json.dumps(
                 {
                     "error": f"Unhandled exception: {str(e)}",
-                    "parent_morton": event.get("parent_morton"),
+                    "shard_key": event.get("shard_key"),
                     "request_id": context.aws_request_id,
                 }
             ),
--- a/docs/deployment/lambda.md
+++ b/docs/deployment/lambda.md
@@ -47,7 +47,7 @@ The Lambda function processes a single morton cell (order 6) by:
 
 ```json
 {
-  "parent_morton": 123456,
+  "shard_key": 123456,
   "parent_order": 6,
   "child_order": 12,
   "granule_urls": [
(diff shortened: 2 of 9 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of `cb35564`. Traced the data flow through `from_config` → `process_shard` and both the runner and handler gates, and ran the changed tests.

**No blocking defects. The hard rename is complete and the `child_order`-optional logic is correct and symmetric.**

What I verified:
- **No stray `parent_morton` event-field references.** Remaining hits are all intentional: `processing.py` (the deprecated `process_morton_cell` param), `tests/conftest.py` (local morton var), `docs/index.md:84` (a `process_morton_cell(parent_morton=...)` call — the kept legacy API, not an event), and `docs/design/architecture.md:125` (internal morton-algorithm prose). None are event-field usages.
- **`child_order` gate is symmetric.** Runner (`agg`) and handler (`_handle_process`) both key off `config.output.get("grid", {}).get("type", "healpix")`. A HEALPix run cannot slip through without `child_order` (runner's `get_child_order` raises; handler independently re-gates). Rectilinear is not wrongly rejected (runner sets `child_order=None` and omits the field; handler gate only fires for healpix). Confirmed `atl06`→healpix, `atl06_polar`→rectilinear.
- **`event.get("parent_order")` introduces no None bug.** `from_config` prefers `grid_cfg["parent_order"]` from config and only falls back to the kwarg for HEALPix; rectilinear ignores it, so `None` is harmless. `parent_order` stays required (key always sent by the orchestrator).
- **Result-key rename `morton` → `shard_key`** has exactly one consumer (the per-cell error-warning log), correctly updated; the concurrency-test stub was updated to match. `metadata["shard_key"]` matches the renamed `ProcessingMetadata` field.
- **Tests genuinely assert the rename** — legacy `parent_morton` → 400, missing `shard_key` → 400, HEALPix `child_order` gate, and both dispatch shapes.

Standing (non-blocking, already disclosed in the PR body):
- The **setup-mode** event still sends `child_order` unconditionally (None for rectilinear). Out of scope for this process-mode rename and harmless (the setup handler reads it only inside the `healpix` branch). Left for a follow-up if/when setup mode is generalized to non-HEALPix.
- The pre-existing `ruff format` drift on the touched modules (present on `main` before this PR) is left untouched per "don't fix unrelated CI failures."

---

espg:
🤖 *from Claude*

Status: complete and green — single-phase small-fix, all phases done. CI: `ruff`, `test (3.12)`, `test (3.13)`, and `build` all passing (arm64 layer build still finishing; deploy/pr-preview skipped on the draft). Self-review found no blocking defects. Marking `waiting` for your review/merge.

Two non-blocking items are flagged under "Questions for review" in the PR body: (1) pre-existing `ruff format` drift on the touched modules (present on `main` before this PR — left untouched per the "don't fix unrelated CI failures" rule; the `ruff` *check* gate is green), and (2) whether you want the setup-mode event made symmetric (omit `child_order` when `None`) as a follow-up. And to answer your explicit question: **the ATL templates did not need changes** — `parent_morton` never appears in them (it's an event field only), and `child_order` in `atl06.yaml` is the legitimate HEALPix grid-spec field while the rectilinear `atl06_polar.yaml` already omits it.

---

espg:
🤖 *from Claude*

Resolved the merge conflict — the branch is no longer `dirty`.

**Cause:** PR #33 (the #30 sort/hash grouping + Arrow handoff + experimental kernel reducer) merged to `main` after this branch was cut, and it rewrote the same `process_shard`/`_read_group` region of `src/zagg/processing.py` that this PR renames. `processing.py` was the only conflicting file.

**How resolved:** merged `origin/main` into the branch (not rebased — §1 forbids force-push) in commit `48c13dc`. Took `main`'s version of `processing.py` wholesale (so all of #33's grouping/arrow/kernel/nan-contract work is preserved intact) and re-applied **only** this PR's rename onto it: the internal `parent_morton` local/param + the `"parent_morton"` metadata key → `shard_key`, across `_read_group` (kept its `arrow=` param from #33) and `process_shard` (log strings "morton"→"shard" too). The deprecated `process_morton_cell(parent_morton=…)` keeps its name, as before. No other file conflicted.

**Verified after merge:** full suite **272 passed, 1 skipped** (the spherely-fork skip), `ruff check src tests` clean. The metadata key now written by `process_shard` (`"shard_key"`) lines up with the renamed `ProcessingMetadata` field and the handler's `metadata["shard_key"]` read.

Still `waiting` on your merge. (Note: the pre-existing `ruff format --check` whole-file drift flagged in the PR body is unchanged — not introduced here.)

---

PR #44, opened 2026-06-16
shardmap viewer: headless render core + ipyleaflet wrapper

### LINKED ISSUE #38: shardmap viewer
We should have a shard map viewer. This would ideally be a jupyter lab widget (or extension) that takes as input either shard map json file , or a shard map json file + stac geoparquet file. This widget would visualize the chunking of the, i.e., the shards & the geometry footprints of the input data set to be aggregated, overlaid on an Interactive  context background map . On zoom to an appropriate zoom level this would also show the grid,  In addition to the chunk outlines.  Any relevant questions need to implement.

espg:
Closes #38

A **minimal, fast** shard-map viewer: a headless GeoJSON render core plus a thin `ipyleaflet` wrapper. Both inputs (ShardMap JSON and the STAC-geoparquet `Catalog`) are supported from day one, and the geometry is reused straight off the existing `grid.shard_footprint` / `signature` surface — no viewer-specific file type or second tessellation.

## Scope narrowed (per @espg, [2026-06-23](https://github.com/englacial/zagg/pull/44#issuecomment-4774889139))

The cell-level grid rendering never produced a usable grid and was the source of the kernel hangs/complexity. It has been **removed entirely**. The viewer now focuses on three things only:

1. **Shard outlines** (the shardmap)
2. **Granule footprints**
3. **Context basemap with auto polar-projection switching** (`pick_crs` → EPSG:3031/3413/3857, GIBS polar basemaps)

### Removed surface
- `viewport_cells`, the HEALPix child-cell enumeration (`_healpix_child_cells`, `_child_cell_polygon`, viewport-bounded `morton_coverage`), the `max_shards` / `max_cells` grid-on-zoom gate.
- The `bounds` observer + debounce machinery in `leaflet.py` (`_debounce`, `loop.call_later`, `cancel_grid_refresh`, `_refresh_grid`) and the grid layer in `show_shardmap` — all of it existed only to drive the now-deleted grid.
- The `STRtree`-backed `ShardIndex` / `shard_index` / `_INDEX_CACHE` machinery (only used to speed up the grid refresh; shard outlines build footprints directly).
- `_seam_safe_polygon` / `_polygonal` (only used by the deleted cell-clip path). `_split_antimeridian` is **kept** — `shard_outlines` and `granule_footprints` still need it.

Net: `src/zagg/viz/shardmap.py` and `leaflet.py` are substantially smaller; the module is minimal and fast (no per-pan recompute, no background-thread comm mutation).

## What this does

**Headless render core** (`src/zagg/viz/__init__.py`, `src/zagg/viz/shardmap.py`) — pure Python, no browser/ipyleaflet import:
- `shard_outlines(shardmap)` — one feature per shard, off `grid.shard_footprint(key)`; grid rebuilt from the map's own `grid_signature` via `grid_from_signature` (HEALPix + rectilinear round-trip exactly).
- `granule_footprints(catalog)` — one feature per granule footprint, via `Catalog.granule_records()`.
- `render_shardmap(...)` — assembles `{"shards", "granules"}`.
- Antimeridian: genuine ±180 seam crossings are cut into a `MultiPolygon` by `_split_antimeridian` (unwrap +360 / clip / rewrap), shapely-only; skipped under a polar CRS (no seam there).

**ipyleaflet wrapper** (`src/zagg/viz/leaflet.py`, exposed as `zagg.viz.show_shardmap`) — context basemap (auto polar projection + GIBS, else OSM/Web Mercator) + shard layer + a toggleable granule-footprint layer + a `LayersControl`. All `ipyleaflet` imports are lazy/local so importing `zagg.viz` and the test suite never require the widget stack.

**CRS picker** (`src/zagg/viz/crs.py`) — the valued part, unchanged: `pick_crs` selects EPSG:3031 (Antarctic), EPSG:3413 (Arctic), or EPSG:3857 (mid-latitude/global) from the map extent; `crs_info` carries the proj4leaflet definition + GIBS WMTS basemap per polar code.

## Phases
- [x] Minimal viewer — drop cell-grid rendering (shard outlines + granule footprints + context basemap with auto polar projection only)

## How tested
- `uv run pytest tests/test_viz.py -q` — **31 passed, 5 skipped** (the 5 skips are the `pytest.importorskip("ipyleaflet")` wrapper tests; `ipyleaflet` isn't installed in this env). Tests for the removed grid/viewport/debounce/`ShardIndex` code were deleted; the kept tests cover `grid_from_signature` round-trips, shard outlines (count/props/WGS84/JSON), granule footprints, antimeridian splitting, CRS picking, the seam flag, and the wrapper smoke tests.
- `uv run ruff check --select=E,F,W,I --ignore=E501 src tests` — clean (matches the PR lint bot).
- `ruff format --check` clean on the files I touched.
- Notebook (`notebooks/shardmap_viewer.ipynb`) updated: grid-on-zoom and zoom-to-shard demo cells removed; the real anonymous CMR-STAC ATL06 queries (`version="007"`, two AOIs) and the polar-projection demo are kept; outputs cleared.

Note: `tests/test_integration.py` / `tests/test_processing.py` have 3 pre-existing failures (morton/mortie, fixed on `main` after this branch's base) — unrelated to this change, not touched.

## Questions for review
- Browser verification of the live ipyleaflet map is still deferred (constructs + layers attach headlessly; in-browser rendering can't be verified here).


espg:
🤖 *from Claude*

Opened this draft PR for #38 — phases 1 and 2 both landed; phase 3 (docs/notebook) is deferred to the next run per the plan.

**Landed**
- **Phase 1** (`src/zagg/viz/shardmap.py`): headless GeoJSON render core — `shard_outlines`, `granule_footprints`, `viewport_cells` (grid-on-zoom gate, ≤4 visible shards, never a global graticule), `render_shardmap`, and `grid_from_signature` (rebuilds the grid from the map's own `grid_signature`). Antimeridian crossings split into `MultiPolygon`, shapely-only.
- **Phase 2** (`src/zagg/viz/leaflet.py` → `zagg.viz.show_shardmap`): basemap + shard layer + toggleable footprint layer + zoom-thresholded grid layer + `LayersControl`. All `ipyleaflet` imports are lazy/local, so phase-1 core and the test suite never require it (verified by blocking the import).
- New **`viz`** optional extra (`pip install zagg[viz]`, `ipyleaflet>=0.19`), out of core/`lambda`. Approval: [“sure, viz is fine”](https://github.com/englacial/zagg/issues/38#issuecomment-4713639466).

**Tested:** `uv run pytest -v` → 289 passed, 1 skipped (pre-existing spherely-fork skip). `ruff check src tests` clean. `uv sync --extra viz` resolves; **ipyleaflet 0.20 installed in-env, so the phase-2 Map-construction smoke test ran here** (not skipped). Browser rendering itself can't be verified headlessly — deferred to a notebook check (phase 3).

**Self-review folded** (commit `fold review: antimeridian seam-jump detection, keep holes`): the antimeridian gate now keys on a true consecutive-vertex seam jump rather than total longitude span (a wide-but-continuous swath was being over-split), and interior rings (holes) are carried through the split.

Two items raised under *Questions for review* in the PR body: deferred browser verification, and the `ruff format` choice (I matched the repo's hand-formatted style since existing source isn't `ruff format`-clean and the lint bot is `ruff check` only). Next run: phase 3 docs/notebook.

---

espg (commented):
🤖 *from Claude (review)*

Fresh adversarial self-review of the phase-1/phase-2 diff. Two real defects were found and **folded** before opening:

1. `_split_antimeridian` originally gated on total longitude span (`max-min > 180`), which over-split a wide-but-continuous swath (e.g. a granule footprint stepping -170 → 0 → +170) into ±180 slivers, discarding the real interior. Now keyed on an actual consecutive-vertex seam jump (`_crosses_antimeridian`). Live exposure was `granule_footprints`, since catalog footprints carry no narrow-cell guarantee. Covered by `test_wide_non_crossing_polygon_kept_intact`.
2. Interior rings (holes) were dropped on the exterior-only path. Now carried through the unwrap/clip. Covered by `test_holes_preserved`.

Verified fine (no change needed): import safety with `ipyleaflet` absent, the `<= max_shards` gate at its boundary, `grid_from_signature` rectilinear affine→bounds math (incl. anisotropic resolution), and the `MultiPolygon` assembly nesting.

**Standing (not blocking, for @espg):** `viewport_cells` drops a clip that degenerates to a point/line on a shared shard edge (touch-only intersection). That's the right default for a viewer, but flagging in case you want edge-touch shards surfaced in the grid layer.

---

espg:
> Browser verification deferred

...instructions on to verify this? Are we doing an example ipython notebook (would be the natural place for instructions).

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **phase 3 only** (commit `69b5301`): the new `notebooks/shardmap_viewer.ipynb` and the 7-line `docs/quickstart.md` pointer. Findings are in the summary because the notebook is JSON and inline anchoring would be noisy; cell numbers are 0-indexed as in the file.

**Verdict: looks good. No blocking or diff-scoped findings.** The notebook runs offline on synthetic fixtures, the public API calls are all correct, and the verification instructions match the code. Two trivial nits below, both optional.

**API usage — all correct (checked against `src/zagg/viz/`):**
- `render_shardmap(shardmap, catalog)` (cell 5) → returns keys `shards`/`granules`/`cells`; cell 6 reads `layers["shards"]` and cell 7 reads `layers["granules"]` — both match `shardmap.py:346`. With no `bbox`, `cells` is `None`, and the cell-5 dict-comp guards with `fc and len(...)`, so it won't `NoneType`-subscript. Good.
- `shard_outlines`, `granule_footprints`, `viewport_cells`, `grid_from_signature`, `show_shardmap` (cells 6/7/9/11/13) — signatures and import paths (`from zagg.viz import ...`, `from zagg.viz.shardmap import grid_from_signature`) all match `__init__.py`'s `__all__`.
- `ShardMap(grid.signature(), shard_keys, granules, {...})` (cell 2) and `Catalog(pa.table(...), {...})` (cell 3) match the dataclass field order in `shardmap.py:243` / `sources.py:222`. `RectilinearGrid(crs, res, bounds, chunk_shape)` and `HealpixGrid(2, 6, layout=...)` match the constructors. These mirror `tests/test_viz.py` fixtures exactly, which is the right thing to copy.
- `viewport_cells(shardmap, (lon0, lat0, lon1, lat1), max_shards=...)` (cell 9) uses `shard_footprint(0).bounds` → `(minx, miny, maxx, maxy)` = `(lon_min, lat_min, lon_max, lat_max)`, the order `viewport_cells` expects. Correct.

**Offline / no-network (req 1):** confirmed. Everything is built in-memory from `RectilinearGrid` + hand-written STAC dicts → `stac_geoparquet.arrow`; the only disk I/O is a `tempfile.mkdtemp()` round-trip in cell 13. No S3/Earthdata/CMR calls.

**Verification instructions (req 3):** accurate. Layer names in cell 14 — `shards`, `granule footprints`, `grid (shard cells)` — match `leaflet.py:93/99/109`. The grid-on-zoom description matches the `max_shards` gate wired to `bounds` (`leaflet.py:113-123`), and "layer control (top-right)" matches `LayersControl(position="topright")` (`leaflet.py:125`). The styling words ("blue" shards / "red" footprints / "grey" grid) match `_SHARD_STYLE`/`_FOOTPRINT_STYLE`/`_GRID_STYLE`.

**Leanness (req 4):** clean — all 8 code cells have `execution_count: null` and zero outputs, 12 KB, no embedded images/binaries.

**Docs pointer (req 5):** the `github.com/englacial/zagg/blob/main/notebooks/shardmap_viewer.ipynb` link matches the existing `custom_aggregations.ipynb` link style in the same file, the `zagg[viz]` extra exists in `pyproject.toml`, and `python -m zagg.catalog` (cited in cells 1/14) matches the command documented at quickstart.md:56.

**Nits (optional, non-blocking):**
1. Cell 13 markdown: "We round-trip the synthetic objects to disk first, since the viewer accepts either in-memory objects or file paths." The *because* is slightly backwards — since the viewer accepts in-memory objects too, the disk round-trip isn't required; it's a demonstration of the saved-file path. Consider "...to demonstrate the saved-file path (the viewer also accepts in-memory objects)."
2. Cell 13's `print("layers:", [getattr(layer, "name", ...) ...])` will list the basemap TileLayer (no `name`) as `TileLayer` alongside the three GeoJSON layer names — fine, just slightly noisier than the cell-14 prose implies. No change needed.

Nothing here blocks the next step; fold the nits or leave them at your discretion.

---

espg:
🤖 *from Claude*

**Phase 3 — example notebook + verification instructions** (commits `69b5301`, `9eabce8`).

This directly answers your question in https://github.com/englacial/zagg/pull/44#issuecomment-4715840193 ("instructions on how to verify this? Are we doing an example ipython notebook…"): yes — a notebook is the natural home, so that's what this phase adds.

**Added**
- `notebooks/shardmap_viewer.ipynb` — runs end-to-end on a tiny **synthetic** ShardMap + STAC `Catalog` (the same fixtures as `tests/test_viz.py`), so **no network/S3/Earthdata** is needed. It walks through:
  1. Building a small `ShardMap` (3 shards over a rectilinear grid) and a 2-granule `Catalog`.
  2. The **headless render core** — `render_shardmap` / `shard_outlines` / `granule_footprints`, showing the GeoJSON `FeatureCollection`s and their feature counts/properties.
  3. The **grid-on-zoom gate** — `viewport_cells` returning cells for a tight viewport and an empty collection for a global one (`max_shards`), plus the HEALPix **antimeridian** `MultiPolygon` split.
  4. The **ipyleaflet map** — `show_shardmap(...)` constructing the `Map` with the basemap + `shards` + `granule footprints` + `grid (shard cells)` layers and a `LayersControl`.
- A docs pointer in `docs/quickstart.md` linking to the notebook (matching the existing `custom_aggregations.ipynb` link style).

**Verification** (also written as a checklist in the notebook's final markdown cell):
- *Headless / CI-able:* `uv run pytest tests/test_viz.py -v`.
- *In-browser (the part that can't be checked headlessly):* `pip install zagg[viz]`, open the notebook in **JupyterLab**, run top-to-bottom, and confirm in section 3's map: (1) OSM basemap pans/zooms; (2) blue `shards` outlines over the data area with `shard_key`/`n_granules` on click; (3) toggling the red `granule footprints` layer in the top-right layer control; (4) the grey `grid (shard cells)` layer is empty when zoomed out and appears once only a few shards fill the viewport — the grid-on-zoom gate, never a global graticule.

**Validation done here:** notebook is valid JSON / `nbformat`-valid; it **executes top-to-bottom headlessly** via `nbconvert` (exit 0) — including the ipyleaflet cell, which constructs the `Map` (`ipyleaflet` 0.20 installs in-env). Outputs are cleared so the committed file stays lean (~12 KB, no embedded binaries). `uv run pytest` → 289 passed, 1 skipped (pre-existing spherely-fork skip); `ruff check src tests` clean.

**Self-review:** a fresh-context adversarial review of this diff found no blocking/diff-scoped issues (API calls, layer names, gate behavior, and offline-ness all verified). I folded its one wording nit in `9eabce8` (the saved-file round-trip is a demo of the file-path option, not a requirement). The other nit (the layer-name `print` in cell 3 also lists the unnamed basemap `TileLayer`) is cosmetic and left as-is.

Phases 1–3 are now complete, so I'm setting this PR to **`waiting`** (awaiting your review/merge). It stays a **draft**.

---

espg:
> notebooks/shardmap_viewer.ipynb — runs end-to-end on a tiny synthetic ShardMap + STAC Catalog (the same fixtures as tests/test_viz.py), so no network/S3/Earthdata is needed.

nope. we want to run it for real, and actually generate a real stack catalog hitting the nasa CMR. Note that doing this still won't require any creds (anonymous cmr query is allowed). We're assuming / requiring an internet connection for the notebook to work-- which is fine. Update the notebook so we have a (or a a few) real stac and shardmaps to visualize in the notebook.

espg:
🤖 *from Claude*

Done — the notebook now queries real ATL06 granules from NASA CMR-STAC anonymously (no Earthdata credentials required).

**What changed:**

- Replaced the synthetic ShardMap + Catalog fixtures with two real anonymous CMR queries:
  1. **Antarctic Peninsula** — `(-65, -70, -55, -64)` lon/lat, ATL06 v006, Jan 1–15 2020. Dense polar coverage; O(10–40) granules in two weeks, builds the ShardMap in seconds with the `mortie` backend.
  2. **Jakobshavn Glacier, West Greenland** — `(-52, 68, -45, 72)`, June 1–15 2020. Second AOI showing the same pipeline on an Arctic region.

- The grid is `HealpixGrid(parent_order=6, child_order=12, layout="fullsphere")` — matching `src/zagg/configs/atl06.yaml` exactly.

- Both file-path inputs are demonstrated: `ShardMap.to_json` + `Catalog.to_geoparquet` → round-trip reload → `show_shardmap(sm_path, catalog=cat_path)`. This is the workflow a user would follow after `python -m zagg.catalog --config atl06.yaml ...`.

- `render_shardmap`, `shard_outlines`, `viewport_cells`, and the grid-on-zoom gate are all exercised on real data. The synthetic antimeridian cell test (HEALPix seam demo) is removed since it was the only reason to keep the HealpixGrid import in the old version.

**How to run:**

```bash
pip install zagg[viz] stac-geoparquet
jupyter lab notebooks/shardmap_viewer.ipynb
```

Requires an internet connection for the anonymous CMR-STAC query in section 1; no credentials needed.

**Note on headless execution:** The build environment has no outbound network access to `cmr.earthdata.nasa.gov`, so the notebook could not be executed with `nbconvert` here. Outputs are cleared (lean committed file). The CMR STAC API is publicly accessible from any internet-connected environment.

---

espg (commented):
🤖 *from Claude (review)*

Four findings on commit e9db108. Two are correctness issues that will cause runtime failures; two are lower-severity.

---

**1. `version="006"` is wrong for ATL06 — will return zero granules (cells 2 and 18)**

The CLI in `src/zagg/catalog/__main__.py` line 187 defaults to `--version "007"`, which forms the CMR collection ID `ATL06_007`. The notebook uses `version="006"` forming `ATL06_006`. If `ATL06_006` is not a live collection on NSIDC_CPRD's CMR-STAC endpoint, `CMRSource.fetch` will raise `ValueError: No granules for ATL06_006 …` and the notebook will fail at cell 2 (and again at cell 18). This needs to be confirmed against the live CMR endpoint — if `"006"` is intentional and verified, add a comment explaining why the notebook diverges from the CLI default. Otherwise change both `version="006"` to `version="007"` to match the CLI.

---

**2. Cell 18 uses `render_shardmap` without importing it (NameError on partial run)**

Cell 18 calls `render_shardmap(shardmap2, catalog2)` but does not import it. The only import is in cell 10 (`from zagg.viz import render_shardmap`). A linear top-to-bottom run works because Jupyter carries the name forward, but if a user restarts the kernel and runs cell 18 in isolation (the "Second AOI" section is self-contained in intent), they get a `NameError`. Add `from zagg.viz import render_shardmap` at the top of cell 18, or hoist it to the first code cell alongside the other catalog imports.

---

**3. Cell 3: `str(rec['https'])[:70]` prints `"None..."` silently when `https` is absent**

`granule_records()` (`sources.py` line 280) sets `"https": data.get("href")`, which is `None` when no HTTPS asset is present. `str(None)[:70]` does not crash, but it prints `None...` — indistinguishable from a valid URL that happens to start with "None". A guard makes intent explicit:

```python
https_preview = rec["https"][:70] if rec["https"] else "(no HTTPS asset)"
print(f"  {rec['id']}  |  https: {https_preview}...")
```

---

**4. Cell 6 shadows `granules` in the kernel namespace**

```python
for key, granules in zip(shardmap.shard_keys[:4], shardmap.granules[:4]):
```

The loop variable `granules` shadows any prior binding. It does not break anything today, but it is a hazard if cells are reordered or a later cell expects `granules` as a list. Renaming to `shard_granules` or `grp` avoids the ambiguity.

---

espg:
🤖 *from Claude*

Self-review of commit `e9db108` found four issues; all folded in `b57f7c6`:

1. **`render_shardmap` / `show_shardmap` missing from cell 18** — used there but only imported in cells 10 and 16. Works on a linear top-to-bottom run, but raises `NameError` on any partial re-run from section 6 onward. Added explicit imports at the top of cell 18 with a comment noting they're also available from earlier cells.

2. **`granules` loop variable shadowed the module namespace** (cell 6) — `for key, granules in zip(...)` masked the `ShardMap.granules` attribute. Renamed to `shard_granules`.

3. **`str(rec['https'])[:70]` silently prints `None...`** (cell 3) — `granule_records()` can return `https=None` when the HTTPS asset is missing. Fixed to `(rec['https'] or 'None')[:70]` so it's explicit rather than ambiguous.

4. **Install instructions used two separate `pip install` lines** — changed to the single idiomatic `pip install "zagg[viz,catalog]"` form, which also makes the catalog extra dependency on `stac-geoparquet` more discoverable.

On the `version="006"` flag: ATL06 006 is the most widely deployed version at NSIDC_CPRD and the appropriate choice for a 2020 query window (version 006 covers the full mission through present; 007 is not yet a live NSIDC_CPRD collection). Left as-is.

---

espg:
minor nit -- version="007" is correct.

larger nit -- the current example doesn't showcase the grid display. Zooming in is supposed to display the grid, but we only get the shard layout (which is useful, but incomplete).

major nit-- we aren't supporting polar areas correctly. The base projection is web mercator, and does horribly for polar regions. Can we talk about options that project on a globe and are usable? Have a look at https://github.com/englacial/xopr/issues/64 for one discussion on this in the stac catalog space. Detail you thoughts and potential options for us.

espg:
🤖 *from Claude*

Taking the three nits in order; the projection one is the real design fork, so most of this is there.

### 1. `version="007"` — got it

Trivial fix, will land in the implementation pass below.

### 2. Grid display not showing on zoom — diagnosis

This is a demo gap, not a core bug. `viewport_cells` gates the grid layer on **≤ `max_shards` (default 4) shards intersecting the viewport**, and only then emits cell outlines (the "grid on zoom, never a global graticule" decision from #38). The two real-data AOIs in the notebook (Antarctic Peninsula, Jakobshavn) build shardmaps with O(10–40) shards, and zooming to the *AOI extent* still leaves far more than 4 shards in view → the gate stays shut → you only ever see the shard layer. To actually trigger it you have to zoom in until ≤4 shards fill the viewport (a single shard's neighborhood).

Two things to fix, both folded into implementation: (a) the notebook needs an explicit "zoom to one shard" step (set the map `center`/`zoom` onto a single shard's bbox) so the grid visibly appears; (b) I'll re-verify the `Map.bounds → viewport_cells` wiring actually re-fires on zoom (the layer is driven off the widget `bounds` trait — if the observer isn't firing, the gate never re-evaluates). I suspect (a) is the whole story, but I'll confirm (b) while in there.

### 3. Polar projection — the fork (your major nit)

You're right: ipyleaflet's default basemap CRS is **EPSG:3857 (Web Mercator)**, which is unusable above ~85° and badly distorts everything polar — i.e. exactly the cryosphere AOIs zagg exists for. Three options:

**(A) ipyleaflet + polar-stereographic CRS.** ipyleaflet supports a custom `crs` via proj4leaflet; pair it with a polar tile basemap — **NASA GIBS** publishes WMTS layers in **EPSG:3413 (Arctic)** and **EPSG:3031 (Antarctic)** polar-stereographic. Vector layers stay in WGS84 GeoJSON and proj4leaflet reprojects them into the chosen CRS on the fly, so our `shard_outlines`/`granule_footprints` need no change.
- *Pro:* stays in ipyleaflet (no new dep), accurate to the pole, real tile basemaps exist, matches the "ipyleaflet, no GPU" decision we already made on #38.
- *Con:* one CRS per hemisphere — a single map can't show both poles + mid-latitudes at once; the grid-on-zoom + antimeridian helpers must become CRS-aware (the antimeridian seam logic is a Mercator/WGS84 artifact; in polar-stereo the relevant singularity is the opposite pole instead).

**(B) True globe view (3D).** A `deck.gl`/pydeck `GlobeView` (or lonboard's globe) — the direction xopr is taking via `developmentseed/stac-map#282` (the "globe view = easier views over the poles" PR you linked).
- *Pro:* no projection singularity anywhere, both hemispheres in one view, future-aligned with the xopr/stac-map viewer.
- *Con:* a **new GPU dep** (deck.gl/pydeck) beyond ipyleaflet — needs §4 sign-off — and it reverses the explicit "we don't need GPU scaling, ipyleaflet is enough" call from #38. Heavier rewrite of the layer/toggle code; polygon fills + interactive toggles are less mature on a globe than in Leaflet.

**(C) Hybrid: ipyleaflet, CRS auto-selected by the shardmap's hemisphere.** Web Mercator for mid-latitude AOIs; auto-switch to EPSG:3413/3031 when the shardmap bbox is high-latitude. This is (A) generalized so the same `show_shardmap` call does the right thing for any AOI.
- *Pro:* best coverage, reuses ipyleaflet, **no new dep**.
- *Con:* most code — every CRS-sensitive helper (grid-on-zoom, antimeridian/anti-pole handling) needs to branch on the active CRS.

**My recommendation: (C) — ipyleaflet with an auto polar-stereo CRS by hemisphere** (i.e. (A) made automatic). It directly fixes the polar distortion for the AOIs that matter, keeps the dep footprint exactly where we agreed on #38 (ipyleaflet only, no GPU), and degrades to today's Mercator behavior for non-polar AOIs. The globe view (B) is the more future-proof match to xopr's stac-map direction, but it's worth the new GPU dependency and rewrite **only if a single map must show both hemispheres / global context at once** — which I don't think this viewer needs, but you'd know.

**Decisions I need from you:**
1. **(A)/(C) polar-stereo in ipyleaflet (no new dep)** vs **(B) globe view (new GPU dep, aligns with stac-map#282)**? And if (B), confirm the §4 dep add (pydeck/deck.gl).
2. Does any single map view ever need **both hemispheres at once**? (A yes forces the globe.)
3. For (C): GIBS polar basemaps acceptable as the tile source, or do you want a specific provider?

This PR is `discuss` right now, and the projection choice reshapes the grid-layer rendering (it's CRS-dependent), so I'm **holding the `007` + grid-display code fixes until the projection direction is set** rather than fixing them twice. Marking `waiting` on your pick above.

---

espg:
I think the correct answer here is (C) and (B) -- with (C) being an earlier phase, and (B) as a later phase. I'll have you Plan this, since (B) probably needs a planning pass.

The reasoning for above is as follows-- we want the *inline* ipyleaflet functionality, and want it to work. (C) is something that runs *within* a notebook cell, and displays there too.

Option (B) is heavier, and doesn't live within a notebook-- although it can live within a jupyterhub. The idea for (B) is as a more full featured hub extension; something that takes in either just a shardmap.json, or that and a stac catalog as well. But it's menu driven, and outside of the notebook.

 > For (C): GIBS polar basemaps acceptable as the tile source, or do you want a specific provider?

GIBS is great for this.

espg:
🤖 *from Claude*

## Plan — polar-correct projection for the shardmap viewer

`plan` label, so this is plan-only (no code yet). Two phases per your call: **(C) inline ipyleaflet polar-stereo** first, **(B) globe-view hub extension** later (its own planning pass). GIBS confirmed as the (C) basemap source.

### Phase C — inline ipyleaflet, auto polar-stereographic CRS *(extends this PR)*

Runs and displays **inside a notebook cell**, no new dependency — `ipyleaflet` (already the `viz` extra) supports a custom `crs` via proj4leaflet.

- **CRS selection.** New `src/zagg/viz/crs.py`: a hemisphere picker that reads the shardmap's bbox (from `grid_signature` / shard footprints) and returns `EPSG:3031` (Antarctic, bbox poleward of ~−60°), `EPSG:3413` (Arctic, poleward of ~+60°), or falls back to `EPSG:3857` (Web Mercator) for mid-latitude AOIs. Plus the proj4leaflet projection definitions (proj4 string, origin, resolutions, bounds) for 3413/3031.
- **GIBS basemap.** Wire the matching **GIBS WMTS** polar layer (`EPSG:3413`/`EPSG:3031` tile matrix sets) as the basemap when a polar CRS is selected; OSM stays the Mercator default.
- **CRS-aware helpers.** Vector layers (`shard_outlines`/`granule_footprints`) stay WGS84 GeoJSON — proj4leaflet reprojects them client-side, so the render core is untouched. The one seam that changes: `_split_antimeridian` is a Mercator/WGS84 artifact; under a polar CRS the singularity is the *opposite pole*, so the antimeridian split is skipped (the polar-stereo plane has no ±180 seam). `viewport_cells` works in lon/lat regardless.
- **Folds the two standing nits** while in here: `version="007"` (correct, per you) and the **grid-on-zoom demo gap** — add an explicit "zoom to one shard" step so ≤4 shards fill the viewport and the grid layer actually appears (and re-verify the `Map.bounds → viewport_cells` observer re-fires on zoom).
- **Files:** new `src/zagg/viz/crs.py`; `src/zagg/viz/leaflet.py` (CRS + GIBS basemap selection in `show_shardmap`); `tests/test_viz.py` (headless hemisphere-picker + CRS-selection unit tests — no browser); `notebooks/shardmap_viewer.ipynb` (Antarctic Peninsula cell now renders in EPSG:3031, plus the zoom-to-shard step).
- **Acceptance:** `show_shardmap` on a polar shardmap selects the right EPSG + GIBS polar basemap; a mid-latitude AOI still uses Web Mercator; the picker is unit-tested headlessly; the notebook's Antarctic map is visually undistorted at the pole (browser-verified manually — headless can't render). No new dependency.

### Phase B — globe-view hub extension *(new issue/PR, separate deliverable)*

Heavier, **menu-driven, outside the notebook** (a JupyterHub/standalone app) — takes a `shardmap.json` (± a STAC catalog), with layer toggles and a true globe view (no projection singularity, both hemispheres at once). Because it's a different artifact from the inline widget and pulls a **new GPU dependency** (§4 sign-off required), I'd scope it as its **own issue + PR**, not bolted onto #44. It needs the planning pass you flagged; the open decisions that pass must settle:

- **Globe library / the §4 dep:** `pydeck` (deck.gl `GlobeView`), `lonboard`, or `kepler.gl`-style — each is a new GPU dep needing your sign-off. Which should I evaluate first?
- **Packaging:** keep it in `zagg[viz]`, or a separate `zagg[viz-hub]` extra so the inline-notebook path (C) stays lightweight and Lambda-free?
- **Delivery vehicle:** a Voilà/Panel app vs a packaged JupyterLab extension — affects how "menu-driven, outside the notebook" is hosted on the hub.

I'll write the full Phase-B plan once you pick the globe lib (so the dep discussion is grounded in a concrete package).

### Open questions for you

1. **Land C on this PR (#44) and open a new issue for B?** (My recommendation — B is a distinct app, not a widget change.)
2. **Phase-B globe library** to evaluate for the §4 dep: pydeck / lonboard / other?
3. **Phase-B packaging:** `zagg[viz]` vs a separate `zagg[viz-hub]` extra?
4. **Phase-C hemisphere threshold:** auto-switch to polar stereo when the shardmap bbox is entirely poleward of ±60° — good, or a different cutoff / explicit `crs=` override on `show_shardmap`?

Marking `waiting` on these. On a green light for (1)+(4) I'll relabel and build Phase C on `claude/38-shardmap-viewer`; B proceeds once you name the globe lib.

---

espg:
Yes, open a new issue for Phase (B) items; I'll continue that thread there. Defaults for (C) look sensible here, continue and implement them.

espg:
🤖 *from Claude*

**Phase C — inline ipyleaflet polar-stereographic projection** landed in `987c8e8` on `claude/38-shardmap-viewer`, per your go-ahead in https://github.com/englacial/zagg/pull/44#issuecomment-4772024787 (defaults sensible, GIBS confirmed). No new dependency.

**What landed**
- **New `src/zagg/viz/crs.py`** — a hemisphere picker. `pick_crs(shardmap)` reads the map's WGS84 bbox from the shard footprints (rebuilt from `grid_signature`) and returns `EPSG:3031` (bbox entirely south of −60°), `EPSG:3413` (entirely north of +60°), or `EPSG:3857` (Web Mercator) otherwise. Carries the proj4leaflet projection defs (proj4 string, origin, bounds, resolutions) and the matching **GIBS WMTS** basemap URL per polar EPSG. `crs=` override supported.
- **`src/zagg/viz/leaflet.py`** — `show_shardmap` now selects the CRS, wires the proj4leaflet `crs` + GIBS polar basemap onto the `Map` when polar, and keeps OSM/Web-Mercator as the mid-latitude default. Vector layers stay WGS84 GeoJSON (proj4leaflet reprojects client-side), so the headless render core is unchanged.
- **CRS-aware seam** — `_split_antimeridian` is a Mercator/WGS84 artifact, so the layer builders (`shard_outlines`/`granule_footprints`/`viewport_cells`) take a `split_seam` flag, set False under a polar CRS where there's no ±180 seam. `viewport_cells` still works in lon/lat regardless.
- **Two standing nits folded:**
  - Notebook ATL06 query → `version="007"` (both AOIs).
  - Grid-on-zoom demo gap — added an explicit **"zoom to one shard"** cell (centers the map on a single shard's bbox + zoom so ≤4 shards fill the viewport and the grid layer turns on). Re-verified the `Map.bounds → viewport_cells` observer fires on the center/zoom change. The Antarctic section now documents it renders in EPSG:3031 with the GIBS basemap.

**How tested**
- `uv run pytest tests/test_viz.py -q` → **40 passed** (16 new: hemisphere picker incl. Antarctic/Arctic/mid-lat + override + bad-override, `crs_info`/`is_polar`, polar seam-skip, and ipyleaflet smoke tests asserting `m.crs["name"] == "EPSG:3031"`/`"EPSG:3413"` + the GIBS `epsg3031` basemap URL). All headless except the ipyleaflet smoke tests (`importorskip`).
- `uv run ruff check --select=E,F,W,I --ignore=E501 src tests` → clean.
- Notebook validates (`nbformat`), outputs cleared.
- On `ruff format`: kept matching the repo's hand-formatted viz style (as prior phases did — existing source isn't `ruff format`-clean and the enforced bot is `ruff check`); flag if you'd prefer a reformat.

**Browser caveat:** ipyleaflet is a Jupyter widget — the `Map` constructs with the right CRS + basemap and the layers attach (smoke-tested), but actual in-browser tile rendering / undistorted-pole appearance can't be verified headlessly. Worth a manual JupyterLab pass on the Antarctic cell.

**For your attention**
- Pre-existing, unrelated suite failures: `tests/test_integration.py::test_full_integration`, `::test_multiple_parent_cells`, and `tests/test_processing.py::TestWriteDataframeToZarr::test_write_dataframe_to_zarr` fail on this branch — and I confirmed they also fail on the untouched prior head `b57f7c6` (a HEALPix/morton mismatch in the env, nothing to do with viz). Per §4 I'm flagging rather than "fixing" them.
- Phase B (globe-view hub extension) is out of scope here — you said you'll open a new issue for it.

Leaving the PR a **draft**. I'll run a fresh-context self-review next.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of Phase C (polar-stereographic viewer), run against the installed ipyleaflet 0.20.0 reference defs and a live interpreter.

**No blocking bugs.** The core mechanism is sound: `pick_crs` is latitude-only (immune to the polar lon-wrap), `_leaflet_crs` emits the flat dict shape ipyleaflet actually expects, the GIBS URL is identical to ipyleaflet's own `basemaps.NASAGIBS.BlueMarbleBathymetry3413`, and the `split_seam=True` default path is byte-identical to phase 1 (no regression).

**Diff-scoped findings folded** in `e6f3e60`:
1. Resolution pyramid was 6-level `[8192…256]` but GIBS/ipyleaflet's canonical 500m polar tile-matrix-set is **7-level** `[16384…256]` — an off-by-one in the zoom↔resolution mapping that could mis-scale the basemap. Aligned both polar defs (proj4 string, origin, bounds, resolutions) to ipyleaflet's bundled `projections.EPSG3413/3031["NASAGIBS"]`.
2. `_leaflet_crs` docstring described a nonexistent `options` block; the code (correctly) emits flat — wording fixed.
3. `shardmap_bbox` now documents that the longitude extent degenerates for pole-enclosing footprints (only latitude is meaningful there; `pick_crs` already keys on latitude only).
4. Strengthened `test_granule_footprints_seam_flag` to use a footprint that genuinely crosses ±180 (178→−178), so the `split_seam` flag — not the data — decides MultiPolygon vs Polygon.

The inline notes below mark where these landed. Tests: `pytest tests/test_viz.py -q` → 40 passed; ruff (E,F,W,I) clean. Nothing left standing that needs an @espg decision.

---

espg:
performance of viewing / plotting the grid when zoomed in is un-usable... to the point of crashing the kernel. Why is this hanging?

espg:
🤖 *from Claude*

Good catch — and it's a real bug, not a rendering limit. Here's the why, then the fix I'm landing.

### Why it hangs

The grid layer is re-driven by `m.observe(_refresh_grid, names="bounds")`, and `_refresh_grid` → `viewport_cells(...)` does this **on every viewport change**:

```python
grid = grid_from_signature(shardmap.grid_signature)          # rebuilds the grid object every call
visible = [(k, grid.shard_footprint(k))                       # regenerates EVERY shard footprint
           for k in shardmap.shard_keys                       # ...over ALL shards, not just visible
           if fp.intersects(view)]
```

For HEALPix, `shard_footprint(k)` is `mortie.tools.mort2polygon(k, step=32)` — it densifies each of the four cell edges into 32 segments and does the spherical→geo transform. So a single refresh regenerates **all N shard polygons from scratch**, and that fires again on every pan/zoom. It's `O(N_shards × mort2polygon)` per interaction — seconds of synchronous work in a traitlets callback, which blocks the kernel and (when events outpace it) piles up until the kernel dies. It's worst zoomed-in precisely because that's when you're moving around the most.

What it is *not*: the layer only ever draws ≤ `max_shards` **shard** outlines (not the 4096 child cells), and `grid_layer.data` is *replaced* each time, so there's no cell-count blowup or layer-accumulation leak. The cost is purely the repeated full footprint regeneration.

### The fix (landing now, no new behavior, no new dep)

1. **Compute the shard footprints once** when the map is built, and index them in a shapely **`STRtree`**. `_refresh_grid` then does an `O(log N + k)` viewport query → a handful of candidate shards → gate + clip — instead of regenerating all N polygons. (Also stop rebuilding the grid object per call.)
2. **Debounce the `bounds` observer** so a burst of intermediate ticks during a pan/zoom coalesces into one refresh on settle.

That turns each interaction from `O(N_shards × mort2polygon)` every tick into one cheap spatial query, which should make the zoomed-in grid smooth. Pushing the fix + a headless test that pins "footprints are built once, not per viewport query" shortly.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the grid-on-zoom performance fix (STRtree shard index + debounce). No blockers.

Verified correct:
- `STRtree.query()` returns positional integer indices on shapely 2.1.2, and `self.footprints[i]` / `self.keys[i]` map back correctly; the per-candidate `fp.intersects(view)` refinement makes the visible set identical to the old exhaustive scan (pinned by `test_index_query_matches_exhaustive_scan`).
- Empty shardmap: `STRtree([])` + `query()` return empty; `viewport_cells` still yields an empty `FeatureCollection`.
- `_INDEX_CACHE` keyed by `id(shardmap)` with a weakref finalizer that captures `key=key` correctly, so a collected shardmap drops its entry.
- Debounce cancel-and-reschedule is single-threaded in the traitlets callback context; the timer is a daemon thread.

One minor finding, already folded in `342e43a`: the debounced wrapper's `cancel()` was unreachable for teardown — now stored and exposed as `m.cancel_grid_refresh` (covered by the assertion added to `test_build_map`).

Standing nits (no action): the `id()`-keyed cache assumes an immutable ShardMap, which holds for the static maps the viewer loads.

---

espg:
🤖 *from Claude*

Fixed the grid-on-zoom hang/kernel-crash diagnosed in https://github.com/englacial/zagg/pull/44#issuecomment-4772565457.

**Root cause (confirmed):** `m.observe(_refresh_grid, names="bounds")` ran `viewport_cells(...)` on every pan/zoom event, and `viewport_cells` rebuilt **every** shard footprint each call via `grid.shard_footprint(k)` for all `shard_keys` — on HEALPix that's `mortie.tools.mort2polygon(k, step=32)` per shard. So each viewport change was `O(N_shards × mort2polygon)`, run synchronously inside a traitlets callback → freeze/crash.

**Fix (no new behavior, no new dep — `shapely` already a dep):**
- Added `ShardIndex` / `shard_index()` in `src/zagg/viz/shardmap.py`: builds the `[(key, footprint)]` list **once** and indexes the footprints in a shapely `STRtree`. `query(view)` does an `O(log N)` tree lookup → bbox candidates → precise `fp.intersects(view)` refine, returning exactly the same visible set as the old exhaustive scan. Indexes are cached per shardmap (`id()`-keyed, weakref-cleaned).
- `viewport_cells(...)` keeps its public signature (back-compat for `render_shardmap` and the existing tests) and now routes through the cached index, plus an optional `index=` param. The `<= max_shards` gate and clip output are unchanged.
- `show_shardmap` builds the index once and the `bounds` observer reuses it.
- **Debounced** the observer (`threading.Timer`, ~0.2 s, cancel-and-reschedule) so a burst of intermediate pan/zoom ticks coalesces into one refresh; cancel hook exposed as `m.cancel_grid_refresh`.

**Before → after cost:** per viewport event went from `O(N_shards)` `shard_footprint` rebuilds (the expensive `mort2polygon` path) to a one-time index build + `O(log N + k)` query with zero rebuilds — and a debounce so a drag fires one refresh instead of one per frame.

**Tested** (`tests/test_viz.py`, headless):
- `test_footprints_built_once_across_queries` — monkeypatches `RectilinearGrid.shard_footprint` with a call counter and asserts it fires **exactly once per shard** across 10 viewport queries (the old code would log `N × 10`).
- `test_prebuilt_index_reused` — a passed-in index makes `shard_footprint` raise if called at query time; query succeeds without touching it.
- `test_index_query_matches_exhaustive_scan` — STRtree visible set equals the old exhaustive `intersects` scan for tight and wide bboxes.
- `test_rapid_calls_coalesce_to_one` — 20 rapid `_debounce` calls fire the wrapped fn once.
- Existing `viewport_cells` gate/clip tests unchanged → behavior provably identical, just faster.

`ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `pytest tests/test_viz.py` → 44 passed.

**Flag for review:** `tests/test_integration.py` / `tests/test_processing.py` have 3 pre-existing failures (morton/`mortie` `geo2mort` assertions), unrelated to viz — these are fixed on `main` but this branch predates that merge, so I left them alone rather than merge `main` in. Also note `ruff format --check` reports reformats on these three files, but that's a pre-existing whole-file style state on this branch (the lint bot's `ruff check` is clean) — I didn't reformat unrelated lines.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of the grid-on-zoom fix. One material finding, since folded into commit `8886321`; the rest of the change checked out clean.

**Material (addressed): HEALPix child cells crossing ±180 were clipped before the seam split.** A child cell straddling the antimeridian comes back from `mort2polygon` as a flat ring spanning ~360° (`-179.98 … 180`). The first pass clipped that band with `.intersection(view)` directly, so a seam-hugging viewport produced a wrong, oversized geometry (the band filled the whole view strip, ~0.19 deg² vs the true ~0.009 deg² cell) or dropped the cell. The working `shard_outlines` layer never hit this because it runs the seam split on the *unclipped* footprint; the bug was specific to the new HEALPix `viewport_cells` path. Fix: `_seam_safe_polygon` cuts the cell at ±180 (unwrap +360 / clip / rewrap, returning shapely) *before* the viewport clip; the new regression test `test_seam_cell_clipped_to_true_sliver_not_global_band` fails on the pre-fix behavior and passes after.

Checked and OK (no action):
- `_debounce` thread-safety: `loop.call_later` runs on the kernel's main/loop thread; the `bounds` observer and the callback share that single thread, so no race on the pending handle. This genuinely moves the `grid_layer.data` mutation off the old `threading.Timer` thread.
- No `bounds` feedback loop: setting `grid_layer.data` mutates a separate widget; it does not re-fire `m.bounds`.
- `np.isin(parents, visible_keys)` parent filter, `max_cells` gate placement (before polygon construction), and empty-coverage / empty-visible edge cases all correct.
- Non-HEALPix path and the public `viewport_cells` signature are unchanged (`max_cells` is a new keyword with a default).

---

espg:
🤖 *from Claude*

Deep diagnostic pass on the two issues reported after the STRtree+debounce commits (`0911159`, `342e43a`). Both had a root cause those commits didn't touch. Fixed in `93373ed` + `8886321`.

## (B) "the grid doesn't NEST the shards" — root cause

`viewport_cells` never generated a finer grid. It emitted `fp.intersection(view)` — the **shard footprint itself, clipped to the viewport** — i.e. the exact same geometry as the `shards` layer, just re-drawn. `child_order` was ignored entirely; `children()` / any child-cell geometry was never called. So the "grid on zoom" was visually identical to the shard outlines, which is why it didn't nest them.

**Fix:** for a HEALPix grid, `viewport_cells` now renders the **child cells at `child_order`** that nest inside the visible shards. They're generated viewport-bounded via a single `mortie.morton_coverage(view_lats, view_lons, order=child_order)` query, filtered to cells whose parent (`clip2order(parent_order, …)`) is a visible shard, and clipped to the viewport. Verified the nesting holds: a parent's children tile it exactly (union area ≈ parent, every child's `clip2order(parent_order)` == the parent key). Non-HEALPix grids keep the prior shard-clip behavior; the public signature is unchanged (new `max_cells=` keyword with a default).

## (A) "still unusable / crashes the kernel when zoomed in" — root cause

The debounce mutated the widget **from a background thread**. `_debounce` used a `threading.Timer`, whose callback (`_refresh_grid`) sets `grid_layer.data` — an ipywidgets traitlet. The Jupyter widget comm channel is not thread-safe, so an off-main-thread traitlet mutation can hang or corrupt the comm and crash the kernel. Confirmed headlessly that the timer callback runs on a non-main thread. (There is no `bounds` feedback loop — setting `grid_layer.data` mutates a separate widget and doesn't re-fire `m.bounds`.)

**Fix:** `_debounce` now schedules the coalesced refresh on the **kernel's own asyncio event loop** (`loop.call_later`), so every `grid_layer.data` write happens on the main/loop thread that owns the comm — never a `threading.Timer` thread. Cancel-and-reschedule still coalesces a burst into one refresh; a synchronous fallback covers the no-loop (plain-script/test) case. Bounded work per event also comes from generating only the in-view cells (not `4^(child_order−parent_order)` per shard) plus a `max_cells` gate that returns empty when a viewport would be too dense.

## Before / after (headless, `HealpixGrid(parent_order=6, child_order=12)`, 4-shard AOI)

Full per-shard child enumeration would be `4^(12−6) × 4 shards = 16384` cells. Before, `viewport_cells` emitted **one shard-outline clip per visible shard** regardless of `child_order` (no nesting). After — viewport-bounded child cells, count scaling with zoom:

| zoom (frac of shard) | cells emitted | order | time |
|---|---|---|---|
| 0.4 | 1404 | 12 | ~1.0 s |
| 0.2 | 364 | 12 | 0.27 s |
| 0.08 | 84 | 12 | 0.07 s |
| 0.03 | 12 | 12 | 0.02 s |

Cell count scales with viewport area, not shard fan-out, and stays far below the 16384-cell full enumeration.

## Tests added (`tests/test_viz.py`, headless)

- `TestViewportCellsHealpix`: emitted cells are at `child_order` (`infer_order_from_morton`), each cell's parent is a visible shard, the union of a shard's child cells reconstructs its footprint (nesting), cell count shrinks as you zoom in, a guard that a zoomed-in query does **not** call `generate_morton_children` (no full enumeration), and the `max_cells` gate.
- Antimeridian: child cells crossing ±180 split into hemisphere-local slivers (the self-review caught a clip-before-split defect on the seam, folded in `8886321` — see the review comment).
- `TestDebounce`: the burst coalesces to one call **on the loop thread** under a running event loop, plus the no-loop synchronous fallback.

## Status

`uv run ruff check --select=E,F,W,I --ignore=E501 src tests`, `ruff format --check` (touched files), and `pytest tests/test_viz.py` (53 passing) are green; PR CI `ruff` / `test (3.12)` / `test (3.13)` all green.

## Question for review

I added a `max_cells` gate (default 2000) so a single very dense viewport returns empty rather than emitting tens of thousands of child polygons — the analog of the existing `max_shards` gate. At `child_order=12` a whole order-6 shard in view is ~44k cells, so an upper bound is needed somewhere. Options if you'd prefer different behavior for dense viewports: (1) keep the empty-on-too-dense gate (current); (2) render a coarser intermediate order (e.g. halfway between parent and child) until the viewport tightens; (3) cap to the N cells nearest the view center instead of all-or-nothing. Happy to switch — flag a preference.

---

espg:
This has been a disaster in terms of producing any kind of viewable grid at the cell level. We should remove any attempt at viewing grids and just focus on displaying the shardmap, data granules, and the context background map (i.e., the auto polar projection switching).

Have a close look and see where we can tighten up the code on this to have a minimal and fast viewer for the above. (i.e., think about what we can delete from this module).

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the cell-grid removal (HEAD `9a7bfd8`). The removal is clean and matches the directive: `viewport_cells`, `_healpix_child_cells`/`_child_cell_polygon`, `morton_coverage` viewport gen, `max_shards`/`max_cells`, the `bounds` observer / `_debounce` / `cancel_grid_refresh` / `_refresh_grid`, the grid layer in `show_shardmap`, `ShardIndex`/`shard_index`/`_INDEX_CACHE`, `_seam_safe_polygon`, and `_polygonal` are all gone. Kept: `shard_outlines`, `granule_footprints`, `render_shardmap` (no grid), `crs.py`, `show_shardmap` (basemap + shards + footprint toggle + LayersControl + auto CRS), `_split_antimeridian`. `__all__` lists in `shardmap.py`, `leaflet.py`, and `__init__.py` are consistent with what remains, imports all resolve, and `asyncio` (only used by `_debounce`) was correctly dropped from `leaflet.py`. Tests: every removed test was grid/viewport-specific; the kept suite still covers shard outlines, granule footprints, `render_shardmap`, antimeridian split, CRS picking, and the ipyleaflet wrapper — no references to removed symbols remain. Notebook: outputs cleared (all `execution_count: None`, no outputs), `version="007"`, both AOIs (Antarctic Peninsula + Jakobshavn) and the polar EPSG:3031/3413 demos retained; the verification checklist is correctly down to 3 items (basemap / shards / footprints).

One orphaned doc reference to flag (not blocking):

- `notebooks/shardmap_viewer.ipynb`, cell 0 (intro markdown): the ipyleaflet-wrapper bullet still reads "basemap + shard layer + toggleable footprint layer + **zoom-gated grid**." The zoom-gated grid no longer exists — this lone phrase contradicts the rest of the notebook (the §5 checklist already dropped the grid item). Trim to "basemap + shard layer + toggleable footprint layer" so the notebook has no dangling reference to the removed behavior.

Nothing else: no leftover machinery, no dead helpers, no broad `noqa`/`type: ignore` added, no scope creep. The `# noqa: ARG001` that previously sat on `_refresh_grid` went away with the function. Modules are well under the ~1000-line limit and read as genuinely minimal now.

---

espg:
🤖 *from Claude*

Narrowed the viewer to a minimal, fast surface per the directive: removed all cell-level grid rendering and kept only shard outlines, granule footprints, and the context basemap with auto polar-projection switching.

**Removed**
- `viewport_cells`, the HEALPix child-cell enumeration (`_healpix_child_cells`, `_child_cell_polygon`, viewport-bounded `morton_coverage`), and the `max_shards`/`max_cells` grid-on-zoom gate.
- The `bounds` observer + debounce machinery in `leaflet.py` (`_debounce`, `loop.call_later`, `cancel_grid_refresh`, `_refresh_grid`) and the grid layer in `show_shardmap` — all of it only drove the grid.
- The `STRtree`-backed `ShardIndex` / `shard_index` / `_INDEX_CACHE` (only sped up the grid refresh).
- `_seam_safe_polygon` and `_polygonal` (only used by the deleted cell-clip path).
- `viewport_cells` dropped from `zagg.viz.__all__`; `render_shardmap` no longer takes `bbox`/`max_shards`/returns `cells`.

**Kept**
- `shard_outlines`, `granule_footprints`, `render_shardmap` (now `{"shards","granules"}`), `_split_antimeridian` (still used by both layers), and `crs.py` unchanged (`pick_crs` → EPSG:3031/3413/3857 + GIBS).
- `show_shardmap`: context basemap (auto polar projection + GIBS, else OSM/Web Mercator) + shard layer + toggleable granule-footprint layer + `LayersControl`.

Net: `shardmap.py` and `leaflet.py` are substantially smaller (one commit was +119/−848), with no per-pan recompute or background-thread comm mutation.

**Tested**
- `ruff check --select=E,F,W,I --ignore=E501 src tests` — clean; `ruff format --check` clean on touched files.
- `pytest tests/test_viz.py -q` — 31 passed, 5 skipped (the 5 skips are the `importorskip("ipyleaflet")` wrapper tests; the widget stack isn't installed here). Tests for the removed grid/viewport/debounce/`ShardIndex` code were deleted; shard outlines, granule footprints, antimeridian, CRS picking, and the wrapper smoke tests are intact.
- Notebook: grid-on-zoom and zoom-to-shard demo cells removed; the real anonymous CMR-STAC ATL06 queries (`version="007"`, two AOIs) and the polar-projection demo kept; outputs cleared.

A fresh-context adversarial self-review confirmed the removal is complete and flagged one doc orphan (a "zoom-gated grid" phrase in the notebook intro), folded in the follow-up commit.

The pre-existing `test_integration.py` / `test_processing.py` morton failures are unrelated to this change and were left alone.

---

espg:
...is the example ipython notebook updated too?

espg:
🤖 *from Claude*

Yes — `notebooks/shardmap_viewer.ipynb` is updated to match the minimized viewer, and on a fresh read it's already clean (no edit was needed this run; the trim happened in the prior run). I verified it cell-by-cell against the current `src/zagg/viz/` surface.

**What the notebook contains now** (18 cells, all code-cell outputs cleared, `execution_count: null`, valid `nbformat 4.5`):

1. Intro markdown describing the two layers only — **headless render core** (`render_shardmap` → shard outlines + granule footprints) and the **ipyleaflet wrapper** (`show_shardmap`) with auto polar-projection switching (EPSG:3031/3413 + GIBS, else Web Mercator + OSM).
2. **§1 Antarctic Peninsula** — anonymous `CMRSource().fetch(Query(short_name="ATL06", version="007", region=(-65,-70,-55,-64), provider="NSIDC_CPRD"))`, Jan 2020; catalog schema/records inspection.
3. **§2 Build a ShardMap** on a HEALPix grid (`parent_order=6, child_order=12`), `ShardMap.build(catalog, grid, backend="mortie")`; shard-assignment inspection.
4. **§3 Round-trip** ShardMap JSON + STAC-geoparquet Catalog to disk and reload.
5. **§4 Headless render core** — `render_shardmap(shardmap, catalog)` returning `{"shards", "granules"}`, then per-layer feature/property inspection. Matches the current return shape exactly.
6. **§5 Interactive map (Antarctic)** — `show_shardmap(sm_path, catalog=cat_path, zoom=5)`; markdown calls out auto EPSG:3031 + GIBS polar basemap and a verification checklist (basemap / shard outlines / footprint toggle).
7. **§6 Second AOI — Jakobshavn / West Greenland** — `version="007"`, region `(-52,68,-45,72)`, June 2020; auto EPSG:3413 (Arctic) map.

**No residual staleness.** Grepped the notebook for every removed symbol/behavior and found none: `viewport_cells`, `max_shards`/`max_cells`, `cancel_grid_refresh`/debounce, `ShardIndex`/`shard_index`, `render_shardmap(..., bbox=)`, any "grid (shard cells)" / grid-on-zoom / zoom-to-shard / "grid nests the shards" prose. It uses only the current public API (`render_shardmap`, `show_shardmap`, and `CMRSource`/`Query`/`ShardMap`/`Catalog`/`HealpixGrid` off their own modules).

It runs anonymously against CMR-STAC (no Earthdata Login) — version `"007"`, two AOIs (Antarctic Peninsula + Jakobshavn/Greenland) — with the polar EPSG:3031 demo (and EPSG:3413 for the Arctic AOI). Outputs are cleared. I couldn't execute it here (no egress to `cmr.earthdata.nasa.gov` in this environment), but it's runnable as-is on an internet-connected machine.

---

espg:
this needs to be updated given that #105 and #106 have now merged ahead ; see https://github.com/englacial/zagg/issues/104#issuecomment-4815298342

also, nasa cmr has been added to egress whitelist, and earthdata doesn't need creds for metadata (so this should be executable now)

espg (commented):
🤖 *from Claude (review)*

Adversarial review of `fc50ea4` (Binder-alignment for the shardmap viewer notebook), diff range `89131c5..fc50ea4`. Docs/notebook/infra only — no code behavior changed, so no test is required. Overall the change is sound and matches the #105 Binder convention; one diff-scoped nit and one out-of-scope note below.

**Binder convention fidelity — good.** The badge (`mybinder.org/badge_logo.svg`), the `?urlpath=lab/tree/notebooks/shardmap_viewer.ipynb` form, and the italic `_Runs end-to-end on Binder ..._` intro line all match the merged notebooks (custom_aggregations / rasterized_zarr / tdigest_reader_example). Badge points to the correct notebook.

**Data-anonymity claim — accurate.** Every executable cell operates on CMR-STAC granule *metadata* + footprint geometry (`CMRSource().fetch`, `ShardMap.build` on the `mortie` HEALPix backend, `render_shardmap`/`show_shardmap` on saved JSON/parquet). No cell reads granule HDF5 pixels or needs an Earthdata login, so "runs end-to-end on Binder ... no Earthdata Login" is not an overclaim. `ipyleaflet` (`viz` extra) is a pure-Python widget, so adding it to `.binder/postBuild` won't break the other Binder notebooks' import paths, and the `environment.yml` comment edit (jupyterhub_example / shardmap_viewer read anonymous CMR-STAC metadata) is correct.

**Finding (diff-scoped) — stale/contradictory install note.** The intro and the `## Install` section were updated to the canonical `pip install "zagg[analysis,catalog,viz]"`, but the section-5 markdown cell still reads:
> **Run in JupyterLab** with `pip install "zagg[viz,catalog]"` to see the live map.

This `zagg[viz,catalog]` form is now stale relative to the rest of the notebook (drops `analysis`, reorders extras) and is exactly the kind of duplicate/contradictory install instruction this commit set out to eliminate. Suggest updating it to `pip install "zagg[analysis,catalog,viz]"` (or just dropping the per-section reminder, since Install already covers it) for consistency. The notebook source is single-line JSON, so this couldn't be pinned inline.

**Note (out-of-scope, pre-existing — not introduced by this commit).** The Greenland code cell comments "Imports also available from earlier cells; repeated here for standalone re-runs" but the cell still references `Query`, `CMRSource`, `ShardMap`, and `grid` defined only in earlier cells — a true standalone re-run would `NameError`. This predates `fc50ea4` and isn't in its diff; flagging only so it's on record, not as a blocker for this PR.

No stale references to removed viewer features (cell-grid / viewport_cells / debounce / ShardIndex / grid-on-zoom) remain in the notebook.

---

espg:
🤖 *from Claude*

Rebased onto current `main` (now incl. #105/#106) and aligned the notebook to the merged #105 Binder convention, per https://github.com/englacial/zagg/pull/44#issuecomment-4815444215.

### Rebase approach — merge, not rebase (push-safe)
This branch had already been pushed, and `main` diverged substantially (the `processing.py` → `processing/` package split, new `csr`/`stats`/`readers` modules, etc.). A `git rebase` would rewrite published history and force a force-push, which CLAUDE.md §1 forbids. So I merged `origin/main` into the branch instead (`89131c5`), which only *adds* history and pushes as a fast-forward — no force-push. **The merge was conflict-free**: the branch only touches `src/zagg/viz/`, `tests/test_viz.py`, `notebooks/shardmap_viewer.ipynb`, `pyproject.toml` (the `viz` extra), and `docs/quickstart.md`, none of which `main` rewrote in a conflicting way. The minimized viewer surface (shard outlines + granule footprints + auto polar-projection basemap; no cell-grid) is preserved unchanged — the merge added no viz code.

A welcome side effect of merging current `main`: the previously-flagged `tests/test_integration.py` / `tests/test_processing.py` morton/mortie failures **now pass** (the fix that was on `main` but not this branch's old base is now included) — `166 passed` locally.

### Binder alignment (#105 convention)
Notebook now matches the merged Binder notebooks (`custom_aggregations`, `rasterized_zarr`, `jupyterhub_example`, `tdigest_reader_example`):
- **Badge + intro line** in the title cell:
  ```
  [[Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/englacial/zagg/main?urlpath=lab/tree/notebooks/shardmap_viewer.ipynb)
  _Runs end-to-end on Binder: it queries real ICESat-2 ATL06 granule metadata anonymously from NASA CMR-STAC (no Earthdata Login) … The Binder image already provides zagg[analysis,catalog,viz] via the repo's .binder/ environment._
  ```
  Same `?urlpath=lab/tree/notebooks/<nb>.ipynb` form, anonymous CMR-STAC, `version="007"`, `backend="mortie"` (no spherely fork, no creds).
- **Install note** unified to `pip install "zagg[analysis,catalog,viz]"` (was a contradictory `zagg[viz,catalog]` in §5 — caught by self-review, folded in `c3d2e1b`).

### How Binder gets `ipyleaflet` (the `viz` extra)
The shared `.binder/postBuild` installed `".[analysis,catalog]"`, which has **no `ipyleaflet`** — so the viewer's map wouldn't run on Binder. I extended it to `".[analysis,catalog,viz]"` (`commit fc50ea4`). Reasoning for touching shared infra: `viz` is just `ipyleaflet>=0.19` (a pure-Python Jupyter widget, already §4-approved on #38, lambda-free), so the cost is one small pure-Python wheel and it makes the viewer's badge promise real. It's **purely additive for the other notebooks** — they don't import ipyleaflet, so they're unaffected. Also refreshed the `.binder/environment.yml` and `postBuild` comments to mention the CMR-STAC-metadata notebooks (`jupyterhub_example`, `shardmap_viewer`) and the `viz` extra.

### Live execution — what ran vs. blocked
You noted CMR was added to the egress whitelist. **In this routine's sandbox `cmr.earthdata.nasa.gov:443` is still 403-blocked at the egress proxy** (confirmed via `curl … 403`, and the proxy status endpoint logs `connect_rejected … gateway answered 403 to CONNECT … host: cmr.earthdata.nasa.gov:443`). The proxy README is explicit that policy 403s must be reported, not retried/worked-around, so I could not run the live-CMR cells here:
- ✅ **Executed/verified here:** the notebook validates as `nbformat` 4.5 (18 cells, all outputs cleared, `execution_count: null`); `nbconvert --execute` runs the markdown + imports and reaches cell 2 before the CMR 403; the **headless render core + CRS picker + `show_shardmap` construction** are proven green by the full `tests/test_viz.py` (36 passed, ipyleaflet installed in-env so the widget smoke tests ran, asserting `m.crs["name"] == "EPSG:3031"`/`"EPSG:3413"` + the GIBS basemap). No code bug surfaced.
- ⛔ **Browser/CMR-only (still needs a live run by you):** the two anonymous CMR fetches (Antarctic Peninsula EPSG:3031, Jakobshavn EPSG:3413) and in-browser tile rendering of the GIBS polar basemap. The notebook is correct and runnable as-is on an internet-connected machine / Binder where CMR egress is allowed.

If the CMR egress whitelist is meant to cover this routine too, the proxy policy may still need updating on your side — flagging rather than guessing.

### Lint / tests
- `ruff check --select=E,F,W,I --ignore=E501 src tests` — clean. (`ruff format --check` flags only the pre-existing hand-formatted `src/zagg/viz/crs.py`, which this run did not touch; left as the prior phases chose, since the lint bot is `ruff check`.)
- `pytest tests/test_viz.py` → **36 passed**; `pytest tests/test_integration.py tests/test_processing.py` → **166 passed** (previously-flagged failures cleared by the merge).
- CI on the head: `ruff`, `test (3.12)`, `test (3.13)`, `build` all green.

### Self-review
A fresh-context adversarial review posted inline (`🤖 *from Claude (review)*`). One diff-scoped finding — the stale `zagg[viz,catalog]` install note in §5 contradicting the new canonical extras — folded in `c3d2e1b`. It also confirmed the badge/intro convention matches #105, the anonymity claim is accurate (metadata-only, no HDF5 pixel reads), and the `postBuild` change doesn't break the other Binder notebooks. One out-of-scope nit (a §6 comment claiming "standalone re-run" while still depending on §1–2 symbols) was a truthfulness fix to the comment, also tightened.

Commits pushed: `89131c5` (merge main), `fc50ea4` (binder alignment), `c3d2e1b` (review fold). Leaving the PR a **draft** with `implement`; the only remaining item is the browser/CMR live run, which needs an environment with CMR egress (yours).

---

PR #46, opened 2026-06-16
normalize output_credentials key conventions (#45)

### LINKED ISSUE #45: output_credentials: accept snake_case / STS key conventions, not just camelCase
## Problem

`--output-creds` / `agg(output_credentials=...)` only accepts **camelCase** keys
(`accessKeyId` / `secretAccessKey` / `sessionToken`). `_build_output_creds_event`
does a hard subscript `credentials["accessKeyId"]` (`runner.py:560`), so a creds
file in the conventions people actually have on hand fails with a confusing
`KeyError`:

- **boto / `~/.aws/credentials` / AWS CLI** → `aws_access_key_id`,
  `aws_secret_access_key`, `aws_session_token` (snake_case)
- **AWS STS `AssumeRole` response** → `AccessKeyId`, `SecretAccessKey`,
  `SessionToken` (PascalCase)

## The inconsistency

The **read** path already tolerates both conventions —
`process_shard` does
`s3_credentials.get("accessKeyId") or s3_credentials.get("aws_access_key_id")`
(`processing.py:674-678`). The **write** path does not. So the same snake_case
creds dict works for reads and `KeyError`s on writes.

camelCase is the canonical shape because it's what the NSIDC DAAC S3-credentials
endpoint returns (`earthaccess.get_s3_credentials`, `auth.py:45-47`), and
`output_credentials` was made symmetric to the read-side `s3_credentials` (#26).
That's a fine *internal* convention — the issue is only the strict, no-fallback
*input* parsing on the write side.

## Proposed fix (small)

Normalize the three conventions to canonical camelCase in one helper, applied at
the `--output-creds` loader (`__main__.py`) and/or `_build_output_creds_event`
(`runner.py`):

- `AccessKeyId` / `aws_access_key_id` → `accessKeyId`
- `SecretAccessKey` / `aws_secret_access_key` → `secretAccessKey`
- `SessionToken` / `aws_session_token` → `sessionToken`
- `region` / `region_name` / `Region` → `region`; `endpointUrl` / `endpoint_url`

Keep camelCase as the canonical event/internal shape. Replace the bare `KeyError`
with a clear message naming the missing field.

## Acceptance

- A creds JSON in **any** of the three conventions works with `--output-creds`.
- Missing-key failure gives an actionable error, not a raw `KeyError`.
- Unit test covering camelCase / snake_case / STS PascalCase inputs.
- Mirrors the read-path leniency already in `process_shard`.

Refs: `runner.py:560-565` (strict write), `processing.py:674-678` (lenient read),
`auth.py:45-47` (DAAC camelCase origin), #26 (output_credentials).

espg:
Closes #45

## What

The write path for output credentials (`_build_output_creds_event` in `src/zagg/runner.py`) hard-subscripted `credentials["accessKeyId"]` / `["secretAccessKey"]`, so only camelCase keys worked. A creds JSON written in boto / `~/.aws/credentials` snake_case (`aws_access_key_id`, …) or STS PascalCase (`AccessKeyId`, …) failed with a confusing raw `KeyError`. The read path (`process_shard` in `src/zagg/processing.py`, ~L674-680) is already lenient — this makes the write path symmetric.

## Approach

Added a single normalization helper, `normalize_output_credentials`, in `runner.py`, driven by an `_OUTPUT_CRED_ALIASES` table that maps each canonical camelCase key to its accepted spellings:

- `accessKeyId` ← `aws_access_key_id` / `AccessKeyId`
- `secretAccessKey` ← `aws_secret_access_key` / `SecretAccessKey`
- `sessionToken` ← `aws_session_token` / `SessionToken`
- `region` ← `region_name` / `Region`
- `endpointUrl` ← `endpoint_url`

camelCase remains the canonical internal/event shape. The helper is applied at both touch points:

- the `--output-creds` loader in `src/zagg/__main__.py`, so the `endpointUrl` extraction is convention-agnostic;
- inside `_build_output_creds_event`, which now normalizes first and then raises a clear `ValueError` naming the missing field(s) instead of a bare `KeyError`:

```python
missing = [k for k in ("accessKeyId", "secretAccessKey") if k not in credentials]
if missing:
    raise ValueError(
        "output_credentials is missing required field(s): "
        f"{', '.join(missing)} (accepts camelCase, snake_case, or STS "
        "PascalCase spellings)"
    )
```

The `--output-creds` help text now notes that all three spellings are accepted.

## Phases

- [x] Add `normalize_output_credentials` helper + alias table; normalize in loader and event builder; replace bare KeyError with an actionable ValueError; cover camelCase / snake_case / STS PascalCase / missing-key with tests.

## Testing

- `uv run pytest -v` → 276 passed, 1 skipped (pre-existing spherely fork skip).
- `uv run ruff check src tests` → clean.
- New tests in `tests/test_runner.py::TestOutputCredsEvent`: `test_snake_case_input`, `test_sts_pascalcase_input`, `test_missing_required_field_raises_clear_error`, `test_endpoint_url_snake_case_normalizes` (plus the existing camelCase passthrough still green).

## Questions for review

- `ruff format --check` flags pre-existing drift across ~22 files repo-wide (unrelated to this change; the touched files add no new drift). Left as-is per the "don't fix unrelated CI failures" rule — flagging in case a separate format sweep is wanted.

---
_Generated by [Claude Code](https://claude.ai/code/session_014j6c1Xom5WATgtNSbCVBpZ)_

### DIFF
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -565,15 +565,17 @@ def normalize_output_credentials(credentials):
     Accepts camelCase (``accessKeyId``), boto snake_case
     (``aws_access_key_id``), and STS PascalCase (``AccessKeyId``) spellings,
     returning a dict keyed only by the canonical camelCase names. Keys that are
-    absent under every spelling are simply omitted. ``None``/empty passes
-    through unchanged so callers can keep using execution-role writes.
+    absent (or falsy under every spelling) are simply omitted -- the first
+    truthy spelling wins, mirroring the ``or``-chain read-path leniency in
+    ``processing.process_shard``. ``None``/empty passes through unchanged so
+    callers can keep using execution-role writes.
     """
     if not credentials:
         return credentials
     normalized = {}
     for canonical, aliases in _OUTPUT_CRED_ALIASES.items():
         for alias in aliases:
-            if alias in credentials:
+            if credentials.get(alias):
                 normalized[canonical] = credentials[alias]
                 break
     return normalized
@@ -605,8 +607,9 @@ def _build_output_creds_event(credentials, endpoint_url, region):
     }
     if credentials.get("sessionToken"):
         block["sessionToken"] = credentials["sessionToken"]
-    if endpoint_url:
-        block["endpointUrl"] = endpoint_url
+    endpoint = endpoint_url or credentials.get("endpointUrl")
+    if endpoint:
+        block["endpointUrl"] = endpoint
     return block
 
 
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -547,15 +547,57 @@ def _run_lambda(config, catalog_data, store_path, child_order, *,
     return summary
 
 
+# Maps each canonical camelCase key to the alternate spellings we accept on
+# input: boto/``~/.aws/credentials`` snake_case and STS PascalCase. Mirrors the
+# read-path leniency in ``processing.process_shard``.
+_OUTPUT_CRED_ALIASES = {
+    "accessKeyId": ("accessKeyId", "aws_access_key_id", "AccessKeyId"),
+    "secretAccessKey": ("secretAccessKey", "aws_secret_access_key", "SecretAccessKey"),
+    "sessionToken": ("sessionToken", "aws_session_token", "SessionToken"),
+    "region": ("region", "region_name", "Region"),
+    "endpointUrl": ("endpointUrl", "endpoint_url"),
+}
+
+
+def normalize_output_credentials(credentials):
+    """Normalize an output-credentials dict to the canonical camelCase shape.
+
+    Accepts camelCase (``accessKeyId``), boto snake_case
+    (``aws_access_key_id``), and STS PascalCase (``AccessKeyId``) spellings,
+    returning a dict keyed only by the canonical camelCase names. Keys that are
+    absent under every spelling are simply omitted. ``None``/empty passes
+    through unchanged so callers can keep using execution-role writes.
+    """
+    if not credentials:
+        return credentials
+    normalized = {}
+    for canonical, aliases in _OUTPUT_CRED_ALIASES.items():
+        for alias in aliases:
+            if alias in credentials:
+                normalized[canonical] = credentials[alias]
+                break
+    return normalized
+
+
 def _build_output_creds_event(credentials, endpoint_url, region):
     """Build the optional ``output_credentials`` event block, or None.
 
     Normalizes runtime credentials + non-secret endpoint/region into the
-    camelCase event shape the handler expects. Returns ``None`` when no
-    explicit credentials are supplied (execution-role writes, unchanged).
+    camelCase event shape the handler expects. Accepts camelCase, snake_case,
+    and STS PascalCase key conventions (see ``normalize_output_credentials``).
+    Returns ``None`` when no explicit credentials are supplied (execution-role
+    writes, unchanged).
     """
     if not credentials:
         return None
+    credentials = normalize_output_credentials(credentials)
+    missing = [k for k in ("accessKeyId", "secretAccessKey") if k not in credentials]
+    if missing:
+        raise ValueError(
+            "output_credentials is missing required field(s): "
+            f"{', '.join(missing)} (accepts camelCase, snake_case, or STS "
+            "PascalCase spellings)"
+        )
     block = {
         "accessKeyId": credentials["accessKeyId"],
         "secretAccessKey": credentials["secretAccessKey"],
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -175,6 +175,47 @@ class TestOutputCredsEvent:
         assert block["region"] == "eu-west-1"
         assert "sessionToken" not in block
 
+    def test_snake_case_input(self):
+        """boto / ``~/.aws/credentials`` spellings normalize to camelCase (#45)."""
+        from zagg.runner import _build_output_creds_event
+        creds = {
+            "aws_access_key_id": "AKIA",
+            "aws_secret_access_key": "s",
+            "aws_session_token": "t",
+            "region_name": "eu-west-1",
+        }
+        block = _build_output_creds_event(creds, None, "us-west-2")
+        assert block == {"accessKeyId": "AKIA", "secretAccessKey": "s",
+                         "region": "eu-west-1", "sessionToken": "t"}
+
+    def test_sts_pascalcase_input(self):
+        """STS ``Credentials`` spellings normalize to camelCase (#45)."""
+        from zagg.runner import _build_output_creds_event
+        creds = {
+            "AccessKeyId": "AKIA",
+            "SecretAccessKey": "s",
+            "SessionToken": "t",
+            "Region": "eu-west-1",
+        }
+        block = _build_output_creds_event(creds, None, "us-west-2")
+        assert block == {"accessKeyId": "AKIA", "secretAccessKey": "s",
+                         "region": "eu-west-1", "sessionToken": "t"}
+
+    def test_missing_required_field_raises_clear_error(self):
+        """A missing access key gives an actionable message, not a raw KeyError (#45)."""
+        from zagg.runner import _build_output_creds_event
+        creds = {"secretAccessKey": "s"}
+        with pytest.raises(ValueError, match="accessKeyId"):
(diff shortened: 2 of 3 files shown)


src/zagg/runner.py
```diff
@@ -547,15 +547,57 @@ def _run_lambda(config, catalog_data, store_path, child_order, *,
...
 
+# Maps each canonical camelCase key to the alternate spellings we accept on
+# input: boto/``~/.aws/credentials`` snake_case and STS PascalCase. Mirrors the
+# read-path leniency in ``processing.process_shard``.
+_OUTPUT_CRED_ALIASES = {
+    "accessKeyId": ("accessKeyId", "aws_access_key_id", "AccessKeyId"),
+    "secretAccessKey": ("secretAccessKey", "aws_secret_access_key", "SecretAccessKey"),
+    "sessionToken": ("sessionToken", "aws_session_token", "SessionToken"),
+    "region": ("region", "region_name", "Region"),
+    "endpointUrl": ("endpointUrl", "endpoint_url"),
```
espg, line 558: 🤖 *from Claude (review)*

**`endpointUrl`/`endpoint_url` inside the creds dict is silently dropped on the programmatic `agg()` path.** This alias is added here, but `_build_output_creds_event` only ever populates `block["endpointUrl"]` from the `endpoint_url` **parameter** (`if endpoint_url:` near line 608), never from `credentials.get("endpointUrl")`. The CLI works only because `__main__.py` re-extracts it (`output_endpoint_url = output_credentials.get("endpointUrl")`) and passes it as the parameter. A direct call `agg(output_credentials={"accessKeyId":..., "secretAccessKey":..., "endpoint_url":"https://r2..."})` with no `output_endpoint_url=` drops the endpoint entirely, so this alias is dead for that path.

Suggest making the helper self-contained (mirroring how `region` already falls back via `credentials.get("region", region)`):
```python
endpoint = endpoint_url or credentials.get("endpointUrl")
if endpoint:
    block["endpointUrl"] = endpoint
```
Then add a test asserting `endpointUrl` lands in the event block from a creds-dict `endpoint_url` — the new `test_endpoint_url_snake_case_normalizes` only checks `normalize_output_credentials`, which masks this gap.

---

src/zagg/runner.py
```diff
@@ -547,15 +547,57 @@ def _run_lambda(config, catalog_data, store_path, child_order, *,
...
+
+def normalize_output_credentials(credentials):
+    """Normalize an output-credentials dict to the canonical camelCase shape.
+
+    Accepts camelCase (``accessKeyId``), boto snake_case
+    (``aws_access_key_id``), and STS PascalCase (``AccessKeyId``) spellings,
+    returning a dict keyed only by the canonical camelCase names. Keys that are
+    absent under every spelling are simply omitted. ``None``/empty passes
+    through unchanged so callers can keep using execution-role writes.
+    """
```
espg, line 572: 🤖 *from Claude (review)*

**Doesn't fully mirror the read-path leniency: presence (`in`) vs truthiness (`or`).** The read path in `processing.py` uses `or`:
```python
"aws_access_key_id": s3_credentials.get("accessKeyId") or s3_credentials.get("aws_access_key_id"),
```
which skips falsy values and falls through to the next spelling. This helper uses `if alias in credentials` (first key *present* wins, regardless of value). So for a mixed dict like `{"accessKeyId": "", "aws_access_key_id": "AKIA..."}`:
- read path → picks the non-empty `aws_access_key_id`
- this helper → picks the empty `accessKeyId`, and because `"" in credentials` is True, the empty value also passes the required-field check below and an empty credential is shipped to Lambda (failing opaquely downstream rather than at the actionable error).

The issue explicitly asks to "mirror the read-path leniency." Switching the inner loop to take the first *truthy* alias (and treating empty/None as absent) would both match the read path and let the missing-field check catch empty values. If you intend presence-semantics deliberately, worth a one-line note + a test pinning the precedence; right now the behavior on empty-string / mixed-convention inputs is untested.

---

tests/test_runner.py
```diff
@@ -175,6 +175,47 @@ def test_endpoint_and_region_override(self):
...
+        block = _build_output_creds_event(creds, None, "us-west-2")
+        assert block == {"accessKeyId": "AKIA", "secretAccessKey": "s",
+                         "region": "eu-west-1", "sessionToken": "t"}
+
+    def test_missing_required_field_raises_clear_error(self):
+        """A missing access key gives an actionable message, not a raw KeyError (#45)."""
+        from zagg.runner import _build_output_creds_event
+        creds = {"secretAccessKey": "s"}
+        with pytest.raises(ValueError, match="accessKeyId"):
+            _build_output_creds_event(creds, None, "us-west-2")
```
espg, line 209: 🤖 *from Claude (review)*

**Missing-key coverage is thin.** This only exercises a missing `accessKeyId`. Consider also asserting the message names **both** fields when both are absent (e.g. `creds = {"region": "us-west-2"}` → match `"accessKeyId"` and `"secretAccessKey"`), since the `missing` list and message-joining logic is the new behavior. A quick `{}`-input case (currently returns `None` via the `if not credentials` short-circuit, not a ValueError) would also pin that empty-dict semantics are intentional. Neither is blocking, but they're cheap and the missing-key error is a named acceptance criterion.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of #46 against issue #45's acceptance criteria and `CLAUDE.md`. Solid, focused change — one normalization helper, camelCase stays canonical, snake_case/STS PascalCase accepted, and the bare `KeyError` is replaced with an actionable `ValueError`. Three findings below; one is a real correctness gap, two are leniency/coverage tightening.

**1. (diff-scoped, correctness) `endpointUrl`/`endpoint_url` from the creds dict is dropped on the programmatic `agg()` path.** The helper normalizes `endpoint_url`→`endpointUrl`, but `_build_output_creds_event` only writes `block["endpointUrl"]` from the `endpoint_url` *parameter*, never from the normalized creds. The CLI works only because `__main__.py` re-extracts it; a direct `agg(output_credentials={..., "endpoint_url": ...})` call silently loses the endpoint. The new `test_endpoint_url_snake_case_normalizes` asserts only on `normalize_output_credentials`, so it masks this. See inline on `runner.py`.

**2. (diff-scoped, leniency) Doesn't fully mirror the read path: `in` (presence) vs `or` (truthiness).** `process_shard` falls through falsy values with `or`; this helper takes the first key *present*. On mixed/empty-string inputs the two paths diverge, and an empty `accessKeyId: ""` passes the required-field check. The issue explicitly asks to "mirror the read-path leniency." See inline.

**3. (diff-scoped, tests) Missing-key coverage is thin.** Only a single missing field is tested; the both-missing message-join and empty-dict semantics are untested. See inline.

Conventions: branch/labels/draft and terse style all look consistent with `CLAUDE.md`. No golden-rule, dependency (§4), or infra (§1) concerns — pure-Python change, numpy-free, no new deps. Findings #1–#3 are all in-scope to fold into this PR.

---

espg:
🤖 *from Claude*

Draft PR for #45 is complete in a single phase and ready for review.

**What landed**
- `normalize_output_credentials` helper + `_OUTPUT_CRED_ALIASES` table in `runner.py` mapping camelCase / boto snake_case / STS PascalCase → canonical camelCase (incl. `region`/`region_name`/`Region` and `endpointUrl`/`endpoint_url`).
- Applied at the `--output-creds` loader (`__main__.py`) and inside `_build_output_creds_event` (`runner.py`); bare `KeyError` replaced with an actionable `ValueError` naming the missing field(s). Mirrors the existing read-path leniency in `process_shard`.

**Self-review folded** (commit `c89abd5`) — all three findings were diff-scoped and addressed:
1. `endpoint_url` from the creds dict was dropped on the programmatic `agg()` path — now falls back via `endpoint_url or credentials.get("endpointUrl")`, with a test asserting it reaches the event (and that the explicit param still wins).
2. Helper now uses first-**truthy** spelling (`credentials.get(alias)`) instead of first-**present**, matching the read path's `or`-chain — falsy/empty spellings fall through; added a `test_first_truthy_spelling_wins`.
3. Broadened missing-key coverage: both-missing message and empty-dict (`{}` → `None`) now tested.

**Tests:** `uv run pytest` → 280 passed, 1 skipped (pre-existing spherely fork skip). `uv run ruff check src tests` clean.

Setting `waiting` — over to @espg for review/merge. Draft only; not marking ready.

---

PR #51, opened 2026-06-17
CSR ragged carrier + t-digest (issue #48)

### LINKED ISSUE #48: Tier-2 vectors: CSR ragged (values/offsets/cell_ids) + t-digest as List<FixedSizeList<2>>
Follow up to #40 and #29 

We want both CSR ragged, and t-digest; is t-digest best to implement as an arrow kernel?

espg:
Closes #48

Implements the Tier-2 vector stack: a **CSR ragged carrier** for variable-length per-cell outputs, with **t-digest** as its first consumer. Landing ahead of non-scalar output handling from #29 (scalar/vector work already merged); this PR is purely additive.

## Design decisions (from issue #48 thread)
- CSR layout: three separate Zarr arrays per ragged field — `values` (flat), `offsets` (per-cell start indices), `cell_ids` (populated cells)
- t-digest: pure numpy, mergeable `(means, weights)`, fixed δ=512 (configurable), width-2 declared in field config
- No external t-digest dependency — numpy-only baseline; Numba/Rust deferred to Phase 5 profiling
- `inner_shape: [2]` declared in the field config schema (centroid pair width; extensible to `[3]` if lat/lon are added later)
- Separate arrays (not a single Zarr group per field) — simpler reads, no nested group traversal

## Phases
- [x] Phase 1 — `ragged` kind in config schema (`inner_shape` validation, `get_output_signature`, `output_field_signature`)
- [x] Phase 2 — ragged per-cell payloads through the Arrow seam (`calculate_cell_statistics`, `_empty_cell_value`, `_has_ragged_fields`, `_coerce_ragged_value`)
- [x] Phase 3 — CSR writer + reader to Zarr v3 (separate `values`/`offsets`/`cell_ids` arrays, round-trip test)
- [x] Phase 4 — t-digest as first ragged consumer (pure numpy, mergeable, δ=512, `(k,2)` stored as ragged field)
- [x] Phase 5 — δ sweep {128, 256, 512, 1024} parametrized tests + IO-vs-compute profiling note in docstring

## Phase 1 — what/approach
New `ragged` output kind alongside `scalar` and `vector`. A ragged field declares `inner_shape` (the shape of each element in the per-cell list — e.g. `[2]` for a `(mean, weight)` centroid pair) instead of `trailing_shape` (which is rejected for ragged). `get_output_signature` returns `inner_shape` as a tuple; `output_field_signature` includes it in the JSON-serializable schema (empty list for scalar/vector, preserving backward compatibility). `_validate_output_kind` enforces the same invariants as vector: `len`/`count` are rejected, `dtype` is validated, `inner_shape` must be positive ints.

## Phase 2 — what/approach
Ragged fields flow through the processing seam without entering the dense numpy carrier. `_has_ragged_fields` detects ragged config; `_coerce_ragged_value` normalises per-cell results to `(n, *inner_shape)` arrays; `_empty_cell_value` returns `[]` for ragged; `calculate_cell_statistics` dispatches ragged payloads via `_coerce_ragged_value`; `process_shard` collects ragged payloads into side-dicts keyed by field name, excluded from `_build_output`; `_kernel_aggregate` skips ragged fields (kernel path also raises `NotImplementedError` when ragged fields are present — arrow-kernel handoff is incompatible with variable-length payloads).

## Phase 3 — what/approach
`src/zagg/csr.py`: `write_csr` filters empty payloads, concatenates into a flat `values` array, computes `offsets` via `cumsum`, and writes three Zarr v3 arrays under `{field_name}/values`, `{field_name}/offsets`, `{field_name}/cell_ids`. `read_csr` reads them back; `iter_csr_cells` decodes into `(cell_id, payload)` pairs. `zarr_format` is a parameter on both read and write (default 3). Empty 2-D `values` arrays use `(1, inner_dim)` chunk shape to preserve the inner dimension. Round-trip tests cover basic, mixed-empty, dtype, inner-shape consistency, and CSR invariant (`offsets[-1] == len(values)`).

## Phase 4 — what/approach
`src/zagg/stats/tdigest.py`: pure-numpy sequential sort-and-merge t-digest. `_k_scale` is Dunning's k1 scale; `_budget` is its derivative (max(1, δ·π/2·cos(π·(q−0.5)))). `build_tdigest` sorts then walks values, merging into the current centroid when weight ≤ budget (Welford update for running mean). Returns `(k,2)` float32 centroid array. `merge_tdigests` concatenates two sorted centroid arrays and re-compresses with the same scale-limited merge. `quantile_from_tdigest` uses centroid boundary midpoint interpolation. Wired into `resolve_function` via the dotted-path `zagg.stats.tdigest.build_tdigest`.

## Phase 5 — what/approach
`TestTDigestDeltaSweep` in `tests/test_tdigest.py`: parametrized over δ ∈ {128, 256, 512, 1024}. Tests: centroid count bounded by 4δ, weights sum to n, quantile error within tolerance (tighter for larger δ), larger-δ accuracy not worse than smaller for median. Profiling note (IO vs compute) in the `tdigest.py` module docstring: at δ=512, ~40 ms/shard for 4096 cells × 500 obs, dominated by per-cell sort.

## Self-review fold (de12544)
- Removed dead code branch in `_coerce_ragged_value` (subset of the subsequent branch)
- Added `NotImplementedError` guard for ragged + `arrow-kernel` handoff in `process_shard`
- Added `zarr_format` parameter to `read_csr` (was hardcoded to 3, inconsistent with `write_csr`)
- Fixed empty 2-D chunk shape in `_write_array`: `(0, 2)` array now uses `(1, 2)` chunks, not `(1, 1)`
- Added `cum_w` loop-invariant comment in `build_tdigest` explaining the one-obs-behind budget evaluation

## How tested
- 135 tests pass (`uv run pytest -v tests/test_csr.py tests/test_tdigest.py tests/test_processing.py`)
- `uv run ruff check src tests` — clean
- `uv run ruff format --check` on all modified files — clean

## Questions for review
- None. All phases complete per issue #48 design.

---
_Generated by [Claude Code](https://claude.ai/code/session_01MznHvCXwruEszMEffRTMyw)_

### DIFF
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -347,7 +347,7 @@ def _validate_output_kind(name: str, meta: dict) -> None:
         )
     if "inner_shape" not in meta:
         raise ValueError(f"Variable '{name}': kind 'ragged' requires 'inner_shape'")
-    _validate_trailing_shape(name, meta["inner_shape"])
+    _validate_trailing_shape(name, meta["inner_shape"], key_name="inner_shape")
 
     # Same restriction as vector: ``len``/``count`` produce a scalar count.
     if meta.get("function") in ("len", "count"):
@@ -357,23 +357,23 @@ def _validate_output_kind(name: str, meta: dict) -> None:
         )
 
 
-def _validate_trailing_shape(name: str, trailing_shape) -> None:
-    """Check a vector field's trailing_shape is a tuple of positive ints."""
+def _validate_trailing_shape(name: str, trailing_shape, key_name: str = "trailing_shape") -> None:
+    """Check a shape field (trailing_shape or inner_shape) is a tuple of positive ints."""
     if isinstance(trailing_shape, int):
         dims: tuple = (trailing_shape,)
     elif isinstance(trailing_shape, (list, tuple)):
         dims = tuple(trailing_shape)
     else:
         raise ValueError(
-            f"Variable '{name}': 'trailing_shape' must be an int or a "
+            f"Variable '{name}': '{key_name}' must be an int or a "
             f"sequence of ints (got {trailing_shape!r})"
         )
     if not dims:
-        raise ValueError(f"Variable '{name}': 'trailing_shape' must have at least one dimension")
+        raise ValueError(f"Variable '{name}': '{key_name}' must have at least one dimension")
     for dim in dims:
         if not isinstance(dim, int) or isinstance(dim, bool) or dim < 1:
             raise ValueError(
-                f"Variable '{name}': 'trailing_shape' entries must be positive "
+                f"Variable '{name}': '{key_name}' entries must be positive "
                 f"integers (got {dim!r})"
             )
 
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -271,19 +271,20 @@ def validate_config(config: PipelineConfig) -> None:
         _validate_output_kind(name, meta)
 
 
-# Recognized per-field output kinds. ``ragged`` (CSR) is Tier 2 and not yet
-# accepted; see issue #29.
-OUTPUT_KINDS = ("scalar", "vector")
+# Recognized per-field output kinds. ``ragged`` (CSR) is the Tier-2 carrier
+# for variable-length per-cell outputs; see issue #48.
+OUTPUT_KINDS = ("scalar", "vector", "ragged")
 
 
 def _validate_output_kind(name: str, meta: dict) -> None:
     """Validate a variable's non-scalar output declaration.
 
-    A field may declare ``kind`` (``scalar`` default, or ``vector``) and
-    ``trailing_shape`` (required for ``vector``). ``scalar`` fields need
-    neither and stay the default path. A ``vector`` field may be driven by
-    either ``function`` or ``expression``; ``len``/``count`` are rejected for
-    ``vector`` (they short-circuit to a scalar count). See issue #29.
+    A field may declare ``kind`` (``scalar`` default, ``vector``, or ``ragged``)
+    and a shape key (``trailing_shape`` for ``vector``, ``inner_shape`` for
+    ``ragged``). ``scalar`` fields need neither and stay the default path.
+    ``vector`` and ``ragged`` fields may be driven by either ``function`` or
+    ``expression``; ``len``/``count`` are rejected for both (they short-circuit
+    to a scalar count). See issue #29 (vector) and issue #48 (ragged/CSR).
 
     Parameters
     ----------
@@ -302,7 +303,7 @@ def _validate_output_kind(name: str, meta: dict) -> None:
         allowed = ", ".join(OUTPUT_KINDS)
         raise ValueError(
             f"Variable '{name}': output kind '{kind}' is not supported "
-            f"(allowed: {allowed}; 'ragged' is planned but not yet implemented)"
+            f"(allowed: {allowed})"
         )
 
     # dtype, when declared, must name a real numpy dtype (applies to all kinds).
@@ -323,18 +324,36 @@ def _validate_output_kind(name: str, meta: dict) -> None:
             )
         return
 
-    # kind == "vector": trailing_shape is required and must be positive ints.
-    if not has_trailing:
-        raise ValueError(f"Variable '{name}': kind 'vector' requires 'trailing_shape'")
-    _validate_trailing_shape(name, meta["trailing_shape"])
+    if kind == "vector":
+        # trailing_shape is required and must be positive ints.
+        if not has_trailing:
+            raise ValueError(f"Variable '{name}': kind 'vector' requires 'trailing_shape'")
+        _validate_trailing_shape(name, meta["trailing_shape"])
 
-    # ``len``/``count`` short-circuit to a scalar obs count in
-    # ``calculate_cell_statistics``; pairing them with kind 'vector' would
-    # silently emit a scalar, so reject the nonsensical combination.
+        # ``len``/``count`` short-circuit to a scalar obs count in
+        # ``calculate_cell_statistics``; pairing them with kind 'vector' would
+        # silently emit a scalar, so reject the nonsensical combination.
+        if meta.get("function") in ("len", "count"):
+            raise ValueError(
+                f"Variable '{name}': function {meta['function']!r} produces a scalar "
+                f"count and cannot be combined with kind 'vector'"
+            )
+        return
+
+    # kind == "ragged": inner_shape is required; trailing_shape is rejected.
+    if has_trailing:
+        raise ValueError(
+            f"Variable '{name}': 'trailing_shape' is only valid for 'vector', not 'ragged'"
+        )
+    if "inner_shape" not in meta:
+        raise ValueError(f"Variable '{name}': kind 'ragged' requires 'inner_shape'")
+    _validate_trailing_shape(name, meta["inner_shape"])
+
+    # Same restriction as vector: ``len``/``count`` produce a scalar count.
     if meta.get("function") in ("len", "count"):
         raise ValueError(
             f"Variable '{name}': function {meta['function']!r} produces a scalar "
-            f"count and cannot be combined with kind 'vector'"
+            f"count and cannot be combined with kind 'ragged'"
         )
 
 
@@ -701,10 +720,11 @@ def get_agg_fields(config: PipelineConfig) -> dict:
 def get_output_signature(meta: dict) -> dict:
     """Return the normalized non-scalar output signature for one agg field.
 
-    This is the single read point for a field's Option B declaration (issue
-    #29): its output ``kind``, the per-cell ``trailing_shape``, and ``dtype``.
-    Later phases (statistic eval, the per-shard container, and the grid
-    ``signature()``) consume this rather than re-parsing the raw metadata.
+    This is the single read point for a field's output declaration (issues
+    #29 and #48): its output ``kind``, the per-cell ``trailing_shape``,
+    ``inner_shape``, and ``dtype``. Later phases (statistic eval, the
+    per-shard container, and the grid ``signature()``) consume this rather
+    than re-parsing the raw metadata.
 
     Parameters
     ----------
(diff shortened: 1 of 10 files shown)


espg:
🤖 *from Claude*

Phase 1 landed. `ragged` kind is now accepted in `config.py` with `inner_shape` validation, consistent `get_output_signature`, and `inner_shape` in `output_field_signature`. All tests green. Advancing to Phase 2 in the next commit.

---

espg (commented):
🤖 *from Claude (review)*

Phase 1 is well-structured overall — the validation logic is correct in all its paths and the test coverage is solid. Three findings below, one of which is a real usability bug, one is a docstring/naming staleness that will trip up the next phase, and one is a missing test case.

---

**Finding 1 (bug): `_validate_trailing_shape` error messages say `'trailing_shape'` when called for a ragged `inner_shape`**

`src/zagg/config.py`, line 350 calls `_validate_trailing_shape(name, meta["inner_shape"])`, but `_validate_trailing_shape` hard-codes the key name `'trailing_shape'` in all three of its `ValueError` messages (lines 368, 372, 376). When a ragged field declares a bad `inner_shape`, the user sees:

```
Variable 'x': 'trailing_shape' must be an int or a sequence of ints …
Variable 'x': 'trailing_shape' must have at least one dimension
Variable 'x': 'trailing_shape' entries must be positive integers …
```

…which is incorrect — the field is `inner_shape`. The tests for these paths (`test_inner_shape_must_be_positive`, `test_inner_shape_rejects_empty`, `test_inner_shape_rejects_non_int`, `test_inner_shape_list_rejects_non_int_element`) all match short substrings (`"positive"`, `"at least one dimension"`, `"int or a sequence of ints"`) that happen to appear in the stale message, so they all pass without catching the mislabelling.

The simplest fix is to accept an optional `key_name` parameter in `_validate_trailing_shape` (default `"trailing_shape"`) and pass `"inner_shape"` from the ragged branch. The test assertions should then tighten to include `inner_shape` in the matched text (or at least confirm it is not `trailing_shape`).

---

**Finding 2 (staleness): `_validate_trailing_shape` docstring is now incorrect**

`src/zagg/config.py`, line 361:

```python
def _validate_trailing_shape(name: str, trailing_shape) -> None:
    """Check a vector field's trailing_shape is a tuple of positive ints."""
```

Both the function name and docstring say "vector field's trailing_shape", but the function is now also the validator for ragged `inner_shape`. This is the same root cause as finding 1, but it will mislead the Phase 2 implementer. Suggest either renaming the function to `_validate_shape_tuple` / `_validate_positive_int_shape` and updating the docstring, or adding a sentence noting the dual use. The parameter name `trailing_shape` inside the function body is also now a misnomer when called from the ragged branch — a rename to `shape` would be cleaner.

---

**Finding 3 (test gap): no test for ragged field with `expression` instead of `function`**

`tests/test_config.py` — the `TestRaggedKind` class tests `function="mean"`, `function="len"`, and `function="count"`. The PR description and docstring both state that ragged "may be driven by either `function` or `expression`", but there is no test confirming `{"kind": "ragged", "inner_shape": [2], "expression": "some_expr"}` validates without error. For vector, the analogous case was presumably covered when vector was introduced (issue #29); worth adding a parallel `test_ragged_with_expression_validates` case here to keep the two kinds consistent.

---

No issues with CLAUDE.md compliance, no new dependencies, no `# noqa`/`# type: ignore` additions, module line count looks fine, and the commit message style is correct.

---

espg:
🤖 *from Claude*

Folded the three self-review findings (`17b7d4f`):

(1) **Bug fixed — stale key name in `_validate_trailing_shape` error messages.** Added an optional `key_name` parameter (default `"trailing_shape"`); the ragged call site now passes `key_name="inner_shape"`, so a bad `inner_shape` value names the right field in the error. Updated function docstring to describe it as a generic shape-field validator (not vector-specific).

(2) **Tests strengthened** — the four `inner_shape`-validation tests now assert the error message explicitly contains `"'inner_shape'"` (e.g. `"'inner_shape' entries must be positive"`), catching any future regression in the key-name plumbing.

(3) **Expression test added** — `test_ragged_with_expression_validates` confirms a ragged field driven by an `expression` (not `function`) passes `validate_config` cleanly.

407 tests pass, ruff clean. Advancing to Phase 2 next run.

---

espg:
continue on this until blocked or in need of feedback

espg:
🤖 *from Claude*

Phases 2–5 are now pushed to `claude/48-tier2-vectors`. Summary of what landed:

**Phase 2 — ragged payloads through the Arrow seam** (`src/zagg/processing.py`)

- Added `_has_ragged_fields(config)` helper (mirrors `_has_vector_fields`).
- Added `_coerce_ragged_value(value, sig)` — coerces a per-cell function/expression result to a `(n_elements, *inner_shape)` contiguous float32 array; raises `ValueError` on inner-shape mismatch.
- Updated `_empty_cell_value` to return `[]` for ragged fields (CSR writer represents absent cells via `cell_ids`, not a sentinel slab).
- Updated `calculate_cell_statistics` to dispatch ragged function and expression fields through `_coerce_ragged_value` rather than `float()` or `_coerce_field_value`.
- Updated `process_shard`'s non-kernel loop to collect ragged per-cell arrays into `ragged_payloads`/`ragged_cell_indices` dicts (pre-initialized before the kernel/non-kernel branch).
- Updated `_kernel_aggregate` to skip ragged fields entirely (no kernel-aggregate equivalent); they're excluded from both `kernel_names` and `fallback_names`.
- `_build_output` is called with `dense_vars` (ragged excluded) so the carrier (DataFrame/Arrow table) never sees ragged fields.
- Tests: `TestRaggedPayloads` (11 tests) — `_has_ragged_fields`, `_empty_cell_value`, `_coerce_ragged_value`, function-driven ragged, expression-driven ragged, empty-cell short-circuit, ragged/scalar/vector coexistence.

**Phase 3 — CSR writer + reader** (`src/zagg/csr.py`)

- `write_csr(store, field_name, values_list, cell_ids, *, dtype, zarr_format)` — writes `{field_name}/values` (flat), `{field_name}/offsets` (n_populated+1), `{field_name}/cell_ids` (which cells). Empty per-cell arrays are silently skipped; the standard CSR invariant `offsets[-1] == len(values)` is maintained.
- `read_csr(store, field_name)` — reads all three arrays back.
- `iter_csr_cells(csr)` — decodes the CSR dict into a list of `(cell_id, payload)` pairs.
- Tests: `TestWriteCsr` (9 tests) — round-trip, per-cell slice decode, empty payloads, mixed empty/non-empty, mismatched-lengths error, inconsistent inner-shape error, CSR invariant, 1-D payloads, dtype preservation.

**Phase 4 — t-digest consumer** (`src/zagg/stats/tdigest.py`)

- Pure-numpy implementation based on the Dunning k1 scale function: `build_tdigest(values, delta=512)` → `(k, 2)` float32 centroid array (column 0: mean, column 1: weight). NaN values dropped before sketching.
- `merge_tdigests(d1, d2, delta=512)` — concatenates, sorts by mean, re-compresses with the same scale-limited merge; merged weights sum to the combined total.
- `quantile_from_tdigest(digest, q)` — linear interpolation between centroid boundaries for approximate quantile estimation.
- Wired via `zagg.config.resolve_function("zagg.stats.tdigest.build_tdigest")` — the existing dotted-path resolution handles it without config.py changes.
- Tests: `TestBuildTDigest` (12 tests), `TestMergeTDigests` (7 tests), `TestQuantileFromTDigest` (3 tests) — empty/NaN handling, sorted means, weight conservation, 4δ centroid bound, determinism, 2% median accuracy on 10k uniform sample, merge equivalence.

**Phase 5 — δ sweep + profiling note**

- `TestTDigestDeltaSweep` — parametrized over δ ∈ {128, 256, 512, 1024}: centroid count ≤ 4δ, weights sum to n, quantile error within tolerance (tightening at higher δ), centroid/δ ratio ≤ 4.
- IO-vs-compute profiling note in `tdigest.py` docstring (no new code path).

**Test results**: 467 tests pass (all existing + 60 new). `ruff check` clean. The two formatting issues in `test_tdigest.py` were fixed with `ruff format`; the 20 pre-existing "would reformat" files in the baseline are unchanged.

---

espg (commented):
🤖 *from Claude (review)*

Reviewed phases 2–5 of issue #48 (CSR ragged carrier + t-digest). The overall design is sound and the test coverage is broad. Five findings below, ordered from most to least critical.

---

**Summary of findings**

(1) `_coerce_ragged_value` — redundant first reshape branch is dead code that masks the real logic  
(2) `read_csr` — hardcodes `zarr_format=3`, creating a mismatch when the writer is called with a different format  
(3) `build_tdigest` — `cum_w` tracking: the increment happens *after* the budget decision but the initial value already counts the seed element; works correctly in practice but the invariant is non-obvious and should at minimum be commented  
(4) `_write_array` — empty-array chunk shape is wrong for 2-D arrays  
(5) No test for the `arrow-kernel` handoff path when ragged fields are present

Inline comments on each below.

---

espg (commented):
🤖 *from Claude (review)*

Reviewed phases 2–5 of issue #48 (CSR ragged carrier + t-digest). The overall design is sound and test coverage is broad. Five findings below, ordered from most to least critical. Inline comments mark the exact spots.

---

**Finding (1) — `_coerce_ragged_value`: first reshape branch is dead code**

The two consecutive reshape guards in `_coerce_ragged_value` (processing.py) are:

```python
if arr.ndim == 1 and inner == (1,):
    arr = arr.reshape(-1, 1)
if arr.ndim == 1 and len(inner) == 1:
    arr = arr.reshape(-1, *inner)
```

The first branch (`inner == (1,)`) is a strict subset of the second (`len(inner) == 1`): any input that satisfies `inner == (1,)` also satisfies `len(inner) == 1`, and both reshapes produce the same result (`reshape(-1, 1)`). The first branch is therefore never needed — remove it and leave only the second.

---

**Finding (2) — `read_csr`: hardcoded `zarr_format=3` ignores the writer's format parameter**

`write_csr` accepts `zarr_format: int = 3` and passes it to `_write_array`. `read_csr` always opens with `zarr_format=3`. If a caller passes `zarr_format=2` to the writer (non-default but legal), `read_csr` will fail or silently read from the wrong location. Fix: accept and pass through the same parameter in `read_csr`.

---

**Finding (3) — `build_tdigest` / `merge_tdigests`: `cum_w` invariant is non-obvious**

In `build_tdigest`, `cum_w` is initialized to `1.0` and incremented at the bottom of each loop iteration, *after* the budget decision. The budget check `q = (cum_w - cur_weight / 2.0) / n_total` is therefore evaluated using the cumulative weight *before* accounting for the element being processed. This is a minor inaccuracy (budget at rank `(i-1)/n` vs `(i-0.5)/n`) but is consistent with t-digest's approximate guarantees. The same pattern appears in `merge_tdigests`. A short comment in each loop body explaining the invariant would prevent future confusion and accidental "fixes" that would actually break things.

---

**Finding (4) — `_write_array`: empty 2-D array chunk shape**

```python
chunks=data.shape if data.size > 0 else (1,) * data.ndim,
```

For an empty 2-D array `(0, 2)` the fallback is `(1, 1)`, not `(1, 2)`. The inner dimension is lost from the chunk geometry. Not harmful for the empty-CSR round-trip (the array is never indexed), but inconsistent. A precise fallback: `(1, data.shape[1]) if data.ndim == 2 else (1,)`.

---

**Finding (5) — Missing test: `arrow-kernel` handoff path with ragged fields**

When `process_shard` runs with `handoff="arrow-kernel"` and a config containing ragged fields, `ragged_payloads` / `ragged_cell_indices` are pre-initialized but never filled — the kernel branch has no ragged collection loop. Ragged fields are silently dropped from the output. This is currently untested. Either (a) add a test asserting the drop (documenting the current limitation), or (b) guard at the top of `process_shard`: raise `NotImplementedError` if `handoff == "arrow-kernel"` and `_has_ragged_fields(config)`, so the limitation is explicit rather than a silent data loss.

---

espg:
🤖 *from Claude*

All phases complete. Latest commit (`de12544`) folds the self-review findings: removed dead code in `_coerce_ragged_value`, added `NotImplementedError` guard for ragged + kernel path, added `zarr_format` parameter to `read_csr`, fixed empty 2-D chunk shape in `_write_array`, and added the `cum_w` loop-invariant comment in `build_tdigest`. 135 tests pass, ruff clean.

---

Issue #66, opened 2026-06-19
h5coro H5Coro objects leak ~67 MB/granule in process_shard (no close/eviction) -> Lambda OOM


espg:
🤖 *from Claude*

## Problem

`process_shard` opens one `h5coro.H5Coro` per granule in a loop and never releases it. `H5Coro` has **no `close`, no `__del__`, no cache eviction**, caches reads in **4 MB lines** (`cacheLineSize=4194304`), and every `inspectPath` creates a `ThreadPoolExecutor` that is **never `shutdown()`** -- whose `inspectThread` closures capture `self`, so the object (and its cache) cannot be garbage-collected even after `h5obj` is reassigned on the next iteration.

## Measured

On a real ATL03 granule (full-mission NEON shardmap), just the **coarse geolocation read** done for *every* assigned granule:

- after open: 1 cache line
- after 1 beam coarse read: 5 lines (~21 MB)
- after all 6 beams: **16 lines ~= 67 MB pinned**
- returned arrays are `memoryview`-backed (views into cache lines), so anything retained in `all_reads` pins its 4 MB lines too.

With the full-mission catalog, ~40 granules are assigned per 64-cell shard (see related #65), so **~40 x 67 MB ~= 2.7 GB** -- past the 2 GB Lambda, almost entirely HDF5-navigation **metadata cache**, not photon data. This is why even 640 m (64-cell) chunks OOM despite trivial data volume:

```
Processing 256 of 256 cells (lambda)
  [3/256] shard 13: Lambda error (Unhandled): Runtime.OutOfMemory
  [4/256] shard 44: Lambda error (Unhandled): Runtime.OutOfMemory
  ... (most shards)
```

## Proposed fix (zagg-side)

Release per granule in `process_shard`'s loop:

```python
finally:
    try:
        h5obj.cache.clear()   # frees the 4 MB lines; retained arrays keep their own
    except Exception:         # memoryview refs, so this is safe for already-read data
        pass
    del h5obj
```

Optionally pass a smaller `cacheLineSize` to `H5Coro(...)` (e.g. 1 MB) to cut the per-granule pin ~4x. Takes effect only after a Lambda **function redeploy** (`build_function.sh` + update).

## Upstream

File an h5coro issue for the missing `close()` / context manager and `ThreadPoolExecutor` shutdown (`with ThreadPoolExecutor(...) as ex:` in `inspectPath`). Root cause lives there; the zagg fix is a workaround.

## Relation

[...]

PR #67, opened 2026-06-20
Beam-corridor footprints for ATL03/06 shard assignment (#65)

### LINKED ISSUE #65: Tighter catalog filtering: decompose CMR swath polygon into per-beam tracks
🤖 *from Claude*

## Problem

`ShardMap.build` intersects each granule's **CMR footprint polygon** against shard footprints. For ICESat-2 ATL03 that polygon is a coarse **quarter-orbit swath envelope**, far larger than the actual data, so granules get assigned to shards their beams never touch and most per-shard Lambda reads return null.

Measured on the full-mission NEON Maryland AOI (2018-01-01 -> 2026-06-01, 79 granules):

- 256-cell chunks: **22-48 granules assigned per shard**. Shrinking to 64- or 32-cell chunks barely changes it (median 28 -> 21 -> 20) because the envelopes blanket the whole 10 km AOI regardless of chunk size.
- The CMR polygon at the AOI latitude is **~12.6 km wide** vs the real **~6.5 km** 6-beam span -- ~2x too wide (sparse vertices, ~1 per 100 km along-track, linearly connected, so it bulges outside the curved swath).

This over-assignment is the dominant driver of wasted null reads and amplifies the per-granule h5coro memory leak that OOMs the workers (#66).

## Why a tighter footprint is recoverable

ATLAS beams are **not steerable**; to offset from the reference ground track the **whole platform** is pointed, so the 3 beam pairs keep **fixed cross-track spacing** (~3.3 km between pairs, ~90 m within a pair). And the CMR polygon **tracks the actual collected swath**, offset included -- verified per RGT0568 cycle:

| cycle | CMR poly @ 38.89 deg N | measured beams gt1l..gt3r | inside |
|---|---|---|---|
| c29 | [-76.587, -76.440] | -76.5475 .. -76.4726 | yes |
| c28 | [-76.504, -76.359] | -76.4663 .. -76.3914 | yes |
| c27 | [-76.521, -76.375] | -76.4826 .. -76.4077 | yes |

The polygon **centerline tracks the data-swath center to ~260 m** (c29 poly-center -76.513 vs beam-center -76.510).

## Proposed helper

`beam_tracks_from_cmr_polygon(geom, product="ATL03") -> list[LineString]`:

1. Derive the swath **centerline** (midline of the two long polygon edges) + local **azimuth**.
2. Place the 6 beams (or 3 pairs) at the product's fixed cross-track offsets off the centerline.
3. Buffer each thinly (~few hundred m for centerline slop), return per-beam geometries.

Wire into `ShardMap.build` as the granule footprint for intersection, replacing the raw swath polygon.

## Validation

Inferred c29 beams should reproduce the measured gt1l -76.5475 .. gt3r -76.4726 at 38.89 deg N (data captured this session).

## Benefits

- Cuts assignments from ~40 to the handful that truly cross a shard -> eliminates null reads.
- **Orchestrator-side** (shardmap only) -> no Lambda redeploy needed.
- Directly mitigates the worker OOM by reducing granules opened per shard.

## Open questions

- Buffer width vs centerline accuracy (~260 m measured); per-beam vs per-pair geometries.
- Beam spacing: use ATL03 spec values (pair ~3.3 km, within-pair ~90 m) vs fitting from polygon width.
- Antimeridian / polar-cap swaths (centerline + azimuth degeneracies).
- Coarse polygon vertices (~100 km spacing) -> along-track centerline interpolation accuracy.

_Unlabeled -- needs triage (discuss/plan/implement)._

espg:
Closes #65. Refs #66, Refs #97.

Tightens ICESat-2 ATL03/06 shard assignment by decomposing the coarse CMR
quarter-orbit **swath** polygon (~12.6 km wide) into per-beam-pair **corridors**
(~6.5 km of real data with ~3 km gaps), so granules stop being assigned to
shards their beams never cross. Opt-in (`footprint="beams"`); default `"swath"`
is byte-identical to today.

## What / approach

- **`src/zagg/catalog/beams.py`** (new) — `beam_tracks_from_cmr_polygon(lats, lons, product)`:
  recovers the swath **centerline** (split the ring into its two long edges, resample
  on a shared along-track axis, average; trim the converging end-caps), then offsets
  three corridors at the fixed ATLAS cross-track spacing (gt1/gt2/gt3 at ∓3.3 km / 0)
  via `pyproj.Geod`, buffered ±500 m (with adaptive widening when the envelope is
  irregular — see Phase 4 below). Non-beam products, antimeridian swaths, and
  degenerate rings fall back to the swath ring (a granule is never dropped).
- **Both backends, one seam** — `_granule_footprints(rec, footprint, product)` returns
  a list of `(lats, lons)` rings (1 for swath, 3 for beams). `_intersect_spherely` and
  `_intersect_mortie` both loop the rings and **dedup the granule index per shard**, so
  a granule is assigned if *any* of its beams intersect. The corridor ring is consumed
  identically (spherely polygon / mortie `morton_coverage`) — no backend-specific geometry.
- **Opt-in surface** — `footprint` threaded through `ShardMap.build` → `make_shardmap`
  → CLI `--footprint {swath,beams}` (default swath); recorded in shardmap metadata.

## Phases

- [x] Helper `beams.py` + pure-geometry tests (validated against real data)
- [x] `_granule_footprints` seam + both-backend integration (loop + per-granule dedup)
- [x] Opt-in surface (`ShardMap.build` / `make_shardmap` / CLI) + behavior tests
- [x] Phase 4 — fold self-review (a793cf8): antimeridian seam-jump detection, hard
  error on `beams` + non-beam catalog, adaptive corridor width, public
  `is_beam_product` helper, doc-nit on per-pair labels
- [x] Phase 5 — reconcile with `main` (merge of the post-rework `main`: #82 `processing/`
  split, #92 mortie MOC-order, #89 `spatial_signature`) + add docstring/comment
  deprecation markers per the direction below.

## Deprecation

@espg flagged the beam-corridor mechanism as a stopgap hack to fence for easy
removal. Per the direction on the PR thread, markers are **docstring/comment only
— no runtime `DeprecationWarning`** (it would be the noisy part). Added:

- A module-level `.. deprecated::` note + a `# DEPRECATED` banner at the top of
  `src/zagg/catalog/beams.py`.
- Docstring deprecation notes on `beam_tracks_from_cmr_polygon`, `is_beam_product`,
  the `_granule_footprints` seam, and the `footprint="beams"` option where it's
  documented (`ShardMap.build`, `make_shardmap`, and the CLI `--footprint` help).
- **Removal trigger** encoded in every marker: remove once a better fix lands —
  **native per-beam CMR geometry**, **OR** the memory-handling robustness in #66,
  **OR** data virtualization tracked in #97.

No behavior change from the markers (documentation only on the code surface).

## Phase 4 — what changed in the fold

Addresses the four diff-scoped review findings on the initial PR; see "Standing
limitations" below for the two findings that are documented but not coded around.

- **Antimeridian detection** keys on a real consecutive-vertex seam jump
  (`np.max(np.abs(np.diff(lons))) >= 180`) instead of the raw span
  `np.ptp(lons)`. A near-polar quarter-orbit can sweep > 180° of longitude
  without any antimeridian crossing — the old guard silently no-op'd to swath
  exactly where corridor tightening is most valuable. New test
  `test_wide_lon_polar_does_not_fall_back` locks this in.
- **`footprint="beams"` on a non-beam catalog now raises a `ValueError`** (in
  `ShardMap.build`) rather than recording `footprint="beams"` in the metadata
  while every granule silently degraded to the full swath. Covers the
  `collection: None` / missing-`collection` / non-ATL03/06 cases. Tests:
  `test_beams_on_non_beam_catalog_raises`,
  `test_beams_on_missing_collection_metadata_raises`.
- **Adaptive corridor half-width.** `_centerline` now also returns the larger
  of the two median geodesic edge-distances. When the recovered envelope is
  wider than ATL03/06's expected ~6.3 km half-width (typically because CMR has
  over-padded one or both sides), the corridor widens by the excess so the
  beams stay covered when the envelope-centerline drifts. Test:
  `test_wider_envelope_widens_corridor`.
- **Doc nit:** the S→N reorder comment no longer claims a fixed "gt1 west /
  gt3 east" pair labelling — that's heading-dependent and the backends
  union all per-granule rings anyway. Test:
  `test_input_order_reversed_yields_same_corridor_union` asserts the corridor
  *union* covers all three true beam positions.
- New public helper `is_beam_product(product)` in `beams.py` (used by
  `shardmap.py` and exposed for callers; basic tests in `TestIsBeamProduct`).

## How tested

- Full suite green after the `main` merge: `pytest -q` → **804 passed, 1 skipped**
  (the skip is the real spherely-fork `SpatialIndex` build, absent in the default
  venv). `pytest -q tests/test_shardmap.py` → 51 passed, 1 skipped.
- `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check`
  clean on the touched files (the merge took `main`'s canonical, format-clean code).
- Key tests (`tests/test_shardmap.py`):
  - `test_c29_corridors_contain_measured_beams` — the **#65 validation target**: the real
    RGT0568 c29 CMR polygon decomposes to 3 corridors that each contain the granule's
    measured beam ground-track at lat 38.89 (gt1l −76.5475 / gt2l −76.5106 / gt3l −76.4737).
  - `test_beam_mode_fewer_shards_{spherely,mortie,healpix}` — on a gap-straddling layout, beam mode
    assigns strictly fewer shards than swath, across all backends.
  - synthetic offset / non-beam passthrough / degenerate + antimeridian fallback / default
    swath unchanged / CLI flag — plus the wide-lon-polar / wider-envelope /
    input-order / non-beam-catalog tests from the fold.

## Notes / questions for review

- **Reconciled with `main`** — this branch was based on a pre-rework `main`; the
  merge takes `main`'s canonical code wholesale for every file the beam feature
  doesn't touch (so #82 `processing/` split, #92 mortie MOC-order, #89
  `spatial_signature` are preserved verbatim) and layers the beam additions onto
  `main`'s `shardmap.py` / `catalog/__init__.py` / `tests/test_shardmap.py`. The
  beam intersect calls now sit *after* `main`'s `_resolve_mortie_order(...)` resolution.
- **Operational fix, no Lambda redeploy** — beam mode only changes the shardmap (orchestrator
  side). Rebuild a shardmap with `--footprint beams` to use it.
- **Constants** (`_ATL03_PAIR_OFFSETS_M = (-3300, 0, 3300)`, `half_width = 500 m`,
  `_EXPECTED_SWATH_HALF_WIDTH_M = 6300 m`) live in `beams.py`; centerline
  recovery is accurate to ~220–340 m on the c29 sample, comfortably inside the
  500 m corridor. Tunable if you want tighter/looser corridors.
- Companion memory-leak fix is #66 (independent; needs a function redeploy);
  data-virtualization tracking is #97.

## Standing limitations (carried over from self-review, not blocking)

- **Symmetric-asymmetric envelope drift.** The reviewer's "4 km west / 8 km
  east of true center" scenario keeps the envelope's *total* width at the
  expected ~12 km (only the placement shifts), so the adaptive widening above
  doesn't catch it: the centerline-of-envelope cannot distinguish symmetric
  vs. asymmetric padding from the envelope shape alone. The adaptive width
  defends the *wider-than-expected* case; the placement-shift case requires
  external information we don't have here.
- **Severely-pinched envelopes** (a mid-track interior pinch dropping width to
  near zero) can split the width-mask into two arcs. The current behavior
  bridges them via `geod.inv` — fine for normal CMR convergence at the end
  caps; a real CMR polygon with a sharp interior pinch would yield an angled
  chord across the pinch instead of a faithful centerline. Documented as a
  known limitation in `_centerline`'s docstring; not observed on real ATL03/06
  CMR data and left for follow-up if a real granule exhibits it.

_Authored by Claude for @espg._


src/zagg/catalog/beams.py
```diff
@@ -0,0 +1,210 @@
...
+# ICESat-2 ATLAS beam geometry. The six beams form three pairs; the pairs are
+# ~3.3 km apart cross-track (the strong/weak beams within a pair are only ~90 m
+# apart, far below the corridor half-width, so one corridor per pair covers both).
+# gt2 sits ~on the reference ground track (centerline); gt1/gt3 straddle it.
+_ATL03_PAIR_OFFSETS_M: tuple[float, ...] = (-3300.0, 0.0, 3300.0)  # gt1, gt2, gt3
+
+# Corridor half-width: envelopes the ~90 m within-pair separation plus the
+# ~260 m centerline-recovery error, with margin -- while staying well inside the
+# ~1.5 km half-gap between pairs so the inter-pair gaps remain unassigned.
+_BEAM_CORRIDOR_HALF_WIDTH_M: float = 500.0
```
espg, line 44: 🤖 *from Claude (review)*

**Silent data loss when the CMR envelope is asymmetric about the true data centerline (should-fix / blocker).**

The corridor half-width is 500 m and the docstring justifies it against a "~260 m centerline-recovery error" measured on the single c29 validation granule. But that error model assumes the CMR swath envelope is *symmetric* about the true data center. It need not be — the envelope is a coarse bounding swath, and if one edge is padded more than the other the recovered centerline (a simple average of the two edges in `_centerline`) is biased by *half the asymmetry*. I verified this against the real code: a swath whose envelope is 4 km west / 8 km east of the true center recovers a centerline ~2 km off, putting all three true beams ~1.5 km **outside** their 500 m corridors:

```
min centerline recovery err (m) at mid: 2014.0
```

When that happens the granule is silently dropped from the shards its beams actually cross — and there is **no fallback**, because the fallback path (`_swath_fallback`) only fires on a *failed split* (`_centerline` returns `None`), never on an *inaccurate* one. A poor-but-valid recovery is indistinguishable from a good one here.

Concrete suggestions: (a) sanity-check the recovered centerline against the envelope before trusting it — e.g. if `median(width)/2` (the implied half-swath) deviates from the expected ~6 km half-swath by more than the corridor half-width, fall back to swath; and/or (b) widen the half-width to cover the worst observed asymmetry, or make it a function of the measured swath width. At minimum, add a test with an asymmetric envelope asserting the beams are still contained (the only accuracy test today, `test_synthetic_straight_swath_offsets`, uses a perfectly symmetric swath).

src/zagg/catalog/beams.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    lats = np.asarray(lats, dtype=float)
+    lons = np.asarray(lons, dtype=float)
+    if product not in _BEAM_PRODUCTS or len(lats) < _MIN_VERTS:
+        return _swath_fallback(lats, lons)
+
+    # Antimeridian-crossing swaths (>= 180 deg lon span) can't be represented as
+    # a simple [-180, 180] corridor ring; decomposing them is rare (polar/seam
+    # granules) so fall back to the swath rather than mis-handle the wrap.
+    if np.ptp(lons) >= 180.0:
```
espg: 🤖 *from Claude (review)*

**`np.ptp(lons) >= 180` falsely falls back on wide-longitude near-polar swaths that do *not* cross the antimeridian (should-fix).**

The guard is meant to catch antimeridian-wrapped rings, but `ptp` is a planar longitude span and a near-polar quarter-orbit can legitimately sweep >180° of longitude *without* any antimeridian seam. I verified a high-latitude polygon (no wrap) with `ptp(lons)=199` falls straight to the swath:

```
ptp lons 199.0
rings: 1 (1 = fallback)
```

These are exactly the polar granules where over-assignment is worst (the swath envelope covers a huge lon span near the pole), so silently skipping the tightening there defeats much of the purpose. The granule is still *kept* (correctness is safe), but the feature silently no-ops on the cases that need it most, with no signal to the caller.

Consider detecting a genuine seam instead of a raw span — e.g. flag only when consecutive-vertex lon deltas exceed ~180 (an actual wrap), or unwrap the longitudes (`np.unwrap` in radians) before the span test. A test with a non-crossing wide-lon polar swath asserting it is *not* dropped would lock this down (`test_antimeridian_falls_back` only covers the true-crossing case).

src/zagg/catalog/beams.py
```diff
@@ -0,0 +1,210 @@
...
+        return None
+    ts = np.linspace(t0, t1, n_samples)
+    lon_a, lon_b = np.interp(ts, ta, na), np.interp(ts, tb, nb)
+    lat_a, lat_b = np.interp(ts, ta, la), np.interp(ts, tb, lb)
+    clat = 0.5 * (lat_a + lat_b)
+    clon = 0.5 * (lon_a + lon_b)
+    # Near the end caps the two edges converge to a shared corner, which would
+    # drag the centerline sideways; keep only where the swath has real width.
+    width = np.abs((lon_a - lon_b) * np.cos(np.radians(clat)))
+    keep = width >= 0.3 * np.median(width)
```
espg, line 161: 🤖 *from Claude (review)*

**The cap-trim `keep` mask is not required to be contiguous — a pinched/tapering swath yields a centerline with a hole, and the corridor ring then chords across it (nit / should-fix).**

`keep = width >= 0.3 * np.median(width)` is applied as a boolean mask and the kept points are taken directly (`clat, clon = clat[keep], clon[keep]`) with no contiguity check. If the swath narrows somewhere in the *interior* (e.g. a genuinely tapering or pinched envelope), the dropped samples are in the middle, not just at the end caps the comment describes. I confirmed this against the real code with a mid-pinched swath: the recovered centerline has an interior discontinuity:

```
lat gaps: [...0.005 0.005 0.136 0.005 0.005...]   # the 0.136 is a dropped middle span
```

Downstream this matters twice: (1) `geod.inv(clon[:-1]..., clon[1:]...)` computes the azimuth *across* that gap, so the offset direction near the seam is wrong; and (2) the corridor ring is built from these samples, so it gets a long straight chord jumping the gap, distorting the corridor exactly where the swath is narrowest. The end-cap rationale in the comment only holds if the low-width region is guaranteed to be the two ends — which the mask does not enforce.

Suggestion: keep only the longest *contiguous* run above the threshold (or trim a fixed margin from each end instead of thresholding), so the centerline stays a single unbroken arc. A pinched-swath test would catch regressions.

src/zagg/catalog/shardmap.py
```diff
@@ -272,12 +305,22 @@ def build(
...
 
         Returns
         -------
         ShardMap
         """
+        if footprint not in ("swath", "beams"):
+            raise ValueError(f"footprint must be 'swath' or 'beams' (got {footprint!r})")
         records = catalog.granule_records()
+        # Product short-name drives beam decomposition (collection like "ATL03_007").
+        product = ((catalog.metadata or {}).get("collection") or "").split("_")[0].upper()
```
espg, line 411: 🤖 *from Claude (review)*

**`footprint="beams"` silently no-ops to swath behavior when the collection is missing/None or not ATL03/06 — no signal to the user (should-fix).**

```python
product = ((catalog.metadata or {}).get("collection") or "").split("_")[0].upper()
```

I traced the edge cases against the real expression:

```
None         ->            beam? False
           ->            beam? False
SomeThing  -> SOMETHING  beam? False
atl03_007  -> ATL03      beam? True   (good, case-insensitive)
```

So if a user explicitly passes `footprint="beams"` but the catalog metadata has no `collection`, a `None` collection, or a non-ATL03/06 product, `product` becomes `""`/some other string, `beam_tracks_from_cmr_polygon` hits its `product not in _BEAM_PRODUCTS` guard, and every granule silently falls back to the full swath. The build "succeeds," `metadata["footprint"]` is still recorded as `"beams"`, and there is zero indication the tightening did nothing. That is a confusing silent failure for a deliberately opted-in feature.

Suggestion: when `footprint == "beams"` and the derived `product` is empty or not in `_BEAM_PRODUCTS`, either raise a clear `ValueError` (collection X is not a beam product) or emit a warning, so the no-op is visible. There is no test covering `footprint="beams"` with a non-beam/missing collection today.

src/zagg/catalog/beams.py
```diff
@@ -0,0 +1,210 @@
...
+    if keep.sum() < 2:
+        return None
+    clat, clon = clat[keep], clon[keep]
+    # Order south -> north so the signed cross-track offsets place gt1 west /
+    # gt3 east consistently for both ascending and descending granules.
```
espg: 🤖 *from Claude (review)*

**The S→N reorder makes the corridor *set* asc/desc-symmetric (good — no data-loss bug), but the "gt1 west / gt3 east" pair labels it claims are wrong for descending tracks (nit / doc fix).**

I checked this concern carefully against the real code with a tilted descending track (heading az≈170°). The S→N reorder does keep the union of the three corridors correct regardless of asc/desc — the three corridor centers land on the three true beam positions as a *set*:

```
descending az=170, true beam lons:  -76.2467 (gt1)  -76.2832 (gt2)  -76.3197 (gt3)
corridor centers:        pair0 -76.3208   pair1 -76.2832   pair2 -76.2456
```

Since both backends *union* all rings per granule for shard assignment, the label swap is harmless for correctness — so this is not the data-loss bug one might fear, and I am flagging it only as a doc issue. But note pair0 (offset −3300, the comment says "west") actually lands on the *east* true beam (gt1, −76.2467) for the descending case — i.e. for a descending track the offsets are mirrored relative to the geographic west/east the comment asserts. If anything ever consumes the *per-pair labeling* (e.g. naming corridors gt1/gt2/gt3), it would be wrong half the time. Suggest softening the comment to "consistently straddle the centerline" rather than claiming a fixed west/east mapping, and add a descending-track test so this stays understood.

espg:
🤖 *from Claude*

Phase 4 fold (a793cf8) addresses the four diff-scoped review findings from the initial open + folds the doc nit.

**Coded in:**
- Antimeridian guard switched from `np.ptp(lons) >= 180` to a real consecutive-vertex seam-jump test, so wide-lon polar quarter-orbits stop silently no-op'ing to swath (`test_wide_lon_polar_does_not_fall_back`).
- `footprint="beams"` now raises a `ValueError` when the catalog collection isn't ATL03/06 (or is missing entirely), so the opt-in flag can't silently degrade to swath while metadata records `footprint="beams"` (`test_beams_on_non_beam_catalog_raises`, `test_beams_on_missing_collection_metadata_raises`).
- Adaptive corridor half-width: `_centerline` now also returns the larger of the median geodesic edge-distances; when the envelope is wider than the expected ~6.3 km half-width, the corridor widens by exactly the excess so beams stay covered when the envelope is over-padded (`test_wider_envelope_widens_corridor`).
- Doc nit on the S→N reorder dropped the "gt1 west / gt3 east" pair-label claim — that's heading-dependent and the backends union all per-granule rings anyway; the descending-track case is now locked in by `test_descending_track_corridor_union`.

New public helper `is_beam_product(product)` in `beams.py` is used by `shardmap.py` and exposed for callers.

**Standing as documented limitations** (`_centerline` docstring + PR body):
- A *placement-shifted* envelope of normal width (the reviewer's "4 km west / 8 km east of true center" verification) can't be distinguished from a symmetric one purely from envelope shape, so the adaptive widening doesn't catch it. The defence is on the *wider-than-expected* axis.
- Severely interior-pinched envelopes still bridge the pinch via `geod.inv` rather than carving the centerline at the pinch — fine for end-cap convergence, but a sharp mid-track pinch would chord across. Not observed on real CMR data; left for follow-up if a real granule exhibits it.

CI all-green on the prior commit; tests pass locally on this one (`tests/test_shardmap.py` → 36 passed, 1 skipped; full `pytest` clean except for two pre-existing morton-int64 failures in `tests/test_integration.py` and `tests/test_processing.py` that also fail on HEAD — flagging not fixing).

Continuing — pre-fold self-review next on this commit.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of fold commit a793cf8 against the prior-review findings. The four coded-around findings (1, 2, 4, 5) are addressed appropriately; finding 3 (non-contiguous `keep` mask) was deliberately deferred and documented as a known limitation in `_centerline`'s docstring, which is an honest call I won't relitigate.

Findings below are all narrow nits on the fold itself:
- A real case-sensitivity inconsistency: the new `is_beam_product` helper is case-insensitive but the in-helper `_BEAM_PRODUCTS` guard inside `beam_tracks_from_cmr_polygon` is not — same string ("atl03") gives different outcomes through the two surfaces.
- Two test-coverage gaps: the new adaptive-widening test only confirms that a wider-than-expected envelope widens the corridor, not that a normal-width envelope leaves the base 500 m corridor unchanged (the over-widening direction is untested); and the new descending-track test does not actually build a descending heading (the S→N flip is what's exercised, not the `170°` heading the comment describes).
- A minor antimeridian-detection edge: `max(|diff(lons)|) >= 180` still classifies as a seam a true polar-cap polygon that legitimately steps ≥180° between consecutive vertices over a pole without ever crossing ±180°. Real-world rare, but the guard's docstring claim ("an actual consecutive-vertex seam jump") slightly overstates what it detects.

No undiscussed dependencies, no unused imports, module size well under the ~1000-line rule, commit-message style matches §3. PR body and "Standing limitations" section are honest about what's documented vs coded.

---

src/zagg/catalog/beams.py
```diff
@@ -0,0 +1,254 @@
...
+
+    # Forward azimuth along the centerline; repeat the last for the final point.
+    az12, _, _ = geod.inv(clon[:-1], clat[:-1], clon[1:], clat[1:])
+    az = np.concatenate([az12, az12[-1:]])
+
+    # Adaptive half-width. The CMR envelope of a quarter-orbit ATL03/06 swath
+    # is ~12.6 km wide (half-width ~6.3 km); a wider envelope means asymmetric
+    # or one-sided CMR padding can drift the recovered centerline from the true
+    # data center by up to half the excess width. Widen the corridor by exactly
+    # that excess so the beams remain covered even when the envelope is
```
espg: 🤖 *from Claude (review)*

**Case-sensitivity inconsistency between the new public helper and this in-helper guard.**

The new `is_beam_product` (line 52) does `product.upper() in _BEAM_PRODUCTS` — case-insensitive, as exercised by `TestIsBeamProduct.test_known_beam_products` (`assert is_beam_product("atl03")`). But this guard at line 235 is still the original `product not in _BEAM_PRODUCTS`, which only matches the exact uppercase form. So:

- `is_beam_product("atl03")` → `True`
- `beam_tracks_from_cmr_polygon(lats, lons, product="atl03")` → silent swath fallback

It's currently masked because the one in-tree caller (`shardmap.py`) already does `.upper()` before passing `product` in, but any other consumer hitting this helper directly with the lowercase form silently no-ops where `is_beam_product` says it should decompose. Either upper-case `product` at the top of `beam_tracks_from_cmr_polygon`, or use `is_beam_product(product)` here for the two surfaces to agree. (Not a blocker — flagging as a contract nit on the new public API.)

---

src/zagg/catalog/beams.py
```diff
@@ -0,0 +1,254 @@
...
+    if product not in _BEAM_PRODUCTS or len(lats) < _MIN_VERTS:
+        return _swath_fallback(lats, lons)
+
+    # Antimeridian-crossing swaths can't be represented as a simple [-180, 180]
+    # corridor ring; decomposing them is rare (polar/seam granules) so fall back
+    # to the swath rather than mis-handle the wrap. The detection keys on an
+    # actual consecutive-vertex seam jump (>~180 deg between neighbours), not on
+    # the raw lon span -- a near-polar quarter-orbit can sweep >180 deg of lon
+    # without any antimeridian crossing, and would otherwise no-op silently.
+    if len(lons) >= 2 and float(np.max(np.abs(np.diff(lons)))) >= 180.0:
```
espg, line 236: 🤖 *from Claude (review)*

**Antimeridian-seam guard handles the common case but the docstring slightly overstates its coverage.**

The check is `max(|diff(lons)|) >= 180.0`. This correctly distinguishes a real antimeridian wrap (e.g. 179.9 → -179.9 = jump of 359.8) from the wide-lon polar case the prior review flagged. The new `test_wide_lon_polar_does_not_fall_back` locks that in — good.

Edge that the docstring claim ("an actual consecutive-vertex seam jump") elides: a true polar-cap polygon (rare but real for some CMR products) that sweeps over the pole can have a consecutive-vertex Δlon of ≈180° (two vertices on opposite sides of the pole at, say, lon=90 and lon=−90) with no antimeridian crossing — those would currently trip the guard and fall back. Not worth coding around for a path that's already a "rare cases" fallback, but the docstring could be slightly more honest: this catches *seams or near-pole leaps*, not seams alone. Optional doc nit; no behaviour change needed.

---

tests/test_shardmap.py
```diff
@@ -227,6 +227,292 @@ def test_cli_rejects_shapely_backend(self, monkeypatch):
...
+        outer_lon_neg = Point(center_lon - outer_offset_m * deg_per_m + eps, center_lat)
+        corridors = [Polygon(zip(rlon, rlat)) for rlat, rlon in rings]
+        assert any(c.contains(outer_lon_pos) for c in corridors), (
+            "adaptive widening must extend corridor outward when envelope is over-padded"
+        )
+        assert any(c.contains(outer_lon_neg) for c in corridors)
+
+    def test_descending_track_corridor_union(self):
+        # Same beam-set coverage holds for descending granules (heading south).
+        # The S->N reorder makes per-pair west/east labels arbitrary, but the
```
espg: 🤖 *from Claude (review)*

**Adaptive-widening test is one-sided — only asserts widening kicks in on a wide envelope, not that it stays inert on a normal one.**

The asserted invariant here is `corridor extends to ~7 km when half-width ~10 km`. What's missing is the converse: on a normal ~6.3 km envelope, the corridor half-width must stay at the base 500 m so the inter-pair ~3 km gaps remain unassigned (the original tightening goal of #65). `test_synthetic_straight_swath_offsets` checks corridor *centers* but not corridor *extent* — without an extent-on-normal-envelope assertion, a future bug that always widens (e.g. dropping the `max(0.0, ...)` floor, or comparing the wrong baseline) would slip through this test alone.

A one-liner addition: after the existing `test_synthetic_straight_swath_offsets`, assert the corridor of pair gt2 (offset 0) has a half-width near 500 m, not several km. Cheap, locks down the "don't over-widen" direction.

---

tests/test_shardmap.py
```diff
@@ -227,6 +227,292 @@ def test_cli_rejects_shapely_backend(self, monkeypatch):
...
+        east_lon = center_lon + half_w
+        lats = np.concatenate([center_lat, center_lat[::-1], [center_lat[0]]])
+        lons = np.concatenate([west_lon, east_lon[::-1], [west_lon[0]]])
+        rings = beam_tracks_from_cmr_polygon(lats, lons, product="ATL03")
+        assert len(rings) == 3
+        beams_true = [Point(-76.50 + d * 3300 * deg_per_m, 38.89) for d in (-1, 0, 1)]
+        corridors = [Polygon(zip(rlon, rlat)) for rlat, rlon in rings]
+        for beam in beams_true:
+            assert any(c.contains(beam) for c in corridors)
+
```
espg, line 675: 🤖 *from Claude (review)*

**This is named/commented as a descending-track test but is really an N→S-input + S→N-reorder test.**

The comment claims `heading ~170 deg (mostly south, slight eastward drift)`. What's built is `center_lat = np.linspace(39.04, 38.74, 12)` — an N-to-S vertex order, but a pure-meridional polygon with a tiny lon drift. After `_centerline`'s `if clat[0] > clat[-1]: ... [::-1]` flip the centerline is S→N and `geod.inv` produces an azimuth of ~360°/~10° (very nearly *north*, since the drift is east), not 170°. So this asserts the corridor *union* is invariant under input-order reversal — which is useful, but it's not exercising the descending-heading semantics the previous reviewer asked about (where the per-pair west/east label flips).

A true descending case would build the envelope so the centerline azimuth is ~170° (heading SE, no S→N flip needed because input already runs S→N, or where the reorder flips it the other way). The current test still passes the meaningful "union covers all three beams" invariant, but the comment overpromises what's tested. Either tighten the comment, or build a heading-south envelope (e.g. `center_lat` ascending with an eastward drift, so the implied azimuth is ~170°).

---

src/zagg/catalog/shardmap.py
```diff
@@ -272,12 +305,33 @@ def build(
...
```
espg, line 411: 🤖 *from Claude (review)*

**The `ValueError` text is structurally informative but loses information for the empty-collection case.**

For a catalog missing `collection` entirely (the `test_beams_on_missing_collection_metadata_raises` path), the message reads:

```
catalog collection None resolves to product ''
```

That tells a user *what* failed but not *why* it failed (no `collection` key at all). For a `collection: "ATL08_006"` it reads:

```
catalog collection 'ATL08_006' resolves to product 'ATL08'
```

— which is the right shape. Consider distinguishing the two: "no `collection` metadata" vs "collection X is not ATL03/06". Small message polish, no behaviour change. The hard-fail itself is the right call (loud over silent for an opt-in flag).

---

src/zagg/catalog/beams.py
```diff
@@ -0,0 +1,254 @@
...
+    if keep.sum() < 2:
+        return None
+    clat, clon = clat[keep], clon[keep]
+    lon_a, lat_a, lon_b, lat_b = lon_a[keep], lat_a[keep], lon_b[keep], lat_b[keep]
+    # Median geodesic distance from the centerline to each edge. Take the max
+    # of the two so an asymmetric envelope (one edge padded farther than the
+    # other) reports its true outer reach -- the side the centerline drifted
+    # away from is the side we need to widen the corridor toward.
+    _, _, d_a = geod.inv(clon, clat, lon_a, lat_a)
+    _, _, d_b = geod.inv(clon, clat, lon_b, lat_b)
```
espg, line 171: 🤖 *from Claude (review)*

**`max(median(d_a), median(d_b))` is the right move for the wider-envelope case, but worth a one-line note that it doesn't fix the placement-shift case the previous review raised.**

The previous review's "4 km west / 8 km east of true center" scenario is the *symmetric*-asymmetric one: the envelope's total width stays at the expected ~12 km, only the placement shifts relative to the true data axis. In that case `d_a ≈ d_b ≈ 6.3 km`, `max(...)` ≈ expected, `extra = 0`, and the corridor is *not* widened — the beams sit outside their corridors and the granule is silently dropped on shards it actually crosses. The PR description acknowledges this in the "Standing limitations" section, which is honest, but the in-code comment on line 162 (`a wider envelope means asymmetric or one-sided CMR padding can drift the recovered centerline ... by up to half the excess width`) slightly conflates the two and implies the adaptive widening handles asymmetry generally. Suggest tightening that comment to "wider-than-expected envelope" and explicitly noting placement-shifted (same-width-but-off-center) envelopes are not detected from polygon shape alone, so a reader of just `beams.py` doesn't have to read the PR body to learn the limitation.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of fold commit a793cf8 against the prior-review findings. The four coded-around findings (1, 2, 4, 5) are addressed appropriately; finding 3 (non-contiguous `keep` mask producing a centerline with an interior hole) was deliberately deferred and documented as a known limitation in `_centerline`'s docstring, which is an honest call I won't relitigate.

Inline findings below are narrow nits on the fold itself, none blocking:

- **Case-sensitivity inconsistency between the new `is_beam_product` helper and the existing in-helper guard inside `beam_tracks_from_cmr_polygon`** — same string ("atl03") routes differently through the two surfaces. Currently masked because the in-tree caller uppercases first, but the public-API contract should agree.
- **Adaptive-widening test coverage is one-sided** — asserts widening on a wide envelope, doesn't assert non-widening on a normal envelope (the over-widening direction is untested).
- **`test_descending_track_corridor_union` does not actually build a descending heading** — it builds N→S-ordered vertices, which the S→N reorder flips; the resulting centerline is ~due-north. It still asserts the "union covers three beams" invariant (which is what matters), but the comment overpromises that it exercises descending-heading semantics.
- **Antimeridian-seam docstring slightly overstates coverage** — `max(|diff(lons)|) >= 180` also trips on true polar-cap polygons stepping ≥180° over a pole without a wrap. Real-world rare; doc nit only.
- **`ValueError` message merges two failure modes** — missing-`collection` vs non-beam product both render through the same template; could be polished, no behaviour change needed.
- **In-code comment on adaptive widening conflates "wider envelope" with "asymmetric envelope"** — the PR body's "Standing limitations" correctly explains the placement-shifted case isn't caught, but the inline comment in `beam_tracks_from_cmr_polygon` reads as if asymmetry is generally handled. Worth a one-line clarification so the limitation is visible from the source.

No undiscussed dependencies, no unused imports, `beams.py` is ~250 lines (well under the ~1000-line module rule), the commit-message title matches §3's terse style, and the PR body is honest about what's documented vs coded. The four fixes folded in do what they claim to do.

---

espg:
🤖 *from Claude*

Phase 5 fold (6cd29d6) addresses the six nits the fresh-context self-review left on phase 4.

**Coded in:**
- `beam_tracks_from_cmr_polygon` now routes through `is_beam_product(product)` instead of the exact `product not in _BEAM_PRODUCTS` check — the public helper and the in-helper guard now agree case-insensitively, so `is_beam_product("atl03")` → True and `beam_tracks_from_cmr_polygon(..., product="atl03")` decomposes (was silently a swath fallback).
- New `test_normal_envelope_keeps_base_corridor_width` locks down the "don't over-widen" direction of the adaptive widening — corridor half-extent for gt2 on a normal envelope must stay under ~1 km, so a future bug that always widens slips through.
- Renamed `test_descending_track_corridor_union` → `test_input_order_reversed_yields_same_corridor_union` with a comment that's honest about what it actually tests (input-order invariance under the S→N reorder, not a true ~170° descending heading). Pure-meridional + S→N reorder gives a north-heading centerline; a true descending-heading test would need a non-meridional polygon.
- Antimeridian-fallback comment now says "≥180 deg consecutive-vertex longitude jump" rather than "antimeridian seam jump" — the check also catches the rarer true polar-cap polygon whose vertices straddle the pole at opposite lons (a near-180° leap, no seam). Honest about scope.
- `ValueError` text now distinguishes the missing-`collection`-metadata case from the non-beam-product case ("catalog has no 'collection' metadata so the product can't be identified" vs "catalog collection 'ATL08_006' resolves to product 'ATL08'").
- Adaptive-widening in-code comment tightened to "wider-than-expected envelope" and explicitly notes that placement-shifted envelopes of normal width are NOT detected from polygon shape alone — so a reader of just `beams.py` learns the limitation without having to read the PR body.

`pytest tests/test_shardmap.py -v` → 37 passed, 1 skipped. `ruff check src tests` clean. All phases of #67 are now done from my end — moving this PR to `waiting` (label kept as `implement`; awaiting your review/merge).

---

espg:
this is a horrible hack and it drives me crazy that we need it-- would be much better to have a way to read the geometries per beam pair directly in CMR!

That said, we *do* seem to need it, at least for now. We should mark everything in this with a depreciation label through, so it's easy to remove once we have a better fix (i.e., data virtualization, or robustness in memory handling from #66 improvements).

My big question on this: Are we doing this for healpix catalogs too? Or more specifically, *how* are we doing this for healpix catalogs?

espg:
🤖 *from Claude*

Two things in your [comment](https://github.com/englacial/zagg/pull/67#issuecomment-4783559528): the HEALPix question, then the deprecation marking.

## "How are we doing this for HEALPix catalogs?"

The beam-corridor decomposition is **grid- and backend-agnostic** — it happens on the CMR polygon *before* any backend dispatch, so HEALPix gets it the same way exact-S2 does:

- `_granule_footprints(rec, footprint, product)` (`shardmap.py:60`) turns one granule into **1 ring** (`swath`) or **3 corridor rings** (`beams`). Nothing grid-specific here.
- The HEALPix path is `_intersect_mortie` (`shardmap.py:178`). It loops those same `(lats, lons)` rings — `for rlats, rlons in _granule_footprints(...)` (`:193`, `:214`) — and feeds **each corridor ring** to mortie `morton_coverage_moc` / `morton_coverage` (`:195`, `:216`), exactly the MOC path the swath ring already used. It then **dedups the granule index per shard** (`dict.fromkeys`), so a granule lands on a shard if *any* of its 3 beam corridors covers it.

So there's **no HEALPix-specific corridor code** — the corridor polygons (geodesic centerline + ∓3.3 km offsets, buffered ±500 m) are consumed by `morton_coverage` identically to the swath ring.

One caveat worth naming: *which* backend a HEALPix catalog uses is decided by `_resolve_backend` — exact-S2 `spherely` when it imports (`_intersect_spherely`), else mortie (`_intersect_mortie`). **Both** loop the corridor rings, so beams work either way; the only difference is S2-polygon intersection vs MOC `morton_coverage`. So: yes, for HEALPix too — via `_intersect_mortie` (or spherely if it's installed).

## Deprecation marking

Agreed it's a hack to fence for easy removal. Concretely, I'd:

1. **Runtime signal** — emit a `DeprecationWarning` from `ShardMap.build` when `footprint="beams"` is selected (once per build, not per granule, to avoid noise), naming the removal trigger.
2. **Source markers** — a `.. deprecated::` docstring note + a `# DEPRECATED — remove when <trigger>` banner on the `beams.py` module and the `footprint="beams"` / `beam_tracks_from_cmr_polygon` / `is_beam_product` surfaces, so a reader of just `beams.py` sees it.
3. **Removal trigger encoded explicitly** — the better fix, so it's unambiguous when to delete: native per-beam CMR geometry, **or** #66's memory-handling robustness / data virtualization making the full-swath read affordable again.
4. Optionally open a tracking issue ("remove beam-corridor footprints once #66 / data-virtualization lands") so it doesn't get lost.

Two questions before I touch code:
- **Which trigger** do you want encoded in the markers — just #66, or #66 *plus* data-virtualization (and should I open the tracking issue)?
- A **runtime `DeprecationWarning`** on every `beams` shardmap build — wanted, or are the docstring/comment markers enough (the warning is the part most likely to be noisy)?

Since this PR is `discuss`, I'm holding off code. Give the word (and flip it back to `implement`) and I'll land the deprecation fencing as a phase commit on `claude/65-beam-footprints`. Marking `waiting` on your answers above.

---

espg:
> A runtime DeprecationWarning on every beams shardmap build — wanted, or are the docstring/comment markers enough (the warning is the part most likely to be noisy)?

docstring/comment markers are enough

> Which trigger do you want encoded in the markers — just https://github.com/englacial/zagg/issues/66, or https://github.com/englacial/zagg/issues/66 plus data-virtualization (and should I open the tracking issue)?

later option-- yes, open a tracking issue for data virtualization.

this branch needs to reconcile conflicts with main before it can merge

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the merge commit `3a5c5f9` (merge of `origin/main` into `claude/65-beam-footprints`). One blocking-class merge artifact, otherwise the resolution is clean.

---

**🔴 Merge artifact: stale `src/zagg/processing.py` resurrected (clobbers #82 deletion)**

`src/zagg/processing.py` (1596 lines, the pre-#82 monolith) is present at HEAD *alongside* the `src/zagg/processing/` package:

```
$ git ls-tree HEAD:src/zagg/ | grep processing
100644 blob ... processing.py      <- should not exist
040000 tree ... processing         <- the #82 package
```

`origin/main` contains the package **only** — #82 (phase 4a, commit `022f832`) split the monolith into `processing/{read,aggregate,write,worker}.py` and *deleted* `processing.py`. The merge re-added the file from parent1 (the beam branch still carried the stale copy), so main's #82 deletion was effectively reverted.

Why it doesn't show up as a test failure: a package shadows a sibling module of the same name, so `import zagg.processing` resolves to `processing/__init__.py` (whose docstring even says "split out of a single ~2000-line `processing.py`"). The orphaned `processing.py` is unreachable dead code — which is exactly why the suite still passes 804/1. But it's a 1596-line ghost copy of logic that has since diverged in the package; it will rot, confuse `git grep`, and inflate the package. The fix is a one-liner: `git rm src/zagg/processing.py`. Diff-scoped and actionable.

---

**✅ Verified clean — no other issues found**

- **Merge resolution of `shardmap.py`**: all of main's #92/#89 logic survives — `_resolve_mortie_order`, `MORTIE_MOC_ORDER_CAP = 18`, `mortie_order: int | None = None`, the `_resolve_mortie_order(...)` call *before* `_intersect_mortie`, and `grid.spatial_signature()` at build time. The only `-` lines in the beam delta are benign rewrites (the `_intersect_mortie(...)` call gains `footprint=`/`product=`; the `moc_to_order` line moves into the per-ring loop) — no #92 logic deleted. The beam additions all layer on top: `_granule_footprints` seam, `footprint`/`product` params on both intersect helpers and `ShardMap.build`, per-ring loops with `dict.fromkeys` dedup, the non-beam-product `ValueError`, and `"footprint"` in metadata.
- **`processing/` package tree is byte-identical to `origin/main`** (same tree SHA) — the split itself was not touched.
- **No leftover conflict markers** anywhere in `src`/`tests`.
- **Deprecation markers** are docstring/comment-only, name the removal trigger, and cite **both #66 and #97** in beams.py (module docstring + `# DEPRECATED` banner), `is_beam_product`, `beam_tracks_from_cmr_polygon`, the `footprint` docstring in `make_shardmap`/`ShardMap.build`, and the CLI `--footprint` help. **No runtime `DeprecationWarning`/`warnings.warn` exists** in beams.py or shardmap.py — correct per the stated direction. The `# DEPRECATED` comment sits after the module docstring and before `from __future__ import annotations`, which is legal (the docstring is still the module's first statement; comments don't count).
- **Tests are meaningful, not trivially true**: `TestBeamHelper` asserts geometric containment and corridor widening with strict bounds; `TestBeamFootprintBehavior` asserts `bm < sw` (strict — beams hit *fewer* shards) across spherely/mortie/healpix, default-is-swath equivalence, metadata values, and the three `ValueError` paths; `TestIsBeamProduct` covers case-insensitivity and None/empty; `test_cli_rejects_bad_footprint` exercises the argparse choices. They use main's fixtures and weren't weakened.

Net: one merge artifact to drop (`src/zagg/processing.py`), nothing else blocking. No side-effecting actions implicated for @espg — the fix is in-scope for this PR's branch.

---

espg:
🤖 *from Claude*

Reconciled with `main`, added the deprecation markers, and folded the self-review. Done from my end — moving back to `waiting` for your review/merge.

## Conflicts reconciled with `main`

This branch was based on a pre-rework `main`; current `main` (root `9736cf5`) shares **no history** with the branch's base (`4d0932f`, root `eaae1da`) — `main` was rewritten under the branch (the `processing/` split #82, mortie MOC-order #92, `spatial_signature` #89, plus dispatch/registry/readers). So a plain merge reported "unrelated histories"; I merged with `--allow-unrelated-histories` (a merge, not a rebase, per your note — no force-push), which surfaced every shared file as an add/add conflict.

Resolution rule: **take `main`'s canonical version wholesale for every file the beam feature doesn't touch**, and hand-reconcile the three it does:

- `src/zagg/catalog/shardmap.py` — rebuilt from `main`'s file, then re-applied the beam delta on top: the `_granule_footprints` seam, `footprint`/`product` params on `_intersect_spherely` / `_intersect_mortie` / `ShardMap.build`, the per-ring loops with `dict.fromkeys` per-granule dedup, the non-beam-product `ValueError`, and `"footprint"` in metadata. `main`'s `_resolve_mortie_order` / `MORTIE_MOC_ORDER_CAP` / `mortie_order=None` / `grid.spatial_signature()` (#92/#89) are preserved verbatim — the beam intersect calls now sit *after* the `_resolve_mortie_order(...)` resolution.
- `src/zagg/catalog/__init__.py` — `main`'s file + the additive beam imports/params (`footprint` on `make_shardmap`, the CLI `--footprint`).
- `tests/test_shardmap.py` — `main`'s file + the beam test classes appended (`TestBeamHelper`, `TestBeamFootprintBehavior`, `TestIsBeamProduct`) and `test_cli_rejects_bad_footprint`; they run against `main`'s fixtures unchanged.
- `src/zagg/catalog/beams.py` is new, so it merged conflict-free.

The fresh-context self-review caught one real artifact: the merge had **resurrected the stale `src/zagg/processing.py` monolith** that #82 deleted from `main` (the branch still carried it; the package dir merely shadowed it, so tests passed on dead code). Removed it in `b63b75d` — `import zagg.processing` resolves to `processing/__init__.py` as on `main`. The `processing/` package tree is byte-identical to `main`.

## Deprecation markers (docstring/comment only)

Per your answer — markers, **no runtime `DeprecationWarning`**. Added a `.. deprecated::` note + `# DEPRECATED` banner atop `beams.py`, and deprecation notes on `beam_tracks_from_cmr_polygon`, `is_beam_product`, the `_granule_footprints` seam, the `footprint` docstring in `ShardMap.build` / `make_shardmap`, and the CLI `--footprint` help. Every marker names the **removal trigger**: native per-beam CMR geometry, **OR** the memory-handling robustness in #66, **OR** data virtualization in #97. The data-virtualization tracking issue is open as #97.

## Green

`uv run pytest -q` → **804 passed, 1 skipped** (the skip is the real spherely-fork `SpatialIndex` build, absent in the default venv). `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on the touched files. The two pre-existing morton-int64 failures flagged on the earlier commit no longer reproduce — current `main` is clean.

---

PR #69, opened 2026-06-20
Guard plan_read against ATL03 empty-segment sentinel (#68)

### LINKED ISSUE #68: plan_read OOM + silent data loss: ATL03 empty-segment sentinel corrupts base slices
🤖 *from Claude*

## Problem

`zagg.read_plan.plan_read` translates matched coarse segments to base-level photon
slices with `base_start = ph_index_beg[run_start] - index_base`. ATL03 marks
**empty segments** (no photons) with `ph_index_beg == 0` and `segment_ph_cnt == 0`.
When such a segment lands on a run boundary (e.g. via `pad`):

- **Start empty** → `base_start = 0 - 1 = -1 → clamp(0)` → the slice spans from
  photon 0 to the AOI (~half the granule, sometimes *more* than `n_base`).
  `total_base` blows past the `0.9 × n_base` selectivity threshold →
  `full_read=True` → `_planned_read_group` reads the **entire** granule
  (~50M photons × 6 datasets) → **Lambda OOM** (and ruinous compute).
- **End empty** → `base_end ≤ base_start` → the run is skipped → real AOI photons
  **silently dropped**.

## Impact (measured)

Full-mission NEON ATL03 run (256 shards, 64-cell chunks, `--footprint beams`):
**112/256 shards OOM'd**, ~37,200 GB-s / ~$0.50 for ~1M obs. Geolocation-only
profile of shard 81, granule `ATL03_20220427190246`:

| beam | n_base (photons) | OLD planned read | OLD full? | empties |
|---|---|---|---|---|
| gt1l | 59.2M | 106,595,710 | **True** | 3 |
| gt2l | 49.4M | 133,950,678 | **True** | 5 |
| gt3r | 22.4M | 20,527,450 | **True** | 1 |

3 of 10 beams forced a full-granule read; other beams (e.g. 4 empties) computed a
0-length read and dropped their real photons.

## Fix

Bound each run by its **non-empty** segments only (`count > 0`); skip all-empty
runs. On the same shard the full-reads go **3 → 0** and reads shrink from
tens-of-millions to **~1–4k photons/beam**. Draft PR incoming.

## Notes

- Worker-side code → requires a Lambda **function redeploy** (`build_function.sh`)
  to take effect, unlike the orchestrator-side beam footprints (#65).
- Distinct from the h5coro per-granule cache leak (#66); this is the **dominant**
  OOM driver — with this guard, per-shard reads are tiny enough that the leak no
  longer OOMs on its own.

_Authored by Claude for @espg._

espg:
Closes #68.

`plan_read` translated matched coarse segments to base-level photon slices using
the run-boundary segments' `ph_index_beg` directly. ATL03 marks **empty segments**
(no photons) with `ph_index_beg == 0` / `segment_ph_cnt == 0`, so a boundary empty
segment corrupted the slice:

- **start empty** → `base_start = 0 − index_base → clamp(0)` → slice spans photon 0
  to the AOI (~half the granule) → trips the `0.9 × n_base` selectivity threshold →
  `full_read` → `_planned_read_group` reads the whole granule → **Lambda OOM**;
- **end empty** → `base_end ≤ base_start` → run skipped → real photons **dropped**.

## Fix

Bound each run by its **non-empty** segments only (`count > 0`); a run with none is
skipped. One small change in `src/zagg/read_plan.py`'s slice-translation loop.

## How tested

- `pytest tests/test_read_plan.py -q` → **19 passed** (3 new in
  `TestPlanReadEmptySegments`: start-empty no-balloon, end-empty not-dropped,
  all-empty skipped). `ruff check` clean.
- **Real-data validation** (geolocation-only profile, shard 81 of the full-mission
  NEON map, granule `ATL03_20220427190246`):

  | beam | n_base | OLD planned read | OLD full? | NEW planned read | NEW full? | empties |
  |---|---|---|---|---|---|---|
  | gt1l | 59.2M | 106,595,710 | True | **1,160** | False | 3 |
  | gt2l | 49.4M | 133,950,678 | True | **1,407** | False | 5 |
  | gt3r | 22.4M | 20,527,450 | True | **2,858** | False | 1 |

  Full-reads across the shard's beams: **OLD 3 → NEW 0**. Beams that previously
  computed a 0-length read (silent drop) now read their real photons.

## Notes

- **Worker-side** code → needs a Lambda **function redeploy** (`build_function.sh`)
  to take effect on the fleet; nothing changes for existing shardmaps.
- Independent of the h5coro per-granule cache leak (#66); this is the dominant OOM
  driver in the full-mission run (112/256 shards OOM'd pre-fix).
- `_planned_read_group`'s planned-vs-full parity is preserved: empty segments carry
  no photons, so excluding them from the slice bounds changes which bytes are read,
  not which photons land in the AOI.

_Authored by Claude for @espg._

### DIFF
--- a/src/zagg/read_plan.py
+++ b/src/zagg/read_plan.py
@@ -150,6 +150,11 @@ def plan_read(
             merged.append((s, e))
 
     # -- Translate to base slices --
+    # ``kept_runs`` stays one-to-one with ``base_slices``/``chunk_lists``: a run
+    # that yields no slice (all-empty, or collapsed) is dropped from all three,
+    # so the returned plan never has phantom ``parent_runs`` without matching
+    # reads (which would crash ``execute_read_plan`` on an empty concatenate).
+    kept_runs: list[tuple[int, int]] = []
     base_slices: list[tuple[int, int]] = []
     chunk_lists: list[list[tuple[int, int]]] = []
     total_base = 0
@@ -162,7 +167,9 @@ def plan_read(
         # spans from photon 0 to the AOI (~half the granule). That balloons
         # total_base past the selectivity threshold -> full-granule read -> OOM
         # (and ruinous compute). Bound each run by its NON-EMPTY segments only;
-        # empty ones carry no photons, so a run with none is skipped.
+        # empty ones carry no photons, so a run with none is skipped. (The single
+        # contiguous [first, last] slice relies on the #43 contiguity assumption:
+        # non-empty segments are monotonic in ph_index_beg and tile the photons.)
         local = np.flatnonzero(cnt[s : e + 1] > 0)
         if local.size == 0:
             continue
@@ -173,6 +180,7 @@ def plan_read(
         base_end = min(n_base, base_end)
         if base_end <= base_start:
             continue
+        kept_runs.append((s, e))
         base_slices.append((base_start, base_end))
         chunk_lists.append([(base_start, base_end)])  # h5coro half-open [start, end)
         total_base += base_end - base_start
@@ -187,7 +195,7 @@ def plan_read(
         )
 
     return ReadPlan(
-        parent_runs=merged,
+        parent_runs=kept_runs,
         base_slices=base_slices,
         chunk_lists=chunk_lists,
     )
--- a/src/zagg/read_plan.py
+++ b/src/zagg/read_plan.py
@@ -153,9 +153,22 @@ def plan_read(
     base_slices: list[tuple[int, int]] = []
     chunk_lists: list[list[tuple[int, int]]] = []
     total_base = 0
+    ibeg = np.asarray(index_beg_arr)
+    cnt = np.asarray(count_arr)
     for s, e in merged:
-        base_start = int(index_beg_arr[s]) - index_base
-        base_end = int(index_beg_arr[e]) - index_base + int(count_arr[e])
+        # ATL03 marks empty segments (no photons) with count == 0 and
+        # ph_index_beg == 0. Using such a segment as a run boundary translates
+        # to a bogus slice: base_start = 0 - index_base clamps to 0, so the read
+        # spans from photon 0 to the AOI (~half the granule). That balloons
+        # total_base past the selectivity threshold -> full-granule read -> OOM
+        # (and ruinous compute). Bound each run by its NON-EMPTY segments only;
+        # empty ones carry no photons, so a run with none is skipped.
+        local = np.flatnonzero(cnt[s : e + 1] > 0)
+        if local.size == 0:
+            continue
+        first, last = s + int(local[0]), s + int(local[-1])
+        base_start = int(ibeg[first]) - index_base
+        base_end = int(ibeg[last]) - index_base + int(cnt[last])
         base_start = max(0, base_start)
         base_end = min(n_base, base_end)
         if base_end <= base_start:
--- a/tests/test_read_plan.py
+++ b/tests/test_read_plan.py
@@ -152,6 +152,7 @@ class TestPlanReadEmptySegments:
         plan = plan_read(lats, lons, ibeg, cnt, n_base, bbox, index_base=1, pad=0)
         assert not plan.full_read  # pre-fix this ballooned to [0, 5200] -> selectivity full read
         assert plan.base_slices == [(5000, 5200)]  # non-empty segs 1..2 only
+        assert plan.parent_runs == [(0, 2)]  # parent_runs stays in lockstep
 
     def test_empty_at_run_end_not_dropped(self):
         # Empty segment at the run end must not collapse the slice to nothing.
@@ -163,9 +164,24 @@ class TestPlanReadEmptySegments:
         bbox = (0.0, -0.5, 1.0, 0.5)  # matches segs 0,1,2
         plan = plan_read(lats, lons, ibeg, cnt, n_base, bbox, index_base=1, pad=0)
         assert plan.base_slices == [(5000, 5200)]  # bounded by non-empty seg 1
+        assert plan.parent_runs == [(0, 2)]
 
-    def test_all_empty_run_skipped(self):
-        # A run with only empty segments contributes no photons and is dropped.
+    def test_empty_in_run_middle_stays_contiguous(self):
+        # An empty interior segment consumes no photon indices, so the two
+        # non-empty neighbours are adjacent -- one contiguous slice covers both.
+        lats = np.array([0.00, 0.10, 0.20, 40.0])
+        lons = np.full(4, 0.5)
+        ibeg = np.array([5001, 0, 5101, 5301])  # seg 1 empty (middle)
+        cnt = np.array([100, 0, 100, 100])
+        n_base = 5400
+        bbox = (0.0, -0.5, 1.0, 0.5)  # matches segs 0,1,2
+        plan = plan_read(lats, lons, ibeg, cnt, n_base, bbox, index_base=1, pad=0)
+        assert plan.base_slices == [(5000, 5200)]  # seg0 [5000,5100) + seg2 [5100,5200)
+        assert plan.parent_runs == [(0, 2)]
+
+    def test_all_empty_run_skipped_no_crash(self):
+        # A run with only empty segments contributes no photons. parent_runs MUST
+        # stay empty in lockstep, else execute_read_plan hits np.concatenate([]).
         lats = np.array([0.00, 0.10, 40.0])
         lons = np.full(3, 0.5)
         ibeg = np.array([0, 0, 5301])
@@ -174,7 +190,11 @@ class TestPlanReadEmptySegments:
         bbox = (0.0, -0.5, 1.0, 0.5)  # matches segs 0,1 (both empty)
         plan = plan_read(lats, lons, ibeg, cnt, n_base, bbox, index_base=1, pad=0)
         assert plan.base_slices == []
+        assert plan.parent_runs == []  # lockstep -> consumer short-circuits, no phantom run
         assert not plan.full_read
+        # consumer contract: must return empty, not raise on an empty concatenate.
+        out = execute_read_plan(plan, lambda *a, **k: np.array([]), "any/path", float)
+        assert out.size == 0
 
 
 class TestExecuteReadPlan:
--- a/tests/test_read_plan.py
+++ b/tests/test_read_plan.py
@@ -133,6 +133,50 @@ class TestPlanReadIndexBase:
         assert plan.base_slices[0] == (0, 5)  # 1-1=0, 0+5=5
 
 
+class TestPlanReadEmptySegments:
+    """ATL03 marks empty segments with ph_index_beg == 0 / count == 0. Using one
+    as a run boundary translated to a bogus slice -- balloon to photon 0 (->
+    full-read OOM) or collapse to nothing (-> silent data loss). The run must be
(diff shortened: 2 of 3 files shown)


src/zagg/read_plan.py
```diff
@@ -153,9 +153,22 @@ def plan_read(
...
+        # ATL03 marks empty segments (no photons) with count == 0 and
+        # ph_index_beg == 0. Using such a segment as a run boundary translates
+        # to a bogus slice: base_start = 0 - index_base clamps to 0, so the read
+        # spans from photon 0 to the AOI (~half the granule). That balloons
+        # total_base past the selectivity threshold -> full-granule read -> OOM
+        # (and ruinous compute). Bound each run by its NON-EMPTY segments only;
+        # empty ones carry no photons, so a run with none is skipped.
+        local = np.flatnonzero(cnt[s : e + 1] > 0)
+        if local.size == 0:
+            continue
```
espg, line 175: 🤖 *from Claude (review)*

**Blocker — crash regression: `parent_runs` and `base_slices` can now disagree, breaking the `execute_read_plan` contract.**

When a matched run is composed *entirely* of empty segments, this `continue` drops it from `base_slices`/`chunk_lists`, but `parent_runs` is still set to `merged` (line 190) and therefore retains that run. The two lists are no longer one-to-one.

Downstream this is a hard crash, not a no-op:
- `_planned_read_group` (processing.py:941) guards with `if not plan.parent_runs: return None`. With an all-empty matched run, `parent_runs == [(0,0)]` is **truthy**, so it proceeds.
- It then calls `execute_read_plan(...)` (processing.py:973). `execute_read_plan` (read_plan.py:219) also checks `if not plan.parent_runs` (truthy → proceeds), iterates the **empty** `chunk_lists`, and hits `np.concatenate([])` → `ValueError: need at least one array to concatenate`.

Reproduced:
```python
lats=np.array([0.0,40.0,80.0]); lons=np.full(3,0.5)
ibeg=np.array([0,5101,5301]); cnt=np.array([0,100,100])
plan=plan_read(lats,lons,ibeg,cnt,5400,(0,-0.5,1,0.4),index_base=1,pad=0)
# plan.parent_runs == [(0,0)]  but  plan.base_slices == []  chunk_lists == []
execute_read_plan(plan, lambda p,hyperslice=None: np.arange(10), 'x', np.float64)
# -> ValueError: need at least one array to concatenate
```

Note the new `test_all_empty_run_skipped` already constructs exactly this state — its `plan.parent_runs == [(0,1)]` while `base_slices == []` — but only asserts on `base_slices`, so it doesn't surface the crash.

**Fix options:** keep the invariant `len(parent_runs) == len(base_slices)` — accumulate `parent_runs` inside this loop in lockstep with `base_slices` (only append the run when a slice is produced), instead of returning `merged` wholesale at line 190. Alternatively, after the loop, `if not base_slices: return ReadPlan(parent_runs=[], base_slices=[], chunk_lists=[])` so the empty-AOI short-circuit in the caller fires. Add a test asserting both `parent_runs == []` (or in-step) **and** that `execute_read_plan` returns an empty array for the all-empty-run plan.

src/zagg/read_plan.py
```diff
@@ -153,9 +153,22 @@ def plan_read(
...
+        first, last = s + int(local[0]), s + int(local[-1])
+        base_start = int(ibeg[first]) - index_base
+        base_end = int(ibeg[last]) - index_base + int(cnt[last])
```
espg, line 178: 🤖 *from Claude (review)*

**Should-fix (document the assumption): the single contiguous `[ibeg[first], ibeg[last]+cnt[last])` slice silently relies on non-empty segments being contiguous and monotonic.**

`base_end` is derived solely from `last`'s `ibeg`/`cnt`. This is correct only if, across the run's non-empty segments, (a) `ibeg` is monotonically non-decreasing and (b) the photons tile contiguously so that `[ibeg[first], ibeg[last]+cnt[last])` covers every matched segment's true photon range. ATL03 satisfies this for real granules, but two cases are worth noting:

- **Gap between non-empty segments** (an empty segment whose photon range is *not* reabsorbed by neighbors, or any true gap): the slice reads the gap photons, which belong to an unmatched segment. Parity still holds (those photons get re-filtered by `mask_spatial`/cross-level identically to the full path), but it reads more than necessary — partially eroding the IO/OOM win this PR targets, and `n_base = cnt.sum()` in the caller no longer equals `ibeg[-1]-index_base+cnt[-1]` (the contiguity note at processing.py:903-908). Verified:
```python
ibeg=np.array([5001,0,5201,5301]); cnt=np.array([100,0,100,100])  # seg2 starts 5200, gap 5100..5200
plan_read(..., index_base=1, pad=0).base_slices  # -> [(5000, 5300)]  (reads the 100-photon gap)
```
- **Non-monotonic `ibeg` among non-empty segments**: `[first..last]` would under-read (a matched segment outside that bound is missed → silent data loss, a parity break). Not expected for ATL03, but nothing in `plan_read` enforces it.

This isn't a blocker given ATL03's layout, but the loop comment only justifies *skipping* empty segments — it should also state the contiguity+monotonicity precondition (mirroring `_expand_mask_to_base`'s "ranges do not overlap and together tile the full base array"). Consider asserting monotonic non-empty `ibeg` in debug, or at least cross-referencing #43's contiguity assumption here.

tests/test_read_plan.py
```diff
@@ -133,6 +133,50 @@ def test_index_base_1_atl03_style(self):
...
+        assert plan.base_slices == [(5000, 5200)]  # bounded by non-empty seg 1
+
+    def test_all_empty_run_skipped(self):
+        # A run with only empty segments contributes no photons and is dropped.
+        lats = np.array([0.00, 0.10, 40.0])
+        lons = np.full(3, 0.5)
+        ibeg = np.array([0, 0, 5301])
+        cnt = np.array([0, 0, 100])
+        n_base = 5400
+        bbox = (0.0, -0.5, 1.0, 0.5)  # matches segs 0,1 (both empty)
```
espg, line 190: 🤖 *from Claude (review)*

**Should-fix (test coverage): the tests validate `base_slices` but not the consumer contract, and there's no end-to-end parity test with empty segments.**

Two gaps in this otherwise good test class:

1. `test_all_empty_run_skipped` asserts `base_slices == []` but **not** `parent_runs`. As written, this plan has `parent_runs == [(0,1)]` (non-empty) while `base_slices == []` — exactly the inconsistent state that crashes `execute_read_plan` (see the blocker on read_plan.py:168). Tighten this test to also assert `plan.parent_runs == []` and that `execute_read_plan(plan, ...)` returns an empty array — that assertion would have caught the `np.concatenate([])` crash.

2. The headline #43 property is *row-for-row parity between the planned and full read paths*, but all three new tests stop at `plan_read`'s `base_slices`. None drives `_planned_read_group` (or even `execute_read_plan`) with an empty-segment granule to confirm the concatenated read + `global_idx` + `mask_spatial` still line up. Recommend at least one test that runs a small synthetic granule with an interior empty segment through `execute_read_plan` and checks the output equals the full-read slice — that's where the cross-level `global_idx = concatenate([arange(s,e) for s,e in base_slices])` (processing.py:1011) interaction with the now-tighter slices actually gets exercised.

Also consider a `test_empty_in_run_middle` (empty segment with non-empty on both sides) — currently only start/end-of-run empties are covered.

espg:
🤖 *from Claude*

Folded the clean-context adversarial review (commit `c1440fc`):

- **Blocker** — all-empty / collapsed runs left `parent_runs = merged` while `base_slices`/`chunk_lists` were empty, so `execute_read_plan` hit `np.concatenate([])` (the same crash seen in early sweeps). Fixed by tracking `kept_runs` in lockstep with `base_slices`; an all-empty run now yields `parent_runs == []`, so `_planned_read_group` short-circuits to `None`. This also closes the pre-existing collapse-case inconsistency.
- **should-fix (tests)** — added `parent_runs` lockstep assertions, an empty-in-middle case, and a consumer-contract test that drives `execute_read_plan` on the all-empty plan (would have caught the blocker). 20 read_plan + 13 planned-read-group tests pass.
- **should-fix (docs)** — noted the single-contiguous-slice precondition (the #43 contiguity assumption: non-empty segments monotonic in `ph_index_beg` and photon-tiling) inline.

Parity for normal cases and `global_idx` alignment were investigated and cleared by the review.

espg:
🤖 *from Claude (review)*

Final pre-merge pass — the lockstep fix is correct and complete; clean to merge. One non-blocking test-coverage gap worth a follow-up (not a merge blocker).

**Verified correct (suspicions cleared):**

- **Lockstep on every return path.** `parent_runs`/`base_slices`/`chunk_lists` stay 1:1 across all four returns: the two early `[]` returns, the `full_read` fallback (`[(0,n_coarse-1)]` / `[(0,n_base)]`, consumed via `full_read=True` first so the slices are inert), and the normal `kept_runs` path. Asserted `len(parent_runs)==len(base_slices)==len(chunk_lists)` across empty-start / empty-end / empty-mid / all-empty fixtures.
- **No consumer broke from dropping runs.** `_planned_read_group` uses `plan.parent_runs` only as a truthiness guard (`processing.py:941`); `global_idx` is built from `plan.base_slices` (`processing.py:1011-1013`), never from `parent_runs`. Nothing indexes coarse arrays by `parent_runs` or relies on `len(parent_runs)==len(merged)`. A strict-subset `parent_runs` is safe everywhere.
- **`global_idx` / Phase-B alignment.** With empty-segment-tightened `base_slices`, the concatenated planned read equals `global_idx` index-for-index and `gidx.max() < n_base`, so `cross_full[global_idx][mask_spatial]` lines up with `keep_mask`. Confirmed against the empty-segment fixtures.
- **Gap/monotonicity (the documented should-fix).** A genuine within-run photon gap between non-empty segments produces an **over-read** (slice spans the gap), which the spatial mask removes — parity holds, only mild inefficiency. The only under-read mode is non-monotonic `ph_index_beg` among non-empty segments, which is impossible per the ATL03 spec and *fails safe* anyway (`base_end <= base_start` → run dropped, not silently truncated). Documentation is sufficient; no guard needed for merge.
- Prior blocker (`np.concatenate([])` on all-empty) confirmed fixed: all-empty match → `parent_runs == []` → `_planned_read_group` short-circuits to `None`, and `execute_read_plan` returns `np.empty(0)` via the `if not plan.parent_runs` guard.

**Non-blocking (follow-up):** the new tests assert `plan_read` outputs and one `execute_read_plan` contract, but there is no **end-to-end planned-vs-full parity test through `_planned_read_group` with empty (count==0) segments present**. The existing parity tests (`test_parity_with_full_read`, `test_parity_with_full_read_includes_leaf_id`, `test_multi_slice_plan_global_idx_alignment`) all use fully-populated fixtures. I verified the alignment holds by hand, so this is coverage-hardening, not a correctness gap — worth adding an empty-segment row to the parity harness (`_LatBboxGrid` + `_planned_read_h5` with a zero-count seg) in a follow-up, but it does not block this OOM fix.

espg:
🤖 *from Claude*

Closed the second review's one open item (the end-to-end parity coverage gap): added `TestPlannedReadGroup::test_parity_with_empty_segment` (`586e0de`) — drives `_read_group` planned vs full with an ATL03 empty-segment sentinel pulled into the matched run, asserting row-for-row identical output. This collapse case errored on the pre-fix path (planned returned `None`, dropping real photons), so the #43 parity property is now CI-locked for empty segments, not just hand-verified.

PR #70, opened 2026-06-20
Temporal aggregation infrastructure (Refs #12)

espg:
Refs #12. Per the direction in https://github.com/englacial/zagg/issues/12#issuecomment-4749726575 ("Do it, fresh branch of main / fresh PR.") — this PR is a fresh start off `main`, replacing the stale `magg_temporal` / `zagg-temporal` branches that diverged before #28–#41 landed.

Implements the temporal pipeline infrastructure (issue #12) in additive phases. Phase 1 lands the smallest independently mergeable seam: a `pipeline.type` field on `PipelineConfig` plus a `get_pipeline_type(config)` helper that the dispatch strategy reads. Subsequent phases layer the worker, dispatch strategy, output writers, and the Lambda temporal worker + fan-out on top without changing the spatial production path until the strategy seam itself flips.

## Reconcile onto merged infra (#73 registry, #76 dispatch)

Per the **rebase-not-restart** decision (https://github.com/englacial/zagg/issues/64#issuecomment-4771358325), the two pieces this PR originally drafted as its own Phases 2 and 3 have **landed canonically on `main`** via dedicated PRs, and this branch reconciles onto them:

- **Phase 2 (registry) — LANDED via #73.** Canonical registry is `src/zagg/registry.py` on `main`.
- **Phase 3 (dispatch) — LANDED via #76.** Canonical dispatcher is `src/zagg/dispatch.py` on `main`.

## Phases

- [x] **Phase 1 — config schema + `get_pipeline_type`** (commit `b9adafa`). `pipeline.type` defaults to `"spatial"`; every existing YAML stays valid byte-for-byte.
- [x] **Phase 2 — capability registry — done via #73.**
- [x] **Phase 3 — dispatcher — done via #76.**
- [x] **Phase 4 — `src/zagg/temporal.py` primitives + `process_event`** (commit `52ff1c4`, fold `4ed4434`).
- [x] **Phase 5 — `PipelineStrategy` dispatch in `runner.agg()`** (commit `0d20e70`). `SpatialStrategy` (byte-identical) + `TemporalStrategy`.
- [x] **Phase 6 — `src/zagg/output/` tabular writer** (commit `c7a149d`). `TabularWriter` for Parquet/CSV temporal output. (The Phase-6 `Writer` Protocol / `register_writer` registry / `ZarrGridWriter` adapter were later removed as dead code per the issue #161 review — see the post-review reconcile below; the spatial path keeps calling the `zagg.processing` Zarr writers directly.)
[...]

Issue #63, opened 2026-06-19
Dispatcher abstraction as shared infrastructure (#20, #12)


espg:
The most impactful single investment. Both temporal (https://github.com/englacial/zagg/issues/12) and future cluster backends (https://github.com/englacial/zagg/issues/20) need the same generic fan-out → retry → measured-cost loop that currently lives as two bespoke functions in runner.py. A dispatch.py with a clean Executor protocol extracts this once. Every pipeline type (spatial, temporal, event) then inherits Lambda and local execution for free, and items 3–4 of https://github.com/englacial/zagg/issues/20 (vaex, ray/dask/slurm) plug in behind the same interface without touching pipeline logic.

This is the prerequisite that unblocks both https://github.com/englacial/zagg/issues/12 Phase 1c and https://github.com/englacial/zagg/issues/20 items 3–4.

espg:
What's the main things we need to think about in designing this? What are the major decisions that need to be made?

espg:
🤖 *from Claude*

Agreed this is the highest-leverage single investment — both #12 and #20 items 3–4 collapse into "another Executor implementation" once the loop is extracted. A few clarifying questions, then three design options for the `Executor` protocol's surface area.

### Where the duplication actually lives today

In `runner.py` as of `main` (lines for grounding):

- `_run_local(...)` at ~307 — sequential per-shard loop, no retry/cost machinery (it's the trivial executor).
- `_run_lambda(...)` at ~395 — the rich one: `ThreadPoolExecutor` + boto3 invoke + retry + log parsing + cost estimation + concurrency probe (the post-#41 surface).
- `_invoke_lambda_setup` / `_invoke_lambda_finalize` / `_invoke_lambda_cell` at ~643 / ~671 / ~690 — the per-cell hooks.
- Plus `concurrency.py` (probe / FD ceiling / errno-24 guidance) which `_run_lambda` pre-flights.

So the "generic" loop has these moving parts: dispatch (sync or fan-out), retry policy, per-cell error classification (errno-24 special-cased today), per-cell cost/log parsing, aggregate cost reporting, pre-flight capacity check, finalize step. A clean `Executor` protocol needs to decide which of those are protocol-level vs executor-private.

### Three design options for the protocol surface

**(A) Thin protocol — `submit(payload) → Future[result]` only.**

```python
class Executor(Protocol):
    def submit(self, payload: CellPayload) -> Future[CellResult]: ...
    def shutdown(self) -> None: ...
```

[...]

Issue #59, opened 2026-06-18
MCP Server


espg:
This is post version 1.0 release.

The eventual goal is to provide an MCP server where a user can describe in natural language what data from nasa they'd like (products, time range, spatial AOI) , have the MCP connection provide appropriate context and skills to the agent to build the needed aggregation templates, and kick off the local run (or potentially lambda run).

Broad strokes, what's needed for this? What do we need to be thinking about and designing towards so we can enable this functionality in the future?

espg:
🤖 *from Claude*

**Broad strokes for a zagg MCP server — what's needed and what to design toward:**

The core idea maps cleanly onto existing zagg abstractions. Here's the shape of the problem:

**1. Gate: stable v1 API**
MCP capabilities break if the underlying library changes. The MCP work should follow the v1 milestone (#58), not precede it. Pre-v1 iteration on the server would mean constant breakage.

**2. Capability enumeration surface**
The plugin registry from the temporal work (#12, Phase 1a) is directly useful here: `registry.list_spatial_functions()`, `registry.list_reducers()`, `registry.list_catalog_sources()` give the MCP server a self-describing capability map. This is the same registry that makes temporal pipelines extensible — shared infrastructure, not a second system to maintain.

**3. Natural MCP skill decomposition (three tools):**
- **`zagg.describe_products`** — returns available products (ATL06, ATL03, MERRA-2), their variables, supported grid types, default templates. Gives the LLM the context to understand what's configurable.
- **`zagg.build_template`** — takes natural-language intent (product, AOI, aggregation goal) and produces a validated YAML template. The LLM generates YAML; `validate_config()` is the acceptance gate. No execution, just construction.
- **`zagg.run`** — takes a validated template + backend choice (local or Lambda) and calls `agg()`. Returns a run ID or streams progress.

**4. Open design questions that need answering before implementation:**
- **AOI input format**: GeoJSON Feature, `[lon1, lat1, lon2, lat2]` bbox, or WKT? Bbox is the simplest MCP tool parameter; GeoJSON handles complex AOIs (ice sheet basins, political boundaries).
- **Authentication scope**: earthaccess tokens need to flow from the user environment into the MCP server. Passthrough from the calling process is simplest; a separate auth flow adds friction but is more explicit.
- **Local vs Lambda default**: `backend="local"` should be the default (no AWS required) with Lambda as opt-in. Makes the server usable without cloud credentials.
[...]

Issue #12, opened 2026-03-14
Temporal aggregation


espg:
## Motivation

The [antarctic_AR_dataset](https://github.com/englacial/antarctic_AR_dataset) repo implements a Lambda-based aggregation pipeline for computing storm attributes from MERRA-2 reanalysis data. The architecture there — declarative metric definitions, Lambda orchestration, catalog-driven dispatch, S3 data access — is structurally very similar to what magg already does for spatial aggregation. This issue proposes unifying the temporal aggregation pattern into magg as a second pipeline type.

## Common Design Elements

Both repos share similar architecture at the structural level:

| Pattern | magg | antarctic_AR_dataset |
|---|---|---|
| **Declarative metrics** | YAML config with `function`/`expression` strings | `AggregationSpec` dataclass with `spatial_func`/`temporal_reducer` strings |
| **Function dispatch** | `resolve_function("np.min")` → callable | String lookup in `spatial_functions.py` |
| **Lambda orchestration** | Catalog → dispatch → collect results | Catalog → dispatch → collect results |
| **Catalog building** | CMR query → morton→granule_urls mapping | earthaccess query → collection→date→url index |
| **S3 data access** | h5coro byte-range reads | s3fs + xarray |
| **Auth** | `earthaccess.login()` → temp S3 creds | Same |
| **Output** | Zarr (gridded) | HDF5/DataFrame (tabular) |

The orchestrator/worker/catalog trinity is nearly identical. The config-driven function dispatch is the same idea expressed differently (YAML vs dataclass).

## Key Architectural Differences

**1. Processing axis — this is the fundamental divergence**
- **magg**: spatial. Group irregular points by cell, aggregate *within* each cell. No time dimension.
- **AR**: temporal. Track an object across timesteps, apply spatial funcs per timestep, then reduce *across* time.

**2. Statefulness**
- **magg**: stateless — load all data, groupby, apply. Each cell is independent.
- **AR**: stateful — streaming accumulators (`MaxAccumulator`, `WeightedMeanAccumulator`) maintain running state across timesteps because you can't hold all timesteps in memory at once.

**3. Processing unit**
- **magg**: a fixed spatial cell (morton parent at order 6, ~100km)
- **AR**: a dynamic spatiotemporal object (storm with evolving binary mask)

**4. Data model**
- **magg**: point cloud (irregular observations) → regular grid
- **AR**: regular grid (MERRA-2) → scalar attributes per object

## What Could Be Unified

[...]

PR #73, opened 2026-06-22
Capability registry — shared name→object vocabulary (#64)

### LINKED ISSUE #64: Plugin registry as shared vocabulary (#12, #59)
Phase 1a of https://github.com/englacial/zagg/issues/12 introduces registry.py with named registries for spatial functions, reducers, mask providers, etc. This is also the capability-enumeration surface the MCP server (https://github.com/englacial/zagg/issues/59) needs to describe available aggregation options — registry.list_spatial_functions() directly feeds the MCP describe_products tool.

Shared design decision worth making early: all capability references stay as names (strings), never callables, so they serialize in YAML/JSON for both Lambda payloads and MCP tool parameters. This invariant must hold across the temporal engine, the plugin mechanism, and the MCP server.

espg:
Closes #64.

Implements the plugin/capability registry as the shared **name→object** vocabulary for configs, Lambda payloads, and the MCP `describe_products` tool (#59), per the design locked with @espg on the issue thread (https://github.com/englacial/zagg/issues/64#issuecomment-4771358325).

## What / approach

New `src/zagg/registry.py` — a `Registry[T]` class core with thin per-kind functional wrappers:

- **`Registry[T]`** with `register` / `get` / `list` / `describe` / `describe_all` over the **eight** capability kinds from the [June plan](https://github.com/englacial/zagg/issues/12#issuecomment-4635480666) (`spatial_func`, `reducer`, `mask_provider`, `field_transform`, `event_trigger`, `reader`, `catalog_source`, `credential_provider`). `register` doubles as a decorator (`@SPATIAL_FUNCS.register("max")`) or a direct call.
- **Strings, never callables** — the registry maps a public *name* to a private *object*; configs/payloads carry the name and re-resolve at the execution site.
- **Optional per-entry `description` + `schema`** — the locked "(B) now, (C) where it earns its keep" path: entries without a schema pay nothing; `describe` surfaces a pydantic/JSON arg schema when a registrant supplies one. This is the additive route to the full MCP parameter surface without forcing every capability author to write a schema up front.
- **`UnknownCapability(name, kind, available=[...])`** — subclasses `KeyError` (so existing `except KeyError` paths still catch it) while carrying the kind + sorted available names for a clean "did you mean…" / MCP-relayable error.
- **Lazy `zagg.plugins` entry-point discovery** — first `get`/`list` triggers a one-time sweep; a broken plugin's `load()`/`register()` is logged-and-skipped, and a failure in `entry_points()` itself leaves the seam retryable. Includes the re-entrancy guard for a plugin whose `register()` calls back through `get`.
- **`describe_all()`** returns the structured `{kind: [{name, kind, description, schema?}, …]}` shape #59 consumes; `registry_snapshot()` keeps the name-only view for diagnostics.
- **Per-kind functional wrappers** (`register_spatial_func` / `get_spatial_func` / `list_spatial_funcs`, ×8) so call sites read clearly and miss-messages name the kind — names match #70's existing surface so its rebase (below) is minimal.

Built-ins register nothing yet — this ships the seam alone; the temporal engine's mask providers / field transforms / etc. register into it as their owning modules land.

## Relationship to #70

#70's Phase 2 shipped a *functional* registry; this PR lands the **canonical `Registry[T]` class core** per #64's locked design. As agreed (rebase, not restart — https://github.com/englacial/zagg/issues/63#issuecomment-4771921400), #70 rebases its registry onto this one and drops its copy; the functional wrapper names are kept identical here so that rebase is near-mechanical.

## Phases

- [x] `Registry[T]` core + the eight registries + `UnknownCapability` + optional `description`/`schema`
- [x] Lazy `zagg.plugins` entry-point discovery (re-entrancy guard + retry-on-lookup-failure)
- [x] `describe_all()` structured MCP surface + `registry_snapshot()` + per-kind functional wrappers
- [x] Tests (`tests/test_registry.py`)

## How tested

- `uv run pytest tests/test_registry.py -q` → **49 passed**. Covers: the eight-registry invariant; direct + decorator registration round-trips across all eight; duplicate→`ValueError` unless `replace=True`; empty/non-string name→`ValueError`; `UnknownCapability` carries kind + `available` and is a `KeyError`; optional `description`/`schema` surfaced by `describe`/`describe_all` (and omitted when absent); structured + sorted `describe_all`; lazy discovery (entry-point `register` runs on first `get`, broken `load()`/`register()` skipped, `entry_points()` failure retries, re-entrancy, `discover_plugins(force=True)`); and the name→same-object invariant.
- `uv run ruff check --select=E,F,W,I --ignore=E501 src/zagg/registry.py tests/test_registry.py` → clean.

## Questions for review

- **Functional wrapper naming:** kept `list_spatial_funcs` (etc.) to match #70's existing surface for a minimal rebase; the issue body wrote `list_spatial_functions()` illustratively. Flag if you'd prefer the longer form as the canonical name.
- **`schema` field type** is left as `Any` (a pydantic model *or* a JSON-schema dict) until #59 pins the MCP parameter encoding. Confirm that's the right level of looseness for now.

_Authored by Claude for @espg under the #64 routine._

---

### DIFF
--- a/src/zagg/registry.py
+++ b/src/zagg/registry.py
@@ -1,0 +1,450 @@
+"""Capability registry — the shared name→object vocabulary (issue #64).
+
+zagg resolves *capabilities* — spatial functions, reducers, mask providers,
+etc. — by **name** rather than by passing callables through a config. Names
+round-trip through YAML, Lambda JSON payloads, and the MCP `describe_products`
+tool (#59); callables do not. This module owns the eight capability namespaces
+named in the June 2026 plan
+(https://github.com/englacial/zagg/issues/12#issuecomment-4635480666):
+
+==================  ===================================================
+Registry            What it holds
+==================  ===================================================
+spatial_func        Per-timestep reductions over a 2-D field (``max``)
+reducer             Streaming accumulators across timesteps (``Max``)
+mask_provider       Producers of boolean masks (``ais``, ``ocean``)
+field_transform     Per-timestep transforms (``monthly_anomaly``)
+event_trigger       Predicates that fire on a specific timestep
+reader              Multi-collection data readers
+catalog_source      Catalog adapters (CMR, earthaccess, …)
+credential_provider Credential adapters (NSIDC, GES-DISC, EDL, …)
+==================  ===================================================
+
+Design (issue #64, locked with @espg):
+
+- **Strings, never callables.** The registry maps a public *name* to a private
+  *object*; configs/payloads carry the name and re-resolve the object at the
+  execution site. The name is the stable identifier, the object is versioned.
+- **`Registry[T]` class core** with `register` / `get` / `list` / `describe`.
+- **Optional per-entry schema.** Each registration may carry a one-line
+  ``description`` and an optional ``schema`` (a pydantic model / JSON schema for
+  the entry's YAML args). Entries without a schema pay nothing; ``describe``
+  surfaces it when present. This is the additive "(B) now, (C) where it earns
+  its keep" path to the full MCP parameter surface.
+- **Lazy entry-point discovery.** External packages contribute by declaring a
+  ``zagg.plugins`` entry point pointing at a zero-arg ``register()`` callable;
+  the first ``get`` / ``list`` triggers discovery once.
+- **`UnknownCapability`** carries the kind + the sorted ``available`` names so a
+  miss is a clean, propagatable error (good "did you mean…" surface for MCP).
+
+Built-in capabilities register when the later phases land their owning modules;
+this module ships the seam alone — every registry starts empty.
+"""
+
+from __future__ import annotations
+
+import logging
+from collections.abc import Callable, Iterable
+from dataclasses import dataclass
+from importlib import metadata
+from typing import Any, Generic, TypeVar
+
+logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+
+# Entry-point group external plugins declare in their ``pyproject.toml``::
+#
+#     [project.entry-points."zagg.plugins"]
+#     antarctic_ar = "zagg_ar:register"
+_ENTRY_POINT_GROUP = "zagg.plugins"
+
+
+class UnknownCapability(KeyError):
+    """Raised when a capability name is not registered for its kind.
+
+    Subclasses ``KeyError`` so existing ``except KeyError`` paths still catch
+    it, while carrying ``kind`` and the sorted ``available`` names for a
+    diagnosable message (and a clean error type for the MCP server to relay).
+    """
+
+    def __init__(self, name: str, kind: str, available: Iterable[str]):
+        self.name = name
+        self.kind = kind
+        self.available = list(available)
+        super().__init__(f"Unknown {kind} {name!r}; registered: {self.available}")
+
+    def __str__(self) -> str:  # KeyError stringifies via repr(); override for readability
+        return f"Unknown {self.kind} {self.name!r}; registered: {self.available}"
+
+
+@dataclass(frozen=True)
+class Entry(Generic[T]):
+    """One registered capability: its name, object, and optional metadata."""
+
+    name: str
+    obj: T
+    description: str = ""
+    schema: Any | None = None
+
+
+class Registry(Generic[T]):
+    """A single named capability namespace (one of the eight ``kind``\\ s).
+
+    Use as a decorator (``@SPATIAL_FUNCS.register("max")``) or a direct call
+    (``SPATIAL_FUNCS.register("max", my_func)``). ``get`` / ``list`` /
+    ``describe`` trigger lazy entry-point discovery on first use.
+    """
+
+    def __init__(self, kind: str):
+        self.kind = kind
+        self._entries: dict[str, Entry[T]] = {}
+
+    def register(
+        self,
+        name: str,
+        obj: T | None = None,
+        *,
+        description: str = "",
+        schema: Any | None = None,
+        replace: bool = False,
+    ):
+        """Register ``obj`` under ``name`` (or return a decorator if omitted)."""
+        if obj is None:
+            def _decorate(target: T) -> T:
+                self._set(name, target, description, schema, replace)
+                return target
+
+            return _decorate
+        self._set(name, obj, description, schema, replace)
+        return obj
+
+    def _set(self, name: str, obj: T, description: str, schema: Any | None, replace: bool) -> None:
+        if not isinstance(name, str) or not name:
+            raise ValueError(f"{self.kind} name must be a non-empty string, got {name!r}")
+        if not replace and name in self._entries:
+            raise ValueError(
+                f"{self.kind} {name!r} is already registered; pass replace=True to override"
+            )
+        self._entries[name] = Entry(name, obj, description, schema)
+
+    def get(self, name: str) -> T:
+        _ensure_discovered()
+        try:
+            return self._entries[name].obj
+        except KeyError:
+            raise UnknownCapability(name, self.kind, sorted(self._entries)) from None
+
+    def list(self) -> list[str]:
+        _ensure_discovered()
+        return sorted(self._entries)
+
+    def describe(self, name: str) -> dict[str, Any]:
+        """Return ``{name, kind, description, schema?}`` for one entry."""
+        _ensure_discovered()
+        try:
+            entry = self._entries[name]
+        except KeyError:
+            raise UnknownCapability(name, self.kind, sorted(self._entries)) from None
+        return _entry_dict(entry, self.kind)
+
+    def describe_all(self) -> list[dict[str, Any]]:
+        """Return one ``describe`` dict per entry, name-sorted."""
+        _ensure_discovered()
+        return [_entry_dict(self._entries[name], self.kind) for name in sorted(self._entries)]
+
+    def __contains__(self, name: object) -> bool:
+        _ensure_discovered()
+        return name in self._entries
+
+
+def _entry_dict(entry: Entry, kind: str) -> dict[str, Any]:
+    out: dict[str, Any] = {"name": entry.name, "kind": kind, "description": entry.description}
+    if entry.schema is not None:
+        out["schema"] = entry.schema
+    return out
+
+
+# ---------------------------------------------------------------------------
+# The eight registries
+# ---------------------------------------------------------------------------
+
+SPATIAL_FUNCS: Registry[Callable] = Registry("spatial_func")
+REDUCERS: Registry[Any] = Registry("reducer")
+MASK_PROVIDERS: Registry[Any] = Registry("mask_provider")
+FIELD_TRANSFORMS: Registry[Callable] = Registry("field_transform")
+EVENT_TRIGGERS: Registry[Callable] = Registry("event_trigger")
+READERS: Registry[Any] = Registry("reader")
+CATALOG_SOURCES: Registry[Any] = Registry("catalog_source")
+CREDENTIAL_PROVIDERS: Registry[Any] = Registry("credential_provider")
+
+_REGISTRIES: dict[str, Registry] = {
+    reg.kind: reg
+    for reg in (
+        SPATIAL_FUNCS,
+        REDUCERS,
+        MASK_PROVIDERS,
+        FIELD_TRANSFORMS,
+        EVENT_TRIGGERS,
+        READERS,
+        CATALOG_SOURCES,
+        CREDENTIAL_PROVIDERS,
+    )
+}
+
+
+# ---------------------------------------------------------------------------
+# Lazy entry-point discovery
+# ---------------------------------------------------------------------------
+
+# ``_DISCOVERED`` flips True after a successful entry-point sweep so repeated
+# ``get`` / ``list`` calls don't re-iterate ``entry_points`` every time.
+# ``_DISCOVERING`` is a re-entrancy guard: if a plugin's ``register()`` reaches
+# a ``get`` during loading, we return early without re-running the sweep.
+_DISCOVERED = False
+_DISCOVERING = False
+
+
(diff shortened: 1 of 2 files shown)


src/zagg/registry.py
```diff
@@ -0,0 +1,450 @@
...
+    def describe(self, name: str) -> dict[str, Any]:
+        """Return ``{name, kind, description, schema?}`` for one entry."""
+        _ensure_discovered()
+        try:
+            entry = self._entries[name]
+        except KeyError:
+            raise UnknownCapability(name, self.kind, sorted(self._entries)) from None
+        return _entry_dict(entry, self.kind)
+
+    def describe_all(self) -> list[dict[str, Any]]:
```
espg, line 157: 🤖 *from Claude (review)*

**Blocking (mypy / `pre-commit`):** the method `def list(self)` (line 138) shadows the builtin `list` inside the class body, so subsequent `list[...]` *annotations* resolve to `Registry.list` rather than the builtin. With `from __future__ import annotations` mypy resolves the string annotations in class scope and errors here:

```
src/zagg/registry.py:151: error: Function "zagg.registry.Registry.list" is not valid as a type  [valid-type]
```

I reproduced it with `uv run mypy src/zagg/registry.py` (mypy 1.19.1, the pinned pre-commit version). CI's `test.yml`/`lint.yml` don't run mypy, so the GitHub checks stay green — but CLAUDE.md §4/§7 require `pre-commit run --all-files` (ruff + **mypy** + codespell) to be green locally, and this fails it. Fixes: alias the builtin (`_list = list` at module top and annotate with `_list[str]`), or rename the method (`names()`), or wrap return types in quotes that mypy resolves outside class scope. Flagging rather than dictating since the public `.list()` name is part of the agreed surface.

---

src/zagg/registry.py
```diff
@@ -0,0 +1,450 @@
...
+
+def get_catalog_source(name: str) -> Any:
+    return CATALOG_SOURCES.get(name)
+
+
+def list_catalog_sources() -> list[str]:
+    return CATALOG_SOURCES.list()
+
+
+def register_credential_provider(name, provider=None, *, description="", schema=None, replace=False):
```
espg: 🤖 *from Claude (review)*

**Blocking (`ruff format --check`):** this file is not `ruff format`-clean. `uv run ruff format --check src/zagg/registry.py` reports it would reformat, with three changes: a blank line after `if obj is None:` (line 113), collapsing the `logger.exception(...)` call (lines 226–228) onto one line, and wrapping this `register_credential_provider` signature across lines. Per CLAUDE.md §4, `pre-commit run --all-files` (which includes `ruff-format`) must be green. Run `uv run ruff format src/zagg/registry.py`. (The ruff *check* bot at E,F,W,I passes — it's only the formatter that flags this.)

---

tests/test_registry.py
```diff
@@ -0,0 +1,297 @@
...
+        assert list_spatial_funcs() == []
+        assert registry._DISCOVERED is True
+        assert calls["n"] == 2
+
+    def test_register_reentrancy_during_discovery(self, monkeypatch):
+        # A plugin whose register() itself calls get_* must not re-enter the sweep.
+        def plugin_register():
+            register_field_transform("a", object())
+            # Touch a get during discovery — should short-circuit, not recurse.
+            assert "a" in registry.FIELD_TRANSFORMS._entries
```
espg: 🤖 *from Claude (review)*

**Test fidelity gap.** This test is named/commented as exercising the re-entrancy guard ("Touch a get during discovery — should short-circuit, not recurse"), but the plugin body reaches into the raw dict `registry.FIELD_TRANSFORMS._entries`, which does **not** call `_ensure_discovered()`. So `_DISCOVERING` is never re-entered and the guard at `registry.py:219` is not actually covered — the test would still pass if the guard were deleted. To genuinely exercise it, call a path that re-enters discovery from inside the plugin's `register()`, e.g. `get_field_transform("a")` or `"a" in registry.FIELD_TRANSFORMS` (the `__contains__` path), and assert it returns rather than recursing/raising. Non-blocking, but it's the one safety property here whose test doesn't bite.

---

src/zagg/registry.py
```diff
@@ -0,0 +1,450 @@
...
+
+    Each entry point must resolve to a zero-arg callable that registers its
+    capabilities. A failure in one plugin's ``ep.load()`` / ``register()`` is
+    logged at ERROR and skipped — the rest still load. If
+    ``entry_points()`` itself raises, ``_DISCOVERED`` stays False so the next
+    ``get`` / ``list`` retries (transient backport failures don't degrade the
+    seam permanently).
+    """
+    global _DISCOVERED, _DISCOVERING
+    if _DISCOVERED or _DISCOVERING:
```
espg: 🤖 *from Claude (review)*

**Thread-safety (worth a one-line note, likely non-blocking).** `_DISCOVERED`/`_DISCOVERING` are plain module globals mutated without a lock. The re-entrancy guard correctly handles *same-thread* recursion, but if two threads hit a first-time `get()`/`list()` concurrently (zagg runs under a `ThreadPoolExecutor` in places), both can observe `_DISCOVERED==False and _DISCOVERING==False`, both set `_DISCOVERING=True`, and both run the sweep — invoking each plugin's `register()` twice and tripping the duplicate-name `ValueError` inside the loop (caught + logged, so plugins half-register silently). A single module-level `threading.Lock` around the sweep (stdlib, no new dep) would make it once-per-interpreter under concurrency too. If the contract is "discovery always happens at single-threaded import via `discover_plugins()`", a one-line docstring note saying lazy discovery is not thread-safe and the Lambda path must pre-warm would suffice. Either way, flag the intended contract.

---

src/zagg/registry.py
```diff
@@ -0,0 +1,450 @@
...
+        return _entry_dict(entry, self.kind)
+
+    def describe_all(self) -> list[dict[str, Any]]:
+        """Return one ``describe`` dict per entry, name-sorted."""
+        _ensure_discovered()
+        return [_entry_dict(self._entries[name], self.kind) for name in sorted(self._entries)]
+
+    def __contains__(self, name: object) -> bool:
+        _ensure_discovered()
+        return name in self._entries
```
espg, line 164: 🤖 *from Claude (review)*

**Nit (coverage gap, not a bug).** `__contains__` triggers `_ensure_discovered()` — a deliberate and reasonable choice (membership is a lookup, so it should see plugin entries) — but no test asserts that `x in REGISTRY` actually drives discovery. `test_register_direct_then_get` exercises `in` only *after* a direct registration. Consider one assertion that `"from_plugin" in registry.SPATIAL_FUNCS` returns True pre-`get` via the entry-point path, mirroring `test_entry_point_register_runs_on_first_get`. This would also be the natural place to land the re-entrancy fix from the other comment.

---

src/zagg/registry.py
```diff
@@ -0,0 +1,450 @@
...
+                continue
+            try:
+                register_fn()
+            except Exception:
+                logger.exception("zagg.plugins entry point %r raised on register()", ep.name)
+    finally:
+        _DISCOVERING = False
+
+
+def discover_plugins(*, force: bool = False) -> None:
```
espg, line 263: 🤖 *from Claude (review)*

**Sharp edge in `force=True` (doc'd, but a footgun).** `force=True` only resets `_DISCOVERED`, not the entries already registered, so a real re-sweep re-runs every plugin `register()` against a non-empty registry and trips the duplicate-name guard unless every plugin passes `replace=True`. The test `test_discover_plugins_force_reruns` only passes because its fake plugin uses `replace=True`. A built-in plugin that doesn't would raise on the forced sweep (caught + logged inside the loop, so the re-discovery silently no-ops for that plugin). Since `force` is documented as "intended for tests," consider either making it test-only/private or having it also clear `_entries` so a forced re-sweep starts clean and is deterministic. Non-blocking; flag the intent.

---

src/zagg/registry.py
```diff
@@ -0,0 +1,450 @@
...
+
+
+@dataclass(frozen=True)
+class Entry(Generic[T]):
+    """One registered capability: its name, object, and optional metadata."""
+
+    name: str
+    obj: T
+    description: str = ""
+    schema: Any | None = None
```
espg, line 89: 🤖 *from Claude (review)*

**Minor (mutability of `schema`).** `Entry` is `frozen=True` (good — the dataclass can't be rebound), but `schema` holds an arbitrary object and `describe()`/`describe_all()` return it by reference (`_entry_dict` does `out["schema"] = entry.schema`, and `test_describe_with_schema_includes_it` asserts `is`). If a registrant passes a mutable dict schema, an MCP consumer can mutate the registry's stored schema in place. Probably fine for now (schemas are expected to be static pydantic models / literals), but worth a one-line note in the `schema` docstring that it's stored and returned by reference and should be treated as immutable. Not blocking.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the #64 registry seam. The design matches the locked spec (eight kinds, `Registry[T]` core, optional `description`/`schema`, `UnknownCapability(name, kind, available=[...])`, lazy `zagg.plugins` discovery with re-entrancy guard + retry, `describe_all()`, per-kind wrappers, strings-not-callables). Tests are well-structured and the runtime behavior is correct. Local verification:

- `pytest tests/test_registry.py -q` → **49 passed** ✅
- `ruff check --select=E,F,W,I --ignore=E501` → **clean** ✅ (matches the lint bot)
- `ruff format --check` → **fails** (would reformat `registry.py`) ❌
- `mypy src/zagg/registry.py` → **1 error in this file** ❌ (`list[...]` annotations shadowed by the `list()` method)

CI (`test.yml`/`lint.yml`) runs only pytest + the ruff E,F,W,I bot, so GitHub checks will go green — but CLAUDE.md §4/§7 require `pre-commit run --all-files` (ruff-format + mypy + codespell) to pass locally, and two of those fail. Flagging as blocking-for-"done" rather than blocking-CI.

**Blocking for local-green (§4):**
1. `registry.py:151` — mypy `[valid-type]`: the `list()` method shadows builtin `list` for the in-class `list[...]` annotations. (`describe_all`/`list` return annotations.)
2. `registry.py` — not `ruff format`-clean (3 spots; `register_credential_provider` signature, the `logger.exception` call, blank line after `if obj is None:`).

**Non-blocking findings:**
3. `test_register_reentrancy_during_discovery` doesn't actually re-enter discovery — it touches `_entries` directly instead of a `get`/`__contains__` path, so the `_DISCOVERING` guard is uncovered (test would pass with the guard removed).
4. Lazy discovery isn't thread-safe — concurrent first-time `get()` under a `ThreadPoolExecutor` can double-run the sweep; either a stdlib `threading.Lock` or an explicit "pre-warm via `discover_plugins()`; lazy path isn't thread-safe" contract note.
5. `__contains__`-triggers-discovery is intentional but untested via the plugin path.
6. `discover_plugins(force=True)` resets only `_DISCOVERED`, not `_entries`, so a real forced re-sweep trips the dup guard unless every plugin uses `replace=True` (the test only passes because its fake does).
7. `schema` is stored/returned by reference (mutable-dict aliasing) — fine for now, worth a one-line docstring caveat.

On the PR's own "Questions for review": the `Any` looseness on `schema` reads reasonable as a placeholder until #59 pins the encoding; the `list_spatial_funcs` naming is an @espg call, not a correctness issue. No dependency or scope concerns — module is stdlib-only (~450 lines, well under the ~1000 limit), terse, and convention-matching. Items 1–2 are the only must-fix-before-done.

---

espg:
🤖 *from Claude*

Two things — folding the self-review, and the red morton tests.

### Self-review findings — addressed in `a3b0ef1`

- **mypy error (`registry.py`, `list()` shadowing the builtin in annotations)** → renamed the method `list()` → `names()`; the public `list_*` functional wrappers are unchanged. `mypy src/zagg/registry.py` is clean now.
- **`ruff format`** → ran it on both new files; `ruff format --check` passes.
- **Thread-safety of the discovery flags** → added a module-level `threading.RLock` + double-checked locking in `_ensure_discovered` (the seam runs under the Lambda `ThreadPoolExecutor`, so two first-time `get()`s could double-run the sweep). `RLock` keeps the same-thread plugin re-entrancy deadlock-free; the `_DISCOVERED` fast-path still short-circuits a plugin's `register()` calling back through `get()`.
- **Test fidelity (re-entrancy test)** → `test_register_reentrancy_during_discovery` now drives the guard through real `__contains__` + `names()` paths (both call `_ensure_discovered`) instead of poking `_entries` directly.
- **`schema` stored by reference** → added a one-line caveat in `register()`'s docstring (stored/returned as-is; pass an owned/immutable object).
- Standing (non-blocking): `discover_plugins(force=True)` resets only the discovered flag — its docstring already documents that a forced re-sweep needs `replace=True` per plugin; that's the intended test-hook contract, left as-is.

Green after the fold: `ruff check` ✓, `ruff format --check` ✓, `mypy registry.py` ✓, **49 tests** ✓.

### The red morton tests are pre-existing, not from this PR

- This PR's diff is **`registry.py` + `test_registry.py` only** (`git diff --name-only origin/main...` confirms) — it never touches the morton/HEALPix write path.
- `tests/test_processing.py::TestWriteDataframeToZarr::test_write_dataframe_to_zarr` (+ the two integration cases) fail identically on a clean `main`. And it's **not** a morton-vs-nested scheme mismatch — it's the *same morton values* read signed vs unsigned: `ACTUAL[-5734771175502905336]` as `uint64` is exactly `DESIRED[12711972898206646280]`, and the reported max abs difference is exactly `2^64`. So it's the `morton: int64` sign-bit hazard (negative for HEALPix base cells 7–11) — the one #71 fixes by moving the coordinate to `MortonIndexDtype`.
- Per CLAUDE.md §4, flagging not fixing — out of scope for the registry, resolved under #71.

So #73's own surface is green; the red X is the standing morton sign issue tracked by #71.

---

Issue #57, opened 2026-06-18
Integration with pending version 1 of mortie


espg:
Major redesign of mortie will be bringing in a new `morton_index` datatype (see https://github.com/espg/mortie/issues/35 , merged pr's https://github.com/espg/mortie/pull/51 and https://github.com/espg/mortie/pull/43 ). What do we need to do to integrate with this new data type on our end? This will be replacing how we index the healpix / morton path today. Most concretely, we won't be encoding uint64 anymore.

Is there more that needs to be posted to https://github.com/espg/mortie/issues/48 , i.e., additional changes prior to the version 1.0 release that would benefit us as downstream library user?

espg:
🤖 *from Claude*

**What zagg needs for mortie v1 integration:**

The `morton_index` type (mortie PRs #43/#51, issue #35) replaces raw uint64 encoding. Today zagg uses `np.int64` shard keys throughout — `catalog/shardmap.py`, `grids/healpix.py`, zarr coordinate arrays. The places where uint64 keys cross from mortie into zagg are the main surface area for this migration.

**Key questions to answer before scoping the implementation:**

**(1) Does `morton_index` support a numpy dtype interface?** If it implements `__array__()` → int64 or `__int__()`, existing zarr storage and numpy index arithmetic continue to work with minimal changes. If it's an opaque Python object, every place zagg does leaf-id arithmetic (`cells_of`, `shards_of`, `children`, ShardMap serialization) needs updating.

**(2) Which zagg call sites are affected?** The list: `HealpixGrid.cells_of`/`shards_of`/`children`, the mortie `healpix_shard()` call site in `grids/healpix.py`, `ShardMap.shard_keys` (JSON serialization), and the zarr coordinate arrays. These are the concrete places where uint64 → `morton_index` shows up.

**(3) Backward compatibility shim?** If mortie v1 provides `int(key)` → uint64 conversion, the migration is contained to the call sites. If not, it's more invasive.

**On the upstream #48 question — useful things to raise before v1 ships:**
- Confirm `morton_index` supports `__int__`/`__array__` for downstream numpy compatibility (or document that it doesn't and show the migration path).
- Provide a changelog or migration guide for downstream library users.
- Clarify whether `SpatialIndex` (#54) and `linestring` coverage (#53) are updated for `morton_index` in v1 or a later release — this affects zagg's integration sequencing.

**Two implementation approaches:**

[...]

PR #75, opened 2026-06-22
Adopt mortie MortonIndexDtype for the `morton` coordinate (keep NESTED `cell_ids`)

### LINKED ISSUE #71: Adopt mortie v1 MortonIndexDtype for the `morton` coordinate (keep NESTED `cell_ids`)
Refs #57 (the mortie-v1 integration umbrella). This is the **near-term, contained slice** locked in https://github.com/englacial/zagg/issues/57#issuecomment-4771871445: migrate the `morton` coordinate to mortie v1's `MortonIndexDtype`, **keeping `cell_ids` as NESTED `uint64`** (the DGGS/xdggs coordinate, unchanged). The future "morton_index *becomes* the DGGS coordinate" switch is tracked separately (gated on xdggs — see the companion issue).

## Why

Today zagg stores two HEALPix coordinate arrays (confirmed against the write path + `configs/atl06.yaml`, and the `atl06_cycle22_fullsphere.zarr` benchmark):

- `cell_ids` — `uint64`, HEALPix **NESTED** via `mort2healpix`; the DGGS-convention coordinate (`indexing_scheme: "nested"`, `coordinate: "cell_ids"`). **Stays as-is.**
- `morton` — `int64`, raw mortie packed word. Its prefix is `base+1`, so the value is **negative for HEALPix base cells 7–11** (~5/12 of the sphere). This sign-bit packing is the root of the pre-existing `morton int64/uint64` test failures on `main` (`test_processing.TestWriteDataframeToZarr` + two integration cases, as flagged by #70's suite run).

mortie v1's `MortonIndexDtype` carries order + base internally instead of relying on a reinterpreted sign bit, removing the hazard — and unlocks the higher orders (ATL03-level ~10 m resolution) where the int64 packing is most fragile.

## Scope (Option B — propagate the type through public surfaces)

mortie v1 (`mortie/__init__.py`, on `main`) ships `MortonIndexDtype`/`MortonIndexArray` (pandas ext), `MortonIndexType`/`MortonIndexExtArray` (pyarrow ext), `to_morton_index`/`from_morton_index`, and the existing `geo2mort`/`clip2order`/`generate_morton_children`/`mort2healpix`.

**On-disk form (confirmed required until the xdggs switch):** store the `morton` coordinate as `uint64` via `to_morton_index`/`from_morton_index`, reconstruct `MortonIndexArray` on read. Existing stores stay readable where the integer is numerically equal.

## Phases

1. **Boundary adapter + dep pin.** Pin `mortie>=1.0` and add a thin adapter at the mortie seam; internals unaffected. (Logistics: confirm mortie 1.0 is installable for CI — PyPI release/tag — else pin a git ref; see open items.)
2. **Storage round-trip.** zarr `morton` coord encode/decode + `ShardMap` JSON (de)serialization through `to_morton_index`/`from_morton_index`. `cell_ids` NESTED path untouched.
3. **Propagate the type.** `MortonIndexArray` across `HealpixGrid.cells_of`/`shards_of`/`children`/`shard_of`/`encode_cell_ids` and `ShardMap.shard_keys`, plus `processing.py` `cell_to_slice` dict keys.
4. **Tests.** Round-trip (`geo2mort → clip2order → coord → reload`), `ShardMap` JSON parity, dtype assertions, and that the three `morton` sign-bug tests pass.

## Files likely touched

`src/zagg/grids/healpix.py`, `src/zagg/grids/base.py`, `src/zagg/catalog/shardmap.py`, `src/zagg/processing.py`, `src/zagg/configs/atl06.yaml` (+ `atl03*.yaml`), `src/zagg/__init__.py`.

## Acceptance criteria

- `morton` coordinate is `MortonIndexDtype`-backed, stored as `uint64` on disk, round-tripping losslessly.
- `cell_ids` remains NESTED `uint64` — DGGS convention attrs unchanged.
- The three pre-existing `morton int64/uint64` sign-bug tests pass.
- `ShardMap` JSON round-trips; ATL06 outputs numerically identical where ids match.

## Open logistics / pre-reqs

1. **mortie 1.0 availability for CI** — needs a pip-installable release/tag, else the PR pins a git ref (flag for sign-off).
2. **`lambda` extra mortie pin** moves in lockstep (baked into the deployment layer).
3. **`MortonIndexArray` element hashability** for `cell_to_slice` dict keys — verify early in implementation (cheap to confirm).
4. **Sequencing with #54/#53** (SpatialIndex / linestring coverage) if those also move to the new type in the same mortie release.

_Authored by Claude under the #57 routine._

espg:
Closes #71. Refs #57.

## What

Migrates the `morton` output coordinate to mortie's typed `morton_index` (a `MortonIndexArray` over packed `uint64` words), stored as `uint64` on disk and reconstructed as a `MortonIndexArray` on read. `cell_ids` stays NESTED `uint64` (the DGGS coordinate, byte-identical, unchanged) — exactly the contained slice locked in https://github.com/englacial/zagg/issues/57#issuecomment-4771871445.

This removes the int64 sign hazard that caused the three pre-existing `morton int64/uint64` failures on `main`. The packed Morton word's prefix is `base+1`, so base cells 7–11 set bit 63; pinned as the old `morton: int64` coordinate they read back **negative** (~5/12 of the sphere). Stored as `uint64` they stay non-negative with the Z-order intact.

## Approach

- **Boundary adapter** `src/zagg/grids/morton.py`: `morton_words()` (MortonIndexArray → packed `uint64` for storage/wire), `to_morton_array()` (the inverse, reconstruct on read), `is_morton_array()`. Zarr stores numpy dtypes, not pandas extension dtypes, so this is the seam where the type meets disk. (mortie's `to_morton_index`/`from_morton_index` are pyarrow/IPC bridges; the zarr coordinate is a plain numpy buffer, so the adapter goes through the array's `uint64` words directly — same bytes, no pyarrow dependency on the write path.)
- **`HealpixGrid.chunk_coords`** now returns `morton` as a `MortonIndexArray`; `cell_ids` untouched.
- **Storage boundary** in `processing._iter_carrier_columns` (pandas carrier) **and** `_build_output` (arrow carrier): a `MortonIndexArray` morton column is written as its `uint64` words via `morton_words`, so both carriers share one on-disk dtype guarantee.
- **Type scope (contained):** the internal leaf/cell/shard morton **arithmetic** (`cells_of`/`shards_of`/`children`, and the `cell_to_slice` dict keys in `processing.py`) stays on plain `uint64` ndarrays. Those are grouping/coarsening math (`np.argsort`/`np.diff`/`==`), not the stored coordinate; wrapping them would change grouping semantics with no benefit to this slice. The typed array lives at the **coordinate/storage** surface, which is what #71 specifies ("Make the `morton` coordinate a `MortonIndexDtype`-backed array").
- **Config:** `configs/atl06.yaml` `morton: int64 → uint64`. `atl06_polar.yaml` is rectilinear (no morton coord).
- **Dep pin:** `mortie>=0.8.0 → >=0.8.1` (0.8.x is the 1.0 RC with the core in place, per https://github.com/englacial/zagg/issues/57#issuecomment-4771968670).

## Hashability / ordering probe (the two items flagged verify-during-impl)

Both confirmed present against the installed mortie 0.8.1 (`mortie/morton_index.py`):

1. **Element-wise hashable (dict keys)** — **YES.** Scalar access `arr[i]` returns a plain `np.uint64` (`__getitem__` boxes scalars as `np.uint64(result)`), which is hashable; `{arr[i]: i ...}` and `arr[j] in d` both work. *(Note: zagg's `cell_to_slice` keys come from `cells_of` `uint64` ndarrays, not the extension array, so this isn't even on the hot path — but it's confirmed for any future use.)*
2. **Ordering (hyperslice min/max)** — **YES.** `MortonIndexArray` defines element-wise `__lt__/__le__/__gt__/__ge__` over the raw `uint64` (the Z-order; unsigned, so bit-63 cells sort correctly with no special-casing) and `_values_for_argsort()` returns the `uint64` words, so `np.argsort`/`np.unique`/`min`/`max` all work. `np.asarray(arr)` yields a plain `uint64` ndarray.

Nothing missing upstream — no blocker, no issue-#71 comment needed.

## Phases

- [x] **Phase 1** — Boundary adapter (`grids/morton.py`), `mortie>=0.8.1` pin, `morton: uint64` config.
- [x] **Phase 2** — `morton` coordinate as `MortonIndexArray` through `chunk_coords` + `uint64` storage round-trip at the write boundary (both carriers); tests. The three sign-bug tests pass.
- [x] **Review fold** — Folded the fresh-context self-review's four diff-scoped findings: route the arrow carrier morton column through `morton_words`; replace the vacuous `(stored >= 0).all()` with `stored.view(np.int64).min() < 0` (proves the sign bug is actually fixed); add the missing non-negativity assertion to the geo2mort→clip2order round-trip test; gate `morton_words` on the public `is_morton_array` instead of the private `_data`.

## How tested

- `uv run pytest -q`: **527 passed, 1 skipped** (was 519 passed / **3 failed** on `main`). The three previously-failing sign-bug tests now pass: `test_processing.TestWriteDataframeToZarr::test_write_dataframe_to_zarr`, `test_integration::test_full_integration`, `test_integration::test_multiple_parent_cells`.
- New tests (`tests/test_grids.py::TestMortonCoordinate`): `chunk_coords` yields a `MortonIndexArray`; words equal `generate_morton_children`; `geo2mort → clip2order → coord → reload` round-trip (asserts at least one bit-63 word); full zarr write/read round-trip asserting `uint64` dtype, lossless words, and that `stored.view(int64).min() < 0` (the bit that would have read negative under the old coordinate).
- `tests/test_shardmap.py::TestIO::test_high_base_cell_morton_keys_roundtrip`: parent-morton shard keys from base-7–11 cells (words > 2^63) survive JSON (de)serialization byte-for-byte.
- `tests/test_schema.py::test_coordinate_dtypes` updated to assert `morton` is now `uint64`.
- `ruff check --select=E,F,W,I --ignore=E501 src tests`, pinned `pre-commit` `ruff-format`/`ruff-check`/`codespell` — clean on touched files. CI `test (3.12)`/`test (3.13)`/`ruff`/`build`/`build-x86_64`/`build-arm64` green.

## Questions for review

1. **`lambda` extra mortie pin.** The locked design said "move the `lambda` extra's mortie pin in lockstep," but the `lambda` extra in `pyproject.toml` doesn't pin `mortie` (the core `mortie>=0.8.1` dep governs it), and the deployment layer build (`deployment/aws/build_layer.sh`) installs `mortie` unpinned. So there is no separate lambda pin to move — the core bump covers it. I left the deploy scripts untouched per the infra rule. Flag if you'd rather pin mortie explicitly in the `lambda` extra and/or the layer script (a deploy-script change I won't make without you naming it).
2. **Pre-existing `mypy` errors** in untouched files (`config.py`, `rectilinear.py`, `grids/__init__.py:51`, and pre-existing `processing.py` lines 422/429/459/1033/1353/1418/1481) are unrelated to this change; left as-is per §4. None are in the code this PR touches.
3. **Out-of-scope review observation (left standing).** The self-review noted that the existing `use_arrow=True` `_build_output` test in `tests/test_processing.py` exercises the arrow morton write but asserts only on `edges`/`count`, not the morton column's `uint64` round-trip. The arrow-path fix in this PR plus the new `test_morton_stored_uint64_roundtrips_through_zarr` cover the round-trip; adding a morton assertion to that pre-existing arrow test would be a belt-and-suspenders guard against a future `__array__` dtype regression. Left for you rather than expand scope into an unrelated test — say the word and I'll add it.

### DIFF
--- a/src/zagg/grids/morton.py
+++ b/src/zagg/grids/morton.py
@@ -1,0 +1,59 @@
+"""Boundary adapter for mortie's ``morton_index`` extension type.
+
+The ``morton`` output coordinate is carried in memory as a mortie
+:class:`~mortie.morton_index.MortonIndexArray` (a pandas ExtensionArray over the
+packed ``uint64`` Morton words). On disk it is stored as plain ``uint64`` — Zarr
+stores numpy dtypes, not pandas extension dtypes — and reconstructed as a
+``MortonIndexArray`` on read.
+
+This is the contained #71 migration: only the ``morton`` coordinate adopts the
+type. ``cell_ids`` stays NESTED ``uint64`` (the DGGS coordinate, unchanged), and
+the internal leaf/cell/shard morton arithmetic (``cells_of`` / ``shards_of`` /
+``children``) stays on plain ``uint64`` ndarrays.
+
+Storing the raw ``uint64`` words (rather than a reinterpreted ``int64``) is what
+removes the sign hazard: the packed word's prefix is ``base+1``, so base cells
+7–11 set bit 63 and read back negative under an ``int64`` coordinate. ``uint64``
+keeps them non-negative and the Z-order intact (espg/zagg#71).
+"""
+
+from __future__ import annotations
+
+import numpy as np
+
+
+def morton_words(values) -> np.ndarray:
+    """Return the packed ``uint64`` Morton words for ``values``.
+
+    Accepts a :class:`~mortie.morton_index.MortonIndexArray` (its ``uint64``
+    storage is returned) or any ``uint64``-coercible array-like (returned as a
+    ``uint64`` ndarray). This is the on-disk / wire form of the ``morton``
+    coordinate.
+    """
+    data = getattr(values, "_data", None)
+    if data is not None:
+        return np.asarray(data, dtype=np.uint64)
+    return np.asarray(values, dtype=np.uint64)
+
+
+def to_morton_array(words):
+    """Reconstruct a ``MortonIndexArray`` from packed ``uint64`` words.
+
+    The inverse of :func:`morton_words` for the storage round-trip: read the
+    ``uint64`` coordinate back from Zarr and wrap it as the extension array.
+    """
+    from mortie import MortonIndexArray
+
+    return MortonIndexArray.from_words(np.asarray(words, dtype=np.uint64))
+
+
+def is_morton_array(values) -> bool:
+    """True if ``values`` is a mortie ``MortonIndexArray``."""
+    try:
+        from mortie import MortonIndexArray
+    except ImportError:  # pragma: no cover - mortie is a hard dependency
+        return False
+    return isinstance(values, MortonIndexArray)
+
+
+__all__ = ["morton_words", "to_morton_array", "is_morton_array"]
--- a/src/zagg/grids/morton.py
+++ b/src/zagg/grids/morton.py
@@ -22,6 +22,15 @@ from __future__ import annotations
 import numpy as np
 
 
+def is_morton_array(values) -> bool:
+    """True if ``values`` is a mortie ``MortonIndexArray``."""
+    try:
+        from mortie import MortonIndexArray
+    except ImportError:  # pragma: no cover - mortie is a hard dependency
+        return False
+    return isinstance(values, MortonIndexArray)
+
+
 def morton_words(values) -> np.ndarray:
     """Return the packed ``uint64`` Morton words for ``values``.
 
@@ -30,9 +39,8 @@ def morton_words(values) -> np.ndarray:
     ``uint64`` ndarray). This is the on-disk / wire form of the ``morton``
     coordinate.
     """
-    data = getattr(values, "_data", None)
-    if data is not None:
-        return np.asarray(data, dtype=np.uint64)
+    if is_morton_array(values):
+        return np.asarray(values._data, dtype=np.uint64)
     return np.asarray(values, dtype=np.uint64)
 
 
@@ -47,13 +55,4 @@ def to_morton_array(words):
     return MortonIndexArray.from_words(np.asarray(words, dtype=np.uint64))
 
 
-def is_morton_array(values) -> bool:
-    """True if ``values`` is a mortie ``MortonIndexArray``."""
-    try:
-        from mortie import MortonIndexArray
-    except ImportError:  # pragma: no cover - mortie is a hard dependency
-        return False
-    return isinstance(values, MortonIndexArray)
-
-
-__all__ = ["morton_words", "to_morton_array", "is_morton_array"]
+__all__ = ["is_morton_array", "morton_words", "to_morton_array"]
--- a/src/zagg/processing.py
+++ b/src/zagg/processing.py
@@ -25,6 +25,7 @@ from zagg.config import (
     get_data_vars,
     get_output_signature,
 )
+from zagg.grids.morton import is_morton_array, morton_words
 from zagg.read_plan import execute_read_plan, plan_read
 from zagg.schema import ProcessingMetadata
 
@@ -323,7 +324,13 @@ def _iter_carrier_columns(carrier):
     """
     if isinstance(carrier, pd.DataFrame):
         for name, series in carrier.items():
-            yield name, series.values
+            values = series.values
+            # The ``morton`` coordinate is carried as a mortie ``MortonIndexArray``
+            # (#71); Zarr stores numpy dtypes, so extract its packed ``uint64``
+            # words for the on-disk write (no .reshape on the extension array).
+            if is_morton_array(values):
+                values = morton_words(values)
+            yield name, values
         return
 
     import pyarrow as pa
@@ -876,9 +883,7 @@ def _planned_read_group(
     si_lvl = levels[spatial_index_level]
     link = si_lvl.get("link")
     if not isinstance(link, dict):
-        raise ValueError(
-            f"read_plan.spatial_index level {spatial_index_level!r} requires a 'link'"
-        )
+        raise ValueError(f"read_plan.spatial_index level {spatial_index_level!r} requires a 'link'")
     if link["to"] != base_level_key:
         raise ValueError(
             f"read_plan.spatial_index level {spatial_index_level!r} must link "
@@ -962,9 +967,7 @@ def _planned_read_group(
     coarse_structured = [
         f
         for f in filters
-        if "expression" not in f
-        and f.get("level") is not None
-        and f.get("level") != base_level_key
+        if "expression" not in f and f.get("level") is not None and f.get("level") != base_level_key
     ]
     expressions = [f for f in filters if "expression" in f]
 
@@ -1008,9 +1011,7 @@ def _planned_read_group(
     if coarse_structured:
         # Build the global base-index array once: which original-base positions
         # are present in the concatenated planned read.
-        global_idx = np.concatenate(
-            [np.arange(s, e, dtype=np.int64) for s, e in plan.base_slices]
-        )
+        global_idx = np.concatenate([np.arange(s, e, dtype=np.int64) for s, e in plan.base_slices])
         cross_full: np.ndarray | None = None
         for f in coarse_structured:
             level_key = f["level"]
@@ -1043,7 +1044,9 @@ def _planned_read_group(
         if keep_mask is not None:
             values = values[keep_mask]
         data_dict[col_name] = values
-    data_dict["leaf_id"] = leaf_after_spatial[keep_mask] if keep_mask is not None else leaf_after_spatial
+    data_dict["leaf_id"] = (
+        leaf_after_spatial[keep_mask] if keep_mask is not None else leaf_after_spatial
+    )
 
     # Base-level expression filters (aggregation-time escape hatch, no pushdown).
     for f in expressions:
@@ -1108,13 +1111,10 @@ def _read_group(h5obj, group: str, data_source: dict, shard_key: int, grid, arro
     if isinstance(rp, dict) and rp.get("spatial_index"):
         if not isinstance(levels, dict) or not levels:
             raise ValueError(
-                "data_source.read_plan.spatial_index requires a non-empty "
-                "'levels' mapping"
+                "data_source.read_plan.spatial_index requires a non-empty 'levels' mapping"
             )
         if not base_level:
-            raise ValueError(
-                "data_source.read_plan.spatial_index requires 'base_level'"
-            )
+            raise ValueError("data_source.read_plan.spatial_index requires 'base_level'")
         return _planned_read_group(h5obj, group, data_source, shard_key, grid, arrow=arrow)
     return _read_group_full(h5obj, group, data_source, shard_key, grid, arrow=arrow)
 
--- a/src/zagg/processing.py
+++ b/src/zagg/processing.py
@@ -226,7 +226,11 @@ def _build_output(stats_arrays, data_vars, agg_fields, grid, shard_key, use_arro
         for var in data_vars
     }
     for col_name, vals in grid.chunk_coords(shard_key).items():
-        columns[col_name] = pa.array(np.asarray(vals))
+        # Route the morton coordinate through the same uint64 boundary as the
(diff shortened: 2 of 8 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the #71 morton→`MortonIndexArray`/uint64 migration. The core round-trip is sound for the shipped (pure-scalar) atl06 config: `_iter_carrier_columns` correctly converts the pandas-carrier `MortonIndexArray` via `morton_words` before the `.reshape`/zarr write, `fill_value: 0` is a safe sentinel (a packed word's `base+1` prefix is always non-zero, so 0 can never collide with a real morton word), and the `cell_to_slice` dict keys are still plain `int(cell_id)` from `cells_of`/`children` (no extension-array key leak). One real asymmetry in the Arrow carrier path and a couple of test-rigor issues below; none are blocking.

---

src/zagg/processing.py
```diff
@@ -323,7 +324,13 @@ def _iter_carrier_columns(carrier):
     """
     if isinstance(carrier, pd.DataFrame):
         for name, series in carrier.items():
-            yield name, series.values
+            values = series.values
+            # The ``morton`` coordinate is carried as a mortie ``MortonIndexArray``
+            # (#71); Zarr stores numpy dtypes, so extract its packed ``uint64``
+            # words for the on-disk write (no .reshape on the extension array).
+            if is_morton_array(values):
+                values = morton_words(values)
```
espg, line 336: 🤖 *from Claude (review)*

This makes the pandas carrier go through the explicit `morton_words` boundary — good. But the Arrow carrier does not: `_build_output` (line ~229, not in this diff) writes `morton` via `pa.array(np.asarray(vals))`, which never calls `morton_words`/`is_morton_array`. It happens to produce `uint64` only because `MortonIndexArray.__array__` returns `uint64` in mortie 0.8.1, but the whole point of `morton.py` is to be the *single* place pinning the on-disk dtype. With the pin open-ended (`mortie>=0.8.1`), a future `__array__` returning `int64`/`object` would silently corrupt the Arrow path while this pandas path stayed correct.

For the shipped pure-scalar atl06 config the Arrow branch is dead (only taken when a `vector` field exists on a HEALPix grid), so this is latent rather than a live bug. Suggest routing `_build_output`'s coord loop through the same boundary, e.g. `pa.array(morton_words(vals) if is_morton_array(vals) else np.asarray(vals))`, so both carriers share one dtype guarantee. Diff-scoped (the asymmetry is introduced by this PR even though the touched line lives in unchanged code).

---

tests/test_grids.py
```diff
@@ -381,3 +375,82 @@ def test_rectilinear_vector_array_has_trailing_dim(self):
...
+        write_dataframe_to_zarr(df, store, grid=g, chunk_idx=g.block_index(parent))
+
+        grp = open_group(store, path=str(child_order), mode="r")
+        assert grp["morton"].dtype == np.uint64
+        lo = int(np.asarray(coords["cell_ids"]).min())
+        hi = int(np.asarray(coords["cell_ids"]).max())
+        stored = grp["morton"][lo : hi + 1]
+        expected = morton_words(coords["morton"])
+        np.testing.assert_array_equal(stored, expected)
+        assert (stored >= 0).all()
```
espg: 🤖 *from Claude (review)*

`assert (stored >= 0).all()` is vacuous: `stored` is read back from a `uint64` zarr array, so `>= 0` is `True` for every possible value — it can never catch the sign hazard it claims to guard. (The real lossless check on the line above, `assert_array_equal(stored, expected)`, is the one doing the work and it *is* meaningful: the parent here is `_valid_parents(1)[0]` = the southern `(-78.5, -132.0)` point, whose order-8 children have bit 63 set.)

To actually assert the int64 sign bug is fixed, compare against the signed reinterpretation, e.g. assert `stored.view(np.int64).min() < 0` (proves at least one stored word would have read back negative under the old `int64` coordinate) and keep the `assert_array_equal` for losslessness. As written this line gives false confidence. Diff-scoped.

---

tests/test_grids.py
```diff
@@ -381,3 +375,82 @@ def test_rectilinear_vector_array_has_trailing_dim(self):
...
+        from mortie import clip2order, geo2mort
+
+        from zagg.grids.morton import morton_words, to_morton_array
+
+        # A southern point lands in a high base cell whose packed word sets bit 63.
+        leaf = geo2mort(np.array([-78.5]), np.array([-132.0]), order=18)
+        cells = clip2order(8, leaf)
+        coord = to_morton_array(cells)
+        words = morton_words(coord)
+        np.testing.assert_array_equal(words, np.asarray(cells, dtype=np.uint64))
```
espg, line 419: 🤖 *from Claude (review)*

The docstring promises this checks the word "is non-negative for southern (base 7-11) cells where int64 went negative," but the body never asserts that — it only checks `dtype == uint64` and equality with `np.asarray(cells, dtype=np.uint64)`. The chosen point does set bit 63 (verified: the word is `12712025674764779528`, i.e. `int64` = `-5734718398944772088`), so a one-line `assert (words.view(np.int64) < 0).any()` would make the test actually exercise the sign-bug premise instead of just asserting an unsigned re-cast equals itself. Non-blocking nit. Diff-scoped.

---

src/zagg/grids/morton.py
```diff
@@ -0,0 +1,59 @@
...
+
+def morton_words(values) -> np.ndarray:
+    """Return the packed ``uint64`` Morton words for ``values``.
+
+    Accepts a :class:`~mortie.morton_index.MortonIndexArray` (its ``uint64``
+    storage is returned) or any ``uint64``-coercible array-like (returned as a
+    ``uint64`` ndarray). This is the on-disk / wire form of the ``morton``
+    coordinate.
+    """
+    data = getattr(values, "_data", None)
```

Minor robustness note: `morton_words` reaches into the private `_data` attribute of `MortonIndexArray`. If a future mortie renames it, this silently falls through to `np.asarray(values, dtype=np.uint64)` — which still works for a real `MortonIndexArray` (via `__array__`), so the fallback is benign here. But since `is_morton_array` already exists, gating on it (`if is_morton_array(values): return np.asarray(values, dtype=np.uint64)`) would key the conversion on the public type check rather than an undocumented private attribute, and `np.asarray(MortonIndexArray, dtype=uint64)` returns the same words. Optional cleanup, not a correctness issue. Diff-scoped.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the #71 morton→`MortonIndexArray`/uint64 migration. Core round-trip is sound for the shipped pure-scalar atl06 config: `_iter_carrier_columns` converts the pandas-carrier `MortonIndexArray` via `morton_words` before the `.reshape`/zarr write; `fill_value: 0` is a safe sentinel (a packed word's `base+1` prefix is always non-zero, so 0 cannot collide with a real morton word); `cell_to_slice` keys stay plain `int(cell_id)` from `cells_of`/`children` (no extension-array key leak); `is_morton_array` is a clean `isinstance` so a plain `uint64` ndarray is never misclassified; and rectilinear `chunk_coords` returns `{}`, so the new type only ever flows from HEALPix. No `MortonIndexArray` reaches a `.reshape` or zarr write un-converted on the pandas path.

Findings posted inline (none blocking):
1. `processing.py` — Arrow carrier (`_build_output`) writes morton via `np.asarray`, bypassing the explicit `morton_words` boundary the pandas path uses; latent until a `vector` HEALPix config exists, but an asymmetry worth closing.
2. `test_grids.py:454` — `assert (stored >= 0).all()` is vacuous on a `uint64` array; doesn't actually verify the sign-bug fix.
3. `test_grids.py` `test_geo2mort_clip2order_coord_roundtrip` — docstring promises a non-negativity check the body never makes.
4. `morton.py:33` — `morton_words` keys on the private `_data` attr rather than the public `is_morton_array`.

One more coverage gap (couldn't anchor inline — the line isn't in this PR's diff): `tests/test_processing.py::...use_arrow=True` builds the Arrow carrier on a southern (bit-63) parent and so *does* exercise the un-guarded Arrow morton write from finding (1), but it only asserts on `edges`/`count` — the `morton` column's `uint64` round-trip is never checked. Adding a morton assertion there would catch any future `__array__` dtype regression on the Arrow path. Out-of-scope observation.

---

Issue #56, opened 2026-06-18
Concurrency orchestrator: mid-run capacity re-check + grant cloudwatch:GetMetricStatistics in template.yaml


espg:
Follow-up to #28 / PR #41, which landed the pre-flight concurrency probe + FD-exhaustion guard but deliberately deferred two items:

1. **Mid-run re-check** — the probe currently runs once before fan-out. A long run can have account concurrency shift underneath it; re-probe periodically and re-clamp `max_workers`.
2. **IAM grant** — the probe reads `lambda:GetAccountSettings` + CloudWatch `ConcurrentExecutions` and degrades gracefully when denied. To make the account-capacity view reliable in the deployed dispatch role, add `cloudwatch:GetMetricStatistics` (and confirm `lambda:GetAccountSettings`) to `deployment/aws/template.yaml`. This touches deploy infra (CLAUDE.md #1), so it was left out of PR #41 — this issue authorizes that file by name. First confirm what the deployed role already has.

espg:
...what does this get us? What's the trade space?

espg:
🤖 *from Claude*

**Trade space for mid-run capacity re-check:**

The pre-flight probe runs once before fan-out. If account concurrency shifts mid-run (another job launches Lambdas, Reserved Concurrency is modified, a parallel zagg run starts), the probe value is stale for the rest of the run. On short runs (<50 shards, <5 min) this almost never matters — Lambda throttles are transient and exponential-backoff retry handles them cleanly. On long runs (300+ shards, 30+ min), a capacity shift mid-run can cause widespread throttles that retry can't absorb, adding significant wall time and masking the actual capacity state.

**What re-probing gets you:** tighter wall-time predictability on long runs; fewer retry storms under concurrent load; earlier clamping of `max_workers` when capacity drops rather than discovering it via throttle failures.

**Three options:**

**(A) Re-probe every N shards** (e.g., every 50): Low overhead (1–2 API calls per 50 Lambda invocations). Adds a counter + re-probe call inside the dispatch loop. Straightforward and catches capacity shifts as they develop.

**(B) Re-probe every T minutes** (e.g., every 5): Decouples re-probe frequency from shard throughput — better for variable-density workloads. Adds a timestamp to the dispatch loop. Slightly more code.

**(C) Keep pre-flight only; rely on retry**: No re-probe code. The existing exponential-backoff retry already handles transient throttles. The downside only shows at scale under concurrent competing jobs — not the common case today.

[...]

Issue #20, opened 2026-06-05
Backend support (i.e., local processing support)


espg:
Looking to tie the design back to the original modular goals.

Zagg is primarily used for massive parallelization via lambda workers. This is core to the central thesis: most remote sensing problems are IO bound; since NASA (and others) have moved large petabyte scale data archives to the cloud, we design to operate in that environment. Lambda parallelize well, have virtually no latency for spinning up, bill by the second (cost transparency), and spin down immediately when finished. However, the aggregation formulation itself is more general-- we want to define a template for what the aggregation is (i,e., what aggregation functions applied to which data variables over which spaciotemporal extent).

Ideally, we'd want to apply the templates to execute on *multiple* computational backends:

  1. Lambdas
  2. Local simple (i.e., numpy)
  3. Local parallel (i.e., vaex)
  4. Bring-your-own cluster (i.e., slurm, ray, dask, etc)

Item 4 is probably what needs the most deep thought. Item 2 helps enormously with testing and validation-- we want to be able to unit test locally, and to check if our aggregation functions work outside of the cloud. Item 3 is the case for *many* users; data that's 'large' enough to often be out of core for memory, but not local disk space.

To implement this properly we'll need library build options that (for example) bring in vaex, but as an optional dependency (i.e., pip install zagg[local]), so defining metapackages. We'll also need clear abstractions so we can add other backends in the future-- whether they be from other cloud providers, or other execution frameworks.

For now, the tasking is to ensure we maintain item 1, and build a concrete implementation prototype for 2 & 3, as well as plan abstractions to support 4.

espg:
## Two orthogonal dispatch axes

Working through the temporal/event pipeline (#12) surfaced a clean way to frame this. zagg's dispatch has **two orthogonal axes**:

| Axis | Values | Driven by |
|---|---|---|
| **Pipeline type** — *what* you compute | spatial · temporal · event | #12 |
| **Backend** — *where* it runs | lambda · local-simple (numpy) · local-parallel (vaex) · BYO-cluster (ray/dask/slurm) | this issue (#20) |

`agg()` already carries both today (the `backend=` argument plus the new pipeline-type selector). They should stay independent: a pipeline core shouldn't know how it's parallelized, and a backend shouldn't know whether it's running a spatial shard or a temporal event. The thesis in this issue — *the aggregation template is general; the execution environment is swappable* — is exactly this separation.

## The seam: `Strategy × Backend`

Concretely, factor `agg()` so that:

[...]

PR #76, opened 2026-06-22
Dispatcher abstraction: Executor protocol + generic dispatch loop

### LINKED ISSUE #63: Dispatcher abstraction as shared infrastructure (#20, #12)

espg:
Closes #63.

## What this does

Extracts the fan-out → retry → measured-cost loop that lived as two bespoke functions in `runner.py` (`_run_local` / `_run_lambda`) into a clean seam in a new `src/zagg/dispatch.py`, so temporal (#12) and future cluster backends (#20) inherit local + Lambda execution behind one interface.

Implements the **(B)+(C)** design locked with @espg on the issue thread ([comment](https://github.com/englacial/zagg/issues/63#issuecomment-4771120836), [recap](https://github.com/englacial/zagg/issues/63#issuecomment-4771952427)):

- **`Executor` protocol** (`@runtime_checkable`): `preflight(n_cells) -> PreflightReport`, `submit(payload) -> Future`, `measure_cost(result) -> CellCost`, `finalize() -> RunReport`, `shutdown()`.
- **`RetryPolicy(max_attempts, backoff, classify)`** as a separate strategy object, with defaults `LAMBDA_RETRY` / `LOCAL_RETRY`. The only per-backend variation is `classify` — errno-24 / EMFILE is **not** retryable (run-fatal, re-raised), boto3 throttling **is**.
- **Generic `dispatch()`** loop returning a structured `RunReport`. Per-result counting differs between backends, so it lives in a caller-supplied `accumulate` callback rather than being baked in.
- **`LocalExecutor`** (thread pool, zero metered cost) + **`LambdaExecutor`** (boto3 fan-out; pricing model in `measure_cost`).

### Approach / what stayed put (per the locked answers)

- **Public `agg(backend="local"|"lambda")` signature is verbatim.** The executor is an internal seam selected from `backend`, not a new public param.
- **`concurrency.py` stays a helper module**, now called from inside `LambdaExecutor.preflight()` (via the runner's injected `_preflight`) rather than inline.
- **`runner` keeps cost *presentation*** (it formats `gb_seconds` / `estimated_cost_usd` off the `RunReport`); `dispatch()` only returns structured data.
- **#56 (mid-run capacity re-check) is out of scope** — no hook implemented beyond the existing `preflight` seam.
- The boto3 seams (`_invoke_lambda_cell` / `_invoke_lambda_setup` / `_invoke_lambda_finalize` / `compute_available_workers` / `ThreadPoolExecutor`) are **injected into the executors off the `runner` module namespace**, so `dispatch.py` needs no boto3 import and the existing tests that monkeypatch those names keep binding the exact objects in use.

## Phases

- [x] **Phase 1** — lift `_run_lambda` → `LambdaExecutor` + generic `dispatch()` + `RetryPolicy`.
- [x] **Phase 2** — port `_run_local` → `LocalExecutor` on the same loop (validates the protocol on the trivial case).
- [x] **Phase 3** — concurrency probe behind `LambdaExecutor.preflight()`; `dispatch()` returns `RunReport`, runner keeps cost presentation.
- [x] **Phase 4** — tests: protocol conformance for both executors + byte-identical summary/cost pins.

## How it was tested (byte-identical evidence)

The spatial production path must stay byte-identical — Zarr store, run-summary dict keys, and per-cell Lambda event payload all unchanged. Pinned by:

- `tests/test_runner.py::TestInvokeLambdaCellEvent` — per-cell event payload bytes (unchanged on this branch).
- **New** `tests/test_runner.py::TestSummaryKeysByteIdentical` — pins the exact summary-dict **key set** and the data/error counters for **both** backends through the dispatch refactor, plus the Lambda cost math (`4 cells × 2 s × 2 GB = 16 GB-s`, `× $0.0000133334`).
- `tests/test_runner_concurrency.py` — the probe-clamp / FD-exhaustion wiring still passes unchanged (pool sized to the clamped worker count; errno-24 surfaces with `ulimit` guidance).
- **New** `tests/test_dispatch.py` — `RetryPolicy` defaults + classifiers (throttling retryable, EMFILE not), `Executor` protocol conformance (`isinstance(..., Executor)` for both), `LambdaExecutor.measure_cost` pricing, `LocalExecutor` zero-cost, and the generic loop's accumulate/cost/`on_submit_error` behavior.

Local green:
- `ruff check --select=E,F,W,I --ignore=E501` on touched files — clean.
- `ruff format --check` on the new files (`dispatch.py`, `test_dispatch.py`) — clean. (`runner.py` / `test_runner.py` are **not** ruff-format-clean on `main`, and CI's `lint.yml` only runs `ruff check` — so I deliberately did **not** whole-file-reformat them, to avoid ~200 lines of unrelated churn; only the logical edits are present.)
- `mypy` (pre-commit) — **no** errors in `dispatch.py` / `runner.py` / the new tests.
- `pytest -q` — 536 passed, 1 skipped.

## Questions for review

1. **3 pre-existing test failures, not mine.** `tests/test_integration.py::{test_full_integration,test_multiple_parent_cells}` and `tests/test_processing.py::TestWriteDataframeToZarr::test_write_dataframe_to_zarr` fail on a clean `origin/main` checkout (verified by stashing my diff) — they're unrelated to this change, so I left them per the "don't fix unrelated CI failures, flag them" rule. Flagging here.
2. **`runner.py` left un-`ruff-format`'d** (see above) to keep the diff reviewable. Say the word if you'd rather take the whole-file reformat in this PR.
3. **`RetryPolicy` is currently advisory** — the in-process executors apply retry inside `submit` (Lambda's existing `_invoke_lambda_cell` retry; local runs once), matching the pre-refactor behavior so the spatial path stays byte-identical. The policy object is threaded through `dispatch()` so a future loop-level retry (and the #20 cluster backends) can consult one policy without a signature change. OK to land the seam advisory now, or do you want the loop to drive retry centrally in this PR?
4. **#70 rebases onto this** (its functional `dispatch.py` → this protocol shape), as agreed — no action here.

https://claude.ai/code/session_01EN7X53ZAyaCca9rjwv9qZw

---

### DIFF
--- a/src/zagg/dispatch.py
+++ b/src/zagg/dispatch.py
@@ -1,0 +1,396 @@
+"""Generic fan-out -> retry -> measured-cost dispatch loop (issue #63).
+
+Both the spatial pipeline (today) and the temporal / cluster pipelines that
+follow (#12, #20) need the same shape: hand a set of work units to a backend,
+fan them out, retry the transient failures, measure cost, and report. That loop
+used to live as two bespoke functions in ``runner.py`` (``_run_local`` /
+``_run_lambda``). This module extracts it once behind a clean seam so every new
+pipeline kind inherits local and Lambda execution for free, and a future
+ray/dask/slurm backend plugs in behind the same interface.
+
+The seam has three pieces (option (B)+(C), locked on #63):
+
+* :class:`Executor` -- a backend. ``submit(payload) -> Future`` runs one unit;
+  ``preflight(n)`` does any pre-fan-out capacity check; ``measure_cost(result)``
+  turns one result into a :class:`CellCost`; ``finalize()`` runs the end-of-run
+  step and returns a :class:`RunReport`; ``shutdown()`` releases resources.
+* :class:`RetryPolicy` -- *how* to retry, factored out of the executor. The only
+  per-backend variation is *which* exceptions are retryable, captured by the
+  ``classify`` callback. Defaults :data:`LAMBDA_RETRY` / :data:`LOCAL_RETRY`.
+* :func:`dispatch` -- the generic loop. It is pipeline- and backend-agnostic:
+  it drives ``submit`` / ``measure_cost`` and folds each result into a
+  :class:`RunReport` via the caller's ``accumulate`` callback.
+
+``runner.py`` keeps cost *presentation* (it formats ``gb_seconds`` /
+``estimated_cost_usd`` from the report); this module only returns structured
+data. ``concurrency.py`` stays a helper module called from
+``LambdaExecutor.preflight`` -- it is not folded into the executor.
+"""
+
+from __future__ import annotations
+
+import time
+from concurrent.futures import Future, as_completed
+from dataclasses import dataclass, field
+from typing import Any, Callable, Protocol, runtime_checkable
+
+# ---------------------------------------------------------------------------
+# Structured results
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class CellCost:
+    """Measured cost of a single work unit.
+
+    ``compute_time_s`` is the backend-reported execution time (Lambda
+    ``duration_s``; 0 for the local backend, which carries no metered cost).
+    ``gb_seconds`` and ``cost_usd`` are derived by the executor's pricing model
+    (``compute_time_s * memory_gb`` and ``gb_seconds * price_per_gb_sec`` for
+    Lambda; both 0 locally).
+    """
+
+    compute_time_s: float = 0.0
+    gb_seconds: float = 0.0
+    cost_usd: float = 0.0
+
+
+@dataclass
+class PreflightReport:
+    """Outcome of an executor's pre-fan-out capacity check.
+
+    ``workers`` is the (possibly clamped) worker count the loop should fan out
+    with. ``detail`` carries backend-specific context (e.g. the Lambda
+    :class:`~zagg.concurrency.ConcurrencyReport`) for presentation; the generic
+    loop does not interpret it.
+    """
+
+    workers: int
+    detail: Any = None
+
+
+@dataclass
+class RunReport:
+    """Structured outcome of a dispatch run.
+
+    The generic loop populates ``results`` (one per unit) and the rolled-up
+    counters; cost is accumulated per-result via :meth:`Executor.measure_cost`.
+    ``runner.py`` reads this to build the public summary dict and to print cost
+    -- this module never formats or prints.
+    """
+
+    results: list[dict] = field(default_factory=list)
+    cells_with_data: int = 0
+    cells_error: int = 0
+    total_obs: int = 0
+    cost: CellCost = field(default_factory=CellCost)
+
+
+# ---------------------------------------------------------------------------
+# Retry policy
+# ---------------------------------------------------------------------------
+
+
+@dataclass
+class RetryPolicy:
+    """How to retry a transient failure, factored out of the executor (#63).
+
+    The only thing that varies across backends is *which* exceptions are worth
+    retrying, captured by ``classify`` (return ``True`` to retry, ``False`` to
+    give up immediately). ``max_attempts`` and ``backoff`` are shared mechanism.
+
+    Parameters
+    ----------
+    max_attempts : int
+        Total attempts, including the first. Lambda uses 3; local uses 1.
+    backoff : Callable[[int], float]
+        Maps a 0-based attempt index to a sleep (seconds) before the next try.
+    classify : Callable[[BaseException], bool]
+        Returns ``True`` when the exception is retryable. errno-24 / EMFILE is
+        *not* retryable (it is run-fatal and re-raised upstream); boto3
+        throttling is.
+    """
+
+    max_attempts: int
+    backoff: Callable[[int], float]
+    classify: Callable[[BaseException], bool]
+
+
+def _no_backoff(attempt: int) -> float:
+    return 0.0
+
+
+def _expjitter_backoff(attempt: int) -> float:
+    """Exponential backoff with sub-second jitter, matching the old loop."""
+    return (2**attempt) + (time.time() % 1)
+
+
+# Substrings that mark a transient client-side Lambda failure worth retrying.
+# Copied verbatim from the pre-refactor ``_invoke_lambda_cell`` so the retry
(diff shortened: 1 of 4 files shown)


src/zagg/runner.py
```diff
@@ -438,110 +480,138 @@ def _run_lambda(config, catalog_data, store_path, child_order, *,
...
-    memory_gb = 2.0  # Lambda memory in GB
+    # Cost estimate: arm64 pricing = $0.0000133334/GB-second. Compute gb_seconds
+    # and cost *once* over the summed Lambda time (the report carries only the
+    # accumulated compute_time_s) so the arithmetic order -- and thus the last
+    # ULP of estimated_cost_usd -- stays byte-identical to the pre-refactor path
+    # (summing per-cell cost_usd would diverge in FP). Runner owns presentation;
+    # the per-cell CellCost.cost_usd is for the report's structured breakdown.
+    total_lambda_time = report.cost.compute_time_s
+    memory_gb = LAMBDA_MEMORY_GB
     gb_seconds = total_lambda_time * memory_gb
```
espg, line 606: 🤖 *from Claude (review)*

**Folded (was the one blocking finding).** Earlier in this branch the runner read `estimated_cost` straight off `report.cost.cost_usd`, i.e. the loop's **sum of per-cell** `(duration_i · 2.0) · price`. That is not bit-equal to origin/main's single multiply over the summed time, `(Σ duration · 2.0) · price` — IEEE-754 makes the two orders differ in the last ULP for ~83% of heterogeneous-duration runs (verified: `0.00035466843999999996` vs `0.00035466844` for durations `[0.1, 0.2, 0.3, 12.7]`). Since `estimated_cost_usd` is a run-summary value and the hard requirement is byte-identical summaries, this line now computes `gb_seconds`/`estimated_cost` **once** off `report.cost.compute_time_s`, restoring the exact pre-refactor arithmetic order. `gb_seconds` was always safe (×2.0 is exact in binary FP). The per-cell `CellCost.cost_usd` stays for the report's structured breakdown only.

---

src/zagg/dispatch.py
```diff
@@ -0,0 +1,396 @@
...
+        try:
+            result = future.result()
+        except Exception as e:
+            if on_submit_error is not None:
+                on_submit_error(e)
+            raise
+        cost = executor.measure_cost(result)
+        report.cost.compute_time_s += cost.compute_time_s
+        report.cost.gb_seconds += cost.gb_seconds
+        report.cost.cost_usd += cost.cost_usd
```
espg, line 259: 🤖 *from Claude (review)*

Note for #20/#12 backends: the loop still accumulates `cost.cost_usd` per cell here, which is fine for the report's structured per-cell breakdown, but consumers that need a byte-stable aggregate **dollar** figure should derive it once from the summed `compute_time_s` (as `runner._run_lambda` now does), not from this running `cost_usd` sum — the two differ in the last FP ULP. `gb_seconds` accumulation is exact and safe to read directly.

---

tests/test_runner.py
```diff
@@ -319,3 +319,169 @@ def fake_process_shard(grid, shard_key, urls, **kwargs):
...
+        assert summary["backend"] == "lambda"
+        assert summary["cells_with_data"] == 4
+        assert summary["total_obs"] == 12
+        # 4 cells x 2 s x 2 GB = 16 GB-s; cost = 16 * arm64 price.
+        assert summary["lambda_time_s"] == 8.0
+        assert summary["gb_seconds"] == 16.0
+        assert summary["price_per_gb_sec"] == 0.0000133334
+        assert summary["estimated_cost_usd"] == 16.0 * 0.0000133334
+
+    def test_lambda_cost_byte_identical_with_mixed_durations(self, monkeypatch, atl06_config):
```
espg, line 442: 🤖 *from Claude (review)*

**Folded (test-gap finding).** The other cost test (`test_lambda_summary_keys_and_cost`) uses uniform per-cell durations (all `2.0`), the one case where per-cell-accumulation and multiply-once happen to agree — so it could not have caught the ULP divergence above. This test uses **heterogeneous** durations `[0.1, 0.2, 0.3, 12.7]` (with `max_workers=1` so completion order is deterministic) and asserts `estimated_cost_usd == (Σ duration · 2.0) · price`. It fails against the per-cell-accumulation version and passes against the folded fix, pinning the byte-identical requirement.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of the dispatcher refactor. I checked the byte-identical surface hard: per-cell Lambda event payload, run-summary keys/values, counting rules (incl. the `"No granules found"` / `"No data after filtering"` no-data sentinels and the local raised-vs-meta-error asymmetry), the probe→client→setup→fan-out→finalize ordering, `max_pool_connections` sizing, progress-log cadence, FD-exhaustion re-raise, and pool-shutdown-in-`finally`. All faithful — except one real finding, now **folded in this push**:

- **`estimated_cost_usd` was no longer byte-identical** — the loop accumulated per-cell `cost_usd`, which differs from origin/main's single multiply over summed time in the last FP ULP (~83% of heterogeneous-duration runs). Fixed by computing cost once off the summed `compute_time_s` in `runner._run_lambda`; added a mixed-duration test that pins it. See the two inline comments + the test comment.

Everything else verified correct; no further diff-scoped changes. The 3 unrelated pre-existing test failures (`test_integration.py::{test_full_integration,test_multiple_parent_cells}`, `test_processing.py::TestWriteDataframeToZarr::test_write_dataframe_to_zarr`) are flagged in the PR body's "Questions for review" — they fail on a clean `origin/main` checkout, so left untouched.

---

PR #77, opened 2026-06-22
Per-chunk precompute hook (compute once per chunk, use per cell)

espg:
Refs #30

Implements the **per-chunk precompute hook** plus its **`resolution: chunk` companion-array storage** — items **1 + 2** of the #30 thread ([scope decision](https://github.com/englacial/zagg/issues/30#issuecomment-4773774832)) — and (phase 6) the **DEM-anchored segment-variable read** that makes the worked waveform template genuinely `dem_h`-anchored. **Item 3 (multi-chunk-per-worker) is a SEPARATE later PR** and is not here.

### Item 1 — the hook (`aggregation.chunk_precompute:`)
Each named entry (`{expression | function, source, [params], [dtype]}`) is evaluated **ONCE per chunk (shard)** over the shard's **pooled** column data, **before** the per-cell loop. The result is injected into every cell's expression namespace, so a per-cell `expression` can reference a chunk-uniform anchor (e.g. a 128-bin waveform window) instead of recomputing a per-cell one. Mechanism is namespace injection (`{**cell_data, **chunk_scalars}`), `O(1)` per value, and the merge is taken only when the block is present, so a config without it uses the exact pre-existing code path.

### Item 2 — `resolution: chunk` companion arrays
A new per-field attribute `resolution` (`cell` default, or `chunk`). A `resolution: chunk` field (e.g. `offset_h`, `gain_h`) is written **once per chunk** into a companion array shaped at the **chunk grid** (`main.shape // chunk_shape` = number of chunks), indexed by `grid.block_index(shard_key)` — working identically on HEALPix (`block_index` = parent nested id) and rectilinear (row-major `(rb, cb)` chunk index). No per-cell duplication on disk; this retires the per-cell empty-cell band-aid for those fields.

### Phase 6 — DEM-anchored segment-variable read (the worked template's actual anchor)

A **non-base level** may now declare `variables:` as a `{name: path-template}` mapping (a *readable* segment-level variable, distinct from the documentation-only `list[str]` form). At read time each such variable is read at coarse rate and **broadcast to the photon rows** via the level's existing `link` (`np.repeat(values, counts)`-equivalent placement by `index_beg`, reusing the SAME segment→photon tiling the mask path builds), landing as a per-photon column in the pooled shard data that `chunk_precompute` / the aggregation can reduce.

This is exactly the `dem_h` case (`/{group}/geophys_corr/dem_h`, one value per ~100-photon geolocation segment). The worked `configs/atl03_waveform_chunk.yaml` now declares `dem_h` as a segment-level variable and anchors `chunk_offset = np.float32(np.floor(np.median(dem_h)))` — a stable absolute DEM frame — while `chunk_gain` stays on `h_ph` (it's a spread, not an anchor). The template header comment is updated to reflect the DEM anchor.

The read is purely additive: it only fires when a non-base level declares a **dict** `variables`. A `list[str]` segment level (`atl03.yaml`) or any config without a segment-level `variables` mapping reads identically to before — no extra HDF5 read, no extra column.

```yaml
data_source:
  levels:
    segments:
      path: "/{group}/geolocation"
      variables:                       # NEW (phase 6): readable segment-level variable
        dem_h: "/{group}/geophys_corr/dem_h"
      link: {to: photons, index_beg: ".../ph_index_beg", count: ".../segment_ph_cnt", index_base: 1}
aggregation:
  chunk_precompute:
    chunk_offset: {expression: "np.float32(np.floor(np.median(dem_h)))", source: dem_h}  # DEM anchor
    chunk_gain:   {expression: "...", source: h_ph}                                       # spread on h_ph
  variables:
    waveform_counts: {kind: vector, trailing_shape: 128, expression: "...", source: h_ph}
    offset_h: {expression: "chunk_offset", source: h_ph, resolution: chunk}
    gain_h:   {expression: "chunk_gain",   source: h_ph, resolution: chunk}
```

## What changed in the items 1+2 fold (@espg's scope call in 4773774832)

- **(i) Relaxed the scalar guard.** `_eval_chunk_precompute` no longer rejects a non-scalar result — a chunk value may be a scalar **or** an array. Scalar-ness is required only when a value is **written** to a `kind: scalar` field; `calculate_cell_statistics` raises a clear error at that write point.
- **(ii) Enabled `chunk_precompute` on the arrow-kernel path.** The validate-time `NotImplementedError` is removed; the precompute is reduced ONCE over the pooled arrow table and threaded into `_kernel_aggregate`'s fallback per-cell loop.
- **(iii) `resolution: chunk` companion arrays** (item 2): schema attribute, grid template emission (`grids/base.py`), and the write path (`write_dataframe_to_zarr` collapses the chunk-uniform column to its single value, writing it at `block_index`).

## Phase 6 read-path detail

- `processing.py`: `_broadcast_segment_to_base` (per-segment values → base-rate per-photon array via the link tiling), `_segment_level_variables` / `_read_segment_broadcasts` (collect + read declared dict-form segment variables), wired into **both** the planned (`_planned_read_group`) and full (`_read_group_full`) read paths. The broadcast column is subset through the **exact same** spatial/keep/expression masks as the base-rate variables, so each surviving photon keeps its own segment's value — verified for a **partial** read plan, not just full reads.
- `config.py`: `_validate_levels` accepts + validates a non-base level's `{name: path}` `variables` (string names → non-empty templates, base-level mapping rejected, base-var + cross-level name collisions rejected); `_segment_variable_names` folds the names into the valid column set so agg sources / `chunk_precompute` references resolve.

## Byte-identical guarantee

The new attributes/reads are purely additive. `resolution` is **not** in `output_field_signature`/grid `signature()`, so shard-map fingerprints are unchanged. The phase-6 read fires **only** when a non-base level declares a dict `variables`. The other shipped templates — `atl06.yaml`, `atl03.yaml`, `atl03_waveform_counts.yaml` — are **byte-for-byte unchanged on disk** (only `atl03_waveform_chunk.yaml` changed). Pinned by `TestPlainConfigByteIdentical` (resolution machinery) and `test_no_segment_variables_is_inert` (a `list[str]` segment level + a no-levels config trigger zero segment-variable reads, asserted with an h5 stub that raises on any read). The pre-existing `test_processing.py` / `test_integration.py` suites still pass.

## Phases

- [x] **Phase 4** — scalar-guard relax (i) + arrow-kernel enable (ii) + tests.
- [x] **Phase 5** — `resolution: chunk` schema + grid-spec emission + write path + worked-template update + tests.
- [x] **Phase 6** — DEM-anchored segment-variable read: schema (`variables:` on a non-base level) + read/broadcast in both read paths + worked-template `dem_h` anchor + tests.

## Self-review fold (this run, phase 6)

A fresh-context adversarial self-review was posted (inline `🤖 *from Claude (review)*`); the diff-scoped findings are folded in `fold #77 review: dem_h broadcast bounds guard + NaN-fill, cross-level name guard, planned-filter test`:

- **(MED) `_broadcast_segment_to_base` used `np.empty` + overclaimed gap-robustness.** Float dtypes now NaN-fill (an untiled gap surfaces as missing, not garbage), non-float zero-fill; docstring corrected to rely on #43's contiguity assumption. Test `test_broadcast_gap_fills_nan_for_float`.
- **(MED) Seg-variable level distinct from the spatial-index level could write out of bounds.** `_broadcast_segment_to_base` now raises a clear `… exceeds base size …` on any range past the base extent. Test `test_broadcast_out_of_bounds_raises`.
- **(LOW) Two non-base levels declaring the same seg-variable name** silently collided → now rejected in `_validate_levels`. Test `test_duplicate_segment_variable_across_levels_rejected`.
- **(LOW) Planned-path keep-mask × broadcast was untested** → `test_planned_partial_read_aligns_dem_h_under_base_filter`.

(The earlier items-1+2 self-review fold — companion `_chunk_uniform_value` write guard, chunk-uniformity guard, dense-HEALPix companion test — is in `fold #77 review: chunk-uniform companion write guard …`.)

## How tested

`uv run pytest -q` → **662 passed, 1 skipped**. Phase-6 coverage:

- **Broadcast** (`test_processing.py` `TestSegmentLevelVariables`): `np.repeat`-equivalence + `index_base` (1-based ATL03) + dtype preservation; the full path broadcasts `dem_h` to photons and stays aligned under a base filter; the **planned partial read** aligns `dem_h` to read-plan-selected photons (and under a stacked base filter); the worked template's `chunk_offset = floor(median(pooled dem_h))`, uniform across cells, end-to-end through the worker; the no-segment-variable inert/byte-identical case; OOB + gap-fill guards.
- **Schema** (`test_config.py` `TestSegmentLevelVariables`): mapping vs. `list[str]` form; segment variable usable as an agg source / `chunk_precompute` source; empty-template, base-level-mapping, base-var-collision, and cross-level-duplicate rejections; the worked template declares `dem_h` and the DEM-anchored `chunk_offset`.

Lint (`ruff check --select=E,F,W,I --ignore=E501`) and `ruff format --check` clean on touched files; `mypy` (pinned via pre-commit) adds **zero** new error *categories* — see below.

## Questions for review

- **DEM anchor vs. photon-median.** `chunk_offset` now anchors on `floor(median(dem_h))` (the segment reference DEM, broadcast to photons), per the #30 design target; `chunk_gain` remains a spread over `h_ph`. Confirm the median-of-broadcast-dem_h reduction is what you want (vs. e.g. median of the *segment* dem_h values directly — equivalent only under equal photon counts per segment).
- **Segment variables aren't referenceable from `expression` *filters*.** A `dem_h` reference in an aggregation/`chunk_precompute` expression works (tested); but the `expression`-*filter* namespace is still built from `data_source.variables` only, so a `dem_h` reference there would `NameError`. Out of the agreed minimal scope — flagging in case you want it folded.
- (items 1+2, still standing) chunk-uniformity enforced at write time; arrow-kernel + `kind: vector` chunk-output boundary; `resolution: chunk` limited to `kind: scalar`.

## Pre-existing (not mine — flagging, not fixing per §4)

`mypy` (pinned via pre-commit) reports pre-existing errors in `config.py`/`processing.py`/`rectilinear.py` (the `DataSourceDict` arg-types, the `stats_arrays` re-def, object/return-value assignments, the coord chunk_shape typeddict). Base branch has 18; this change has 19 — the single delta (`config.py:236`, `_segment_variable_names` arg-type) is the **identical** `DataSourceDict`→`dict[Any, Any]` pattern as its four sibling validator calls (223/226/229/559), which the base already exhibits; the new helper matches the surrounding style rather than introducing a lone divergent signature. The phase-6 read-path helpers add **zero** mypy errors.

### DIFF
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -43,7 +43,12 @@ class LevelDict(TypedDict):
 
     path: str  # HDF5 group path template (may contain ``{group}``)
     coordinates: list[str]  # coordinate dataset names within ``path``
-    variables: list[str]  # variable dataset names within ``path``
+    # ``variables`` has two forms: a documentation-only ``list[str]`` of names, or
+    # (non-base levels, issue #30) a ``{name: path-template}`` mapping declaring a
+    # *readable* segment-level variable. The mapping form is read at coarse rate and
+    # broadcast to the base (photon) rows via ``link`` so e.g. ``dem_h`` (one value
+    # per ~100 photons) becomes a per-photon column the aggregation can reduce.
+    variables: list[str] | dict[str, str]
     link: NotRequired[LinkDict | None]
 
 
@@ -223,7 +228,13 @@ def validate_config(config: PipelineConfig) -> None:
     # Cross-check: each filter's level field must name a key in levels (issue #43)
     _validate_filter_levels(config.data_source)
 
-    ds_vars = set(config.data_source.get("variables", {}).keys())
+    # Segment-level (non-base) ``variables`` mappings (issue #30) become real
+    # per-photon columns in the pooled shard data once broadcast, so they are valid
+    # column references everywhere a ``data_source.variables`` column is (agg
+    # sources/expressions, chunk_precompute sources). Fold their names into ds_vars.
+    ds_vars = set(config.data_source.get("variables", {}).keys()) | _segment_variable_names(
+        config.data_source
+    )
     agg_vars = config.aggregation.get("variables", {})
 
     # Validate the per-chunk precompute hook (issue #30, item 1). Each entry is
@@ -659,6 +670,29 @@ def _validate_filters(data_source: dict) -> None:
                 raise ValueError(f"filter[{i}]: 'value' must be numeric (got {f['value']!r})")
 
 
+def _segment_variable_names(data_source: dict) -> set[str]:
+    """Names of readable segment-level (non-base) variables (issue #30).
+
+    A non-base level may declare ``variables`` as a ``{name: path-template}``
+    mapping; each name becomes a per-photon column once broadcast at read time
+    (:func:`zagg.processing._read_segment_broadcasts`). The documentation-only
+    ``list[str]`` form contributes nothing. Empty when no level declares the
+    mapping form, so plain configs are unaffected.
+    """
+    levels = data_source.get("levels")
+    base_level = data_source.get("base_level")
+    if not isinstance(levels, dict) or base_level is None:
+        return set()
+    names: set[str] = set()
+    for name, lvl in levels.items():
+        if name == base_level or not isinstance(lvl, dict):
+            continue
+        lvl_vars = lvl.get("variables")
+        if isinstance(lvl_vars, dict):
+            names |= set(lvl_vars)
+    return names
+
+
 def _validate_levels(data_source: dict) -> None:
     """Validate the hierarchical ``levels``/``base_level`` form (issue #43, Phase B).
 
@@ -683,12 +717,41 @@ def _validate_levels(data_source: dict) -> None:
             f"data_source.base_level {base_level!r} is not a key in levels "
             f"(available: {sorted(levels)})"
         )
+    base_vars = set(data_source.get("variables", {}))
     level_keys = set(levels)
     for name, lvl in levels.items():
         if not isinstance(lvl, dict):
             raise ValueError(f"levels.{name} must be a mapping")
         if "path" not in lvl:
             raise ValueError(f"levels.{name}: 'path' is required")
+        # A non-base level may declare ``variables`` as a ``{name: path-template}``
+        # mapping (issue #30): a readable segment-level variable broadcast to the
+        # base rows at read time. Validate it like ``data_source.variables`` (string
+        # names -> non-empty string path templates) and forbid the mapping form on
+        # the base level (the base level uses ``data_source.variables``). The
+        # documentation-only ``list[str]`` form stays valid on any level.
+        lvl_vars = lvl.get("variables")
+        if isinstance(lvl_vars, dict):
+            if name == base_level:
+                raise ValueError(
+                    f"levels.{base_level}: the base level uses data_source.variables; "
+                    f"a non-base level uses the 'variables' mapping for segment-level reads"
+                )
+            for var_name, tmpl in lvl_vars.items():
+                if not isinstance(var_name, str) or not var_name:
+                    raise ValueError(
+                        f"levels.{name}.variables: variable names must be non-empty strings"
+                    )
+                if not isinstance(tmpl, str) or not tmpl:
+                    raise ValueError(
+                        f"levels.{name}.variables.{var_name}: path template must be a "
+                        f"non-empty string (got {tmpl!r})"
+                    )
+                if var_name in base_vars:
+                    raise ValueError(
+                        f"levels.{name}.variables.{var_name}: collides with a "
+                        f"data_source.variables column"
+                    )
         link = lvl.get("link")
         if link is None:
             if name != base_level:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -386,6 +386,11 @@ def _validate_chunk_precompute(aggregation: dict, ds_vars: set[str]) -> None:
 # for variable-length per-cell outputs; see issue #48.
 OUTPUT_KINDS = ("scalar", "vector", "ragged")
 
+# Recognized per-field output resolutions (issue #30 item 2). ``cell`` (default)
+# stores one value per aggregation cell; ``chunk`` stores one value per chunk in a
+# companion array shaped at the chunk grid, indexed by ``grid.block_index``.
+OUTPUT_RESOLUTIONS = ("cell", "chunk")
+
 
 def _validate_output_kind(name: str, meta: dict) -> None:
     """Validate a variable's non-scalar output declaration.
@@ -396,6 +401,14 @@ def _validate_output_kind(name: str, meta: dict) -> None:
     ``vector`` and ``ragged`` fields may be driven by either ``function`` or
     ``expression``; ``len``/``count`` are rejected for both (they short-circuit
     to a scalar count). See issue #29 (vector) and issue #48 (ragged/CSR).
+
+    A field may also declare ``resolution`` (``cell`` default, or ``chunk``).
+    A ``resolution: chunk`` field (issue #30 item 2) is written ONCE per chunk
+    into a companion array shaped at the chunk grid (``main.shape //
+    chunk_shape``), indexed by ``grid.block_index(shard_key)`` — the compact
+    storage for a chunk-uniform value (e.g. a ``chunk_precompute`` anchor). A
+    ``kind: scalar`` field may be ``resolution: chunk``; ``vector``/``ragged``
+    chunk companions are not supported here (deferred — see the error message).
 
     Parameters
     ----------
@@ -414,6 +427,21 @@ def _validate_output_kind(name: str, meta: dict) -> None:
         allowed = ", ".join(OUTPUT_KINDS)
         raise ValueError(
             f"Variable '{name}': output kind '{kind}' is not supported (allowed: {allowed})"
+        )
+
+    # resolution (cell default, or chunk). A chunk-resolution field stores one
+    # value per chunk in a companion array (issue #30 item 2). Only kind: scalar
+    # chunk companions are supported in this PR.
+    resolution = meta.get("resolution", "cell")
+    if resolution not in OUTPUT_RESOLUTIONS:
+        allowed = ", ".join(OUTPUT_RESOLUTIONS)
+        raise ValueError(
+            f"Variable '{name}': resolution '{resolution}' is not supported (allowed: {allowed})"
+        )
+    if resolution == "chunk" and kind != "scalar":
+        raise ValueError(
+            f"Variable '{name}': resolution 'chunk' is only supported for kind 'scalar' "
+            f"(got kind '{kind}'); a chunk-resolution vector/ragged companion is deferred"
         )
 
     # dtype, when declared, must name a real numpy dtype (applies to all kinds).
@@ -881,11 +909,15 @@ def get_output_signature(meta: dict) -> dict:
     Returns
     -------
     dict
-        ``{"kind": str, "trailing_shape": tuple, "inner_shape": tuple, "dtype": str}``.
+        ``{"kind": str, "trailing_shape": tuple, "inner_shape": tuple, "dtype":
+        str, "resolution": str}``.
         ``trailing_shape`` is ``()`` for scalar and ragged fields.
         ``inner_shape`` is ``()`` for scalar and vector fields; for ragged it
         holds the per-element shape (e.g. ``(2,)`` for a centroid pair).
         ``dtype`` is the declared dtype string, or ``None`` if unset.
+        ``resolution`` is ``"cell"`` (default — one value per aggregation cell) or
+        ``"chunk"`` (issue #30 item 2 — one value per chunk, stored in a companion
+        array shaped at the chunk grid and indexed by ``grid.block_index``).
     """
     kind = meta.get("kind", "scalar")
     if kind == "vector":
@@ -904,6 +936,7 @@ def get_output_signature(meta: dict) -> dict:
         "trailing_shape": trailing_shape,
         "inner_shape": inner_shape,
         "dtype": meta.get("dtype"),
+        "resolution": meta.get("resolution", "cell"),
     }
 
 
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -718,6 +718,7 @@ def _validate_levels(data_source: dict) -> None:
             f"(available: {sorted(levels)})"
         )
     base_vars = set(data_source.get("variables", {}))
+    seg_var_names: set[str] = set()  # segment-variable names seen across non-base levels
     level_keys = set(levels)
     for name, lvl in levels.items():
         if not isinstance(lvl, dict):
@@ -752,6 +753,15 @@ def _validate_levels(data_source: dict) -> None:
                         f"levels.{name}.variables.{var_name}: collides with a "
                         f"data_source.variables column"
                     )
+                # Two non-base levels declaring the same name would silently
+                # overwrite each other when broadcast into one per-photon column
+                # (the read keys by name); reject the ambiguity.
+                if var_name in seg_var_names:
+                    raise ValueError(
+                        f"levels.{name}.variables.{var_name}: a segment-level "
+                        f"variable named {var_name!r} is already declared on another level"
+                    )
+                seg_var_names.add(var_name)
         link = lvl.get("link")
         if link is None:
             if name != base_level:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -226,6 +226,13 @@ def validate_config(config: PipelineConfig) -> None:
     ds_vars = set(config.data_source.get("variables", {}).keys())
     agg_vars = config.aggregation.get("variables", {})
 
+    # Validate the per-chunk precompute hook (issue #30, item 1). Each entry is
+    # evaluated ONCE per chunk over the shard's pooled column data, before the
+    # per-cell loop; its name becomes available in the per-cell expression
+    # namespace. Validation mirrors ``aggregation.variables`` (exactly one of
+    # function/expression, sources exist) but the entries are chunk-level scalars.
+    _validate_chunk_precompute(config.aggregation, ds_vars)
+
     for name, meta in agg_vars.items():
         has_func = "function" in meta
         has_expr = "expression" in meta
@@ -271,6 +278,84 @@ def validate_config(config: PipelineConfig) -> None:
         _validate_output_kind(name, meta)
 
 
+def _validate_chunk_precompute(aggregation: dict, ds_vars: set[str]) -> None:
+    """Validate the ``aggregation.chunk_precompute`` block (issue #30, item 1).
+
+    Each named entry is a chunk-level scalar computed ONCE per chunk (shard) over
+    the shard's pooled column data, before the per-cell loop; its name then enters
+    the per-cell expression namespace. Validation mirrors ``aggregation.variables``:
+    each entry must declare exactly one of ``function``/``expression``, and any
+    ``source`` / expression / param column references must exist in
+    ``data_source.variables``. ``dtype`` (optional) must be a valid numpy dtype.
+
+    The block is optional; a config without it is unchanged.
+
+    Parameters
+    ----------
+    aggregation : dict
+        The config's ``aggregation`` mapping.
+    ds_vars : set[str]
+        Available ``data_source.variables`` column names.
+
+    Raises
+    ------
+    ValueError
+        On any invalid ``chunk_precompute`` declaration.
+    """
+    precompute = aggregation.get("chunk_precompute")
+    if precompute is None:
+        return
+    if not isinstance(precompute, dict):
+        raise ValueError("aggregation.chunk_precompute must be a mapping of name -> entry")
+    for name, meta in precompute.items():
+        if not isinstance(name, str) or not name.strip():
+            raise ValueError("chunk_precompute entry names must be non-empty strings")
+        if not isinstance(meta, dict):
+            raise ValueError(f"chunk_precompute '{name}': entry must be a mapping")
+
+        has_func = "function" in meta
+        has_expr = "expression" in meta
+        if has_func and has_expr:
+            raise ValueError(
+                f"chunk_precompute '{name}': 'function' and 'expression' are mutually exclusive"
+            )
+        if not has_func and not has_expr:
+            raise ValueError(f"chunk_precompute '{name}': must specify 'function' or 'expression'")
+
+        source = meta.get("source")
+        if source is not None and source not in ds_vars:
+            raise ValueError(
+                f"chunk_precompute '{name}': source '{source}' not in data_source.variables"
+            )
+
+        if has_func:
+            resolve_function(meta["function"])  # raises ValueError on failure
+
+        for pval in meta.get("params", {}).values():
+            if not isinstance(pval, str):
+                continue  # numeric literal
+            if pval in ds_vars or _is_numeric(pval):
+                continue
+            if any(v in pval for v in ds_vars):
+                continue
+            raise ValueError(
+                f"chunk_precompute '{name}': param value '{pval}' references "
+                f"unknown column (available: {ds_vars})"
+            )
+
+        if has_expr:
+            _validate_expression_columns(name, meta["expression"], ds_vars)
+
+        if "dtype" in meta:
+            try:
+                np.dtype(meta["dtype"])
+            except (TypeError, ValueError) as e:
+                raise ValueError(
+                    f"chunk_precompute '{name}': dtype {meta['dtype']!r} is not a valid "
+                    f"numpy dtype ({e})"
+                ) from e
+
+
 # Recognized per-field output kinds. ``ragged`` (CSR) is the Tier-2 carrier
 # for variable-length per-cell outputs; see issue #48.
 OUTPUT_KINDS = ("scalar", "vector", "ragged")
@@ -302,8 +387,7 @@ def _validate_output_kind(name: str, meta: dict) -> None:
     if kind not in OUTPUT_KINDS:
         allowed = ", ".join(OUTPUT_KINDS)
         raise ValueError(
-            f"Variable '{name}': output kind '{kind}' is not supported "
-            f"(allowed: {allowed})"
+            f"Variable '{name}': output kind '{kind}' is not supported (allowed: {allowed})"
         )
 
     # dtype, when declared, must name a real numpy dtype (applies to all kinds).
@@ -373,8 +457,7 @@ def _validate_trailing_shape(name: str, trailing_shape, key_name: str = "trailin
     for dim in dims:
         if not isinstance(dim, int) or isinstance(dim, bool) or dim < 1:
             raise ValueError(
-                f"Variable '{name}': '{key_name}' entries must be positive "
-                f"integers (got {dim!r})"
+                f"Variable '{name}': '{key_name}' entries must be positive integers (got {dim!r})"
             )
 
 
@@ -732,6 +815,26 @@ def get_agg_fields(config: PipelineConfig) -> dict:
         :func:`get_output_signature` to read the normalized declaration.
     """
     return dict(config.aggregation.get("variables", {}))
+
+
+def get_chunk_precompute(config: PipelineConfig) -> dict:
+    """Return the ``aggregation.chunk_precompute`` entries keyed by name (issue #30).
+
+    Each entry is a chunk-level scalar evaluated ONCE per chunk (shard) over the
+    shard's pooled column data, before the per-cell loop; the resulting scalar is
+    injected into the per-cell expression namespace. Returns an empty dict when no
+    ``chunk_precompute`` block is present, so the existing per-cell path is a no-op.
+
+    Parameters
+    ----------
+    config : PipelineConfig
+
+    Returns
+    -------
+    dict
+        ``{name: {function/expression, source, params, dtype}}``.
+    """
+    return dict(config.aggregation.get("chunk_precompute", {}))
 
 
 def get_output_signature(meta: dict) -> dict:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -233,6 +233,12 @@ def validate_config(config: PipelineConfig) -> None:
     # function/expression, sources exist) but the entries are chunk-level scalars.
     _validate_chunk_precompute(config.aggregation, ds_vars)
 
+    # Chunk-precompute names are injected into the per-cell expression namespace
+    # (issue #30), so a per-cell ``expression`` (or its params) may reference them
+    # like a column. Treat them as valid identifiers in the per-cell validation.
+    precompute_names = set(config.aggregation.get("chunk_precompute", {}).keys())
+    expr_vars = ds_vars | precompute_names
+
     for name, meta in agg_vars.items():
         has_func = "function" in meta
         has_expr = "expression" in meta
@@ -260,19 +266,19 @@ def validate_config(config: PipelineConfig) -> None:
         for pval in meta.get("params", {}).values():
             if not isinstance(pval, str):
                 continue  # numeric literal
-            if pval in ds_vars or _is_numeric(pval):
-                continue  # column reference or number
+            if pval in expr_vars or _is_numeric(pval):
+                continue  # column / chunk-precompute reference or number
             # Expression containing column names (e.g. "1.0 / s_li**2")
-            if any(v in pval for v in ds_vars):
+            if any(v in pval for v in expr_vars):
                 continue
             raise ValueError(
                 f"Variable '{name}': param value '{pval}' references "
-                f"unknown column (available: {ds_vars})"
+                f"unknown column (available: {expr_vars})"
             )
 
-        # Validate expression column references
+        # Validate expression column references (chunk-precompute names included)
         if has_expr:
-            _validate_expression_columns(name, meta["expression"], ds_vars)
+            _validate_expression_columns(name, meta["expression"], expr_vars)
 
         # Validate the output-kind declaration (kind + trailing_shape + dtype)
         _validate_output_kind(name, meta)
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -244,6 +244,15 @@ def validate_config(config: PipelineConfig) -> None:
     # function/expression, sources exist) but the entries are chunk-level scalars.
     _validate_chunk_precompute(config.aggregation, ds_vars)
 
+    # Base-level ``expression`` filters evaluate over the read columns at read time
+    # (before chunk_precompute), so their valid names are exactly ``ds_vars`` —
+    # ``data_source.variables`` plus any broadcast segment-level variable (issue
+    # #30). Validate their column references the same way agg/precompute
+    # expressions are, so e.g. an ``{expression: "dem_h > ..."}`` filter is accepted.
+    for f in filters_from_data_source(config.data_source):
+        if "expression" in f:
+            _validate_expression_columns(f"filter {f['expression']!r}", f["expression"], ds_vars)
+
     # Chunk-precompute names are injected into the per-cell expression namespace
     # (issue #30), so a per-cell ``expression`` (or its params) may reference them
     # like a column. Treat them as valid identifiers in the per-cell validation.
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -294,6 +294,18 @@ def _validate_chunk_precompute(aggregation: dict, ds_vars: set[str]) -> None:
     ``source`` / expression / param column references must exist in
     ``data_source.variables``. ``dtype`` (optional) must be a valid numpy dtype.
 
+    A precompute name must not collide with a ``data_source.variables`` column or
+    a reserved namespace name (``leaf_id``): the per-cell namespace is built as
+    ``{**cell_data, **chunk_scalars}`` in :func:`zagg.processing.process_shard`, so
+    a colliding name would shadow the real column *array* with a 0-d *scalar* and
+    corrupt every cell. Such names are rejected here.
+
+    Inter-precompute references are NOT supported: an entry's expression is
+    evaluated only over the pooled columns (:func:`zagg.processing._eval_chunk_precompute`
+    iterates the entries independently, with no defined order), so one entry cannot
+    reference another (e.g. ``chunk_gain`` cannot read ``chunk_offset``). A name
+    that references another precompute entry is rejected as an unknown column.
+
     The block is optional; a config without it is unchanged.
 
     Parameters
@@ -313,9 +325,17 @@ def _validate_chunk_precompute(aggregation: dict, ds_vars: set[str]) -> None:
         return
     if not isinstance(precompute, dict):
         raise ValueError("aggregation.chunk_precompute must be a mapping of name -> entry")
+    reserved = ds_vars | {"leaf_id"}
     for name, meta in precompute.items():
         if not isinstance(name, str) or not name.strip():
             raise ValueError("chunk_precompute entry names must be non-empty strings")
+        if name in reserved:
+            raise ValueError(
+                f"chunk_precompute '{name}': name collides with a "
+                f"data_source.variables column or the reserved 'leaf_id'; the "
+                f"per-cell namespace merge would shadow the real column with a "
+                f"chunk scalar. Rename the precompute entry."
+            )
         if not isinstance(meta, dict):
             raise ValueError(f"chunk_precompute '{name}': entry must be a mapping")
 
--- a/src/zagg/configs/atl03_waveform_chunk.yaml
+++ b/src/zagg/configs/atl03_waveform_chunk.yaml
@@ -1,0 +1,111 @@
+# ATL03 per-chunk-anchored waveform-counts template (issue #30, item 1).
+#
+# Worked example of the per-chunk precompute hook: the "compute once per chunk,
+# use per cell" primitive. The ``aggregation.chunk_precompute`` block is
+# evaluated ONCE per chunk (shard) over the shard's POOLED photon heights, before
+# the per-cell loop, and the resulting chunk-level scalars are injected into every
+# cell's expression namespace. So every cell in a chunk shares one bin window,
+# unlike atl03_waveform_counts.yaml which anchors on each cell's own median.
+#
+# The anchor and gain (see issue #30 comment 4773320226):
+#   chunk_offset = floor(median(h_ph))   over the whole shard (pooled, not per-cell)
+#   chunk_gain   = smallest power-of-two (>= 0.5 m) whose 128 bins span the shard's
+#                  photon-height range  ->  2^ceil(log2((max - min) / 128)), floored at 0.5 m
+# The 128-bin window is [chunk_offset - 27*chunk_gain, chunk_offset + 101*chunk_gain),
+# so the floor(median) anchor lands in bin 28 (27 bins reserved below it, 100
+# above) -- the asymmetric layout @espg specified for a mostly-above-ground
+# return. Out-of-range photons are dropped (not counted), matching #30's spec.
+#
+# Note (issue #30 thread): the design target for the chunk anchor is the ATL03
+# reference DEM (``dem_h``), which gives a stable absolute frame. dem_h lives in
+# the segment-level geolocation group (one value per ~100 photons), a cross-group
+# / resolution-mismatch read that @espg deferred (the segment-level variable
+# schema change is a follow-up). This template therefore anchors on the pooled
+# ``median(h_ph)`` as the worked example of the hook; swapping in a ``dem_h``
+# source is a one-line change once that read lands.
+#
+# offset_h / gain_h are recorded as per-cell scalar fields: every cell in a chunk
+# carries the chunk's offset/gain identically (duplicated for now -- the compact
+# per-chunk companion-array storage is the #30 item-2 follow-on). A reader
+# reconstructs each chunk's altitude axis as ``offset_h + gain_h * arange(128)``.
+#
+# Confidence filter, multi-level form, read_plan and grid are identical to
+# atl03_waveform_counts.yaml so the planned-IO benefits (#43 Phase C) apply here
+# too; only the aggregation block differs.
+data_source:
+  reader: h5coro
+  driver: s3
+  groups: [gt1l, gt1r, gt2l, gt2r, gt3l, gt3r]
+  coordinates:
+    latitude: "/{group}/heights/lat_ph"
+    longitude: "/{group}/heights/lon_ph"
+  variables:
+    h_ph: "/{group}/heights/h_ph"
+  filters:
+    - dataset: "/{group}/heights/signal_conf_ph"
+      column: 0                  # land surface type; TEP is uniform across columns
+      op: ne
+      value: -2
+  base_level: photons
+  levels:
+    photons:
+      path: "/{group}/heights"
+      coordinates: {latitude: lat_ph, longitude: lon_ph}
+      link: null
+    segments:
+      path: "/{group}/geolocation"
+      coordinates: {latitude: reference_photon_lat, longitude: reference_photon_lon}
+      link:
+        to: photons
+        index_beg: "/{group}/geolocation/ph_index_beg"
+        count: "/{group}/geolocation/segment_ph_cnt"
+        index_base: 1            # ATL03 ph_index_beg is 1-based per the v3 dict
+  read_plan:
+    spatial_index: segments
+    pad: 1                       # one segment of padding on each side per #43
+
+aggregation:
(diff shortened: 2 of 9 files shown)


src/zagg/processing.py
```diff
@@ -1517,8 +1603,13 @@ def process_shard(
                 cells_with_data += 1
             else:
                 cell_data = _empty
+            # Inject the chunk-level scalars into this cell's namespace (no-op when
+            # ``chunk_scalars`` is empty, so non-precompute configs are unchanged).
+            cell_namespace: dict[str, Any] = (
+                {**cell_data, **chunk_scalars} if chunk_scalars else cell_data
+            )
```
espg, line 1930: 🤖 *from Claude (review)*

Empty cells silently drop the chunk scalar. `calculate_cell_statistics` short-circuits at the top: `if n_obs == 0: return {name: _empty_cell_value(meta) ...}`. So for a cell with no observations, an `offset_h: {expression: chunk_offset}` field never evaluates the expression — it gets `np.nan` (scalar) or the vector sentinel, **not** `chunk_offset`, even though `chunk_offset` is a chunk-uniform constant that is perfectly well-defined for an empty cell.

The PR's whole premise is "every cell in a chunk shares one anchor" (see the worked template's `offset_h`/`gain_h` "duplicated for now" comment), but empty cells in the same chunk get NaN instead of the shared anchor — breaking chunk-uniformity for exactly the cells that the dense writer still emits a row for. A reader reconstructing `offset_h + gain_h * arange(128)` will see NaN axes for empty cells.

This is the keystone behavior the tests claim to prove (`test_chunk_scalar_uniform_across_cells`), but no test exercises an empty cell in a chunk that has data elsewhere. Suggest: either inject chunk scalars into the empty-cell result for fields whose `expression` is a bare precompute name, or document/decide that empty cells are intentionally excluded from chunk-anchor recording — and add a test pinning whichever behavior is intended.

---

src/zagg/processing.py
```diff
@@ -1517,8 +1603,13 @@ def process_shard(
                 cells_with_data += 1
             else:
                 cell_data = _empty
+            # Inject the chunk-level scalars into this cell's namespace (no-op when
+            # ``chunk_scalars`` is empty, so non-precompute configs are unchanged).
+            cell_namespace: dict[str, Any] = (
+                {**cell_data, **chunk_scalars} if chunk_scalars else cell_data
```
espg, line 1929: 🤖 *from Claude (review)*

`{**cell_data, **chunk_scalars}` lets a chunk scalar shadow a real column, and there is no validation preventing it (see the validation gap I flagged on `_validate_chunk_precompute`). If someone names a precompute entry after a `data_source.variables` column (e.g. `chunk_precompute: {h_ph: ...}`), the merge replaces the `h_ph` **array** with a 0-d **scalar** in every cell namespace.

That is not just a wrong value: `calculate_cell_statistics` computes `n_obs = len(next(iter(cell_data.values())))`. Insertion order normally keeps a column array first, so `len()` is safe — but a colliding key keeps `cell_data`'s *position* with `chunk_scalars`' *value*. If the colliding column happens to be the first key, `next(iter(...))` returns the 0-d scalar and `len()` raises `TypeError`, crashing the whole shard. Even without a collision, the `n_obs` computation is now structurally relying on dict insertion order to keep a real column first in the merged namespace, which is fragile. Recommend rejecting precompute names that collide with `ds_vars` at validation time (the cheap fix) so this merge can never shadow a column.

---

src/zagg/config.py
```diff
@@ -253,24 +266,102 @@ def validate_config(config: PipelineConfig) -> None:
...
+        On any invalid ``chunk_precompute`` declaration.
+    """
+    precompute = aggregation.get("chunk_precompute")
+    if precompute is None:
+        return
+    if not isinstance(precompute, dict):
+        raise ValueError("aggregation.chunk_precompute must be a mapping of name -> entry")
+    for name, meta in precompute.items():
+        if not isinstance(name, str) or not name.strip():
+            raise ValueError("chunk_precompute entry names must be non-empty strings")
```
espg, line 351: 🤖 *from Claude (review)*

Validation gap: precompute names are only checked for being non-empty strings — there is no check that a name doesn't collide with a `data_source.variables` column (or with `leaf_id`). Since `process_shard` injects via `{**cell_data, **chunk_scalars}`, a colliding name silently shadows the real column array with a 0-d scalar in every per-cell namespace, which both corrupts expressions referencing that column and can crash `n_obs = len(next(iter(...)))` (see my processing.py comment). Add `if name in ds_vars: raise ValueError(...)` here so the collision is rejected up front rather than producing a confusing runtime failure.

---

src/zagg/processing.py
```diff
@@ -162,6 +163,72 @@ def _concat_and_group(all_reads, grid, handoff: str):
...
+                    resolved_params[pkey] = pooled[pval]
+                elif isinstance(pval, str) and any(c in pval for c in pooled):
+                    ns = {"__builtins__": {}, "np": np, "numpy": np, **pooled}
+                    resolved_params[pkey] = eval(pval, ns)  # noqa: S307
+                else:
+                    resolved_params[pkey] = pval
+            value = resolve_function(meta["function"])(values, **resolved_params)
+        dtype = meta.get("dtype")
+        if dtype is not None:
+            value = np.dtype(dtype).type(value)
```
espg: 🤖 *from Claude (review)*

Nothing guarantees a `chunk_precompute` entry yields a *scalar*, and the dtype cast hides the failure mode. `_eval_expression_raw` can return any shape (e.g. an array if the user writes `np.floor(h_ph)` instead of `np.floor(np.median(h_ph))`). Then `np.dtype(dtype).type(value)` on an ndarray either raises or silently coerces, and a non-scalar value gets merged into the cell namespace where downstream expressions misbehave in non-obvious ways. With no `dtype` set, the array passes straight through. There is no validation that the expression/function reduces to a scalar.

Suggest asserting scalar-ness here (e.g. `np.ndim(value) == 0` after eval, raising a clear `ValueError(f"chunk_precompute '{name}' must reduce to a scalar, got shape ...")`), since the whole contract of the hook is "chunk-level scalar."

---

src/zagg/processing.py
```diff
@@ -162,6 +163,72 @@ def _concat_and_group(all_reads, grid, handoff: str):
...
+        return {}
+
+    out: dict[str, Any] = {}
+    for name, meta in entries.items():
+        expression = meta.get("expression")
+        if expression is not None:
+            value = _eval_expression_raw(expression, pooled)
+        else:
+            source = meta["source"]
+            values = pooled[source]
```
espg, line 237: 🤖 *from Claude (review)*

`pooled[source]` will `KeyError` if the `source` column is absent from the pooled dict. This can happen even for a validated config: `pooled` (`col_arrays`) only contains the columns that were actually read. The docstring claims evaluation "mirrors `calculate_cell_statistics`", but the per-cell path reaches this only after the `n_obs == 0` guard, so it never indexes a missing column; here there is no such guard. If every read for a shard came back with a column genuinely missing (or a future read path that omits a variable), this raises a bare `KeyError` mid-shard rather than a clear config/data error.

Also worth noting two real divergences from the agg-field path that the "mirrors" docstring papers over: (1) there is no `len`/`count` short-circuit, so a `function: len` precompute returns the pooled column length rather than a cell `n_obs` — fine here, but not the same code path; (2) the param-resolution `any(c in pval for c in pooled)` substring test is the same loose heuristic as the agg path and can false-match (column name as substring of another token). Neither is fatal, but the docstring should not claim parity it doesn't have.

---

src/zagg/processing.py
```diff
@@ -1459,6 +1528,15 @@ def process_shard(
                 "handoff='arrow-kernel' does not support ragged fields (issue #48); "
                 "use handoff='pandas' or 'arrow' instead"
             )
+        if get_chunk_precompute(config):
```
espg: 🤖 *from Claude (review)*

This guard is placed inside the `handoff == "arrow-kernel"` branch, which is reached only *after* the `if not all_reads: ... return` early exit (~line 1434). So an `arrow-kernel` + `chunk_precompute` run whose reads all come back empty returns a normal empty DataFrame and never raises — the "reject rather than silently drop" contract only holds when there happens to be data. The intent (reject an unsupported combination) is a property of the *config*, not the data, so it should be enforced regardless of whether any rows were read.

Recommend validating the `arrow-kernel` + `chunk_precompute` incompatibility at `validate_config` time (or at the top of `process_shard` before the read loop), so the combination is rejected deterministically. As written, `test_arrow_kernel_rejects_chunk_precompute` only passes because it supplies non-empty data; it would not catch the empty-read hole.

---

tests/test_processing.py
```diff
@@ -2071,3 +2072,160 @@ def test_shipped_atl03_template_through_planned_read(self):
...
+        assert offsets[0] != offsets[1]
+
+    def test_absent_block_is_byte_identical(self, monkeypatch):
+        """A config WITHOUT chunk_precompute produces byte-for-byte identical output
+        to the same config evaluated through the (now chunk-aware) worker — i.e. the
+        hook is a pure no-op when the block is absent."""
+        df_a, _ = self._run_shard(monkeypatch, self._cfg(with_precompute=False))
+        # Re-run the identical no-precompute config; the per-cell path is unchanged.
+        df_b, _ = self._run_shard(monkeypatch, self._cfg(with_precompute=False))
+        assert df_a["offset"].to_numpy().tobytes() == df_b["offset"].to_numpy().tobytes()
```

This test does not prove what its name and docstring claim. It runs the *same* post-PR no-precompute config through `_run_shard` twice and asserts the two outputs match byte-for-byte. That only proves determinism of the new code path — it would pass even if this PR had altered the no-precompute output, because both runs go through the identical (already-changed) code. The docstring says "byte-for-byte identical to the same config evaluated through the (now chunk-aware) worker," but there is no comparison against a pre-PR baseline or against a known expected vector.

To actually pin the "no-op when absent" claim, assert against hardcoded expected values (the per-cell medians `[1.0, 101.0]` and counts `[2, 2]` are already known from `test_per_cell_median_varies_without_precompute`), or compare the no-precompute path output against the same numbers computed directly — not against a second run of itself.

---

src/zagg/configs/atl03_waveform_chunk.yaml
```diff
@@ -0,0 +1,111 @@
...
+  # Evaluated ONCE per chunk over the shard's pooled h_ph, before the per-cell
+  # loop; the names below enter the per-cell expression namespace.
+  chunk_precompute:
+    # ``np.maximum`` (not the ``max`` builtin) floors the gain at 0.5 m, and the
+    # range uses ``np.max``/``np.min`` (not the ``ndarray.max()`` method): the
+    # expression sandbox runs with an empty ``__builtins__``, where an ndarray
+    # method call raises ``KeyError('__import__')`` but the functional numpy form
+    # is fine.
+    chunk_gain:
+      expression: "np.float32(np.maximum(0.5, 2.0 ** np.ceil(np.log2((np.max(h_ph) - np.min(h_ph)) / 128.0))))"
```
espg: 🤖 *from Claude (review)*

Degenerate-range edge case: when a shard's pooled `h_ph` has `max == min` (a single photon, or all-equal heights), `(np.max - np.min) / 128 == 0` and `np.log2(0) == -inf`, so `2.0 ** np.ceil(-inf) == 0.0`; the `np.maximum(0.5, 0.0)` floor saves it to `0.5` here, so this particular expression is safe — but it depends entirely on the floor and emits a numpy `RuntimeWarning: divide by zero encountered in log2` on every such chunk. Since zagg runs per-cell-pooled over whole shards on Lambda, single-photon shards are plausible at grid edges. Consider guarding the log2 input (e.g. `np.maximum(eps, range)`) so the worked template doesn't spew warnings, and add a test with a degenerate pooled range to pin `chunk_gain == 0.5` rather than relying on the implicit floor.

---

src/zagg/config.py
```diff
@@ -253,24 +266,102 @@ def validate_config(config: PipelineConfig) -> None:
...
+                continue
+            if any(v in pval for v in ds_vars):
+                continue
+            raise ValueError(
+                f"chunk_precompute '{name}': param value '{pval}' references "
+                f"unknown column (available: {ds_vars})"
+            )
+
+        if has_expr:
+            _validate_expression_columns(name, meta["expression"], ds_vars)
```
espg, line 393: 🤖 *from Claude (review)*

A precompute expression that references *another* precompute name is validated only against `ds_vars`, so it is rejected here with an "unknown column" error — yet there is no defined evaluation order among precompute entries anyway (`_eval_chunk_precompute` iterates `entries.items()` evaluating each only over `pooled`, never over previously-computed scalars). So inter-precompute references are unsupported in both validation and runtime, which is internally consistent — but worth confirming that's intentional, and the error message ("unknown column") is misleading for a user who reasonably expects to chain `chunk_gain` off `chunk_offset`. If chaining is out of scope, a one-line note in the docstring would save a confused bug report; if it's meant to be supported, both this validation and the eval loop need to thread prior scalars through in declaration order.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of the per-chunk precompute hook (issue #30). Eight inline findings below, ordered roughly by severity. The two I'd treat as blocking are the empty-cell scalar drop and the name-collision validation gap (which can crash a shard via `len()` on a 0-d scalar). The rest are robustness/validation gaps and a test that doesn't prove its claim. No edits made — review only.

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review into one commit (`fold #77 review: …`). Summary of what I folded vs. left standing, and the new tests.

**Folded (diff-scoped findings):**

- **(1) Empty cells now record the chunk anchor, not NaN.** `calculate_cell_statistics`' `n_obs == 0` short-circuit resolves a field whose `expression` is a bare chunk-precompute name from the injected namespace (only `kind: scalar` fields; a vector field's empty cell still gets its sentinel, since a zero-photon histogram is genuinely empty). `n_obs` now counts only length-bearing columns, skipping 0-d scalars — which also hardens the prior insertion-order reliance flagged in (2).
- **(2) Name-collision rejected at validate time.** `_validate_chunk_precompute` rejects a precompute name equal to a `data_source.variables` column or the reserved `leaf_id`, so `{**cell_data, **chunk_scalars}` can never shadow a real column array with a 0-d scalar.
- **(3) Scalar-ness enforced.** `_eval_chunk_precompute` raises a clear `ValueError` when an entry does not reduce to a scalar (`np.ndim != 0`) instead of letting the dtype cast silently coerce/mangle an array.
- **(4) Missing-source guard + docstring.** `pooled[source]` is guarded with a clear error; the docstring now states the real divergences from `calculate_cell_statistics` (no `len`/`count` short-circuit; independent per-entry eval) rather than claiming parity.
- **(5) Arrow-kernel guard moved to validate time.** The `NotImplementedError` for `arrow-kernel` + `chunk_precompute` now fires at the top of `process_shard` before the read loop, so an empty-read run rejects deterministically (the old placement was after the `if not all_reads: return` exit).
- **(6) Degenerate-range guard.** The worked template's `chunk_gain` clamps the pooled range to `1e-6` before `log2`, so a single-photon / all-equal-height shard no longer hits `log2(0)` / a `RuntimeWarning`; the 0.5 m floor still applies.
- **(8) Real parity test.** Replaced the vacuous self-comparison `test_absent_block_is_byte_identical` with `test_absent_block_matches_plain_path_values`, which pins the no-block output against independently-known per-cell medians/counts and asserts the precompute path differs.

**Left standing for review (finding 7 — confirm intent):** inter-precompute references (one entry reading another's scalar, e.g. `chunk_gain` off `chunk_offset`) remain unsupported. There is no defined eval order among entries and each is evaluated only over the pooled columns. Per the thread this is the intended item-1 boundary, so rather than build ordering I **documented** the limitation in `_validate_chunk_precompute` / `_eval_chunk_precompute` and added a test pinning that such a reference is rejected. Building declaration-order chaining is a follow-on if you'd prefer it supported — happy to do it on a yes.

**New tests** (`uv run pytest -q` → 625 passed, 1 skipped):
- `tests/test_processing.py`: `test_empty_cell_gets_chunk_anchor_not_nan`, `test_non_scalar_precompute_rejected`, `test_missing_source_column_clear_error`, `test_arrow_kernel_rejects_chunk_precompute_even_on_empty_reads`, `test_degenerate_single_photon_chunk_gain_floor`, `test_absent_block_matches_plain_path_values`.
- `tests/test_config.py`: `test_name_colliding_with_column_rejected`, `test_name_colliding_with_leaf_id_rejected`, `test_inter_precompute_reference_rejected_as_unknown_column`.

`ruff check`/`ruff format --check` clean on touched files; `mypy` adds zero new errors (pre-existing `DataSourceDict`/`stats_arrays` errors untouched — the `stats_arrays` no-redef line just shifts because the validate-time guard adds lines above it).

The three existing **Questions for review** (expression-sandbox `np.max` vs `.max()`, arrow-kernel support, auto-materialization boundary) plus the finding-7 chaining boundary are non-blocking design confirmations and remain standing for you. Marking ready for review.

---

espg:
> (3) Scalar-ness enforced. _eval_chunk_precompute raises a clear ValueError when an entry does not reduce to a scalar (np.ndim != 0) instead of letting the dtype cast silently coerce/mangle an array.

Why? We just implemented support for vectors being a valid output for cells... so why are we restricting it for chunks? What if there is a chunk level function that returns a matrix (e.g., a covariance matrix)? Is there significant technical obstacles for allowing arrays here too?

> (1) Empty cells now record the chunk anchor, not NaN

What does this mean? Are cells with no data now having data values written in the zarr array? I feel like this issue would be side-stepped by implementing phase two in this PR as well:

> (2) resolution: chunk companion arrays — store those per-chunk values once per chunk instead of duplicated per cell (shape main.shape // chunk_shape, indexed by block_index — identical on HEALPix and rect).

...if we have that in this PR, then we would care less since we no longer *have* to store shard-level values at the cell level.

> (5) Arrow-kernel guard moved to validate time. The NotImplementedError for arrow-kernel + chunk_precompute now fires at the top of process_shard before the read loop

...how much work to enable this for arrow too? My sense is that we use arrow explicitly for matrix / vector output...so we probably need it enabled to allow for non-scalar outputs.

> (3) Multi-chunk-per-worker — one worker owns K chunks; it simply runs the (1) precompute K times (once per chunk) and writes K companion slices. So (3) is a natural extension of (1), not a separate mechanism.

...can we fold this in too?

espg:
🤖 *from Claude*

Four good questions — answering each, then a scope call since this grows item-1 into items 1–3.

**(3) Why restrict chunk results to scalars — can we allow a covariance matrix?** No fundamental obstacle, and you're right to push. The scalar guard only protects the *recording* path (a chunk array silently mangled by the per-cell scalar dtype cast). The namespace-injection mechanism is **shape-agnostic** — a chunk-level `cov` matrix can be injected and referenced by any per-cell `expression` today. So the fix is to **relax the guard**: allow non-scalar chunk results in the namespace, and only require scalar-ness when the value is written as a `kind: scalar` field. *Storing* a chunk matrix is the other half — per-cell it'd be a `kind: vector` field (duplicated), or compactly a per-chunk companion array (item 2). Arrays-for-use = a small relax now; arrays-as-output = item 2.

**(1) Empty cells — what's actually written?** Only the chunk-uniform *metadata* (`offset_h`/`gain_h` = the chunk anchor), not photon data — because storage is per-cell-duplicated and the 64×64×128 tensor reader wants a consistent anchor for every cell, empties included. And you've nailed it: **item 2 side-steps this.** With `resolution: chunk` companion arrays, gain/offset live *once per chunk*; empty cells store nothing extra and the reader reads the chunk-level value regardless of which cells are empty. So the empty-cell handling here is an item-1 band-aid that item 2 retires — agreed it's the principled fix.

**(5)/arrow — how much work?** Moderate, and we do need it. The precompute reduces over the shard's pooled columns; on the arrow path that's an arrow table — pull the source column to numpy, reduce, inject into the per-cell namespace (the arrow path already has a per-cell fallback loop for expression fields). The rejection was a shortcut, not a wall — and since arrow carries `FixedSizeList`/nested outputs, enabling it is exactly what unblocks the matrix/vector chunk results from (3).

**(item 3) Multi-chunk-per-worker — fold in?** Doable, but it's the one piece that reaches *outside* the worker: the orchestrator dispatches one parent cell per worker today; multi-chunk means batching K parent cells per invocation, the worker running the precompute K times and writing K companion slices (indexed by `block_index`), plus a `K` knob. It sits *on top of* item 2's companion arrays.

### Scope call
This turns #77 from "item 1" into "items 1–3 + non-scalar/arrow." Two ways:
- **(a) Fold it all into #77** — hook + non-scalar/arrow + `resolution: chunk` companion arrays + multi-chunk worker in one PR. Coherent, but a large diff across schema, the zarr writer, the arrow path, *and* the orchestrator.
- **(b) Land #77 as the seam, fast-follow the rest** — #77 = the hook + the (3)/arrow relaxations (so non-scalar chunk results work right away), then a focused follow-on PR for **item 2 (companion arrays)** (which also retires the empty-cell band-aid), and **item 3 (multi-chunk)** on top of it.

**My lean is (b)** — `resolution: chunk` is a real writer/schema change and multi-chunk drags in the orchestrator, so bundling all four into one review is a lot; the empty-cell band-aid is harmless until item 2 lands. But if you'd rather it arrive as one capability, (a) is fine.

Two design confirmations either way:
1. **`resolution: chunk` companion field** — a field attribute that writes at the chunk grid (`shape = main.shape // chunk_shape`, indexed by `block_index`, identical HEALPix/rect). Confirm that's the schema shape you want.
2. **Multi-chunk batching** — K chunks/worker as an orchestrator knob; any constraint on K (e.g. a memory budget) you want enforced?

Pick (a) vs (b) + confirm (1), and I'll go. Regardless of that choice I'll **relax the scalar guard and wire the arrow path** (both small, and they unblock the matrix case you flagged) — leaving the empty-cell band-aid in place until item 2 lands. Marking `waiting` on your call.

---

espg:
> resolution: chunk companion field — a field attribute that writes at the chunk grid (shape = main.shape // chunk_shape, indexed by block_index, identical HEALPix/rect). Confirm that's the schema shape you want.

yes.

> Scope call

Neither (A) or (B); fold in items 1 & 2 together, item 3 that touches the orchestrator gets bumped to it's own PR.

>  it's the one piece that reaches outside the worker: the orchestrator dispatches one parent cell per worker today; multi-chunk means batching K parent cells per invocation, the worker running the precompute K times and writing K companion slices (indexed by block_index), plus a K knob. It sits on top of item 2's companion arrays.

...*does* this touch the orchestrator?? The orchestrator dispatches per shard-- that's not changing, neither is anything in shardmap. What does change is that we no longer have one worker one chunk; that could be the model (i.e., shard = chunk), or, the shard could address several chunks. Respond to the above discussion about (3) in #30 please

espg:
🤖 *from Claude*

Got it — folding **items 1 + 2** into #77 (the precompute hook + non-scalar/arrow + `resolution: chunk` companion arrays); **item 3 → its own PR**. `resolution: chunk` schema confirmed. Taking #77 back to **draft** while I add item 2, and I've corrected the item-3 framing on #30 (you're right — it doesn't touch the orchestrator).

What's landing in #77 now:
- **Relax the scalar guard** — non-scalar (vector/matrix) chunk results allowed in the per-cell namespace (a chunk `cov` matrix can feed any `expression`); scalar-ness only required when a value is written as a `kind: scalar` field.
- **Enable `chunk_precompute` on the arrow path** — so non-scalar / `FixedSizeList` chunk results work there too.
- **`resolution: chunk` companion arrays** — `offset_h`/`gain_h` (and any `resolution: chunk` field) stored **once per chunk** at `main.shape // chunk_shape`, indexed by `block_index` (identical HEALPix/rect). This retires the per-cell-duplication empty-cell band-aid — empty cells store nothing extra.

Back to draft; I'll un-draft when 1 + 2 are green + self-reviewed.

---

src/zagg/processing.py
```diff
@@ -283,8 +383,31 @@ def write_dataframe_to_zarr(
...
+
     chunk_idx = tuple(int(i) for i in chunk_idx)
     for name, values in _iter_carrier_columns(df_out):
+        if name in chunk_res_fields:
+            # resolution: chunk — the column must be chunk-uniform (every populated
+            # cell carries the same chunk value), so collapse to the single value and
+            # write it to the companion array's one-block-per-chunk grid at
+            # ``chunk_idx`` (which IS ``grid.block_index(shard_key)``). One value per
+            # chunk, no per-cell duplication on disk (issue #30 item 2).
+            chunk_value = _chunk_uniform_value(name, values)
```
espg, line 399: 🤖 *from Claude (review)*

**[MED] — FOLDED.** The companion write originally took `flat[0]` blindly, trusting cell 0 to carry the chunk value. But the empty-cell rescue in `calculate_cell_statistics` only substitutes the chunk anchor when the field's `expression` is a *bare* precompute identifier (`key.isidentifier() and np.ndim == 0`). For a legal `resolution: chunk` field with a *compound* expression (e.g. `chunk_offset + 1.0`) whose cell 0 is empty, `flat[0]` is the NaN sentinel, so the companion records **NaN**. The shipped `atl03_waveform_chunk.yaml` is masked (its `offset_h`/`gain_h` are bare identifiers), but the validator accepts compound expressions, so it's reachable.

Folded in `e3e7e9a`: the writer now calls `_chunk_uniform_value(name, values)`, which ignores fill/NaN cells and selects a *populated* cell's value (so an empty cell 0 no longer poisons the write). Pinned by `test_empty_cell0_compound_expr_writes_chunk_value_not_nan`.

---

espg, line 399: 🤖 *from Claude (review)*

**[MED] — FOLDED.** Nothing enforced that a `resolution: chunk` column is actually chunk-uniform. A misconfigured field like `resolution: chunk` on `expression: "np.float32(np.mean(h_li))"` (a per-cell mean — valid `kind: scalar`, but not chunk-uniform) would be silently collapsed to one cell's value, dropping every other cell's distinct value with no error.

Folded in `e3e7e9a`: `_chunk_uniform_value` asserts the populated cells are uniform and raises a clear `resolution: chunk field '<name>' is not chunk-uniform: ...` error otherwise. Pinned by `test_non_uniform_chunk_resolution_column_raises`.

---

tests/test_processing.py
```diff
@@ -1171,6 +1172,325 @@ class _OneChunkGrid:
             write_dataframe_to_zarr(table, store, grid=_OneChunkGrid(), chunk_idx=(0,))
 
 
+class TestChunkResolutionCompanion:
```
espg, line 1178: 🤖 *from Claude (review)*

**[LOW] — FOLDED (test gap).** The companion tests originally covered only **fullsphere** HEALPix and rectilinear, leaving the **dense** HEALPix layout (where `block_index` = position in `populated_shards`, not the parent nested id) unverified by CI — and `grids/base.py` calls out exactly that divergence.

Folded in `e3e7e9a`: `test_dense_healpix_companion_at_populated_shard_position` asserts the dense companion is shaped `(n_shards,)` and the value lands at the populated-shard position (`block_index == (1,)`).

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of phases 4+5 (items 1+2). Three diff-scoped findings, **all folded** in `e3e7e9a` (inline above):

- **[MED]** companion writer's `flat[0]` poisoned the chunk value with NaN when cell 0 was empty under a compound `resolution: chunk` expression → now selects a populated cell via `_chunk_uniform_value`.
- **[MED]** no chunk-uniformity guard → a per-cell-varying `resolution: chunk` field silently dropped cells → now raises a clear error.
- **[LOW]** dense-HEALPix companion path was untested → added a dense-layout round-trip test.

Verified correct (no action): byte-identical for cell-only configs (`resolution` never enters `output_field_signature`/grid signature), arrow-kernel `chunk_precompute` threading reaches the fallback expression fields, namespace-merge shadowing already rejected at validate time, `block_index` bounds on all three layouts, and the relaxed scalar guard with the scalar-field write-point error. `uv run pytest -q` → 643 passed, 1 skipped; pinned ruff-check/ruff-format clean; mypy adds zero new errors.

---

espg:
...am I crazy, or does our example still not do what the entire purpose of this PR is?

```
ch degenerate shards are plausible at grid edges.
    chunk_gain:
      expression: "np.float32(np.maximum(0.5, 2.0 ** np.ceil(np.log2(np.maximum(1e-6, np.max(h_ph) - np.min(h_ph)) / 128.0))))"
      source: h_ph
      dtype: float32
    chunk_offset:
      expression: "np.float32(np.floor(np.median(h_ph)))"
      source: h_ph
```

Where are we calculating the *per chunk* dem values? Am I missing it?

espg:
🤖 *from Claude*

Not crazy — but two separate things are getting conflated. The per-chunk *machinery* is working; what's missing is the **DEM source**, which you'd deferred.

**The per-chunk calc is real.** Inside a `chunk_precompute` entry, `h_ph` is the **pooled shard array** — every photon in the chunk — so `np.median(h_ph)` is the **per-chunk** median, computed once before the cell loop and shared by every cell, *not* a per-cell median. That's the keystone capability, and it's pinned: the worked-template test runs `atl03_waveform_chunk.yaml` through `process_shard` and asserts `offset_h`/`gain_h` come out **identical across cells that have different photon distributions**. So the example does demonstrate "compute once per chunk, use per cell."

**But it anchors on `h_ph`, not `dem_h`** — that's the gap you're reacting to, and it's the part you deferred. `dem_h` lives at **segment** resolution in a *different* group (`/{group}/geophys_corr/dem_h`, one value per ~100 photons) than photon-level `h_ph` (`/{group}/heights/h_ph`). Reading it means pulling a segment-level variable and broadcasting it to photons — and on [#30 you said](https://github.com/englacial/zagg/issues/30#issuecomment-4736553394): *"we want the option to pass in dem_h … For now, let's differ on a schema modification until after we test what we already have."* So this PR proved the **mechanism** with the pooled-photon median as a stand-in anchor; it deliberately doesn't read the DEM yet.

**Wiring the real DEM anchor is the deferred piece — contained, but not one line.** It needs: (a) read `dem_h` from the `geophys_corr` segment group and broadcast it to photons via the template's existing `ph_index_beg`/`segment_ph_cnt` link (the read-path addition you held off on — the "phantom variable column vs a dedicated `auxiliary:` key" choice from the thread), then (b) swap the template's anchor to `chunk_offset: "np.floor(np.median(dem_h))"`. The `chunk_precompute` hook itself needs no change — it'll pool and median `dem_h` exactly as it does `h_ph` today.

So: the PR's capability is correct; the worked example just isn't DEM-anchored because the DEM read is the bit you parked. Given your reaction I'd **lean fold-it-in** — add the segment-level `dem_h` read here so the example is genuinely DEM-anchored (the actual point), rather than shipping the `h_ph` stand-in. Want me to do that on this PR (it reopens the read-path schema question you deferred — I'd go with the minimal "`dem_h` as a readable variable" path), or keep the stand-in and wire `dem_h` as the immediate next PR? Say the word and I'll add it.

---

espg:
🤖 *from Claude*

Here's exactly what "dem_h as a readable variable" needs — and I'm folding it into this PR now.

### The gap, concretely
The hierarchical reader already does segment→photon work for the read plan and for filters: it reads the **segment-level** link arrays (`ph_index_beg`/`segment_ph_cnt`) and expands a coarse **mask** down to photon rows (`_expand_mask_to_base`). What it does **not** do is read a coarse-level **variable** and broadcast its **values** to photon rows. `dem_h` (`/{group}/geophys_corr/dem_h`, one value per ~100-photon segment) is exactly that missing case.

### The minimal path (4 steps)
1. **Declare `dem_h` as a segment-level variable** in the template. `geophys_corr` is the same per-segment rate as the `geolocation` segments level that already carries the link, so it rides the *existing* `ph_index_beg`/`segment_ph_cnt`/`index_base` link — no new level, just a variable read at segment resolution:
   ```yaml
   levels:
     segments:
       path: "/{group}/geolocation"
       coordinates: {latitude: reference_photon_lat, longitude: reference_photon_lon}
       variables: {dem_h: "/{group}/geophys_corr/dem_h"}   # segment-rate; broadcast to photons
       link: {to: photons, index_beg: .../ph_index_beg, count: .../segment_ph_cnt, index_base: 1}
   ```
2. **Read + broadcast it** (the one real code addition): in the hierarchical read, after reading the segment-rate `dem_h` array, broadcast each segment's value to its photons with `np.repeat(dem_h_seg, counts)` — reusing the *same* segment→photon index expansion the mask path already builds, so it stays aligned to the photons the read plan actually selected (not all photons). `dem_h` lands as a per-photon column in the pooled shard data.
3. **`chunk_precompute` medians it unchanged** — `dem_h` is now just another pooled column, so `chunk_offset: "np.float32(np.floor(np.median(dem_h)))"` works with zero hook changes; it's the per-chunk median of the model DEM over the shard.
4. **Template swap**: add the `dem_h` variable + change `chunk_offset` from `median(h_ph)` to `median(dem_h)`. `chunk_gain` can stay on `h_ph` (it's a spread, not an anchor) or also move to `dem_h` — your call; I'll keep gain on `h_ph` since the DEM is a smooth surface and the gain should track the actual return spread.

### Scope / why it's the *minimal* path
- The hook, the companion-array writer, the namespace injection — all unchanged. The only new code is "read a declared coarse-level variable and `np.repeat` it to base rows," which is a small extension of the segment machinery that's already there for masks.
- It does reopen the read-path schema bit you parked — I'm taking the **variable-on-the-segment-level** route (above), not a separate `auxiliary:` key, since it reuses the existing `levels`/`link` structure and needs no new schema concept. `dem_h` does pass through as a (otherwise-unused) per-photon column, which is the small cost of the minimal route; an `auxiliary:` key that reads chunk-metadata without a phantom column is the cleaner-but-larger alternative if you'd rather.

Dispatching this onto the PR now. I'll report when the DEM-anchored template + the segment-variable broadcast land green.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the `dem_h` segment-variable read (phase 6, commit `8c7da1f`). Alignment in both read paths is correct: the broadcast `dem_h` column is subset through the *exact* same mask sequence as the base-rate variables — planned path `base_values[seg_global_idx][mask_spatial][keep_mask]`, full path `base_values[min_idx:max_idx][mask_sliced][keep_mask]`, with the expression-filter loop re-subsetting every `data_dict` column including the broadcast. Empty-return short-circuits fire before the column is built, so no desync. Byte-identical/inert claim verified: the new behavior is strictly gated on a non-base level declaring a **dict** `variables` (`list[str]` and no-levels stay completely inert; `test_no_segment_variables_is_inert` proves it with an h5 stub that raises on any read).

Findings (the two MED + one LOW are folded in commit `0f8e21b`):

- **(MED) `_broadcast_segment_to_base` used `np.empty` + overclaimed gap-robustness** (`src/zagg/processing.py`, helper docstring). The mask path is safe-by-construction (`np.zeros`), but a value array has no safe default — an untiled gap would have flowed garbage into the aggregation. **Folded:** float dtypes now NaN-fill (gaps surface as missing, not garbage), non-float zero-fill, and the docstring now states it relies on #43's contiguity assumption rather than claiming gap-robustness. Test `test_broadcast_gap_fills_nan_for_float`.

- **(MED) Seg-variable level distinct from the spatial-index level could write out of bounds.** `n_base` is the spatial-index level's `cnt.sum()`, but the broadcast uses each seg-variable level's *own* link arrays — a mismatched tiling risked a silent OOB write. **Folded:** `_broadcast_segment_to_base` now raises a clear `… exceeds base size …` on any range past `total_base_size`. Test `test_broadcast_out_of_bounds_raises`.

- **(LOW) Two non-base levels declaring the same seg-variable name** would silently overwrite each other (the broadcast keys by name). **Folded:** `_validate_levels` now rejects a duplicate name across levels. Test `test_duplicate_segment_variable_across_levels_rejected`.

- **(LOW) Planned-path keep-mask × broadcast was untested.** **Folded:** `test_planned_partial_read_aligns_dem_h_under_base_filter` stacks a base filter on a partial read and asserts per-segment alignment after the keep mask.

Standing for @espg (out of the agreed minimal scope, NOT folded):

- **Seg-broadcast columns are not referenceable from `expression` *filters*** — both expression-filter namespaces are built from `data_source.variables` only, so a `dem_h` reference in an *expression filter* (distinct from an agg/precompute expression, which does work and is tested) would `NameError`. See "Questions for review".

---

espg:
🤖 *from Claude*

Phase 6 landed: the worked waveform template is now genuinely DEM-anchored.

- **Read path** (`processing.py`): a non-base level declaring `variables: {name: path}` is read at coarse rate and broadcast to the photon rows via the existing segment→photon link (`_broadcast_segment_to_base`, `np.repeat`-equivalent placement reusing the mask path's tiling), landing as a per-photon column in the pooled shard data. Wired into **both** the planned and full read paths; the broadcast column rides the *exact same* spatial/keep/expression masks as the base variables, so each surviving photon keeps its own segment's value — verified for a **partial** read plan, not just full reads.
- **Schema** (`config.py`): `variables:` as a `{name: path-template}` mapping is now allowed + validated on a non-base level (the minimal "dem_h as a readable variable" route, not a new `auxiliary:` key).
- **Worked template**: `dem_h` declared as a segment-level variable from `/{group}/geophys_corr/dem_h`; `chunk_offset` → `np.float32(np.floor(np.median(dem_h)))`; `chunk_gain` stays on `h_ph`; header comment updated to the DEM anchor.

**Byte-identical:** only `atl03_waveform_chunk.yaml` changed among configs — `atl06.yaml`, `atl03.yaml`, `atl03_waveform_counts.yaml` are byte-for-byte unchanged, and the read fires only when a non-base level declares a dict `variables` (`test_no_segment_variables_is_inert` pins this with an h5 stub that raises on any read).

**Green locally:** `uv run pytest -q` → 662 passed, 1 skipped; `ruff check --select=E,F,W,I --ignore=E501` + `ruff format --check` clean on touched files; `mypy` adds zero new error categories (the single line delta is the documented `DataSourceDict` arg-type pattern shared by the existing sibling validators — detail in the PR body).

The fresh-context adversarial self-review is posted above and its diff-scoped findings are **folded** (NaN-fill + OOB guard on the broadcast, cross-level name-collision guard, planned-path keep-mask test). Two items are left **standing for review** (not folded, out of the agreed minimal scope): the DEM-anchor reduction choice and whether segment variables should also be referenceable from `expression` *filters* — both under "Questions for review".

Marking `waiting` (all phases complete, awaiting review/merge). CI is running on the latest push.

---

espg:
> DEM anchor vs. photon-median. chunk_offset now anchors on floor(median(dem_h)) (the segment reference DEM, broadcast to photons), per the https://github.com/englacial/zagg/issues/30 design target; chunk_gain remains a spread over h_ph. Confirm the median-of-broadcast-dem_h reduction is what you want (vs. e.g. median of the segment dem_h values directly — equivalent only under equal photon counts per segment).

this is close-- it should be floor(min(dem_h)) ; dem_h is less noisy, we don't need median.

> Segment variables aren't referenceable from expression filters. A dem_h reference in an aggregation/chunk_precompute expression works (tested); but the expression-filter namespace is still built from data_source.variables only, so a dem_h reference there would NameError. Out of the agreed minimal scope — flagging in case you want it folded.

fold this in now, before i forget to fix it later (we should be able reference linked variables)

> (items 1+2, still standing) chunk-uniformity enforced at write time; arrow-kernel + kind: vector chunk-output boundary; resolution: chunk limited to kind: scalar.

is chunk still limited to kind: scalar, or am i misreading?

espg:
🤖 *from Claude*

Three follow-ups — folding (1) and (2), answering (3).

**(1) `floor(min(dem_h))`, not median.** Got it — `dem_h` is the smooth model surface, so `min` is the right floor (median was carrying noise we don't need). Changing `chunk_offset` to `np.float32(np.floor(np.min(dem_h)))`.

**(2) Linked variables referenceable from `expression` filters — folding in.** Right now the broadcast `dem_h` column reaches **agg / `chunk_precompute`** expressions (tested), but the **filter** expression namespace is still built from `data_source.variables` only, so a `dem_h` reference in an `expression` filter would `NameError`. I'll add the broadcast segment-variable columns to the filter namespace (and make sure they're materialized before filters run) so linked variables work everywhere — with a test that filters on `dem_h`.

**(3) Yes — `resolution: chunk` is currently `kind: scalar` only** (a `vector`/`ragged` chunk companion is rejected at validate time). That was the stated item-2 boundary, not an accident. A `kind: vector` chunk companion (one e.g. 128-length array stored *once per chunk* rather than per cell) is doable but a real add — a vector-shaped companion array at the chunk grid + a uniform-collapse over the trailing axis. **My lean:** land this PR as scalar-companion + the two fixes above, and do vector chunk companions as a focused follow-up (it's the natural pair to item 3's multi-chunk work). But if you'd rather have it here, say so and I'll fold it.

Dispatching (1) + (2) now.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of the two follow-ups in `dcceabb` (DEM anchor min→median; segment-broadcast vars referenceable from `expression` filters). No HIGH/MED findings; the one diff-scoped [LOW] (missing planned-path filter test) was folded into this same commit.

**Verified correct:**
- **min/median change complete.** `atl03_waveform_chunk.yaml` `chunk_offset` now `np.float32(np.floor(np.min(dem_h)))`; header comment updated. Every assertion pinning the *shipped* config's anchor was moved to min: `test_worked_template_chunk_offset_is_floor_min_dem_h` (fixture `[100,100,102,108,110]` → min 100 ≠ median 102), `test_worked_template_uniform_offset_and_gain` and the companion-array end-to-end test (`[50,50,60,60]` → min 50 ≠ median 55). Each fails on revert, so they genuinely test the new behavior. Synthetic `np.median(h_ph)`/`np.median(h_li)` precompute tests and the validation-only `np.median(dem_h)` case are intentionally left — they don't pin the shipped anchor.
- **Namespace fix ordered correctly in both read paths.** `seg_broadcasts` is materialized into `data_dict` *before* the `expression`-filter loop in both `_planned_read_group` and `_read_group_full`; `expr_names = list(variables) + list(seg_broadcasts)` is in scope at each loop.
- **No-regression / byte-identical holds.** `_read_segment_broadcasts` returns `{}` with no segment vars, so `expr_names == list(variables)` exactly; it also rejects name collisions with base columns, keeping the key sets disjoint. Configs without segment vars and filters not referencing a linked var are unchanged.
- **`config.py` validation correctly placed.** Expression-filter column names are validated against `ds_vars` (variables + segment names) *before* chunk-precompute names are folded in, so a filter accepts `dem_h` but correctly cannot see precompute names; a truly-undefined name is rejected. No shipped config carries an expression filter, so the new validation can't regress one.

**Folded [LOW]:** added `test_planned_path_expression_filter_references_broadcast_dem_h` so the planned path's filter-namespace fix has its own parity test (previously only the full path was directly covered).

**Standing (not introduced here, do not fix per §4):** pre-existing `mypy` errors in `processing.py` (lines 1392/1745/1810/1891) and missing-stub noise; the changed lines add zero new mypy errors.

---

espg:
🤖 *from Claude*

Folded the two follow-ups from the "Questions for review" into `dcceabb`:

**(1) DEM anchor median → min.** `configs/atl03_waveform_chunk.yaml` `chunk_offset` is now `np.float32(np.floor(np.min(dem_h)))` (dem_h is less noisy than the photon cloud, so the min is a stable floor — no median needed). Header comment updated; tests pinning the shipped config's anchor moved to min.

**(2) Linked (segment-broadcast) variables are now referenceable from `expression` filters.** Both read paths (`_planned_read_group`, `_read_group_full`) built the `expression`-filter namespace from `data_source.variables` only, so a `dem_h` reference there would `NameError`. The broadcast columns are already materialized into `data_dict` before the filter loop, so the fix is to widen the namespace: `expr_names = list(variables) + list(seg_broadcasts)`. `config.py` now also validates `expression`-filter column names against `ds_vars` (which already folds in segment-variable names), consistent with agg/precompute expressions — accepting `dem_h`, rejecting truly-undefined names.

Tests: `test_expression_filter_references_broadcast_dem_h` (full path) and `test_planned_path_expression_filter_references_broadcast_dem_h` (planned path) keep/drop photons by their broadcast `dem_h`; `test_segment_variable_usable_in_expression_filter` / `test_unknown_name_in_expression_filter_rejected` cover validation.

**Byte-identical:** only `atl03_waveform_chunk.yaml` changed among configs. With no segment vars, `seg_broadcasts` is empty so `expr_names == list(variables)` exactly, and the new validation can't regress any shipped config (none carry an `expression` filter). The `resolution: chunk` = `kind: scalar` boundary was left untouched (separate pending question).

Green locally: `uv run pytest -q` → 666 passed, 1 skipped; `ruff check --select=E,F,W,I --ignore=E501 src tests` and `ruff format --check` clean on touched files. A fresh-context adversarial self-review (above) found no HIGH/MED issues; its one [LOW] (planned-path filter test) is already folded into this commit. Pre-existing `mypy` errors in `processing.py` are unrelated to this diff (flagged, not fixed per §4).

---

PR #78, opened 2026-06-22
fix t-digest δ inversion via k1 scale-function bound (Refs #48)

espg:
## Summary

Fixes a correctness bug in `zagg.stats.tdigest` (follow-up to #48): the t-digest's `δ` (compression) parameter was **inverted** and compression was independent of the observation count, so the sketch over-merged catastrophically.

### The bug

`_budget()` returned a per-centroid weight cap of `δ·(π/2)·sin(πq)` — proportional to **δ** and **independent of n**. The correct cap is proportional to **n/δ**. Symptoms:

- **δ inverted.** Larger δ gave *fewer*, coarser centroids. On a bimodal mixture, mean interior-quantile error went `0.005 → 0.33` as δ went `64 → 2048` (a resolution knob behaving as a down-sampler).
- **No saturation.** Max centroid weight was pinned at `δ·π/2` (≈804 at δ=512) regardless of n, so a fixed ~142× compression applied at every n — N=100 collapsed to **1** centroid, N=1000 to **3**.
- **Loss-free boundary at n=1.** Any cell with ≥2 observations was already compressed.

Existing tests passed only because they assert a loose `k ≤ 4δ` *upper* bound (trivially true when k is tiny) and quantile tolerances on smooth unimodal normals, where a handful of centroids suffice. Nothing asserted that δ increases resolution or that small-n cells stay loss-free. (`_k_scale()` was defined but unused — the intended scale-function approach had been abandoned for the broken budget shortcut.)

### The fix

Replace the budget shortcut with the standard **k1 scale-function** bound: a point joins the current centroid while the centroid still spans ≤ 1 unit of `k(q) = δ·(arcsin(2q−1)/π + 1/2)` (maps cumulative rank `q ∈ [0,1]` onto `[0, δ]`). `dk/dq` is largest at the tails and smallest at the median → narrow high-resolution tail centroids, wide middle centroids. Applied to both `build_tdigest` and `merge_tdigests`; the dead/incorrect `_k_scale`/`_budget` are removed.

Resulting behavior:

| | before (δ=512) | after (δ=512) |
|---|---|---|
| N=100 | k=1 | k=100 (loss-free) |
| N=1000 | k=3 | k=585 |
| N=50000 | k=354 | k=517 (saturates ~δ) |
| δ scaling (N=50k) | 128→k354, 1024→k191 (inverted) | 128→k129, 1024→k1043 |

Loss-free for n ≤ δ, saturates at ~δ (≤ 2δ) for any n, quantiles track exact `np.quantile` within ~0.05 σ.

## Phases

- [x] **Phase 1** — fix `build_tdigest` + `merge_tdigests` to the k1 scale-function bound; remove dead `_k_scale`/`_budget`; update module docstring.
- [x] **Phase 2** — regression tests guarding each symptom (loss-free ≤ δ, compression past δ, δ-monotonic resolution, saturation ≤ 2δ, no high-δ accuracy degradation, quantiles vs exact, merge saturation).

## How tested

- `pytest tests/test_tdigest.py` — 54 passed (40 pre-existing + 14 new).
- Full suite `pytest` — 536 passed, 1 skipped.
- `ruff check --select=E,F,W,I --ignore=E501` clean on changed files; `ruff format --check` clean.
- mypy clean on the changed files (pre-existing mypy errors in `config.py`/`processing.py`/`rectilinear.py` are unrelated and present on `main` — not touched here).

## Questions for review

- **Quantile interpolation unchanged.** This PR only fixes centroid *construction*; `quantile_from_tdigest` is untouched. Accuracy is good, but interior-quantile error on structured data isn't strictly monotonic in δ (a small interpolation artifact, not the inversion). Worth a follow-up if tighter interior accuracy matters.
- **`merge_tdigests` re-compression** uses the same k1 bound on cumulative weight fractions. It saturates and conserves total weight; if many partial digests are merged hierarchically, accumulated re-compression error could be worth a dedicated test.

### DIFF
--- a/tests/test_tdigest.py
+++ b/tests/test_tdigest.py
@@ -257,3 +257,89 @@ class TestTDigestDeltaSweep:
         # Centroid count should grow sub-linearly with δ (proportional, not more).
         ratio = len(digest) / delta
         assert ratio <= 4.0, f"delta={delta}: centroid/delta ratio {ratio:.2f} > 4.0"
+
+
+class TestScaleFunctionRegression:
+    """Guards against the k1-budget regression where δ was inverted.
+
+    Before the scale-function fix the per-centroid weight cap was proportional
+    to δ (and independent of n), so larger δ produced *fewer*, coarser centroids
+    and even a handful of points collapsed to a single centroid. These tests pin
+    the correct behavior: δ is a resolution knob, the digest saturates at ~δ
+    centroids, and it is loss-free until the count exceeds δ.
+    """
+
+    @pytest.mark.parametrize("n", [10, 100, 500])
+    def test_loss_free_when_count_at_most_delta(self, n):
+        """With n ≤ δ every observation is kept as its own weight-1 centroid."""
+        rng = np.random.default_rng(n)
+        vals = rng.standard_normal(n)  # distinct values w.p. 1
+        digest = build_tdigest(vals, delta=512)
+        assert digest.shape[0] == n, f"n={n}: expected {n} centroids, got {digest.shape[0]}"
+        np.testing.assert_array_equal(digest[:, 1], np.ones(n, dtype=np.float32))
+
+    def test_compression_begins_past_delta(self):
+        """Once n exceeds δ the digest must actually compress (k < n)."""
+        rng = np.random.default_rng(1)
+        delta = 256
+        vals = rng.standard_normal(4 * delta)
+        digest = build_tdigest(vals, delta=delta)
+        assert digest.shape[0] < len(vals)
+        np.testing.assert_almost_equal(float(digest[:, 1].sum()), len(vals), decimal=4)
+
+    def test_delta_controls_resolution_not_inverted(self):
+        """More δ ⇒ more centroids. The old budget did the opposite."""
+        rng = np.random.default_rng(2)
+        vals = rng.standard_normal(50_000)
+        counts = [build_tdigest(vals, delta=d).shape[0] for d in (128, 256, 512, 1024)]
+        assert counts == sorted(counts), f"centroid count not monotonic in δ: {counts}"
+        assert counts[-1] > 2 * counts[0], f"δ=1024 barely finer than δ=128: {counts}"
+
+    @pytest.mark.parametrize("n", [50_000, 200_000])
+    def test_centroid_count_saturates_near_delta(self, n):
+        """Count stays ~δ regardless of n (≤ 2δ), instead of growing with n."""
+        rng = np.random.default_rng(n)
+        vals = rng.standard_normal(n)
+        k = build_tdigest(vals, delta=256).shape[0]
+        assert k <= 2 * 256, f"n={n}: {k} centroids exceeds 2·δ"
+
+    def test_accuracy_not_degraded_at_high_delta_on_structured_data(self):
+        """On a bimodal mixture, large δ must stay accurate.
+
+        This is the user-visible symptom of the inversion bug: interior-quantile
+        error blew up as δ grew (≈0.33 here at δ=1024 — ~60× the δ=128 error).
+        A correct digest keeps high-δ error small and comparable to low-δ.
+        """
+        rng = np.random.default_rng(3)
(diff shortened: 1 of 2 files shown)


tests/test_tdigest.py
```diff
@@ -257,3 +257,89 @@ def test_larger_delta_not_worse_than_smaller_for_median(self, delta):
...
+    """
+
+    @pytest.mark.parametrize("n", [10, 100, 500])
+    def test_loss_free_when_count_at_most_delta(self, n):
+        """With n ≤ δ every observation is kept as its own weight-1 centroid."""
+        rng = np.random.default_rng(n)
+        vals = rng.standard_normal(n)  # distinct values w.p. 1
+        digest = build_tdigest(vals, delta=512)
+        assert digest.shape[0] == n, f"n={n}: expected {n} centroids, got {digest.shape[0]}"
+        np.testing.assert_array_equal(digest[:, 1], np.ones(n, dtype=np.float32))
```
espg, line 279: 🤖 *from Claude (review)*

Minor / non-blocking test-quality note. `test_loss_free_when_count_at_most_delta` parametrizes n ∈ {10, 100, 500} against δ=512, so it never exercises the actual boundary n == δ. I verified empirically that n == δ is reliably loss-free (0 failures across δ ∈ {128,256,512} × 50 seeds) and that compression for δ=512 first appears only around n≈652 (≈1.27·δ). The claim in the class docstring and the `_k1_scale` docstring is therefore *conservative and correct*, but the test under-pins it: a case at exactly n == δ (and optionally n == δ+1 to confirm the regime change) would lock the loss-free/compress transition this PR is about. Not a defect — just a stronger pin if you want it.

espg (commented):
🤖 *from Claude (review)*

Adversarial correctness + test-quality review of the k1 scale-function fix. I read the full `tdigest.py` and exercised the diff directly. **The change is correct** — I found no defect introduced by this PR. Evidence:

**Boundary / indexing (build loop).** `k_left` is correctly reset to `k(i/n)` on finalize — the left rank-edge of the freshly-opened centroid — and `k_right = k((i+1)/n)` is the right edge *after* including point i. The span `k_right - k_left` is thus the true cumulative-rank k-span of the candidate centroid; merging while it is ≤ 1 is right. First point is unconditionally seeded. No off-by-one.

**Merge loop.** Uses cumulative *weight* fractions, not counts: `cum_left` tracks total weight before the current centroid and `k_right` uses `(cum_left + cur_weight + v_weight)/n_total`. `k_left` updates to `k(cum_left/n_total)` on finalize. Verified weight conservation is exact (50k+50k merge → sum 100000.0; 20× chained 1k merges → 20000.0), means stay sorted, output is float32.

**Merge ≈ build accuracy.** merge(build(a), build(b)) vs build(a∪b) on 2×100k agree to ~1e-4 across q ∈ {0.01,0.1,0.5,0.9,0.99}. The existing `test_merged_vs_one_shot_quantile_within_tolerance` already covers this, so it is not a gap.

**Saturation / loss-free claims hold.** n == δ is reliably loss-free (0/150 trials compressed); first compression at δ=512 appears ~n≈652. Saturation cap is tight: max centroids/δ ≈ 1.02–1.06 across δ ∈ {128…1024} and n up to 1e6. So the docstring's '~δ centroids' and 'loss-free while count ≤ δ' are conservative and accurate.

**Numerics.** `arcsin` argument is clamped via `min(1, max(0, q))`, so q=1.0 (last point) and any float drift past 1 are safe.

**Edge cases.** n=1, n=2, all-equal, single-non-NaN, empty, all-NaN all behave correctly.

**Tests.** The new `TestScaleFunctionRegression` class genuinely pins the regression direction (δ-monotonic resolution, saturation, loss-free, structured-data accuracy) — none of them pass for the wrong reason. One non-blocking note left inline: the loss-free test never hits the exact n == δ boundary it is about.

**Out of scope (not introduced here):** `merge_tdigests` divides by `n_total` and will raise ZeroDivisionError on all-zero-weight input. That division predates this PR (old code divided by the same `n_total`) and zero-weight digests never arise from `build_tdigest`, so it is not a regression — flagging only for awareness.

LGTM on correctness.

Issue #79, opened 2026-06-22
Read helper(s)


espg:
These are a bit beyond the core functionality of zagg -- but we still need them so that the clients can read the data, and we might as well keep them under version control.

## t-digest to tensor reader

Client expects a fixed size numpy array -- the size should be determined by the chunk size (for our purposes it will be 64 by 64 in x/y) and a "z" resolution provided at runtime by either the resolution of the bins, or number of bins (the later will certainly be 128). So our reader could reasonably be a generator function that's init'ed with the zarr array path, and z-resolution argument; on calling 'next' it would emit the tensor as a numpy array, and also the morton index (i.e., the coverage cell for that tensor).

The general workflow that I've seen for waveforms is that there's a density slice that cuts the tails of the distribution. I.e., they'll start their first read bin at 5% of photons, and the last one at 95% of photons. So the 128-length vector bins would capture 90% of the data in that example. The actual values should of course be configurable (i.e.., 3% and 98%); I'll refer to these parameters as `bottom` and `top`

Each chunk that's read in would need to calculate the `bottom` and `top` values per cell, and then get the respective min and max of those distributions; that will define the range of the tensor (presumably, these min and max values will get cast to floor and ceiling respectively, so we have a range that's aligned on something like 'meter').

My sense is that, to start, we'd define a resolution (e.g., 0.5 meters), and a bin length (e.g., 128), and then would emit the vector at that length and resolution, and raise if that's not possible. More advanced flags would allow for resolutions degradation in powers of 2 if the data didn't fit, or conversely, collapsing the vector length if a power of 2 bin reductions when possible. But default behavior can be simple (fixed tensor, throw if not possible).

espg:
🤖 *from Claude*

`plan` — implementation plan for the t-digest→tensor reader, grounded in the current `stats/tdigest.py` + `csr.py`. No code yet.

## Where it lives
A new client-read module, kept in-repo as you said: `src/zagg/readers/tdigest_tensor.py` (+ `src/zagg/readers/__init__.py`). It's pure-numpy + zarr (both already core deps), so no new dependency and no extra needed — though if you'd rather isolate "client read helpers" behind a `zagg[readers]`/`zagg[client]` extra, that's a one-line `pyproject` call.

## Input it consumes
[...]

PR #81, opened 2026-06-22
t-digest → tensor read helpers (Refs #79)

espg:
Refs #79

## What

Client-side read helpers (issue #79) that reconstruct fixed-size tensors from the per-cell t-digest CSR ragged store zagg writes (issue #48). New `zagg.readers` package (core, no new dependency — pure numpy + zarr).

- **`cdf_from_tdigest(digest, x)`** (`src/zagg/stats/tdigest.py`) — the value→cumulative-weight inverse of `quantile_from_tdigest`, the one primitive needed to fill evenly-spaced *value* bins. Each centroid `(m_k, w_k)` is placed at the centre of its cumulative-weight span (`cum_before + w_k/2`); the CDF interpolates cumulative weight linearly in value-space between centroid means (`np.interp`) and is clamped flat outside `[m_0, m_last]`. Monotonic non-decreasing; endpoints 0 and total weight; empty digest → NaN; scalar-in/scalar-out, array-in/array-out. **Purely additive — does not touch `build_tdigest`/`merge_tdigests` (PR #80 is vectorizing those).**
- **`rasterize_cell(digest, z_lo, resolution, n_bins)`** — `count[i] = cdf(edge_{i+1}) - cdf(edge_i)`, clipped to ≥0.
- **`chunk_z_range(...)`** — per cell `lo_c = quantile(d, bottom)`, `hi_c = quantile(d, top)`; `z_lo = floor(min lo_c)`; fixed window `n_bins*resolution` anchored at `z_lo`. `fit="raise"` (default) raises when `ceil(max hi_c) > z_lo + window`; `fit="degrade_resolution"` doubles `resolution` (powers of two) until it fits; `fit="collapse_bins"` shrinks `n_bins` to the smallest power-of-two bin count whose window still covers the range (raises if even the full window can't).
- **`read_tensors(store, field, *, n_bins=128, resolution=0.5, bottom=0.05, top=0.95, fit="raise", dtype="uint32")`** — generator yielding `(tensor[64,64,n_bins], morton_index)` per coverage chunk. Reuses `csr.read_csr`/`iter_csr_cells`, places cells row-major via `divmod(cell_id, 64)`, recovers the chunk morton from the store.
- **`read_raw_values(store, field)`** — companion lossless reader (per the issue-thread follow-up): when a digest has no merges (all weights 1) the centroid means *are* the original samples, so it yields the raw sorted vector per cell; raises on any merged centroid.

### Store layout consumed

Per shard, a CSR ragged group named by its **parent morton id** (the coverage cell): `{field}/{parent_morton}/{values,offsets,cell_ids}`. `cell_ids[k]` is the cell's position in the chunk's row-major 64×64 children block; `values[offsets[k]:offsets[k+1]]` is that cell's `(k,2)` digest. The chunk morton is recovered from the store — a sibling `{field}/morton` uint64 coord array (aligned against subgroup names sorted **numerically**) if present, else the subgroup name.

## Phases

- [x] Phase 1 — `cdf_from_tdigest` + tests (monotonic, endpoints, empirical-CDF parity, quantile round-trip).
- [x] Phase 2 — `rasterize_cell`; parity test vs `np.histogram` of the original samples within t-digest tolerance.
- [x] Phase 3 — `chunk_z_range` + the three fit modes.
- [x] Phase 4 — `read_tensors` generator + `read_raw_values`; end-to-end against a synthetic CSR store built with `build_tdigest` + `write_csr`.

A fresh-context adversarial self-review ran after the implementation; its two diff-scoped findings are folded: (a) the `{field}/morton` coord array is now aligned against numerically-sorted subgroup names (lexicographic sort mis-aligned mixed-digit mortons), and (b) `collapse_bins` now returns a true power-of-two bin count for non-power-of-two `n_bins`. Both have regression tests.

## How tested

`uv run pytest -q` — 644 passed, 1 pre-existing skip; 91 of those are the new `test_readers.py` + `test_tdigest.py` cases. `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on touched files; `pre-commit run mypy` clean on the new modules (pre-existing mypy errors elsewhere in the tree are not introduced by this PR). The reader is exercised end-to-end offline via a synthetic per-shard t-digest CSR store built with `build_tdigest` + `write_csr`.

Note: `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` is flaky under the full-suite run (fails intermittently, passes in isolation) — it is the Lambda build-script test, unrelated to this pure-Python change, and is left untouched.

## Questions for review

The plan's 4 open questions were implemented with these defaults — please confirm:

1. **Tensor values** — per-bin **counts** reconstructed from the digest CDF. dtype is a **flag** (`uint16` | `uint32` | `float32`), defaulting to `uint32`; per the thread note that the client currently consumes `uint16`, should the **default** be `uint16` instead?
2. **Window positioning** — fixed-width window **anchored at `z_lo = floor(min bottom-quantile)`**, span `n_bins*resolution`, `fit="raise"` default (vs a window centred on the trimmed range).
3. **Per-chunk morton id** — recovered from the store: sibling `{field}/morton` uint64 coord array if present, else the shard subgroup name (= parent morton id). Confirm the store-layout assumption (one CSR subgroup per shard, keyed by parent morton) matches the writer once #48's store integration lands — if the real layout differs the iteration will need adapting.
4. **Module home** — `zagg.readers` in core (no extra), per "readers in core is fine."

Also flagging two follow-ups from the thread: (a) the non-synthetic t-digest store check is not yet wired (needs a real product fixture); (b) `read_raw_values` is the raw-vector reader requested, with a `raise` on merged centroids — confirm the lossless-only semantics.

---

### DIFF
--- a/src/zagg/readers/tdigest_tensor.py
+++ b/src/zagg/readers/tdigest_tensor.py
@@ -171,15 +171,17 @@ def chunk_z_range(
     window = n_bins * resolution
 
     if fit == "collapse_bins":
-        # Shrink n_bins to the smallest power of two whose window still covers
-        # the trimmed range. Only ever reduces the bin count, so it cannot help
-        # a range that already exceeds the full n_bins window.
+        # Shrink to the smallest power-of-two bin count (≤ the largest power of
+        # two that fits within n_bins) whose window still covers the trimmed
+        # range. Only ever reduces the bin count, so it cannot help a range that
+        # already exceeds the full n_bins window.
         if needed > window:
             raise ValueError(
                 f'fit="collapse_bins" cannot grow the window: trimmed span {needed} '
                 f"exceeds {n_bins} bins × {resolution} = {window}"
             )
-        n = n_bins
+        # Largest power of two ≤ n_bins (the collapsed count is always pow2).
+        n = 1 << (int(n_bins).bit_length() - 1)
         while n // 2 >= 1 and (n // 2) * resolution >= needed:
             n //= 2
         return float(z_lo), n, resolution
@@ -209,6 +211,11 @@ def _resolve_chunk_morton(
     Prefers a sibling ``{field}/morton`` ``uint64`` coordinate array (chunk
     order → morton id); falls back to parsing the subgroup name as the parent
     morton id (the shard key is the parent morton — see :func:`process_shard`).
+
+    The coordinate array is aligned against the subgroup names sorted
+    **numerically** (the canonical ascending-morton chunk order), not
+    lexicographically — string sorting would mis-align names of differing digit
+    counts (e.g. ``"1000"`` before ``"99"``).
     """
     try:
         arr = zarr.open_array(store, path=f"{field}/morton", mode="r", zarr_format=zarr_format)
@@ -218,7 +225,7 @@ def _resolve_chunk_morton(
     if len(morton) != len(shard_keys):
         # Coordinate present but not 1:1 with subgroups — fall back to the names.
         return {k: int(k) for k in shard_keys}
-    return {k: int(m) for k, m in zip(sorted(shard_keys), morton)}
+    return {k: int(m) for k, m in zip(sorted(shard_keys, key=int), morton)}
 
 
 def read_tensors(
@@ -259,7 +266,9 @@ def read_tensors(
         (default ``"raise"``).
     dtype : {"uint16", "uint32", "float32"}, optional
         Output tensor dtype (default ``"uint32"``).  ``uint16``/``uint32`` round
-        counts to the nearest integer; ``float32`` keeps fractional counts.
+        counts to the nearest integer; ``float32`` keeps fractional counts.  A
+        per-bin count exceeding the dtype's max (65535 for ``uint16``) wraps on
+        cast — keep ``uint32`` for dense cells with many observations per bin.
     zarr_format : int, optional
         Zarr format version (default 3).
 
--- a/src/zagg/readers/tdigest_tensor.py
+++ b/src/zagg/readers/tdigest_tensor.py
@@ -1,0 +1,366 @@
+"""Client-side reader: per-cell t-digests → fixed ``(64, 64, n_bins)`` tensors.
+
+These are *read helpers* (issue #79) that live in-repo but sit just outside
+zagg's core write path: a client consuming a gridded Zarr product wants a dense,
+fixed-size tensor per coverage chunk, reconstructed from the per-cell t-digest
+ragged field that zagg writes (issue #48).
+
+Store layout consumed
+---------------------
+The t-digest field is stored per shard as a CSR ragged group (see
+:mod:`zagg.csr`): under the field prefix, one subgroup per shard, named by the
+shard's **parent morton id** (the coverage cell), each holding the three CSR
+arrays ``values`` / ``offsets`` / ``cell_ids``::
+
+    {field}/{parent_morton}/values
+    {field}/{parent_morton}/offsets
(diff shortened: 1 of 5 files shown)


src/zagg/readers/tdigest_tensor.py
```diff
@@ -0,0 +1,375 @@
...
+    """
+    try:
+        arr = zarr.open_array(store, path=f"{field}/morton", mode="r", zarr_format=zarr_format)
+        morton = np.asarray(arr[...])
+    except (FileNotFoundError, KeyError, ValueError):
+        return {k: int(k) for k in shard_keys}
+    if len(morton) != len(shard_keys):
+        # Coordinate present but not 1:1 with subgroups — fall back to the names.
+        return {k: int(k) for k in shard_keys}
+    return {k: int(m) for k, m in zip(sorted(shard_keys, key=int), morton)}
```
espg, line 228: 🤖 *from Claude (review)*

**Folded (was a correctness bug).** The first pass zipped the `{field}/morton` coord array against `sorted(shard_keys)` — lexicographic string sort. For subgroup names of differing digit counts that mis-aligns: keys `99/100/1000` sort to `['100','1000','99']`, so a coord `[900,901,902]` (ascending-morton chunk order) would mis-assign `100↦900, 1000↦901, 99↦902`. Now sorted `key=int` so the coord aligns with numeric chunk order; regression test `test_morton_coord_array_mixed_digit_keys` pins it. The plain-name fallback was already order-independent. Open item for review: confirm the writer emits `{field}/morton` in ascending-parent-morton order so this alignment holds.

---

src/zagg/readers/tdigest_tensor.py
```diff
@@ -0,0 +1,375 @@
...
+        # two that fits within n_bins) whose window still covers the trimmed
+        # range. Only ever reduces the bin count, so it cannot help a range that
+        # already exceeds the full n_bins window.
+        if needed > window:
+            raise ValueError(
+                f'fit="collapse_bins" cannot grow the window: trimmed span {needed} '
+                f"exceeds {n_bins} bins × {resolution} = {window}"
+            )
+        # Largest power of two ≤ n_bins (the collapsed count is always pow2).
+        n = 1 << (int(n_bins).bit_length() - 1)
```
espg, line 184: 🤖 *from Claude (review)*

**Folded (doc/contract mismatch).** The `collapse_bins` docstring promises "smallest power of two," but simply halving a non-power-of-two `n_bins` (e.g. 100→25, 127→31) doesn't yield one. Now the collapse starts from the largest power of two ≤ `n_bins` (`1 << (bit_length-1)`) and halves from there, so the result is always a power of two; `test_collapse_bins_pow2_for_non_pow2_n_bins` (n_bins=100) covers it. Note this changes the semantics slightly for non-pow2 `n_bins` (the *count* is rounded down to a power of two rather than preserved) — flag for confirmation, though default `n_bins=128` is unaffected.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of #81. One correctness bug (lexicographic morton-coord alignment) and one doc/contract mismatch (`collapse_bins` power-of-two) were found and **folded into this PR** with regression tests — see the two inline comments. Remaining notes (non-blocking, left standing): `cdf_from_tdigest`/`quantile_from_tdigest` are approximate (not exact) inverses with an O(1/n) round-trip offset, by design; and `uint16` output can wrap if a single bin's count exceeds 65535 (documented; `uint32` default avoids it). All 91 new tests pass.

---

PR #93, opened 2026-06-23
issue #92: key mortie shardmap MOC order to grid chunk_order

### LINKED ISSUE #92: mortie shardmap backend: order-8 MOC upsamples every granule onto every shard (OOM)
🤖 *from Claude*

## Summary

The `mortie` shardmap backend assigns **~every granule to ~every shard** whenever the grid's `parent_order` is finer than the MOC order (`mortie_order`, default **8**). On the NEON ATL03 AOI this produced shardmaps where **65 of 66 granules appear in every one of the 196 shards** — so each Lambda worker reads ~65 full granules' worth of in-AOI photons instead of the 1–6 its shard actually needs, and OOMs at the 2048 MB Lambda (`Runtime.OutOfMemory`).

## Root cause

`_intersect_mortie` (in `src/zagg/catalog/shardmap.py`) computes each granule footprint as a MOC at `order=mortie_order` (default 8), then `moc_to_order(moc, parent_order)` maps it to the order-`parent_order` shard cells. When `mortie_order < parent_order`, that call **upsamples**: every coarse MOC cell becomes all `4^(parent_order - mortie_order)` order-`parent_order` descendants — fattening a thin ground track to fill every shard under each coarse cell.

The NEON grid is `parent_order=13`, so order-8 cells each expand to `4^5 = 1024` shards. The whole ~0.1° AOI sits inside ~2 order-8 cells, so every granule grazing the AOI is assigned to all ~196 shards.

Measured directly against `mortie` on the AOI bbox ring:

```
AOI footprint -> order-13 shards covering AOI
  mortie order= 8: MOC cells=    2  -> order-13 shards = 2048   <-- degenerate
  mortie order=11: MOC cells=   13  -> order-13 shards =  352
  mortie order=13: MOC cells=   55  -> order-13 shards =  205
  mortie order=15: MOC cells=  262  -> order-13 shards =  205
  mortie order=17: MOC cells=15168  -> order-13 shards =  205
```

Confirmed by the shardmaps themselves (same AOI, different builders):

| shardmap | backend | granules | shards | granules/shard (mean) | granules in *every* shard |
|---|---|---|---|---|---|
| `neon_*_shardmap.json` | mortie (order 8) | 66 | 196 | **65.97** | **65** |
| `shardmap_atl03_serc_2025.json` | shapely | 10 | 400 | 2.78 | 0 |

The SERC map (built by the old shapely path) is healthy; the NEON maps (built by mortie at order 8) are degenerate.

## Fix

Stop defaulting `mortie_order` to a fixed 8. Pin the MOC order to the grid's **`child_order`**, clamped to mortie's order-18 `morton_coverage_moc` cap (`min(child_order, 18)`), and raise loudly if a requested/derived order is coarser than `parent_order` (which always upsamples). This resolves footprints at ~leaf resolution so a granule lands only in the shards its track actually crosses.

Note `morton_coverage_moc` caps at order 18 while `child_order` can be 19 — hence the `min(.., 18)` clamp; order 18 still comfortably exceeds typical `parent_order` (13 here).

## Scope / related

- Separate, milder over-coverage from the CMR swath polygon being a fat multi-beam ribbon is tracked in #65; this issue is only the MOC-order upsampling, which is far more severe.
- Existing degenerate NEON shardmaps need rebuilding after the fix lands.

espg:
Closes #92.

## What

The `mortie` shardmap backend computed each granule footprint as a MOC at a fixed `mortie_order=8`, then mapped it to the order-`parent_order` shard cells via `moc_to_order`. When `mortie_order < parent_order`, that call **upsamples**: each coarse cell becomes all `4^(parent_order - order)` shards beneath it, fattening a thin granule track onto every shard. On the NEON ATL03 AOI every order-8 cell expanded to many shards, so ~every granule landed in ~every shard and the Lambda workers OOMed (`Runtime.OutOfMemory`).

This pins the MOC order to the grid instead of a fixed coarse constant. Per @espg's review (https://github.com/englacial/zagg/pull/93#issuecomment-4783625863), the order is keyed to the grid's **inner-chunk order** (`grid.chunk_order`, set by `chunk_inner`, defaulting to `parent_order` when unset) — the unit work is dispatched at:

- `mortie_order` defaults to `None`, resolved by `_resolve_mortie_order(mortie_order, grid)` to `min(grid.chunk_order, MORTIE_MOC_ORDER_CAP)`. The shipped ATL03 HEALPix configs (`atl03_gain_bias_healpix.yaml`, `atl03_tdigest_healpix.yaml`) use `parent_order=11`, `chunk_inner=13`, `child_order=19`, so this resolves to **13** — comfortably above `parent_order=11` (no upsampling) and under mortie's order-18 coverage cap.
- A grid with `chunk_inner` unset has `chunk_order == parent_order`, so the MOC resolves to the bare shard order — the "else the shard order" branch of the directive.
- Any order (explicit or derived) coarser than `parent_order` still **raises** rather than silently fattening — a guard against this whole bug class. The cap clamp is applied first, then the guard, so a clamp landing at/under `parent_order` still raises.
- Non-HEALPix grids keep the legacy default of 8 (byte-identical).
- The resolved order is recorded in `metadata["mortie_order"]`.

This supersedes the earlier `child_order - 3` (= 16 for the production grid) default; see the benchmark below for why the coarser chunk-order default loses no footprint precision.

## Benchmark (per @espg, https://github.com/englacial/zagg/pull/93#issuecomment-4783435855 + https://github.com/englacial/zagg/pull/93#issuecomment-4783625863)

`benchmarks/mortie_order_sweep.py` sweeps `_intersect_mortie` over morton orders `[9, 12, 13, 15, 16, 17, 18]` on the NEON Maryland AOI against the production grid (`parent_order=11`, `chunk_inner=13`, `child_order=19`), reporting wall-time, granules selected, and mean granules/shard.

**Offline note:** this environment has no Earthdata/CMR credentials, so the real catalog fetch is not reachable. The granule set is a **synthetic but representative** catalog of thin (~90 m) RGT-style swaths over the AOI — the geometry that drove the #92 degeneracy. Absolute counts are representative, not measured from production granules; the *relative* cost across orders is the comparison @espg asked for. Re-run in a credentialed session against a real `Catalog` for production absolutes.

20 order-11 shards cover the AOI; 12 synthetic swaths:

| morton order | run time (median) | granules selected | granules/shard (mean) | guard |
|---|---|---|---|---|
| 9  | 2.8 ms  | 12 | 9.65 | **rejected** (< parent_order) |
| 12 | 3.2 ms  | 12 | 2.95 | ok |
| 13 | 3.2 ms  | 12 | 2.95 | ok |
| 15 | 4.2 ms  | 12 | 2.95 | ok |
| 16 | 6.8 ms  | 12 | 2.95 | ok |
| 17 | 10.9 ms | 12 | 2.95 | ok |
| 18 | 21.1 ms | 12 | 2.95 | ok |

Reading: **granules/shard is flat (2.95) for every order ≥ `parent_order`**, while wall-time grows monotonically (~7× from order 12 → 18). The shard footprints are order-11 cells, so a MOC resolved at 13 vs 16 vs 18 collapses to the *same* order-11 shards via `moc_to_order` — a finer MOC buys precision the shard grid can't see, at strictly more compute. Below `parent_order` (order 9) the footprint upsamples to 9.65 granules/shard (the #92 fattening) and the guard rejects it. So keying to `chunk_order = 13` gets the tight-footprint result at ~⅓ the wall-time of the old `child_order - 3 = 16` default, with identical granule placement.

## Tests

`tests/test_shardmap.py::TestMortieOrder`:
- `test_default_keys_to_chunk_order` — `chunk_inner=13` resolves the MOC to 13.
- `test_default_falls_back_to_parent_order` — `chunk_inner` unset → `chunk_order == parent_order`, MOC resolves to the bare shard order.
- `test_default_under_mortie_cap_at_leaf_order_19` — the shipped production grid resolves to 13, under mortie's cap.
- `test_coarse_order_rejected` — an explicit order below `parent_order` raises.
- `test_derived_order_clamped_to_cap` — `chunk_inner=19` (> cap, parent 15) clamps to 18, never an illegal order.
- `test_no_fattening_west_east_disjoint` — a west and an east granule occupy disjoint shard sets (the regression).
- `test_non_healpix_keeps_legacy_default` — rectilinear grids keep order 8.

`pytest tests/test_shardmap.py -q`: 25 passed, 1 skipped (spherely-fork SpatialIndex test). Full suite `pytest -q`: 767 passed, 1 skipped. `ruff check --select=E,F,W,I --ignore=E501` + `ruff format --check` clean on all touched files (incl. `benchmarks/mortie_order_sweep.py`).

## Follow-ups (not in this PR)

- The degenerate NEON shardmaps need rebuilding once this lands.
- Tighter-still footprint edges depend on espg/mortie#60 (coverage order cap) — moot at the chunk-order default, which sits well under the cap.
- Milder multi-beam swath over-coverage is tracked separately in #65.

## Questions for review

None outstanding — the offset judgment call from the prior revision is resolved by keying to `chunk_order`. Flagging only that the benchmark numbers are from a synthetic (offline) catalog as noted above; a credentialed re-run would confirm the production absolutes (the cross-order *trend* is geometry-driven and won't change).

### DIFF
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -83,20 +83,27 @@ def _resolve_mortie_order(mortie_order, grid) -> int:
     order resolves to 13. Keying the MOC to the chunk order matches the unit work
     is dispatched at: footprints resolve no finer than the chunk the worker reads,
     which is enough to keep ``moc_to_order`` from upsampling onto neighbor shards
-    (#92) at near-minimal compute -- a benchmark sweep showed granules/shard is
-    flat for every order >= ``parent_order`` while wall-time grows with order, so
-    a finer MOC buys precision the order-``parent_order`` shard cells can't see.
+    (#92) at near-minimal compute -- the order-sweep benchmark
+    (``benchmarks/mortie_order_sweep.py``) shows granules/shard flat for every
+    order >= ``parent_order`` while wall-time grows with order, so a finer MOC
+    buys precision the order-``parent_order`` shard cells can't see.
     The order is still clamped to ``MORTIE_MOC_ORDER_CAP`` (mortie's order-18
     coverage cap) before the ``parent_order`` guard, so an exotic ``chunk_order``
     past the cap can't make mortie raise into the swallowing ``except`` (silent
-    coverage loss). An explicit ``mortie_order`` is honored but still validated
-    against ``parent_order``. Non-HEALPix grids (no ``parent_order`` /
-    ``child_order``) keep the legacy default of 8.
+    coverage loss). The clamp comes *before* the guard, so a ``parent_order``
+    itself above the cap (the clamp then lands at 18 < ``parent_order``) still
+    trips the raise rather than passing an order coarser than the shards. An
+    explicit ``mortie_order`` is honored but still validated against
+    ``parent_order``. Non-HEALPix grids (no ``parent_order`` / ``child_order``)
+    keep the legacy default of 8.
     """
     is_healpix = hasattr(grid, "parent_order") and hasattr(grid, "child_order")
     if mortie_order is not None:
         order = int(mortie_order)
     elif is_healpix:
+        # ``chunk_order`` is the inner-chunk order on HealpixGrid (always set;
+        # == parent_order when chunk_inner is unset). The getattr default only
+        # covers a duck-typed grid that exposes parent/child but not chunk_order.
         chunk_order = getattr(grid, "chunk_order", grid.parent_order)
         order = min(int(chunk_order), MORTIE_MOC_ORDER_CAP)
     else:
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -35,14 +35,21 @@ from typing import Dict, List
 
 import numpy as np
 
-# Resolve the granule-footprint MOC this many HEALPix levels ABOVE the leaf
-# (``child_order``). A few levels coarser than the leaf keeps footprint edges
-# reasonably tight while staying well above the shard order (``parent_order``) --
-# so ``moc_to_order`` never upsamples a footprint onto every shard (#92) -- and
+# Resolve the granule-footprint MOC this many HEALPix levels *coarser than* the
+# leaf (``child_order``) -- i.e. subtracting N gives an order N levels above the
+# leaf. A few levels coarser than the leaf keeps footprint edges reasonably tight
+# while staying well above the shard order (``parent_order``) -- so
+# ``moc_to_order`` never upsamples a footprint onto every shard (#92) -- and
 # comfortably under mortie's order-18 coverage cap (child_order 19 -> order 16).
 # Once mortie lifts that cap (espg/mortie#60) this offset can shrink for tighter
 # edges.
 MOC_LEVELS_BELOW_LEAF = 3
+
+# Upper bound on the MOC order mortie's ``morton_coverage`` /
+# ``morton_coverage_moc`` accept; a higher order raises inside mortie. The
+# derived order is clamped to this so a large ``child_order`` can't push the MOC
+# order past the cap and silently lose coverage (#92).
+MORTIE_MOC_ORDER_CAP = 18
 
 # ── granule footprint helpers ────────────────────────────────────────────────
 
@@ -78,18 +85,22 @@ def _resolve_mortie_order(mortie_order, grid) -> int:
     fixed default of 8 against ``parent_order=13`` expanded each cell to 1024
     shards, putting ~every granule in ~every shard and OOMing the workers (#92).
 
-    ``None`` (the default) pins the order to ``child_order -
-    MOC_LEVELS_BELOW_LEAF`` so footprints resolve a few levels above the leaf --
-    tight enough to track the granule yet under mortie's order-18 coverage cap.
-    An explicit ``mortie_order`` is honored but still validated against
-    ``parent_order``. Non-HEALPix grids (no ``parent_order``/``child_order``)
-    keep the legacy default of 8.
+    ``None`` (the default) pins the order to ``min(child_order -
+    MOC_LEVELS_BELOW_LEAF, MORTIE_MOC_ORDER_CAP)`` so footprints resolve a few
+    levels above the leaf -- tight enough to track the granule yet never past
+    mortie's order-18 coverage cap (a large ``child_order`` would otherwise push
+    the order past the cap and make mortie raise, silently losing coverage).
+    The clamp is applied first, then the order is still validated against
+    ``parent_order`` -- so if the clamp ever lands at or below ``parent_order``
+    the raise still fires. An explicit ``mortie_order`` is honored but still
+    validated against ``parent_order``. Non-HEALPix grids (no
+    ``parent_order``/``child_order``) keep the legacy default of 8.
     """
     is_healpix = hasattr(grid, "parent_order") and hasattr(grid, "child_order")
     if mortie_order is not None:
         order = int(mortie_order)
     elif is_healpix:
-        order = int(grid.child_order) - MOC_LEVELS_BELOW_LEAF
+        order = min(int(grid.child_order) - MOC_LEVELS_BELOW_LEAF, MORTIE_MOC_ORDER_CAP)
     else:
         order = 8
     if is_healpix and order < grid.parent_order:
@@ -316,10 +327,12 @@ class ShardMap:
             mortie for HEALPix grids (non-HEALPix grids require spherely and
             raise an ``ImportError`` with an install pointer when it is absent).
         mortie_order : int, optional
-            MOC order for the mortie backend. ``None`` (default) pins it to the
-            grid's ``child_order`` (clamped to mortie's order-18 cap) so granule
-            footprints resolve at leaf resolution; a coarser order upsamples
-            every footprint onto all shards (#92). Must be >= ``parent_order``.
+            MOC order for the mortie backend. ``None`` (default) pins it to
+            ``min(child_order - MOC_LEVELS_BELOW_LEAF, 18)`` -- a few levels above
+            the leaf, never past mortie's order-18 coverage cap -- so footprints
+            resolve tight to the granule; a coarser order upsamples every
+            footprint onto all shards (#92). Raises if the resolved order is
+            coarser than ``parent_order``.
 
         Returns
         -------
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -35,19 +35,9 @@ from typing import Dict, List
 
 import numpy as np
 
-# Resolve the granule-footprint MOC this many HEALPix levels *coarser than* the
-# leaf (``child_order``) -- i.e. subtracting N gives an order N levels above the
-# leaf. A few levels coarser than the leaf keeps footprint edges reasonably tight
-# while staying well above the shard order (``parent_order``) -- so
-# ``moc_to_order`` never upsamples a footprint onto every shard (#92) -- and
-# comfortably under mortie's order-18 coverage cap (child_order 19 -> order 16).
-# Once mortie lifts that cap (espg/mortie#60) this offset can shrink for tighter
-# edges.
-MOC_LEVELS_BELOW_LEAF = 3
-
 # Upper bound on the MOC order mortie's ``morton_coverage`` /
 # ``morton_coverage_moc`` accept; a higher order raises inside mortie. The
-# derived order is clamped to this so a large ``child_order`` can't push the MOC
+# derived order is clamped to this so an exotic ``chunk_order`` can't push the MOC
 # order past the cap and silently lose coverage (#92).
 MORTIE_MOC_ORDER_CAP = 18
 
@@ -85,22 +75,30 @@ def _resolve_mortie_order(mortie_order, grid) -> int:
     fixed default of 8 against ``parent_order=13`` expanded each cell to 1024
     shards, putting ~every granule in ~every shard and OOMing the workers (#92).
 
-    ``None`` (the default) pins the order to ``min(child_order -
-    MOC_LEVELS_BELOW_LEAF, MORTIE_MOC_ORDER_CAP)`` so footprints resolve a few
-    levels above the leaf -- tight enough to track the granule yet never past
-    mortie's order-18 coverage cap (a large ``child_order`` would otherwise push
-    the order past the cap and make mortie raise, silently losing coverage).
-    The clamp is applied first, then the order is still validated against
-    ``parent_order`` -- so if the clamp ever lands at or below ``parent_order``
-    the raise still fires. An explicit ``mortie_order`` is honored but still
-    validated against ``parent_order``. Non-HEALPix grids (no
-    ``parent_order``/``child_order``) keep the legacy default of 8.
+    ``None`` (the default) pins the order to the grid's **inner-chunk order**
+    (``grid.chunk_order``) -- the Zarr-chunk order between the shard order
+    (``parent_order``) and the leaf (``child_order``), set by ``chunk_inner`` and
+    defaulting to ``parent_order`` when unset (so chunk == shard). The shipped
+    ATL03 HEALPix configs use ``chunk_inner=13`` (parent 11, child 19), so the
+    order resolves to 13. Keying the MOC to the chunk order matches the unit work
+    is dispatched at: footprints resolve no finer than the chunk the worker reads,
+    which is enough to keep ``moc_to_order`` from upsampling onto neighbor shards
+    (#92) at near-minimal compute -- a benchmark sweep showed granules/shard is
+    flat for every order >= ``parent_order`` while wall-time grows with order, so
+    a finer MOC buys precision the order-``parent_order`` shard cells can't see.
+    The order is still clamped to ``MORTIE_MOC_ORDER_CAP`` (mortie's order-18
+    coverage cap) before the ``parent_order`` guard, so an exotic ``chunk_order``
+    past the cap can't make mortie raise into the swallowing ``except`` (silent
+    coverage loss). An explicit ``mortie_order`` is honored but still validated
+    against ``parent_order``. Non-HEALPix grids (no ``parent_order`` /
+    ``child_order``) keep the legacy default of 8.
     """
     is_healpix = hasattr(grid, "parent_order") and hasattr(grid, "child_order")
     if mortie_order is not None:
         order = int(mortie_order)
     elif is_healpix:
-        order = min(int(grid.child_order) - MOC_LEVELS_BELOW_LEAF, MORTIE_MOC_ORDER_CAP)
+        chunk_order = getattr(grid, "chunk_order", grid.parent_order)
+        order = min(int(chunk_order), MORTIE_MOC_ORDER_CAP)
     else:
         order = 8
     if is_healpix and order < grid.parent_order:
@@ -327,11 +325,12 @@ class ShardMap:
             mortie for HEALPix grids (non-HEALPix grids require spherely and
             raise an ``ImportError`` with an install pointer when it is absent).
         mortie_order : int, optional
-            MOC order for the mortie backend. ``None`` (default) pins it to
-            ``min(child_order - MOC_LEVELS_BELOW_LEAF, 18)`` -- a few levels above
-            the leaf, never past mortie's order-18 coverage cap -- so footprints
-            resolve tight to the granule; a coarser order upsamples every
-            footprint onto all shards (#92). Raises if the resolved order is
+            MOC order for the mortie backend. ``None`` (default) pins it to the
+            grid's inner-chunk order ``grid.chunk_order`` (the ``chunk_inner``
+            order, defaulting to ``parent_order`` when unset), clamped to mortie's
+            order-18 coverage cap -- the dispatch chunk's own resolution, enough
+            to keep ``moc_to_order`` from upsampling a footprint onto neighbor
+            shards (#92) at near-minimal compute. Raises if the resolved order is
             coarser than ``parent_order``.
 
         Returns
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -24,6 +24,7 @@ shapely is no longer an intersection backend -- its WGS84 STRtree path had
 antimeridian/near-pole correctness bugs (#36). shapely remains a dependency for
 WKB decode (``sources.py``) and footprint geometry (``grids/``).
 """
+
 from __future__ import annotations
 
 import importlib
@@ -34,7 +35,17 @@ from typing import Dict, List
 
 import numpy as np
 
+# Resolve the granule-footprint MOC this many HEALPix levels ABOVE the leaf
+# (``child_order``). A few levels coarser than the leaf keeps footprint edges
+# reasonably tight while staying well above the shard order (``parent_order``) --
+# so ``moc_to_order`` never upsamples a footprint onto every shard (#92) -- and
+# comfortably under mortie's order-18 coverage cap (child_order 19 -> order 16).
+# Once mortie lifts that cap (espg/mortie#60) this offset can shrink for tighter
+# edges.
+MOC_LEVELS_BELOW_LEAF = 3
+
 # ── granule footprint helpers ────────────────────────────────────────────────
+
 
 def _to_spherely_polygon(lats, lons):
     """Build a closed sphere-aware polygon, or None on validation failure.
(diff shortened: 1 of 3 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of the core change (pin mortie MOC order to `child_order - 3`, raise on coarse orders). The central fix is sound: `moc_to_order(moc, parent_order)` upsamples iff `order < parent_order`, so resolving footprints at `child_order - 3` (which sits well above `parent_order`) and guarding `order < parent_order` directly kills the order-8-vs-order-13 fattening/OOM in #92. The guard boundary is correct (`>=` is the no-upsample set: `==` is identity, `>` coarsens granule cells without fattening). Tests cover the default offset, the order-19 leaf case, the explicit-coarse rejection, the west/east disjoint regression, and the non-HEALPix legacy default — a good set.

Findings below; none are hard diff-scoped *bugs that produce wrong output today*, but two are correctness-relevant and I'd want them resolved before this goes to `waiting`:

- **MED — the order-18 coverage cap is not actually guaranteed.** Issue #92's prescribed fix was `min(child_order, 18)` (an explicit clamp). This PR substitutes `child_order - 3`, which is *not* a clamp: for `child_order > 21` the derived order exceeds 18 and `morton_coverage_moc` will reject it. Worse, that rejection is swallowed by the bare `except Exception: continue` in `_intersect_mortie`, so every granule is silently dropped → empty shardmap → silent data loss rather than a loud failure. Today's grids (≤19) are fine, but the invariant the PR title/body claims ("under mortie's order-18 cap") isn't enforced. Consider `min(child_order - MOC_LEVELS_BELOW_LEAF, 18)` or an explicit cap check that raises. (Inline.)
- **MED — naming/comment contradiction.** The constant is `MOC_LEVELS_BELOW_LEAF` but the header comment says "this many HEALPix levels ABOVE the leaf", and the value is *subtracted* from `child_order` → a coarser (above-the-leaf) order. Name says BELOW, semantics are ABOVE. Pick one; this will confuse the next reader. (Inline.)
- **MED — `build` docstring is wrong on two counts.** It says `None` "pins it to the grid's `child_order` (clamped to mortie's order-18 cap) so granule footprints resolve at leaf resolution." It does neither: it subtracts 3 (not leaf resolution) and does not clamp. `_resolve_mortie_order`'s own docstring is accurate; the `build` one drifted. (Inline.)
- **LOW — missing test: a *derived* coarse order that raises.** `test_coarse_order_rejected` only exercises an explicit `mortie_order=8`. There's no test that a tight grid (e.g. `HealpixGrid(13, 14)` → derived `11 < 13`) trips the same guard via the `None` path. Cheap to add and it locks in the guard for the derived branch.

Everything else (the ruff-format churn — blank lines after the module docstring, `to_json`/`from_json` rewraps) is benign and matches `ruff format`. Module stays well under the ~1000-line limit; no new dependencies. The open `MOC_LEVELS_BELOW_LEAF = 3` question is correctly surfaced for @espg.

---

src/zagg/catalog/shardmap.py
```diff
@@ -57,6 +68,39 @@ def _to_spherely_polygon(lats, lons):
...
+    tight enough to track the granule yet under mortie's order-18 coverage cap.
+    An explicit ``mortie_order`` is honored but still validated against
+    ``parent_order``. Non-HEALPix grids (no ``parent_order``/``child_order``)
+    keep the legacy default of 8.
+    """
+    is_healpix = hasattr(grid, "parent_order") and hasattr(grid, "child_order")
+    if mortie_order is not None:
+        order = int(mortie_order)
+    elif is_healpix:
+        order = int(grid.child_order) - MOC_LEVELS_BELOW_LEAF
```

MED — coverage cap not enforced. `child_order - MOC_LEVELS_BELOW_LEAF` is not the `min(child_order, 18)` clamp #92 asked for. For `child_order > 21` this derives order > 18, which `morton_coverage_moc` rejects — and the rejection is swallowed by the bare `except Exception: continue` in `_intersect_mortie`, silently dropping every granule (empty shardmap, silent data loss) instead of failing loud. Today's grids (≤19 → ≤16) are fine, but the "under mortie's order-18 cap" invariant the PR claims isn't actually guaranteed. Suggest `min(int(grid.child_order) - MOC_LEVELS_BELOW_LEAF, 18)`, or an explicit `order > 18` raise.

---

src/zagg/catalog/shardmap.py
```diff
@@ -34,8 +35,18 @@
...
 import numpy as np
 
+# Resolve the granule-footprint MOC this many HEALPix levels ABOVE the leaf
+# (``child_order``). A few levels coarser than the leaf keeps footprint edges
+# reasonably tight while staying well above the shard order (``parent_order``) --
+# so ``moc_to_order`` never upsamples a footprint onto every shard (#92) -- and
+# comfortably under mortie's order-18 coverage cap (child_order 19 -> order 16).
+# Once mortie lifts that cap (espg/mortie#60) this offset can shrink for tighter
+# edges.
+MOC_LEVELS_BELOW_LEAF = 3
```

MED — name/semantics contradiction. The constant is `MOC_LEVELS_BELOW_LEAF`, but the header comment says "this many HEALPix levels ABOVE the leaf" and the value is *subtracted* from `child_order`, yielding a coarser (above-the-leaf) order. In HEALPix a smaller order number is coarser, i.e. above the leaf. Name says BELOW, behavior is ABOVE — rename to `MOC_LEVELS_ABOVE_LEAF` (or flip the comment) so the next reader isn't misled.

---

src/zagg/catalog/shardmap.py
```diff
@@ -270,8 +315,11 @@ def build(
             Geometry backend. ``"auto"`` -> spherely when importable, else
             mortie for HEALPix grids (non-HEALPix grids require spherely and
             raise an ``ImportError`` with an install pointer when it is absent).
-        mortie_order : int
-            MOC order for the mortie backend.
+        mortie_order : int, optional
+            MOC order for the mortie backend. ``None`` (default) pins it to the
+            grid's ``child_order`` (clamped to mortie's order-18 cap) so granule
+            footprints resolve at leaf resolution; a coarser order upsamples
```

MED — `build` docstring is inaccurate. It says `None` "pins it to the grid's `child_order` (clamped to mortie's order-18 cap) so granule footprints resolve at leaf resolution." Neither is true: the resolver subtracts `MOC_LEVELS_BELOW_LEAF` (so not leaf resolution) and does not clamp to 18 (it subtracts). `_resolve_mortie_order`'s own docstring is correct; this one drifted from the implementation. Bring it in line (e.g. "pins it to `child_order - MOC_LEVELS_BELOW_LEAF`, a few levels above the leaf").

---

tests/test_shardmap.py
```diff
@@ -245,6 +245,51 @@ def test_cli_rejects_shapely_backend(self, monkeypatch):
...
+        # child_order 17 - MOC_LEVELS_BELOW_LEAF (3) = 14.
+        sm = ShardMap.build(catalog, hp_grid, backend="mortie")
+        assert sm.metadata["mortie_order"] == 17 - MOC_LEVELS_BELOW_LEAF == 14
+
+    def test_default_under_mortie_cap_at_leaf_order_19(self, catalog):
+        # child_order 19 (the ~10 m leaf) -> order 16, under mortie's order-18 cap.
+        g = HealpixGrid(13, 19, layout="fullsphere")
+        sm = ShardMap.build(catalog, g, backend="mortie")
+        assert sm.metadata["mortie_order"] == 16
+
```
espg, line 282: 🤖 *from Claude (review)*

LOW — missing edge case. `test_coarse_order_rejected` only exercises an *explicit* `mortie_order=8`. There's no test that a *derived* coarse order trips the same guard via the `None` default — e.g. a tight grid like `HealpixGrid(13, 14)` derives `14 - 3 = 11 < 13` and should raise. Adding it locks the guard in for the `None`/derived branch, which is the path the fix actually changes.

---

espg (commented):
🤖 *from Claude (review)*

Inline findings for the #92 fix (see line comments). Summary of severities: MED — coverage cap not actually enforced (`child_order - 3` is not the `min(.., 18)` clamp #92 prescribed; a too-high derived order is silently swallowed by `except Exception: continue`); MED — `MOC_LEVELS_BELOW_LEAF` name contradicts its "above the leaf" comment/semantics; MED — `build` docstring claims leaf-resolution + clamp, neither of which the code does; LOW — no test for a *derived* coarse order tripping the guard. The core upsampling fix and guard boundary are correct; none of these produce wrong output on today's grids (≤ child_order 19).

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review findings (all diff-scoped, in `src/zagg/catalog/shardmap.py` + `tests/test_shardmap.py`):

- **(1) Enforce mortie's order-18 coverage cap.** Added `MORTIE_MOC_ORDER_CAP = 18` and changed the derived order in `_resolve_mortie_order` from `child_order - MOC_LEVELS_BELOW_LEAF` to `min(child_order - MOC_LEVELS_BELOW_LEAF, MORTIE_MOC_ORDER_CAP)`. Without the clamp a large `child_order` (>21) pushed the derived order past mortie's `morton_coverage`/`morton_coverage_moc` cap; mortie then raised and the bare `except Exception: continue` in `_intersect_mortie` swallowed it → empty/partial shardmap (silent data loss). Clamp is applied **before** the `parent_order` guard, so if `min(...)` ever lands at/under `parent_order` the existing raise still fires.
- **(2) Naming/comment contradiction.** The `MOC_LEVELS_BELOW_LEAF` comment said "ABOVE the leaf" while the value is *subtracted*. Reworded so name and comment agree: subtracting N gives an order N levels *coarser than* / above the leaf.
- **(3) `ShardMap.build` docstring.** It wrongly claimed `None` → `child_order` "clamped to mortie's order-18 cap … at leaf resolution". Now describes the real behavior: `None` → `min(child_order - MOC_LEVELS_BELOW_LEAF, 18)`, with the raise-if-coarser-than-`parent_order` guard.
- **(4) Missing tests.** Added `test_derived_coarse_order_rejected` (the `None` path itself trips the guard: `HealpixGrid(13, 14)` → derived 11 < 13 raises — the existing test only covered an explicit `mortie_order=8`) and `test_derived_order_clamped_to_cap` (`HealpixGrid(15, 22)` → would-be 19 clamps to 18, not an illegal order).

Tested: `uv run pytest tests/test_shardmap.py -q` → 25 passed, 1 skipped (spherely-fork SpatialIndex test). `ruff check --select=E,F,W,I --ignore=E501 src tests` and `ruff format --check` clean on both files.

The remaining `MOC_LEVELS_BELOW_LEAF = 3` offset is a design call left standing under "Questions for review"; the cap is now enforced in code regardless of the offset chosen.


---

espg:
> MOC_LEVELS_BELOW_LEAF = 3 is a judgment call (resolves to 16 for the order-19 leaf). Comfortable with that, or prefer a different offset / an explicit per-config knob? This is the one remaining design call on the PR.

The 'most correct' approach here is to set this to whatever the child order is (i.e., 19 for this example). I specified the MOC_LEVELS_BELOW_LEAF at 3 on the assumption that this is more computationally efficient, and has a low error of commission rate. 

We should test / benchmark this explicitly. While we can't do is for order 19 yet, we can do it for the other orders. Run the shardmap generation at orders [9,12,15,16,17,18] and time the execution at each morton order. Report back a table with the run time for each order as well as granules selected, and granules per shard.

espg:
> The 'most correct' approach here is to set this to whatever the child order is (i.e., 19 for this example). 

Actually, I think this is wrong for what we're doing-- we should set this to match inner_chunk order (if it exists, or else the shard order).

I still want the above benchmarks though; add in order 13 too (i.e., test / benchmark [9,12,13,15,16,17,18])

espg:
🤖 *from Claude*

Addressed both review directives — the MOC-order benchmark (https://github.com/englacial/zagg/pull/93#issuecomment-4783435855) and the switch to inner-chunk-order keying (https://github.com/englacial/zagg/pull/93#issuecomment-4783625863). Pushed as one commit (`key mortie MOC order to grid chunk_order for issue #92`).

### Inner-chunk-order investigation

The grid **does** have an inner-chunk order: `HealpixGrid.chunk_order`, set by the `chunk_inner` constructor arg and always in `[parent_order, child_order]`. It defaults to `parent_order` (chunk == shard) when `chunk_inner` is unset. The shipped ATL03 HEALPix configs (`atl03_gain_bias_healpix.yaml`, `atl03_tdigest_healpix.yaml`) declare `parent_order: 11`, `chunk_inner: 13`, `child_order: 19`, so the real production inner-chunk order is **13**.

So "inner_chunk order (if it exists, or else the shard order)" maps cleanly to `grid.chunk_order`, and resolves to **13** for the production workload — above `parent_order=11` (no upsampling, clears the guard) and under mortie's order-18 cap. No regression to a degenerate bare-shard-order case: even on a `chunk_inner`-unset grid the fallback is `parent_order` itself, which is the guard floor, not below it.

For reference, the old `child_order - 3` default resolved to 16 on that grid; the new chunk-order default is 13. The benchmark below shows that's a pure compute win with identical granule placement.

### Benchmark

`benchmarks/mortie_order_sweep.py` sweeps the requested orders `[9, 12, 13, 15, 16, 17, 18]` on the NEON Maryland AOI against the production grid (`parent_order=11`, `chunk_inner=13`, `child_order=19`).

**Honesty flag:** no Earthdata/CMR credentials in this environment, so the real catalog fetch isn't reachable. The granules are a **synthetic but representative** set of thin (~90 m) RGT-style swaths over the AOI — the geometry that drove #92. Absolute counts are representative, not measured from production granules; the *relative* cost across orders (what was asked) is what this isolates. A credentialed re-run against a real `Catalog` would give production absolutes — the cross-order trend is geometry-driven and won't change.

20 order-11 shards cover the AOI; 12 synthetic swaths; run times are medians of 7:

| morton order | run time | granules selected | granules/shard (mean) | guard |
|---|---|---|---|---|
| 9  | 2.8 ms  | 12 | 9.65 | rejected (< parent_order) |
| 12 | 3.2 ms  | 12 | 2.95 | ok |
| 13 | 3.2 ms  | 12 | 2.95 | ok |
| 15 | 4.2 ms  | 12 | 2.95 | ok |
| 16 | 6.8 ms  | 12 | 2.95 | ok |
| 17 | 10.9 ms | 12 | 2.95 | ok |
| 18 | 21.1 ms | 12 | 2.95 | ok |

Findings:
- **granules/shard is flat (2.95) for every order ≥ `parent_order`** — the shard cells are order-11, so a MOC at 13 vs 16 vs 18 collapses to the *same* order-11 shards via `moc_to_order`. A finer MOC buys precision the shard grid can't resolve.
- **wall-time grows monotonically with order** (~7× from 12 → 18). So a higher order is strictly more expensive with no placement benefit on this AOI — which is exactly the efficiency hypothesis behind keying to the (coarser) chunk order.
- **order 9 (< parent_order=11)** upsamples to 9.65 granules/shard — the #92 fattening — and the guard rejects it. (Orders 9 and below trip the guard; the sweep measures them directly to quantify the degeneracy.)

Net: `chunk_order = 13` lands the tight-footprint result at ~⅓ the wall-time of the old `child_order - 3 = 16` default, with byte-identical granule placement. Your directive is the better default on both axes.

### Implemented

- `_resolve_mortie_order(None, grid)` now returns `min(grid.chunk_order, MORTIE_MOC_ORDER_CAP)` (was `min(child_order - 3, 18)`); dropped the `MOC_LEVELS_BELOW_LEAF` constant. Cap clamp still precedes the `parent_order` guard; explicit-order and non-HEALPix paths unchanged; `metadata["mortie_order"]` records the resolved value.
- Tests in `TestMortieOrder` updated to the chunk-order behavior (default → chunk_order; chunk_inner-unset → parent_order; clamp-to-cap). Dropped `test_derived_coarse_order_rejected` — with chunk-order keying the derived order is `chunk_order ≥ parent_order` by grid construction, so the derived path can no longer trip the guard (the explicit-order path still does, and stays tested).
- Added `benchmarks/mortie_order_sweep.py` (reproducible, alongside the existing `region_timing.py` / `handoff_bias.py` benchmarks).

Status: one open question from the prior revision (the offset) is now resolved; nothing is blocked or waiting. Keeping the PR in draft. `pytest -q` 767 passed / 1 skipped; ruff + format clean.


---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of `3ffdab3` (keying the mortie MOC default to `grid.chunk_order`). The core change is sound — `chunk_order >= parent_order` is invariant in `HealpixGrid.__init__`, so the derived `min(chunk_order, 18)` clears the strict `<` guard in the normal case, and the `chunk_inner` unset case correctly resolves to `order == parent_order` (guard is `<`, not `<=`). Inline anchoring isn't available through the review tool here, so findings are collected below with file/line references.

**1. `getattr(grid, "chunk_order", grid.parent_order)` fallback is effectively dead for real grids — `src/zagg/catalog/shardmap.py:100`.** `HealpixGrid.__init__` (`src/zagg/grids/healpix.py:97`) sets `self.chunk_order` unconditionally, so every real `HealpixGrid` has the attribute. The `getattr` default is only reachable by a duck-typed object that has `parent_order` + `child_order` (enough to make `is_healpix` true at line 96) but lacks `chunk_order`. That's a defensive branch for a type that doesn't exist in-tree. Not a bug, but worth a decision: either drop the default to `grid.chunk_order` (matching how `parent_order`/`child_order` are accessed as plain attributes two lines up and in the guard at line 104), or keep it and say in a comment that it's guarding the duck-typed non-`HealpixGrid` case. As written it reads as if `HealpixGrid` might not define `chunk_order`, which is misleading.

**2. Removing `test_derived_coarse_order_rejected` drops the only coverage of the derived-path guard — `tests/test_shardmap.py` (deleted block).** Under the old `child_order - 3` default, the derived order could land below `parent_order` (the removed test's `HealpixGrid(13, 14)` → 11 < 13). Under the new keying that exact scenario is impossible, so removing *that* test is correct. But the derived path can *still* trip the `order < parent_order` guard — when `parent_order > MORTIE_MOC_ORDER_CAP (18)`: then `chunk_order >= parent_order > 18`, the clamp pins `order = 18`, and `18 < parent_order` raises. No test now exercises the derived path hitting the guard. Consider adding one (e.g. `HealpixGrid(19, 20, layout="fullsphere")` → derived 18, raises) so the guard-on-derived-path branch stays covered. This is the inverse edge of `test_derived_order_clamped_to_cap` (which only checks the clamp succeeding because parent_order 15 < 18).

**3. Docstring omits the `parent_order > cap` interaction — `src/zagg/catalog/shardmap.py:89-92`.** The docstring says the clamp runs "before the `parent_order` guard, so an exotic `chunk_order` past the cap can't make mortie raise into the swallowing `except`." True, but it frames the clamp purely as preventing a mortie-side raise; it doesn't mention that when `parent_order` itself exceeds the cap, the clamp drives `order` *below* `parent_order` and the guard then raises a `ValueError` on the default path. That's arguably correct behavior (a `parent_order > 18` grid is unsupported by mortie), but the prose currently implies the default path is always safe. One sentence noting "a grid with `parent_order > cap` raises the guard on the default path" would make the contract complete.

**4. `test_default_falls_back_to_parent_order` name/comment conflate two different fallbacks — `tests/test_shardmap.py:270-275`.** The comment says "chunk_inner unset -> chunk_order == parent_order, so the MOC order is the bare shard order." That's accurate as to *behavior*, but nothing in `_resolve_mortie_order` "falls back" here: `grid.chunk_order` genuinely equals 11, so the code reads the attribute directly — the `getattr` default (finding 1) is never exercised. The "falls back to parent_order" naming reads as if it tests the code's fallback path; it actually tests `HealpixGrid`'s own `chunk_inner=None → chunk_order=parent_order` default. Minor, but the name invites a future reader to think the `getattr` default has coverage (it doesn't).

**5. Production docstring cites a synthetic-data benchmark as evidence — `src/zagg/catalog/shardmap.py:86-88` vs `benchmarks/mortie_order_sweep.py:10-19`.** The docstring states "a benchmark sweep showed granules/shard is flat for every order >= `parent_order` while wall-time grows with order." The benchmark it refers to is candid (good) that its catalog is "synthetic but representative" and "absolute numbers are therefore representative, not measured." The docstring's unqualified "showed" promotes that synthetic result to a load-bearing justification for the design choice. Suggest softening to note the relationship was observed on a synthetic sweep (or run the credentialed real-catalog sweep the benchmark's OFFLINE NOTE describes and cite that). The benchmark file itself is honest and well-documented — this is only about how its result is quoted upstream.

None of these block the change; (1) and (4) are cosmetic/clarity, (2) is a real (if narrow) coverage gap, (3) and (5) are docstring-accuracy. All are diff-scoped — none imply a side-effecting directive.

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review findings (commit `fold #92 review: cover clamp-below-parent guard + docstring/comment`). All diff-scoped:

- **(substantive) Coverage gap on the derived guard path.** The review noted that with chunk-order keying the derived (`None`) path can *still* trip the `parent_order` guard in one case: when `parent_order` itself exceeds the order-18 cap, the clamp drives the order to 18 < `parent_order`. Added `test_derived_order_clamped_below_parent_rejected` (`HealpixGrid(19, 20)` → derived order clamps to 18 < parent 19 → raises), restoring coverage of that branch that the removed `test_derived_coarse_order_rejected` used to provide via a now-impossible scenario.
- **Docstring: clamp-vs-guard interaction.** `_resolve_mortie_order`'s docstring now spells out that the clamp comes *before* the guard, so a `parent_order` above the cap still trips the raise (it previously read as if the clamp always made the default safe).
- **Docstring: benchmark provenance.** Reworded the "a benchmark sweep showed…" line to point at `benchmarks/mortie_order_sweep.py` by name rather than citing it as unqualified load-bearing evidence (the script itself flags its catalog as synthetic/offline).
- **getattr fallback intent.** Added a comment that `grid.chunk_order` is always set on `HealpixGrid`, so the `getattr(..., grid.parent_order)` default only covers a duck-typed grid exposing parent/child but not chunk_order — i.e. it's defensive, not a live path.

The review's remaining note (test name conflating the two fallbacks) — the `test_default_falls_back_to_parent_order` test explicitly asserts `chunk_inner`-unset → `chunk_order == parent_order` (the real HealpixGrid default it exercises), so the name is accurate as written; left as-is.

`pytest -q`: 768 passed, 1 skipped. ruff + format clean. No side-effecting findings — nothing raised for separate decision.

Status: implemented cleanly, no open question — leaving `implement` (continuing), draft.

---

PR #94, opened 2026-06-23
Release per-granule h5coro cache in process_shard (issue #66)

### LINKED ISSUE #66: h5coro H5Coro objects leak ~67 MB/granule in process_shard (no close/eviction) -> Lambda OOM

espg:
Closes #66

## The leak

`process_shard` opens one `h5coro.H5Coro` per granule in its read loop and never releases it (`src/zagg/processing/worker.py` ~`:197`). h5coro caches reads in 4 MB lines with no eviction and leaves a `ThreadPoolExecutor` whose worker closures capture `self`, so each granule's HDF5-navigation cache stays resident for the whole loop even after `h5obj` is reassigned. Per the measurements in #66, the per-granule coarse geolocation read alone pins ~16 lines ≈ **67 MB**, and with the full-mission catalog ~40 granules are assigned per 64-cell shard → **~40 × 67 MB ≈ 2.7 GB > 2 GB Lambda → `Runtime.OutOfMemory`**, even for trivial data volumes (the pinned memory is HDF5 metadata, not photon data).

## Investigation — what's retained in `all_reads`, copy vs view

Traced the read path (`src/zagg/processing/read.py`: `_read_group` → `_planned_read_group` / `_read_group_full`). **Every array placed into the returned DataFrame/Table is a numpy boolean-mask copy, not a memoryview into h5coro's cache lines:**

- Base-rate variable columns: `values = arrays_by_path[path][mask_spatial]` (then optionally `[keep_mask]`) — boolean fancy-indexing, which **always copies** in numpy.
- `leaf_id`: `leaf_after_spatial[keep_mask]` (or `leaf_ids[mask_spatial]` when `keep_mask is None`) — boolean-index copy.
- In `_read_group_full`: `leaf_ids[min_idx:max_idx]` is a view, but the subsequent `[mask_sliced]` boolean index copies; same for the variable columns.
- Segment-level broadcasts (`_read_segment_broadcasts` / `_broadcast_segment_to_base`): freshly `np.full`/`np.zeros`-allocated, then boolean-sliced — never an h5coro view.
- Expression-filter path: `{k: v[emask] for ...}` — boolean-index copy.

`mask_spatial` / `mask_sliced` are always boolean arrays (results of `==`), so there is no slice-view-only retention path. On top of that, `pd.DataFrame(data_dict)` / `pa.table(data_dict)` also copy their numpy inputs at construction. **Conclusion: everything retained past a granule's lifetime is detached from the cache (a two-layer guarantee: boolean-mask copy AND construction-time copy)**, so freeing the cache after each granule cannot corrupt already-read data — copy-before-clear is satisfied by construction and needs no extra `np.array(..., copy=True)`.

## The fix

In `process_shard`'s granule loop, wrap each granule's read in `try/finally` and release the cache per granule (`src/zagg/processing/worker.py`):

```python
finally:
    if h5obj is not None:
        try:
            if hasattr(h5obj, "close"):
                h5obj.close()
            elif getattr(h5obj, "cache", None) is not None:
                h5obj.cache.clear()
        except Exception:
            logger.debug("h5coro cache release failed", exc_info=True)
    h5obj = None
```

- **Per granule, after extraction** — the release runs in the `finally` of each loop iteration, after the granule's data is fully copied into `all_reads`, not once at the end. `h5obj` is set to `None` at the top of each iteration, so a constructor failure leaves nothing to release (the `if h5obj is not None` guard skips it).
- **`close()` is the live path on the pinned h5coro 1.0.4.** 1.0.4 already ships `close()` (its `_cleanup` does `cache.clear()` + `cache_locks.clear()` + `driver.close()`), so the `hasattr(h5obj, "close")` branch is taken against the real dependency and genuinely frees the cache lines. The `cache.clear()` branch is a defensive fallback for any build lacking `close()`. The `hasattr` guard also stays correct when the pin later moves to h5coro 1.0.5 — no `pyproject.toml` change here.

## Tests (`tests/test_processing.py`, no network / no real S3 / no RSS sampling)

New `TestProcessShardCacheRelease`, mirroring the existing stub-H5Coro / `monkeypatch.setattr("zagg.processing.h5coro.H5Coro", ...)` pattern, driving the **real** `_read_group` → aggregate → `_build_output` path so the loop's release actually runs:

- `test_close_called_once_per_granule` — close-bearing stub over 3 granules; asserts `close()` recorded exactly 3× (once per granule, inside the loop — not zero, not once at the end). Because a fresh stub is minted per granule, a release moved outside the loop would close only the last object → 1 ≠ 3, so this also catches the "once at the end" regression.
- `test_cache_clear_fallback_on_1_0_4` — stub with no `close()` and a `cache` whose `clear()` is recorded; asserts the `hasattr` fallback fires once per granule.
- `test_retained_data_survives_cache_clear` — copy-before-clear regression guard. A stub hands out `np.frombuffer` views into private buffers and **zeroes every buffer on `close()`** (simulating cache-line eviction); asserts the worker's aggregation output reflects the **original** bytes (`h_min == 100.0`, `count == 2`), proving no retained array references the evicted cache. (This guards the system-level invariant; a DataFrame-returning read helper can't isolate the read-site copy layer because pandas copies at construction anyway — the test docstring says so.)

Existing `test_processing.py` (163) and `test_integration.py` stay byte-identical and green. Full suite: **763 passed, 1 skipped** (pre-existing).

## Lint / format / types

- `ruff check --select=E,F,W,I --ignore=E501 src tests` — clean.
- `ruff format --check` — my two changed files are clean. The format check also flags pre-existing drift in unrelated files (`src/zagg/catalog/shardmap.py`, `src/zagg/catalog/sources.py`, `src/zagg/runner.py`, `src/zagg/schema.py`, `tests/test_catalog.py`, `tests/test_integration.py`, `tests/test_rectilinear.py`, `tests/test_runner.py`) — left untouched per "don't fix unrelated CI failures"; flagging here.
- mypy: no new error categories (remaining errors are pre-existing import-stub/assignment/arg-type at untouched lines 142/209/284).

## Deployment note

Worker-side only (`src/zagg/processing/worker.py`); no `deployment/aws/` or `pyproject.toml` pin changes. Needs a Lambda **function redeploy** (`build_function.sh` + update) to take effect on Lambda — the maintainer's step.

## Questions for review

- On the pinned h5coro 1.0.4, `close()` frees the cache lines but does **not** shut down the leaked `ThreadPoolExecutor` (that gap is tracked upstream on `fix/resource-leak-1.0.5`, per the latest #66 comment). Bounding peak RSS by the freed cache lines is the dominant win here; the executor-thread pin is out of zagg's scope.
- `test_retained_data_survives_cache_clear` is a system-level (whole-worker) regression guard rather than a read-site copy isolation test — because `pd.DataFrame`/`pa.table` copy at construction, no DataFrame-returning test can distinguish a read-site view from a copy. Flagging the test's altitude in case a tighter read-internal assertion is wanted.

### DIFF
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -225,13 +225,16 @@ def process_shard(
             # ThreadPoolExecutor pinning ``self``, so without an explicit release
             # each granule's ~tens-of-MB HDF5-navigation cache stays resident for
             # the whole loop (~40 granules × ~67 MB ≈ 2.7 GB → Lambda OOM, even for
-            # trivial data volumes). Every array retained in ``all_reads`` is a
-            # boolean-mask copy made in ``_read_group`` (numpy boolean indexing
-            # always copies — never a memoryview into the cache lines), so freeing
-            # the cache here cannot corrupt already-read data. Prefer the upstream
-            # ``close()`` (h5coro 1.0.5's fix/resource-leak branch) and fall back to
-            # ``cache.clear()`` on 1.0.4; the ``hasattr`` guard keeps this
-            # forward-compatible without bumping the h5coro pin.
+            # trivial data volumes). Nothing retained in ``all_reads`` aliases the
+            # cache lines: ``_read_group`` builds every column by boolean indexing
+            # (a numpy copy) and returns ``pd.DataFrame(data_dict)`` (which copies
+            # its numpy inputs into managed blocks), so freeing the cache here
+            # cannot corrupt already-read data. The pinned h5coro 1.0.4 has
+            # ``close()`` (it does ``cache.clear()`` + ``cache_locks.clear()`` +
+            # ``driver.close()``), so that is the live path; the ``cache.clear()``
+            # branch is a defensive fallback for any build lacking ``close()``. The
+            # ``hasattr`` guard keeps this forward-compatible without bumping the
+            # h5coro pin.
             if h5obj is not None:
                 try:
                     if hasattr(h5obj, "close"):
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -191,6 +191,7 @@ def process_shard(
 
     # Read files and filter spatially
     for s3_url in granule_urls:
+        h5obj = None
         try:
             resource_path = _rewrite_url(s3_url)
 
@@ -218,6 +219,28 @@ def process_shard(
         except Exception as e:
             logger.warning(f"  Error processing file {s3_url}: {e}")
             continue
+        finally:
+            # Release this granule's h5coro cache before moving on (issue #66).
+            # h5coro caches reads in 4 MB lines with no eviction and leaves a
+            # ThreadPoolExecutor pinning ``self``, so without an explicit release
+            # each granule's ~tens-of-MB HDF5-navigation cache stays resident for
+            # the whole loop (~40 granules × ~67 MB ≈ 2.7 GB → Lambda OOM, even for
+            # trivial data volumes). Every array retained in ``all_reads`` is a
+            # boolean-mask copy made in ``_read_group`` (numpy boolean indexing
+            # always copies — never a memoryview into the cache lines), so freeing
+            # the cache here cannot corrupt already-read data. Prefer the upstream
+            # ``close()`` (h5coro 1.0.5's fix/resource-leak branch) and fall back to
+            # ``cache.clear()`` on 1.0.4; the ``hasattr`` guard keeps this
+            # forward-compatible without bumping the h5coro pin.
+            if h5obj is not None:
+                try:
+                    if hasattr(h5obj, "close"):
+                        h5obj.close()
+                    elif getattr(h5obj, "cache", None) is not None:
+                        h5obj.cache.clear()
+                except Exception:
+                    logger.debug("h5coro cache release failed", exc_info=True)
+            h5obj = None
 
     logger.info(f"  Processed {files_processed}/{len(granule_urls)} files")
     metadata["files_processed"] = files_processed
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -220,21 +220,11 @@ def process_shard(
             logger.warning(f"  Error processing file {s3_url}: {e}")
             continue
         finally:
-            # Release this granule's h5coro cache before moving on (issue #66).
-            # h5coro caches reads in 4 MB lines with no eviction and leaves a
-            # ThreadPoolExecutor pinning ``self``, so without an explicit release
-            # each granule's ~tens-of-MB HDF5-navigation cache stays resident for
-            # the whole loop (~40 granules × ~67 MB ≈ 2.7 GB → Lambda OOM, even for
-            # trivial data volumes). Nothing retained in ``all_reads`` aliases the
-            # cache lines: ``_read_group`` builds every column by boolean indexing
-            # (a numpy copy) and returns ``pd.DataFrame(data_dict)`` (which copies
-            # its numpy inputs into managed blocks), so freeing the cache here
-            # cannot corrupt already-read data. The pinned h5coro 1.0.4 has
-            # ``close()`` (it does ``cache.clear()`` + ``cache_locks.clear()`` +
-            # ``driver.close()``), so that is the live path; the ``cache.clear()``
-            # branch is a defensive fallback for any build lacking ``close()``. The
-            # ``hasattr`` guard keeps this forward-compatible without bumping the
-            # h5coro pin.
+            # Release this granule's h5coro cache before the next one (issue #66):
+            # without it each granule's unevicted cache stays resident for the whole
+            # loop → Lambda OOM. ``close()`` is the live path; ``cache.clear()`` is a
+            # fallback for builds lacking it. Retained ``all_reads`` data is already
+            # copied off the cache lines (see PR #94), so releasing here is safe.
             if h5obj is not None:
                 try:
                     if hasattr(h5obj, "close"):
@@ -243,7 +233,6 @@ def process_shard(
                         h5obj.cache.clear()
                 except Exception:
                     logger.debug("h5coro cache release failed", exc_info=True)
-            h5obj = None
 
     logger.info(f"  Processed {files_processed}/{len(granule_urls)} files")
     metadata["files_processed"] = files_processed
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -4218,12 +4218,20 @@ class TestProcessShardCacheRelease:
         assert len(clears) == 2
 
     def test_retained_data_survives_cache_clear(self, monkeypatch):
-        """Copy-before-clear correctness: the stub backs its returned arrays with
-        memoryviews into a buffer it ZEROES on ``close()``. If anything retained in
-        ``all_reads`` were a view into that buffer (rather than a numpy boolean-mask
-        copy), the aggregation output would be corrupted after the release. Assert
-        the output reflects the ORIGINAL bytes, proving the retained columns are
-        copies."""
+        """End-to-end copy-before-clear: drive the full ``process_shard`` loop with a
+        stub that hands out buffer-backed views (as real h5coro does — memoryviews
+        into 4 MB cache lines) and ZEROES every buffer on ``close()`` (the
+        per-granule release simulating cache-line eviction). Assert the aggregation
+        output reflects the ORIGINAL bytes — the worker's retained data is detached
+        from the cache before the release fires, so per-granule release is safe.
+
+        Note this guards the SYSTEM-LEVEL invariant (the worker's output survives the
+        release), which is what the fix relies on. It is a two-layer guarantee:
+        ``_read_group`` builds columns by boolean-mask indexing (a numpy copy) AND
+        ``pd.DataFrame``/``pa.table`` copy their numpy inputs at construction. A test
+        on a DataFrame-returning helper cannot isolate the read-site layer (pandas
+        copies regardless), so this asserts the property that actually matters: no
+        retained array references the evicted cache."""
 
         class _ViewBackedH5:
             """Returns memoryview-backed arrays into a private buffer per dataset;
@@ -4260,7 +4268,7 @@ class TestProcessShardCacheRelease:
         )
         # The single cell pooled both photons (h_li = 100, 200) BEFORE the buffer
         # was zeroed; h_min must be the original 100.0, not 0.0 (corrupted) and the
-        # count must be 2. A retained view would read back 0.0 here.
+        # count must be 2.
         assert meta["total_obs"] == 2
         assert df_out["count"].to_numpy()[0] == 2
         assert df_out["h_min"].to_numpy()[0] == np.float32(100.0)
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -4052,3 +4052,215 @@ class TestChunkPrecompute:
         assert not np.array_equal(
             df_chunk["offset"].to_numpy(), np.array([1.0, 101.0], dtype=np.float32)
         )
+
+
+# ---------------------------------------------------------------------------
+# Per-granule h5coro cache release in process_shard (issue #66)
+# ---------------------------------------------------------------------------
+
+
+class _ReleaseGrid:
+    """Grid stub for the #66 release tests: leaf id == row index, every row maps
+    to ``shard_key`` 0 (so the flat read keeps all rows), and the post-read
+    methods (``cells_of``/``children``/``chunk_coords``) collapse every row onto a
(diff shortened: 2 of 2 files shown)


espg:
🤖 *from Claude*

Status: single phase, complete and green on `a25012f`. CI — **Tests ✅, Lint ✅, Docs ✅**; **Lambda Build** is in progress and passed on the prior commit `63fbde7` (it is the dual-arch size-gate workflow, unrelated to this worker-side logic change).

Self-review folded in (`a25012f`): corrected the code comment — the pinned h5coro **1.0.4 already ships `close()`** (its `_cleanup` does `cache.clear()` + `cache_locks.clear()` + `driver.close()`), so `close()` is the live path and `cache.clear()` is the defensive fallback; and recharacterized `test_retained_data_survives_cache_clear` honestly as a system-level (whole-worker) regression guard, since `pd.DataFrame`/`pa.table` copy at construction and no DataFrame-returning test can isolate the read-site copy layer.

Two standing items for review (also under "Questions for review" in the PR body): (1) on 1.0.4, `close()` frees the cache lines but does not shut down the leaked `ThreadPoolExecutor` (tracked upstream on `fix/resource-leak-1.0.5`); (2) the correctness test's altitude. Left as **draft** for the orchestrator to un-draft on green.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the per-granule h5coro cache release (Closes #66). Overall this is correct and well-scoped — recommend addressing the two MED items (error-path test coverage) before flipping to `waiting`; the LOW items are optional polish. No blocking diff-scoped bugs found.

**Verified correct (no change needed):**
- **Per-iteration, not once-at-end.** The `try/finally` is inside the `for s3_url in granule_urls:` body and `h5obj` is reset at the top of each iteration, so the release fires once per granule. `test_close_called_once_per_granule` mints a fresh stub per granule and asserts `== 3`, which genuinely catches a "moved outside the loop" regression (would be `1`).
- **`close()` is the live path on the pinned 1.0.4.** Confirmed against the #66 thread (issue comment 4776518439): 1.0.4 ships `close()` → `_cleanup` → `cache.clear()`. The original "no close/eviction" text in the issue body was stale; the PR body's framing is the corrected one. The `cache.clear()` branch is a genuine defensive fallback, and the `hasattr` guard keeps the pin untouched.
- **Copy-before-clear holds.** Traced `src/zagg/processing/read.py`: in `_read_group_full` (the flat path the tests drive) every column is `data[path][mask_sliced]` (boolean fancy-index → numpy copy), `leaf_id` is `leaf_sliced[keep_mask]` (copy), and the carrier is `pd.DataFrame(data_dict)` / `pa.table(data_dict)` (copies at construction). `_planned_read_group` is the same shape (`arrays_by_path[path][mask_spatial]`). Nothing retained in `all_reads` aliases a cache line, so the release can't corrupt read data.
- **No `deployment/aws/` or `pyproject.toml` changes.** Confirmed — diff touches only `src/zagg/processing/worker.py` and `tests/test_processing.py`. The PR's "no pin change" claim is accurate.
- **Monkeypatch path resolves.** `zagg/processing/__init__.py` binds `import h5coro`, so `monkeypatch.setattr("zagg.processing.h5coro.H5Coro", ...)` patches the exact symbol the worker calls via `_processing.h5coro.H5Coro`.

**Scope note (non-blocking, already flagged in the PR body):** `close()` on 1.0.4 frees the cache lines but does not shut down the leaked `ThreadPoolExecutor` that pins `self` (per #66 comments 4776348693 / 4776518439). Freeing the 4 MB lines is the dominant RSS win, so this is the right zagg-side fix; the executor shutdown is correctly left upstream. Good to keep it called out under "Questions for review."

See inline comments for the two test-coverage gaps (MED) and two polish items (LOW).

---

src/zagg/processing/worker.py
```diff
@@ -218,6 +219,31 @@ def process_shard(
...
+            # h5coro pin.
+            if h5obj is not None:
+                try:
+                    if hasattr(h5obj, "close"):
+                        h5obj.close()
+                    elif getattr(h5obj, "cache", None) is not None:
+                        h5obj.cache.clear()
+                except Exception:
+                    logger.debug("h5coro cache release failed", exc_info=True)
+            h5obj = None
```

LOW — dead assignment. This trailing `h5obj = None` at the end of the `finally` is redundant: the next iteration sets `h5obj = None` at the top of the loop body, and after the last granule `h5obj` is never read again. §4 ("no dead code") — drop this line; the loop-top reset is the one that does the work.

---

src/zagg/processing/worker.py
```diff
@@ -218,6 +219,31 @@ def process_shard(
...
+            # trivial data volumes). Nothing retained in ``all_reads`` aliases the
+            # cache lines: ``_read_group`` builds every column by boolean indexing
+            # (a numpy copy) and returns ``pd.DataFrame(data_dict)`` (which copies
+            # its numpy inputs into managed blocks), so freeing the cache here
+            # cannot corrupt already-read data. The pinned h5coro 1.0.4 has
+            # ``close()`` (it does ``cache.clear()`` + ``cache_locks.clear()`` +
+            # ``driver.close()``), so that is the live path; the ``cache.clear()``
+            # branch is a defensive fallback for any build lacking ``close()``. The
+            # ``hasattr`` guard keeps this forward-compatible without bumping the
+            # h5coro pin.
```

LOW — comment length vs §4 terseness. This ~18-line rationale block duplicates the PR description nearly verbatim (the 2.7 GB math, the copy-before-clear two-layer argument, the 1.0.4 `close()` provenance). §4 favors terse, reviewable code and §2 says "the PR description is where the substance lives." Consider trimming the inline comment to the 2-3 lines a future reader needs at the call site (what is freed, and why it's safe) and letting the PR body / #66 carry the full derivation. Non-blocking.

---

tests/test_processing.py
```diff
@@ -4052,3 +4052,223 @@ def test_absent_block_matches_plain_path_values(self, monkeypatch):
...
+    out = {}
+    for d in datasets:
+        if isinstance(d, str):
+            out[d] = arrays[d]
+            continue
+        path = d["dataset"]
+        arr = arrays[path]
+        hs = d.get("hyperslice")
+        if hs is not None:
+            lo, hi = hs[0]
```
espg, line 4119: 🤖 *from Claude (review)*

MED — untested error paths that the fix explicitly claims behavior on. The PR body asserts two error-path behaviors that no test exercises:

1. **Constructor failure → nothing to release.** The `h5obj = None` at the loop top is precisely what makes `if h5obj is not None:` skip release when `H5Coro(...)` raises. A test with a factory that raises (over multiple granules) asserting `files_processed == 0` and that no `close`/`clear` was logged would lock this in — and would catch a future refactor that moves the `None` init or releases unconditionally.

2. **Exception during read → release still fires in `finally`.** The whole point of the `try/finally` (vs a plain post-read `close()`) is that the cache is freed even when `_read_group` raises. A stub whose `readDatasets` raises, asserting `close()` was still recorded once per granule, would prove the `finally` actually buys you something over a straight-line close.

§4 requires tests for behavioral changes; right now both happy-path branches (`close` / `cache.clear`) are covered but neither failure branch is. Non-blocking, but worth adding before this goes to `waiting`.

---

tests/test_processing.py
```diff
@@ -4052,3 +4052,223 @@ def test_absent_block_matches_plain_path_values(self, monkeypatch):
...
+        )
+        clears = [e for e in log if e[0] == "clear"]
+        assert len(clears) == 2
+
+    def test_retained_data_survives_cache_clear(self, monkeypatch):
+        """End-to-end copy-before-clear: drive the full ``process_shard`` loop with a
+        stub that hands out buffer-backed views (as real h5coro does — memoryviews
+        into 4 MB cache lines) and ZEROES every buffer on ``close()`` (the
+        per-granule release simulating cache-line eviction). Assert the aggregation
+        output reflects the ORIGINAL bytes — the worker's retained data is detached
```
espg, line 4225: 🤖 *from Claude (review)*

LOW (verify) — `_ViewBackedH5.readDatasets` builds arrays with `np.frombuffer(buf, ...)` where `buf` is a `bytearray`. `np.frombuffer` on a mutable `bytearray` yields a writable view, so the later `buf[i] = 0` corruption is observable through any surviving alias — good, the test's premise is sound. Just flagging to confirm: if a future refactor switched `buf` to `bytes`, `np.frombuffer` would return a read-only array and the "zero the buffer" step could raise / silently no-op, weakening the guard. A one-line assert that the pre-close view reads `100.0` (so the corruption step is known to actually flip live bytes) would make the test self-checking. Optional.

---

espg (commented):
🤖 *from Claude (review)*

Inline findings for the per-granule cache release (#66) — 2 MED (error-path test coverage), 2 LOW (polish). See line comments. No blocking diff-scoped bugs; the core fix is correct (per-iteration release, sound copy-before-clear, `close()` confirmed on the pinned h5coro 1.0.4).

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review findings (all diff-scoped) in commit `fe6458f`:

- **MED — constructor-failure test** (`test_constructor_failure_releases_nothing_and_others_proceed`): a stub whose `H5Coro(...)` raises for the 2nd of 3 granules. Asserts the loop-top `h5obj = None` guard means the `finally` releases nothing for the failed granule (no spurious `close()` on `None`), the granule is skipped via the outer `except → continue`, and the other two still read end-to-end (`close` recorded 2×, `files_processed == 2`, `total_obs == 4`). This matches the worker's actual behavior — a constructor failure is swallowed-and-skipped, not propagated.
- **MED — exception-during-read test** (`test_read_exception_still_releases_in_finally`): a stub that constructs fine but raises in `readDatasets` *after* construction. Asserts the `finally` still releases that granule's cache exactly once (`close` recorded 1×), with the shard falling through to the `"No data after filtering"` path. This is the case that motivates `try/finally` over a plain post-read `close()`.
- **LOW — removed the dead trailing `h5obj = None`** in the `finally`: it was a genuinely dead assignment, since the loop-top `h5obj = None` already resets it for the next iteration.
- **LOW — trimmed the ~18-line inline rationale block** down to a terse 3-line note (per §2/§4 — the full rationale lives in the PR description above).

No worker behavior change beyond the dead-assignment removal; the tests assert the real behavior.

**Tested:** `pytest tests/test_processing.py -q` → 161 passed (3 prior cache-release tests + the 2 new ones all green). `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on both touched files.

All phases plus self-review folds are complete; setting `waiting` for review.

---

Issue #60, opened 2026-06-18
Revisit `plan_read` mortie linestring vs shapely after mortie v1 (#57)


espg:
## Summary

Tracking follow-up from #53 (closed with Option A: keep shapely `LineString` in `plan_read` unconditionally for now).

When `mortie` v1 ships (#57), revisit whether `plan_read`'s within-granule segment->AOI crossing check should branch on grid type and call into mortie's linestring coverage for HEALPix grids — for end-to-end S2 consistency with mortie-driven shard dispatch.

## Why deferred (per #53 analysis)

- **Euclidean error is negligible at ATL03 photon spacing.** ~10 m per 100 km of segment, well inside any usable HEALPix cell — the within-granule check is a crossing filter, not a precision geometry op.
- **Avoid premature binding to a moving API.** `mortie` is mid-redesign for v1 (`morton_index` datatype per espg/mortie#35, etc.); the linestring coverage interface may shift. Binding `plan_read` now risks rewriting it once v1 lands.

## When to revisit

After **#57** (mortie v1 integration) resolves and the v1 linestring API surface is stable. At that point evaluate:

1. **mortie v1 linestring coverage signature** — what does the call look like; does it match `plan_read`'s segment-rep-point input shape (`(lat, lon)` pairs)?
2. **Grid-type branch in `plan_read`** — is the right seam at `plan_read(...)` (passing a grid handle), or in `_planned_read_group` (computing the crossing mask before calling `plan_read`)?
3. **Measurable benefit on a real region** — does the S2 path actually change which segments match the AOI for a 10 km × 10 km HEALPix shard, vs the current Euclidean path? If parity is byte-identical at ATL03 cell scales, the consistency argument alone may not justify the dependency.

## Refs

- Original discussion: #53
- Phase C `plan_read` implementation: #49 (and PR #47's phase 5 wiring it into `_read_group`)
- Mortie v1 integration: #57
- `plan_read` source: `src/zagg/read_plan.py`

PR #96, opened 2026-06-23
issue #95: replace shapely per-segment AOI loop with mortie segment->shard mask

### LINKED ISSUE #53: read_plan: select mortie linestring vs shapely by grid type (HEALPix -> mortie)
Follow-up to #43 / PR #49. `plan_read` (`src/zagg/read_plan.py`) does the within-granule segment->AOI crossing check with `shapely.LineString` (euclidean, confirmed fine at 20 m ATL03 spacing). It currently uses shapely unconditionally, with no grid-type awareness.

Open question from the PR thread: make the within-granule check grid-aware — when the output grid is HEALPix/mortie, use mortie linestring coverage instead of shapely.

Decide whether the two concerns stay separate (mortie for shard dispatch upstream, shapely within-granule) or whether `plan_read` should branch on grid type and call into mortie for the HEALPix case.

### LINKED ISSUE #95: plan_read: shapely per-segment AOI loop dominates shard runtime; use mortie segment->shard mask
🤖 *from Claude*

## Summary

The issue-#43 read pre-filter `plan_read` decides which segment byte-ranges to fetch with a **Python loop over every coarse segment** that builds a shapely `LineString` and calls `.intersects()` against the shard's **bbox** (`read_plan.py:104-118`). At ATL03 segment counts this dominates per-shard runtime and pushes the heavy shards toward the 900 s Lambda timeout.

## Measurement

One beam of one real ATL03 granule (worst-case NEON shard 5347404518843744269):

```
segments in ONE beam (n_coarse): 181,024
(a) shapely plan_read : 1069.4 ms   ← per beam
(b) mortie assign     :    3.8 ms   ← grid.assign + grid.shards_of
speedup: 279x
projected shapely cost, 47 granules × 6 beams: ~302 s
```

A full worst-shard run sat at 99% CPU for 9+ minutes locally (fast Mac) without finishing — the ~300 s AOI-match is region-independent, so it's at least as bad on Lambda's slower cores.

## Why it's wrong

The **exact** shard membership test is already mortie, 50 lines later at `read.py:401`:

```python
leaf_ids = grid.assign(lats, lons)
mask_spatial = grid.shards_of(leaf_ids) == shard_key   # photon level, vectorized
```

`plan_read` reinvents that test on segments with a per-element shapely loop against a **loose bbox** (the bounding rectangle of the HEALPix shard, wider than the shard). The segment lat/lons are already in memory when `plan_read` runs, so the coarse filter should be the same `grid.assign`/`grid.shards_of` call — vectorized, ~280x faster, and **tighter** (exact HEALPix cell, not a bbox envelope), so it also fetches fewer wasted segments and is less likely to trip the `full_read` fallback.

## Fix

- `plan_read`: accept a precomputed boolean `coarse_mask` (segment→shard, from the grid) and use it directly instead of the shapely/bbox loop; vectorize the contiguous-run detection. Keep the bbox path as a grid-free fallback so the existing offline `plan_read` tests are unchanged.
- `_planned_read_group`: compute `coarse_mask = grid.shards_of(grid.assign(seg_lat, seg_lon)) == shard_key` and pass it — removing shapely from the production read path (and the per-group `shard_footprint`/bbox + wide-bbox antimeridian bail, which mortie handles correctly).

Expected: the ~300 s AOI-match collapses to <1 s/shard; heavy shards stop approaching the timeout.

## Secondary (separate)

The S3 driver builds a fresh `boto3.Session()` + client **per granule** (h5coro `s3driver.__init__`) — ~0.2-0.28 s each × 47 ≈ 10 s of botocore model-loading per shard, plus 47x duplicated service-model memory. Worth a follow-up (likely upstream in h5coro, or pass a shared session).

Refs: #43 (the read-plan this fixes), #92 (the shardmap-order fix that exposed this by making per-shard granule counts realistic).

espg:
Closes #95.

## What

`plan_read` (the issue-#43 read pre-filter) decided which segment byte-ranges to fetch with a **Python loop over every coarse segment** that built a shapely `LineString` and called `.intersects()` against the shard's **bbox**. On a 181k-segment ATL03 beam that's ~1.07 s/beam — ~300 s of pure, region-independent CPU for a 47-granule × 6-beam shard, the dominant slice of the 900 s Lambda timeout.

This replaces the segment match with the **same mortie test the photon path already uses** (`grid.shards_of(grid.assign(seg_lat, seg_lon)) == shard_key`, `read.py:401`): exact to the leaf cell, vectorized, ~280× faster.

```
one ATL03 beam, 181,024 segments:
  shapely plan_read : 1069.4 ms
  mortie assign     :    3.8 ms   (279x)
```

## Changes

- **`read_plan.plan_read`**: new optional `coarse_mask` (boolean, one entry per segment). When given it's used directly — no shapely, no bbox. The contiguous-run detection is now vectorized (`np.diff` on the sentinel-padded mask) instead of a per-segment Python loop. The bbox/linestring path is kept as a **grid-free fallback** so the offline `plan_read` unit tests are unchanged; `bbox` is now optional and the function raises if neither `coarse_mask` nor `bbox` is given.
- **`processing._planned_read_group`**: computes `coarse_mask = grid.shards_of(grid.assign(seg_lat, seg_lon)) == shard_key` and passes it. Drops the per-group `shard_footprint`→bbox derivation and the wide-bbox antimeridian/polar bail — mortie `assign` is globally correct, and a shard that genuinely owns most segments is still caught by `plan_read`'s selectivity `full_read` fallback.

## Errors of omission — explicit

The mask is **rep-point** based, so a boundary segment whose ~20 m of photons straddle the shard edge but whose rep-point lands in a neighbor could be skipped. This is **one-directional** (the exact photon-level filter at `read.py:401` still runs, so we never *over*-include — only risk *under*-reading) and **bounded by `pad`**: the segments with ≥1 in-shard photon form a contiguous along-track run; a segment with in-shard photons but out-of-shard rep-point is within ~one segment of the edge and is recovered by `pad` (config `pad: 1`). Recovery holds only when the owning segment is within `pad` of a matched rep-point segment — a stray photon ≥2 segments from any match is *not* recovered at `pad=1` (pinned by `test_pad_does_not_recover_segment_two_runs_away`); at dense ATL03 segment spacing that is effectively unreachable. The residual is a track skimming nearly parallel to a cell edge — a few edge photons, negligible at shard scale; `pad: 2` is cheap insurance. The old bbox path was a looser superset (marginally lower omission) but ~280× slower and itself leaned on `pad`. `pad` is the omission knob.

## Tests

- `tests/test_read_plan.py::TestPlanReadCoarseMask` (7): mask matches the equivalent bbox plan; vectorized multi-run / edge-touching detection; pad; precedence over bbox; shape-mismatch and neither-arg errors. Existing bbox tests unchanged (fallback still covered).
- `tests/test_processing.py`: the `_planned_read_group` tests now drive segment selection through a lat-mask grid (the new production path). Added `test_pad_recovers_boundary_segment_and_matches_full` — the omission guard: a boundary photon whose segment rep-point is out of band is dropped at `pad=0` and **recovered at `pad=1`, matching the full read exactly**. Added `test_pad_does_not_recover_segment_two_runs_away` — pins the upper edge of that bound: a stray photon two segments from the match stays omitted at `pad=1`.

`pytest -q`: 766 passed, 1 skipped. The one failure (`test_lambda_build.py` — build script can't resolve `zarr>=3.1.5` under Python 3.10) is pre-existing and unrelated (not touched, §4). `ruff check` + `ruff format --check` clean.

## Scope / follow-ups (not here)

- The residual heavy-shard cost is **#65**: a 47-granule shard still reads each granule's full segment-coordinate arrays even when the granule contributes **0 photons** (doesn't actually cross the shard) — tighter granule→shard assignment skips them. Measured: most granules on the worst shard return 0 photons.
- The S3 driver builds a fresh boto3 client per granule (~10 s/shard of botocore model-loading) — separate, likely upstream in h5coro.

Refs #43 (the read-plan this fixes), #92 (the shardmap-order fix that made per-shard granule counts realistic and surfaced this).

Closes #53 — this PR moves the HEALPix segment→shard match to mortie (`grid.shards_of(grid.assign(...))`), the core question raised in #53 (use exact mortie cell membership rather than a bbox envelope for the coarse filter). Per the go-ahead on the #53 thread, merging this closes it; the residual "should the grid-free fallback path also be mortie-aware" stays tracked in #60.

## Questions for review

- Dropping the wide-bbox antimeridian bail: I'm relying on mortie `assign` being globally correct (it is) plus the selectivity `full_read` fallback for genuinely-wide shards. Comfortable with that, or keep an explicit guard?

### DIFF
--- a/src/zagg/processing/read.py
+++ b/src/zagg/processing/read.py
@@ -271,8 +271,10 @@ def _planned_read_group(
 
     Issue #43 Phase C: when ``data_source.read_plan.spatial_index`` names a coarse
     level whose ``link`` points at the base level, we read the coarse coordinates
-    + link arrays once (small), call :func:`zagg.read_plan.plan_read` to compute
-    which base-rate slices the AOI bbox actually touches, and read base-rate
+    + link arrays once (small), call :func:`zagg.read_plan.plan_read` with the
+    mortie segment->shard mask (``grid.shards_of(grid.assign(...)) == shard_key``,
+    the same exact test the photon path applies) to compute which base-rate
+    slices the shard actually touches, and read base-rate
     coords + variables + filter datasets only over those slices via
     :func:`zagg.read_plan.execute_read_plan`. This avoids the
     ``lat_ph`` + ``lon_ph`` full-coord read (up to ~245 MB per ATL03 beam) that
@@ -344,7 +346,7 @@ def _planned_read_group(
     # shard edge is recovered by ``pad`` (and the photon-level filter below never
     # over-includes); residual omission is bounded to a few edge photons (#95).
     coarse_leaf = grid.assign(np.asarray(coarse_lats), np.asarray(coarse_lons))
-    coarse_mask = grid.shards_of(coarse_leaf) == int(shard_key)
+    coarse_mask = grid.shards_of(coarse_leaf) == shard_key
 
     plan = plan_read(
         np.asarray(coarse_lats),
@@ -526,9 +528,10 @@ def _read_group(h5obj, group: str, data_source: dict, shard_key: int, grid, arro
     filters are unchanged.
 
     *Hierarchical (planned) read* (``read_plan.spatial_index`` set, in addition
-    to ``levels``/``base_level``): the AOI bbox is computed from the grid's
-    shard footprint, the coarse-level spatial-index coordinates are read fully
-    (cheap), and base-rate coords + variables + filter datasets are read only
+    to ``levels``/``base_level``): the coarse-level spatial-index coordinates
+    are read fully (cheap), matched to the shard with the mortie segment->shard
+    mask (``grid.shards_of(grid.assign(...)) == shard_key``), and base-rate
+    coords + variables + filter datasets are read only
     over the planned hyperslices via :func:`zagg.read_plan.execute_read_plan`.
     Empty-AOI groups short-circuit to ``None``. Selectivity above the configured
     threshold falls back to the full-read path; the planned and full paths
--- a/src/zagg/processing/read.py
+++ b/src/zagg/processing/read.py
@@ -333,21 +333,18 @@ def _planned_read_group(
     if n_base <= 0:
         return None
 
-    # Compute the shard's WGS84 bbox from the grid (every grid's
-    # ``shard_footprint`` returns a shapely ``Polygon`` or ``MultiPolygon``).
-    # An antimeridian-crossing HEALPix shard's footprint can come back as a
-    # split ``MultiPolygon`` (see ``zagg.viz.shardmap._split_antimeridian``),
-    # in which case ``.bounds`` spans ~360 deg in lon and would neutralize
-    # the IO bound (the AOI would intersect every segment). Same for
-    # globe-spanning polar caps. Detect the wide-bbox case up front and fall
-    # back to ``_read_group_full`` so we don't pretend to optimize.
-    poly = grid.shard_footprint(shard_key)
-    min_lon, min_lat, max_lon, max_lat = poly.bounds
-    if (max_lon - min_lon) >= 180.0:
-        # Hand off to the full-read path; the planned-IO benefit is gone for
-        # this shard and trying to plan would waste the coarse-coord read.
-        return _read_group_full(h5obj, group, data_source, shard_key, grid, arrow=arrow)
-    bbox = (float(min_lon), float(min_lat), float(max_lon), float(max_lat))
+    # Match segments to this shard with the SAME mortie test the photon path
+    # applies below (``grid.shards_of(grid.assign(...)) == shard_key``), not a
+    # loose bbox + per-segment shapely scan (issue #95). It is exact to the leaf
+    # cell, vectorized (~280x faster than the shapely loop on a 181k-segment
+    # ATL03 beam), and antimeridian/polar-correct -- so the wide-bbox bail the
+    # old bbox path needed is gone; a shard that genuinely spans most segments is
+    # still caught by ``plan_read``'s selectivity ``full_read`` fallback. The
+    # mask is rep-point based, so a boundary segment whose photons straddle the
+    # shard edge is recovered by ``pad`` (and the photon-level filter below never
+    # over-includes); residual omission is bounded to a few edge photons (#95).
+    coarse_leaf = grid.assign(np.asarray(coarse_lats), np.asarray(coarse_lons))
+    coarse_mask = grid.shards_of(coarse_leaf) == int(shard_key)
 
     plan = plan_read(
         np.asarray(coarse_lats),
@@ -355,10 +352,10 @@ def _planned_read_group(
         np.asarray(ibeg_arr),
         np.asarray(cnt_arr),
         n_base,
-        bbox,
         index_base=index_base,
         pad=pad,
         full_read_threshold=full_read_threshold,
+        coarse_mask=coarse_mask,
     )
 
     if not plan.parent_runs:
--- a/src/zagg/read_plan.py
+++ b/src/zagg/read_plan.py
@@ -1,9 +1,11 @@
 """Offline-computable read plan for AOI-based hyperslice selection (issue #43, Phase C).
 
-``plan_read`` computes which coarse-level segments (e.g. ATL03 ``land_ice_segments``)
-overlap a bounding-box AOI, merges adjacent runs, translates them to base-level
-(photon) slices, and packages them as h5coro-compatible hyperslice lists.
-``execute_read_plan`` then drives a caller-supplied read function with those slices.
+``plan_read`` takes which coarse-level segments (e.g. ATL03 ``land_ice_segments``)
+match the AOI -- preferably a precomputed mortie segment->shard ``coarse_mask``
+(issue #95), or a bbox fallback -- merges adjacent runs, translates them to
+base-level (photon) slices, and packages them as h5coro-compatible hyperslice
+lists. ``execute_read_plan`` then drives a caller-supplied read function with
+those slices.
 
 Both functions are pure and offline-testable — no h5coro, S3, or credentials needed.
 """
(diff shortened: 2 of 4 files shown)


espg:
> Dropping the wide-bbox antimeridian bail: I'm relying on mortie assign being globally correct (it is) plus the selectivity full_read fallback for genuinely-wide shards. Comfortable with that, or keep an explicit guard?

yes. mortie avoids antimerdian issues explicitly by design.

This looks good, but it needs an adversarial review before it's ready for merge.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the mortie segment->shard mask swap (Closes #95). Net: the core mechanics are correct. The vectorized run detection is provably equivalent to the old loop (exhaustive parity verified for all bool masks up to n=7, including empty / all-True / single / runs touching index 0 and the last index / alternating), no dead imports, both modules stay under 1000 lines, no new dependencies. Findings below, ordered by severity. None block; itemized for the author.

**1. [FOLD] Stale docstring in `_read_group` — `src/zagg/processing/read.py:531-538`.** The class-level dispatch docstring still reads "the AOI bbox is computed from the grid's shard footprint." That path is gone — the planned read now derives `coarse_mask` from `grid.shards_of(grid.assign(...))`, never from `shard_footprint`. Update this paragraph (and the `_planned_read_group` docstring at read.py:276-279, which still says "which base-rate slices the AOI **bbox** actually touches") so the narrative matches the mask path. The `read_plan.plan_read` docstring itself was updated correctly; these two in read.py were missed.

**2. [NOTE] Omission guarantee in the comment overstates the bound — `src/zagg/processing/read.py:24-33` and `src/zagg/read_plan.py` docstring (the `coarse_mask` bullet).** The comment asserts a boundary photon is "recovered by `pad`" and "residual omission is bounded to a few edge photons." Precisely: a photon physically in shard S is recovered only if its owning segment lies within `pad` segments of some matched (rep-point-in-S) segment. With the production default `pad=1` that covers distance-1 only; a photon whose owning segment is >=2 away from every matched rep-point is silently omitted. For dense ~20 m ATL03 segments this is effectively unreachable in practice, so I'm leaving this standing rather than calling it a bug — but the comment should be softened from an absolute ("bounded to a few edge photons") to "bounded by `pad` segments," since the new test only exercises the distance-1 case (see finding 4). Flagging for @espg's call on whether the wording is acceptable.

**3. [NOTE] NaN / fill-value rep-points feed `grid.assign` unfiltered — `src/zagg/processing/read.py:34` (`coarse_leaf = grid.assign(np.asarray(coarse_lats), np.asarray(coarse_lons))`).** The old bbox matcher was NaN-safe by construction (`bbox[0] <= lon <= bbox[2]` is `False` for NaN). The new mask delegates to `grid.assign`, whose NaN behavior is grid-dependent (real mortie may raise or return a garbage leaf id for an empty-segment fill-value rep-point). This is partially de-risked downstream: empty segments (`cnt==0`) are excluded as run boundaries by the `cnt[s:e+1] > 0` filter in `plan_read` (read_plan.py:173), so a spurious NaN-driven match on a zero-photon segment contributes no photons and at worst widens a run by one empty segment. Still, no test covers a NaN/fill rep-point on a non-empty or boundary segment, and the photon-level re-filter at read.py:401-402 is the real safety net. Leave for @espg; if mortie's `assign` is known NaN-tolerant this is a non-issue.

**4. [FOLD] New pad test only covers distance-1 recovery — `tests/test_processing.py:3255` (`test_pad_recovers_boundary_segment_and_matches_full`).** Good test, and the pad0-omits / pad1-recovers + full-read parity assertion is exactly the right shape. But it only demonstrates the distance-1 case the comment's guarantee actually holds for. To make the test match the claim in finding 2 (and document the limit), add a short assertion that a photon whose owning segment is 2 away from any matched rep-point is NOT recovered at `pad=1` (i.e. the omission is real beyond `pad`), so the bound is pinned by a test rather than only by prose.

**5. [FOLD] `int(shard_key)` cast is inconsistent with the photon-level compare — `src/zagg/processing/read.py:35` vs `:402`.** The new coarse mask uses `grid.shards_of(coarse_leaf) == int(shard_key)` while the photon-level spatial mask one line of reasoning later uses `grid.shards_of(leaf_ids) == shard_key` (no cast, read.py:402). For the two filters to be "the SAME mortie test" the comment claims, the comparison should be identical on both sides. Either drop the `int()` (rely on numpy's broadcast equality, as the photon path does) or add it to both — pick one so a non-int `shard_key` can't make the coarse and photon filters disagree.

**6. [NOTE] `n_coarse = len(index_beg_arr)` change is correct — `src/zagg/read_plan.py:149`.** Confirmed sound: under the mask path `lat_arr`/`lon_arr` are ignored for matching, and `index_beg_arr` is required and one-per-parent in both paths, so keying `n_coarse` off it is the right invariant. The shape guard (`in_aoi.shape != (n_coarse,)`, read_plan.py:169) correctly validates the mask against this length. No change needed — noting because it was called out for verification.

**7. [NOTE] `test_low_selectivity_falls_back_to_full_read` no longer tests an antimeridian/polar case — `tests/test_processing.py` (renamed from `test_antimeridian_grid_falls_back_to_full_read`).** Dropping the wide-bbox bail is the intended behavior per #95, so removing the antimeridian-specific assertion is correct. The replacement exercises only the selectivity fallback via a permissive grid. There is now no test asserting the mask path is actually *correct* across the antimeridian/poles (the whole justification for dropping the bail) — the `_LatBboxGrid`/`_BboxGrid` stubs are lat-band fakes, not mortie, so finding 4's parity test and these swaps verify the plumbing, not mortie's global correctness. That's acceptable for an offline unit suite, but the "antimeridian/polar-correct" claim in the read.py:27 comment rests entirely on mortie, untested here. Leave for @espg.


---

espg:
🤖 *from Claude*

Ran the fresh-context adversarial self-review you asked for before merge. A separate review subagent posted its findings as a `COMMENT` review on this PR (prefixed `🤖 *from Claude (review)*`).

**Folded** (commit `fc33b3c`, `fold #95 review: ...`):
- **Cast consistency** — the coarse mask compared `grid.shards_of(...) == int(shard_key)` while the photon-level mask at `read.py:401` uses `== shard_key`. Dropped the `int()` so the "same mortie test" claim holds literally for both filters.
- **Stale docstrings** — `_planned_read_group` and `_read_group` still described the AOI as "computed from the grid's shard footprint / bbox". Rewrote both to describe the mortie segment→shard mask path (the `plan_read` docstring was already correct).
- **Pinned the omission bound** — added `test_pad_does_not_recover_segment_two_runs_away`: a stray in-band photon whose owning segment is two segments from the matched rep-point is *not* recovered at `pad=1` (the full read keeps it), confirming the recovery guarantee is exactly "within `pad` segments", not unconditional. Complements the existing `test_pad_recovers_boundary_segment_and_matches_full` (distance-1 recovery).

**Left standing for review** (not folded — judgment calls / out of scope):
- The rep-point omission prose was softened to "bounded by `pad` segments" in the PR body; the inline comment in `read.py` keeps the "a few edge photons" framing, which is accurate at dense ATL03 segment spacing. No code change implied.
- `grid.assign` now consumes rep-points that could be NaN/fill on empty (`cnt==0`) segments; empty segments are dropped as run boundaries (`read_plan.py:173`) and the photon-level re-filter is the net, but there's no direct NaN-rep-point test. Flagged, not added — let me know if you want one.
- Dropping the wide-bbox antimeridian/polar bail removes mortie's actual antimeridian correctness from this (offline) suite's coverage; intended per #95 and your confirmation, left as-is.

**Local status:** `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on the touched files; `pytest -q tests/test_read_plan.py tests/test_processing.py` → 184 passed (incl. the two pad-bound tests). The prior commit's CI was fully green (test 3.12/3.13, ruff, build); the fold push re-triggers them.

With the adversarial review complete, this PR is awaiting merge.

---

PR #103, opened 2026-06-24
Build the Lambda setup template via from_config so chunk_inner can't drift

### LINKED ISSUE #99: Lambda setup handler drops chunk_inner: template chunked coarser than workers write (block index out of bounds)
## Summary

The Lambda **setup** handler builds its grid by hand-constructing `HealpixGrid(...)` and never forwards `chunk_inner` from the config. The **worker** handler builds its grid via `from_config(...)`, which *does* read `chunk_inner`. In a multi-chunk-per-worker run (`chunk_inner` set, K>1) the two paths disagree: setup emits a one-chunk-per-shard (coarse) template, then the workers write finer inner-chunk block indices into it and every shard fails with a Zarr "block index out of bounds" error.

## Reproduction

```
python -m zagg \
  --config src/zagg/configs/atl03_tdigest_healpix.yaml \
  --catalog neon_tdigest_shardmap.json \
  --store s3://<bucket>/test_neon_order11_inner_chunk13.zarr \
  --region us-west-2 \
  --backend lambda --overwrite
```

Config grid: `parent_order: 11`, `chunk_inner: 13`, `child_order: 19` (K = 4^(13-11) = 16 inner chunks per shard).

Every shard fails identically:

```
[1/22] shard 5347396461485096971: Failed to write zarr: block index out of bounds for dimension with 50331648 chunk(s)
...
[22/22] ...
```

This was a freshly created array (not a stale store) — confirmed.

## Root cause

The two handler paths build the grid from different inputs:

| path | how the grid is built | `chunk_inner` | resulting chunk order |
|---|---|---|---|
| **setup** (template) | `HealpixGrid(parent_order=…, child_order=…, layout=…, config=…, populated_shards=…)` — `deployment/aws/lambda_handler.py:138` | **dropped** → defaults to `None` → `chunk_order = parent_order` | order **11**, `12·4^11` = **50,331,648** chunks |
| **worker** (writes) | `from_config(config, parent_order=…)` — `deployment/aws/lambda_handler.py:243` | `chunk_inner=grid_cfg.get("chunk_inner")` = 13 — `src/zagg/grids/__init__.py:55` | order **13**, block indices up to `12·4^13` = **805,306,368** |

So the setup invoke writes a 50,331,648-chunk (order-11) template, then the K=16 workers compute inner-chunk block indices via `grid.iter_chunks()` (`src/zagg/grids/healpix.py:192-196`), which are nested cell ids at `chunk_order=13` — up to 805,306,368. Those exceed the template's 50,331,648 chunks → out of bounds.

The numbers are exact and verified locally:
- order-11 geometry → exactly **50,331,648** chunks (the value in the error).
- `chunk_inner: 13` → **805,306,368** chunks, and `iter_chunks` emits order-13 indices.
- `emit_template(overwrite=True)` correctly re-chunks an existing store on both `LocalStore` and the obstore/S3 backend — so `--overwrite` is not the culprit; the template is simply built at the wrong order.

`#82` (multi-chunk-per-worker) is atomic — `iter_chunks` and the chunk-order `emit_template` shipped together — so any single deployed Lambda is internally consistent on the worker side; the divergence is purely that **setup bypasses `from_config`**.

Why it can hide: with `chunk_inner` unset (K=1) `iter_chunks` yields a single `block_index()` (parent-order, in-bounds), so the mismatch only surfaces when `chunk_inner` is set.

## Fix

Build the setup grid the same way the worker does so the two paths can't drift.

**Minimal** — forward the missing kwarg at `deployment/aws/lambda_handler.py:138`:

```python
grid = HealpixGrid(
    parent_order=event["parent_order"],
    child_order=event["child_order"],
    layout=layout,
    config=config,
    populated_shards=populated,
    chunk_inner=config.output.get("grid", {}).get("chunk_inner"),
)
```

**Better** — collapse the hand-built healpix branch onto `from_config` (which already reads `chunk_inner`), so setup and worker share one construction path. Requires `from_config` to accept the dense-layout `populated_shards` / `n_parent_cells` that setup passes — confirm it threads those before switching.

Either fix needs:
- a regression test asserting the setup template's chunk count equals `grid.chunk_grid_shape` when `chunk_inner` is set (would have caught this), and
- a Lambda layer rebuild/redeploy to take effect.

espg:
Closes #99.

## What / approach

The Lambda **setup** handler hand-built its HEALPix grid with `HealpixGrid(...)` and never forwarded `chunk_inner`, while the **worker** handler builds its grid via `from_config(...)` (which *does* read `chunk_inner`). In a multi-chunk-per-worker run (`chunk_inner` set, K>1) the two paths disagreed: setup emitted a one-chunk-per-shard (coarse, `parent_order`) template, then workers wrote finer inner-chunk block indices into it → every shard failed with a Zarr `block index out of bounds` error.

This takes the **"Better"** fix from the issue (per @espg's "'better' sounds cleaner"): collapse the hand-built HEALPix branch onto `from_config`, so **setup and worker share one construction path** and can't drift. `from_config` already accepts `populated_shards` and reads `chunk_inner` + `layout` from the config, so the change is contained:

```python
populated = (
    list(range(event["n_parent_cells"]))
    if event.get("n_parent_cells") is not None
    else None
)
grid = from_config(config, parent_order=event.get("parent_order"), populated_shards=populated)
```

- `chunk_inner` now flows into the template (the bug), so its chunk grid is `12·4^chunk_inner`, exactly what workers write.
- `child_order` / `layout` / grid `type` are now sourced from the config (single source of truth) rather than re-specified on the setup event; `n_parent_cells` still threads through as `populated_shards` for the dense layout (only its count matters for `emit_template`). The setup-mode docstring is updated to reflect this.

**Note for review — a subtle, intended behavior refinement:** the old setup derived `layout` from *whether* `n_parent_cells` was passed (`dense` if so, else `fullsphere`), independent of the config. Setup now takes `layout` from the config, matching the worker. This *removes* a latent drift: if an orchestrator passed `n_parent_cells` while the config said `fullsphere`, the old code built a dense template against fullsphere workers. A genuine dense run must (and already does, for the worker) set `layout: dense` in the config, where `populated_shards` is still threaded correctly. Flagging in case any caller relied on the old presence-of-`n_parent_cells` ⇒ dense coupling.

## Phases

- [x] Fix the setup grid construction + regression test pinning the template chunk count against the worker grid's `chunk_grid_shape` when `chunk_inner` is set.

## How tested

- New `tests/test_lambda_handler.py::TestSetupTemplate` — runs `_handle_setup` against an in-memory Zarr store for both chunk_inner configs (`atl03_tdigest_healpix`, `atl03_gain_bias_healpix`; parent_order 11 / chunk_inner 13 / child_order 19) and asserts the emitted template's chunk count equals `from_config(cfg).chunk_grid_shape` (`12·4^13`), the invariant the bug violated. Confirmed the test **fails on the pre-fix handler** and passes on the fix.
- `uv run pytest -q` → **814 passed, 1 skipped**.
- `uv run ruff check --select=E,F,W,I --ignore=E501 deployment/aws/lambda_handler.py tests/test_lambda_handler.py` → clean.

## Questions for review

- **`ruff format`** reports it would reformat `lambda_handler.py`, but every hunk is on pre-existing lines (one-arg-per-line `raise`/`return` blocks at L91/147/162/227) untouched by this change — the new `_handle_setup` body is already format-clean. Left unreformatted to avoid churning unrelated lines, matching the repo's existing hand-formatted style and the `ruff check`-only lint bot. Flag if you'd prefer a whole-file reformat.
- The issue notes the fix "needs a Lambda layer rebuild/redeploy to take effect" — out of scope here (no infra/deploy change in this PR; `deployment/aws/lambda_handler.py` is the handler code the issue names directly, not a build/standup script).

---
_Generated by [Claude Code](https://claude.ai/code/session_01SmHdkcUoStnFi8ZQtj656J)_

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -30,11 +30,11 @@ Setup mode (creates the zarr template once before per-cell fan-out):
 {
     "mode": "setup",
     "store_path": str,
-    "parent_order": int,
-    "child_order": int,
-    "n_parent_cells": int,
+    "parent_order": int,        # HEALPix fallback; config.output.grid wins
+    "n_parent_cells": int,      # OPTIONAL -- dense layout only (populated count)
     "overwrite": bool,
-    "config": dict,
+    "config": dict,             # single source of truth: child_order, chunk_inner,
+                                #   layout, and grid type all come from here
     "output_credentials": dict (optional, same shape as process mode),
 }
 
@@ -119,31 +119,30 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
 
 def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
     """Create the zarr template at ``event['store_path']``."""
-    from zagg.grids import HealpixGrid, from_config
+    from zagg.grids import from_config
 
     logger.info(f"Setup mode: creating template at {event.get('store_path')}")
     try:
         config = load_config_from_dict(event["config"])
         store = open_store(event["store_path"], **_output_store_kwargs(event))
-        grid_type = config.output.get("grid", {}).get("type", "healpix")
-        if grid_type == "healpix":
-            # n_parent_cells signals dense layout; populated_shards identities
-            # don't matter for emit_template (only the count does).
-            populated = (
-                list(range(event["n_parent_cells"]))
-                if event.get("n_parent_cells") is not None
-                else None
-            )
-            layout = "dense" if populated is not None else "fullsphere"
-            grid = HealpixGrid(
-                parent_order=event["parent_order"],
-                child_order=event["child_order"],
-                layout=layout,
-                config=config,
-                populated_shards=populated,
-            )
-        else:
-            grid = from_config(config, parent_order=event.get("parent_order"))
+        # Build the grid exactly as the worker does (from_config), so the
+        # template's chunk structure can't drift from what workers write. The
+        # old hand-built HEALPix branch dropped chunk_inner, under-chunking the
+        # template at parent_order while workers wrote finer chunk_inner block
+        # indices -> "block index out of bounds" (issue #99). from_config reads
+        # chunk_inner + layout from the config; n_parent_cells (dense layout)
+        # still threads through as populated_shards (only its count matters for
+        # emit_template).
+        populated = (
+            list(range(event["n_parent_cells"]))
+            if event.get("n_parent_cells") is not None
+            else None
+        )
+        grid = from_config(
+            config,
+            parent_order=event.get("parent_order"),
+            populated_shards=populated,
+        )
         grid.emit_template(store, overwrite=event.get("overwrite", False))
         return {"statusCode": 200, "body": json.dumps({"ok": True, "mode": "setup"})}
     except Exception as e:
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -130,9 +130,9 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
         # old hand-built HEALPix branch dropped chunk_inner, under-chunking the
         # template at parent_order while workers wrote finer chunk_inner block
         # indices -> "block index out of bounds" (issue #99). from_config reads
-        # chunk_inner + layout from the config; n_parent_cells (dense layout)
-        # still threads through as populated_shards (only its count matters for
-        # emit_template).
+        # chunk_inner + layout from the config. n_parent_cells is inert unless
+        # the config selects layout: dense, where it threads through as
+        # populated_shards (only its count matters for emit_template).
         populated = (
             list(range(event["n_parent_cells"]))
             if event.get("n_parent_cells") is not None
--- a/tests/test_lambda_handler.py
+++ b/tests/test_lambda_handler.py
@@ -15,8 +15,11 @@ from unittest.mock import MagicMock
 
 import pandas as pd
 import pytest
+from zarr import open_group
+from zarr.storage import MemoryStore
 
 from zagg.config import default_config
+from zagg.grids import HEALPIX_BASE_CELLS, from_config
 
 REPO_ROOT = Path(__file__).parent.parent
 HANDLER_PATH = REPO_ROOT / "deployment" / "aws" / "lambda_handler.py"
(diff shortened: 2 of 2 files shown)


tests/test_lambda_handler.py
```diff
@@ -212,3 +215,52 @@ def test_k_eq_1_ragged_keyed_by_shard_key(self, handler_mod, monkeypatch):
...
+    def test_setup_template_matches_worker_grid_chunk_count(self, handler_mod, monkeypatch):
+        # Same invariant via the gain/bias config (different schema, same grid).
+        cfg = default_config("atl03_gain_bias_healpix")
+        resp, store = self._setup(handler_mod, monkeypatch, asdict(cfg))
+        assert resp["statusCode"] == 200, json.loads(resp["body"])
+
+        worker_grid = from_config(cfg)
+        group = open_group(store, path=str(worker_grid.child_order), mode="r")
+        offset_arr = group["offset_h"]  # a resolution: chunk companion array
+        assert offset_arr.shape == worker_grid.chunk_grid_shape
```

Coverage gap (non-blocking, but worth a follow-up commit): both new tests use a `fullsphere` config (no `layout` key) and neither passes `n_parent_cells`, so the `populated_shards`/dense branch this PR explicitly threads through `from_config` is never exercised. The two tests are also near-duplicates — same grid (`parent_order=11/chunk_inner=13/child_order=19`), only the schema differs — so the second adds little over the first on the grid-invariant front.

Uncovered branches introduced/touched by the change:
- dense layout (`n_parent_cells` set → `populated_shards=list(range(...))`); at minimum assert it doesn't regress and that fullsphere ignores `populated_shards`.
- K==1 / `chunk_inner` unset (the byte-identical path) — a config without `chunk_inner` would lock in that the template chunks at `parent_order`.
- the rectilinear/non-HEALPix path, which the old code routed through a separate branch and now shares one path.

The chunk-count assertion itself is faithful: it pins to `worker_grid.chunk_grid_shape` (= `12·4^13`), which the old `parent_order`-chunked template (`12·4^11`) would have failed, so it *would* have caught #99.

---

deployment/aws/lambda_handler.py
```diff
@@ -30,11 +30,11 @@
 {
     "mode": "setup",
     "store_path": str,
-    "parent_order": int,
-    "child_order": int,
-    "n_parent_cells": int,
+    "parent_order": int,        # HEALPix fallback; config.output.grid wins
+    "n_parent_cells": int,      # OPTIONAL -- dense layout only (populated count)
```
espg, line 34: 🤖 *from Claude (review)*

Doc/event drift (minor, diff-scoped): this docstring block drops `child_order` from the setup event, which is correct now that `from_config` derives it via `get_child_order(config)`. But `runner.py::_invoke_lambda_setup` still puts `child_order` (and `parent_order`) into the setup event payload (see `runner.py` ~L853-861). After this PR `child_order` in the setup event is vestigial — read by nothing. Not a bug (it's harmless extra payload, and `parent_order` is still a legitimate fallback), but consider either dropping `child_order` from `_invoke_lambda_setup` in this PR or noting in the PR body that it's now ignored so a future reader doesn't assume setup honors it.

---

deployment/aws/lambda_handler.py
```diff
@@ -119,31 +119,30 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
...
+        # emit_template).
+        populated = (
+            list(range(event["n_parent_cells"]))
+            if event.get("n_parent_cells") is not None
+            else None
+        )
+        grid = from_config(
+            config,
+            parent_order=event.get("parent_order"),
+            populated_shards=populated,
```
espg, line 144: 🤖 *from Claude (review)*

Correctness — verified sound, recording the check so it's grounded. The prompt's worry ("orchestrator passes `n_parent_cells` but config omits `layout` → `from_config` builds fullsphere while caller thinks dense") does **not** occur with the real orchestrator: `runner.py::_run_lambda` sets `n_parent_cells` only `if grid_type == "healpix" and layout == "dense"` (L677), where `layout = get_layout(config)` reads `output.grid.layout`. `from_config` reads the same key with the same `fullsphere` default (`grids/__init__.py` L34-43). So `n_parent_cells` is set **iff** the config already says `layout: dense`, and both paths resolve to dense together — this PR is strictly *more* consistent than the hand-built branch it replaces (which forced `layout="dense"` from the mere presence of `n_parent_cells`, independent of config). Threading `populated_shards` unconditionally is safe: `set_populated_shards` is a no-op under fullsphere (`healpix.py` L117-118).

One residual sharp edge, not introduced here but newly reachable via this path: a *direct* setup invocation (not through `runner`) that passes `n_parent_cells` with a config whose `output.grid` omits `layout` will silently build a fullsphere template and ignore the count. Given setup is orchestrator-driven this is acceptable, but the inline comment two lines up asserts "n_parent_cells (dense layout) still threads through" without noting it's inert unless the config also selects dense — consider tightening that comment to say "...threads through **when the config selects dense**".

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of #103 (fixes #99). Net: the core fix is correct and well-targeted — routing setup through `from_config` makes the template's chunk grid share the worker's construction path, so `chunk_inner` can no longer drift. I verified the chunk-count invariant (`12·4^13` for the K>1 configs) and traced the dense-layout threading end-to-end. **Nothing blocking.** Findings, all diff-scoped/optional:

1. **Correctness — verified, not a defect.** The `layout`-from-config change is safe and strictly more consistent than the old hand-built branch. `runner.py` sets `n_parent_cells` only when `get_layout(config) == "dense"`, and `from_config` reads the same `output.grid.layout` key with the same default, so the two paths can't disagree. Threading `populated_shards` unconditionally is harmless (no-op under fullsphere). (inline on `_handle_setup`)

2. **Test coverage gap (recommend a follow-up commit).** Both new tests are fullsphere with no `n_parent_cells`, so the dense `populated_shards` branch this PR touches is untested; the K==1 (`chunk_inner` unset) and rectilinear paths are also uncovered; and the two tests are near-duplicates (same grid). The chunk-count assertion that *is* present is faithful and would have caught #99. (inline on the test class)

3. **Minor event drift.** `child_order` is correctly dropped from the documented setup event, but `runner.py::_invoke_lambda_setup` still ships it in the payload where it's now read by nothing — consider dropping it or noting it's ignored. (inline on the docstring)

4. **Comment nit.** The "n_parent_cells still threads through as populated_shards" comment reads as if dense always engages; it's inert unless the config also selects `layout: dense`. Worth one clarifying clause.

Module size fine (346 lines), no new dependencies, ruff (`--select=E,F,W,I --ignore=E501`) clean, all 9 handler tests pass locally.

---

espg:
🤖 *from Claude*

Folded the fresh-context self-review (`b30deb3`).

**Addressed (diff-scoped):**
- **Test coverage gap** — the review flagged that both original tests were fullsphere/K>1 near-duplicates, leaving the `populated_shards` (dense) branch this PR threads, and the K==1 path, uncovered. Replaced the duplicate gain/bias test with two new cases: `test_setup_dense_layout_threads_populated_shards` (asserts `n_parent_cells` → template chunk count == populated count, exercising the dense `populated_shards` thread) and `test_setup_template_k1_chunked_at_parent_order` (chunk_inner unset → chunked at `parent_order`, the path the fix must leave byte-identical). Extracted a `_template_chunk_count` helper. Now 3 setup tests, distinct branches.
- **Comment clarity** — tightened the `n_parent_cells` comment to say it's inert unless the config selects `layout: dense`, rather than reading as if dense always engages.

**Verified, no change needed:** the review independently confirmed the `layout`-from-config refinement is safe and *strictly more consistent* than the old code — `runner.py:677` sets `n_parent_cells` only when `get_layout(config) == "dense"`, the same `output.grid.layout` key `from_config` reads, so setup and worker resolve dense/fullsphere together; the hypothesized "n_parent_cells with fullsphere config" drift can't occur. That resolves the layout note I'd flagged under Questions for review.

**Left standing for @espg (out of scope for this bugfix):**
- `runner.py::_invoke_lambda_setup` still ships `child_order` in the setup payload, which the handler now ignores (it reads `child_order` from the config). Harmless, but the contract is slightly stale. Didn't drop it here to avoid touching the orchestrator signature in a focused fix — flagging for a follow-up tidy.

`uv run pytest tests/test_lambda_handler.py -q` → 10 passed; full suite 814 passed / 1 skipped; ruff `check` clean.

---

PR #106, opened 2026-06-25
Example notebook for the t-digest read helpers

### LINKED ISSUE #79: Read helper(s)

espg:
Closes #79.

**Blocked by #105** (merge/rebase order: #105 lands first). #105 establishes the repo-wide Binder infra (`.binder/environment.yml` + `.binder/postBuild` that `pip install ".[analysis]"`, plus the Binder badge/intro convention). This PR's notebook now relies on that env for `zagg[analysis]` instead of a per-notebook git-URL `%pip install`, so it should merge after #105.

## What this does

Adds the example notebook that issue #79 was waiting on to close out. The reader code itself landed earlier via PR #81 (the `zagg.readers` package: `read_tensors`, `read_raw_values`, `rasterize_cell`, `chunk_z_range`, plus `cdf_from_tdigest` in `zagg.stats.tdigest`). The only remaining deliverable was a notebook demonstrating that API end to end.

`notebooks/tdigest_reader_example.ipynb` walks through:

- **Building a synthetic t-digest store** with the same public API zagg uses to write the real product (`build_tdigest` + `write_csr`), laid out as the per-shard CSR ragged groups the readers consume (`{field}/{parent_morton}/{values,offsets,cell_ids}`).
- **`read_tensors`** — the `(64, 64, n_bins)` generator: shape/dtype, row-major cell placement, the morton id recovered from the store, a matplotlib plot of one cell&#39;s reconstructed histogram, and the `dtype` flag (`uint16`/`uint32`/`float32`).
- **The `fit` policy** — a deliberately wide chunk triggers the default `raise`, then `degrade_resolution` and `collapse_bins`.
- **`read_raw_values`** — lossless recovery from a merge-free digest, and the raise when a digest carries merged centroids.

Also adds a short &#34;Reading the output&#34; pointer in `docs/quickstart.md` linking the notebook, matching how the custom-aggregations notebook is referenced there.

## Approach / key decision: data source

Per CLAUDE.md §4, a notebook must be binder-runnable with data that is either publicly/anonymously reachable or an in-tree fixture. The realistic real-product source (ICESat-2 via earthaccess/NSIDC S3) needs auth, so the notebook instead **builds a self-contained synthetic in-memory store** in the first cell. The read path it exercises is identical to a real product&#39;s — only the `store` argument would change — and the notebook says so explicitly.

## Binder alignment with #105

Aligned to #105's Binder convention so the two PRs are consistent and the notebook runs on the shared `.binder/` image rather than carrying its own install mechanism:

- **Dropped the per-notebook `%pip install 'zagg[analysis] @ git+...'` cell.** Once #105 merges, `.binder/postBuild` already does `pip install ".[analysis]"`, so the matplotlib dep (the only `[analysis]`-only import) is provided by the Binder env. The git-URL reinstall was redundant.
- **Badge URL matches #105** — `?urlpath=lab/tree/notebooks/tdigest_reader_example.ipynb` (was `?labpath=...`), plus an italic `_Runs end-to-end on Binder ..._` intro line like #105's notebooks.
- **Standalone hint, not #105's mechanism** — the install section is now a brief "## 0. Requirements" note: nothing to do on Binder; `pip install 'zagg[analysis]'` for a standalone run.

## How tested

- Executed in full with `jupyter nbconvert --to notebook --execute` against the current `main` reader code — all 10 executed code cells run clean, the matplotlib figure renders, and the printed outputs are committed in the notebook (real, not fabricated; re-verified after the #105 alignment — a fresh execution reproduces all 10 committed stream outputs exactly). Sample outputs: tensors `(64, 64, 128) uint32`; `fit=&#34;raise&#34;` raises on a 400 m span; `collapse_bins` yields `(64, 64, 32)`; `read_raw_values` recovers `[1 2 3 4 5]` exactly and raises on a merged digest.
- `ruff check --select=E,F,W,I --ignore=E501 src tests` — passes (no `src`/`tests` changes); notebook validates as nbformat v4.

## Phases

- [x] Phase 1 — example notebook + quickstart pointer; executed against current reader API.
- [x] Phase 2 — fold self-review: add the `zagg[analysis]` install cell so the matplotlib plot is Binder-runnable; re-executed.
- [x] Phase 3 — align with #105's `.binder/` env: drop the git-URL `%pip install` cell, match #105's Binder badge/intro convention, keep a standalone install hint; re-executed against `main` (outputs unchanged).

## Questions for review

- **`Closes #79` vs leaving open for the real-product check.** PR #81 flagged a follow-up: a check against a *non-synthetic* (real-product) t-digest store, which isn&#39;t wired here (no anonymous fixture available). I used `Closes #79` since the notebook is the deliverable the issue was explicitly blocked on, but if you&#39;d rather keep #79 open to track that real-store check I can switch to `Refs #79` and/or split the real-store fixture into its own issue.
- **Pre-existing `ruff format` drift (not from this PR):** `ruff format --check src tests` reports 8 files needing reformat (e.g. `tests/test_catalog.py`, `tests/test_integration.py`, `tests/test_rectilinear.py`, `tests/test_runner.py`) — unchanged by this PR (present on a clean `main` checkout) and left untouched per the &#34;don&#39;t fix unrelated CI failures&#34; rule. Flagging so it&#39;s visible.


espg (commented):
🤖 *from Claude (review)*

Adversarial re-review of the t-digest reader notebook + quickstart pointer. I read the four reader/stat sources on `main` (`readers/tdigest_tensor.py`, `readers/__init__.py`, `csr.py`, `stats/tdigest.py`) and checked every code cell's API usage and every committed output against them.

**Verdict: the notebook is correct and the committed outputs are real, not fabricated.** I found one genuine Binder-runnability gap worth fixing, and a couple of minor notes. Nothing blocks the next phase.

### (1) Binder gap: cell 5 imports `matplotlib`, which is not in core deps and has no Binder env to pull it in — `ModuleNotFoundError` on Binder. (real issue; details inline)

This is the one substantive finding — see the inline comment on the matplotlib cell. Short version: `matplotlib` lives only in the `analysis` optional extra (`pyproject.toml:71`), not in core `dependencies` (`pyproject.toml:27-43`). The repo has no `binder/`, `requirements.txt`, or `environment.yml`, so mybinder.org will detect `pyproject.toml` and run a bare `pip install .` (core only). Result: the badge launches, but cell 5's `import matplotlib.pyplot as plt` raises `ModuleNotFoundError`. The PR body's "Questions for review" flags the missing Binder env as an optional follow-up, but it's actually load-bearing for the notebook's own "runs anywhere (including Binder)" claim. The existing `notebooks/cryocloud_example.ipynb` already solves exactly this with a first-cell `pip install --quiet 'zagg[analysis] ...'` — mirroring that here (or adding a `binder/requirements.txt` with `zagg[analysis]`) closes the gap.

### What I verified as correct (so it's on the record)

- **API signatures all match `main`.** `build_tdigest(values, delta=...)`, `write_csr(store, field_name, values_list, cell_ids, *, dtype, zarr_format)`, `read_csr` / `iter_csr_cells`, `read_tensors(store, field, *, n_bins, resolution, bottom, top, fit, dtype, ...)`, `read_raw_values(store, field, ...)`, and `chunk_z_range(digests, *, n_bins, resolution, bottom, top, fit)` are all called positionally/kw exactly as defined. No nonexistent kwargs, no return-shape misuse (`iter_csr_cells` → list of `(cell_id, payload)`; `read_tensors` → `(tensor, morton)`; `read_raw_values` → `(morton, cell_id, values)`, all unpacked correctly).
- **Outputs are internally consistent.** chunk 100 `total counts=6404` vs 6500 input is the dropped lower tail below the window floor (cell 4095 shows 1424 vs 1500 input, same cause); cell 5's full 2000 survive because `normal(22, 1.5)` sits well inside the window — both correct given that `chunk_z_range` only sets the window while `rasterize_cell` keeps all weight inside it. `collapse_bins` → `(64,64,32)` checks out: needed span 10 m, smallest pow2 `n` with `n*0.5 ≥ 10` is 32. `uint16/uint32` peak 295 vs `float32` 295.21 is consistent with `np.rint`. `read_raw_values` recovering `[1 2 3 4 5]` (sorted) matches the "centroids stored by mean" contract.
- **Binder badge URL is well-formed:** `mybinder.org/v2/gh/englacial/zagg/main?labpath=notebooks%2Ftdigest_reader_example.ipynb`.
- **No external data, no auth, no missing files** — store is fully synthetic/in-memory. (The only thing keeping it from actually running on Binder is the matplotlib dep in (1).)

### Minor notes (non-blocking)

- **(2) The `chunk_z_range` recompute in cell 5 is a documented drift risk, and it's handled well.** The comment "We recompute the floor here only to label the x-axis; the reader sets it internally" is exactly the right caveat. It passes the same `bottom=0.05, top=0.95, fit="raise"` defaults `read_tensors` uses, so it won't drift for this data. Fine as-is; flagging only so it's a conscious choice.
- **(3) quickstart pointer is correct.** The "Reading the output" section (`docs/quickstart.md:152-159`) links the right path and reads naturally after "Configuration"; the link target matches the added file. Good placement.
- **(4) No stray/placeholder/dead code.** Imports are placed at point-of-use (a notebook idiom, intentional given the narrative); not a defect.

Solid notebook overall — the only thing I'd actually fix before calling it Binder-runnable is (1).

---

espg:
🤖 *from Claude*

Phase 2 folds the one substantive self-review finding (1): the notebook imported `matplotlib` (only in the `analysis` extra), so a bare Binder `pip install .` would `ModuleNotFoundError` on the plot cell. Added a leading **install cell** mirroring `notebooks/cryocloud_example.ipynb`:

```
%pip install --quiet 'zagg[analysis] @ git+https://github.com/englacial/zagg.git'
```

It's tagged `skip-execution` so re-running `nbconvert` here doesn't trigger a network reinstall (and so the committed outputs stay real, not fabricated — the cell carries its source with no fabricated output). On Binder it pulls zagg + matplotlib so the plot cell runs.

Re-executed the full notebook against current `main`: all 10 executed cells pass, the figure renders, outputs committed. The non-blocking notes (2)–(4) from the review (axis-label recompute is caveated, quickstart pointer placement, point-of-use imports) needed no change.

This still leaves the optional follow-up of a pinned `binder/requirements.txt` if you'd prefer Binder not to build from the git URL — flagged as (2) in the PR body's Questions for review.

---

espg:
this will merge after #105 ; that will resolve the matplotlib dep issue . Have a look at #105 and update with the knowledge that will merge first.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of the #105-alignment commit (`bb206d3`, + `bf03924` newline fix). No blocking findings.

- **Binder badge** — both the badge and inline link now use `?urlpath=lab/tree/notebooks/tdigest_reader_example.ipynb`, matching #105's notebooks (e.g. `custom_aggregations.ipynb`). The old `?labpath=...` form is gone; path/filename correct.
- **git-URL `%pip install` removed cleanly** — no remaining cell references `pip`/`git+`, and nothing assumes a prior install ran. `matplotlib` (the only `[analysis]`-only import) is in `pyproject.toml` `[analysis]` (`matplotlib>=3.10.8`) and #105's `.binder/postBuild` runs `pip install ".[analysis]"`, so it is guaranteed on Binder.
- **"## 0. Requirements" cell** — accurate: defers to `.binder/` on Binder, gives a standalone `pip install 'zagg[analysis]'` hint only; no duplication of #105's mechanism and no stale "the cell below" wording.
- **nbformat** — validates as v4; the three touched cells are markdown-only (no committed outputs affected). A fresh `nbconvert --execute` against current `main` reader code runs clean and reproduces the committed stream outputs (10/10 match).
- **No leftover refs** elsewhere — `docs/quickstart.md` links the notebook by URL only.

Out-of-scope note (recorded, not actioned here): once #105 merges, its `.binder/environment.yml` comment listing the three Binder-runnable notebooks will be slightly stale since this PR adds a fourth — a #105 wording matter, not a #106 diff change.

---

espg:
🤖 *from Claude*

Aligned this notebook with #105 per your note that #105 merges first and resolves the matplotlib dep. Changes in `bb206d3` (+ a one-byte trailing-newline tidy in `bf03924`):

- **Dropped the per-notebook `%pip install 'zagg[analysis] @ git+...'` cell.** After #105 merges, `.binder/postBuild` already runs `pip install ".[analysis]"`, so matplotlib (the only `[analysis]`-only import here) comes from the shared Binder env — the git-URL reinstall was redundant.
- **Matched #105's Binder convention** — badge/link now use `?urlpath=lab/tree/notebooks/tdigest_reader_example.ipynb` (was `?labpath=...`), with an italic `_Runs end-to-end on Binder ..._` intro line like the #105 notebooks.
- **Kept it standalone-runnable** — the install section is now a brief "## 0. Requirements" note (nothing to do on Binder; `pip install 'zagg[analysis]'` otherwise), without duplicating #105's mechanism.

Re-executed end-to-end against current `main` reader code; the committed outputs are unchanged (a fresh `nbconvert --execute` reproduces all 10 stream outputs exactly).

**Merge order:** this PR is now marked **Blocked by #105** in the body and labeled `blocked` — it should land after #105 so the `.binder/` env is in place. Keeping it in draft.

---

PR #107, opened 2026-06-25
Worker-runtime stats + opt-in --profile timing (issue #100)

### LINKED ISSUE #100: Possible order driven mismatch?
Two runs over the same AOI at different resolutions produce divergent observation messages:

```
(zagg) espg@Max ~/s/zagg (main)> python -m zagg \                                                           
                                       --config src/zagg/configs/atl03_tdigest_healpix.yaml \
                                       --catalog neon_tdigest_shardmap.json \
                                       --store s3://sliderule-public/test_neon_order11.zarr \
                                       --region us-west-2 \
                                       --backend lambda --overwrite
Processing 22 of 22 cells (lambda)
You're now authenticated with NASA Earthdata Login
Using token with expiration date 08/18/2026
Loading cached SSO token for smdc
Account concurrency: limit=1000, current=1, padding=100, available=899 -> using 22 workers
Done: 22 cells, 5,391,489 obs, 0 errors, 188.3s
Lambda compute: 2594s total, 5189 GB-s, ~$0.07

Done: 22 cells with data, 5,391,489 obs, 0 errors, 188.3s
Lambda compute: 2594s total, 5189 GB-s, ~$0.07
Output: s3://sliderule-public/test_neon_order11.zarr
```
versus...

```
(zagg) espg@Max ~/s/zagg (main)> python -m zagg \                                         
                                       --config src/zagg/configs/atl03_tdigest_healpix.yaml \
                                       --catalog neon_tdigest_shardmap.json \
                                       --store s3://sliderule-public/test_neon_full.zarr \
                                       --region us-west-2 \
                                       --backend lambda --overwrite
Processing 196 of 196 cells (lambda)
You're now authenticated with NASA Earthdata Login
Using token with expiration date 08/18/2026
Loading cached SSO token for smdc
Account concurrency: limit=1000, current=1, padding=100, available=899 -> using 196 workers
  [  50/196] 0.8 cells/s
  [ 100/196] 1.5 cells/s
  [ 150/196] 1.9 cells/s
Done: 196 cells, 2,602,459 obs, 0 errors, 165.4s
Lambda compute: 13726s total, 27451 GB-s, ~$0.37

Done: 196 cells with data, 2,602,459 obs, 0 errors, 165.4s
Lambda compute: 13726s total, 27451 GB-s, ~$0.37
Output: s3://sliderule-public/test_neon_full.zarr
```

only difference between the two runs is the parent order (11 and 13); one claims 2,602,459 obs, the other 5,391,489 obs. Both agg to the same resolution grid cells, over the same area.

Possible causes--

- false / meaningless output claim on the obs (i.e., the output arrays have the same data)
- missing data in order 13 run

Need to check each of the output s3 arrays to track this down...

espg:
Closes #100.

Instruments the lambda backend with worker-runtime statistics so shardmaps can be sized correctly and the &#34;more workers, lower wall time&#34; paradox is answerable from the run summary. Design is the plan in issue #100 ([plan](https://github.com/englacial/zagg/issues/100#issuecomment-4804172485), [approval](https://github.com/englacial/zagg/issues/100#issuecomment-4804297901)) — both (1) and (2), with (1) always-on (no probe tax) and (2) behind an opt-in `--profile` flag.

## What this does

Wall time on a parallel fan-out tracks the **straggler**, not the mean — so the sum we already keep (`lambda_time_s`) hides the distribution. This surfaces it.

**Phase 1 — always-on (no probe tax):**
- After `dispatch()` returns in `_run_lambda`, computes over `durations = [r["lambda_duration"] for r in report.results if r.get("lambda_duration")]`:
  - `worker_max_s = max(durations)`, `worker_median_s = statistics.median(...)`, `worker_pstdev_s = statistics.pstdev(...)`, `worker_pct_timeout = worker_max_s / function_timeout_s`.
- `function_timeout_s` read once via `get_function_configuration(FunctionName=...)["Timeout"]` on the existing dispatch client, falling back to `720` (the `template.yaml` Timeout default) on permission error / missing key / non-int response (`_get_function_timeout_s`).
- Orchestrator phase brackets `setup_s` / `fanout_s` / `finalize_s` (just `time.time()` deltas around calls that already happen — zero worker tax, so always-on per the approved answer to plan open-question (a)).
- All keys added to the lambda `summary` dict; empty-`durations` degrades to `None`. `wall_time_s` semantics unchanged.
- One extra printed line by `Lambda compute:`: `Workers: max {…}s ({…}% of {…}s timeout), median {…}s, pstdev {…}s`.

**Phase 2 — opt-in `--profile` (CLI only):**
- `--profile` (store_true, default off) in `__main__.py`, threaded `profile=` through `agg(...)` → `_run_lambda`.
- When set, forwards `"profile": true` in the per-cell event; `process_shard(profile=True)` populates a `phase_timings` sub-dict (read / index / aggregate `time.time()` deltas) in its result metadata. The runner rolls the per-phase **max across cells** into `worker_phase_max` (added to the summary only when profiling) and prints a `Worker phases (max across cells): …` line.
- **When off, the worker path and per-cell event payload are byte-identical to before** — no `phase_timings` dict, no `time.time()` calls, no `profile` event key, no extra summary key.

**Phase 3 — deployment-side handler bridge (authorized by @espg, [comment](https://github.com/englacial/zagg/pull/107#issuecomment-4813952501)):**
- `deployment/aws/lambda_handler.py` now (a) reads `event["profile"]` (default `False`), (b) forwards `profile=` into `process_shard`, (c) brackets the **write** phase it owns with `time.time()` and adds `write` to the same `metadata["phase_timings"]` sub-dict, and (d) lets that dict serialize into the response body the runner already consumes. So with `--profile` on, `worker_phase_max` now carries the full read/index/aggregate/**write** breakdown end-to-end.
- `write` is recorded **only on a clean write** (inside the `try`, after the template-missing early-return): every failure path (no-data shard, missing template, raising write loop) leaves `write` absent, so the runner's per-phase `max(...)` rollup never folds a time-to-failure in as a real write duration.
- **Default (no `profile` key) is unchanged:** `process_shard(profile=False)` and no `phase_timings` in the body.

Uses stdlib `statistics` only — no new dependency.

## Phases
- [x] **Phase 1** — always-on worker stats (max/median/pstdev/pct-timeout) + `function_timeout_s` + orchestrator brackets (`setup_s`/`fanout_s`/`finalize_s`) in the lambda summary and one printed line. Tests pin all four stats against a synthetic `report.results` with a stubbed timeout, the empty-durations None-degrade, the brackets, and `_get_function_timeout_s` read/fallback.
- [x] **Phase 2** — `--profile` flag plumbed CLI → `agg` → `_run_lambda` → per-cell event; worker `phase_timings` (read/index/aggregate) only when set; `worker_phase_max` rollup + print only when set; default runs leave the event payload + worker path byte-identical. Tests assert plumbing, byte-identical default event, worker `phase_timings`-only-when-set on success + no-data paths, and the rollup = per-phase max across cells.
- [x] **Phase 3** — handler bridge in `deployment/aws/lambda_handler.py`: read `event["profile"]`, forward `profile=` into `process_shard`, bracket the handler-owned `write` phase, merge into the response body. Recorded only on a clean write (failure paths omit it). New `TestProcessEventProfile` in `tests/test_lambda_handler.py` (import-by-path) covers: profile forwarded + worker timings pass through + `write` added; default-off (no key, no timings); and `write` absent on the empty-`chunk_results`, raising-write, and missing-template paths.

## How tested
- `uv run pytest -q` — full suite green: **838 passed, 1 skipped**.
- `uv run ruff check src/zagg/runner.py src/zagg/__main__.py src/zagg/processing/worker.py deployment/aws/lambda_handler.py tests/test_runner.py tests/test_lambda_handler.py` — clean (full `E,F,W,I,N`). The only `ruff check src tests` finding is a **pre-existing** `N818` in `src/zagg/registry.py` (untouched file) — flagged, not fixed (§4).
- New tests in `tests/test_runner.py`: `TestWorkerRuntimeStats`, `TestGetFunctionTimeout`, `TestProfilePlumbing`, `TestWorkerPhaseTimings`, plus two new cases in `TestInvokeLambdaCellEvent`. New tests in `tests/test_lambda_handler.py`: `TestProcessEventProfile` (7 cases).
- Each phase self-reviewed adversarially in a fresh context (see review comments); Phase 1's `N803` lint blocker and a timeout-stub hygiene fix were folded into the Phase 2 commit, and the Phase 3 review&#39;s write-on-failure finding + failure-path coverage gaps were folded into a follow-up commit.

## Questions for review

1. ~~**`phase_timings` write-phase + handler bridge are deployment-side (§1 boundary) — should I wire them?**~~ **RESOLVED.** @espg green-lit editing the deploy path ([comment](https://github.com/englacial/zagg/pull/107#issuecomment-4813952501): &#34;yes. you can modify the deploy path to enable this for us.&#34;), which satisfies §1&#39;s &#34;unless the issue names it&#34; carve-out for this change. Landed as **Phase 3** above: the 3-line bridge (read `event["profile"]`, forward `profile=`, merge `phase_timings`) plus a handler-owned `write` bracket, recorded only on a clean write.
2. ~~**`runner.py` is now 1100 lines, over the §4 ~1000-line soft limit.**~~ **Closed** per @espg (&#34;ignore it for now&#34;). No split in this PR.
3. ~~**`ruff format` is not repo-clean today.**~~ **Closed** per @espg (&#34;ok&#34;). New code stays in the file&#39;s existing hand-wrapped style; unrelated lines are left unformatted (the enforced bot is `ruff check` per §7, which is clean). `lambda_handler.py`&#39;s remaining `ruff format` diff is entirely pre-existing style on untouched lines.

All three phases complete and green; held in **draft** awaiting @espg review.

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -53,6 +53,7 @@ full pipeline using only lambda:InvokeFunction.
 import json
 import logging
 import os
+import time
 from typing import Any, Dict
 
 # Import cloud-agnostic processing
@@ -247,6 +248,11 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         # Zarr chunk through the sink. At K==1 the sink holds exactly one entry whose
         # ``block_index`` equals ``event["chunk_idx"]``, so the write is unchanged.
         from zagg.processing import process_shard
+        # Opt-in per-phase timing (issue #100). When the orchestrator forwards
+        # ``profile``, ``process_shard`` fills ``metadata["phase_timings"]`` with
+        # read/index/aggregate deltas; the write phase runs here, so we bracket it
+        # below and merge it in. Default (no key) leaves the worker path unchanged.
+        profile = event.get("profile", False)
         chunk_results: list = []
         _df_out, metadata = process_shard(
             grid,
@@ -255,10 +261,12 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
             s3_credentials=s3_creds,
             config=config,
             chunk_results=chunk_results,
+            profile=profile,
         )
 
         # Write Zarr to store: one dense region per chunk plus its ragged (CSR)
         # companion. Mirrors the local runner's K>1 write loop (``_process_and_write``).
+        _write_t0 = time.time() if profile else None
         if chunk_results:
             store_path = event["store_path"]
             store = open_store(store_path, **_output_store_kwargs(event))
@@ -305,6 +313,13 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
             except Exception as e:
                 logger.error(f"Failed to write zarr to {store_path}: {e}")
                 metadata["error"] = f"Failed to write zarr: {e}"
+
+        # Record the write-phase timing (issue #100). read/index/aggregate come
+        # from ``process_shard``; ``write`` is owned here, so it joins the same
+        # sub-dict only when profiling and a write actually ran (non-empty
+        # ``chunk_results``). The no-data path leaves ``write`` absent.
+        if profile and chunk_results and "phase_timings" in metadata:
+            metadata["phase_timings"]["write"] = time.time() - _write_t0
 
         # Log structured result
         logger.info(
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -310,16 +310,18 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
                         grid=grid,
                         shard_key=ragged_key,
                     )
+                # Record the write-phase timing (issue #100) only on a clean write:
+                # read/index/aggregate come from ``process_shard``; ``write`` is owned
+                # here and joins the same sub-dict. Recording inside the ``try`` (and
+                # after the early-return template-missing 500) keeps ``write`` absent
+                # on every failure path, so the runner's per-phase max rollup never
+                # folds in a time-to-failure as a real write duration. The no-data
+                # path returns earlier without writing, so ``write`` stays absent too.
+                if profile and "phase_timings" in metadata:
+                    metadata["phase_timings"]["write"] = time.time() - _write_t0
             except Exception as e:
                 logger.error(f"Failed to write zarr to {store_path}: {e}")
                 metadata["error"] = f"Failed to write zarr: {e}"
-
-        # Record the write-phase timing (issue #100). read/index/aggregate come
-        # from ``process_shard``; ``write`` is owned here, so it joins the same
-        # sub-dict only when profiling and a write actually ran (non-empty
-        # ``chunk_results``). The no-data path leaves ``write`` absent.
-        if profile and chunk_results and "phase_timings" in metadata:
-            metadata["phase_timings"]["write"] = time.time() - _write_t0
 
         # Log structured result
         logger.info(
--- a/tests/test_lambda_handler.py
+++ b/tests/test_lambda_handler.py
@@ -223,9 +223,21 @@ class TestProcessEventProfile:
     write phase it owns, merging ``phase_timings`` into the response body. When
     ``profile`` is absent the worker call and body are unchanged."""
 
-    def _patch(self, handler_mod, monkeypatch, *, worker_phase_timings, chunks):
+    def _patch(
+        self,
+        handler_mod,
+        monkeypatch,
+        *,
+        worker_phase_timings,
+        chunks,
+        template_exists=True,
+        write_raises=False,
+    ):
         """Patch process_shard to record the ``profile`` kwarg and (when given)
-        seed ``metadata['phase_timings']``; fill the sink with ``chunks``."""
+        seed ``metadata['phase_timings']``; fill the sink with ``chunks``.
+        ``template_exists`` toggles the store's existence check (template-missing
+        500 path); ``write_raises`` makes the dense write blow up (failed-write
+        500 path)."""
         import zagg.grids as grids
         import zagg.processing as processing
 
@@ -246,36 +258,46 @@ class TestProcessEventProfile:
             return pd.DataFrame(), meta
 
         store = MagicMock()
-        store.exists.return_value = True
+        store.exists.return_value = template_exists
         grid_stub = MagicMock()
         grid_stub.group_path = "8"
         grid_stub.chunk_grid_shape = (4,)
 
+        def fake_write_dense(*a, **k):
+            if write_raises:
+                raise RuntimeError("boom")
+
         monkeypatch.setattr(processing, "process_shard", fake_process_shard)
         monkeypatch.setattr(grids, "from_config", lambda *a, **k: grid_stub)
         monkeypatch.setattr(handler_mod, "open_store", lambda *a, **k: store)
-        monkeypatch.setattr(handler_mod, "write_dataframe_to_zarr", lambda *a, **k: None)
+        monkeypatch.setattr(handler_mod, "write_dataframe_to_zarr", fake_write_dense)
         monkeypatch.setattr(handler_mod, "write_ragged_to_zarr", lambda *a, **k: None)
         return cap
 
-    def test_profile_forwarded_and_write_bracketed(self, handler_mod, monkeypatch):
-        # profile=True flows into process_shard, and the handler-owned write phase
-        # is appended to the worker's read/index/aggregate timings in the body.
-        cap = self._patch(
-            handler_mod,
-            monkeypatch,
-            worker_phase_timings={"read": 1.0, "index": 0.5, "aggregate": 0.25},
-            chunks=[((0,), pd.DataFrame(), {})],
-        )
+    def _profile_event(self):
         event = _base_event(_healpix_config_dict())
         event["child_order"] = 12
         event["profile"] = True
-        resp = handler_mod._handle_process(event, _context())
+        return event
+
+    def test_profile_forwarded_and_write_bracketed(self, handler_mod, monkeypatch):
+        # profile=True flows into process_shard, the worker's read/index/aggregate
+        # timings pass through unchanged, and the handler-owned write phase is added.
+        worker = {"read": 1.0, "index": 0.5, "aggregate": 0.25}
+        cap = self._patch(
+            handler_mod,
+            monkeypatch,
+            worker_phase_timings=worker,
+            chunks=[((0,), pd.DataFrame(), {})],
+        )
+        resp = handler_mod._handle_process(self._profile_event(), _context())
         assert resp["statusCode"] == 200
         assert cap["profile"] is True
         timings = json.loads(resp["body"])["phase_timings"]
         assert set(timings) == {"read", "index", "aggregate", "write"}
-        assert timings["write"] >= 0.0
+        # worker-seeded phases pass through untouched; write is a non-negative delta.
+        assert {k: timings[k] for k in worker} == worker
+        assert isinstance(timings["write"], float) and timings["write"] >= 0.0
 
     def test_default_off_no_profile_no_timings(self, handler_mod, monkeypatch):
         # No profile key -> process_shard gets profile=False and the body carries
@@ -292,6 +314,50 @@ class TestProcessEventProfile:
         assert resp["statusCode"] == 200
         assert cap["profile"] is False
         assert "phase_timings" not in json.loads(resp["body"])
+
+    def test_profile_no_data_omits_write(self, handler_mod, monkeypatch):
+        # Empty chunk_results (no-data shard): the worker still seeds read but no
+        # write runs, so the body carries phase_timings WITHOUT a write key.
+        self._patch(
+            handler_mod,
+            monkeypatch,
+            worker_phase_timings={"read": 1.0},
+            chunks=[],
+        )
+        resp = handler_mod._handle_process(self._profile_event(), _context())
+        assert resp["statusCode"] == 200
+        timings = json.loads(resp["body"])["phase_timings"]
+        assert "write" not in timings
+
+    def test_profile_failed_write_omits_write(self, handler_mod, monkeypatch):
+        # A raising write loop (500): write timing is recorded only on a clean
+        # write, so a time-to-failure never leaks in as a real write duration.
+        self._patch(
+            handler_mod,
+            monkeypatch,
+            worker_phase_timings={"read": 1.0, "index": 0.5, "aggregate": 0.25},
+            chunks=[((0,), pd.DataFrame(), {})],
+            write_raises=True,
+        )
+        resp = handler_mod._handle_process(self._profile_event(), _context())
+        assert resp["statusCode"] == 500
+        timings = json.loads(resp["body"])["phase_timings"]
+        assert "write" not in timings
+
+    def test_profile_missing_template_omits_write(self, handler_mod, monkeypatch):
+        # Template-missing early return (500) bypasses the write loop, so write
+        # stays absent (the worker-seeded phases still ride along in the body).
+        self._patch(
+            handler_mod,
+            monkeypatch,
+            worker_phase_timings={"read": 1.0, "index": 0.5, "aggregate": 0.25},
+            chunks=[((0,), pd.DataFrame(), {})],
+            template_exists=False,
+        )
+        resp = handler_mod._handle_process(self._profile_event(), _context())
+        assert resp["statusCode"] == 500
+        timings = json.loads(resp["body"])["phase_timings"]
+        assert "write" not in timings
 
 
 class TestSetupTemplate:
--- a/tests/test_lambda_handler.py
+++ b/tests/test_lambda_handler.py
@@ -217,6 +217,83 @@ class TestProcessEventWriteLoop:
         assert cap["ragged"][0][0] == event["shard_key"]
 
 
+class TestProcessEventProfile:
+    """Issue #100 phase 3: the handler bridges the opt-in ``profile`` event key
+    into ``process_shard`` (read/index/aggregate timing) and brackets the
+    write phase it owns, merging ``phase_timings`` into the response body. When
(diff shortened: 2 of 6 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of **Phase 1** (commit `5940964`).

**Blocker (folded into the phase-2 commit):**
- `tests/test_runner.py` — a test stub `def get_function_configuration(self, FunctionName):` tripped `N803` under the full `ruff check src tests` (pyproject `select=[E,F,W,I,N]`). The PR lint bot uses `--select=E,F,W,I` (no `N`) so it wouldn't catch it, but local/CI `ruff check` would fail. Fixed by switching the stub to `**kwargs` (the production call passes `FunctionName=` as a boto3 keyword, so the name must stay on the real path; the stub doesn't assert on it).

**Major (hygiene, folded):**
- The two pre-existing lambda tests (`test_lambda_summary_keys_and_cost`, `test_lambda_cost_byte_identical_with_mixed_durations`) ran with `function_timeout_s == 1` because a bare `MagicMock` client makes `int(MagicMock()["Timeout"])` return `1` (not the 720 fallback). They still passed (assert only the key set / cost), but the timeout value flowing through was nonsense. Added `monkeypatch.setattr(runner, "_get_function_timeout_s", lambda *a, **k: 720)` to both for determinism.

**Verified clean (no action):** `wall_time_s` stays byte-identical (brackets are independent `time.time()` deltas, `start_time`/`wall_time` unmoved); empty-durations → all four keys `None` + print skipped; `worker_pct_timeout` guarded against `function_timeout_s` falsy; `_get_function_timeout_s` falls back on AccessDenied / missing-key / non-int; `pstdev` (not `pvariance`) is pinned (≈35.36 vs 1250.0, so a regression to variance fails the test); the truthy `if r.get("lambda_duration")` filter correctly drops the `0` durations failed cells report.

**Flagged for review (see PR body):** `runner.py` crossed the §4 ~1000-line soft limit (1008 → 1068 in Phase 1); pre-existing breach, pushed further.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial self-review of **Phase 2** (commit `ce5d565`).

**Verdict: ship-ready — no blockers, no majors.** The byte-identical-when-off invariant holds on every path checked:
- Event payload: `event["profile"]` set only under `if profile`; no key otherwise (pinned by `test_default_event_has_no_profile_key`).
- Worker path: `phase_timings`/`_read_t0` are `None` via ternaries that call `time.time()` only when profiling; every write is `if profile`-guarded — no dict alloc, no timing calls on the default path. `metadata["phase_timings"]` is set only when profiling on all return paths (no-granules, no-data, success, K==1, K>1). No `UnboundLocalError` risk (`_index_t0`/`_aggregate_t0` are referenced and assigned symmetrically under `if profile`).
- Default lambda summary key set unchanged (`worker_phase_max` added only when profiling); `TestSummaryKeysByteIdentical` stays green.
- Rollup handles missing `body`/`phase_timings` (`or {}`) and computes per-phase max across cells; tests pin empty-when-no-timings, max-across-cells, and default-omits-key.
- Worker-test stubs genuinely traverse the index + aggregate brackets (default `handoff="pandas"` exercises `_concat_and_group`; `del grid.iter_chunks` forces the K==1 fallback loop), so the brackets are really run, not short-circuited.

**Nits (non-blocking, no change made):**
- `--profile` is silently ignored on `--backend local` (threaded only to `_run_lambda`), matching how `handoff` is local-only. Consistent with the locked "lambda backend" design.
- `runner.py` now 1100 lines, over the §4 soft limit — flagged in the PR body for a maintainer call.

**Not a bug (intentional §1 deferral):** the handler→worker `profile` bridge in `deployment/aws/lambda_handler.py` is deliberately unwired — see Question (1) in the PR body. `test_profile_run_with_no_phase_timings_is_empty` pins the graceful empty-rollup for that interim state.

---

espg:
🤖 *from Claude*

Both phases landed and CI is green (`test (3.12)`, `test (3.13)`, `ruff`, `build-x86_64` all passing; arm64 build still running but unaffected by this pure-Python change). Full suite: 833 passed, 1 skipped.

- **Phase 1** (`5940964`): always-on worker stats (`worker_max_s` / `worker_median_s` / `worker_pstdev_s` / `worker_pct_timeout`), `function_timeout_s` (read once from `get_function_configuration`, 720 fallback), and orchestrator brackets (`setup_s` / `fanout_s` / `finalize_s`) in the lambda summary + one printed `Workers:` line.
- **Phase 2** (`ce5d565`): opt-in `--profile` plumbed CLI → `agg` → `_run_lambda` → per-cell event; `process_shard(profile=True)` emits `phase_timings` (read/index/aggregate); `worker_phase_max` rollup + print, all gated so default runs stay byte-identical.

Adversarial self-reviews for both phases are posted above; Phase 1's one `N803` lint blocker and a timeout-stub hygiene fix were folded into the Phase 2 commit.

**Holding in draft + `waiting` on one decision (Question (1) in the PR body):** the `phase_timings` write-phase bracket and the handler→worker `profile` bridge live in `deployment/aws/lambda_handler.py`, which §1 says I shouldn't touch unless the issue names it. So `phase_timings` is fully implemented worker-side and orchestrator-side, but won't surface in production until the handler forwards `event["profile"]` into `process_shard` and merges the result (3 lines + an optional `write` bracket). Want me to make that handler edit as a follow-up phase, or keep `deployment/` untouched and merge the worker-side support dormant? (Also flagged: `runner.py` is now 1100 lines, over the §4 soft limit — Question (2).)

---

espg:
> phase_timings write-phase + handler bridge are deployment-side (§1 boundary) — should I wire them? 

yes. you can modify the deploy path to enable this for us.

> runner.py is now 1100 lines, over the §4 ~1000-line soft limit.

ignore it for now

> ruff format is not repo-clean today

ok

deployment/aws/lambda_handler.py
```diff
@@ -306,6 +314,13 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
                 logger.error(f"Failed to write zarr to {store_path}: {e}")
                 metadata["error"] = f"Failed to write zarr: {e}"
 
+        # Record the write-phase timing (issue #100). read/index/aggregate come
+        # from ``process_shard``; ``write`` is owned here, so it joins the same
+        # sub-dict only when profiling and a write actually ran (non-empty
+        # ``chunk_results``). The no-data path leaves ``write`` absent.
```

(1) Medium — `write` is recorded even when the write **failed**. The write loop's `except` (line ~315) sets `metadata["error"]` but does not return; execution falls through to this block, so `metadata["phase_timings"]["write"]` is populated with the time-to-failure of a partial/aborted write rather than a real write duration. The runner then folds it into `worker_phase_max["write"]` via `max(...)` (runner.py ~L780), so a single failed cell can skew the per-phase straggler with a number that isn't a successful write at all — and on that path `statusCode` is 500, so the body that carries the bogus `write` is still consumed by `_accumulate`.

Is timing a failed write intended? If yes, a one-line comment would settle it; if not, gate on write success (e.g. a `wrote_ok` flag set after the loop completes without exception). Either way this path has no test — the three profile cases all use a no-op writer that never raises.

---

deployment/aws/lambda_handler.py
```diff
@@ -306,6 +314,13 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
                 logger.error(f"Failed to write zarr to {store_path}: {e}")
                 metadata["error"] = f"Failed to write zarr: {e}"
 
+        # Record the write-phase timing (issue #100). read/index/aggregate come
+        # from ``process_shard``; ``write`` is owned here, so it joins the same
+        # sub-dict only when profiling and a write actually ran (non-empty
```

(2) Low — the `"phase_timings" in metadata` guard silently absorbs a worker-contract break. When `profile=True`, `process_shard` always seeds `metadata["phase_timings"]` (it does so on every return path, including the no-data early return). So the only way `profile and chunk_results` is true but this key is absent is if the worker's contract regresses — and in that case you'd silently drop the `write` timing rather than surfacing the mismatch. That's defensible as defensive coding, but worth a one-word comment ("defensive: worker always sets it under profile") so a future reader doesn't think there's a real path where it's missing. Not blocking.

---

deployment/aws/lambda_handler.py
```diff
@@ -255,10 +261,12 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
             s3_credentials=s3_creds,
             config=config,
             chunk_results=chunk_results,
+            profile=profile,
         )
 
         # Write Zarr to store: one dense region per chunk plus its ragged (CSR)
         # companion. Mirrors the local runner's K>1 write loop (``_process_and_write``).
+        _write_t0 = time.time() if profile else None
         if chunk_results:
```
espg, line 270: 🤖 *from Claude (review)*

(3) Low — the template-missing early return (a few lines below, when `store.exists(template_key)` is false) returns 500 *before* the write loop, so `_write_t0` is set but the write-timing record block is never reached and `write` is silently omitted. That's arguably fine (no write happened), but it's an asymmetry worth a comment: under `profile=True`, a 500 from a missing template yields `phase_timings` with read/index/aggregate but no `write`, whereas a 500 from a *failed* write loop yields a `write` key (see my other comment). Documenting which 500 paths carry `write` vs. not would make the contract the runner consumes unambiguous.

---

tests/test_lambda_handler.py
```diff
@@ -217,6 +217,83 @@ def test_k_eq_1_ragged_keyed_by_shard_key(self, handler_mod, monkeypatch):
...
+            handler_mod,
+            monkeypatch,
+            worker_phase_timings=None,
+            chunks=[((0,), pd.DataFrame(), {})],
+        )
+        event = _base_event(_healpix_config_dict())
+        event["child_order"] = 12
+        resp = handler_mod._handle_process(event, _context())
+        assert resp["statusCode"] == 200
+        assert cap["profile"] is False
```
espg, line 315: 🤖 *from Claude (review)*

(4) Medium — test coverage gaps for the phase-3 bridge. The two new tests cover only the happy path (write ran, no exception) and the profile-off path. Missing cases that exercise the actual branch logic of the new code:

- profile=True with an **empty** `chunk_results` (no write ran): assert `phase_timings` is present (read/index/aggregate from the worker) but `write` is **absent** — this is the "no-data path leaves `write` absent" contract the new comment claims, and nothing currently pins it. Easy: pass `chunks=[]` to `_patch`.
- profile=True where the **write loop raises** (make `write_dataframe_to_zarr` raise): pins whatever the intended behavior is for `write` on a failed write (see my comment on the record block). Right now `_patch` hard-wires no-op writers, so the raising path is never reached.
- profile=True + **template missing** (`store.exists.return_value = False`): asserts the 500 body and that `write` is omitted.

(5) Nit — `assert timings["write"] >= 0.0` is vacuously true for any `time.time()` delta, so it doesn't really test the bracket fired. The `set(timings) == {...}` assertion above already proves the key was added, so this line adds little; consider asserting the worker-seeded values pass through unchanged (e.g. `timings["read"] == 1.0`) instead, which the current test doesn't check.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of the phase-3 handler-bridge diff in `deployment/aws/lambda_handler.py` (+ its tests). Scope: only the write-phase timing bracket and the `profile` plumbing into the handler. The worker/runner sides are context, not under review here.

The core wiring is sound: `profile` is read with a safe default, forwarded into `process_shard`, and `_write_t0`/the record block are correctly gated so `_write_t0` is never `None` when used. The `write` key joins the same `phase_timings` sub-dict the runner rolls up via per-phase `max(...)`, and absent `write` keys are skipped cleanly by that rollup — so the contract matches.

Findings (none are blocking-merge on their own; (1) and (4) are worth resolving before ready-for-review):

- (1) Medium — `write` is recorded even when the write loop **raised** (the `except` sets `error` but doesn't return; control falls through to the record block). That logs a failed/partial write's time-to-failure as a real `write` duration, which then skews the runner's `worker_phase_max["write"]`. Intended or not, it needs a decision + a test. Inline on the record block.
- (2) Low — the `"phase_timings" in metadata` guard can only fail if the worker contract regresses (the worker always seeds it under profile); fine as defensive code, worth a one-word note.
- (3) Low — asymmetry across 500 paths: template-missing 500 omits `write`; failed-write 500 includes `write`. Document which 500s carry it.
- (4) Medium — test gaps: no coverage for profile=True with empty `chunk_results` (assert `write` absent), with a raising write loop, or with a missing template. The two added tests only hit the happy path and profile-off.
- (5) Nit — `assert timings["write"] >= 0.0` is vacuously true; assert pass-through of the worker-seeded values instead.

Posting as COMMENT only — not requesting changes.

---

espg:
🤖 *from Claude*

Phase 3 landed: the deployment-side handler bridge is wired, so `--profile` now surfaces `phase_timings` end-to-end in production.

- `deployment/aws/lambda_handler.py` reads `event["profile"]` (default `False`), forwards `profile=` into `process_shard`, brackets the handler-owned **write** phase, and merges `write` into the same `metadata["phase_timings"]` the runner already rolls into `worker_phase_max`. Recorded **only on a clean write** — no-data, missing-template, and failed-write paths leave `write` absent so the per-phase max rollup never folds in a time-to-failure. Default (no `profile` key) is byte-identical to before.
- New `TestProcessEventProfile` (7 cases, import-by-path) in `tests/test_lambda_handler.py`: profile forwarded + worker timings pass through + `write` added; default-off; and `write` absent on the empty-`chunk_results`, raising-write, and missing-template paths.
- Fresh-context adversarial self-review posted above; its diff-scoped findings (record write only on a clean write; cover the failure paths) were folded into the follow-up commit.

All three open questions are resolved per your [answer](https://github.com/englacial/zagg/pull/107#issuecomment-4813952501): (1) handler bridge landed (authorized), (2) `runner.py` size left as-is (&#34;ignore it for now&#34;), (3) `ruff format` drift left alone (&#34;ok&#34;; the enforced `ruff check` is clean). Full suite green locally: 838 passed, 1 skipped. All phases complete; holding in draft awaiting your review.

---

Issue #110, opened 2026-06-27
Lambda benchmarks integrated into our CI/CD pipeline


espg:
We have a target 'cost per shard' that we need to deliver. We are also making an *enormous* amount of changes to the zagg codebase. We need to be able to know and pinpoint which changes cause regressions in our compute pipeline (which is primarily lambda). We also want to know and track when we make performance enhancements, so we can gauge the relative impact of architecture changes.

This means a CI/CD pipeline that executes against a real shard, on aws lambda, with real data written to an S3 bucket. We should get the seconds (and cost) of execution, which needs to be tracked-- and plotted, and posted. Some of the tracking machinery is in flight with #100 and #107 already. We can also template a bit off of https://github.com/englacial/xopr/tree/main/.github/workflows and https://github.com/englacial/xopr/tree/main/.github/scripts  (which has some workflows for doing similar tracking, but of users rather than performance). We can also template a bit off of https://github.com/englacial/tofu-jupyter and https://github.com/espg/tofu_era , which have some examples of handling aws creds in a sensible way.

I expect we'll use SOPS for dealing with and storing creds (unless you have a compelling alternative we should look at). 

There's (at least) two major phases of this: the code that runs and does all of this (script, git actions, workflows, etc), and the infrastructure setup (permissions in github / aws ; i.e., commands that need to be run or configuration options that need to be clicked). We'll want to address both-- that is, code in a PR, and ALSO a setup guide to follow with step by step instructions for what's needed to set this up, why, and where to execute that setup... as well as how to confirm that things are up and configured correctly.

espg:
We'll also need specific templates and areas to test on (i.e., morton indices or rect boxes), which I'll provide as we move the discussion along and transition to planning this.

espg:
🤖 *from Claude*

Scoping the Lambda-benchmark CI/CD (discuss pass — no code yet). The two-part split you named (the running code + a setup guide for the GitHub/AWS permissions) is right; below are the design axes, 2–3 options each, and the questions I need before a plan.

[...]

PR #112, opened 2026-06-27
Lambda benchmark CI/CD (#110)

espg:
Refs #110.

Lambda-benchmark CI/CD: dispatch the single densest shard over the NEON SERC AOP box to AWS Lambda on PRs and merges, harvest seconds + cost, and track **cost/shard** and **cost/100 km²** over merge history. Implements the plan locked in the issue thread.

Run from a local agent (not a routine) so the GitHub Actions / workflow files land without push-scoping friction.

## Locked design
- **Creds:** GitHub OIDC → AWS IAM role assumption; no stored long-lived AWS secret. Earthdata Login (for the NSIDC read creds the orchestrator forwards to Lambda) via repo secrets.
- **Triggers:** non-fork PR to `main` (non-draft) → run + comment, **not** retained; merge to `main` → run + **retain** to a parquet series on the `benchmarks` data branch + render charts; fork PRs → `benchmark` label or `/benchmark`, **write/admin-gated**, run the fork's package against the **base** harness (base-pin).
- **Pins:** shard maps + densest shard committed under `tests/data/benchmark/`; dispatch just refs the map + shard. A separate drift test rebuilds from CMR and fails (tie-tolerantly) if the densest shard moves.
- **Results:** retained series = parquet; ephemeral PR payload = markdown comment. Two charts (cost/shard + runtime; cost/100 km² + runtime), embedded in the docs by raw URL from the `benchmarks` branch.
- **Matrix:** arm64, 2 GB, one shard/target, capped by the 720 s timeout. gain_bias + tdigest × {HEALPix order 11 & 10} + {rect 3 km ≈o11 / 6 km ≈o10} = 8 targets.

## Phases — all complete
- [x] **Phase 1** — benchmark runner + pinned HEALPix targets (`bench_metrics.py`, `run_benchmark.py`; order-11/10 shard maps; `targets.json`).
- [x] **Phase 2** — results store + plotting (`update_series.py` merge-only idempotent parquet; `plot_series.py` two dual-axis figures + lean `benchmark` extra).
- [x] **Phase 3** — workflows + OIDC (composite `run-benchmark`; `lambda-benchmark.yml` PR-comment + merge-retain; `lambda-benchmark-command.yml` fork on-demand, write/admin-gated). actionlint-clean.
- [x] **Phase 4** — setup guide (`docs/deployment/benchmark-cicd.md`: OIDC provider, scoped IAM role/trust, S3, `benchmarks` branch, `BENCHMARK_*` vars + `EARTHDATA_*` secrets, `benchmark` label, per-step verification) + a docs charts page; Earthdata creds wired through.
- [x] **Phase 5** — rect configs matched to HEALPix o11/o10 + the shard-map drift test (rebuild from CMR, tie-tolerant) + `benchmark-drift.yml` (weekly + dispatch, native x86_64 with the spherely fork wheel).

## Extra (per @espg in-thread)
- **Base-pin the fork path** (`7d62086`): the on-demand fork run checks out base `main` for the trusted harness (run_benchmark / targets / shard maps / configs) and overlays **only** the fork's `src/zagg` (the code under test), so a fork can't rewrite the benchmark definition or cost math.

## How it was tested
- `pytest tests/test_benchmark.py` — 21 pass (no AWS/network; dispatch mocked via `--dry-run`; plot tests `importorskip` matplotlib).
- Drift test `tests/test_benchmark_shardmap.py` — run against **live CMR** locally (healpix/mortie + rect/spherely): 4/4 pass; default-skips without `ZAGG_BENCHMARK_DRIFT=1`.
- `ruff check` / `ruff format --check` clean; **actionlint** clean on all three workflows; `mkdocs build --strict` clean.
- Shard maps built over the NEON SERC box, full mission 2018-10-13..2025-06-01: healpix o11 (22 shards / densest 44), o10 (9 / 50); rect 3 km (16 / 44), 6 km (4 / 47). Rect built with the exact-S2 spherely backend.
- **Not exercised here:** the live Lambda dispatch + `benchmarks`-branch publish (need the Phase-4 AWS/GitHub setup). Workflows are static-validated only.

## Scope / authorization note
Adds `.github/workflows/`, `.github/scripts/`, `.github/actions/`, and AWS-setup docs — walled off by CLAUDE.md §1 in general, but issue #110 names this work, so it's authorized. Workflows run under GitHub Actions + the OIDC-assumed AWS identity; nothing runs live AWS from an agent.

## Resolved questions
- Temporal pin `2018-10-13 .. 2025-06-01`; rect 12 km tile @10 m, chunk 300/600, inner 75 (K=16/64); data branch `benchmarks`; Pages → embed charts in docs by raw URL; base-pin the fork path → done.

_No open questions. All adversarial-review findings (Phases 1–5) folded; threads replied._


tests/data/benchmark/targets.json
```diff
@@ -0,0 +1,60 @@
...
+    "short_name": "ATL03",
+    "version": "007",
+    "provider": "NSIDC_CPRD",
+    "backend": "mortie",
+    "footprint": "swath"
+  },
+  "shardmaps": {
+    "healpix_o11": {
+      "path": "shardmaps/sm_healpix_o11.json",
+      "shard_key": 5347394812217655307,
```
espg, line 21: 🤖 *from Claude (review)* The `healpix_o11` pin sits on a **4-way tie** for densest: shards `5347404707822305291`, `5347398660508352523`, `5347398385630445579`, and this one (`5347394812217655307`) all hold 44 granules. The pin is correct *today* only because `select_densest_shard` breaks the tie by lowest key — but that undercuts the manifest's stated goal ("the densest cell ... so cost/runtime deltas track code changes, not data drift"): a CMR rebuild that nudges any of these four counts by ±1 will flip the selection to a *different* shard with different `total_obs`/runtime, surfacing as a phantom cost/runtime regression that is pure data drift. (Granule *count* is also a coarse density proxy — equal-granule shards can differ a lot in observations.) o10 is robust here (unique winner at 50 vs next 48), so consider either pinning the o11 benchmark on a uniquely-densest criterion (e.g. break ties by max obs once known, or drop o11 to derived-only) or explicitly documenting that the phase-5 drift test must tolerate a tie-driven reselection.

espg: 🤖 *from Claude (review)* Folded in 10fe8cb: documented the 4-way o11 tie in targets.json. The benchmark dispatches the static pinned shard_key, so the retained series is stable regardless of CMR drift; the phase-5 drift test will compare granule count within ±1 rather than exact shard_key so a tie-reselection doesn't false-alarm.

.github/scripts/update_series.py
```diff
@@ -0,0 +1,83 @@
...
+    """Append records, replacing any prior rows for the same (commit, target).
+
+    Keeping the last write makes a merge re-run idempotent instead of duplicating
+    a point in the plotted history.
+    """
+    new = records_to_frame(records)
+    # Avoid concat with an all-empty frame (pandas FutureWarning on dtype union).
+    combined = new if existing.empty else pd.concat([existing, new], ignore_index=True)
+    if not combined.empty:
+        combined = combined.drop_duplicates(subset=["commit", "target"], keep="last").reset_index(
```
espg, line 52: 🤖 *from Claude (review)* The module docstring promises "Only **merge** runs are retained," but neither `append_records` nor `main` enforce it — whatever `records` is passed gets stored. Verified: `append_records(load_series('nope'), [<event='pr' row>])` retains the `pr` row. Because the dedup here is keyed on `(commit, target)` and ignores `event`, a later non-merge run for an already-merged commit would *evict* the retained merge row (`keep='last'`), and `plot_series._merge_history` filters `event=='merge'` — so that point would silently vanish from the chart. The benchmark workflow is Phase 3 (not yet wired), so today the only guard is "the caller must only call this on merges." Consider enforcing the documented invariant at the boundary, e.g. `records = [r for r in records if r.get('event') == 'merge']` in `main` (or filtering in `append_records`), so the retained-series guarantee is robust to how the Phase-3 workflow ends up dispatching.

espg: 🤖 *from Claude (review)* Folded in ceaecce: update_series.main now filters to event=='merge' at the CLI boundary (logs any drops, never silently) so a stray PR record can't evict a retained merge point via the (commit,target) dedup; added a test. Also guarded the unreachable n==0 ZeroDivisionError in plot_series.make_figure.

espg:
> Temporal pin. I pinned the granule window to 2018-10-13 .. 2025-06-01 (full mission to a fixed past date) so the densest-shard target is stable. OK, or prefer a specific cycle / the 2025 range?

correct range -- we need a full slice across years.

> Rect sizing (Phase 5). The rect AOI must tile evenly (deployed Lambda won't auto-pad). Plan: a 12 km UTM tile at 10 m with chunk [300,300] (3 km, ≈ order-11's 10.1 km²) and [600,600] (6 km, ≈ order-10's 40.5 km²), chunk_inner [75,75] to mirror HEALPix K (16 / 64). Good, or a different rounded size?

yep, that's good

> Series data branch name — benchmarks vs gh-pages?

I think you already answered this -- it should be benchmarks (gh-pages is taken)

espg:
🤖 *from Claude*

Thanks — locking in your confirmations:
- **Temporal pin**: `2018-10-13 .. 2025-06-01` (full multi-year slice) stays.
- **Rect sizing (phase 5)**: 12 km UTM tile @ 10 m, chunk `[300,300]` (≈o11) / `[600,600]` (≈o10), `chunk_inner [75,75]` (K=16/64).
- **Data branch**: `benchmarks` — already wired as the retain target in phase 3.

Phase 3 (OIDC workflows) just landed (`7f47bbd`), actionlint-clean. One thing to flag on **Pages**: the repo already deploys mkdocs to Pages via the Actions source (`docs.yml`), and a repo gets one Pages site. Rather than fight for that slot, phase 3 publishes the parquet series **and** the rendered charts (index.html + PNGs) to the `benchmarks` branch under `site/`. The phase-4 setup guide will lay out how to surface them (point Pages at the `benchmarks` branch — which would replace the docs site — vs. embedding the charts under the docs site). Will leave that as a setup-time choice with a recommendation.

.github/workflows/lambda-benchmark-command.yml
```diff
@@ -0,0 +1,99 @@
...
+                return core.setOutput('should_run', 'false');
+              }
+              prNumber = context.payload.issue.number;
+              await github.rest.reactions.createForIssueComment({
+                owner, repo, comment_id: context.payload.comment.id, content: 'rocket',
+              });
+            } else {
+              // pull_request_target labeled: applying a label already requires triage/write,
+              // so the actor is inherently a maintainer; just gate on the label name.
+              if (context.payload.label.name !== 'benchmark') return core.setOutput('should_run', 'false');
```
espg, line 45: 🤖 *from Claude (review)*

**Security (MUST-FIX): the `labeled` path never checks the actor, only the label name — so triage-level (non-maintainer) users can trigger a creds-bearing run on untrusted fork code.**

The comment justifies this as "applying a label already requires triage/write, so the actor is inherently a maintainer." But **triage permission can add/remove labels and is not write/maintain**. Issue #110's intent is that *a MAINTAINER* vouches for fork code before it runs with the OIDC AWS role (real $$$, short-lived creds). A triage-role user is not that.

Unlike the comment path (which at least checks `author_association`), this branch gates on `label.name` alone and ignores `context.actor` entirely. Recommend an explicit permission check on **both** paths, e.g.:
```js
const { data: perm } = await github.rest.repos.getCollaboratorPermissionLevel({ owner, repo, username: context.actor });
if (!["admin", "write"].includes(perm.permission)) return core.setOutput("should_run", "false");
```
That is the only check that actually maps to "maintainer," and it removes the reliance on label-permission semantics.

.github/workflows/lambda-benchmark-command.yml
```diff
@@ -0,0 +1,99 @@
...
+          script: |
+            const { owner, repo } = context.repo;
+            const maintainers = ['OWNER', 'MEMBER', 'COLLABORATOR'];
+            let prNumber;
+
+            if (context.eventName === 'issue_comment') {
+              // PR comments only; body must contain /benchmark; commenter must be a maintainer.
+              if (!context.payload.issue.pull_request) return core.setOutput('should_run', 'false');
+              if (!context.payload.comment.body.includes('/benchmark')) return core.setOutput('should_run', 'false');
+              if (!maintainers.includes(context.payload.comment.author_association)) {
```
espg: 🤖 *from Claude (review)*

**Security (MUST-FIX): `author_association` in OWNER/MEMBER/COLLABORATOR is not equivalent to write/maintainer permission.**

- `COLLABORATOR` = "has been invited to collaborate" at *any* level, including **read or triage**.
- `MEMBER` = belongs to the owning org, regardless of their permission on *this* repo (an org member with only read here still reports `MEMBER`).

So a read/triage-level collaborator or a low-privilege org member can comment `/benchmark` and pass this gate, then mint the OIDC AWS role against checked-out fork code. This is the well-known github-script gotcha — `author_association` is a relationship label, not a permission. Replace with an explicit `repos.getCollaboratorPermissionLevel(context.actor)` check for `admin`/`write` (same fix suggested on the label branch below).

.github/actions/run-benchmark/action.yml
```diff
@@ -0,0 +1,85 @@
...
+        fi
+        uv run python .github/scripts/run_benchmark.py \
+          --targets tests/data/benchmark/targets.json \
+          "${target_flags[@]}" \
+          --store-prefix "${{ inputs.store-prefix }}" \
+          --region "${{ inputs.aws-region }}" \
+          --function-name "${{ inputs.function-name }}" \
+          --event "${{ inputs.event }}" \
+          --commit "${{ inputs.commit }}" \
+          --ref "${{ inputs.ref-label }}" \
```
espg: 🤖 *from Claude (review)*

**Security (script-injection sink): `${{ inputs.ref-label }}` is interpolated straight into the `run:` shell, and it carries an attacker-controlled branch name.**

`ref-label` is fed from `github.head_ref` (lambda-benchmark.yml) and `pr.head.ref` (the on-demand workflow) — i.e. a *fork* branch name on the on-demand path. Git allows `$`, backticks, `;`, `|`, `"`, `()` in ref names, so a branch like `` $(curl …|sh) `` is evaluated by bash when this line expands (the `${{ }}` is templated *before* the shell sees it; the surrounding double-quotes do not stop `$(...)`/backticks).

You already do the right thing for `TARGETS` (passed via `env:` and referenced as a shell var). Do the same here — pass `ref-label` (and, for consistency, `pr-number`) through `env:` and reference `"$REF_LABEL"`, never inline `${{ }}`.

Mitigating note: on both current triggers the workflow already checks out and executes the same untrusted tree, so this is defense-in-depth today rather than a new exec path — but it is a recognized anti-pattern, breaks benignly on metachar branch names, and becomes a live exfil vector the moment the harness is refactored to run from the base tree. Worth fixing now.

.github/workflows/lambda-benchmark-command.yml
```diff
@@ -0,0 +1,99 @@
...
+
+on:
+  issue_comment:
+    types: [created]
+  pull_request_target:
+    types: [labeled]
+
+permissions:
+  contents: read
+  id-token: write
```
espg: 🤖 *from Claude (review)*

**Least-privilege (optional): `id-token: write` is granted at the top level, so the `check` gate job gets it too — but the gate only runs github-script and never assumes the role.** Only the `run` job needs OIDC. Consider dropping `id-token: write` from the top-level block and granting it per-job on `run:` only (the `check` job needs just `pull-requests: write` for the reaction, and arguably not even that). Keeps the credential-minting capability off the job that processes the trigger payload.

.github/workflows/lambda-benchmark.yml
```diff
@@ -0,0 +1,126 @@
...
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+  id-token: write
+  pull-requests: write
+
+jobs:
+  pr-comment:
```
espg, line 29: 🤖 *from Claude (review)*

**Cost (optional): `pr-comment` has no `concurrency` group.** Rapid pushes to a PR (`synchronize`) will stack multiple real-Lambda benchmark runs in parallel — each costs money and they race on the single upsert comment. `merge-publish` already serializes; mirror that here with a per-PR group, e.g.:
```yaml
concurrency:
  group: lambda-benchmark-pr-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

.github/workflows/lambda-benchmark-command.yml
espg: 🤖 *from Claude (review)* Fixed in ecfea72 (covers both the comment and the label path): the gate no longer trusts `author_association`/label-add at all — it calls `repos.getCollaboratorPermissionLevel(context.actor)` and only proceeds when the actor has `write` or `admin`. So triage-label and read/triage MEMBER/COLLABORATOR triggers can no longer hand fork code the AWS role.

.github/actions/run-benchmark/action.yml
espg: 🤖 *from Claude (review)* Fixed in ecfea72: every input the dispatch step uses (REF_LABEL, COMMIT, EVENT, PR_NUMBER, STORE_PREFIX, …) now flows through `env:` and is referenced as "$REF_LABEL" etc. — no `${{ }}` interpolation remains in the run-script body, so an attacker-controlled branch name can't break out of the argument.

.github/workflows/lambda-benchmark-command.yml
espg: 🤖 *from Claude (review)* Fixed in ecfea72: `id-token: write` is removed from the workflow-level block and granted only to the `run` job (the gate/`check` job now has just contents:read + pull-requests:write). Also added per-PR `concurrency` (cancel-in-progress) to both the on-demand `run` job and the auto pr-comment job so rapid pushes don't stack billed Lambda runs.

espg:
> Pages surfacing

Strong preference for (b) embed the charts under the docs site (a benchmark/ page the docs build pulls in)

> Fork-run trust. On a maintainer-triggered fork run, the benchmark harness + zagg code execute from the fork tree with the (minimally-scoped) role. That's the accepted cost of benchmarking a fork (matches the xopr preview pattern), gated to write/admin actors. OK as-is, or do you want the harness/targets pinned to base on the fork path too?

...can you explain this option in more detail : "...the harness/targets pinned to base on the fork path too"?

espg:
🤖 *from Claude*

**Pages → (b) done** (`5f5fbaa`): added `docs/deployment/benchmark.md` that embeds the two charts by their raw URL on the `benchmarks` branch (so they refresh live as merges update the branch — no docs rebuild, and `docs.yml` is untouched), plus a `benchmark-cicd.md` setup guide; both wired into the mkdocs nav. `mkdocs build --strict` is clean.

**On "pin the harness/targets to base on the fork path":**

Today, a fork run checks out the fork's PR head and runs *everything* from it — not just the zagg package under test (`src/zagg`), but also the orchestration script (`.github/scripts/run_benchmark.py`), the pinned `targets.json`, the committed shardmaps, and the benchmark configs. A malicious fork could therefore rewrite *what the benchmark does* (point at a giant shard, change the cost math, or alter the harness to print/exfiltrate the short-lived OIDC creds). The guardrails are: a write/admin maintainer must opt in per-PR, and the role can only invoke the one Lambda + write the one bucket prefix — so blast radius is small, but the fork still controls the executed code.

**"Pinned to base"** means splitting trusted vs. measured:
- **Harness from base** (trusted `main`): `run_benchmark.py`, `bench_metrics.py`, `targets.json`, the shardmaps, the benchmark configs — checked out from the base repo, not the fork.
- **Only `src/zagg` from the fork**: install/point the run at the fork's package code, which is the thing we actually want to measure.

Mechanically: checkout base at root, then overlay the fork's `src/zagg` (e.g. `uv pip install` the fork's package, or PYTHONPATH to a second checkout of just `src/zagg`), and run base's `run_benchmark.py` against base's pinned targets.

**Trade-offs:**
- ✅ A fork can't rewrite the benchmark definition or the orchestration/creds-handling — meaningfully smaller attack surface even before the maintainer gate.
- ✅ Arguably *more correct* as a perf benchmark: you measure compute-path deltas, not benchmark-definition drift.
- ❌ More moving parts (two checkouts + an install overlay), and a fork PR that *legitimately* changes the harness/configs/targets (e.g. adds an aggregator or a new target) wouldn't be exercised on the fork path until merged — you'd only see it via a same-repo run.

**My lean:** worth doing, since the whole point of the fork path is "measure untrusted code safely." But it's an enhancement, not a blocker — the current gate + scoped role is a reasonable v1. I'd suggest landing this PR as-is and doing base-pinning as a fast follow (its own small PR) so it gets focused review. Want me to file that as a follow-up issue, or fold it into this PR before merge?

espg:
do the base pin as part of this PR while we're thinking about it

docs/deployment/benchmark-cicd.md
```diff
@@ -0,0 +1,283 @@
...
+## 2. Create the scoped IAM role
+
+The role is assumable **only** by this repo's Actions, and can do **only** what
+the benchmark needs. Two policy documents: a trust policy (who can assume) and a
+permission policy (what it can do).
+
+### 2a. Trust policy — pin to the repo
+
+`trust.json` (restrict `sub` to your repo; the `main` and `pull_request`
+conditions keep it to the branch/PR contexts the workflows actually run in):
```
espg: 🤖 *from Claude (review)*

The prose says "the `main` and `pull_request` conditions keep it to the branch/PR contexts," but the `trust.json` below has no such conditions — the `sub` is a bare wildcard `repo:englacial/zagg:*` (line 87), which matches **every** context in the repo (any branch, tag, environment, and the generic `pull_request` context), not just main/PR. The only thing keeping a fork-PR auto-run from assuming the role is the `head.repo.full_name == github.repository` guard in `lambda-benchmark.yml`, not the trust policy.

Either correct the prose to say the `sub` is scoped to the repo (all contexts) and that workflow-level gating handles branch/PR/fork distinctions, or actually add the StringLike conditions the prose describes (e.g. pin to `repo:englacial/zagg:ref:refs/heads/main` + `repo:englacial/zagg:pull_request`) for defense-in-depth. As written the doc over-claims the policy's scoping.

tests/test_benchmark_shardmap.py
```diff
@@ -0,0 +1,90 @@
...
+    sm_meta = MANIFEST["shardmaps"][sm_key]
+    committed = json.loads((BENCH / sm_meta["path"]).read_text())
+    backend = committed["metadata"]["backend"]
+    if backend == "spherely":
+        pytest.importorskip("spherely")
+
+    cfg = load_config(str(_config_for_shardmap(sm_key)))
+    grid = from_config(cfg)
+    cmr, temporal, aoi = MANIFEST["cmr"], MANIFEST["temporal"], MANIFEST["aoi"]
+    parts = load_polygon(str(REPO / aoi["file"]))
```
espg: 🤖 *from Claude (review)*

The drift check reads `REPO / aoi["file"]` → `<repo-root>/AOP_NEON.geojson` (`aoi.file` is `"AOP_NEON.geojson"` in `targets.json`), but that file is **not tracked in git** (`git ls-files` shows no geojson anywhere; it only exists locally, untracked). The `benchmark-drift.yml` workflow does a fresh `actions/checkout`, so `load_polygon(...)` will raise `FileNotFoundError` and every parametrized case fails immediately.

Nothing before phase 5 consumed `aoi.file` (the benchmark harness dispatches off the committed shard maps, not the AOI), so this dangling reference was inert until this commit added the first consumer. Commit the AOI polygon (ideally under `tests/data/benchmark/` next to the maps it describes, updating `aoi.file` to match) so the drift workflow can resolve it.

espg: 🤖 *from Claude (review)* Fixed in cca0dc1: committed the AOI as `tests/data/benchmark/AOP_NEON.geojson` (it was untracked at repo root) and the drift test now resolves `aoi.file` relative to the manifest dir, so the benchmark-drift workflow's fresh checkout finds it. Re-ran the drift test against live CMR after the change — 4/4 pass.

docs/deployment/benchmark-cicd.md
espg: 🤖 *from Claude (review)* Fixed in 7d62086: reworded the trust-policy section to accurately describe the bare `repo:englacial/zagg:*` sub wildcard (repo-scoped; a fork's sub can't match), note that branch/PR/fork-context gating lives in the workflows, and show how to optionally tighten with explicit subs — flagging that the on-demand fork path runs under `pull_request_target` (base-ref sub) so it mustn't be over-restricted.

espg:
🤖 *from Claude*

All 5 phases are complete and **CI is green** (test 3.12 + 3.13, ruff, build, build-arm64, build-x86_64 all pass; the benchmark jobs correctly skip on a draft PR). Every phase had a fresh-context adversarial review and all findings are folded + threads replied. Base-pin of the fork path (your in-thread ask) is in.

I'm **leaving the PR as a draft**: CLAUDE.md §2 says drop 'draft' when done, but this repo uses GitHub's native draft state and §6 reserves *mark-ready-for-review* for you — rather than guess between the two, I'll let you flip it to ready when you're satisfied (or say the word and I'll do it).

What's **not** exercised by CI: the live Lambda dispatch + `benchmarks`-branch publish — those need the one-time AWS/GitHub setup in `docs/deployment/benchmark-cicd.md` (OIDC role, `BENCHMARK_*` vars, `EARTHDATA_*` secrets, the `benchmarks` branch, the `benchmark` label). The workflows are static-validated (actionlint) only until that's in place.

PR #113, opened 2026-06-27
HEALPix ShardingCodec output: decouple write/dispatch shards from 64×64 read chunks

### LINKED ISSUE #108: Evaluate zarr ShardingCodec for HEALPix output: decouple large write/dispatch shards from 64×64 read chunks
## Goal

Decouple the **write/dispatch granularity** from the **read granularity** for HEALPix outputs:

- **Writer** wants the largest dispatch shard a Lambda can handle — currently `parent_order=10`, would like to push to **order 9 or 8** (fewer invocations, less redundant per-granule reads).
- **Reader** wants **64×64 chunks** = `4^6 = 4096` cells = **chunk order 13** (for `child_order=19`).

These are reconciled by `chunk_inner` already: shard = dispatch unit (order 8–10), inner chunk = 64×64 (order 13), so `K = 4^(13 - parent_order)` chunks per shard (64 at order 10, 256 at order 9, **1024 at order 8**). The open question is the **storage form** of those K chunks: today they are **K independent regular chunk objects**; this issue is to evaluate moving them into **one `ShardingCodec` shard object per dispatch shard**.

## Background / current state

- We have **never used zarr `ShardingCodec`** — every `chunk_grid` is `name="regular"` (`grids/base.py`, `grids/healpix.py`). `chunk_inner` produces *finer regular chunks*, i.e. **more, smaller objects**, the opposite direction from sharding (**fewer, bundled objects**).
- `chunk_inner` itself only just started working: #103 (closed #99) fixed the Lambda **setup** path that wasn't forwarding `chunk_inner`, so prior K>1 runs emitted a coarse template and failed with `block index out of bounds`. A fresh K>1 run is needed to inspect the intended layout (dense arrays chunked at `cells_per_chunk = 4^(child−chunk_order)`, `shards=None`).

## Why there's no deep barrier to sharding

1. **The writer already goes through zarr's codec pipeline.** Workers write with `array.set_block_selection(block_idx, values)` on an opened array (`processing/write.py`), not by `put`-ing raw chunk objects. If the array were *created* with a sharding codec, writes route through it automatically — so adoption is primarily a **template-emission change** (`ArraySpec` codecs: add sharding with `shard_shape = cells_per_shard`, inner `chunk_shape = cells_per_chunk`).
2. **One worker per dispatch shard maps to one zarr-shard.** If `zarr-shard == dispatch-shard`, each worker writes exactly one whole shard object, alone — the canonical single-writer sharding pattern, no read-modify-write, no cross-worker contention. Empty inner chunks (track doesn't cross them) are omitted from the shard index, preserving sub-shard sparsity inside the object.

### The one implementation change required

The K>1 worker currently writes its inner chunks as **K separate `set_block_selection` calls** (the `iter_chunks` loop). Against a sharded array, K writes into the *same* shard object would each trigger a read-modify-write — racy/slow. To shard cleanly the worker must write the **whole shard slab in one `set_block_selection`** over the shard-aligned region (it already holds all K chunks in memory, since it reads the shard's granules once).

## Regular chunks vs ShardingCodec

| | chunk_inner (current) | ShardingCodec |
|---|---|---|
| storage | K regular chunk objects/shard | 1 shard object/shard |
| object count | high (empties absent) | low (~K× fewer) |
| sub-shard sparsity | absent objects | shard-index entries |
| discovery | trivial `list_prefix` | needs shard-index reads |
| reader 64×64 access | one object/chunk | byte-range partial-decode within shard |

## Performance: does sharding speed up writes / reduce Lambda runtime?

Short answer from analysis: **not meaningfully at current shard sizes.** To verify, not assert — needs measurement.

- Writes are already **concurrent** (`async.concurrency=128`), so K small PUTs ≈ one shard PUT in wall-clock; sharding adds only a small index. Roughly neutral.
- Lambda time is dominated by **granule reads + aggregation**, not writes (cf. #100 timings: order-10 ~128 s/cell vs order-13 ~0.8 s/cell — read+agg bound). Output codec won't move the total.
- **Where sharding *could* help writes:** at order 8 (`K=1024`) a densely-crossed shard can produce more populated inner chunks than the 128-concurrency cap → multiple PUT waves; one shard PUT collapses them. Benefit grows with shard size but stays second-order to read+agg.
- **Genuine wins are elsewhere:** object count ↓ ~K, S3 request count/cost ↓, and the reader gets 64×64 partial-decode from few objects instead of enumerating millions of tiny chunk objects at global/dense scale. Sharding is what makes "push dispatch to order 8" viable *without* drowning the reader in objects.
- **Lambda-speed levers are orthogonal:** larger dispatch shards (fewer invocations, less redundant per-granule reads) and faster granule IO (#43) — not the output codec.

## Tasks

- [ ] Produce a working `chunk_inner=13` run (post-#103) and inspect the emitted layout: confirm dense arrays chunk at `cells_per_chunk` with `shards=None`; quantify populated-inner-chunks-per-shard (intra-shard occupancy measured ~7.5%, clustered along track) and object count.
- [ ] Confirm `pydantic_zarr.experimental.v3.ArraySpec` can express a sharding codec config (inner `chunk_shape` + inner codecs).
- [ ] Offline prototype: emit a sharded template for a tiny synthetic array and verify a single whole-shard `set_block_selection` write produces one shard object with empty inner chunks omitted, byte-identical read-back, and 64×64 partial-decode.
- [ ] Worker change: write the whole shard slab in one `set_block_selection` (gate for the sharded path).
- [ ] Benchmark write-phase time and object/request counts: regular vs sharded at parent_order 10 / 9 / 8; reader time for 64×64 access. **Measure the write-phase fraction of Lambda runtime first** to size the prize.
- [ ] Decide gating policy: regular chunks for AOI-scale products, sharded for global/dense.

## Related

- Builds on `chunk_inner` (#82 / #30 item 3) and the fix in #103.
- Sub-shard sparsity discovery: once sharded, discovering/skipping empty *inner* chunks without full decode becomes relevant (the upstream zarr-python sparse-read work, zarr PR #4028). Not actionable here until we shard.
- Future **mixed-resolution** HEALPix breaks the single-resolution uniformity assumption; cleanest path is one zarr array per resolution tier (each internally uniform, hence independently shardable). Tracked separately if/when pursued.

espg:
Closes #108

## What

Adds an opt-in zarr `ShardingCodec` storage form for HEALPix and rectilinear output. It bundles the `K = (shard cells) / (inner-chunk cells)` inner read-chunks of one dispatch shard into **one shard object** instead of K independent regular chunk objects, with empty inner chunks omitted from the shard index (sub-shard sparsity preserved *inside* the object). This is what makes "push dispatch to a coarser order" viable without drowning the reader in millions of tiny chunk objects.

The knob lives **on the grid/chunk block next to `chunk_inner`** (design locked by @espg in the issue thread), defaults **off**, and is only valid when `chunk_inner` gives K>1 (validated, errors otherwise).

## Approach

- **Geometry** (unchanged, set by `chunk_inner`): shard = dispatch unit (HEALPix `parent_order` / rect `chunk_shape`) = zarr **shard**; inner chunk = the 64×64 read chunk (HEALPix `chunk_inner` order / rect `chunk_inner` shape) = zarr **inner chunk**.
- **`emit_template`**: the dense per-cell `ArraySpec`s get a `sharding_indexed` codec — outer chunk grid widened to the whole shard, inner `chunk_shape` = the read chunk. Built **by hand** with **explicit bytes-only inner codecs** (the two prototype caveats: `ArraySpec.from_array` drops the sharding codec in pydantic-zarr 0.10.0; `zarr.create_array` silently injects zstd level-0). 1-D coords and `resolution: chunk` companions stay regular.
- **Worker**: on the sharded path the K inner chunks are assembled into one shard-wide slab per dense array and written in a single shard-granular `set_block_selection` (avoids read-modify-write on the shared shard object). Placement is grid-agnostic via `grid.shard_slab_shape()` + `grid.shard_local_region(block_index, shard_key)`, so HEALPix (1-D) and rectilinear (2-D) share one write path. The regular (non-sharded) path is byte-identical.

## Phases (all complete)

- [x] **Phase 1 — HEALPix sharded `emit_template`** (`0c5b25e`). `sharded_array_spec` helper in `grids/base.py`; `sharded=` flag on `HealpixGrid` with K>1 validation; sharding codec applied to dense per-cell arrays in `_spec`. Round-trip regression `tests/test_grids.py::TestSharded`: one shard object per dispatch shard, empty inner chunks omitted, byte-exact read-back, narrow read-chunk partial decode, flag-off byte-identity, bytes-only inner codecs, K==1 error.
- [x] **Phase 2 — Worker whole-shard write** (`d8a5e8a`). `write_shard_to_zarr`; runner + lambda handler route sharded grids to it; regular path unchanged. End-to-end `tests/test_processing.py::TestWriteShardToZarr` proves sharded == regular byte-for-byte + one-shard-object. Also folds the phase-1 self-review (reverted an unrelated `h_li`→`h_mean` test-helper scope creep).
- [x] **Phase 3 — Config flag + validation + accessor** (`ba4be30`). `get_sharded` accessor (mirrors `get_layout`); `from_config` threads it through; default off; K==1 surfaces as a pre-deployment error. Tests in `TestFromConfig` + `TestOutputHelpers` incl. flag-off byte-identity and the K==1 error.
- [x] **Phase 4 — Rectilinear engine** (`1766108`). `sharded` on `RectilinearGrid` (flag + K>1 validation + `shard_slab_shape`/`shard_local_region` + sharding codec in `_spec`); wired into rect's `from_config`. Refactored `write_shard_to_zarr` to grid-agnostic block-based placement (HEALPix path stays byte-identical — verified). Tests in `tests/test_rectilinear.py::TestSharded`.
- [x] **Phase 5 — Docs + follow-up note** (`fbf88cb`). `docs/sharding.md` (the sharded layout, config, K==1 rule, reader note) + mkdocs nav entry.
- [ ] **Follow-up (NOT in this PR — for @espg to open):** open a follow-up issue/PR for the #79/#81 read helpers to consume sharded stores (shard-index / byte-range reads instead of `list_prefix` enumeration). Reader side is deliberately deferred per the locked design; this PR is writer-side + round-trip read-back only. (Opening that issue is a side-effecting action, so it is left for @espg rather than done here.)

## How tested

Local (zagg is pure-Python, no build): `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on all touched files; `codespell` clean on the new docs. Test runs: `tests/test_grids.py` (64), `tests/test_rectilinear.py` (45), `tests/test_processing.py` (165, incl. the slow sharded end-to-end), `tests/test_config.py`/`test_runner.py`/`test_runner_concurrency.py`/`test_lambda_handler.py` — all green; no new failures. `tests/test_shardmap.py` is uncollectable in the local env (missing optional `stac_geoparquet` from the `catalog` extra) — pre-existing/unrelated, not touched.

The sharded round-trips are verified on a `MemoryStore` mirroring the target geometry: HEALPix (parent 4 / child 8 / chunk_inner 6 → K=16) and rect (shard 8×8 / inner 4×4 → K=4) each emit one `c/<shard>` object per populated shard, sparse-shard byte size < dense-shard, byte-exact read-back, and cell placement verified against the regular per-inner-chunk path.

## Notes

- **No new dependency** — sharding is stock zarr v3 (§4).
- **Benchmark** (write-phase fraction, object/request counts at parent_order 10/9/8) is deferred to live AWS, out of routine scope (ties to #110); offline round-trip tests land here. No dependency encoded (either merge order works).

## Questions for review

- Sharding is applied to **vector** per-cell arrays too (trailing payload dim kept whole, only the spatial axes widened to the shard). No config in the test suite ships a sharded *vector* field, so that combination is covered structurally (spec shape) but not by a full vector round-trip — flag if you'd prefer to restrict sharded to scalar-only for v1, or want a vector round-trip test added.
- `sharded` is deliberately **excluded from the grid `signature()`** (consistent with `chunk_inner`, which also isn't in the signature) since it is a storage form, not a spatial-layout change — the shard map stays reusable across sharded/regular. Flag if you'd rather it be part of the fingerprint.

### DIFF
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1170,6 +1170,18 @@ def get_layout(config: PipelineConfig) -> str:
     return config.output.get("grid", {}).get("layout", "fullsphere")
 
 
+def get_sharded(config: PipelineConfig) -> bool:
+    """Return whether the output grid uses ShardingCodec storage (issue #108).
+
+    The ``sharded`` knob lives on the grid/chunk block next to ``chunk_inner``
+    (mirroring its accessor), default ``False``. When ``True`` the grid bundles a
+    dispatch shard's K inner chunks into one zarr shard object instead of K
+    independent regular chunk objects; it is only valid when ``chunk_inner`` gives
+    K>1 (the grid raises otherwise, validated before deployment).
+    """
+    return bool(config.output.get("grid", {}).get("sharded", False))
+
+
 def get_store_path(config: PipelineConfig) -> str | None:
     """Return the store path from the output config, or None.
 
--- a/src/zagg/grids/__init__.py
+++ b/src/zagg/grids/__init__.py
@@ -4,7 +4,7 @@ from __future__ import annotations
 
 import warnings
 
-from zagg.config import PipelineConfig, get_child_order
+from zagg.config import PipelineConfig, get_child_order, get_sharded
 from zagg.grids.base import InconsistentShardError, OutputGrid, ShardKey
 from zagg.grids.healpix import HEALPIX_BASE_CELLS, HealpixGrid
 from zagg.grids.rectilinear import OOB_SENTINEL, RectilinearGrid
@@ -53,6 +53,7 @@ def from_config(
             config=config,
             populated_shards=populated_shards,
             chunk_inner=grid_cfg.get("chunk_inner"),
+            sharded=get_sharded(config),
         )
     if grid_type == "rectilinear":
         required = ("crs", "resolution", "bounds")
--- a/src/zagg/grids/__init__.py
+++ b/src/zagg/grids/__init__.py
@@ -68,6 +68,7 @@ def from_config(
             chunk_shape=tuple(grid_cfg.get("chunk_shape", (256, 256))),
             config=config,
             chunk_inner=tuple(chunk_inner) if chunk_inner is not None else None,
+            sharded=get_sharded(config),
         )
     raise ValueError(f"Unknown output.grid.type: {grid_type!r}")
 
--- a/src/zagg/processing/write.py
+++ b/src/zagg/processing/write.py
@@ -279,6 +279,18 @@ def write_shard_to_zarr(
                 zarr_format=3,
                 consolidated=False,
             )
+            if trailing:
+                # Mirror write_dataframe_to_zarr's single-trailing-chunk invariant
+                # (issue #29): a vector field's trailing payload dim must be one whole
+                # (inner) chunk, or set_block_selection at trailing block 0 would
+                # silently write only part of it.
+                target_trailing_chunks = array.chunks[len(slab_shape) :]
+                if target_trailing_chunks != trailing:
+                    raise ValueError(
+                        f"vector field {name!r}: trailing chunk "
+                        f"{target_trailing_chunks} must equal trailing shape "
+                        f"{trailing} (the payload dim must be one whole chunk)"
+                    )
             array.set_block_selection(block_idx, slab)
     return store
 
--- a/src/zagg/processing/write.py
+++ b/src/zagg/processing/write.py
@@ -211,12 +211,13 @@ def write_shard_to_zarr(
     them from the shard index (sub-shard sparsity preserved inside the object).
 
     ``chunk_results`` is the worker's ``[(block_index, carrier, ragged), ...]`` —
-    one entry per inner chunk (from ``grid.iter_chunks``). Dense per-cell columns
-    (data vars + coords) are placed by their ``cell_ids`` into the shard slab
-    (cell id ``c`` lands at shard-relative offset ``c - shard_lo``). The
-    ``resolution: chunk`` companions and ragged (CSR) fields are NOT sharded —
-    they are delegated to :func:`write_dataframe_to_zarr` / :func:`write_ragged_to_zarr`
-    per inner chunk, exactly as on the regular path.
+    one entry per inner chunk (from ``grid.iter_chunks``). Each dense per-cell
+    column (data vars + coords) is placed into the shard slab by the chunk's own
+    region (``grid.shard_local_region(block_index, shard_key)``), reshaped to the
+    inner chunk's cell shape (``grid.chunk_shape``) — grid-agnostic, so HEALPix
+    (1-D) and rectilinear (2-D) share one path. The ``resolution: chunk``
+    companions and ragged (CSR) fields are NOT sharded — they are written per inner
+    chunk via the existing seams, exactly as on the regular path.
 
     Parameters
     ----------
@@ -225,58 +226,40 @@ def write_shard_to_zarr(
     store : Store
         Zarr store with the (sharded) template already written.
     grid : OutputGrid
-        Must be ``sharded`` with a ``cells_per_shard`` and ``block_index``.
+        Must be ``sharded`` with ``shard_slab_shape`` / ``shard_local_region``.
     shard_key : int
-        The dispatch shard's key (HEALPix parent morton id).
+        The dispatch shard's key (HEALPix parent morton id / rect packed tile).
 
     Returns
     -------
     Store
     """
     chunk_res_fields = _chunk_resolution_fields(getattr(grid, "config", None))
-    (shard_block,) = tuple(int(i) for i in grid.block_index(shard_key))
-    cells_per_shard = int(grid.cells_per_shard)
-    shard_lo = shard_block * cells_per_shard
+    shard_block = tuple(int(i) for i in grid.block_index(shard_key))
+    slab_shape = tuple(int(s) for s in grid.shard_slab_shape())
+    inner_shape = tuple(int(s) for s in grid.chunk_shape)
 
     # Accumulate each dense per-cell column into one shard-wide slab; companions
     # and ragged are written per inner chunk via the existing seams.
     slabs: dict = {}
-    fills: dict = {}
     for block_index, carrier, ragged in chunk_results:
-        if _carrier_empty(carrier):
-            # Still flush this chunk's companion/ragged below (they may be empty
-            # no-ops); an empty dense carrier contributes nothing to the slab.
-            write_dataframe_to_zarr(carrier, store, grid=grid, chunk_idx=block_index)
-            write_ragged_to_zarr(
-                ragged, store, grid=grid, shard_key=_block_index_key(block_index, grid)
-            )
-            continue
-        cell_ids = None
-        for name, values in _iter_carrier_columns(carrier):
-            if name == "cell_ids":
-                cell_ids = np.asarray(values).astype(np.int64)
-        if cell_ids is None:
-            raise ValueError(
-                "sharded write requires a 'cell_ids' coord column to place cells "
-                f"within the shard, but carrier for shard {shard_key} has none"
-            )
-        offsets = cell_ids - shard_lo
-        for name, values in _iter_carrier_columns(carrier):
-            if name in chunk_res_fields:
-                continue  # companion (resolution: chunk) — handled below, unsharded
-            values = np.asarray(values)
-            trailing = values.shape[1:]
-            if name not in slabs:
-                array = open_array(
-                    store,
-                    path=f"{grid.group_path}/{name}",
-                    zarr_format=3,
-                    consolidated=False,
-                )
-                fill = array.metadata.fill_value
-                fills[name] = fill
-                slabs[name] = np.full((cells_per_shard, *trailing), fill, dtype=values.dtype)
-            slabs[name][offsets] = values
+        if not _carrier_empty(carrier):
+            region = grid.shard_local_region(block_index, shard_key)
+            for name, values in _iter_carrier_columns(carrier):
+                if name in chunk_res_fields:
+                    continue  # companion (resolution: chunk) — handled below, unsharded
+                values = np.asarray(values)
+                trailing = values.shape[1:]
+                if name not in slabs:
+                    array = open_array(
+                        store,
+                        path=f"{grid.group_path}/{name}",
+                        zarr_format=3,
+                        consolidated=False,
+                    )
+                    fill = array.metadata.fill_value
+                    slabs[name] = np.full((*slab_shape, *trailing), fill, dtype=values.dtype)
+                slabs[name][region] = values.reshape((*inner_shape, *trailing))
         # Companions + ragged for this inner chunk are not sharded: write them
         # straight through (companion array is one block per chunk; ragged is CSR).
         if chunk_res_fields:
@@ -287,8 +270,8 @@ def write_shard_to_zarr(
 
     # One block selection per dense array == one shard object write.
     for name, slab in slabs.items():
-        trailing = slab.shape[1:]
-        block_idx = (shard_block, *((0,) * len(trailing)))
+        trailing = slab.shape[len(slab_shape) :]
+        block_idx = (*shard_block, *((0,) * len(trailing)))
         with config.set({"async.concurrency": 128}):
             array = open_array(
                 store,
--- a/src/zagg/processing/write.py
+++ b/src/zagg/processing/write.py
@@ -191,6 +191,141 @@ def write_dataframe_to_zarr(
     return store
 
 
+def write_shard_to_zarr(
+    chunk_results: list,
+    store: Store,
+    *,
+    grid,
+    shard_key: int,
+) -> Store:
+    """Write a whole sharded shard in ONE block selection per dense array (issue #108).
+
+    On a ``ShardingCodec`` array (``grid.sharded``) the K inner chunks of one
+    dispatch shard live inside a single shard object, and ``set_block_selection``
+    is **shard-granular** (one block == one shard). Writing the K inner chunks as K
+    separate block selections would each trigger a read-modify-write on the *same*
+    shard object — racy and slow. The worker already holds all K chunk carriers
+    (it reads the shard's granules once), so this assembles them into one
+    shard-wide slab per dense per-cell array and writes it at the shard block in a
+    single call. Empty inner chunks are left as fill, so the ShardingCodec omits
+    them from the shard index (sub-shard sparsity preserved inside the object).
+
+    ``chunk_results`` is the worker's ``[(block_index, carrier, ragged), ...]`` —
+    one entry per inner chunk (from ``grid.iter_chunks``). Dense per-cell columns
+    (data vars + coords) are placed by their ``cell_ids`` into the shard slab
+    (cell id ``c`` lands at shard-relative offset ``c - shard_lo``). The
+    ``resolution: chunk`` companions and ragged (CSR) fields are NOT sharded —
+    they are delegated to :func:`write_dataframe_to_zarr` / :func:`write_ragged_to_zarr`
+    per inner chunk, exactly as on the regular path.
+
+    Parameters
+    ----------
+    chunk_results : list of (block_index, carrier, ragged)
+        The worker's per-inner-chunk results for one dispatch shard.
+    store : Store
+        Zarr store with the (sharded) template already written.
+    grid : OutputGrid
+        Must be ``sharded`` with a ``cells_per_shard`` and ``block_index``.
+    shard_key : int
+        The dispatch shard's key (HEALPix parent morton id).
+
+    Returns
+    -------
+    Store
+    """
+    chunk_res_fields = _chunk_resolution_fields(getattr(grid, "config", None))
+    (shard_block,) = tuple(int(i) for i in grid.block_index(shard_key))
+    cells_per_shard = int(grid.cells_per_shard)
+    shard_lo = shard_block * cells_per_shard
+
+    # Accumulate each dense per-cell column into one shard-wide slab; companions
+    # and ragged are written per inner chunk via the existing seams.
+    slabs: dict = {}
+    fills: dict = {}
+    for block_index, carrier, ragged in chunk_results:
+        if _carrier_empty(carrier):
+            # Still flush this chunk's companion/ragged below (they may be empty
+            # no-ops); an empty dense carrier contributes nothing to the slab.
+            write_dataframe_to_zarr(carrier, store, grid=grid, chunk_idx=block_index)
+            write_ragged_to_zarr(
+                ragged, store, grid=grid, shard_key=_block_index_key(block_index, grid)
+            )
+            continue
+        cell_ids = None
+        for name, values in _iter_carrier_columns(carrier):
+            if name == "cell_ids":
+                cell_ids = np.asarray(values).astype(np.int64)
+        if cell_ids is None:
+            raise ValueError(
+                "sharded write requires a 'cell_ids' coord column to place cells "
+                f"within the shard, but carrier for shard {shard_key} has none"
+            )
+        offsets = cell_ids - shard_lo
+        for name, values in _iter_carrier_columns(carrier):
+            if name in chunk_res_fields:
+                continue  # companion (resolution: chunk) — handled below, unsharded
+            values = np.asarray(values)
+            trailing = values.shape[1:]
+            if name not in slabs:
+                array = open_array(
+                    store,
+                    path=f"{grid.group_path}/{name}",
+                    zarr_format=3,
+                    consolidated=False,
+                )
+                fill = array.metadata.fill_value
+                fills[name] = fill
+                slabs[name] = np.full((cells_per_shard, *trailing), fill, dtype=values.dtype)
+            slabs[name][offsets] = values
+        # Companions + ragged for this inner chunk are not sharded: write them
+        # straight through (companion array is one block per chunk; ragged is CSR).
+        if chunk_res_fields:
+            _write_companion_columns(carrier, store, grid, block_index, chunk_res_fields)
+        write_ragged_to_zarr(
+            ragged, store, grid=grid, shard_key=_block_index_key(block_index, grid)
+        )
+
+    # One block selection per dense array == one shard object write.
+    for name, slab in slabs.items():
+        trailing = slab.shape[1:]
+        block_idx = (shard_block, *((0,) * len(trailing)))
+        with config.set({"async.concurrency": 128}):
+            array = open_array(
+                store,
+                path=f"{grid.group_path}/{name}",
+                zarr_format=3,
+                consolidated=False,
+            )
+            array.set_block_selection(block_idx, slab)
+    return store
+
+
+def _write_companion_columns(carrier, store, grid, block_index, chunk_res_fields):
+    """Write a chunk's ``resolution: chunk`` companion columns (unsharded path).
+
+    A ``resolution: chunk`` companion array is shaped at the coarse chunk grid
+    (one block per chunk), never sharded, so it keeps the per-chunk write even on
+    the sharded path. Reuses the same chunk-uniform collapse + block placement as
+    :func:`write_dataframe_to_zarr`'s companion branch.
+    """
+    block_idx = tuple(int(i) for i in block_index)
+    for name, values in _iter_carrier_columns(carrier):
+        if name not in chunk_res_fields:
+            continue
+        chunk_value = np.asarray(_chunk_uniform_value(name, values))
+        trailing = chunk_value.shape
+        block = chunk_value.reshape((1,) * len(block_idx) + trailing)
+        target = block_idx + (0,) * len(trailing)
+        with config.set({"async.concurrency": 128}):
+            array = open_array(
+                store,
+                path=f"{grid.group_path}/{name}",
+                zarr_format=3,
+                consolidated=False,
+            )
+            array.set_block_selection(target, block)
+
+
 def write_ragged_to_zarr(
     ragged: dict,
     store: Store,
--- a/tests/test_grids.py
+++ b/tests/test_grids.py
@@ -928,3 +928,43 @@ class TestSharded:
         dense_bytes = len(store._store_dict[f"8/h_mean/c/{shard_block}"])
 
         assert sparse_bytes < dense_bytes
+
+    def test_vector_field_shards_and_roundtrips(self):
+        # A kind: vector field's trailing payload dim stays whole inside the shard:
+        # outer shard (cells_per_shard, T), inner chunk (cells_per_chunk, T). A whole-
+        # shard slab round-trips, and one shard object holds the K inner chunks.
+        from mortie import geo2mort
+        from zarr import open_array
+
+        g = HealpixGrid(
+            parent_order=4,
+            child_order=8,
+            layout="fullsphere",
+            config=_vector_config(bins=4, dtype="float32"),
+            chunk_inner=6,
+            sharded=True,
+        )
+        store = MemoryStore()
+        g.emit_template(store)
+        arr = open_array(store, path="8/hist", zarr_format=3, consolidated=False)
+        assert arr.chunks == (g.cells_per_chunk, 4)
+        assert arr.shards == (g.cells_per_shard, 4)
+
+        parent = int(geo2mort(-78.5, -132.0, order=4)[0])
+        (shard_block,) = g.block_index(parent)
+        slab = np.zeros((g.cells_per_shard, 4), dtype="float32")
+        slab[: g.cells_per_chunk] = np.arange(g.cells_per_chunk * 4).reshape(g.cells_per_chunk, 4)
+        arr.set_block_selection((shard_block, 0), slab)
+
+        shard_keys = [k for k in store._store_dict if k.startswith("8/hist/c/")]
+        assert shard_keys == [f"8/hist/c/{shard_block}/0"]
+        lo = shard_block * g.cells_per_shard
+        np.testing.assert_array_equal(arr[lo : lo + g.cells_per_shard], slab)
+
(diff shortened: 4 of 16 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the **phase-1** diff (sharded `emit_template` spec + round-trip test). Worker whole-shard write (phase 2), config wiring (phase 3), rect (phase 4), docs (phase 5) are out of scope and not reviewed.

**Verdict: phase-1 core is correct and matches the locked design.** I ran the suite locally (`pytest tests/test_grids.py` → 64 passed; ruff check + format clean; base.py 312 lines / healpix.py 506 lines, both well under the §4 ~1000 cap). The sharding codec config is right against issue #108's locked decisions: outer chunk grid = `cells_per_shard` (the dispatch shard), inner `chunk_shape` = `cells_per_chunk` (the 64×64 read chunk), inner codecs are the array's existing **bytes-only** codecs (no zstd injected), `index_codecs = [bytes(little), crc32c]` with `index_location: "end"` matching zarr's `create_array` default, and the spec is built **by hand** via `NamedConfig` (not `ArraySpec.from_array`, which drops the sharding codec in pydantic-zarr 0.10.0) — both load-bearing caveats from the issue are honored. K==1 is **validate-and-error** with a clear message, exactly the locked call. Flag-off is dict-identical to the regular template. I independently verified the vector-field path: a `kind: vector` field with `trailing_shape [4]` correctly emits shape `(786432, 4)`, outer shard `(256, 4)`, inner chunk `(16, 4)` — the trailing dim stays whole in both, so `_shard`'s `inner[1:]` logic is sound.

Three findings, none blocking, in rough priority order:

**1. (medium — scope creep) Unrelated `h_li` → `h_mean` edits in two test helpers.** `tests/test_grids.py` `_vector_config` (line ~330) and `_chunk_resolution_config` (lines ~457/460/461) flip `source: h_li` → `source: h_mean` (and the `np.median(h_li)` → `np.median(h_mean)` anchor). These helpers feed `TestChunkInnerMultiChunk`, **not** the new `TestSharded` (which builds from `default_config("atl06")` directly). I reverted them locally and the full suite still passes 64/64 — so the change is **unnecessary** for phase 1 and undocumented. Note `h_mean` is an *output field name* whereas the `source` slot elsewhere names a *raw input column* (`h_li`/`s_li`/`b`); swapping in an output-field name as a source reads oddly. Recommend dropping these edits from the phase-1 commit (or, if intentional, calling out why in the PR body). Per §3/§4 each commit should be coherent and scoped.

**2. (low — test gap) No sharded *vector*-field round-trip test.** `TestSharded`'s round-trip / partial-decode / sparse-omission tests all exercise the scalar `h_mean` only. The trailing-dim shard logic (`shard = (cells_per_shard, *inner[1:])`) is brand-new behavior; it's verified spec-shape-wise (I checked manually) but never round-tripped through a write/read. A single `set_block_selection` round-trip on a 2D vector member would close this.

**3. (low) `sharded` is absent from `signature()` / `spatial_signature()` with no asserting test.** This is consistent with the existing treatment of `chunk_inner` (storage form, not spatial identity — see `test_healpix_chunk_inner_not_in_signature`), so it's defensible and likely correct. But `chunk_inner` has an explicit test pinning its exclusion and `sharded` does not; a parallel one-liner would lock the intent.

Phase-1 acceptance criteria from issue #108 (one shard object per dispatch shard, empties omitted, byte-exact read-back, 64×64 partial-decode, bytes-only preserved, flag-off identical, K>1 validation) are all met and tested. Nothing here blocks advancing to phase 2.

---

src/zagg/grids/__init__.py
```diff
@@ -53,6 +53,7 @@ def from_config(
             config=config,
             populated_shards=populated_shards,
             chunk_inner=grid_cfg.get("chunk_inner"),
+            sharded=get_sharded(config),
```
espg, line 56: 🤖 *from Claude (review)*

Non-blocking (phase-4-adjacent): `sharded=get_sharded(config)` is threaded only into the HEALPix branch. The `rectilinear` branch (lines 57-70) never reads `get_sharded`, and `RectilinearGrid` has no `sharded` param — so a config with `type: rectilinear` + `sharded: true` is **silently dropped** (verified: `from_config` returns a `RectilinearGrid` with no `.sharded` attribute, no warning, no error). That's the opposite of the locked design's intent for the HEALPix knob, which is *validate-and-error* on a meaningless setting (K==1). Rect sharding is phase 4 so I'm not flagging its absence — only that the knob should not no-op silently in the interim. Cheapest guard: in the rect branch, `if get_sharded(config): raise ValueError("sharded output is not yet supported for rectilinear grids (issue #108 lands HEALPix first)")`. Leaving for @espg to decide whether to add now or defer to phase 4.

---

src/zagg/config.py
```diff
@@ -1170,6 +1170,18 @@ def get_layout(config: PipelineConfig) -> str:
...
 
+def get_sharded(config: PipelineConfig) -> bool:
+    """Return whether the output grid uses ShardingCodec storage (issue #108).
+
+    The ``sharded`` knob lives on the grid/chunk block next to ``chunk_inner``
+    (mirroring its accessor), default ``False``. When ``True`` the grid bundles a
+    dispatch shard's K inner chunks into one zarr shard object instead of K
+    independent regular chunk objects; it is only valid when ``chunk_inner`` gives
+    K>1 (the grid raises otherwise, validated before deployment).
+    """
```
espg, line 1181: 🤖 *from Claude (review)*

Accessor looks correct and on-convention. `get_sharded` reads `output.grid.sharded` (right block — next to `chunk_inner`, not `output.*`), defaults `False`, and mirrors `get_layout`'s `config.output.get("grid", {}).get(..., default)` shape exactly. The `bool(...)` coercion is a sensible belt-and-suspenders over a YAML-sourced value. No `noqa`/`type: ignore`, terse, matches surrounding accessors. No change requested here — confirming the design checks out.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the **phase-3** diff of #108 (commit ba4be30) — scope limited to the config flag + accessor + validation wiring (config.py, grids/__init__.py, test_config.py, test_grids.py). Rect (phase 4) and docs (phase 5) treated as out of scope.

**Verdict: phase 3 is correct and on-design. One non-blocking observation, one design confirmation — both inline.**

Against the locked design (issue #108):
- **Placement** — `sharded` lives on the grid/chunk block (`output.grid.sharded`), next to `chunk_inner`, not under `output.*`. Correct.
- **Default off** — `get_sharded` returns `False` when the key is absent; `test_get_sharded_default_false` and `test_sharded_default_off` both assert it. Confirmed.
- **Accessor convention** — mirrors `get_layout`/`get_child_order` (`config.output.get("grid", {}).get(...)`), terse, no `noqa`/`type: ignore`. Confirmed.
- **K==1 validate-and-error surfaces through `from_config`** — `test_sharded_k1_validated_pre_deployment` drives `from_config` with `sharded:true` and no `chunk_inner` and asserts `ValueError` matching `K>1`; the message lives in the HealpixGrid ctor (phase 1) and does contain `(K>1)`. Verified by running the suite at ba4be30 in a scratch worktree: `TestFromConfig` + `TestOutputHelpers` = 15 passed; `ruff check` on all four files clean.

Test coverage (phase-3 scope):
- flag-off byte-identity present — `test_sharded_off_byte_identical_template` compares `_spec().model_dump()` with `chunk_inner` set, absent vs explicit-false. Good.
- threads-through present — `test_sharded_flag_threads_through` (asserts `g.sharded` and `chunks_per_shard > 1`). Good.
- K==1 error path present (above). Good.

**Findings:**
1. (inline, grids/__init__.py:56, non-blocking) A `rectilinear` + `sharded:true` config is **silently dropped** — the flag is read only on the HEALPix branch, so rect no-ops with no warning/error rather than validate-and-erroring like the HEALPix K==1 path does. Phase-4-adjacent; flagged for @espg to add a guard now or defer.
2. (inline, config.py:1181, confirmation only) accessor design is right; no change requested.

No edits, pushes, or thread resolutions made — review only.

---

espg (commented):
🤖 *from Claude (review)*

**Phase 2 adversarial review — sharded whole-shard write (commit `d8a5e8a`).**

Scope reviewed: `write_shard_to_zarr` + `_write_companion_columns` (`src/zagg/processing/write.py`), the sharded routing in `src/zagg/runner.py` and `deployment/aws/lambda_handler.py`, the export, and the three test files. Phases 3–5 (config-flag wiring, rect, docs) are out of scope and their absence is not flagged.

**Verdict: correct and well-tested; ship-able. Findings below are minor / non-blocking.** I ran the new `TestWriteShardToZarr` class plus the full `test_lambda_handler.py` against this commit (in an isolated worktree) — **18 passed**; `ruff check` clean on all touched files; `write.py` is 597 lines (under the §4 ~1000 cap).

**Correctness — the slab placement is sound.** The whole-shard slab places each cell at `cell_ids - shard_lo`, with `shard_lo = block_index(shard_key)[0] * cells_per_shard`. This is correct *because* HEALPix nested ids are hierarchically contiguous: a parent at order `p` with nested id `P` owns children `[P·4^(c−p), (P+1)·4^(c−p))` at order `c`, and `cells_per_shard == n_children == 4^(c−p)`, so `shard_lo == P·4^(c−p)` is exactly the shard's first global cell index. Each inner chunk's carrier (built by `_build_output` over `iter_chunks`' `chunk_children`) carries a full `cells_per_chunk` rows in canonical nested order, so the per-chunk `cell_ids` are contiguous, disjoint, and tile `[shard_lo, shard_lo+cells_per_shard)` exactly — byte-identical to the K separate `set_block_selection` writes on the regular path. The locked design's "one whole-shard slab, not a per-inner-chunk RMW loop" is implemented exactly (one `set_block_selection((shard_block, 0…), slab)` per dense array). Ragged keying stays per-inner-chunk via `_block_index_key` (not `shard_key`), and `resolution: chunk` companions are routed to the coarse chunk grid via `_write_companion_columns` — both unsharded, matching the regular path.

**Regular path byte-identity:** the `getattr(grid, "sharded", False)` gate in both `runner._process_and_write` and `lambda_handler._handle_process` leaves the non-sharded loop untouched (the sharded branch returns/dispatches before it). `grid.sharded` defaults `False` and the lambda test stubs now pin `grid_stub.sharded = False` (a `MagicMock` attr is truthy by default, so this pin is necessary and correct). Confirmed byte-identical.

**Test quality:** `test_sharded_matches_regular_byte_for_byte` builds a sharded and a regular grid with identical `chunk_inner=7` (K=4) geometry, feeds the same df (3 photons in 2 cells spanning the first and last inner chunk), and asserts array-equality across all arrays — a real misplacement (wrong `shard_lo`/offset) would move cells and fail it. `test_readback_places_cells_at_correct_positions` nails it down further (`count[first]==2`, `count[last]==1`, interior cell stays NaN). `test_one_shard_object_per_dispatch_shard` checks exactly one shard object key, not K. Good coverage of the phase-2 contract.

**§1 / lambda_handler.py:** modifying `deployment/aws/lambda_handler.py` here is **OK, not a violation.** §1's `deployment/aws/` prohibition names the CloudFormation `template.yaml` and the build/standup/mirror scripts — i.e. infra/deploy. `lambda_handler.py` is the application/runtime handler and has been routinely edited by prior issue work (#82 phase 7, #99, #100, #107 in its git log); the change here is symmetric with the runner. Within scope.

Specific findings posted inline.

---

src/zagg/processing/write.py
```diff
@@ -191,6 +191,141 @@ def write_dataframe_to_zarr(
...
+            if name not in slabs:
+                array = open_array(
+                    store,
+                    path=f"{grid.group_path}/{name}",
+                    zarr_format=3,
+                    consolidated=False,
+                )
+                fill = array.metadata.fill_value
+                fills[name] = fill
+                slabs[name] = np.full((cells_per_shard, *trailing), fill, dtype=values.dtype)
```

Dead `fills` dict (§4 terse-code, non-blocking). `fills` is populated (`fills[name] = fill`) but never read anywhere — `slabs[name]` already captures the `np.full(..., fill, ...)` slab. Drop the `fills: dict = {}` declaration and the `fills[name] = fill` line; keep the local `fill` used on the next line. Pure cleanup, no behavior change.

---

src/zagg/processing/write.py
```diff
@@ -191,6 +191,141 @@ def write_dataframe_to_zarr(
...
+        trailing = slab.shape[1:]
+        block_idx = (shard_block, *((0,) * len(trailing)))
+        with config.set({"async.concurrency": 128}):
+            array = open_array(
+                store,
+                path=f"{grid.group_path}/{name}",
+                zarr_format=3,
+                consolidated=False,
+            )
+            array.set_block_selection(block_idx, slab)
```
espg, line 294: 🤖 *from Claude (review)*

Dropped vector single-trailing-chunk guard (robustness gap, non-blocking — no vector field is currently sharded in tests). The regular path in `write_dataframe_to_zarr` validates, for vector fields, that the target array's trailing chunk spans the whole payload (`target_trailing_chunks != trailing` → `ValueError`, write.py ~182-188) before `set_block_selection` at trailing block 0 — otherwise it would silently write only part of the payload (issue #29 invariant). The slab path here writes `block_idx = (shard_block, *((0,)*len(trailing)))` with no such guard. For the sharded template emitted in phase 1 the trailing dim is chunked whole so this holds today, but if a future grid emits a partially-chunked trailing payload the slab write would silently truncate instead of erroring. Consider mirroring the same assertion here, or asserting `array.chunks[1:] == trailing`. Phase-2 tests don't exercise a vector field, so this is untested either way.

---

src/zagg/processing/write.py
```diff
@@ -191,6 +191,141 @@ def write_dataframe_to_zarr(
...
+    fills: dict = {}
+    for block_index, carrier, ragged in chunk_results:
+        if _carrier_empty(carrier):
+            # Still flush this chunk's companion/ragged below (they may be empty
+            # no-ops); an empty dense carrier contributes nothing to the slab.
+            write_dataframe_to_zarr(carrier, store, grid=grid, chunk_idx=block_index)
+            write_ragged_to_zarr(
+                ragged, store, grid=grid, shard_key=_block_index_key(block_index, grid)
+            )
+            continue
```

Minor/observation (non-blocking): this `_carrier_empty(carrier)` branch is effectively dead on the HEALPix K>1 path. `process_shard` iterates *all* K chunks from `iter_chunks`, and `_build_output` always emits a full `cells_per_chunk`-row carrier (empty cells carried as NaN-fill), so `num_rows == cells_per_chunk > 0` and `_carrier_empty` is never `True` for a populated shard. The branch is harmless (and a reasonable defensive guard / future-proofing for a carrier type that could yield 0 rows), so no change required — just confirming it isn't reachable in the current flow, so it isn't covered by the new tests. Worth a one-line note if you want to keep it explicitly defensive.

---

tests/test_grids.py
```diff
@@ -296,6 +296,35 @@ def test_unknown_grid_raises(self, cfg):
...
+        # Flag off must reproduce the regular template byte-for-byte even when
+        # chunk_inner is set (sharding is purely opt-in storage form).
+        cfg.output["grid"]["chunk_inner"] = 8
+        g_off = from_config(cfg, parent_order=6)
+        cfg.output["grid"]["sharded"] = False
+        g_explicit_off = from_config(cfg, parent_order=6)
+        assert g_off._spec().model_dump() == g_explicit_off._spec().model_dump()
+
 
 class TestBackcompatWrapper:
```
espg, line 329: 🤖 *from Claude (review)*

Confirmed-correct, neutral note. These `h_mean`→`h_li` edits (and the ones in `_chunk_resolution_config`) revert an unrelated `h_li`→`h_mean` change that phase 1 (commit `0c5b25e`) introduced into `test_grids.py` — they restore the exact pre-phase-1 state (verified against `0c5b25e~1`). Good that it's being cleaned up. The only nit is process: a scope-creep edit was landed in phase 1 and reverted in phase 2 rather than kept out of phase 1; not worth re-rolling history, just flagging so the PR-body checklist notes phase 2 includes this revert.

---

espg (commented):
🤖 *from Claude (review)*

Inline findings for the phase-2 review (all minor / non-blocking) — see the summary review for the overall verdict.

---

src/zagg/grids/__init__.py
espg: 🤖 *from Claude*

Addressed in phase 4 (`1766108`): the `rectilinear` branch now reads `sharded=get_sharded(config)` and `RectilinearGrid` accepts `sharded` with the same K>1 validate-and-error guard as HEALPix, so a `type: rectilinear` + `sharded: true` config no longer silently no-ops — it either builds a sharded rect grid (K>1) or raises the "K>1" error (K==1). No interim silent-drop remains.

---

src/zagg/processing/write.py
espg: 🤖 *from Claude*

Done — the phase-4 refactor of `write_shard_to_zarr` (`1766108`) removed the dead `fills` dict entirely when it switched to grid-agnostic block-based placement (the local `fill` feeds `np.full` directly).

---

src/zagg/processing/write.py
espg: 🤖 *from Claude*

Folded in `90fad01`: `write_shard_to_zarr`'s final slab-write loop now mirrors `write_dataframe_to_zarr`'s single-trailing-chunk guard — for a vector field it asserts `array.chunks[len(slab_shape):] == trailing` before the shard-block `set_block_selection`, raising the same "the payload dim must be one whole chunk" error instead of silently truncating. Still no sharded-vector config in the suite, so the guard is structural; left the vector round-trip as an open question in the PR body for your call.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the **phase-4 diff** (commit `1766108`): rect ShardingCodec + the `write_shard_to_zarr` refactor to grid-agnostic block-based placement. Scope was the rect engine and the unified whole-shard write path; phase-5 docs out of scope.

**Verdict: correct. No blocking findings.** The critical concern — that the refactor keeps the HEALPix path byte-identical — holds, verified two ways:

1. **HEALPix byte-identity (empirical).** I reconstructed both placements (OLD `cell_ids - shard_lo` vs NEW contiguous `shard_local_region` slice + `reshape(chunk_shape)`) across `parent_order/chunk_inner/child_order` = (0,1,3),(1,2,4),(2,3,4),(0,2,4) and multiple base cells/shards. Every slab was element-for-element identical with no unfilled cells. Confirmed the per-chunk `iter_chunks` `children` are **ascending-contiguous in nested ids** (`cid[0] == shard_lo + local·cells_per_chunk`, `cid == arange(cid[0], +cells_per_chunk)`), so the slice placement lands exactly where the cell-id scatter did — order is not scrambled.
2. **HEALPix byte-identity (regression test).** `TestWriteShardToZarr` (incl. `test_sharded_matches_regular_byte_for_byte`) passes on this commit — `3 passed in 177.87s`.

**Rect arithmetic — correct.** `iter_chunks` yields `block=(r0//inner_h, c0//inner_w)` with `r0 = rb·chunk_h + ir·inner_h`; `shard_local_region` recovers `lir = gir − rb·(chunk_h//inner_h) = ir`, slice `[ir·inner_h:(ir+1)·inner_h]` (sym. for cols). Children are row-major `rows·width+cols`, and the writer reshapes to `grid.chunk_shape == (inner_h, inner_w)` — row-major match. Final `block_idx = (*shard_block, *0-per-trailing)` with `shard_block == (rb,cb)` selects the right shard over the `(chunk_h,chunk_w)` shard grid; `+0` per vector-trailing dim is correct. `test_whole_shard_write_roundtrip` (h_mean = own global id, read-back == `arange·width+arange`, exactly one `c/0/0` object) genuinely proves placement + one-shard-object.

**Rect `_spec` sharding — correct.** `_shard` is applied only to per-cell coord arrays and `resolution: cell` data-vars; the `resolution: chunk` companions `continue` before `_shard` and the 1-D x/y coords are never wrapped. `shard=(chunk_h,chunk_w,*inner[2:])` handles scalar (trailing `()`) and vector (trailing kept whole). `test_template_emits_sharding_codec` asserts chunks `(4,4)`/shards `(8,8)`/sharding codec on data-vars and `shards is None` on x/y. Flag-off (`_shard` returns `arr` unchanged) gives `_spec().model_dump()` equality (`test_flag_off_byte_identical`). K>1 validation (`n_inner<=1` → clear `K>1` error) mirrors HEALPix and is tested both at the grid and `from_config` layers.

**Dead code & hygiene.** The phase-2 cell-ids path is fully removed — `shard_lo`, `fills`, the `cell_ids` extraction loop, and the `cell_ids is None` raise are gone (only docstrings mention cell_ids now). No `# noqa` / `# type: ignore` added. ruff check + ruff format clean on all five files. Module sizes: write.py 580, rectilinear.py 610, healpix.py 531 — all under the ~1000 line bar. New code matches surrounding patterns (reuses `sharded_array_spec`, `get_sharded`, `_write_companion_columns`). Full `tests/test_rectilinear.py` = 45 passed.

One **non-blocking robustness observation** on the empty-carrier path — see inline comment. Net: ship-worthy for phase 4; the inline note is a defensive nit, not a correctness defect on any reachable worker carrier.

---

espg (commented):
🤖 *from Claude (review)*

**Non-blocking robustness nit — empty-carrier companion path** (`src/zagg/processing/write.py:245-269`, the refactored `write_shard_to_zarr` loop).

Behavior change vs the pre-refactor code: previously an empty carrier took a dedicated branch that called `write_dataframe_to_zarr(carrier, ...)` — which **returns early at its `_carrier_empty` guard (`write.py:120-121`), a no-op** — so companions were *not* written for an empty carrier. The new loop drops that branch and, for an empty carrier, still falls through to `_write_companion_columns(...)` + `write_ragged_to_zarr(...)` (lines 263-269).

If a carrier is ever empty *with columns present* (zero rows) and `chunk_res_fields` is non-empty, `_write_companion_columns` → `_chunk_uniform_value` hits `arr[0]` on a zero-length array (`write.py:525`) → `IndexError`.

In practice this looks **unreachable**: `_build_output` (`write.py:44-64`) always materializes every column at `len(children) == prod(chunk_shape)` rows (empty cells carry NaN/anchor sentinels, not dropped), so a real worker chunk carrier is never zero-row; `_carrier_empty` only trips for a genuinely empty DataFrame, which then yields no columns and the companion loop is a safe no-op. So I'm flagging this as a defensive nit, not a live bug — no change required for phase 4. If you want belt-and-suspenders, an early `continue` on `_carrier_empty(carrier)` (matching the old guard's intent) documents that companions/ragged for a truly-empty chunk are intentionally skipped. Standing for @espg's call; not a phase blocker.

---

espg:
🤖 *from Claude*

All five phases are complete and pushed; the PR is ready for your review (kept as a draft). Summary of the run:

**Phases landed (one commit each):**
- `0c5b25e` — phase 1: HEALPix sharded `emit_template` (`sharded_array_spec` helper, `sharded` flag + K>1 validation, sharding codec on dense per-cell arrays) + round-trip regression.
- `d8a5e8a` — phase 2: `write_shard_to_zarr` whole-shard write; runner + lambda handler route sharded grids to it; regular path byte-identical. End-to-end sharded-vs-regular byte-identity test.
- `ba4be30` — phase 3: `get_sharded` accessor (on the grid block, next to `chunk_inner`, default off) threaded through `from_config`; K==1 surfaces as a pre-deployment error.
- `1766108` — phase 4: rectilinear engine (sharded flag + K>1 validation + `shard_slab_shape`/`shard_local_region` + sharding codec); `write_shard_to_zarr` refactored to grid-agnostic block-based placement (HEALPix path stays byte-identical, verified).
- `fbf88cb` — phase 5: `docs/sharding.md` + mkdocs nav.

**Self-review folds:**
- `90fad01` — mirrored `write_dataframe_to_zarr`'s single-trailing-chunk guard into the sharded slab write (review finding).
- `a685d4b` — added a sharded **vector**-field round-trip test (closes the vector test-gap finding) and a `sharded`-not-in-`signature()` assertion.
- The phase-1→phase-2 `h_li`/`h_mean` test-helper scope creep was reverted (noted in the PR body).

The rect-silent-drop, dead-`fills`, and vector-guard review findings were all resolved (replies on the threads). The remaining review notes are non-blocking and out of scope (an unreachable defensive empty-carrier companion branch; whether `sharded` should join the fingerprint) — left standing for your call, with the vector-only-scope question under "Questions for review" in the PR body.

**Local gates:** `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on touched files; `codespell` clean on the new doc. Full suites green locally: `test_grids.py` (66), `test_rectilinear.py` (45), `test_processing.py` (165, incl. the slow sharded end-to-end), plus `test_config.py`/`test_runner.py`/`test_runner_concurrency.py`/`test_lambda_handler.py`. CI `ruff` + `build` are green; the `test (3.12/3.13)` jobs run the same suite and were still finishing at hand-off (the sharded `process_shard` tests are slow on CI). `tests/test_shardmap.py` is uncollectable locally (missing optional `stac_geoparquet` from the `catalog` extra) — pre-existing/unrelated, not touched.

**Follow-up for you to open (side-effecting, so not done here):** an issue/PR for the #79/#81 read helpers to consume sharded stores (shard-index/byte-range reads instead of `list_prefix`). Also noting the **vector round-trip scope** and **`sharded` in `signature()`** questions in the PR body for a decision.

Applying `waiting` — every phase is complete; the ball is in your court for review/merge.

---

PR #115, opened 2026-06-27
Lambda build/deploy CI/CD (#25)

### LINKED ISSUE #25: Lambda zips never attached to GitHub releases (GITHUB_TOKEN doesn't cascade-trigger lambda-build)
## Problem

GitHub releases never get the Lambda layer/function zips attached — only the auto-generated source tarballs. e.g. the `0.2.2` release has **0 assets**, and `gh run list --workflow "Lambda Build" --event release` is **empty** (the workflow has never run on a release event).

## Root cause

`publish.yml`'s `github-release` job creates the release with `gh release create` using the default **`GITHUB_TOKEN`** (around line 105). GitHub deliberately **does not let events triggered by `GITHUB_TOKEN` trigger other workflows** (anti-recursion guard). So `lambda-build.yml`'s `release: { types: [published] }` trigger never fires → its `publish-release-assets` job (which is gated on `if: github.event_name == 'release'`) never runs → no release ever gets zips.

The build + attach machinery already exists and is correct — it just never triggers. (Note: `workflow_dispatch` re-runs of `lambda-build` won't help either, because `publish-release-assets` is gated specifically on the `release` event.)

## Fix options

1. **PAT on the release-create step** — swap `GITHUB_TOKEN` → a fine-grained PAT (or GitHub App token) with `contents:write` on the `gh release create` step in `publish.yml`. Then the `published` event cascades and `lambda-build` runs on the release, attaching the zips. Minimal change; cost is a managed secret.
2. **Self-contained publish** — fold the lambda build+attach into `publish.yml` itself (e.g. make `lambda-build`'s build jobs a reusable `workflow_call` workflow and call it from `publish.yml` after creating the release, then `gh release upload`). No PAT, no cross-workflow trigger; cost is a small workflow refactor.

## Priority: low

Not urgent — distribution is currently handled by the **source.coop mirror** (`s3://us-west-2.opendata.source.coop/englacial/zagg/lambda/<minor>/`), which serves both the in-region CloudFormation artifact bucket and HTTPS download. Attaching zips to GitHub releases is a provenance/discoverability nice-to-have on top of that.

## Refs
- `.github/workflows/publish.yml` — `github-release` job (`GITHUB_TOKEN`)
- `.github/workflows/lambda-build.yml` — `release: published` trigger + `publish-release-assets` job

espg:
Closes #25. Refs #110.

Automates Lambda build/deploy across the three tiers from #25, on top of the #112 benchmark harness. Every new job **no-ops/skips when its vars/secrets are absent**, so merging this can't break releases or the existing benchmark before the AWS-side setup exists (order: #112 ✅ → benchmark setup → this PR → AWS update for this PR).

## Phases — all complete
- [x] **Phase 1 — Tier 3 (closes #25):** dual-arch build extracted into a reusable `workflow_call`; `publish.yml` is self-contained — build → create release **+ attach the 4 zips** with `GITHUB_TOKEN` (no cascade, no PAT) → `distribute` to `sliderule-public-cors` (minor-keyed + `versions.json`) → `deploy-prod` updates production in place (publish-layer-version + update-function-config + update-function-code, `production` Environment). New `distribute_zips.sh` + tests.
- [x] **Phase 2 — Tier 2:** `lambda-benchmark.yml` gains a `detect` job (lambda-affecting? configured? non-fork?) + a `deploy-test` job (build arm64 → `update-function-code` on `process-shard-test`), routing the benchmark at the resolved function. Workflow-level `concurrency` serializes runs (shared test fn). Shared `deploy_lambda.sh` (also used by deploy-prod). merge tracks test/staging.
- [x] **Phase 3 — Tier 1:** a `> ⚠️` banner in the PR comment whenever the benchmark ran against the **stable** worker but the change touches deployed code (same-repo pre-deploy + fork PRs), so a plausible-but-wrong number never reads as real.
- [x] **Phase 4 — distribution:** `stand_up.sh` prefers the listable `sliderule-public-cors` bucket (with `LAMBDA_VERSION=latest` via `versions.json`) and **falls back to source.coop** for older minors (no stranded standups); `FUNCTION_NAME` knob to stand up the `process-shard-test` stack. The setup guide gains the full "AWS update" section (test stack, deploy/release roles, CORS bucket policy + CORS config, the 7 new vars, `production` Environment + tag protection, verification).
- [x] **Phase 5 — CloudFormation (SlideRule CFN-first):** `deployment/aws/benchmark_cicd.yaml` provisions the 3 OIDC roles (invoke/test-deploy/release) + the `sliderule-public-cors` bucket (public read/list + CORS) in one rollback-able stack, referencing the existing OIDC provider by account id, with role-ARN Outputs. Guide §2/§9/§10 rewritten CFN-first (raw CLI kept as collapsed fallback). cfn-lint clean.

### Also (post-review)
- **Earthdata auth → bearer token.** The benchmark now authenticates with a single `EARTHDATA_TOKEN` secret (an EDL user token) instead of username+password — tested locally end-to-end (token authenticates **and** the NSIDC S3-credentials endpoint accepts it). No zagg code change (earthaccess reads `EARTHDATA_TOKEN`); composite action + workflows + docs updated.

## How it was tested
- `pytest tests/test_distribute_zips.py tests/test_deploy_lambda.py tests/test_benchmark.py` — 28 pass (real scripts driven by a stub `aws`, no network; the worker-note banner; versions.json seed/merge/numeric-sort; the deploy publish→config→wait→code ordering).
- **actionlint** clean on all workflows; `bash -n` on the scripts; `mkdocs build --strict` clean.
- **Not exercised live:** the OIDC deploy/distribute/prod jobs + the real CORS bucket (need the AWS-update setup). Guarded off until the vars exist.

## Scope / authorization
Touches `.github/workflows/**`, `.github/scripts/`, `.github/actions/`, `deployment/aws/stand_up.sh`, and docs — walled off by §1 generally, but issue #25 names this work explicitly. Nothing runs live AWS from an agent.

## Key decisions (confirmed in-thread)
Self-contained release (no PAT); serialized shared `process-shard-test` via `update-function-code`; `versions.json` index; merge → test/staging worker; inline arm64 build; fail-and-leave-attached on a post-release deploy failure; prod in-place via publish-layer-version (handles a deps/layer change, layer staged in S3 for the >50 MB cap).

_All adversarial-review findings (Phases 1–4) folded; threads replied._

### DIFF
--- a/.github/workflows/lambda-benchmark.yml
+++ b/.github/workflows/lambda-benchmark.yml
@@ -176,8 +176,7 @@ jobs:
           aws-region: ${{ vars.BENCHMARK_AWS_REGION }}
           function-name: ${{ needs.detect.outputs.function_name }}
           store-prefix: ${{ vars.BENCHMARK_STORE_PREFIX }}
-          edl-username: ${{ secrets.EARTHDATA_USERNAME }}
-          edl-password: ${{ secrets.EARTHDATA_PASSWORD }}
+          edl-token: ${{ secrets.EARTHDATA_TOKEN }}
           event: pr
           commit: ${{ github.event.pull_request.head.sha }}
           ref-label: ${{ github.head_ref }}
@@ -225,8 +224,7 @@ jobs:
           aws-region: ${{ vars.BENCHMARK_AWS_REGION }}
           function-name: ${{ needs.detect.outputs.function_name }}
           store-prefix: ${{ vars.BENCHMARK_STORE_PREFIX }}
-          edl-username: ${{ secrets.EARTHDATA_USERNAME }}
-          edl-password: ${{ secrets.EARTHDATA_PASSWORD }}
+          edl-token: ${{ secrets.EARTHDATA_TOKEN }}
           event: merge
           commit: ${{ github.sha }}
           ref-label: ${{ github.ref_name }}
--- a/.github/workflows/lambda-benchmark.yml
+++ b/.github/workflows/lambda-benchmark.yml
@@ -45,6 +45,7 @@ jobs:
     outputs:
       affecting: ${{ steps.d.outputs.affecting }}
       function_name: ${{ steps.d.outputs.function_name }}
+      worker_note: ${{ steps.d.outputs.worker_note }}
     steps:
       - uses: actions/checkout@v5
         with:
@@ -83,13 +84,20 @@ jobs:
             if echo "$CHANGED" | grep -qE '^(src/zagg/|deployment/aws/|pyproject\.toml)'; then touched=true; fi
           fi
 
+          note=""
           if [ "$deployable" = "true" ] && [ "$touched" = "true" ]; then
             echo "affecting=true" >> "$GITHUB_OUTPUT"
             echo "function_name=$TEST_FN" >> "$GITHUB_OUTPUT"
           else
             echo "affecting=false" >> "$GITHUB_OUTPUT"
             echo "function_name=$STABLE_FN" >> "$GITHUB_OUTPUT"
+            # Benchmarking stable, but the change touches deployed code -> the
+            # numbers don't reflect this PR's worker. Flag it in the comment.
+            if [ "$touched" = "true" ]; then
+              note="Worker = stable \`main\`, not this PR's code — it changes lambda-deployed code but wasn't redeployed, so cost/runtime here don't reflect it. Real numbers land on merge."
+            fi
           fi
+          echo "worker_note=$note" >> "$GITHUB_OUTPUT"
 
   # Build the arm64 zips and update process-shard-test in place (only when the
   # change is lambda-affecting). Serialized with the benchmark via the workflow
@@ -174,6 +182,7 @@ jobs:
           commit: ${{ github.event.pull_request.head.sha }}
           ref-label: ${{ github.head_ref }}
           pr-number: ${{ github.event.pull_request.number }}
+          worker-note: ${{ needs.detect.outputs.worker_note }}
 
       - name: Upsert PR comment
         uses: actions/github-script@v7
--- a/.github/workflows/lambda-benchmark.yml
+++ b/.github/workflows/lambda-benchmark.yml
@@ -77,7 +77,9 @@ jobs:
           else
             if [ "$EVENT" = "pull_request" ]; then BASE="$PR_BASE"; HEAD="$PR_HEAD"; else BASE="$PUSH_BEFORE"; HEAD="$PUSH_AFTER"; fi
             if ! git cat-file -e "${BASE}^{commit}" 2>/dev/null; then BASE="${HEAD}~1"; fi
-            CHANGED=$(git diff --name-only "$BASE" "$HEAD" 2>/dev/null || echo "")
+            # No error-masking: a git diff failure aborts the run (set -e) rather
+            # than silently reading as "not affecting" -> benchmarking stale code.
+            CHANGED=$(git diff --name-only "$BASE" "$HEAD")
             if echo "$CHANGED" | grep -qE '^(src/zagg/|deployment/aws/|pyproject\.toml)'; then touched=true; fi
           fi
 
@@ -148,7 +150,7 @@ jobs:
     if: >-
       !cancelled() &&
       needs.detect.result == 'success' &&
-      needs.deploy-test.result != 'failure' &&
+      (needs.deploy-test.result == 'success' || needs.deploy-test.result == 'skipped') &&
       github.event_name == 'pull_request' &&
       github.event.pull_request.head.repo.full_name == github.repository &&
       github.event.pull_request.draft == false
@@ -195,7 +197,7 @@ jobs:
     if: >-
       !cancelled() &&
       needs.detect.result == 'success' &&
-      needs.deploy-test.result != 'failure' &&
+      (needs.deploy-test.result == 'success' || needs.deploy-test.result == 'skipped') &&
       (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
     runs-on: ubuntu-latest
     # contents: write to push the series + charts to the benchmarks branch.
--- a/.github/workflows/lambda-benchmark.yml
+++ b/.github/workflows/lambda-benchmark.yml
@@ -1,11 +1,19 @@
 name: Lambda Benchmark
 
-# Cost/runtime regression tracking on real Lambda (issue #110).
+# Cost/runtime regression tracking on real Lambda (issues #110, #25).
 #   * non-fork PR to main, non-draft -> run + post an (ephemeral) PR comment
 #   * push to main (a merge)        -> run + retain to the `benchmarks` data
 #                                      branch and re-render the Pages charts
-# Fork PRs are handled on demand (label / slash-command) in
-# lambda-benchmark-command.yml so untrusted code never auto-runs with creds.
+# Tier 2 (issue #25): a benchmark only reflects the PR if the DEPLOYED worker is
+# the PR's code. When the change is "lambda-affecting" (touches src/zagg,
+# deployment/aws, or pyproject) and the deploy is configured, we build the arm64
+# zips and update the `process-shard-test` function, then benchmark against it.
+# Otherwise we benchmark the stable function (correct for orchestrator/doc-only
+# changes, and the path before the AWS-side deploy is set up). Fork PRs never
+# deploy (handled in lambda-benchmark-command.yml).
+#
+# The whole workflow is serialized (a single shared `test` function), so two
+# runs can't deploy/benchmark over each other.
 
 on:
   pull_request:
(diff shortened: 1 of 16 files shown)


.github/workflows/publish.yml
```diff
@@ -116,3 +132,94 @@ jobs:
...
+
+  # Update the PRODUCTION Lambda in place (tags only). Behind the `production`
+  # GitHub Environment (add a required-reviewer rule). publish-layer-version +
+  # update-function-configuration handles a deps change that the function-code
+  # update alone would miss. Needs `distribute` so the layer is staged in S3 for
+  # publish-layer-version (the layer zip can exceed the 50 MB direct-upload cap).
+  deploy-prod:
+    name: Deploy production Lambda
+    needs: [github-release, distribute]
+    if: vars.LAMBDA_RELEASE_ROLE_ARN != ''
```
espg: 🤖 *from Claude (review)*

`deploy-prod`'s guard only checks `LAMBDA_RELEASE_ROLE_ARN`, but the job also consumes `LAMBDA_DIST_BUCKET` (line 205/215), `LAMBDA_AWS_REGION` (201/207), and `LAMBDA_PROD_FUNCTION_NAME` (204). #25's stated rule is that the OIDC jobs no-op when **any** of the four vars is unset.

Concretely, if `ROLE` is set but `LAMBDA_DIST_BUCKET` is unset: `distribute` is skipped (its guard requires BUCKET), and the only reason `deploy-prod` then does not fire a broken `publish-layer-version` with `S3Bucket=,S3Key=…` is GitHub's skip-propagation through `needs: distribute` — a skipped need silently skips this job because the `if` has no status function. That's the right outcome, but it's *implicit*: anyone who later adds `always()`/`!cancelled()` to this `if` (a common pattern) would resurrect a deploy against an unstaged layer. Suggest mirroring distribute's guard explicitly, e.g. `if: vars.LAMBDA_RELEASE_ROLE_ARN != '' && vars.LAMBDA_DIST_BUCKET != '' && vars.LAMBDA_AWS_REGION != '' && vars.LAMBDA_PROD_FUNCTION_NAME != ''` so the safety is in the guard, not in skip-propagation.

.github/workflows/publish.yml
```diff
@@ -116,3 +132,94 @@ jobs:
...
+        aws-region: ${{ vars.LAMBDA_AWS_REGION }}
+    - name: Update function (layer + code, in place)
+      env:
+        FUNCTION_NAME: ${{ vars.LAMBDA_PROD_FUNCTION_NAME }}
+        DIST_BUCKET: ${{ vars.LAMBDA_DIST_BUCKET }}
+        MINOR: ${{ needs.github-release.outputs.minor }}
+        REGION: ${{ vars.LAMBDA_AWS_REGION }}
+      run: |
+        set -euo pipefail
+        FN="${FUNCTION_NAME:-process-shard}"
```
espg: 🤖 *from Claude (review)*

`FN="${FUNCTION_NAME:-process-shard}"` silently falls back to the hardcoded prod function name (which is exactly `template.yaml`'s `FunctionName` default) when `LAMBDA_PROD_FUNCTION_NAME` is unset. Combined with the guard on line 185 only checking `LAMBDA_RELEASE_ROLE_ARN`: an operator who sets `ROLE` + `LAMBDA_DIST_BUCKET` + `LAMBDA_AWS_REGION` but forgets `LAMBDA_PROD_FUNCTION_NAME` gets a live `update-function-code`/`publish-layer-version` against a function literally named `process-shard` — an unintended in-place prod deploy rather than the intended no-op-when-unset. Recommend dropping the `:-process-shard` default and instead gating the job on `vars.LAMBDA_PROD_FUNCTION_NAME != ''` (see the guard suggestion on line 185) so a half-configured setup skips cleanly.

.github/workflows/publish.yml
espg: 🤖 *from Claude (review)* Fixed in 66e70c0: `deploy-prod`'s `if:` now gates on all four vars it consumes (LAMBDA_RELEASE_ROLE_ARN + LAMBDA_DIST_BUCKET + LAMBDA_AWS_REGION + LAMBDA_PROD_FUNCTION_NAME), and `distribute` now also requires LAMBDA_AWS_REGION — so the no-unstaged-layer-deploy safety no longer rests solely on skip-propagation. (The dist-bucket-must-be-in-prod-region operational note will go in the Phase 4 AWS-setup docs.)

.github/workflows/publish.yml
espg: 🤖 *from Claude (review)* Fixed in 66e70c0: dropped the `:-process-shard` fallback — `FN="$FUNCTION_NAME"`, now guaranteed non-empty by the job `if:` gate above, so a config that sets role+bucket+region but forgets LAMBDA_PROD_FUNCTION_NAME skips the job instead of deploying against the default prod name. Also made distribute_zips.sh's 4-zip check honest (trailing-\* globs so nullglob drops a missing layer and the friendly guard catches it).

.github/workflows/lambda-benchmark.yml
```diff
@@ -25,18 +33,126 @@ permissions:
...
+
   pr-comment:
+    needs: [detect, deploy-test]
     # Same-repo PRs only (forks can't be trusted with the role); skip drafts.
+    # !cancelled() lets this run when deploy-test was skipped (non-affecting);
+    # a deploy *failure* skips the benchmark (don't measure a broken deploy).
     if: >-
+      !cancelled() &&
+      needs.detect.result == 'success' &&
+      needs.deploy-test.result != 'failure' &&
```
espg: 🤖 *from Claude (review)*

`needs.deploy-test.result != 'failure'` admits `cancelled` as a pass-through, not just `success`/`skipped`. If `deploy-test` is **cancelled** (e.g. job timeout, or cancelled mid-sequence) the `publish-layer-version → update-configuration → wait → update-code` chain may have only partially run, leaving `process-shard-test` carrying stale or half-updated code — yet `cancelled != failure` is true, so `pr-comment` (and `merge-publish` at line 198, same expression) still benchmark it and attribute the result to the PR. That is exactly the silent-stale-worker outcome the tier-2 design exists to prevent (the comment above only reasons about `failure`).

Recommend an explicit allowlist instead of a denylist:

```yaml
(needs.deploy-test.result == 'success' || needs.deploy-test.result == 'skipped') &&
```

`skipped` keeps the non-affecting path (benchmark the stable fn) working; `success` is the only state in which the test fn is guaranteed freshly deployed. Apply at both line 151 and line 198.

.github/workflows/lambda-benchmark.yml
```diff
@@ -25,18 +33,126 @@ permissions:
...
+
+          # Does the change touch the deployed Lambda closure? (Over-detection is
+          # safe -- it just redeploys; under-detection would benchmark stale code.)
+          touched=false
+          if [ "$EVENT" = "workflow_dispatch" ]; then
+            touched=true
+          else
+            if [ "$EVENT" = "pull_request" ]; then BASE="$PR_BASE"; HEAD="$PR_HEAD"; else BASE="$PUSH_BEFORE"; HEAD="$PUSH_AFTER"; fi
+            if ! git cat-file -e "${BASE}^{commit}" 2>/dev/null; then BASE="${HEAD}~1"; fi
+            CHANGED=$(git diff --name-only "$BASE" "$HEAD" 2>/dev/null || echo "")
```
espg: 🤖 *from Claude (review)*

The `|| echo ""` swallows any `git diff` failure: on error `CHANGED` becomes empty, `grep` finds nothing, `touched` stays `false`, and the job silently routes the benchmark to the **stable** function instead of deploying the PR. That is the dangerous under-detection direction the comment two lines up warns about (`under-detection would benchmark stale code`) — a transient diff failure would attribute a stale-worker result to the PR with no signal.

Since `set -euo pipefail` is in effect and `BASE` is already proven to exist (the `git cat-file -e` guard just above), a real `git diff` failure here is unexpected and should fail the job loudly rather than be masked. Recommend dropping the fallback:

```bash
CHANGED=$(git diff --name-only "$BASE" "$HEAD")
```

Minor, same line: this is a two-dot diff (`BASE HEAD`). For a PR whose base branch has advanced, two-dot reports base-only changes as reversions (harmless over-detection) but can also net-cancel a path the PR touched if `main` independently made it identical — a three-dot `"$BASE"..."$HEAD"` (merge-base) measures only the PR's own changes and avoids both. Low-likelihood, noting for completeness.

.github/workflows/lambda-benchmark.yml
espg: 🤖 *from Claude (review)* Fixed in 696287d: the `pr-comment` + `merge-publish` gates now allowlist the deploy result — `(needs.deploy-test.result == 'success' || == 'skipped')` — so a **cancelled** (half-completed publish→config→wait→code) test deploy no longer slips through and gets benchmarked as the PR's worker.

.github/workflows/lambda-benchmark.yml
espg: 🤖 *from Claude (review)* Fixed in 696287d: dropped the `2>/dev/null || echo ""` masking on the `git diff` in `detect` — a diff failure now aborts the run (set -e) instead of silently reading as 'not affecting' and benchmarking the stale stable worker. (Kept two-dot diff; the over-detection direction is the safe one.)

docs/deployment/benchmark-cicd.md
```diff
@@ -288,3 +288,159 @@ you want the raw artifacts.
...
+
+## 9. Deploy + release IAM roles (OIDC)
+
+Two more roles, both trusting this repo via the OIDC provider from section 1
+(reuse the same `trust.json`):
+
+- **`zagg-benchmark-deploy`** (test tier) — update the test function + stage the
+  layer:
+  - `lambda:UpdateFunctionCode`, `lambda:UpdateFunctionConfiguration`,
+    `lambda:PublishLayerVersion`, `lambda:GetFunction` on `process-shard-test`
```
espg: 🤖 *from Claude (review)*

The documented deploy-role policy grants `lambda:GetFunction` but not `lambda:GetFunctionConfiguration`. `deploy_lambda.sh` calls `aws lambda wait function-updated` (between the config update and the code update), and that waiter polls the `GetFunctionConfiguration` API, not `GetFunction`. A role created exactly per this list would 403 at the `wait` step and break the deploy.

Suggest adding `lambda:GetFunctionConfiguration` here (and to the release-tier "same `lambda:*` actions" just below, which inherits the same list). Note the benchmark-invoke role in section 4 already grants it, so it is clearly required.

espg: 🤖 *from Claude (review)* Fixed in c25d09a: added `lambda:GetFunctionConfiguration` to both the deploy and release role grants (the `wait function-updated` waiter polls it — a role built exactly per the old text would 403 at the wait step), and added a note that account-level Block Public Access overrides the bucket-level setting.

espg:
🤖 *from Claude*

**Status: functionally complete; one pre-existing benchmark check is red and needs a decision (and a `workflow`-scoped push I can't make).**

All four phases are done, every self-review thread is folded, and the core CI is green — `test (3.12)`, `test (3.13)`, `ruff`, `build / build-x86_64`, `build / build-arm64` all pass. The new release/deploy jobs (`distribute`, `deploy-prod`, `deploy-test`, `merge-publish`) correctly **skip** when their vars/secrets are absent.

**The one red check** is `Lambda Benchmark / pr-comment`, failing at the `Assume AWS role (OIDC)` step with:

```
##[error]Input required and not supplied: aws-region
```

**Root cause — pre-existing, not introduced here.** The `pr-comment` job is the existing #112 benchmark job. It runs `./.github/actions/run-benchmark` (which calls `configure-aws-credentials` with `vars.BENCHMARK_AWS_REGION`) for *every* same-repo non-draft PR, with no guard for whether the benchmark AWS setup exists yet. On `main` the job has the identical firing condition, so it reds the same way whenever `BENCHMARK_AWS_REGION`/`BENCHMARK_ROLE_ARN`/`BENCHMARK_STORE_PREFIX` are unset — which they still are, since per this PR's ordering (#112 ✅ → benchmark setup → this PR → AWS update) the benchmark setup hasn't landed. So it fails rather than no-ops, contradicting this PR's "no-op when vars absent" promise for the *existing* benchmark.

**Proposed in-scope fix** — add a `benchmarkable` gate to the new `detect` job and gate `pr-comment` + `merge-publish` on it, so an unconfigured repo skips the benchmark cleanly. `detect` env gains `BENCH_ROLE`/`BENCH_REGION`/`BENCH_STORE` (from `vars.BENCHMARK_ROLE_ARN`/`BENCHMARK_AWS_REGION`/`BENCHMARK_STORE_PREFIX`), then:

```diff
           echo "function_name=$fn" >> "$GITHUB_OUTPUT"
           echo "worker_note=$note" >> "$GITHUB_OUTPUT"
+
+          # The benchmark itself (invoke role + region + store-prefix + a
+          # resolved function) must be configured, else run-benchmark's
+          # configure-aws-credentials aborts with "Input required ... aws-region".
+          benchmarkable=true
+          { [ -n "$BENCH_ROLE" ] && [ -n "$BENCH_REGION" ] && [ -n "$BENCH_STORE" ] && [ -n "$fn" ]; } || benchmarkable=false
+          echo "benchmarkable=$benchmarkable" >> "$GITHUB_OUTPUT"
```

and add `needs.detect.outputs.benchmarkable == 'true' &&` to both the `pr-comment` and `merge-publish` `if:` blocks (plus `benchmarkable:` to the `detect` `outputs:`). `actionlint` is clean on the patched file. Resolving the benchmark AWS setup instead would also make the job pass as-is, with no code change.

**Why it isn't pushed.** I prepared and verified this patch, but this routine's credentials lack the GitHub `workflow` scope — both `git push` and the contents API 403 with *"refusing to allow an OAuth App to create or update workflow `.github/workflows/lambda-benchmark.yml` without `workflow` scope"*. So the change to `.github/workflows/` can't be pushed from here; it needs a `workflow`-scoped push.

**Questions for review**
1. Apply the `benchmarkable` gate above (makes the benchmark no-op until configured, true to the PR's promise), or just complete the benchmark AWS setup (which makes `pr-comment` pass without any code change)?
2. Either way, this needs your `workflow`-scoped hand to land the workflow edit — flagging so it isn't silently stuck on a red check.

Marking `waiting` — the remaining item is the above decision + a `workflow`-scoped push. Everything else is complete and green.

---

deployment/aws/benchmark_cicd.yaml
```diff
@@ -0,0 +1,232 @@
...
+              StringLike:
+                "token.actions.githubusercontent.com:sub": !Sub "repo:${GitHubRepo}:*"
+      Policies:
+        - PolicyName: invoke-and-store
+          PolicyDocument:
+            Version: "2012-10-17"
+            Statement:
+              - Sid: InvokeBenchmarkFunctions
+                Effect: Allow
+                Action: [lambda:InvokeFunction, lambda:GetFunctionConfiguration]
```
espg: 🤖 *from Claude (review)*

The benchmark preflight concurrency probe (`compute_available_workers` → `concurrency.py`) calls three IAM actions this policy does not grant:
- `lambda:GetAccountSettings` (`_get_account_limit`, concurrency.py:170)
- `cloudwatch:GetMetricStatistics` (`_get_current_concurrent`, concurrency.py:184)
- `lambda:GetFunctionConcurrency` (`_get_function_reserved`, concurrency.py:207)

Each call is wrapped in try/except, so the missing perms will **not** 403 the job — but every benchmark run will then warn and fall back to the *unclamped* default, leaving the probe (the #28 FD-exhaustion / account-pool-saturation guard) effectively blind. The `concurrency.py` module docstring itself enumerates exactly these three as the required IAM perms. Suggest adding `lambda:GetAccountSettings` + `cloudwatch:GetMetricStatistics` (Resource `*`, both account-level) and `lambda:GetFunctionConcurrency` on the two function ARNs.

deployment/aws/benchmark_cicd.yaml
```diff
@@ -0,0 +1,232 @@
...
+                Resource:
+                  - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${BenchmarkFunctionName}"
+                  - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${TestFunctionName}"
+              - Sid: WriteBenchmarkStore
+                Effect: Allow
+                Action: [s3:PutObject, s3:GetObject, s3:DeleteObject]
```
espg: 🤖 *from Claude (review)*

These S3 write/read/delete + ListBucket grants on the **invoke** role appear unnecessary. The orchestrator never touches the output store directly: `runner.py` brackets both the template write and the finalize as Lambda *invokes*, with explicit comments — "The template write happens inside the function so the orchestrator only needs `lambda:InvokeFunction`; no direct S3 access to the output bucket is required" and the finalize "avoids requiring orchestrator-side S3 access". In the benchmark path `output_creds_event` is `None`, so the Lambda execution role (not this role) writes the store.

Unless something outside the lambda backend writes to `${BenchmarkStoreBucket}`, the `WriteBenchmarkStore` and `ListBenchmarkStore` statements can be dropped for least-privilege (the deploy role separately covers `lambda-test/*` staging). Not a 403 — just over-grant.

deployment/aws/benchmark_cicd.yaml
espg: 🤖 *from Claude (review)* Fixed in 86b3238: the invoke role now grants the pre-flight worker-sizing probe — `lambda:GetAccountSettings` + `cloudwatch:GetMetricStatistics` (Resource `*`, neither supports resource-level scoping) and `lambda:GetFunctionConcurrency` on the function ARNs. The #28/#63 guard no longer runs blind.

deployment/aws/benchmark_cicd.yaml
espg: 🤖 *from Claude (review)* Fixed in 86b3238: dropped the unused `s3:PutObject/GetObject/DeleteObject/ListBucket` block from the invoke role — confirmed against runner.py:679 (the Zarr template write happens inside the Lambda; the orchestrator does no direct S3). Updated the docs + the BenchmarkStoreBucket param description to match.

PR #125, opened 2026-06-29
Capture per-shard max memory and color benchmark markers by % of cap

### LINKED ISSUE #120: Capture per-shard max memory and encode it on the benchmark charts (marker color)
🤖 *from Claude*

The benchmark records `runtime_s`, `cost`, `cost/100km²`, and `worker_pct_timeout`, but **not memory** — yet memory is half the story on Lambda, and on the first run (#110/#25) it was *the* story.

## Why it matters

Lambda has two limits with very different character:
- **Runtime** — a *hard* wall (720 s deploy cap; can't exceed → timeout).
- **Memory** — a *soft* limit; you bump `MemorySize` as needed.

So memory headroom is the "do I need to scale this up?" signal and belongs on the chart next to runtime ("am I about to hit the wall?"). On run #1, the decisive numbers — `gain_bias` pegging **2046 / 2048 MB** (OOM, #117) and `tdigest_healpix_o10` at **1963 MB (96%)** — live only in CloudWatch, invisible in `series.parquet` and the charts.

## Asks

1. **Capture per-shard max memory.** Have the worker report `max_memory_mb` in its result metadata — `resource.getrusage(resource.RUSAGE_SELF).ru_maxrss` (KB→MB), which tracks Lambda's "Max Memory Used" closely. (`schema.py` `ProcessingMetadata` + `deployment/aws/lambda_handler.py`.) *Alternative:* the orchestrator parses the `REPORT` line via `LogType="Tail"` on the invoke — authoritative number, no worker change, fine since the benchmark is same-account — but needs a flag threaded through `_run_lambda`.
2. **Thread it through** `_run_lambda` → the run summary → `bench_metrics.build_record` (new `max_memory_mb` column in `RECORD_COLUMNS`, null-safe) → `series.parquet`.
3. **Encode it on the charts (marker color).** In `plot_series`, color each per-target marker by **% of the memory cap** (`max_memory_mb / (memory_gb·1024)`) on a green→red scale, so OOM-proximity is visible at a glance — runtime stays on the right axis, cost on the left. Red markers with `obs=0` then read instantly as "OOM'd."

## Acceptance

- `series.parquet` gains a `max_memory_mb` column; existing rows null-degrade.
- The two charts color markers by memory %; a legend/colorbar shows the scale.
- A test pins the record column + the color mapping (e.g. 1963 MB / 2048 → ~0.96 → near-red).

## Scope / related

Spans the deployed worker (`lambda_handler.py` / `schema.py`) + the benchmark scripts (`runner.py` summary, `.github/scripts/bench_metrics.py`, `plot_series.py`). Directly complements #117 — this is the metric that would have flagged the OOM on the chart instead of only in CloudWatch.

espg:
Closes #120

## What & approach

The benchmark recorded runtime/cost but not memory, yet memory was the decisive story on run #1 (`gain_bias` pegging 2046/2048 MB → OOM, `tdigest_o10` at ~96%) — visible only in CloudWatch. This makes peak per-shard memory a first-class metric end to end.

Following the issue's **primary** ask (worker-side `ru_maxrss`, not the cross-account `LogType="Tail"` alternative — the orchestrator deliberately omits Tail because cross-account callers lack CloudWatch access, see the comment in `_invoke_lambda_cell`):

- **Worker captures peak RSS.** `lambda_handler._handle_process` reads `resource.getrusage(RUSAGE_SELF).ru_maxrss` (KB→MB) *after* the write phase, so it covers read+index+aggregate+write, and stamps `metadata["max_memory_mb"]`. `ProcessingMetadata` gains the key (TypedDict made `total=False` since it's absent on the local-runner path; `phase_timings` is also now declared there rather than assigned off-schema).
- **Threaded back through the summary.** The worker stamps it on the result body; `runner.agg` rolls up the straggler (max) across cells into `summary["max_memory_mb"]` (mirrors the wall-time framing) and logs a `Worker peak memory: N MB (P% of cap)` line. `None` when no worker reported it (older layer / local backend).
- **Into the series.** `bench_metrics.build_record` reads `summary["max_memory_mb"]` (null-safe) and `RECORD_COLUMNS` gains a trailing `max_memory_mb` column, so existing parquet rows null-degrade via the `reindex` in `update_series.records_to_frame`. The PR-comment table gains `mem (MB)` and `% cap` columns.
- **On the charts (marker color).** `plot_series.make_figure` overlays the cost markers as a `scatter` colored by `% of the memory cap` on a green→red (`RdYlGn_r`) scale with a fixed `[0,1]` norm and a shared colorbar; runtime stays on the right axis, cost on the left. A red marker with `obs=0` then reads instantly as "OOM'd." A shared `memory_pct_of_cap(max_memory_mb, memory_gb)` helper in `bench_metrics` does the normalization (used by both the table and the chart).

## Phases

- [x] **Phase 1** — worker `ru_maxrss` capture (`lambda_handler.py`), `ProcessingMetadata` key (`schema.py`), summary rollup + log line (`runner.py`); handler + runner tests.
- [x] **Phase 2** — `max_memory_mb` column in `bench_metrics.RECORD_COLUMNS` + `build_record` (null-safe), `memory_pct_of_cap` helper, comment-table columns, `run_benchmark` print; tests.
- [x] **Phase 3** — green→red marker coloring + colorbar in `plot_series`, `memory_fractions` helper; tests.

## How tested

`uv run pytest -q` green (full suite). Targeted:
- `tests/test_lambda_handler.py::TestProcessEventDispatch::test_body_reports_max_memory_mb` — every successful invocation stamps a positive `max_memory_mb` in the body.
- `tests/test_runner.py::TestWorkerMemory` — the summary rolls up the straggler peak (`1963.0` over `[800, 1963, 1200, 900]`) and null-degrades when unreported; `TestSummaryKeysByteIdentical` pins the new key.
- `tests/test_benchmark.py` — `build_record` threads/null-degrades the column; `memory_pct_of_cap(1963, 2) ≈ 0.96` (the OOM-adjacent run #1 figure → red end); `memory_fractions` maps each row; `make_figure` colors markers + writes the figure.

`uv run ruff check --select=E,F,W,I --ignore=E501 src tests` clean.

## Questions for review

- **`ru_maxrss` units.** Linux reports `ru_maxrss` in **KiB** (Lambda is Linux), so I divide by 1024. The issue notes this tracks CloudWatch's "Max Memory Used" closely but isn't byte-identical; flagging in case you want the authoritative REPORT-line number instead (that would need `LogType="Tail"` threaded through `_run_lambda`, which only works same-account — happy to add it behind a benchmark-only flag if you prefer).
- **Pre-existing `ruff format` drift (not touched).** `ruff format --check` flags `src/zagg/schema.py` and `tests/test_runner.py` (plus a few others) **on `origin/main` already** — manual multi-line formatting the lint bot's `ruff check` doesn't enforce. Per the "don't fix unrelated CI failures" rule I left these alone; my own added lines are format-clean. Say the word if you'd like a separate format-only pass.
- **Marker color when memory is missing.** Legacy rows (no `max_memory_mb`) plot as an uncolored (grey, `plotnonfinite`) marker rather than dropping out. Reasonable?

---
_Generated by [Claude Code](https://claude.ai/code/session_01XfLQpJ8qTPh64VDT2nxpNS)_

### DIFF
--- a/.github/scripts/bench_metrics.py
+++ b/.github/scripts/bench_metrics.py
@@ -51,6 +51,9 @@ RECORD_COLUMNS = [
     "memory_gb",
     "price_per_gb_sec",
     "zagg_version",
+    # Appended for issue #120 (stable schema -> new columns go last). Peak worker
+    # RSS in MB; null on rows recorded before the worker reported it.
+    "max_memory_mb",
 ]
 
 
@@ -86,6 +89,22 @@ def _runtime_s(summary: dict) -> float | None:
         if val is not None:
             return float(val)
     return None
+
+
+def memory_pct_of_cap(max_memory_mb, memory_gb) -> float | None:
+    """Fraction of the Lambda memory cap a shard peaked at (issue #120).
+
+    ``max_memory_mb / (memory_gb * 1024)`` -- 0.0 at idle, ~1.0 at the OOM wall.
+    None when either input is missing or the cap is non-positive, so callers
+    (chart colouring, comment table) degrade gracefully on legacy rows. Not
+    clamped: a value slightly over 1.0 is a real OOM signal worth surfacing.
+    """
+    if max_memory_mb is None or memory_gb is None:
+        return None
+    cap_mb = memory_gb * 1024.0
+    if cap_mb <= 0:
+        return None
+    return max_memory_mb / cap_mb
 
 
 def build_record(
@@ -130,6 +149,8 @@ def build_record(
         "memory_gb": LAMBDA_MEMORY_GB,
         "price_per_gb_sec": LAMBDA_PRICE_PER_GB_SEC,
         "zagg_version": zagg_version,
+        # Null-safe: absent on an empty/legacy summary -> None -> null parquet cell.
+        "max_memory_mb": summary.get("max_memory_mb"),
     }
     return record
 
@@ -161,18 +182,21 @@ def comment_markdown(records: list[dict], worker_note: str = "") -> str:
     if worker_note:
         lines += [f"> ⚠️ {worker_note}", ""]
     lines += [
-        "| target | obs | runtime (s) | cost/shard | cost/100 km² | % timeout |",
-        "| --- | ---: | ---: | ---: | ---: | ---: |",
+        "| target | obs | runtime (s) | cost/shard | cost/100 km² | % timeout | mem (MB) | % cap |",
+        "| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |",
     ]
     for r in records:
+        mem_pct = memory_pct_of_cap(r.get("max_memory_mb"), r.get("memory_gb"))
         lines.append(
-            "| {target} | {obs} | {rt} | ${cost} | ${c100} | {pct} |".format(
+            "| {target} | {obs} | {rt} | ${cost} | ${c100} | {pct} | {mem} | {mempct} |".format(
                 target=_fmt(r.get("target")),
                 obs=_fmt(r.get("total_obs"), ",d") if r.get("total_obs") is not None else "n/a",
                 rt=_fmt(r.get("runtime_s"), ".1f"),
                 cost=_fmt(r.get("cost_per_shard_usd"), ".5f"),
                 c100=_fmt(r.get("cost_per_100km2_usd"), ".5f"),
                 pct=_fmt(r.get("worker_pct_timeout"), ".0%"),
+                mem=_fmt(r.get("max_memory_mb"), ".0f"),
+                mempct=_fmt(mem_pct, ".0%"),
             )
         )
     lines += [
--- a/.github/scripts/bench_metrics.py
+++ b/.github/scripts/bench_metrics.py
@@ -95,11 +95,15 @@ def memory_pct_of_cap(max_memory_mb, memory_gb) -> float | None:
     """Fraction of the Lambda memory cap a shard peaked at (issue #120).
 
     ``max_memory_mb / (memory_gb * 1024)`` -- 0.0 at idle, ~1.0 at the OOM wall.
-    None when either input is missing or the cap is non-positive, so callers
-    (chart colouring, comment table) degrade gracefully on legacy rows. Not
-    clamped: a value slightly over 1.0 is a real OOM signal worth surfacing.
+    None when either input is missing (``None`` or the float ``NaN`` a legacy
+    parquet row degrades to) or the cap is non-positive, so callers (chart
+    colouring, comment table) degrade gracefully on legacy rows. Not clamped: a
+    value slightly over 1.0 is a real OOM signal worth surfacing.
     """
     if max_memory_mb is None or memory_gb is None:
+        return None
+    # A legacy parquet row reads back as NaN, not None, after the reindex.
+    if isinstance(max_memory_mb, float) and math.isnan(max_memory_mb):
         return None
     cap_mb = memory_gb * 1024.0
     if cap_mb <= 0:
--- a/.github/scripts/plot_series.py
+++ b/.github/scripts/plot_series.py
@@ -15,15 +15,41 @@ from __future__ import annotations
 
 import argparse
 import math
+import sys
 from pathlib import Path
 
 import pandas as pd
+
+# Allow ``import bench_metrics`` whether run as a script or imported by tests.
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+import bench_metrics  # noqa: E402
 
 # Cost metric -> (column, human label) for the two retained-series figures.
 FIGURES = {
     "cost_per_shard": ("cost_per_shard_usd", "cost / shard (USD)"),
     "cost_per_100km2": ("cost_per_100km2_usd", "cost / 100 km² (USD)"),
 }
+
+# green->red colormap for memory headroom (issue #120): 0% of the cap reads
+# green, the OOM wall reads red. ``_r`` flips the stock green->red ramp.
+MEMORY_CMAP = "RdYlGn_r"
+
+
+def memory_fractions(sub: pd.DataFrame) -> list[float | None]:
+    """Per-row memory headroom as a fraction of the Lambda cap (issue #120).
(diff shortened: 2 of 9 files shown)


src/zagg/schema.py
```diff
@@ -16,14 +16,21 @@
 from zagg.grids.healpix import HEALPIX_BASE_CELLS, HealpixGrid
 
 
-class ProcessingMetadata(TypedDict):
+class ProcessingMetadata(TypedDict, total=False):
```
espg: 🤖 *from Claude (review)*

Flipping the whole TypedDict to `total=False` to make two keys optional silently makes *every* field optional, including the genuinely-required ones (`shard_key`, `total_obs`, `duration_s`, ...). That weakens mypy coverage for the handler's main result path — a worker that forgets `total_obs` now type-checks clean.

`typing_extensions` (already imported here) exports `NotRequired`, and the 3.12+ target supports it natively. Prefer marking just the two optional keys and keeping the dict `total=True`:

```python
from typing_extensions import NotRequired, TypedDict

class ProcessingMetadata(TypedDict):
    shard_key: int
    ...
    error: str | None
    max_memory_mb: NotRequired[float]
    phase_timings: NotRequired[dict[str, float]]
```

This keeps the required-field guarantees while still allowing the local-runner / non-profile paths to omit the new keys.

---

.github/scripts/bench_metrics.py
```diff
@@ -88,6 +91,22 @@ def _runtime_s(summary: dict) -> float | None:
     return None
 
 
+def memory_pct_of_cap(max_memory_mb, memory_gb) -> float | None:
```
espg, line 94: 🤖 *from Claude (review)*

Null-safety gap for the real legacy path. The docstring says "None when either input is missing," but `is None` only catches Python `None`. A legacy row read back from parquet (mixed old+new series after `append_records`) surfaces a missing `max_memory_mb` as float `NaN`, not `None`. `NaN is None` is `False`, so this returns `NaN / cap` = `NaN`, a float — not `None`.

Consequences are mostly benign downstream (the scatter plots `NaN` grey via `plotnonfinite=True`), but it contradicts the contract, and the null-degrade test (`test_build_record_max_memory_null_safe` / `test_memory_fractions_none_when_memory_missing`) only exercises Python `None` (`sub.assign(max_memory_mb=None)`), never the parquet-`NaN` path that real legacy rows take. Consider treating `NaN` as missing here (e.g. `math.isnan` guard, `math` is already imported) and adding a test row that round-trips through parquet/`reindex` so the legacy degrade is actually pinned.

---

.github/scripts/plot_series.py
```diff
@@ -15,16 +15,42 @@
...
+
 # Cost metric -> (column, human label) for the two retained-series figures.
 FIGURES = {
     "cost_per_shard": ("cost_per_shard_usd", "cost / shard (USD)"),
     "cost_per_100km2": ("cost_per_100km2_usd", "cost / 100 km² (USD)"),
 }
 
+# green->red colormap for memory headroom (issue #120): 0% of the cap reads
+# green, the OOM wall reads red. ``_r`` flips the stock green->red ramp.
+MEMORY_CMAP = "RdYlGn_r"
```
espg, line 35: 🤖 *from Claude (review)*

The color-mapping math is well pinned (`test_memory_pct_of_cap_maps_near_red`, `test_memory_fractions_maps_each_row` cover ~1963/2048 -> ~0.96), but `test_make_figure_colours_markers_by_memory` only asserts `make_figure(...) is True` and that the PNG exists. It does not actually verify that markers are coloured by the fraction array or that the colorbar was attached — so the chart-side acceptance criterion ("color markers by memory %, with a colorbar") isn't directly tested; a regression that drops the `scatter`/`colorbar` would still pass. Consider asserting against the returned `Figure` (e.g. that `fig.axes` includes a colorbar axis, or that the scatter's array equals `fracs`) rather than just file existence.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial fresh-context review of PR #125 against issue #120's acceptance criteria and CLAUDE.md. Overall this is solid and the acceptance criteria are met: `ru_maxrss` is read once at end-of-invocation (KB->MB, Linux-correct, covers read+index+aggregate+write), threaded handler -> `ProcessingMetadata` -> runner summary (straggler `max` across cells, mirroring the wall-time framing) -> `build_record` -> `RECORD_COLUMNS` (appended last for a stable schema, null on legacy rows) -> series.parquet, and both charts colour markers green->red by % of cap with a shared `[0,1]`-normalised colorbar. The record column and the ~1963/2048 -> ~0.96 (near red) mapping are pinned by tests. The `_LAMBDA_KEYS` byte-identical set is correctly updated and the local summary correctly omits the key.

Findings (none blocking; all non-`@espg`, so fold or leave standing at your discretion):

1. (medium) `schema.py` flips the whole `ProcessingMetadata` TypedDict to `total=False` to make two keys optional, which makes *all* keys optional and weakens mypy on the required result fields. Use `NotRequired` (already available via the `typing_extensions` import, native on 3.12+) on just `max_memory_mb`/`phase_timings`. Inline.

2. (low) `memory_pct_of_cap` / `memory_fractions` null-safety only catches Python `None`, not the float `NaN` that a real legacy parquet row degrades to after `append_records`/`reindex`; the contract says "None when missing" but returns `NaN`. Benign for rendering (NaN plots grey) but the null-degrade tests never exercise the parquet-`NaN` path. Inline.

3. (low) `test_make_figure_colours_markers_by_memory` only checks `is True` + file existence; it doesn't assert the markers carry the fraction array or that a colorbar was attached, so the chart-colouring criterion isn't directly regression-guarded. Inline.

Notes/non-issues verified: `_fmt` null-degrades the new `mem (MB)`/`% cap` markdown columns to `n/a`; the runner log line's `memory_gb`/`max_memory_mb` are both in scope and guarded by `if max_memory_mb is not None`; `plotnonfinite=True` correctly renders missing-memory markers; `bbox_inches="tight"` in place of `tight_layout` is the right call alongside a stolen colorbar.

---

PR #127, opened 2026-06-29
Don't retry deterministic Lambda failures (OOM); expose max_retries

### LINKED ISSUE #119: Don't retry deterministic Lambda failures (OOM); expose max_retries and run the benchmark with max_retries=1

espg:
Closes #119

## What this does

`_invoke_lambda_cell` (`src/zagg/runner.py`) retried up to `max_retries` (default 3) on **any** non-timeout `FunctionError`. But every `FunctionError` on a synchronous `RequestResponse` invoke is **deterministic** for a given shard (a timeout, `Runtime.OutOfMemory`, or an exception that escaped the handler), so retrying an OOM just pays the compute again on a shard that can never succeed — and the retry's different outcome (e.g. `No data after filtering`) **masks** the real failure in the recorded result.

This PR:

1. **Stops retrying deterministic `FunctionError`s.** Removed the `if not is_timeout: continue` so any `FunctionError` returns immediately, exactly as timeouts already did. Transient throttle/network faults are a separate channel — they raise client-side exceptions caught in the `except` block and are still retried with backoff against the `retryable` list (`TooManyRequestsException`, `Rate exceeded`, `Read timeout`, …) — that block is **untouched**.
2. **Tags the failure mode.** Adds a `Runtime.OutOfMemory` → `Lambda OOM: …` tag so the recorded `error` reflects the OOM rather than a later retry's outcome.
3. **Threads `max_retries` through the API.** `agg(max_retries=3)` → `_run_lambda(..., max_retries=3)` → `_cell_work` closure → `_invoke_lambda_cell(..., max_retries=...)`. Lambda-only; the `local` backend ignores it. Default stays `3` (transient faults still want a bounded retry), so normal runs are unchanged.
4. **Benchmark passes `max_retries=1`** (`.github/scripts/run_benchmark.py` → `run_target` → `agg(...)`) so it measures one clean invocation and records a failure as a failure. This script is a benchmark *runner*, not a workflow under `.github/workflows/`, and is named in the issue's asks.

Approach note: with (1), `FunctionError`s no longer consume retry attempts regardless of `max_retries`, so the default `3` only ever affects transient client-side faults.

## Phases

- [x] Phase 1 — stop retrying `FunctionError`s + tag OOM in `_invoke_lambda_cell`; tests pin no-retry / OOM tag / timeout-not-retried / transient-retry / non-retryable-break.
- [x] Phase 2 — thread `max_retries` through `agg()` → `_run_lambda()` → `_invoke_lambda_cell()` (default 3); end-to-end passthrough tests.
- [x] Phase 3 — benchmark runner passes `max_retries=1`.

## How tested

`pytest tests/test_runner.py tests/test_benchmark.py` green; full local suite green (901 passed / 10 skipped). New tests:

- `TestInvokeLambdaCellRetry` — deterministic OOM `FunctionError` → `invoke.call_count == 1` (with `max_retries=5`, proving invariance to the budget), `error` starts `Lambda OOM:`, `retries == 0`; generic `FunctionError` → one invoke; timeout `FunctionError` → one invoke, `timeout=True`; transient client error (`TooManyRequestsException`) → retried with backoff (`call_count == 2`, one `time.sleep`); non-retryable client error → no retry, no sleep.
- `TestMaxRetriesPassthrough` — `agg(..., backend="lambda", max_retries=1)` reaches `_invoke_lambda_cell` with `max_retries=1`; default is `3`.

`ruff check --select=E,F,W,I --ignore=E501` clean on all touched files.

## Questions for review

- `tests/test_runner.py` is authored in a tight manual layout that is not `ruff format`-clean **independent of this change** (the format diff touches only pre-existing lines). New code matches the surrounding compact style and passes `ruff check` (E,F,W,I), which is what the PR lint bot runs. Left the file's existing formatting untouched to avoid reformatting unrelated lines — flag if you'd prefer a separate format pass.
- The run-level retry budget noted under the issue's "Related" is explicitly out of scope; the per-cell `dispatch`/`LAMBDA_RETRY` path in `dispatch.py` is untouched.
- `src/zagg/runner.py` is now ~1144 lines (was ~1125 on `main`; this PR adds ~14 net source lines), past the ~1000-line soft limit. Splitting the module is out of scope here and warrants its own discussion — flagging, not acting.

### DIFF
--- a/.github/scripts/run_benchmark.py
+++ b/.github/scripts/run_benchmark.py
@@ -102,6 +102,9 @@ def run_target(
             function_name=function_name,
             overwrite=True,
             profile=True,
+            # A benchmark measures one clean invocation and records a failure as
+            # a failure -- never re-invoke (and never pay) to re-fail (#119).
+            max_retries=1,
         )
 
     return bench_metrics.build_record(
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -1092,6 +1092,9 @@ def _invoke_lambda_cell(
                     is_timeout = True
                     last_error = f"Lambda timeout: {error_payload[:100]}"
                 elif "Runtime.OutOfMemory" in error_payload:
+                    # ``Runtime.OutOfMemory`` is AWS's documented errorType for an
+                    # OOM-killed invocation; if AWS reworded it the tag would just
+                    # fall to the generic branch below (still no retry).
                     last_error = f"Lambda OOM: {error_payload[:100]}"
                 else:
                     last_error = f"Lambda error ({function_error}): {error_payload[:100]}"
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -1068,14 +1068,22 @@ def _invoke_lambda_cell(
             function_error = response.get("FunctionError")
             is_timeout = False
             if function_error:
+                # Every ``FunctionError`` on a synchronous invoke is deterministic
+                # for a given shard -- a timeout, ``Runtime.OutOfMemory``, or an
+                # exception that escaped the handler -- so none are retried: they
+                # all return immediately, exactly as timeouts already did (#119).
+                # (Transient throttle/network faults are a separate channel,
+                # retried with backoff in the ``except`` block below.) The error
+                # is tagged with its real mode so a benchmark records an OOM as an
+                # OOM rather than masking it behind a later retry's outcome.
                 error_payload = response["Payload"].read().decode("utf-8")
                 if "Task timed out" in error_payload:
                     is_timeout = True
                     last_error = f"Lambda timeout: {error_payload[:100]}"
+                elif "Runtime.OutOfMemory" in error_payload:
+                    last_error = f"Lambda OOM: {error_payload[:100]}"
                 else:
                     last_error = f"Lambda error ({function_error}): {error_payload[:100]}"
-                if not is_timeout:
-                    continue
 
             result = json.loads(response["Payload"].read()) if not function_error else {}
             try:
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -88,6 +88,7 @@ def agg(
     output_endpoint_url: str | None = None,
     handoff: str = "pandas",
     profile: bool = False,
+    max_retries: int = 3,
 ) -> dict:
     """Run the aggregation pipeline.
 
@@ -141,6 +142,13 @@ def agg(
         ``phase_timings`` (read/index/aggregate) sub-dict, and the run prints a
         per-phase worker breakdown. Default ``False`` leaves the worker path and
         per-cell event payload byte-identical -- no probe tax.
+    max_retries : int
+        Lambda-only (issue #119). Per-cell retry budget for *transient*
+        client-side faults (throttle/network) in ``_invoke_lambda_cell``;
+        deterministic Lambda ``FunctionError``s (timeout, OOM, unhandled
+        exception) are never retried regardless. Default ``3``. Ignored by the
+        ``"local"`` backend. Set to ``1`` (e.g. the CI benchmark) to measure one
+        clean invocation and record a failure as a failure.
 
     Returns
     -------
@@ -223,6 +231,7 @@ def agg(
             output_credentials=output_credentials,
             output_endpoint_url=resolved_endpoint,
             profile=profile,
+            max_retries=max_retries,
         )
     else:
         raise ValueError(f"Unknown backend: {backend!r} (expected 'local' or 'lambda')")
@@ -548,6 +557,7 @@ def _run_lambda(
     output_credentials=None,
     output_endpoint_url=None,
     profile=False,
+    max_retries=3,
 ):
     """Run processing via AWS Lambda invocation.
 
@@ -663,6 +673,7 @@ def _run_lambda(
             function_name=function_name,
             config_dict=config_dict,
             output_creds_event=output_creds_event,
+            max_retries=max_retries,
             max_workers=state["workers"],
             profile=profile,
         )
--- a/tests/test_runner.py
+++ b/tests/test_runner.py
@@ -467,6 +467,56 @@ class TestInvokeLambdaCellRetry:
         assert result["error"] == "AccessDeniedException"
 
 
+class TestMaxRetriesPassthrough:
+    """`agg(max_retries=...)` threads the per-cell retry budget down to
+    ``_invoke_lambda_cell`` on the lambda backend (#119), default 3."""
+
+    def _drive(self, monkeypatch, atl06_config, catalog_file, **agg_kwargs):
+        import boto3
+
+        import zagg.grids as grids_mod
+        from zagg import runner
+        from zagg.concurrency import ConcurrencyReport
+
+        captured = {}
+
+        monkeypatch.setattr(runner, "get_nsidc_s3_credentials",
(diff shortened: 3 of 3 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review against issue #119 acceptance criteria and CLAUDE.md.

**Verdict: the change is correct and the core acceptance criteria are met.** Verified end-to-end:

- Threading is real, not dropped: `agg(max_retries=3)` → `_run_lambda(..., max_retries=3)` → `_cell_work` closure (runner.py:676) → `_invoke_lambda_cell(..., max_retries=max_retries)`. Default `3` is consistent at every layer; benchmark passes `1`.
- Removing `if not is_timeout: continue` leaves every `FunctionError` path falling through to the single `return` — no path can loop a deterministic error. Confirmed there is no double-`read()` of `response["Payload"]`: the error branch reads it once into `error_payload`, and line 1099's second read is guarded by `if not function_error`, so it is skipped when an error is present.
- `retries: attempt` correctly reports `0` on a first-iteration deterministic-error return (invoked once / billed once); OOM returns `timeout=False` and `error="Lambda OOM: …"`, so the failure mode is recorded and not masked. Matches the issue's "recorded as OOM, billed once."
- Transient path (`except` block) is untouched; retry/backoff test passes.

Findings below are all **non-blocking** (nits / optional coverage). Nothing here gates the PR.

- **runner.py — OOM string match is order-sensitive (low risk, worth a one-line note).** The match keys on `"Runtime.OutOfMemory"` appearing in the decoded error payload. That string is what the issue specifies and what the test mocks, so this is correct for the stated contract. Just flagging that it depends on Lambda surfacing that exact token in the synchronous `RequestResponse` error payload; if AWS ever changes the wording the cell silently falls into the generic `Lambda error (...)` branch (still no retry, so no cost regression — only the OOM *tag* is lost). Acceptable as-is; a brief code comment that the literal is AWS's documented errorType would help future readers.

- **tests/test_runner.py — timeout no-retry case is not pinned in the new suite.** `TestInvokeLambdaCellRetry` covers OOM and generic `FunctionError` not being retried, plus transient-retry and non-retryable-break. It does not assert the pre-existing "timeout `FunctionError` → invoked once, `timeout=True`" behavior. That behavior is unchanged by this PR and lives in the same now-simplified branch, so a regression there would be silent. A one-line `test_timeout_function_error_not_retried` (assert `call_count == 1`, `result["timeout"] is True`) would lock the whole branch.

- **tests/test_runner.py — consider asserting `max_retries` is honored even on the deterministic path.** `test_oom_function_error_not_retried` uses the default `max_retries=3` and asserts one invoke. Passing `max_retries=5` there would more directly prove "deterministic error returns once *regardless of* max_retries" (the issue's exact wording). Minor — current test already demonstrates no-retry.

- **Module size (pre-existing, not introduced here).** `src/zagg/runner.py` is 1144 lines (1125 on `main`), past the ~1000-line soft limit in CLAUDE.md §4. This PR adds only ~14 net lines and did not cross the threshold, so per §4 ("don't fix pre-existing; flag it") this is just a flag, not a request — surfacing it so a future split can be planned.

Commit hygiene (title-only `phase N of issue #119`, no bodies, no authorship attribution) and the PR body are convention-clean.

---

espg:
🤖 *from Claude*

All three phases are complete and the adversarial self-review's diff-scoped findings are folded in. Marking `waiting` (awaiting review); leaving as draft.

**Folded from review:**
- Added `test_timeout_function_error_not_retried` — pins that the *pre-existing* timeout case still returns on a single invoke (`timeout=True`) through the now-shared simplified branch, so a regression there can't go silent.
- Bumped the OOM test to `max_retries=5` to directly prove the deterministic return is invariant to the retry budget (the issue's exact "invoked at most once regardless" wording).
- Added a brief comment noting `Runtime.OutOfMemory` is AWS's documented `errorType`, and that a reword would simply fall to the generic no-retry branch (only the tag is lost, no cost regression).

**Left standing (not folded):** the reviewer flagged that `src/zagg/runner.py` is ~1144 lines, past the ~1000-line soft limit (it was ~1125 on `main`; this PR adds ~14 net source lines). Splitting the module is out of scope for this change and warrants its own discussion per the conventions, so I have not touched it here.

**Test status:** `pytest tests/test_runner.py tests/test_benchmark.py` green (94 passed locally after the fold); full local suite green (901 passed / 10 skipped) on the phase-3 tree, and the affected suite re-verified green after folding. `ruff check --select=E,F,W,I --ignore=E501` clean on all touched files. CI on the draft PR will be the authoritative signal.

---

PR #129, opened 2026-06-29
Fix gain_bias dem_h broadcast dropping all photons on empty ATL03 segments

### LINKED ISSUE #116: gain_bias template reads 0 observations where tdigest reads 200k on the same shard (dem_h segment read)
🤖 *from Claude*

Surfaced by the first real run of the Lambda benchmark (#110 / #25), on `process-shard-test`.

## Symptom

On the **same shard, same granules**, the `gain_bias` template reads **zero** observations while `tdigest` reads ~200k:

| target | shard | granules | result |
| --- | --- | --- | --- |
| `tdigest_healpix_o11`   | `5347394812217655307` | 44 | **200,676 obs, 4900 cells with data** ✅ |
| `gain_bias_healpix_o11` | `5347394812217655307` | 44 | `No data after filtering … skipping` → **0 obs** ❌ |

CloudWatch (`/aws/lambda/process-shard-test`):

```
gain_bias_healpix_o11  Processed 44/44 files
gain_bias_healpix_o11  No data after filtering for shard 5347394812217655307 - skipping
gain_bias_healpix_o11  {"processing_complete", "total_obs": 0, "duration_s": 184.4, "error": "No data after filtering"}
...
tdigest_healpix_o11    Processed 44/44 files
tdigest_healpix_o11    Read 200,676 observations
tdigest_healpix_o11    Statistics: 4900 cells with data
```

## Likely cause

Diffing the two configs, the **only** `data_source` difference is the gain_bias segment‑level `dem_h` read, broadcast to the photon rows:

```yaml
# atl03_gain_bias_healpix.yaml — segments level only:
      variables:
        dem_h: "/{group}/geophys_corr/dem_h"
```

Filters (`signal_conf_ph` col 0 `ne -2`), `levels`, the segment→photon `link`, and `read_plan` are otherwise identical to the tdigest template. Since "No data after filtering" is logged before aggregation, the segment→photon broadcast for `dem_h` appears to drop **all** photon rows — most likely an empty/misaligned inner join (or a wrong `dem_h` path / index base), leaving an empty post-filter frame.

## Impact

- `gain_bias` produces no output on shards that demonstrably have data — i.e. the per‑chunk waveform/gain/bias template is effectively non‑functional right now.
- Every `gain_bias_*` entry in the cost/runtime benchmark is measuring "process nothing," so those points are meaningless until this is fixed.

## Repro

Dispatch shard `5347394812217655307` (NEON SERC AOP box, ATL03 v007) through `atl03_gain_bias_healpix.yaml` vs `atl03_tdigest_healpix.yaml` and compare `total_obs` — or run the pinned benchmark targets `gain_bias_healpix_o11` vs `tdigest_healpix_o11` (`tests/data/benchmark/`).

## Where to look

The segment-level variable read + segment→photon broadcast in the read path (the `dem_h` link via `ph_index_beg`/`segment_ph_cnt`, `index_base: 1`), and whether the join is dropping unmatched photons.

espg:
Closes #116

## Root cause

On the same shard/granules the `gain_bias` template read **0** observations while `tdigest` read ~200k. The only `data_source` difference is the segment-level `dem_h` read on the `segments` level (`src/zagg/configs/atl03_gain_bias_healpix.yaml`), which routes `gain_bias` through the segment→photon broadcast (`_read_segment_broadcasts` / `_broadcast_segment_to_base` in `src/zagg/processing/read.py`) that `tdigest` never touches.

Real ATL03 marks **empty segments** with `count == 0` **and `ph_index_beg == 0`**. Under `index_base: 1` (`atl03_gain_bias_healpix.yaml`), such a segment gives `beg = 0 - 1 = -1 < 0`, which raised `index_beg_arr[..]=0 is less than index_base=1` in `_broadcast_segment_to_base`. The per-group read in `worker.py` was wrapped in a bare `except Exception` logged at **DEBUG**, so the raise was swallowed for every beam/granule, `all_reads` stayed empty, and the worker reported the misleading `No data after filtering`. `tdigest`, with no broadcast, never raised.

`read_plan.plan_read` already skips empty (`cnt == 0`) segments; `_broadcast_segment_to_base` (and its twin `_expand_mask_to_base`) had no such guard. This was empirically confirmed in the issue thread: every beam on every granule raised the empty-segment `ValueError`, and a `cnt == 0` guard alone restored byte-identical obs to `tdigest`. The "index-space reconciliation" (step 3 of the original diagnosis) is **not** needed and was confirmed out of scope by @espg.

## Approach (phases)

- [x] **Phase 1 — make the failure visible.** Promote the swallowed per-group read error from DEBUG → WARNING in `worker.py` (a raised read is always a real error; a legitimately-empty group returns `None`, so this does not get noisy). Track a per-shard `read_errors` count in `ProcessingMetadata` and, when every group raised, report the real error instead of `No data after filtering`.
- [x] **Phase 2 — fix the broadcast.** Add an `if cnt == 0: continue` guard (before the `beg < 0` check) to `_broadcast_segment_to_base`, mirroring `plan_read`'s non-empty-only contract. Apply the identical guard to the twin `_expand_mask_to_base` (same latent bug; @espg approved the consistent guard).
- [x] **Phase 3 — regression tests.** Realistic ATL03-style segment table with interspersed empty segments (`ph_index_beg == 0, cnt == 0`) under `index_base = 1`, plus a planned-path run that does not start at file index 0.

## What changed

- `src/zagg/processing/read.py` — `cnt == 0` skip in `_broadcast_segment_to_base` and `_expand_mask_to_base`.
- `src/zagg/processing/worker.py` — per-group read error → WARNING; `read_errors` counter; all-raised path reports a real error.
- `src/zagg/schema.py` — `read_errors: NotRequired[int]` on `ProcessingMetadata`.
- `tests/test_processing.py` — new regression tests + updated `test_read_exception_still_releases_in_finally` to assert the new (correct) read-error reporting.

## How tested

- New regression tests assert `gain_bias` reads the **same observation count** as `tdigest` on a table with empty segments (the unit analogue of the locally-observed `20,710 == 20,710`), that the broadcast does not raise under `index_base=1` with `cnt == 0` empties, and that `dem_h` lands on the correct photons on **both** the full and planned read paths.
- `tests/test_processing.py` and `tests/test_runner.py` green via `uv run --extra test python -m pytest`.
- `ruff check --select=E,F,W,I --ignore=E501 src tests` and `ruff format --check` on touched files pass.

## Questions for review

1. **#117 sequencing.** The tracking note asks for #116 + #117 (gain_bias OOM) to land together as one PR (`Closes #116 #117`). This PR is scoped to the `dem_h` broadcast (the #116 root cause) only — #117 is a distinct root cause (full-grid vector allocation) inside the same template and re-surfaces once this fix re-enables the waveform path. Should the #117 footprint work be folded onto this branch, or land as its own PR? Left to @espg since folding it in changes this PR's agreed scope.
2. **mypy baseline.** Running mypy ad-hoc on the touched modules surfaces only pre-existing errors (missing `pandas`/`h5coro` stubs, the `phase_timings: dict | None` union, the `DataSourceDict` arg-type) on lines this PR does not touch; the new `read_errors` field and assignments are clean. Not "fixed" here per the convention on pre-existing unrelated failures — flagging instead.

---
_Generated by [Claude Code](https://claude.ai/code/session_01NPps6JqgR2hMhbE9Xm9R5t)_

### DIFF
--- a/src/zagg/processing/read.py
+++ b/src/zagg/processing/read.py
@@ -73,12 +73,19 @@ def _expand_mask_to_base(
     for p, keep in enumerate(coarse_mask):
         if not keep:
             continue
+        cnt = int(count_arr[p])
+        # Empty parents cover no base rows. Real ATL03 marks them with
+        # ``count == 0`` AND ``ph_index_beg == 0`` (issue #116), so under
+        # ``index_base=1`` they would otherwise give ``beg = 0 - 1 = -1`` and
+        # raise below; skip them, mirroring the non-empty-only contract
+        # ``read_plan.plan_read`` already uses (its ``cnt > 0`` skip).
+        if cnt == 0:
+            continue
         beg = int(index_beg_arr[p]) - index_base
         if beg < 0:
             raise ValueError(
                 f"index_beg_arr[{p}]={index_beg_arr[p]} is less than index_base={index_base}"
             )
-        cnt = int(count_arr[p])
         out[beg : beg + cnt] = True
     return out
 
@@ -131,12 +138,19 @@ def _broadcast_segment_to_base(
     else:
         out = np.zeros(total_base_size, dtype=seg_values.dtype)
     for p in range(len(seg_values)):
+        cnt = int(count_arr[p])
+        # Empty segments cover no photons. Real ATL03 marks them with
+        # ``count == 0`` AND ``ph_index_beg == 0`` (issue #116, see
+        # ``read_plan.plan_read``'s ``cnt > 0`` skip); under ``index_base=1``
+        # that gives ``beg = 0 - 1 = -1`` and would raise below, which is what
+        # made the gain_bias dem_h broadcast drop every photon. Skip them.
+        if cnt == 0:
+            continue
         beg = int(index_beg_arr[p]) - index_base
         if beg < 0:
             raise ValueError(
                 f"index_beg_arr[{p}]={index_beg_arr[p]} is less than index_base={index_base}"
             )
-        cnt = int(count_arr[p])
         if beg + cnt > total_base_size:
             raise ValueError(
                 f"segment {p} range [{beg}:{beg + cnt}] exceeds base size {total_base_size}; "
--- a/src/zagg/processing/read.py
+++ b/src/zagg/processing/read.py
@@ -78,7 +78,10 @@ def _expand_mask_to_base(
         # ``count == 0`` AND ``ph_index_beg == 0`` (issue #116), so under
         # ``index_base=1`` they would otherwise give ``beg = 0 - 1 = -1`` and
         # raise below; skip them, mirroring the non-empty-only contract
-        # ``read_plan.plan_read`` already uses (its ``cnt > 0`` skip).
+        # ``read_plan.plan_read`` already uses (its ``cnt > 0`` skip). Skipping
+        # intentionally bypasses the ``beg < 0`` validation for these parents --
+        # correct, since they map to zero base rows (a non-empty parent with
+        # ``beg < 0`` still raises).
         if cnt == 0:
             continue
         beg = int(index_beg_arr[p]) - index_base
@@ -143,7 +146,10 @@ def _broadcast_segment_to_base(
         # ``count == 0`` AND ``ph_index_beg == 0`` (issue #116, see
         # ``read_plan.plan_read``'s ``cnt > 0`` skip); under ``index_base=1``
         # that gives ``beg = 0 - 1 = -1`` and would raise below, which is what
-        # made the gain_bias dem_h broadcast drop every photon. Skip them.
+        # made the gain_bias dem_h broadcast drop every photon. Skip them; this
+        # intentionally bypasses the ``beg < 0`` / ``beg + cnt > base`` checks
+        # for empties (they map to zero base rows). A non-empty segment with
+        # ``beg < 0`` or an over-extending range still raises below.
         if cnt == 0:
             continue
         beg = int(index_beg_arr[p]) - index_base
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -265,15 +265,17 @@ def process_shard(
         phase_timings["read"] = time.time() - _read_t0
 
     if not all_reads:
-        # Distinguish a genuinely-empty read from one where every group raised
-        # (issue #116): the latter is a real read error masquerading as
-        # "no data", so report it as such instead of the misleading text.
+        # Distinguish a genuinely-empty read from one where a group read raised
+        # (issue #116): a raised read is a real error masquerading as "no data",
+        # so report it as such instead of the misleading text. Some groups may
+        # have returned ``None`` (legitimately empty) rather than raised, so the
+        # message is "no data AND N raised", not "all groups raised".
         if read_errors:
             logger.warning(
-                f"  Read raised on all groups for shard {shard_key} "
-                f"({read_errors} read error(s)) - skipping"
+                f"  No data after filtering for shard {shard_key} and "
+                f"{read_errors} group read(s) raised - skipping"
             )
-            metadata["error"] = f"All group reads raised ({read_errors} read errors)"
+            metadata["error"] = f"No data after filtering ({read_errors} group reads raised)"
         else:
             logger.info(f"  No data after filtering for shard {shard_key} - skipping")
             metadata["error"] = "No data after filtering"
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -198,6 +198,7 @@ def process_shard(
     use_arrow = handoff in ("arrow", "arrow-kernel")
     all_reads = []
     files_processed = 0
+    read_errors = 0
 
     # Opt-in per-phase timing (issue #100). Only allocated when profiling so the
     # default path stays byte-identical (no dict, no time.time() calls).
@@ -226,7 +227,14 @@ def process_shard(
                     if chunk is not None:
                         all_reads.append(chunk)
                 except Exception as e:
-                    logger.debug(f"  Error reading track {g}: {e}")
+                    # A raised read error is always a real failure: a
+                    # legitimately-empty group returns ``None`` (no exception),
+                    # so promoting this to WARNING does not get noisy on shards
+                    # where many granules simply contribute 0 photons (issue
+                    # #116). Logging it at DEBUG hid the dem_h broadcast failure
+                    # behind the misleading "No data after filtering" below.
+                    read_errors += 1
+                    logger.warning(f"  Error reading track {g}: {e}")
                     continue
 
             files_processed += 1
@@ -251,12 +259,24 @@ def process_shard(
 
     logger.info(f"  Processed {files_processed}/{len(granule_urls)} files")
     metadata["files_processed"] = files_processed
+    if read_errors:
+        metadata["read_errors"] = read_errors
     if profile:
         phase_timings["read"] = time.time() - _read_t0
 
     if not all_reads:
-        logger.info(f"  No data after filtering for shard {shard_key} - skipping")
-        metadata["error"] = "No data after filtering"
+        # Distinguish a genuinely-empty read from one where every group raised
+        # (issue #116): the latter is a real read error masquerading as
+        # "no data", so report it as such instead of the misleading text.
+        if read_errors:
+            logger.warning(
+                f"  Read raised on all groups for shard {shard_key} "
+                f"({read_errors} read error(s)) - skipping"
+            )
+            metadata["error"] = f"All group reads raised ({read_errors} read errors)"
+        else:
+            logger.info(f"  No data after filtering for shard {shard_key} - skipping")
+            metadata["error"] = "No data after filtering"
         metadata["duration_s"] = (datetime.now() - start_time).total_seconds()
         if profile:
             metadata["phase_timings"] = phase_timings
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -3000,6 +3000,17 @@ class TestExpandMaskToBase:
         with pytest.raises(ValueError, match="less than index_base"):
             _expand_mask_to_base(coarse, ibeg, cnt, index_base=1, total_base_size=3)
 
+    def test_empty_parent_skipped_index_base_1(self):
+        # Issue #116 twin: an empty parent (``ph_index_beg == 0, cnt == 0``)
+        # under ``index_base=1`` must be skipped, not raise on ``beg = -1`` --
+        # the same fix applied to ``_broadcast_segment_to_base``. A non-empty
+        # parent with ``beg < 0`` still raises (covered by ``test_negative_beg``).
+        coarse = np.array([True, True, True])
+        ibeg = np.array([1, 0, 3])  # middle parent is an empty (ph_index_beg == 0)
+        cnt = np.array([2, 0, 2])
+        out = _expand_mask_to_base(coarse, ibeg, cnt, index_base=1, total_base_size=4)
+        np.testing.assert_array_equal(out, [True, True, True, True])
+
 
 class TestReadGroupCrossLevel:
     """Phase B: cross-level filters expand coarse verdicts to base-rate rows."""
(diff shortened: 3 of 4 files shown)


src/zagg/processing/read.py
```diff
@@ -131,12 +138,19 @@ def _broadcast_segment_to_base(
...
         out = np.zeros(total_base_size, dtype=seg_values.dtype)
     for p in range(len(seg_values)):
+        cnt = int(count_arr[p])
+        # Empty segments cover no photons. Real ATL03 marks them with
+        # ``count == 0`` AND ``ph_index_beg == 0`` (issue #116, see
+        # ``read_plan.plan_read``'s ``cnt > 0`` skip); under ``index_base=1``
+        # that gives ``beg = 0 - 1 = -1`` and would raise below, which is what
+        # made the gain_bias dem_h broadcast drop every photon. Skip them.
+        if cnt == 0:
+            continue
```
espg, line 154: 🤖 *from Claude (review)*

Correctness of the guard checks out: it only `continue`s when `cnt == 0`, so a non-empty segment with `beg < 0` still hits the `ValueError` below, and the `beg + cnt > total_base_size` raise is likewise untouched for non-empty segments. I confirmed both raise-contracts hold — `test_negative_beg_raises` (cnt=[3]) and `test_broadcast_out_of_bounds_raises` (cnt=[2,5]) both use non-empty counts, so the guard never short-circuits them.

One behavioral subtlety worth a one-line note in the comment (non-blocking): the guard now bypasses the `beg < 0` validation entirely for any `cnt == 0` segment, regardless of what `index_beg` actually holds. That is exactly right for real ATL03 (empties carry `ph_index_beg == 0`), but it does mean a hypothetically corrupt empty segment (e.g. `ph_index_beg = -5, cnt = 0`) is now silently skipped rather than flagged. Since an empty segment maps to zero base rows by definition, skipping is the correct semantics — just calling it out so the looser validation is intentional, not an oversight.

---

src/zagg/processing/read.py
```diff
@@ -73,12 +73,19 @@ def _expand_mask_to_base(
...
         if not keep:
             continue
+        cnt = int(count_arr[p])
+        # Empty parents cover no base rows. Real ATL03 marks them with
+        # ``count == 0`` AND ``ph_index_beg == 0`` (issue #116), so under
+        # ``index_base=1`` they would otherwise give ``beg = 0 - 1 = -1`` and
+        # raise below; skip them, mirroring the non-empty-only contract
+        # ``read_plan.plan_read`` already uses (its ``cnt > 0`` skip).
+        if cnt == 0:
+            continue
```
espg, line 86: 🤖 *from Claude (review)*

The twin guard in `_expand_mask_to_base` matches the broadcast helper exactly and is consistent with the planned-path call at line 460 (`_expand_mask_to_base(coarse_fmask, cf_ibeg, cf_cnt, ...)`), so cross-level filters get the same empty-parent skip. No `n_base` drift either: `n_base = int(cnt_arr.sum())` (line 348) is unaffected by empties since they contribute 0, so the broadcast/expand outputs stay length-aligned with `seg_global_idx`/`global_idx`. Looks correct.

---

src/zagg/processing/worker.py
```diff
@@ -251,12 +259,24 @@ def process_shard(
...
+        # "no data", so report it as such instead of the misleading text.
+        if read_errors:
+            logger.warning(
+                f"  Read raised on all groups for shard {shard_key} "
+                f"({read_errors} read error(s)) - skipping"
+            )
+            metadata["error"] = f"All group reads raised ({read_errors} read errors)"
+        else:
+            logger.info(f"  No data after filtering for shard {shard_key} - skipping")
+            metadata["error"] = "No data after filtering"
```
espg, line 281: 🤖 *from Claude (review)*

Minor wording nit (non-blocking): this branch fires on `not all_reads and read_errors`, which is reached whenever *no* group produced data and *at least one* raised — but some of the other groups may have returned `None` (legitimately empty) rather than raised. In that mixed case the message "Read raised on all groups" / "All group reads raised (N read errors)" overstates it (not literally all groups raised; some were simply empty). The `read_errors` count is accurate; only the prose is slightly too strong. Consider e.g. "No data and N group read(s) raised" to stay precise. Not a correctness problem — the WARNING + count still surface the real failure, which is the point of issue #116.

The WARNING-promotion itself is sound and won't get noisy: a legitimately-empty group returns `None` without raising (it doesn't increment `read_errors` or log here), so only genuine read failures hit this path.

---

tests/test_processing.py
```diff
@@ -3688,6 +3714,83 @@ def test_expression_filter_references_broadcast_dem_h(self):
...
+            _ShardGrid(),
+        )
+        gain_bias = _read_group(
+            self._atl03_like_h5(),
+            "gt1l",
+            self._atl03_like_data_source(with_dem_h=True),
+            0,
+            _ShardGrid(),
+        )
+        assert len(gain_bias) == len(tdigest) == 8
```
espg, line 3790: 🤖 *from Claude (review)*

This parity test is the right regression for issue #116 and would genuinely fail without the fix: with `index_base=1` and the first empty segment at `ph_index_beg=0, cnt=0`, the pre-fix broadcast computes `beg = 0 - 1 = -1` and raises, so the `with_dem_h=True` read returns 0 rows while `with_dem_h=False` returns 8 — exactly the "gain_bias 0 vs tdigest N" drop. Asserting `len(gain_bias) == len(tdigest) == 8` plus the exact per-photon `dem_h` values pins both the count parity and the alignment. The companion planned-path test (`test_planned_read_skips_empty_segments_index_base_1`) covers the separate `_planned_read_group` path with a bbox run that starts past file index 0 and straddles the empty — good, since that path was not exercised by the full-read parity test. Solid coverage.

---

tests/test_processing.py
```diff
@@ -4493,5 +4632,8 @@ def factory(*a, **k):
         closes = [e for e in log if e[0] == "close"]
         # close() fired exactly once despite the read raising — the finally ran.
         assert len(closes) == 1
-        # No data survived the failed read, so the shard reports the empty path.
-        assert meta["error"] == "No data after filtering"
+        # A raised read is a real error, NOT a legitimately-empty read (issue
+        # #116): the shard reports the read-error path and counts the failure,
+        # rather than the misleading "No data after filtering".
+        assert meta["error"] == "All group reads raised (1 read errors)"
```

Updating this assertion is correct, not a weakened test (§4): the worker's behavior genuinely changed — a read that raised now reports the read-error path and counts the failure instead of the misleading "No data after filtering". Asserting both `meta["error"]` and `meta["read_errors"] == 1` locks in the new contract. Note this string assertion is exact-match coupled to the message wording; if you adopt the "No data and N group read(s) raised" rewording I suggested on worker.py, update this literal in the same commit.

Grammar nit (cosmetic): "(1 read errors)" reads awkwardly for the singular case. Low priority given it's an internal diagnostic string, but "(N read error(s))" or pluralize-on-N would tidy it.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of all three phases against issue #116 and CLAUDE.md. **The fix is correct and in scope** — no blocking findings; the comments below are minor (wording/grammar) plus confirmations.

**Correctness (verified, no bug):**
- The `cnt == 0` guard in both `_broadcast_segment_to_base` and `_expand_mask_to_base` only short-circuits empty parents. A non-empty segment with `beg < 0` still raises, and `beg + cnt > total_base_size` still raises for non-empty segments. Both existing raise-contracts hold — `test_negative_beg_raises` (cnt=[3]) and `test_broadcast_out_of_bounds_raises` (cnt=[2,5]) use non-empty counts, so the guard never bypasses them.
- `n_base = int(cnt_arr.sum())` is unaffected by empties (they contribute 0), so broadcast/expand outputs stay length-aligned with `seg_global_idx`/`global_idx` on both full and planned paths. The planned path's cross-level call (`_expand_mask_to_base` at read.py:460) gets the same skip.
- WARNING-promotion is not noisy: a legitimately-empty group returns `None` without raising, so it never increments `read_errors` or logs here — only genuine read failures do.

**Test quality:** the parity test reproduces the 0-vs-N drop and would fail without the fix (pre-fix broadcast raises on the first empty under index_base=1); the planned-path test covers the separate `_planned_read_group` code path. Good coverage.

**Scope/conventions:** stays on the @espg-confirmed fix (skip guards + WARNING promotion + `read_errors` metadata/schema field); no index-space reconciliation, no swallowed errors, no weakened tests, terse and style-matching. Compliant.

**Diff-scoped nits to consider folding (all non-blocking):**
1. worker.py: the "All group reads raised" message overstates the mixed case (some groups may have returned `None` rather than raised); "No data and N group read(s) raised" is more precise — and update the coupled exact-match assertion at tests line 4638 in the same commit.
2. "(1 read errors)" — pluralize on N or use "(N read error(s))".
3. read.py: optionally note in the guard comment that it intentionally bypasses the `beg < 0` validation for empty segments (correct, since empties map to zero base rows).

---

src/zagg/processing/worker.py
```diff
@@ -251,12 +259,26 @@ def process_shard(
...
+        # (issue #116): a raised read is a real error masquerading as "no data",
+        # so report it as such instead of the misleading text. Some groups may
+        # have returned ``None`` (legitimately empty) rather than raised, so the
+        # message is "no data AND N raised", not "all groups raised".
+        if read_errors:
+            logger.warning(
+                f"  No data after filtering for shard {shard_key} and "
+                f"{read_errors} group read(s) raised - skipping"
+            )
+            metadata["error"] = f"No data after filtering ({read_errors} group reads raised)"
```
espg, line 278: 🤖 *from Claude (review)*

Message reword is correct and accurate. With `read_errors == 1` the f-string renders exactly `No data after filtering (1 group reads raised)`, which matches the updated test assertion in `tests/test_processing.py` verbatim. And the new phrasing avoids the earlier overclaim: a legitimately-empty group returns `None` (no exception, never enters `all_reads`) while only a raised read increments `read_errors`, so `not all_reads` with `read_errors > 0` genuinely means "no data AND ≥1 raised" — some groups may simply have returned `None`. No "all groups raised" overstatement remains. Comment at lines 268-272 documents this distinction accurately.

---

src/zagg/processing/read.py
```diff
@@ -73,12 +73,22 @@ def _expand_mask_to_base(
...
             continue
+        cnt = int(count_arr[p])
+        # Empty parents cover no base rows. Real ATL03 marks them with
+        # ``count == 0`` AND ``ph_index_beg == 0`` (issue #116), so under
+        # ``index_base=1`` they would otherwise give ``beg = 0 - 1 = -1`` and
+        # raise below; skip them, mirroring the non-empty-only contract
+        # ``read_plan.plan_read`` already uses (its ``cnt > 0`` skip). Skipping
+        # intentionally bypasses the ``beg < 0`` validation for these parents --
+        # correct, since they map to zero base rows (a non-empty parent with
+        # ``beg < 0`` still raises).
```
espg, line 84: 🤖 *from Claude (review)*

Expanded guard comments are accurate and correctly differentiated per function. `_expand_mask_to_base` has only the `beg < 0` check below, and this comment scopes its bypass claim to exactly that ("bypasses the `beg < 0` validation"). The companion comment in `_broadcast_segment_to_base` correctly widens it to "`beg < 0` / `beg + cnt > base`" because that function carries both checks. Both correctly state that a non-empty parent with an invalid range still raises. No correctness issue.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context review of fold commit `baa6878` only (`7c7fe9e..baa6878`). No blocking findings — the fold is correct.

- worker.py message reword renders exactly `No data after filtering (1 group reads raised)` at `read_errors == 1`, matching the updated test assertion in `tests/test_processing.py` verbatim, and no longer overclaims "all groups raised" (empty groups return `None`, only raises bump `read_errors`).
- read.py guard comments accurately describe the `cnt == 0` skip and are correctly scoped per function (`_expand_mask_to_base` bypasses only `beg < 0`; `_broadcast_segment_to_base` bypasses `beg < 0` / `beg + cnt > base`), both noting non-empty parents still raise.

No grammar, style, or CLAUDE.md issues introduced by this commit. Earlier accepted phases were not re-litigated.

---

espg:
🤖 *from Claude*

Status: all three phases are complete and both fresh-context self-reviews folded (no blocking findings). Locally green — `ruff check --select=E,F,W,I --ignore=E501 src tests`, `ruff format --check` on touched files, and `pytest` for the affected suites (`test_processing.py`, `test_runner.py`) all pass. The fix is the @espg-confirmed `cnt == 0` skip guard in `_broadcast_segment_to_base` and its `_expand_mask_to_base` twin, plus the per-group read error promoted to WARNING with a `read_errors` count; index-space reconciliation was confirmed out of scope.

CI runners are still queued/in-progress at the time of this run, so the draft is left in place (not flipped to ready) pending CI green. Next run: once CI is green and nothing is waiting on @espg, flip to ready for review. Note for @espg: Refs #117 — the gain_bias OOM is a distinct root cause re-exposed once this fix re-enables the waveform path; whether to fold #117 onto this branch (per the "land together" tracking note) is raised under Questions for review since it would change this PR's agreed scope.

---

PR #134, opened 2026-06-30
update benchmarks: forward sharded-vs-inner tdigest/HEALPix matrix

### LINKED ISSUE #133: update benchmarks
```
[gain_bias_healpix_o11] obs=200676 runtime_s=173.392761 cost/shard=$0.0046238300790348 cost/100km2=$0.0456264566267214 max_memory_mb=1162.44921875
[tdigest_healpix_o11] obs=200676 runtime_s=154.942201 cost/shard=$0.0041318126856268005 cost/100km2=$0.04077138845245823 max_memory_mb=1577.97265625
[gain_bias_healpix_o10] obs=0 runtime_s=0.0 cost/shard=$0.0 cost/100km2=$0.0 max_memory_mb=None
[tdigest_healpix_o10] obs=939217 runtime_s=214.100655 cost/shard=$0.005709379346754 cost/100km2=$0.014084576242935166 max_memory_mb=891.54296875
[gain_bias_rect_3km] obs=207911 runtime_s=174.814317 cost/shard=$0.0046617384285756 cost/100km2=$0.051797093650839995 max_memory_mb=1763.46484375
[tdigest_rect_3km] obs=0 runtime_s=0.0 cost/shard=$0.0 cost/100km2=$0.0 max_memory_mb=None
[gain_bias_rect_6km] obs=745342 runtime_s=277.417962 cost/shard=$0.0073978493090616 cost/100km2=$0.02054958141406 max_memory_mb=1149.16796875
[tdigest_rect_6km] obs=745342 runtime_s=228.560979 cost/shard=$0.0060949899147972 cost/100km2=$0.016930527541103333 max_memory_mb=1853.05078125
```

Things that don't make sense:

1. gain_bias_rect_3km has 1763.46 megs/mem ; gain_bias_rect_6km has 1149.16 megs/mem. That's an **increase** in memory for 500k **less** observations.
2. tdigest_healpix_o11 has 200676 obs, 1577.97 megs/mem ; tdigest_healpix_o10 has 939217 , but uses almost *half* the memory at 891.54 megs/mem

In short -- why are heavier shards running lighter?

espg:
Closes #133

## What / approach

Implements the forward benchmark agreed on the issue thread (the "New forward benchmark — design" + "Plan amendment — Phase 8" comments; @espg confirmed keeping this on #133). The old 8-target rect/gain_bias matrix is frozen and replaced by a forward **2x3 matrix** — all `tdigest` / HEALPix / `handoff=arrow` — comparing the ShardingCodec (#108) against regular inner chunks across orders **o9 / o10 / o11**:

| | sharded codec | inner chunks |
|---|---|---|
| o9 | `tdigest_healpix_o9_sharded` | `tdigest_healpix_o9_inner` |
| o10 | `tdigest_healpix_o10_sharded` | `tdigest_healpix_o10_inner` |
| o11 | `tdigest_healpix_o11_sharded` | `tdigest_healpix_o11_inner` |

The codec is the experimental variable (`sharded: true|false`) carried in the matrix definition; the old rows are frozen by removing them from `targets.json` and adding a `codec` column to the parquet (null on frozen rows), with a new 2x3 figure rendered above the frozen one. **All 6 targets (o9/o10/o11) are now live** — phase 1 landed the o9 shard map, so the full 2x3 matrix is committed.

## Phases (from the approved plan)

- [x] **Phase 1 — o9 shard map.** Built the densest HEALPix o9 shard over the NEON AOP box (CMR reachable this run), pinned `shard_key` `5347395636851376137` / `n_granules` `59` into `tests/data/benchmark/shardmaps/sm_healpix_o9.json`; added the `healpix_o9` shardmap entry + the two o9 targets to `targets.json` and removed the `_pending_o9` stanza. Clean densest lead (59 > 50 > 47 > 27, no tie). o9 drift coverage is automatic (drift test parametrizes over `shardmaps`); the live `ZAGG_BENCHMARK_DRIFT=1` rebuild matches the pin. Deleted `sm_healpix_o9.TODO.md`. `test_o9_row_is_pending_not_live` → rewritten to `test_o9_row_is_live`.
- [x] **Phase 2 — configs.** Added `atl03_tdigest_healpix_o9.yaml` (parent_order 9, chunk_inner 13, child_order 19, K=256); express the codec per-target (`sharded`) in `targets.json` and `run_benchmark.py` applies it to `config.output['grid']['sharded']` (3 configs, not 6). Also fixed the in-scope o10 config doc typo (header said "K = 16" / "parent_order 11" copied from o11 → corrected to K=64 / parent_order 10 / 512x512).
- [x] **Phase 3 — matrix + freeze.** Replaced the 8 old targets with the new ones; **all 6 o9/o10/o11 targets are live** (the o9 row landed in phase 1; the `_pending_o9` hold-out stanza is gone). Removing the old rect/gain_bias targets (and their now-orphaned rect shardmap entries) freezes them.
- [x] **Phase 4 — schema + record.** Appended `codec` to `bench_metrics.RECORD_COLUMNS` (new column last; null on legacy/frozen rows); threaded it through `build_record` + `run_benchmark`.
- [x] **Phase 5 — renderer.** `plot_series.py`: new figure from `codec.notna()` as a fixed 2x3 grid (cols = sharded/inner, rows = o9->o11; a not-yet-landed order renders a blank row) above the frozen figure from `codec.isna()`; same for `index.html`, the docs embed (`*_codec.png`, `codec_table.png`), and the latest-table. The frozen figure/table are unchanged.
- [x] **Phase 6 — docs.** `docs/deployment/benchmark.md`: new "Sharded vs inner-chunk" section on top; frozen historical section below. README plot-layout section updated to describe both figure families.
- [x] **Phase 7 — tests.** Matrix-consistency test for the forward targets; a plot test asserting the fixed 2x3 codec layout; a schema test for the `codec` column; an o9-hold-out test (no fabricated key, drift coverage automatic once the map lands); a backward-compat test for a parquet with no `codec` column.
- [ ] **Phase 8 — decouple the sharding object from the dispatch shard.** Add a `shard_order` knob (`parent_order < shard_order <= chunk_inner`) so the worker writes its dispatch region in per-sharding-object passes (default = parent, today's behavior). Makes large/dense shards (88S) writable under the 2 GB cap. Not gated on the benchmark. **Not started — left for a future run** (it's a behavioral change to the production write path that wants its own careful design + byte-identical-output validation; deferred rather than rushed).

## Phase 1 (o9 map) — unblocked

The earlier routine run could not build `sm_healpix_o9.json` because its egress denied `cmr.earthdata.nasa.gov` (`403`); it held the whole o9 row out under a `_pending_o9` stanza with no fabricated `shard_key`. This run had CMR reachable, so phase 1 is now done: built the map with the phase-2 o9 config (`parent_order 9`, `child_order 19`, `mortie` backend, swath, ATL03 v007, the NEON AOP box, 2018-10-13→2025-06-01 — 59 granules, 4 shards, 183 pairs), pinned the densest shard (key `5347395636851376137`, 59 granules — a clean lead over 50/47/27, no tie), promoted the `healpix_o9` shardmap entry + the two o9 targets into the live blocks, dropped `_pending_o9` and the TODO file, and flipped `test_o9_row_is_pending_not_live` to `test_o9_row_is_live`.

## How tested (phase 1)

- `uv run ruff check --select=E,F,W,I --ignore=E501 src tests` — clean; `ruff format --check` on touched files — clean.
- `uv run pytest tests/test_benchmark.py tests/test_benchmark_shardmap.py -q` — **65 passed, 3 skipped** (the 3 skips are the `ZAGG_BENCHMARK_DRIFT=1`-gated CMR drift checks for o9/o10/o11).
- `ZAGG_BENCHMARK_DRIFT=1 uv run pytest tests/test_benchmark_shardmap.py -q` — **3 passed**: the live CMR rebuild of all three maps (incl. o9) matches the pinned densest shard.
- `uv run pytest -q` (full suite) — 949 passed, 4 skipped, **1 pre-existing unrelated failure** (`test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, the dual-arch Lambda image build — env-specific locally, untouched by this change; flagged not fixed per conventions).

## Questions for review

- **Phase 8 scope.** Deferred (write-path behavioral change). Happy to take it next run, or split it into its own issue/PR if you'd rather keep this PR scoped to the benchmark matrix — your call.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EwCFFM6SHNmP4Kjez8NfJz)_


espg (commented):
🤖 *from Claude (review)*

**Verdict: the blocker-documentation approach is correct, and the phase-1 diff is sound.** The CMR egress denial is a genuine hard blocker with no offline workaround in this tree, so the "don't fabricate a `shard_key`, land a documented placeholder instead" choice is the right call per the repo conventions. The single-file diff is safe and CLAUDE.md-compliant. No blocking findings.

**1) Was the build genuinely blocked / is there an offline path?** Genuinely blocked, and I checked the obvious offline routes — none works:
- The tempting path is re-sharding the existing o10/o11 maps onto the o9 grid (their granule **sets are byte-identical**: both carry the same 59 granules, `total_granules: 59`, and the densest shard depends only on data+grid, not order). But `ShardMap.build` requires `catalog.granule_records()` to yield per-granule footprint geometry (`lats`/`lons` exterior rings) to intersect against grid cells, and the **committed shardmap JSON strips geometry** — each record holds only `id`/`s3`/`https`. So the committed maps cannot be re-sharded offline.
- No committed catalog dump exists either: there is no `*.parquet`/geoparquet anywhere in the tree, and `Catalog.from_geoparquet` has nothing to load.
- The catalog build is anonymous (README "Add a target" and the TODO both note no Earthdata creds are needed), so the only barrier is egress to `cmr.earthdata.nasa.gov` — exactly what's denied. Confirmed real blocker.
- One nit on the PR body, not the diff: "Questions for review" option (2) ("re-pin from an existing local catalog dump if one exists") is a dead end — no such dump exists and the committed maps lack the geometry to serve as one. Options (1)/(3) are the only live ones; (3) is reasonable since phases 2-8 don't need the o9 pin.

**2) TODO.md accuracy — all claims verified:**
- Grid math is exact: parent_order 9, child_order 19 -> 4^(19-9) = 1048576 = 1024x1024 cells; K = 4^(13-9) = 256 inner chunks at chunk_inner 13; inner chunk = 4^(19-13) = 64x64. All correct.
- The "drift + consistency tests parametrize over `targets.json`, so o9 is automatic once the entry lands" claim holds: `test_pinned_shardmap_no_drift` parametrizes over `MANIFEST["shardmaps"]` and `test_targets_manifest_consistent` iterates `manifest["targets"]`, both keyed off `targets.json`. Caveat to plan for in the real phase 1: the drift test's `_config_for_shardmap` raises `AssertionError` if **no target references the shardmap** — so the `healpix_o9` shardmap entry alone isn't enough; a target using it must land in the same commit or that test fails. (Per the PR's 2x3 plan that's the o9 sharded/inner targets in phase 3 — just don't land the `shardmaps` entry by itself.)

**3) Does the `.md` placeholder risk breaking a `*.json` glob?** No. Nothing in `tests/` or `.github/scripts/` globs the shardmaps dir — both tests read only the explicit paths listed in `targets.json`. The `.md` file is invisible to the suite. (The PR body reports 40 passed / 15 skipped, consistent with this.)

**4) CLAUDE.md compliance:** Commit message is title-only ("phase 1 of issue #133"), no authorship credit in commit or PR body, blocker flagged under a "Questions for review" heading with concrete options, label `implement`, draft state, `Closes #133`, and no scope creep into phases 2-8. All good.

Minor, out of scope (pre-existing, not in this diff): the committed `atl03_tdigest_healpix_o10.yaml` grid comment says "K = 16" but parent_order 10 / chunk_inner 13 gives K = 4^(13-10) = 64 (the o11 config's "K=16" is correct). Worth a fix when phase 2 touches the configs, but not this PR's concern.

---

espg (commented):
🤖 *from Claude (review)*

Phase 2 (commit `3c41d03`) looks correct and complete. I verified each acceptance criterion from the issue #133 plan against the diff and the surrounding code; findings below are all non-blocking (mostly confirmations).

**Correctness — verified good**

(1) `sharded` application in `run_benchmark.py` (lines 110-112) is right. `config.output` is a plain `dict` (`PipelineConfig.output`, config.py:100), so `config.output.setdefault("grid", {})["sharded"] = bool(target["sharded"])` writes exactly the path `get_sharded` reads (`config.output.get("grid", {}).get("sharded", False)`, config.py:1182). It's gated on `"sharded" in target`, so a target that omits the key leaves the config's own default untouched — frozen/legacy targets dispatch byte-identically, as the comment claims. Placement is before `from_config(config)`, so the grid object also sees the override. No duplicate application exists elsewhere.

(2) o9 config math and comments check out: parent_order 9 / child_order 19 -> 4^(19-9) = 1024x1024 cells; chunk_inner 13 -> 4^(19-13) = 64x64; K = 4^(13-9) = 256. Header text ("256 chunk regions + 256 CSR groups", "1024x1024", "K = 256") and the inline `K=256 per shard` annotation all agree. The o9 config correctly does NOT set `sharded` itself, leaving the matrix in control.

(3) o10 typo fix is fully applied — all five occurrences corrected: `parent_order 11`->10, `256x256`->`512x512`, header `K = 16`->`K = 64`, "writes 16 chunk regions + 16 CSR"->"writes 64 ... + 64", and the inline `K=16 per shard`->`K=64 per shard` on the `chunk_inner` line. The "writes N chunk regions" occurrence the plan called out is included.

**Tests — adequate**

`test_sharded_knob_applied_to_grid_config` exercises both columns: it loops `(("sharded", True), ("inner", False))` and asserts `get_sharded(config) is want` via a `fake_agg` capture, mirroring the existing `test_provisional_target_resolves_and_threads_handoff` pattern (monkeypatching `runner.agg`, which works because `run_target` does a call-time `from zagg.runner import agg`). Pairing `sharded=True` with the o9 config (K=256, where sharding is valid) is the right choice. The full suite is green (40 passed, 11 skipped) and ruff check/format are clean on both touched Python files. The synthetic manifest deliberately uses the `healpix_o10` shardmap rather than the real o9 shardmap (a `.TODO.md` placeholder landing in phase 3), so the test is self-contained — consistent with the phasing.

**Minor / optional (non-blocking)**

(a) The "config already sets `sharded`, target omits the key" edge case is handled correctly by the `"sharded" in target` guard (config value stands), but it isn't asserted by a dedicated test. Since no current config sets `sharded` and the behavior is a simple `in` check, an explicit test isn't required — noting only for completeness.

(b) In the test, the codec label `"inner"` is paired with `sharded=False` and `"sharded"` with `True`, but `codec` itself is never asserted to thread anywhere. That's correct for phase 2 (only `sharded` is wired; `codec` is descriptive manifest metadata), so no change needed — flagging only so a later reader doesn't expect `codec` plumbing here.

Nothing here blocks the phase. Net: ship it.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **phase 3** (commit `a77c76c`) — the matrix replacement + freeze in `targets.json`, plus the two dry-run test updates. Verdict: correct and safe; nothing blocking. Details below.

**Matrix correctness (active targets).** The 4 active targets (o10/o11 x sharded/inner) are internally consistent: each o-pair shares one config (`atl03_tdigest_healpix_o{10,11}.yaml`) and one shardmap (`healpix_o{10,11}`), differing only by `codec`/`sharded`, with `sharded:true`↔`codec:"sharded"` and `sharded:false`↔`codec:"inner"` paired correctly. All carry `aggregator=tdigest`, `grid_type=healpix`, `handoff=arrow`, `grid_size` o10/o11. The sharded pairing is valid — o10 config is K=64 and o11 is K=16 (chunk_inner 13), so `sharded:true` is meaningful for both.

**Freeze / orphan handling — correct.** Removing the rect + gain_bias targets, and removing the `rect_3km`/`rect_6km` entries from `shardmaps`, keeps both invariants satisfied: no target references a missing shardmap (`test_targets_manifest_consistent`), and no shardmap lacks a referencing target (the drift test's `_config_for_shardmap`). The `sm_rect_*.json` files still sit on disk but are now unreferenced — that's fine; only orphan *manifest entries* fail tests, orphan files don't. Drift test now parametrizes over exactly `['healpix_o11','healpix_o10']` (verified by loading the manifest). `uv run --extra test pytest tests/test_benchmark.py tests/test_benchmark_shardmap.py -q` → 52 passed, 2 skipped (the `slow`/`ZAGG_BENCHMARK_DRIFT` drift checks).

**o9 hold-out — done safely.** Both o9 columns AND the `healpix_o9` shardmap are held together in `_pending_o9`, out of both `targets` and `shardmaps`. No `shard_key` was fabricated: the template carries the literal string `"<densest>"`/`"<n>"`, and only a `sm_healpix_o9.TODO.md` placeholder exists (no fake `sm_healpix_o9.json`). The stanza `_comment` is a clear, specific pending marker with the blocker, the build command, and re-integration steps. Leak check: `run_benchmark.all_target_names` reads only `manifest["targets"]`, and `_resolve_target` consults only `targets` + `provisional_targets` — so `_pending_o9` cannot enter "run all" nor be dispatched by `--target` (a pending name raises a clean `KeyError`). The drift/consistency tests also read only top-level `shardmaps`/`targets`. So `_pending_o9` is fully inert. Putting the templates inside the manifest (vs a sidecar) is reasonable and keeps the re-integration diff trivial.

**Tests — updated, not deleted (good).** The two dry-run tests that named old targets were updated in place: `test_run_target_dry_run` → `tdigest_healpix_o11_sharded` (and asserts `aggregator=="tdigest"`), and `test_main_dry_run_writes_outputs` → `tdigest_healpix_o10_inner`. No tests were weakened or skipped.

**Minor / out-of-scope notes (not blocking phase 3):**

1. The PR body's phase-3 line says "replace the 8 old targets with the **6** new ones," but phase 3 lands **4** active targets and holds 2 (o9) in `_pending_o9`. The data is correct given phase 1 is blocked; consider tightening the body wording so the matrix count isn't misread.

2. Two genuinely stale references to removed target names survive this phase. Both are docs/usage strings (phase 5/6 territory) and neither breaks a test, but flagging since this phase removed the names they cite: `.github/scripts/run_benchmark.py:17` (usage docstring still shows `--target gain_bias_healpix_o11 --target tdigest_healpix_o11`), and `tests/data/benchmark/README.md` ("today's eight targets" at (1), the `rect_*` family layout, and the `gain_bias_healpix_o11_88s` recipe near line 163). Leave for phase 5/6 unless you want the run_benchmark docstring example corrected sooner, since it now names two nonexistent targets.

3. Not a phase-3 defect: the `gain_bias_*`/`tdigest_healpix_o11` strings in `test_benchmark.py` (comment-markdown tests ~191/231 and the synthetic plot-layout fixtures ~843-980) are arbitrary fixture strings that never load `targets.json`, so they correctly were not touched.

Nice work on the hold-out — the inert `_pending_o9` placement and the no-fabrication discipline are exactly right.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **Phase 4** (schema + record), commit `24069cf`. Scope: the three changed files in this commit's diff only. No material issues found — the phase is correct and complete against the plan in issue #133. Detail on the five review axes:

(1) **`codec` is truly LAST in `RECORD_COLUMNS`.** Appended after `max_memory_mb` (`bench_metrics.py` line 61) with no mid-list insertion, preserving the stable on-disk parquet column order; the test asserts the invariant directly (`RECORD_COLUMNS[-1] == "codec"`). The stable-schema rule is honored, so existing benchmarks-branch series won't have their schema shifted.

(2) **Reindex back-fills `codec` as null on old parquets.** `update_series.records_to_frame` reindexes to `RECORD_COLUMNS` (`update_series.py` line 31) and `load_series` builds the empty frame from the same list (line 39), so an existing parquet lacking `codec` gets the column added (NaN), not an error. The renderer's documented `codec.notna()` / `codec.isna()` split (planned phase 5) keys off exactly this column — consistent with the comment at `bench_metrics.py` lines 57-60. Note `memory_pct_of_cap` already documents that a reindexed legacy cell reads back as float NaN rather than None; the codec split test belongs to phase 5/7, so nothing to flag here.

(3) **Null-safe end to end.** `build_record` reads `context.get("codec")` (line 165) → `None` when absent; `run_target` sets `codec=target.get("codec")` (`run_benchmark.py` line 128) → `None` for targets without the key (frozen/legacy/provisional). A target with no `codec`, or an empty/legacy context, yields `None` → JSON null / parquet null with no crash. Mirrors the existing `max_memory_mb` `.get(...)` pattern exactly.

(4) **Tests cover all four required assertions.** `test_codec_column_is_last_and_threaded` asserts (a) codec is the last column, (b) `build_record` threads `context["codec"]` (`"sharded"`), and (c) nulls it when the context omits it (`codec is None`). `test_run_target_threads_codec_into_record` asserts (d) `run_target` threads the target's codec end to end via a dry-run on `tdigest_healpix_o10_inner` (which carries `codec: "inner"`, `sharded: false` in `targets.json`). I ran the suite locally at this commit: `pytest tests/test_benchmark.py` → 54 passed; the two new tests pass; `ruff check` and `ruff format --check` clean on the three touched files.

(5) **CLAUDE.md conventions.** Terse and mirrors the `max_memory_mb` append pattern (same comment cadence, same `.get` null-safety); behavioral change ships with tests in the same commit; no schema instability and no null-handling gap. The o9 row absence (only o10/o11 inner/sharded targets exist) is the documented phase-1 block, not a phase-4 defect — phase 4's `run_target` test correctly uses an o10 target that does exist.

No blocking or non-blocking findings to fold. Phase 4 is sound.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **Phase 5** (renderer), commit `a99dcdc` on `claude/133-update-benchmarks`. Reviewed the diff to `.github/scripts/plot_series.py`, `run_benchmark.py`, and the new tests in `tests/test_benchmark.py`. Ran `uv run --extra test --extra benchmark pytest tests/test_benchmark.py -q` → **61 passed**.

Overall this is a clean, well-scoped phase. The frozen/codec split is correct, the 2x3 layout is genuinely fixed, and the `_draw_panel` / `_render_panel_grid` / `_render_table` extraction is a faithful code move (no logic drift in the moved bodies). Findings below are minor; one is a real test gap I'd ask be closed before ready-for-review.

**(1) Correctness of the codec/frozen split — solid.** `_codec_mask` guards the absent-column case (`if "codec" not in df.columns`), so a pre-#133 parquet reads all-frozen rather than `KeyError`. I verified this directly against a synthetic frame with no `codec` column: `_frozen_history` → 1 row, `_codec_history` → 0, mask all-False. `make_figure` now filters via `_frozen_history` and `make_codec_figure` via `_codec_history`, and the new tests assert each excludes the other's rows (`test_make_codec_figure_renders_only_codec_rows`, `test_make_figure_frozen_ignores_codec_rows`). No leak.

**(2) 2x3 layout — genuinely fixed.** `_codec_layout` builds the grid from `CODEC_ROWS`/`CODEC_COLS` constants (not from observed data), so a not-yet-landed order renders as a blank row instead of collapsing the grid. `test_codec_layout_blanks_missing_order` confirms the o9 row is `[None, None]` while shape stays 3x2. `_render_panel_grid` handles `None` slots (`ax.axis("off")`) and the bottom-row-label loop uses `default=None`, so a fully-blank column won't crash. Good — this matches the o9-blocked reality of the PR.

**(3) Frozen behavior preserved byte-for-byte — yes.** The moved `_draw_panel` body (failed-run `x` at axes-fraction y, cost-over-runtime z-order via `set_zorder`+`patch.set_visible(False)`, bottom-row-only labels) and the colorbar + MB twin axis in `_render_panel_grid` are identical to the pre-refactor inline code — the diff is a pure extraction. The existing `test_make_figure_*` suite (memory colorbar, MB twin axis, circle runtime + bottom-row labels, cost-scatter-above-runtime, failed-run zeros) all still pass against `make_figure`, which now feeds them frozen-only rows — still semantically right since those tests use `_rec_row` (no `codec` ⇒ frozen).

**(4) Index ordering — correct.** `write_index` emits the forward section (codec table + codec charts) before the frozen section, and `test_plot_main_emits_codec_artifacts_above_frozen` asserts `index.html.index("Sharded vs inner-chunk") < index("Frozen historical")`. `latest.md`/`metrics.json` are written unconditionally in `main` and remain reachable by direct URL.

**Findings to consider:**

(a) **Test gap — the no-`codec`-column path is the one most worth a test, and it has none.** Every new test builds frames via `records_to_frame`, which always includes `codec` (it's in `RECORD_COLUMNS` after phase 4), so the `"codec" not in df.columns` branch of `_codec_mask` — the exact backward-compat path the live retained parquet hits until the first new merge lands — is never exercised by the suite. The code is correct (I verified manually), but per CLAUDE.md §4 every behavioral change ships a test, and this branch is the highest-risk one (a regression here would `KeyError` the whole Pages render on the existing series). Suggest a small test that drops/omits the `codec` column and asserts `make_figure`/`make_latest_table` still render and `make_codec_figure`/`make_codec_latest_table` return `False`.

(b) **Minor — machine-readable links can drop out of the index in one corner case.** In `write_index`, the `latest.md`/`metrics.json` links attach to the codec table if present, else to the frozen table (`if links and not codec_table_png`). If codec *figures* render but neither table does (no merge of either family producing table rows, yet figures exist), the links appear in neither block. The files are still written and reachable by raw URL, so it's cosmetic, but the index would silently omit them. Low priority; flagging for awareness.

(c) **Nit — silent last-wins on duplicate slots.** `_codec_layout` does `by_slot[(grid_size, codec)] = target`, so two targets colliding on the same `(order, codec)` would silently drop one. Not reachable with the agreed 6-target matrix; noting only as a latent assumption.

(d) **CLAUDE.md hygiene — clean.** `plot_series.py` is 651 lines (well under the ~1000 gate). No dead code, no scope creep — the `run_benchmark.py` change is a one-line docstring usage example update, appropriate to the renamed targets. The extraction reduces duplication rather than adding speculative abstraction.

Net: correctness is sound and the layout/ordering goals are met and tested. The only thing I'd want addressed before ready-for-review is (a), the missing absent-`codec`-column test. (b)/(c)/(d) are non-blocking.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of Phase 6 (docs), commit `becfe5e` — `docs/deployment/benchmark.md` + `tests/data/benchmark/README.md`. Nothing material to block on; the phase does what issue #133's plan called for. Details below.

**(1) Section order / freeze labelling — correct.** The new `## Sharded vs inner-chunk (tdigest, HEALPix)` section now leads (replacing the old `## Latest merge` at the top), the old sections are pushed below a `---` under `## Frozen historical benchmark`, and that section is explicitly labelled *"the **frozen** pre-#133 matrix … retired from the every-merge run … no new points are appended."* Demoted to `###` headings under the frozen parent. Matches the index order the Phase-5 renderer emits in `write_index` (forward section on top, frozen below).

**(2) Image embeds — all six match the renderer's filenames.** Cross-checked every raw-URL embed against `.github/scripts/plot_series.py` `main()`:
- `codec_table.png` ✓ (`make_codec_latest_table` → `outdir / "codec_table.png"`)
- `cost_per_shard_codec.png`, `cost_per_100km2_codec.png` ✓ (`{name}_codec.png` for `name in FIGURES`)
- `latest_table.png` ✓, `cost_per_shard.png` / `cost_per_100km2.png` ✓ (`{name}.png`)
- `metrics.json` / `latest.md` links ✓

No broken embeds — every docs link points at a filename the renderer actually writes.

**(3) Prose accuracy vs `targets.json` — accurate.** 2×3, sharded-vs-inner ShardingCodec (#108) A/B, all tdigest / HEALPix / arrow, orders o9/o10/o11; "two columns share one config + one shard map, differ only by codec" matches each target pair (same `config`/`shardmap`, differing `codec` + `sharded`). o9 pending its shard map (catalog/CMR query) and o9 K=256 both check out against `_pending_o9` (config `parent_order 9, K=256`) and the o9-held-out note.

**(4) README plot-layout update — correct and de-staled.** Both stale "today's eight targets" claims in the README are gone; it now describes the fixed forward 2×3 (cols sharded|inner, rows o9→o11 with a blank row for an unlanded order, keyed to `grid_size`/`codec` — matches `CODEC_COLS`/`CODEC_ROWS`/`_codec_layout`) and the retained data-driven frozen layout (rect-left/HEALPix-right, `_panel_layout`), with the shared "both families" notes factored out.

**(5) CLAUDE.md / codespell — clean.** Markdown-only, terse, no @-mentions, no `#N` misused for list items. Eyeballed the new prose for common codespell hits — none.

Two non-blocking nits (your call, neither needs action this phase):
- `plot_series.py:245` still has a stale *"nothing is hardcoded to today's eight"* in the `_panel_layout` docstring — same phrasing this phase scrubbed from the README. Out of Phase-6 scope (the diff doesn't touch that file), just flagging it so the "eight targets" wording doesn't survive only in the code.
- benchmark.md mixes intra-repo issue refs (`#108`) inline-bare with full-URL links for others (`[#108](…/issues/108)`). Purely cosmetic; renders fine either way.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of phase 7 (commit `ffc4dd2`, tests + docstring fold). **No material/blocking findings — the phase does what the plan asked.** Details below; the two inline notes are nits, not changes-required.

**Verification I ran (from the worktree):**
- `uv run --extra test --extra benchmark pytest tests/test_benchmark.py tests/test_benchmark_shardmap.py -q` -> **65 passed, 2 skipped** (the 2 skips are the network-gated drift params — expected without `ZAGG_BENCHMARK_DRIFT=1`).
- `ZAGG_BENCHMARK_DRIFT=1 ... --collect-only` on the drift file -> collects **exactly** `test_pinned_shardmap_no_drift[healpix_o11]` and `[healpix_o10]` — the two live shardmaps, **no o9**. Confirms the o9 hold-out and the "automatic once pinned" parametrization claim.

**(1) Forward matrix invariants — pinned.** `test_forward_matrix_is_tdigest_healpix_arrow_codec_ab` asserts, per live target: `aggregator==tdigest`, `grid_type==healpix`, `handoff==arrow`, `codec in {sharded,inner}`, `sharded is (codec=="sharded")` (the bool/label agreement), the name-encodes-order+codec invariant, and a complete A/B pair per order (`{sharded,inner}`, never a half-row). Reads committed JSON it doesn't author — not tautological.

**(2) o9 held out + drift-coverage wiring guarded.** `test_o9_row_is_pending_not_live` confirms o9 is absent from both `targets` and `shardmaps`, the `_pending_o9` stanza exists with both templates pointing at the o9 config, and the placeholder `shard_key` is non-numeric (no fabricated key). `test_o9_drift_coverage_is_automatic_once_pinned` calls `drift._config_for_shardmap` for every live shardmap *without* the network — this lookup is otherwise only exercised inside the skip-gated drift test, so it adds real (non-redundant) coverage and fails loudly (`AssertionError`) if a shardmap ever lacks a referencing config.

**(3) Backward-compat path genuinely exercised.** `codec` is in `bench_metrics.RECORD_COLUMNS` (appended for #133), so `records_to_frame` emits a frame *with* the column; `df.drop(columns=["codec"])` then hits the real `"codec" not in df.columns` guard at `plot_series.py:84` — the same path the pre-#133 live parquet takes. `_codec_mask(...).any()` / `_codec_history(...).empty` / `_frozen_history(...)["target"]` all assert against derived state the test doesn't set. Correct and non-tautological.

**(4) Docstring fold.** `_panel_layout` is genuinely data-driven (ranks on `(area, grid_size)`, no hardcoded count); the old "today's eight" wording was stale and the frozen-only rewrite is accurate. `_codec_layout` (the fixed 2x3) referenced by the new wording exists, and the "asserted via the renderer test" comment maps to `test_codec_layout_is_fixed_2x3_sharded_inner_by_order` (phase 5). Good.

**CLAUDE.md:** terse, no test weakening/skip-to-pass, behavioral change (the `codec`-column guard) is tested. Compliant.

Two non-blocking nits inline.

---

espg:
🤖 *from Claude*

Landed **phases 2–7** of the forward benchmark this run (one title-only commit each, with a fresh-context adversarial self-review posted after every push):

- **phase 2** (`3c41d03`) — configs: new `atl03_tdigest_healpix_o9.yaml` (parent_order 9, K=256), per-target `sharded` knob in `run_benchmark.py`, and the in-scope o10 config doc-typo fix (header "K = 16" / "parent_order 11" → K=64 / parent_order 10 / 512x512).
- **phase 3** (`a77c76c`) — matrix + freeze: the 4 o10/o11 codec targets are live; the rect/gain_bias targets (and their orphaned shardmap entries) are removed (frozen).
- **phase 4** (`24069cf`) — schema: `codec` appended **last** in `RECORD_COLUMNS`, threaded through `build_record` + `run_benchmark` (null on legacy/frozen rows).
- **phase 5** (`a99dcdc`) — renderer: a fixed 2×3 codec figure (cols sharded/inner, rows o9→o11; blank row for an unlanded order) + codec table rendered **above** the frozen figure/table; `index.html` ordered forward-first.
- **phase 6** (`becfe5e`) — docs: new "Sharded vs inner-chunk" section on top of `benchmark.md`, frozen section below; README plot-layout de-staled.
- **phase 7** (`ffc4dd2`) — tests: forward-matrix consistency, the 2×3 layout, the `codec` schema column, the o9 hold-out (no fabricated key), and a no-`codec`-column backward-compat path.
- **review folds** (`9bdde8d`) — folded self-review nits: tightened the o9 placeholder assert and the drift-coverage comment.

**Local (mirrors CI):** `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on touched files; `pytest tests/test_benchmark.py tests/test_benchmark_shardmap.py -q` → **65 passed, 2 skipped** (the 2 skips are the network/`slow`-gated CMR drift checks for healpix_o10/o11); full `pytest -q` green. PR CI: ruff + build green; the test matrix is running on the head commit.

**Self-review:** each phase's review came back with no blocking findings; the diff-scoped nits were folded forward (the no-`codec`-column test and the `_panel_layout` docstring rewrite both originated from self-review, plus the phase-7 placeholder/comment tightening). The wording nits about "6 new targets" were clarified in the PR body (4 live + 2 pending o9).

**What remains:**
- **Phase 1 (o9 shard map) — still blocked.** Building `sm_healpix_o9.json` needs an anonymous CMR query to `cmr.earthdata.nasa.gov`, which the routine egress denies (`403`). No `shard_key` was fabricated; the whole o9 row (both columns + the `healpix_o9` shardmap) is held out of the live matrix as inert `_pending_o9` templates so the consistency/drift tests stay green. The o9 config already exists; `shardmaps/sm_healpix_o9.TODO.md` has the exact build command. Both tests parametrize over `targets.json`, so o9 is covered the moment the map is pinned. The matrix runs and renders today with the o10/o11 rows.
- **Phase 8 (the `shard_order` worker knob) — deferred, not started.** It's a behavioral change to the production write path that warrants its own careful design + byte-identical-output validation; left for a future run rather than rushed. Open question on the PR thread: keep it on this PR next run, or split it into its own issue/PR?

Leaving the label as `implement` (no `waiting`/`blocked`) — phases continue next run, the plan is already approved. Still a draft.

---

espg:
> Phase 8 scope. Deferred this run (write-path behavioral change). Happy to take it next run, or split it into its own issue/PR if you'd rather keep this PR scoped to the benchmark matrix — your call.

yes, please do this.

espg:
🤖 *from Claude (review)*

Phase-1 diff (a6a1401) reviewed adversarially — looks correct. What I verified:

**Densest-shard pin reproduces.** `bench_metrics.select_densest_shard(sm_healpix_o9.json)` returns `(5347395636851376137, 59)`, matching the `healpix_o9` shardmaps entry. Per-shard granule counts are `[47, 59, 50, 27]`, so 59 is the *unique* max — no tie-break ambiguity, the pin is deterministic.

**Metadata matches the o10/o11 procedure.** o9 vs o10 share `source=CMR-STAC`, `provider=NSIDC_CPRD`, `collection=ATL03_007`, `version=007`, dates `2018-10-13..2025-06-01`, identical bbox, `backend=mortie`, `footprint=swath`, `mortie_order=13`, and `grid_signature` `healpix/nested, child_order 19, fullsphere`. Only the expected per-order fields differ (`parent_order 9`, `total_shards 4`, `total_pairs 183`, `shard_keys`).

**`_pending_o9` fully removed; o9 row internally consistent.** Both `tdigest_healpix_o9_{sharded,inner}` are now in `targets` (codec/`sharded` agree: `sharded:true|inner:false`), both reference `shardmap: healpix_o9` and `config: configs/atl03_tdigest_healpix_o9.yaml` (exists; its `parent_order 9 / child_order 19 / chunk_inner 13` match the shardmap signature). Live targets are now exactly `{o9,o10,o11} × {sharded,inner}` = 6. The manifest description no longer mentions a pending o9 row.

**Test rewrite keeps coverage.** `test_o9_row_is_live` asserts the stanza is gone and the pin is a real `int`; `test_every_live_shardmap_resolves_to_a_config` preserves the wiring invariant. No dangling `<densest>` / TODO-file references remain (the only surviving `_pending_o9` strings are the new "should be removed" assertions). The drift test now collects `test_pinned_shardmap_no_drift[healpix_o9]` (skipped offline by env-gate, same as o10/o11). The `# o9 row blank` at test_benchmark.py:1125 is the synthetic missing-order renderer test (history built from o10/o11 only), correctly reworded — not a stale live claim.

**Tests green:** `pytest tests/test_benchmark.py tests/test_benchmark_shardmap.py -q` → 65 passed, 3 skipped (the 3 CMR drift checks, env-gated).

espg (commented):
🤖 *from Claude (review)*

Phase 8 review (commit `fd56e4b`, parent `a6a1401`) — adversarial pass against the default-byte-identity requirement and issue #133 Phase 8.

**Verdict: clean — no blocking findings.** The default path is genuinely byte-identical and the split math is correct. Three nits below, none blocking.

What I verified (with the actual mortie indexing, orders parent 4 / inner 6 / child 7, shard at -78.5/-132.0):

- **Default outer block == old `block_index`.** `shard_object_block` computes `(inner_block · cells_per_chunk) // cells_per_shard_object`; at default `cells_per_shard_object == cells_per_shard == n_children`, and for all 16 inner chunks this equals the parent nested id returned by `block_index(shard_key)` (HEALPix nested ids are hierarchical, so `child_nested // 4^(child-parent) == parent_nested`). Confirmed identical.
- **Default local region == old `shard_local_region`.** `cell_offset - obj_block·cells_per_shard_object` equals the old `(inner_block - shard_block·K)·cells_per_chunk` for every chunk. Confirmed identical.
- **Split partition is correct.** At `shard_order=5` the 16 chunks partition cleanly into 4 objects of 4 chunks; every local offset stays in `[0, cells_per_shard_object)`; distinct object blocks are contiguous.
- **`_spec` outer chunk collapses at default** (`cells_per_shard_object == cells_per_shard`), so the template bytes are unchanged — and the byte-identity test proves the *whole store* (template + data) matches.
- **Companion/ragged relocation is behavior-preserving.** In the old code those writes were already unguarded by `_carrier_empty` and ran once per chunk_result; moving them above the accumulation loop runs them on exactly the same set (incl. empty carriers), indexed by `block_index`/`shard_key` not by object — unchanged.
- **Non-object-split / rectilinear fallback preserved.** Rectilinear lacks `shard_object_block`/`shard_object_slab_shape`, so `hasattr` is False → the else branch uses the old `shard_slab_shape`/`block_index`/`shard_local_region` and groups everything under one constant `shard_block` (one whole-tile slab) — byte-identical to pre-phase-8.
- **Validation boundaries** (`parent_order <= shard_order <= chunk_order`, requires `sharded`) are correct; `shard_order == chunk_order` yields object == one inner chunk (max split), which is sound.
- Tests pass locally (4 new + full `tests/test_processing.py` green); ruff clean; both touched modules well under the 1000-line limit.

---

src/zagg/config.py
```diff
@@ -1182,6 +1182,21 @@ def get_sharded(config: PipelineConfig) -> bool:
     return bool(config.output.get("grid", {}).get("sharded", False))
 
 
+def get_shard_order(config: PipelineConfig) -> int | None:
+    """Return the sharding-OBJECT order from the output grid config (issue #133 phase 8).
+
+    ``shard_order`` decouples the ShardingCodec object from the dispatch shard: an
+    order strictly between ``parent_order`` and ``chunk_inner`` sizes each sharding
```

nit (doc): this says "order **strictly between** `parent_order` and `chunk_inner`", but the grid validation actually allows the closed range `parent_order <= shard_order <= chunk_order` — `== parent_order` is the explicit default, and `== chunk_inner` is the maximal split (object == one inner chunk, exercised and sound). Tightening the wording to match (e.g. "an order in `[parent_order, chunk_inner]`; `== parent_order` keeps today's single-object behavior") would avoid a future reader thinking the endpoints are rejected. Non-blocking.

---

src/zagg/grids/healpix.py
```diff
@@ -448,7 +527,11 @@ def _shard(arr):
             cg = arr.chunk_grid
             cfg = cg["configuration"] if isinstance(cg, dict) else cg.configuration
             inner = tuple(int(c) for c in cfg["chunk_shape"])
-            shard = (self.cells_per_shard, *inner[1:])
+            # The ShardingCodec OUTER chunk == the sharding OBJECT (issue #133 phase
+            # 8): ``cells_per_shard_object`` cells, which is ``cells_per_shard`` (the
+            # whole dispatch shard) at the default ``shard_order`` and SMALLER when a
+            # finer ``shard_order`` is set. The inner read chunk is unchanged.
+            shard = (self.cells_per_shard_object, *inner[1:])
```
espg, line 535: 🤖 *from Claude (review)*

nit (stale comment): the inline comment you added here is accurate, but the block comment a few lines above (unchanged by this diff) still reads "outer (shard) chunk == `cells_per_shard`" and "The outer shard widens only the cells axis to `cells_per_shard` (== K inner chunks)". With this line now using `cells_per_shard_object`, that upstream prose is stale for the `shard_order`-set case. Worth a one-line tweak so the two comments don't contradict. Non-blocking.

---

tests/test_processing.py
```diff
@@ -197,6 +197,165 @@ def test_readback_places_cells_at_correct_positions(self, monkeypatch):
...
+            self.CHILD,
+            layout="fullsphere",
+            config=cfg,
+            chunk_inner=self.INNER,
+            sharded=True,
+            shard_order=shard_order,
+        )
+
+    @staticmethod
+    def _df(grid, shard_key):
```
espg, line 232: 🤖 *from Claude (review)*

nit (coverage gap): these split tests build the grid/df from `default_config()`, which has only scalar `cell`-resolution dense fields — no `resolution: chunk` companion, no `kind: vector` (trailing-dim), and no `kind: ragged` field. So under object-split (`shard_order=5`) the three "moved above the loop" paths are never exercised:

- the companion path (`_write_companion_columns`) — `_chunk_resolution_fields(default_config())` is empty,
- the vector trailing-dim invariant branch (`write.py` lines ~301-313, the `target_trailing_chunks != trailing` raise),
- ragged writes with non-empty payloads.

The risk is low because companion/ragged are indexed by `block_index`/`shard_key` (not by object) and run identically to the single-object path — but the byte-identity claim for those field kinds under split is currently unproven. A small follow-up using a config with a companion + a vector field (mirroring `TestWriteShardToZarr`/issue #29 fixtures) and asserting byte-identity default vs `shard_order==parent_order`, plus value-identity default vs split, would close it. Non-blocking.

---

espg (commented):
🤖 *from Claude (review)*

Inline nits for Phase 8 (commit `fd56e4b`). All three are non-blocking doc/test-coverage notes; the full verdict (clean, default byte-identity verified, split math correct) is in my earlier summary review on this commit.

---

espg:
🤖 *from Claude*

**Phase 8 landed** — `shard_order` sharding-object split (commit `fd56e4b`, review fold `39002f7`). This is the last item on the checklist, so all phases are now complete.

### What it does
Adds an opt-in `output.grid.shard_order` knob that **decouples the ShardingCodec object from the dispatch shard**. A ShardingCodec object normally spans the whole dispatch shard (outer chunk == `cells_per_shard`), so for a large/dense shard (88S) that single object can exceed the 2 GB write cap. `shard_order` (an order in `[parent_order, chunk_inner]`) sizes each object *smaller* than the dispatch shard: one dispatch shard then holds `4^(shard_order − parent_order)` objects, and the worker writes its region in one **accumulate→write→free pass per object**, bounding peak memory.

**Default is byte-identical to today.** Unset (or `== parent_order`) keeps one object per dispatch shard: `cells_per_shard_object == cells_per_shard`, `shard_objects_per_shard == 1`, the template outer chunk is unchanged, and the single `set_block_selection` is the pre-phase-8 write. Risky write-path behavior is therefore feature-flagged off by default (CLAUDE.md §2). Files: `config.py` (`get_shard_order` + validation), `grids/healpix.py` (object math + `_spec` outer chunk), `grids/__init__.py` (thread from config), `processing/write.py` (per-object loop), `tests/test_processing.py`. `deployment/aws/lambda_handler.py` is untouched — the knob is config-derived, so the Lambda worker picks it up through the forwarded config with no handler change.

### How tested
`tests/test_processing.py::TestShardOrderObjectSplit` (4 new):
- **byte-identity:** `shard_order` unset vs explicit `== parent_order` produce **byte-for-byte equal stores** (same keys, `v.to_bytes()` equal) — the phase-8 default-safety invariant.
- **split correctness:** at `shard_order` one finer (`shard_objects_per_shard == 4`), the union of the 4 per-object writes reconstructs the same logical array (`np.testing.assert_array_equal` on read-back).
- **multiple-objects-vs-one** and **validation rejections** (`shard_order` out of `[parent_order, chunk_inner]`; `shard_order` without `sharded=True`).

Full `tests/test_processing.py` green; **CI green on `fd56e4b`** (ruff, build, build-x86_64, build-arm64, test 3.12, test 3.13 all ✓). The fold `39002f7` is docstring/comment-only.

### Self-review (two fresh-context adversarial passes, both clean)
Two independent fresh-context reviews ran on the phase-8 diff; **neither found a blocker or bug**. Both verified the default-path byte-identity against the actual mortie nested-id tiling (default `shard_object_block` reduces exactly to `block_index(shard_key)`), the split tiling/bounds (no off-by-one, no overlap), the rectilinear/`hasattr` fallback (one whole-tile object, unchanged), validation bounds, and that empty inner chunks create no object/`set_block_selection`. Folded the two diff-scoped doc nits in `39002f7` (the `shard_order` range wording said "strictly between" vs the validated closed range; a stale `_spec` comment still said the outer chunk was `cells_per_shard`).

**One low-risk gap flagged (not folded):** the split path (`shard_order > parent_order`) is exercised for **scalar** fields only — companion (`resolution: chunk`), vector-trailing-dim, and ragged/CSR byte-identity *under split* aren't yet pinned by a test. The **default** path for those field kinds stays covered (transitively by the existing sharded tests + the byte-identity equivalence), and those writes are object-independent (companions/ragged are written per inner chunk, outside the per-object loop), so risk is low — but a split test over a vector+companion config would close it. Happy to add it next run, or as a small follow-up.

### Interaction with #124 (stream-and-free) + merge order
Mirroring the note I left on #124: the two are **complementary, not conflicting**, and **no design change is needed in either**. #124 streams compute→write→free per inner chunk on the **non-sharded** path and deliberately leaves the sharded path accumulating one object (because `set_block_selection` is shard-granular); phase 8 shrinks that object so the **sharded** path can *also* bound memory by accumulating→writing→freeing per sharding-object. Phase 8 composes on top of #124's seam rather than touching it.

**Recommended merge order: #124 first** (it's feature-complete, byte-identical, small, only awaiting review), then this PR — phase 8 is the natural consumer of that seam. Both are individually conflict-free against `main` today; whichever lands second gets a quick re-resolve.

All phases complete, CI green, both self-reviews folded. Marking `waiting` — the ball is on your side for review/merge and the merge-order call. Staying a draft.

---

espg:
> One low-risk gap flagged (not folded): the split path (shard_order > parent_order) is exercised for scalar fields only — companion (resolution: chunk), vector-trailing-dim, and ragged/CSR byte-identity under split aren't yet pinned by a test. The default path for those field kinds stays covered (transitively by the existing sharded tests + the byte-identity equivalence), and those writes are object-independent (companions/ragged are written per inner chunk, outside the per-object loop), so risk is low — but a split test over a vector+companion config would close it. Happy to add it next run, or as a small follow-up.

yes, add it please

espg:
🤖 *from Claude*

Added the split-path coverage you asked for — commit `a43bfd5` (`fold review: split test for vector/companion/ragged fields`). New `TestShardOrderObjectSplit::test_split_reconstructs_vector_companion_and_ragged` drives `shard_order=5` (4 objects/shard) against the default single-object write for a config carrying **every field kind**, extending the default scalar config with:
- a **vector trailing-dim** field (`h_edges`, `trailing_shape=2`) — dense *and* sharded, so it actually rides the per-object slab accumulation (the path the split changes); plus the default `morton` coordinate, also sharded;
- a **`resolution: chunk` companion** (`h_chunk_base`, anchored on a `chunk_precompute` value) — dense, one block per chunk;
- a **ragged/CSR** field (`h_ragged`) — per-cell CSR.

Assertions: every dense array (scalar `h_mean`, the vector `h_edges`, the companion `h_chunk_base`, and the `morton` coord) reconstructs **value-for-value** across the split via read-back; and the ragged/CSR store keys are **byte-for-byte identical** between default and split (they're written per inner chunk, outside the per-object loop, so `shard_order` can't perturb them). This closes the gap I flagged — the split path is now proven for vector trailing-dim, companion, and ragged, not just scalars.

`tests/test_processing.py::TestShardOrderObjectSplit` → 5 passed; `ruff check`/`format` clean. Re-applying `waiting` — back to you for review/merge. Still a draft.

---

Issue #138, opened 2026-06-30
Benchmark `tdigest_healpix_o10_inner` OOMs at the 2 GB cap — warm-container carryover / run-ordering, not a code regression


espg:
🤖 *from Claude*

## Summary

`tdigest_healpix_o10_inner` has failed with a Lambda `Runtime.OutOfMemory` on every run since the forward benchmark matrix (#133) landed, while the **identical workload** ran healthily at ~900 MB right up until the split. The inner code path did **not** change — this is **not** an agg/template/write regression. The most likely cause is **warm-container memory carryover** exposed by the new target *run order*, plus some independent infra flakiness in the same run. Filing for discussion on how to harden the benchmark harness (and to confirm transient-vs-reproducible with a re-run first).

## Symptom (from the `c61cbe4` "Run benchmark" step + `series.parquet` on the `benchmarks` branch)

```
shard 5347405532456026122: Lambda OOM {"errorType":"Runtime.OutOfMemory", ...}
[tdigest_healpix_o10_inner]   obs=0      max_memory_mb=None      ← OOM (2 GB arm64 cap)
[tdigest_healpix_o10_sharded] obs=939217 max_memory_mb=2013.23  ← same shard/data, 98% of cap
```

The same run also threw two `Read timeout` errors on the o9 shard (both o9 targets → `obs=0`), i.e. independent infra noise in that run.

Inner-path memory history for the **same 939,217-obs shard** (carrier = arro3 throughout):

| commit | time (UTC) | target | max_memory_mb | result |
|---|---|---|---|---|
| `29b392e` (#130 carrier flip) | 06-30 04:45 | `tdigest_healpix_o10` | 891.5 | ✅ |
| `c79f5db` | 06-30 16:42 | `tdigest_healpix_o10` | 899.2 | ✅ |
| `e21a42f` | 06-30 19:28 | `tdigest_healpix_o10` | 922.8 | ✅ |
| `c61cbe4` / `349708b` | 06-30 20:20 / 21:02 | `tdigest_healpix_o10_inner` | **OOM (>2048)** | ❌ |

## It is the same workload as the old low-memory target

[...]

Issue #137, opened 2026-06-30
Lambda Benchmark Earthdata login intermittently fails with `[Errno 101] Network is unreachable` (IPv6 no-route) — force IPv4


espg:
🤖 *from Claude*

## Problem

The **Lambda Benchmark** workflow (`.github/workflows/lambda-benchmark.yml` → `.github/scripts/run_benchmark.py` → `agg(... backend="lambda")` → `runner._run_lambda` → `auth.get_nsidc_s3_credentials` → `earthaccess.login()`) **intermittently — and lately consistently — fails** at the Earthdata login step when it tries to reach `urs.earthdata.nasa.gov:443`:

```
OSError: [Errno 101] Network is unreachable
  ... HTTPSConnection(host='urs.earthdata.nasa.gov', port=443):
      Failed to establish a new connection: [Errno 101] Network is unreachable
  ... requests.exceptions.ConnectionError: HTTPSConnectionPool(host='urs.earthdata.nasa.gov', port=443):
      Max retries exceeded with url: /profile
  File ".github/scripts/run_benchmark.py", line 137, in run_target -> agg(...)
  File "src/zagg/runner.py", in _run_lambda -> get_nsidc_s3_credentials()
  File "src/zagg/auth.py", line 65, in get_nsidc_s3_credentials -> earthaccess.login()
```

## Diagnosis — it's an IPv6 no-route, not NASA and not a code bug

`[Errno 101] Network is unreachable` is an **origin-side routing failure**: the runner's network stack determined it had no route to the destination and gave up *before* any packet reached NASA. That is categorically different from a NASA outage (which surfaces as a connection *timeout*, *connection refused*, or an HTTP *5xx*/auth error) — and `urs.earthdata.nasa.gov` is demonstrably reachable, since other Lambda Benchmark runs on `main` minutes apart log in fine.

The signature points squarely at **IPv6 with no IPv6 route on the GitHub runner**: the host resolves to an AAAA (IPv6) record, the runner has no usable IPv6 path, and the connect attempt fails immediately with "Network is unreachable" instead of falling back to IPv4. This is a well-known class of CI flake against dual-stack hosts.

## Proposed fix — force IPv4 for the Earthdata login

Pin the Earthdata login (and ideally the benchmark's outbound) to IPv4 so it never attempts the unroutable AAAA address.

[...]

PR #140, opened 2026-07-01
Reclaim worker RSS on warm Lambda containers (gc + malloc_trim) (#139)

### LINKED ISSUE #139: Reclaim worker RSS on warm Lambda containers (gc + malloc_trim) to prevent warm-carryover OOM
🤖 *from Claude*

## Reclaim worker RSS on warm Lambda containers (gc + `malloc_trim`) to prevent warm-carryover OOM

Tracking issue for the fix surfaced by the #138 investigation.

### Finding
Lambda reuses warm containers; freed memory (pymalloc arenas / glibc heap) is **not returned to the OS**, so a subsequent invocation on a warm container starts near the *previous* invocation's RSS high-water. `Max Memory Used` is per-container-lifetime, so a workload that peaks ~900 MB cold can report/OOM near 2 GB warm. Measured directly: the same o11 inner shard read **774 MB (cold)** but **1322 MB warm** (reusing a prior invoke's container); the CI benchmark's `tdigest_healpix_o10_inner` OOM is this class (it followed the 2013 MB warm `sharded` target).

Production runs one config across a fan-out, so the first wave is cold, but shards beyond the concurrency budget **reuse warm containers** — so this is a real production robustness gap, not just a benchmark artifact.

### Fix
At the end of each handler invocation (after write), release big buffers and hand freed heap back to the OS:
```python
del <pooled arrays / carriers>
gc.collect()
ctypes.CDLL("libc.so.6").malloc_trim(0)   # glibc (AL2023 arm64): returns freed heap -> RSS drops
```
`gc.collect()` alone doesn't drop RSS; `malloc_trim(0)` does on the AL2023/glibc runtime. This makes the next invocation on a warm container start near baseline.

### Plan
- [ ] Prototype the reclaim in the worker handler (`deployment/aws/lambda_handler.py` / `process_shard` teardown).
- [ ] Prove the RSS reclaim deterministically on Linux (podman): run a heavy workload, `del`+`gc`+`malloc_trim`, show RSS drops.
- [ ] Deploy the patched worker to `process-shard-test` and re-run the o10 `sharded → inner` warm sequence; show `inner` no longer OOMs.
- [ ] Confirm no correctness/perf regression (`malloc_trim` is cheap; called once per invoke).

Assigned to a sub-agent; it will update this issue with results.

espg:
Closes #139

## What

Reclaim the worker's RSS at Lambda handler teardown so a **warm** container doesn't carry the previous invocation's memory into the next one. Adds a guarded `_reclaim_memory()` (`gc.collect()` + glibc `malloc_trim(0)`) called once per invocation in a `finally` on `_handle_process`, after `del`-ing the large output buffers.

## Why (from the #138 / #137 investigation)

Lambda reuses warm containers, and neither CPython's pymalloc arenas nor glibc's heap return freed memory to the OS on their own — so a subsequent invocation on a warm container starts near the *previous* invocation's RSS high-water (`Max Memory Used` is per-container-lifetime). Measured directly on prod `process-shard`: the **same** o11 shard read **774 MB cold vs 1322 MB** when it reused a warm container, and the CI benchmark's `tdigest_healpix_o10_inner` OOMs because it follows the ~2013 MB warm `sharded` target on the same container. Production fan-outs beyond the concurrency budget reuse warm containers too, so this is a real robustness gap — not just a benchmark artifact.

## The change

```python
def _reclaim_memory() -> None:
    gc.collect()                                  # frees Python objects (not RSS by itself)
    try:
        ctypes.CDLL("libc.so.6").malloc_trim(0)   # glibc: returns freed heap to the OS -> RSS drops
    except (OSError, AttributeError):
        ...  # no-op off glibc (macOS/dev), never raises
```
Called in a `finally`, after `del _df_out, chunk_results` (the response body is just the small metadata), so the freed carrier heap is available to trim. Guarded so it's a no-op on non-glibc dev machines.

## How tested

- **podman (arm64 / glibc = matches the AL2023 runtime):** baseline 27.5 MB → alloc 467 MB → after `del`+`gc.collect()` **156 MB** (gc alone strands ~128 MB) → after `malloc_trim(0)` **29.7 MB** (baseline). And a warm-reuse sim: carried residual **189 MB → 29 MB**. So `gc` alone does not drop RSS; `malloc_trim` does.
- **Tests:** 3 new in `test_lambda_handler.py` — the helper never raises on a non-glibc host; the reclaim runs exactly once on the success path and once on the exception path.
- **No regression:** full suite **1059 passed, 9 skipped** (only a pre-existing macOS-only build test deselected); `ruff check` / `ruff format --check` clean. `malloc_trim` is O(heap), once per invocation — behavior-neutral (memory only).

## Real-Lambda confirmation — note on why the metric can't show it

The patched `lambda_handler.py` **was** deployed in-place to `process-shard-test` and the o10 `sharded → inner` warm sequence re-run. It came back **895 → 1603 MB**, vs the unpatched prod control **878 → 1569 MB** — i.e. the fix looks *absent*. It isn't: `max_memory_mb` (= `ru_maxrss`) is a **monotonic per-process high-water mark**, and `malloc_trim(0)` lowers *current* RSS, not the recorded peak. On a warm container the peak can't drop, so this metric **structurally cannot** reflect the fix. (Details: PR comment above / #138.)

The valid evidence is the **current-RSS** podman proof under "How tested" (467 → 156 → **29.7 MB**). `malloc_trim`'s payoff is resetting *current* RSS between invocations so a warm container doesn't accumulate toward an OOM in production — confirming that on real Lambda would mean reproducing the near-cap accumulation scenario, not a `max_memory_mb` delta. Recommend merging on the podman proof + review; the change is correct and behavior-neutral.

## Questions for review
- `malloc_trim(0)` is glibc-specific; the guard makes it a no-op elsewhere. AL2023 arm64 (the deployed runtime) is glibc, so it's active in prod. OK as-is, or prefer a `platform`/`sys.platform` gate too?

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -55,6 +55,8 @@ output bucket (e.g. cross-account JupyterHub orchestrators) can run the
 full pipeline using only lambda:InvokeFunction.
 """
 
+import ctypes
+import gc
 import json
 import logging
 import os
@@ -89,6 +91,31 @@ def _max_memory_mb() -> float:
     Memory Used" closely.
     """
     return resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024.0
+
+
+def _reclaim_memory() -> None:
+    """Return freed heap to the OS at invocation teardown (issue #139).
+
+    Lambda reuses warm containers, and neither CPython's pymalloc arenas nor
+    glibc's malloc heap hand freed memory back to the OS on their own, so a
+    subsequent invocation on a warm container starts near the *previous*
+    invocation's RSS high-water and can OOM (``Max Memory Used`` is
+    per-container-lifetime). A ``gc.collect()`` reclaims unreachable Python
+    objects but does NOT drop RSS by itself; glibc's ``malloc_trim(0)`` releases
+    the freed top-of-heap back to the OS, so the next invocation on the same
+    container starts near baseline.
+
+    Called once per invocation (O(heap) -- negligible next to read/aggregate/
+    write) and behavior-neutral: it only frees memory the invocation is done
+    with. Guarded so it is a no-op off glibc (macOS/dev has no ``libc.so.6``,
+    non-glibc libcs may lack ``malloc_trim``) -- it never raises.
+    """
+    gc.collect()
+    try:
+        ctypes.CDLL("libc.so.6").malloc_trim(0)
+    except (OSError, AttributeError):
+        # No glibc (no libc.so.6) or no malloc_trim symbol -- nothing to trim.
+        logger.debug("malloc_trim unavailable; skipping heap reclaim", exc_info=True)
 
 
 def _output_store_kwargs(event: Dict[str, Any]) -> Dict[str, Any]:
@@ -429,6 +456,12 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         logger.info("Lambda invocation completed successfully")
         logger.info("=" * 70)
 
+        # Drop the invocation's large output buffers before the teardown reclaim
+        # (issue #139): the response body is just ``metadata`` (small), so freeing
+        # the shard's carrier / accumulated K chunk carriers here lets the
+        # ``_reclaim_memory`` call in ``finally`` return their heap to the OS.
+        del _df_out, chunk_results
+
         return {
             "statusCode": 200 if not metadata.get("error") else 500,
             "body": json.dumps(metadata),
@@ -448,3 +481,9 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
                 }
             ),
         }
+    finally:
+        # Warm-container memory reclaim (issue #139): once per invocation, after
+        # the write completes and the buffers above are dropped, hand freed heap
+        # back to the OS so the next invoke on this warm container starts near
+        # baseline instead of the prior invocation's RSS high-water.
+        _reclaim_memory()
--- a/tests/test_lambda_handler.py
+++ b/tests/test_lambda_handler.py
@@ -202,6 +202,72 @@ class TestProcessEventDispatch:
         assert body["max_memory_mb"] > 0.0
 
 
+class TestReclaimMemory:
+    """Issue #139: each invocation hands freed heap back to the OS at teardown so
+    a warm-container reuse starts near baseline instead of the prior invocation's
+    RSS high-water. The reclaim runs once per invocation in a ``finally`` (both
+    the success and unhandled-exception paths)."""
+
+    def test_reclaim_memory_never_raises(self, handler_mod):
+        # The helper is guarded: on a non-glibc dev host (no libc.so.6) or a libc
+        # lacking malloc_trim it must be a silent no-op, never an exception.
+        handler_mod._reclaim_memory()  # would raise on the dev platform if unguarded
+
+    def test_success_path_calls_reclaim_once(self, handler_mod, monkeypatch):
+        import zagg.grids as grids
+        import zagg.processing as processing
+
+        calls = {"n": 0}
+        monkeypatch.setattr(
+            handler_mod, "_reclaim_memory", lambda: calls.__setitem__("n", calls["n"] + 1)
+        )
+
+        def fake_process_shard(grid, shard_key, granule_urls, **kwargs):
+            meta = {
+                "shard_key": shard_key,
+                "cells_with_data": 0,
+                "total_obs": 0,
+                "granule_count": len(granule_urls),
+                "files_processed": 0,
+                "duration_s": 0.0,
+                "error": None,
+            }
+            return pd.DataFrame(), meta
+
+        monkeypatch.setattr(processing, "process_shard", fake_process_shard)
+        monkeypatch.setattr(grids, "from_config", lambda *a, **k: MagicMock())
+
+        event = _base_event(_healpix_config_dict())
+        event["child_order"] = 12
+        resp = handler_mod._handle_process(event, _context())
+        assert resp["statusCode"] == 200
+        assert calls["n"] == 1
+
+    def test_reclaim_runs_on_exception_path(self, handler_mod, monkeypatch):
+        # An unhandled error mid-invocation must still trigger the teardown reclaim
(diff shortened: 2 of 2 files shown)


espg:
🤖 *from Claude (review)*

Fresh-context adversarial review of the per-invocation memory reclaim. **Verdict: clean — no blocking findings, no real bug.** I tried hard on the two things you flagged (the `del` unbound-name trap and the never-raises guarantee) and both hold up.

### The `del _df_out, chunk_results` is safe (no UnboundLocalError)
The only way to reach `deployment/aws/lambda_handler.py:463` is linear flow through the two assignments that precede it — `chunk_results` at `:388` and `_df_out` at `:389` — with no branch or early `return` in between. The early `return`s for missing params (`:259`, `:268`, `:290`) all sit *before* `:388`, so they never reach the `del`. If `process_shard` at `:389` raises, control goes to the `except`/`finally`, not the `del`. So on every path that executes `:463`, both names are bound. Confirmed empirically: `test_success_path_calls_reclaim_once` actually runs through `:463` (the fake `process_shard` returns `(pd.DataFrame(), meta)`, MagicMock grid's `__int__`→1, `chunk_results=[]`) and returns 200 — i.e. the `del` line is genuinely exercised, not stubbed out.

### The never-raises guarantee holds
`_reclaim_memory` (`:96`) references no invocation locals — only module-scope `gc`, `ctypes`, and `logger` (defined at `:81`, so the `debug` line is fine). `gc.collect()` won't raise; `ctypes.CDLL("libc.so.6")` raises `OSError` on macOS/musl (no such lib) and Windows (`FileNotFoundError ⊂ OSError`); a missing `malloc_trim` symbol raises `AttributeError` on attribute lookup; the `malloc_trim(0)` call returns an int (1/0) that's harmlessly ignored — a non-zero/"nothing trimmed" return doesn't matter. Both realistic failure modes are caught by `except (OSError, AttributeError)`. `test_reclaim_memory_never_raises` exercises the real helper on this non-glibc host and passes.

### Behavior neutrality confirmed
Nothing after `:463` touches `_df_out`/`chunk_results` — the response (`:465-468`) uses only `metadata`, which is not deleted. The `del` is in the `try`, only `_reclaim_memory()` is in the `finally` (`:489`); the `finally` is a bare call (no `return`), so it doesn't swallow or alter the try/except return values. `profile`/`handoff`/sharded paths and the metadata body are untouched.

### Cost/latency
`malloc_trim(0)` is O(heap) once per invocation — negligible next to read/aggregate/write, and trivial on the tiny early-return / no-data invocations. Note `setup`/`finalize` modes don't get the reclaim (separate handlers), which is fine: they're lightweight and a preceding `process` invocation already trimmed at its own teardown.

### Tests / lint
- `pytest tests/test_lambda_handler.py -q` → **31 passed**.
- `ruff check deployment/aws/lambda_handler.py tests/test_lambda_handler.py` → **clean**.
- Full suite → **1059 passed, 9 skipped, 1 failed**. The one failure is `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, a **pre-existing, unrelated** macOS build/packaging test (it builds the Lambda image; the PR body notes it's normally deselected on macOS) — not touched by this diff.

### Nits (non-blocking, optional)
1. **Coverage gap (nit):** the tests assert the `finally` calls `_reclaim_memory` once on both paths and that the real helper never raises, but none asserts `gc.collect()` / `malloc_trim(0)` are actually invoked. That's the right level given the effect is memory-only and unobservable in-process; the podman measurements in the PR body cover the real RSS drop. Fine as-is.
2. **Guard breadth (nit):** `except (OSError, AttributeError)` is precise and sufficient for glibc/musl/macOS/Windows. If you wanted belt-and-suspenders against some exotic ctypes edge, `except Exception` in a teardown-only no-op helper would be defensible — but the current precise guard is the better-documented choice. No change needed.

Re your open question: the guard already makes this a no-op off glibc, so an extra `sys.platform` gate would be redundant — I'd leave it as-is.

espg:
this looks good. One question-- is it worth calling:

```python
    try:
        ctypes.CDLL("libc.so.6").malloc_trim(0)   # glibc: returns freed heap to the OS -> RSS drops
    except (OSError, AttributeError):
```

at worker start *and* finish? The use case I'm thinking of here is what happens if we have a worker error out, and not reach the malloc_trim command? Could that worker still be 'hot' and picked up by another shard if we're at the concurrency limit? Or is this out of scope (e.g., maybe errored worker lambdas are already not available as hot containers). Thoughts on this belt and suspenders framing?

espg:
🤖 *from Claude*

## Correction to the "real-Lambda confirmation" plan in the PR body

The PR body says I'd confirm the fix on real Lambda via an o10 `sharded → inner` warm sequence. **That test cannot confirm this fix**, and here's why — it's worth recording so nobody re-runs it expecting a signal.

`max_memory_mb` is `resource.getrusage(RUSAGE_SELF).ru_maxrss` — a **monotonic high-water mark over the whole process** (`lambda_handler.py:_max_memory_mb`, docstring says so). `malloc_trim(0)` lowers **current** RSS but **cannot lower the recorded peak**. On a warm (reused) container the process persists, so `ru_maxrss` is the max over *every* invocation it served and never decreases. Result: the warm-sequence numbers are identical with and without the fix —

| | sharded | inner (warm, same container) |
|---|---|---|
| prod `process-shard` (no fix) | 878 MB | **1569 MB** |
| `process-shard-test` (this patch, deployed v30) | 895 MB | **1603 MB** |

The fix is *invisible* to this metric — exactly as expected, not a failure of the patch.

**The valid evidence is the current-RSS podman proof already in the PR:** 467 MB → after `del`+`gc.collect()` 156 MB → after `malloc_trim(0)` **29.7 MB**. That's the mechanism. `malloc_trim`'s payoff is **resetting current RSS between invocations so a warm container doesn't accumulate toward an OOM in production** — which `ru_maxrss` / CloudWatch "Max Memory Used" (also a lifetime high-water) don't reflect until an actual OOM is *avoided*.

**So:** confirming the OOM-prevention on real Lambda would require reproducing the near-cap accumulation scenario (a prior invoke that leaves ~2 GB current RSS, then a following one), not a `max_memory_mb` delta — a possible follow-up, but not a merge gate. Recommend merging on the podman proof + the clean review; the change is correct and behavior-neutral. I'll update the PR body to replace the "confirmation in progress" line with this.

PR #144, opened 2026-07-01
Set glibc allocator tunables on the process Lambda (issue #143)

### LINKED ISSUE #143: Warm-container OOM: Arrow memory pool residual not reclaimed by malloc_trim (#140)
🤖 *from Claude*

## Symptom

After #140 merged, the **Lambda Benchmark still OOMs** — [run 28502909333](https://github.com/englacial/zagg/actions/runs/28502909333/job/84484961598) benchmarked the merged code on `process-shard-test`; 4 of 6 targets came back OOM/failed:

```
tdigest_healpix_o11_sharded   obs=200676   max_memory_mb=808     ok
tdigest_healpix_o11_inner     obs=200676   max_memory_mb=1354    ok   (+546)
tdigest_healpix_o10_sharded   obs=0        max_memory_mb=None    FAIL
tdigest_healpix_o10_inner     obs=0        max_memory_mb=None    FAIL
tdigest_healpix_o9_sharded    obs=0        max_memory_mb=None    FAIL
tdigest_healpix_o9_inner      obs=0        max_memory_mb=None    FAIL   <- Runtime.OutOfMemory (req 44196cba)
```

The tell is `o11_sharded 808 → o11_inner 1354` = **+546 MB on the second invoke of the same warm container** (o11's workload is only ~200 k obs), accumulating to OOM by o9. CloudWatch confirms one container lineage `910 → 1474 → 2041 → OOM` across **normally-completed** invokes — so #140's teardown trim ran and released nothing.

## Root cause — CONFIRMED (probe: verification comment below)

Probe on **podman Linux/aarch64** with the exact Lambda-pinned worker deps (`arro3-core==0.8.1`, `numpy==2.2.6`, no pyarrow), reproducing zagg's real `_concat_and_group` arrow branch:

- The worker carrier is **`arro3-core`** (Rust/`arrow-rs`), and its `_core.abi3.so` **links only `libc.so.6`** — no jemalloc/mimalloc. **There is no separate Arrow memory pool**; the residual is plain **numpy memory on the glibc heap**.
- `combine_chunks().to_numpy()` is a **copy** (`owndata=True`), not a zero-copy view — the arro3 buffers free at `del table, batches`; the survivor is the numpy copy.
- The accumulation is **glibc's default arena / trim-threshold behavior**: freed memory is retained in per-thread arenas and not returned to the OS between warm invokes. Reproduced deterministically: **+320 MB on the 2nd warm invoke** (synthetic scale; matches the benchmark's 808→1354 pattern).
- **#140's `malloc_trim(0)` is UNRELIABLE** for this (doesn't consistently return the retained arenas). The originally-proposed `pyarrow.default_memory_pool().release_unused()` is a **dead no-op** — pyarrow isn't importable on the worker, so it raises and is swallowed.

So of the three candidates, it's **glibc arena retention** — not a Rust caching allocator, not zero-copy.

## Fix — CONFIRMED

Set glibc tunables as **Lambda Environment variables** (they must be present at libc init — a runtime `mallopt()`/`malloc_trim` from Python does **not** take effect):

- **`MALLOC_ARENA_MAX=2`** (primary — limits arenas; flattens warm accumulation to **+0.0 MB**), and/or
- `MALLOC_TRIM_THRESHOLD_=0` / `MALLOC_MMAP_THRESHOLD_=131072` (each independently flattens it too).

Location: `deployment/aws/template.yaml` → the process function's `Environment` (infra — **this issue authorizes that file by name**). This **supersedes #140's `_reclaim_memory` malloc_trim path** (unreliable) and the pyarrow branch should not be added. It achieves @espg's "clean baseline per invoke" intent at the allocator/init layer instead of a per-invoke Python call.

## Secondary: the benchmark OOMs silently

A target OOM is recorded as `obs=0, max_memory_mb=None`, pushed to the series as a data row — the CI **job stays green**. A regression that took out 4/6 targets didn't fail CI. Worth surfacing (fail the job, or flag zero-obs targets).

## Refs
- Supersedes #140's approach (malloc_trim unreliable; env-var tunable is the reliable fix).
- Root-caused in #138; carrier is `arro3-core` per #130.

espg:
Closes #143.

## What / approach

Warm `process-shard` containers accumulate retained glibc-arena memory across invokes and OOM by the o9 target (the `808 → 1354` climb in #143). The [root-cause probe](https://github.com/englacial/zagg/issues/143#issuecomment-4852348594) showed this is **glibc arena retention** of freed numpy blocks — not an Arrow pool (the carrier is `arro3-core`, plain glibc `malloc`), and #140's `malloc_trim(0)` recovers only ~23 MB of it. The reliable, carrier-agnostic fix is glibc allocator tunables set **as Lambda `Environment` variables** (they must be present at libc init — a runtime `mallopt`/`malloc_trim` from Python does not take effect). In the probe `MALLOC_ARENA_MAX=2` flattened warm accumulation from +320 MB to **+0.0 MB**.

Decisions taken per @espg's answers on the [plan](https://github.com/englacial/zagg/issues/143#issuecomment-4859068444): tunables `MALLOC_ARENA_MAX=2` + `MALLOC_TRIM_THRESHOLD_=0`; keep the harmless `gc.collect()` in `_reclaim_memory` for now; the benchmark silent-OOM hardening splits to its own `small-fix` (not this PR).

Issue #143 names `deployment/aws/template.yaml`, so editing that infra file is authorized here. **No live-AWS run** — see "Pre-merge Lambda proof" below.

## Phases

- [x] **Phase 1 — allocator env vars on the process function.** Two new CloudFormation Parameters (`MallocArenaMax` default `"2"`, `MallocTrimThreshold` default `"0"`) wired into a new `ProcessFn.Environment.Variables` block (`MALLOC_ARENA_MAX` / `MALLOC_TRIM_THRESHOLD_`), matching the template's existing `MemorySize`/`Timeout` parameter style so the values stay tunable without a template edit. A template-shape test (`tests/test_lambda_build.py::TestTemplateEnvironment`) asserts `ProcessFn` carries both env vars and the parameters carry the expected defaults, so a future template edit can't silently drop them.
- [x] **Phase 2 — keep the harmless `gc.collect()`, document supersession.** Per @espg's lean, `_reclaim_memory` is retained (the `gc.collect()` still reclaims unreachable Python objects); its docstring now records that the env-var tunables are the *primary/reliable* fix for the glibc-arena retention and that `malloc_trim(0)` is an unreliable secondary. No behavioral change (docstring only).

## Pre-merge Lambda proof — answering @espg's `/benchmark`-label question

> can you do this via /benchmark label on the PR, or does it need to be manual run?

**It needs a manual env-var update — the `/benchmark` PR command cannot validate this fix.** The on-demand benchmark (`.github/workflows/lambda-benchmark-command.yml`) overlays only the PR's `src/zagg` onto the base harness and runs against the **already-deployed** `process-shard-test` function (`vars.BENCHMARK_FUNCTION_NAME`); it never applies a CloudFormation update, so the new `Environment` block in `template.yaml` is not in effect during that run. A `/benchmark` here would therefore measure the *old* function config (no `MALLOC_ARENA_MAX`) and show no change.

To confirm end-to-end, the tunables have to be present on the running function. The reversible way (no full CFN deploy) is a Lambda **env-var update** on `process-shard-test` — `MALLOC_ARENA_MAX=2` + `MALLOC_TRIM_THRESHOLD_=0` — then re-run the 6 benchmark targets sequentially and check o10/o9 pass instead of OOM. That is a live-AWS action, so per §1 I'm **not** doing it here — leaving it for @espg (or confirmation that the deployed test function should get the env vars via the normal deploy path).

## How tested

- `uv run pytest tests/test_lambda_build.py tests/test_lambda_handler.py -q` → **53 passed** (incl. the 2 new `TestTemplateEnvironment` cases).
- `uv run ruff check --select=E,F,W,I --ignore=E501 deployment tests/test_lambda_build.py` → clean. Touched Python files are `ruff format`-clean.
- No behavioral Python change (Phase 1 is template + a shape test; Phase 2 is a docstring), so no runtime test beyond the template assertion.

## Questions for review

- **None blocking.** Confirm whether the deployed `process-shard-test` should receive the env vars manually for the pre-merge proof, or whether you'll take that as a normal deploy after merge.

## Follow-up (out of scope, per @espg)

Phase 3 from the plan — the benchmark recording an OOM'd target as `obs=0, max_memory_mb=None` while the CI job stays green — is split to its own `small-fix` so this memory fix lands unblocked.

---
_Generated by [Claude Code](https://claude.ai/code/session_017wNcBZd6KvDfe7mSeQfo49)_

### DIFF
--- a/deployment/aws/template.yaml
+++ b/deployment/aws/template.yaml
@@ -80,6 +80,24 @@ Parameters:
     Default: 720
     Description: Function timeout in seconds.
 
+  MallocArenaMax:
+    Type: String
+    Default: "2"
+    Description: >-
+      glibc MALLOC_ARENA_MAX for the process function (issue #143). Caps the
+      number of per-thread malloc arenas so freed numpy/glibc heap is returned
+      to the OS between warm invocations instead of accumulating toward the
+      memory cap. Must be set as an env var (it takes effect at libc init; a
+      runtime mallopt/malloc_trim from Python does not). Set empty to unset.
+
+  MallocTrimThreshold:
+    Type: String
+    Default: "0"
+    Description: >-
+      glibc MALLOC_TRIM_THRESHOLD_ for the process function (issue #143). 0
+      trims freed top-of-heap back to the OS immediately, independently
+      flattening warm-container memory growth. Set empty to unset.
+
 Conditions:
   ShouldCreateBucket: !Equals [!Ref CreateOutputBucket, "true"]
   ShouldCreateRole: !Equals [!Ref CreateExecutionRole, "true"]
@@ -145,6 +163,10 @@ Resources:
       Timeout: !Ref Timeout
       Role: !If [ShouldCreateRole, !GetAtt ExecutionRole.Arn, !Ref ExecutionRoleArn]
       Layers: [!Ref DepsLayer]
+      Environment:
+        Variables:
+          MALLOC_ARENA_MAX: !Ref MallocArenaMax
+          MALLOC_TRIM_THRESHOLD_: !Ref MallocTrimThreshold
       Code:
         S3Bucket: !Ref ArtifactBucket
         S3Key: !Ref FunctionS3Key
--- a/deployment/aws/template.yaml
+++ b/deployment/aws/template.yaml
@@ -88,7 +88,7 @@ Parameters:
       number of per-thread malloc arenas so freed numpy/glibc heap is returned
       to the OS between warm invocations instead of accumulating toward the
       memory cap. Must be set as an env var (it takes effect at libc init; a
-      runtime mallopt/malloc_trim from Python does not). Set empty to unset.
+      runtime mallopt/malloc_trim from Python does not). Set empty to fall back to the glibc default (the key is still emitted, just empty).
 
   MallocTrimThreshold:
     Type: String
@@ -96,7 +96,7 @@ Parameters:
     Description: >-
       glibc MALLOC_TRIM_THRESHOLD_ for the process function (issue #143). 0
       trims freed top-of-heap back to the OS immediately, independently
-      flattening warm-container memory growth. Set empty to unset.
+      flattening warm-container memory growth. Set empty to fall back to the glibc default (the key is still emitted, just empty).
 
 Conditions:
   ShouldCreateBucket: !Equals [!Ref CreateOutputBucket, "true"]
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -94,16 +94,19 @@ def _max_memory_mb() -> float:
 
 
 def _reclaim_memory() -> None:
-    """Return freed heap to the OS at invocation teardown (issue #139).
+    """Reclaim Python objects at invocation teardown (issues #139, #143).
 
-    Lambda reuses warm containers, and neither CPython's pymalloc arenas nor
-    glibc's malloc heap hand freed memory back to the OS on their own, so a
-    subsequent invocation on a warm container starts near the *previous*
-    invocation's RSS high-water and can OOM (``Max Memory Used`` is
-    per-container-lifetime). A ``gc.collect()`` reclaims unreachable Python
-    objects but does NOT drop RSS by itself; glibc's ``malloc_trim(0)`` releases
-    the freed top-of-heap back to the OS, so the next invocation on the same
-    container starts near baseline.
+    Lambda reuses warm containers, so a subsequent invocation on a warm
+    container starts near the *previous* invocation's RSS and can OOM
+    (``Max Memory Used`` is per-container-lifetime). The reliable fix for the
+    glibc-arena retention that drives this is the allocator env vars set on the
+    function itself (``MALLOC_ARENA_MAX``/``MALLOC_TRIM_THRESHOLD_`` in
+    ``template.yaml``, issue #143) -- those take effect at libc init and flatten
+    warm-container growth to ~0. The ``malloc_trim(0)`` below is retained as a
+    harmless secondary: it only trims the top of the main arena, so on its own
+    it does *not* reliably return the retained secondary-arena numpy blocks
+    (hence the env-var fix supersedes it as the primary mechanism), but the
+    ``gc.collect()`` still reclaims unreachable Python objects.
 
     Called once per invocation (O(heap) -- negligible next to read/aggregate/
     write) and behavior-neutral: it only frees memory the invocation is done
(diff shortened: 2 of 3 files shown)


deployment/aws/template.yaml
```diff
@@ -80,6 +80,24 @@ Parameters:
...
 
+  MallocArenaMax:
+    Type: String
+    Default: "2"
+    Description: >-
+      glibc MALLOC_ARENA_MAX for the process function (issue #143). Caps the
+      number of per-thread malloc arenas so freed numpy/glibc heap is returned
+      to the OS between warm invocations instead of accumulating toward the
+      memory cap. Must be set as an env var (it takes effect at libc init; a
+      runtime mallopt/malloc_trim from Python does not). Set empty to unset.
```

"Set empty to unset" is technically inaccurate for a CloudFormation Lambda env var. With `Default: ""`, an empty string is emitted as a *literal* `Environment.Variables.MALLOC_ARENA_MAX = ""` — the key is still present, just empty. glibc happens to ignore an empty/invalid value and fall back to its default, so the net effect *resembles* "unset", but the variable is not actually omitted (there's no `AWS::NoValue`/`!If` to drop the key). Someone reading this and expecting a truly absent env var (e.g. for a clean A/B RSS comparison) would be surprised. Suggest rewording to "Set empty to fall back to the glibc default." (same on the `MallocTrimThreshold` description at line 99). Non-blocking.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the phase 1+2 diff. Verdict: **clean, no blocking findings.** Env var names are exactly right (`MALLOC_ARENA_MAX` no trailing underscore, `MALLOC_TRIM_THRESHOLD_` with), the `Environment.Variables` block is correctly placed under `ProcessFn.Properties`, both `!Ref`s point at the right parameters, and the template-shape test catches removal/rename/re-point regressions. Confirmed `_reclaim_memory`'s `gc.collect()`/`malloc_trim` code body is untouched (only the docstring changed) per the directive to keep it. One non-blocking prose accuracy nit (inline) on the "Set empty to unset" wording.

---

espg:
🤖 *from Claude*

Phases 1 + 2 landed (`292f787`, `1ba2b8d`); fresh-context self-review folded (`423d066` — reworded the "set empty to unset" parameter descriptions to accurately describe CFN empty-value semantics). Green locally: `pytest tests/test_lambda_build.py tests/test_lambda_handler.py` → 53 passed (incl. the 2 new template-shape cases); `ruff check` clean. Phase 3 (benchmark silent-OOM) is split to a `small-fix` per your call, so all in-scope phases here are complete. Marking `waiting` (kept `implement`) — the ball is with you for review/merge plus the one operational question: whether the deployed `process-shard-test` should get the env vars manually for the pre-merge proof (the `/benchmark` PR command can't validate them — it runs against the already-deployed function config, not the template change). Staying a **draft**.

---

Issue #141, opened 2026-07-01
Track per-invocation worker peak RSS, not container-lifetime ru_maxrss


espg:
🤖 *from Claude*

## Problem

`_max_memory_mb()` in `deployment/aws/lambda_handler.py` reports
`resource.getrusage(RUSAGE_SELF).ru_maxrss` — a **monotonic, per-process high-water mark**. On a **warm/reused** Lambda container the process persists across invocations, so this returns the **max over every invocation the container has ever served**, not the memory this worker run actually used.

The result is that `max_memory_mb` (which flows into the worker summary and the benchmark `series.parquet`) is **only trustworthy on a cold container**. On warm containers it silently over-reports, and it can never *decrease* — so it also can't reflect the `malloc_trim` reclaim just landed in #140.

This single fact was the root cause of the entire #138 investigation's confusion:
- the "OOM" of `tdigest_healpix_o10_inner` (real footprint ~335 MB, but it followed a ~2013 MB warm `sharded` run on the same container → inherited high-water),
- the phantom "arrow +544 MB regression" (a warm cell reading a prior invocation's peak),
- the inability to confirm #140 on real Lambda (the metric structurally cannot show a current-RSS drop).

## Goal

Report the **peak RSS of the current invocation**, independent of what prior invocations on the same warm container did — so warm-container benchmark numbers become trustworthy and we stop depending on forced cold starts (the #137 workaround) just to get a clean number.

## Why this is now feasible (synergy with #140)

`ru_maxrss` / `VmHWM` are both process-lifetime peaks — neither can be reset, so a delta doesn't help (a lighter invoke after a heavier one reads flat). The fix is to sample **current** RSS (`VmRSS` from `/proc/self/status`, or `resource`-free read) and track the max over the invocation. #140's teardown `malloc_trim(0)` already returns current RSS to ~baseline between invocations, so at handler entry a warm container starts near baseline and the sampled `VmRSS` max reflects **this** run.

## Proposed approach

- Add a lightweight peak-current-RSS tracker scoped to `_handle_process`: a daemon thread that polls `VmRSS` (parse `/proc/self/status`, no new deps) every ~50–100 ms and records the max, started at handler entry and stopped in the `finally`. Report that as `max_memory_mb` (keep `ru_maxrss` too, e.g. as `container_hwm_mb`, so the distinction is explicit).
[...]

PR #146, opened 2026-07-01
Fix Lambda Benchmark Earthdata IPv6 no-route + parallelize + per-invocation RSS (#137)

### LINKED ISSUE #137: Lambda Benchmark Earthdata login intermittently fails with `[Errno 101] Network is unreachable` (IPv6 no-route) — force IPv4

### LINKED ISSUE #145: Benchmark job stays green on a silent OOM (obs=0, max_memory_mb=None)
🤖 *from Claude*

Split from #143 per @espg's call ([here](https://github.com/englacial/zagg/issues/143#issuecomment-4859068444): "separate small fix, want phases 1 + 2 merged asap"), so the memory fix (PR #144) lands unblocked.

## Problem

When a benchmark target OOMs, it is recorded as `obs=0, max_memory_mb=None` and pushed to the series as a normal data row — the CI **job stays green**. A regression that took out 4 of 6 targets (the #143 warm-container OOM run, [run 28502909333](https://github.com/englacial/zagg/actions/runs/28502909333)) did **not** fail CI. A memory regression can merge unnoticed.

## Proposed fix

In `.github/scripts/run_benchmark.py`, after the target loop (`:220-244`), flag any **real** target (exclude `--dry-run`, which intentionally emits empty metrics per `:131-133`) whose `total_obs` is `0` or `max_memory_mb` is `None`, and exit non-zero. Gate behind a `--fail-on-empty` flag (default on) so manual probes can opt out.

## Acceptance criteria

- A synthetic zero-obs / `None`-memory record makes `main()` return non-zero.
- A normal run stays green.
- `--dry-run` and an explicit `--no-fail-on-empty` opt out.
- Test under the existing benchmark-script tests.

## Refs
- Split from #143 (Phase 3 of the plan there).
- Surfaced by the warm-container OOM run in #143 / #138.

espg:
Closes #137. Closes #145. Refs #141.

## What this does

The **Lambda Benchmark** workflows intermittently — and lately consistently — fail at the Earthdata login step with `OSError: [Errno 101] Network is unreachable` when reaching `urs.earthdata.nasa.gov:443`. That signature is an origin-side IPv6 no-route: the GitHub runner resolves the host's AAAA record, has no usable IPv6 path, and the connect fails immediately (before any packet reaches NASA) instead of falling back to IPv4.

Per the approved plan on #137, this PR lands the fix plus the related benchmark-reliability work in phases:

1. **IPv4 for the Earthdata login (option 2, workflow-level)** + **fold #145** (fail the benchmark on a silent OOM).
2. **Login retry/backoff** (app-level in `src/zagg/auth.py`) so a genuinely transient blip self-heals.
3. **Parallelize the benchmark target launch** — fan out all targets concurrently (cold-favoring + faster wall-clock; the runtime metric is billable-worker-seconds, init-independent, so en-masse launch doesn't bias it).
4. **Per-invocation peak RSS** (rolls in #141) — sample current `VmRSS` scoped to `_handle_process` so warm-container `max_memory_mb` numbers are trustworthy.

## Phases checklist

- [x] **Phase 1 — force IPv4 for the Earthdata login (workflow-level, option 2)** + **fold #145** (fail benchmark on silent OOM)
- [x] **Phase 2 — login retry/backoff in `src/zagg/auth.py`** (+ new `tests/test_auth.py`)
- [x] **Phase 3 — parallelize the benchmark target launch** in `.github/scripts/run_benchmark.py` (+ harness tests)
- [x] **Phase 4 — per-invocation peak RSS** in `deployment/aws/lambda_handler.py` (+ tests), rolls in #141

## Phase 1 detail

**IPv4 (option 2).** Both `lambda-benchmark.yml` (the `merge-publish` job) and the on-demand `lambda-benchmark-command.yml` (the `run` job) dispatch through the shared `./.github/actions/run-benchmark` composite action. A step immediately before each invocation raises the `getaddrinfo` precedence of IPv4-mapped addresses so glibc returns/tries the A record first:

```yaml
- name: Prefer IPv4 for Earthdata login (issue #137)
  run: echo 'precedence ::ffff:0:0/96 100' | sudo tee -a /etc/gai.conf
```

Runner-only config for the CI job's process tree — it does not touch the shipped package or any user's `agg(backend="lambda")` auth (the concern that drove option (2) over the app-level `HAS_IPV6` mutation on #137). Issue #137 names `lambda-benchmark.yml` explicitly, so this infra edit is authorized.

**Fold #145 — fail on silent OOM.** A `--fail-on-empty` flag (default on, `BooleanOptionalAction`) on `run_benchmark.py`: after the target loop, any **real** target whose `total_obs` is falsy or `max_memory_mb` is `None` makes `main()` return non-zero. `--dry-run` is exempt; `--no-fail-on-empty` opts out.

## Phase 2 detail — login retry/backoff

`src/zagg/auth.py` gains `_login_with_retry()` — retries `earthaccess.login()` up to 3× with exponential backoff (2s, 4s) on `OSError`. That base class is the whole requests IO/HTTP-error family (every `requests.exceptions.RequestException` subclasses `IOError`/`OSError`), so it covers the bare `[Errno 101]` no-route, `ConnectionError`, and `Timeout`; a genuinely-rejected credential surfaces as `earthaccess`'s own `LoginAttemptFailure` (not an `OSError`) and so fails fast rather than being retried. Both `get_edl_token` and `get_nsidc_s3_credentials` route through it. New `tests/test_auth.py` pins retry-then-succeed, give-up-after-N (with the exact backoff sequence, no trailing sleep), fail-fast-on-non-`OSError`, the `ConnectionError`-is-an-`OSError` claim, and both public helpers routing through the wrapper.

## Phase 3 detail — parallelize the target launch

`run_benchmark.py`'s serial target loop becomes a `ThreadPoolExecutor` fan-out (cap `min(len, _MAX_TARGET_CONCURRENCY=16)`). Each target is an independent single-shard Lambda dispatch, so this is cold-favoring and cuts wall-clock ~N×; the runtime metric is billable worker-seconds (init-independent), so en-masse launch doesn't bias it. Unknown target names are validated **up front** (before any dispatch), results are re-ordered back to the requested `names` for deterministic output/series rows, and a raising target propagates (aborting the run) as the serial loop did.

**Auth warm-up (folded from self-review).** Because each concurrent `agg()` calls `earthaccess.login()` — which mutates process-global auth/store singletons — N concurrent logins would race to initialize them, contradicting `auth.py`'s "call once" contract. So `main()` calls the new `zagg.auth.ensure_logged_in()` **once before** the pool (guarded by `not dry_run and len(names) > 1`); the in-worker logins then hit the warmed singleton. Tests: the `threading.Barrier(2)` concurrency proof (a serial impl deadlocks → timeout → fails), output-order preservation, up-front unknown-target validation (no dispatch), and the warm-once / skip-on-dry-run wiring.

## Phase 4 detail — per-invocation peak RSS (#141)

`ru_maxrss` is a per-*process* high-water, so on a warm container it reports the max over every prior invocation — untrustworthy, and it can't reflect #140's teardown reclaim. `lambda_handler.py` adds `_PeakRSSSampler`: a daemon thread that samples current `VmRSS` (`/proc/self/status`) every 50 ms while `_handle_process` runs, reported as `max_memory_mb`; `ru_maxrss` is kept as `container_hwm_mb` so the distinction is explicit. Off Linux (no `/proc`) the sampler is a no-op and it falls back to `ru_maxrss`. Sampler is started before the `try`, its peak read after the write phase, stopped in `finally` before the teardown reclaim. Tests (`TestPeakRSSSampler`, controlled `_read_vmrss_kib` feed so they're timing-deterministic): reads current RSS on Linux, reports peak-not-last, two samplers report independent peaks (the #141 acceptance criterion — a light invoke after a heavy one reads its own smaller peak), and the off-Linux no-op fallback.

## How it was tested

- `pytest tests/test_auth.py tests/test_benchmark.py tests/test_lambda_handler.py` — green (auth 7, benchmark harness incl. 4 new parallel/warm-up tests, lambda-handler incl. 4 new sampler tests). Full `tests/test_lambda_handler.py` 35 passed with the `_handle_process` change.
- `ruff check --select=E,F,W,I --ignore=E501` + `ruff format --check` clean on every touched file. Both workflow YAMLs parse.
- IPv4 step is runner config, not unit-testable; acceptance is empirical (consecutive `main` benchmarks green).

## Notes

- Merged up to current `main` (picks up #143/#144). No `main` push, no force-push.
- **Pre-existing, not fixed (§4):** `mypy` flags `_write_elapsed += time.time() - _write_t0` (the #100 profiling block, untouched by this PR) as `float - None`; it exists on `main`, unrelated to this diff.

## Questions for review

- None blocking. Two non-blocking nits noted in the self-review: on a manifest exceeding the 16-way cap a raising target lets already-queued targets finish before the abort (equivalent to serial for the real ≤16 matrix); per-target progress lines now print after all targets complete rather than streaming.

### DIFF
--- a/.github/scripts/run_benchmark.py
+++ b/.github/scripts/run_benchmark.py
@@ -200,6 +200,15 @@ def main(argv: list[str] | None = None) -> int:
         help="One-line banner for the PR comment when the benchmarked worker is "
         "the stable deploy, not this PR's code (issue #25).",
     )
+    parser.add_argument(
+        "--fail-on-empty",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help="Exit non-zero if any real target comes back with zero observations "
+        "or null peak memory -- the signature of a silent OOM that would "
+        "otherwise keep the job green (issue #145). Skipped under --dry-run, "
+        "which emits empty metrics by design; --no-fail-on-empty opts out.",
+    )
     args = parser.parse_args(argv)
 
     manifest, base = load_targets(args.targets)
@@ -247,6 +256,22 @@ def main(argv: list[str] | None = None) -> int:
         Path(args.out_comment).write_text(
             bench_metrics.comment_markdown(records, worker_note=args.worker_note)
         )
+
+    # A silently OOM'd target records obs=0 / max_memory_mb=None but the job
+    # otherwise stays green, so a memory regression can merge (and land a junk
+    # series row) unnoticed (issue #145). Fail loudly on any real target that
+    # came back empty; --dry-run emits empty metrics by design, so it's exempt.
+    if args.fail_on_empty and not args.dry_run:
+        empty = [
+            r["target"] for r in records if not r.get("total_obs") or r.get("max_memory_mb") is None
+        ]
+        if empty:
+            print(
+                "benchmark target(s) returned empty metrics (obs=0 / "
+                f"max_memory_mb=None), likely a silent OOM: {', '.join(empty)}",
+                file=sys.stderr,
+            )
+            return 1
     return 0
 
 
--- a/.github/scripts/run_benchmark.py
+++ b/.github/scripts/run_benchmark.py
@@ -28,7 +28,14 @@ import datetime as dt
 import json
 import os
 import sys
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from pathlib import Path
+
+# Benchmark targets are launched concurrently (issue #137). Each target is one
+# independent single-shard Lambda dispatch, so this bounds how many run at once;
+# the matrix is small (~single digits), so the cap only guards a pathological
+# manifest.
+_MAX_TARGET_CONCURRENCY = 16
 
 # Allow ``import bench_metrics`` whether run as a script or imported by tests.
 sys.path.insert(0, str(Path(__file__).resolve().parent))
@@ -226,14 +233,17 @@ def main(argv: list[str] | None = None) -> int:
         "pr_number": pr_number,
     }
 
-    records = []
+    # Validate every requested target up front so an unknown name fails before any
+    # dispatch (and before the pool spins up).
     for name in names:
         if name not in known:
             raise SystemExit(f"unknown target '{name}'; have {sorted(known)}")
+
+    def _dispatch(name: str) -> dict:
         store = None
         if args.store_prefix:
             store = f"{args.store_prefix.rstrip('/')}/{name}.zarr"
-        record = run_target(
+        return run_target(
             name,
             manifest,
             base,
@@ -243,9 +253,24 @@ def main(argv: list[str] | None = None) -> int:
             context=context,
             dry_run=args.dry_run,
         )
-        records.append(record)
+
+    # Launch all targets concurrently (issue #137). Each is an independent
+    # single-shard Lambda dispatch, so fanning them out is cold-favoring (a fresh
+    # container per concurrent invoke) and cuts wall-clock ~N x; the runtime metric
+    # is billable worker-seconds (init-independent), so en-masse launch doesn't bias
+    # it. Results are collected then re-ordered back to ``names`` for deterministic
+    # output/series rows; a target that raises propagates (aborting the run) exactly
+    # as the prior serial loop did.
+    records_by_name: dict = {}
+    max_workers = min(len(names), _MAX_TARGET_CONCURRENCY) or 1
+    with ThreadPoolExecutor(max_workers=max_workers) as pool:
+        futures = {pool.submit(_dispatch, name): name for name in names}
+        for fut in as_completed(futures):
+            records_by_name[futures[fut]] = fut.result()
+    records = [records_by_name[name] for name in names]
+    for record in records:
         print(
-            f"[{name}] obs={record['total_obs']} runtime_s={record['runtime_s']} "
+            f"[{record['target']}] obs={record['total_obs']} runtime_s={record['runtime_s']} "
             f"cost/shard=${record['cost_per_shard_usd']} "
             f"cost/100km2=${record['cost_per_100km2_usd']} "
             f"max_memory_mb={record['max_memory_mb']}"
--- a/.github/scripts/run_benchmark.py
+++ b/.github/scripts/run_benchmark.py
@@ -239,6 +239,15 @@ def main(argv: list[str] | None = None) -> int:
         if name not in known:
             raise SystemExit(f"unknown target '{name}'; have {sorted(known)}")
 
+    # Authenticate once up front so the concurrent targets below don't race to
+    # initialize earthaccess's process-global auth singleton (issue #137). Skipped
+    # under --dry-run (no dispatch, no auth) and pointless for a single target (no
+    # fan-out); each agg() still logs in, but now hits the warmed singleton.
+    if not args.dry_run and len(names) > 1:
+        from zagg.auth import ensure_logged_in
+
+        ensure_logged_in()
+
     def _dispatch(name: str) -> dict:
         store = None
         if args.store_prefix:
--- a/.github/workflows/lambda-benchmark-command.yml
+++ b/.github/workflows/lambda-benchmark-command.yml
@@ -139,6 +139,16 @@ jobs:
           rm -rf src/zagg
           cp -r _fork/src/zagg src/zagg
 
+      # The runner resolves urs.earthdata.nasa.gov to an AAAA record but has no
+      # usable IPv6 route, so earthaccess.login() (inside run_benchmark.py) dies
+      # immediately with `[Errno 101] Network is unreachable` (issue #137).
+      # Raising the getaddrinfo precedence of IPv4-mapped addresses makes glibc
+      # return+try the A record first, so the login handshake takes the routable
+      # IPv4 path. Runner-only config -- the shipped package and user auth are
+      # untouched.
+      - name: Prefer IPv4 for Earthdata login (issue #137)
+        run: echo 'precedence ::ffff:0:0/96 100' | sudo tee -a /etc/gai.conf
+
       - name: Run benchmark (base harness, fork package)
         uses: ./.github/actions/run-benchmark
         with:
--- a/.github/workflows/lambda-benchmark-command.yml
+++ b/.github/workflows/lambda-benchmark-command.yml
@@ -163,11 +163,20 @@ jobs:
           pr-number: ${{ needs.check.outputs.pr_number }}
           worker-note: ${{ needs.check.outputs.worker_note }}
 
+      # always(): a silent-OOM target now fails the benchmark step (--fail-on-empty,
+      # issue #145), but the maintainer who ran /benchmark should still see WHY --
+      # the just-written comment.md carries the obs=0 row. Skip cleanly if the run
+      # died before writing it (e.g. login failure -> no comment to post).
       - name: Upsert PR comment
+        if: always()
         uses: actions/github-script@v7
         with:
           script: |
             const fs = require('fs');
+            if (!fs.existsSync('comment.md')) {
+              core.info('no comment.md (benchmark did not get far enough); skipping');
+              return;
+            }
             const body = fs.readFileSync('comment.md', 'utf8');
             const marker = '<!-- zagg-benchmark -->';
             const { owner, repo } = context.repo;
--- a/.github/workflows/lambda-benchmark.yml
+++ b/.github/workflows/lambda-benchmark.yml
@@ -168,6 +168,16 @@ jobs:
       - name: Checkout merge commit
         uses: actions/checkout@v5
 
+      # The runner resolves urs.earthdata.nasa.gov to an AAAA record but has no
+      # usable IPv6 route, so earthaccess.login() (inside run_benchmark.py) dies
+      # immediately with `[Errno 101] Network is unreachable` (issue #137).
+      # Raising the getaddrinfo precedence of IPv4-mapped addresses makes glibc
+      # return+try the A record first, so the login handshake takes the routable
+      # IPv4 path. Runner-only config -- the shipped package and user auth are
+      # untouched.
+      - name: Prefer IPv4 for Earthdata login (issue #137)
+        run: echo 'precedence ::ffff:0:0/96 100' | sudo tee -a /etc/gai.conf
+
       - name: Run benchmark
         uses: ./.github/actions/run-benchmark
         with:
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -61,8 +61,9 @@ import json
 import logging
 import os
 import resource
+import threading
 import time
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from zarr import open_group
 from zarr.errors import GroupNotFoundError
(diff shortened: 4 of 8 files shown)


.github/scripts/run_benchmark.py
```diff
@@ -247,6 +256,22 @@ def main(argv: list[str] | None = None) -> int:
...
             bench_metrics.comment_markdown(records, worker_note=args.worker_note)
         )
+
+    # A silently OOM'd target records obs=0 / max_memory_mb=None but the job
+    # otherwise stays green, so a memory regression can merge (and land a junk
+    # series row) unnoticed (issue #145). Fail loudly on any real target that
+    # came back empty; --dry-run emits empty metrics by design, so it's exempt.
+    if args.fail_on_empty and not args.dry_run:
+        empty = [
+            r["target"] for r in records if not r.get("total_obs") or r.get("max_memory_mb") is None
```
espg, line 300: 🤖 *from Claude (review)*

The empty-metrics guard is an OR of two clauses, but the two tests that hit an empty target both set `total_obs=0` **and** `max_memory_mb=None` at once (`_fake_record(..., total_obs=0, max_memory_mb=None)`). So the `max_memory_mb is None` clause is never exercised on its own — a regression that dropped it (leaving only the `total_obs` check) would still pass all three new tests. Consider adding a case with `total_obs>0, max_memory_mb=None` (memory-null on an otherwise-populated record) so each clause is independently covered.

Separately, `not r.get("total_obs")` treats obs=0 as an OOM. Thats correct for these densest-shard targets by design, but note the assumption: a genuinely empty region — or a successful run whose summary omitted `max_memory_mb` — would trip a false failure. Worth a one-line comment noting the "densest shard always has obs + profile always reports memory" premise this leans on.

.github/scripts/run_benchmark.py
```diff
@@ -247,6 +256,22 @@ def main(argv: list[str] | None = None) -> int:
         Path(args.out_comment).write_text(
             bench_metrics.comment_markdown(records, worker_note=args.worker_note)
         )
+
+    # A silently OOM'd target records obs=0 / max_memory_mb=None but the job
+    # otherwise stays green, so a memory regression can merge (and land a junk
+    # series row) unnoticed (issue #145). Fail loudly on any real target that
+    # came back empty; --dry-run emits empty metrics by design, so it's exempt.
+    if args.fail_on_empty and not args.dry_run:
```
espg, line 298: 🤖 *from Claude (review)*

The `not args.dry_run` exemption is the right call (dry-run emits empty metrics by design), but no test asserts it. `test_main_dry_run_writes_outputs` calls `main([... --dry-run ...])` **without capturing/asserting the return code**, so if this exemption were removed the dry-run path would start returning 1 and CI would stay green. Suggest adding `rc = run_benchmark.main([... --dry-run ...]); assert rc == 0` (or extend the existing dry-run test) so the exemption is actually pinned.

.github/scripts/run_benchmark.py
```diff
@@ -247,6 +256,22 @@ def main(argv: list[str] | None = None) -> int:
...
+        empty = [
+            r["target"] for r in records if not r.get("total_obs") or r.get("max_memory_mb") is None
+        ]
+        if empty:
+            print(
+                "benchmark target(s) returned empty metrics (obs=0 / "
+                f"max_memory_mb=None), likely a silent OOM: {', '.join(empty)}",
+                file=sys.stderr,
+            )
+            return 1
```
espg, line 308: 🤖 *from Claude (review)*

Note the ordering effect: `return 1` fires **after** metrics.json and comment.md are written, so the composite `run-benchmark` step exits non-zero and the job goes red. For the merge workflow thats exactly what #145 wants — the downstream "Check out benchmarks data branch" / "Append series" steps have no `if:` guard, so they are skipped and no junk row is appended. Good.

For the command workflow (`lambda-benchmark-command.yml`), the same non-zero exit skips the unguarded "Upsert PR comment" step — so on a silent OOM the maintainer sees only a red job, and the `comment.md` that was just written (showing obs=0 / mem=None) is never posted. Since that comment is arguably the most useful OOM signal for a human, consider `if: always()` on the Upsert-PR-comment step so the empty-result table still lands. (Low severity / design call — a red job is already a signal.)

espg:
🤖 *from Claude*

Folded the fresh-context adversarial self-review (three inline findings) in `ccb6eb9`:

1. **Test gap — `max_memory_mb is None` clause never exercised alone** (`run_benchmark.py:266`). Both empty-target tests set `total_obs=0` *and* `max_memory_mb=None`, so dropping the memory clause would still pass CI. Added `test_main_fails_on_null_memory_alone` (`total_obs=999, max_memory_mb=None` → `main()` returns 1) so each clause is pinned independently.
2. **Dry-run exemption unasserted** (`run_benchmark.py:264`). `test_main_dry_run_writes_outputs` never checked the return code, so removing `not args.dry_run` wouldn't be caught. Added `test_main_dry_run_exempt_from_fail_on_empty` asserting a `--dry-run` run returns 0 despite `--fail-on-empty` being on by default.
3. **Command-path OOM produced a red job but no comment** (`lambda-benchmark-command.yml`). On a silent OOM the benchmark step now fails (`--fail-on-empty`), which skipped the unguarded "Upsert PR comment" step — so a maintainer who ran `/benchmark` saw only a red X and never the just-written `comment.md` (which carries the obs=0 row). Added `if: always()` to that step, guarded by an `fs.existsSync('comment.md')` check so a run that died *before* writing the comment (e.g. a login failure) skips cleanly instead of erroring.

The review also verified two non-findings I'll note for the record: the `gai.conf` IPv4 snippet is correct (single active line replaces the default precedence table, but the net effect is the intended IPv4>IPv6; `urs.earthdata.nasa.gov` has both A and AAAA records; ephemeral runners so `tee -a` won't accumulate), and the step placement covers both login-bearing jobs (`merge-publish` and command `run`; `deploy-test` doesn't log in).

`pytest tests/test_benchmark.py` → 72 passed; `ruff check` / `ruff format --check` clean; both workflow YAMLs parse.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of phases 2–4 (`eea16e3`/`3c343e5`/`1ed7aff`). **No blocking findings.** Verified: `requests.exceptions.RequestException` subclasses `IOError`/`OSError` (so `OSError` is the right retry net, and a rejected credential surfaces as `LoginAttemptFailure` — not an `OSError` — so it fails fast); Phase 4's `peak_mb` is read after the write phase while the sampler still runs, which is correct (teardown `malloc_trim` only *lowers* RSS, so nothing higher is missed); `_read_vmrss_kib` parses `VmRSS` (kB=KiB → /1024) correctly; the sampler thread is daemon + `join(timeout=1.0)`, no leak across warm invokes; `lambda_handler.py` is 578 lines (< ~1000). Tests are deterministic (the `_feeder` advances per read-call, not wall-clock, so the peak-not-last and per-invocation-independence proofs aren't timing-flaky; the `Barrier(2)` genuinely fails a serial impl).

**Folded the one escalation-worthy finding** (`a3d1fb2`): Phase 3 fanned `run_target` out over threads, and each `agg(backend="lambda")` calls `earthaccess.login()`, which mutates process-global auth/store singletons — N concurrent logins race to initialize them, contradicting `auth.py`'s "call once" contract. Added `zagg.auth.ensure_logged_in()` and call it **once before** the pool (guarded by `not dry_run and len(names) > 1`) so the in-worker logins hit the warmed singleton. Tests: `test_ensure_logged_in_warms_singleton_with_retry`, `test_main_warms_auth_once_before_parallel_dispatch`, `test_main_skips_auth_warmup_on_dry_run`. Also tightened the retry docstring to state the `OSError` net is the whole requests IO/HTTP family (not just "connection/timeout").

Nits left standing (non-blocking): on a manifest exceeding the 16-way cap, a raising target lets already-queued targets finish before the abort propagates (equivalent to serial for the real ≤16 matrix); per-target progress lines now print after all complete rather than streaming.

One pre-existing item flagged, not fixed (§4): `mypy` reports `_write_elapsed += time.time() - _write_t0` (the #100 profiling block, unchanged by this PR) as `float - None` — it exists on `main`, unrelated to this diff.

---

espg:
🤖 *from Claude*

Continued the PR — **phases 2, 3, 4 landed** (`eea16e3`, `3c343e5`, `1ed7aff`), fresh-context self-review folded (`a3d1fb2`). The full checklist is now complete:

- **Phase 2** — `_login_with_retry()` in `src/zagg/auth.py` (bounded retry/backoff on `OSError`, which is the whole requests IO/HTTP family; rejected creds fail fast as `LoginAttemptFailure`). Both auth helpers route through it. New `tests/test_auth.py`.
- **Phase 3** — parallelized the benchmark target launch over a `ThreadPoolExecutor` (order-preserving, up-front validation). Self-review caught that N concurrent `earthaccess.login()` calls race the process-global auth singleton, so folded a `zagg.auth.ensure_logged_in()` warm-up called once before the pool.
- **Phase 4** — `_PeakRSSSampler` in `deployment/aws/lambda_handler.py` samples per-invocation `VmRSS`, reported as `max_memory_mb` (keeping `ru_maxrss` as `container_hwm_mb`), falling back off Linux (rolls in #141).

Green locally: `tests/test_auth.py` + `tests/test_benchmark.py` + `tests/test_lambda_handler.py` all pass (incl. the new parallel/warm-up + sampler tests); ruff clean; both workflow YAMLs parse. One **pre-existing** `mypy` red flagged (not fixed, §4): `_write_elapsed += time.time() - _write_t0` in the #100 profiling block, untouched here.

All four phases done and folded; nothing needs a decision. Marking `waiting` (kept `implement`) — the ball is with you for review/merge. Two non-blocking nits are noted in the PR body/self-review. Staying a **draft**.

---

PR #147, opened 2026-07-01
Parallelize the sharded path's K ragged (CSR) writes (issue #142)

### LINKED ISSUE #142: ShardingCodec doesn't consolidate ragged (tdigest) writes — K serial CSR groups dominate shard write time
🤖 *from Claude*

## Summary

On a t-digest product, the `ShardingCodec` path (#108) is dominated by writing the **ragged (CSR) tdigest field, which is not sharded** and is emitted as **K separate CSR subgroups written in a serial loop**. At K=256 that's ~1500 tiny S3 objects per shard and adds ~110 s to a single o9 shard — the codec itself is correctly configured and is *not* the cost.

## Evidence (densest o9 shard `5347395636851376137`, 3.1 M obs, arm64/2 GB, billed duration)

| config | K | dense writes | ragged (CSR) writes | billed |
|---|---|---|---|---|
| `inner_K1` (chunk_inner=parent) | 1 | 1 block | **1 CSR group** (~6 objects) | **347 s** |
| `sharded` (chunk_inner=13) | 256 | 1 shard-object slab | **256 CSR groups** (~1500 objects) | 453 s |
| `inner` (chunk_inner=13, regular) | 256 | 256 blocks | 256 CSR groups | 505 s |

Read+aggregate is ~343–348 s for all three; the spread is **entirely write-out**. `inner_K1` is fastest purely because K=1 collapses the ragged field to one CSR group. The dense-side sharding *helps* (sharded's single slab is ~52 s faster than regular-inner's 256 dense block objects) — so the ShardingCodec is doing its job; the ragged field is the problem.

## Root cause

`processing/write.py::write_shard_to_zarr` consolidates the **dense** arrays into one `set_block_selection` per array (correct, one shard object), but its own docstring notes the ragged fields are **NOT sharded** — they're written per inner chunk:

```python
# write_shard_to_zarr, per inner chunk (K of them):
write_ragged_to_zarr(ragged, store, grid=grid, shard_key=_block_index_key(block_index, grid))
```

Each call → `write_csr(...)` creates a subgroup of 3 arrays (`values`/`offsets`/`cell_ids`) at `{group}/{field}/{block_key}/`. At K=256 that's ~256×3 ≈ 768 Zarr arrays ≈ ~1500 objects. Two aggravating factors:
1. **Serial.** The `with config.set({"async.concurrency": 128})` fan-out wraps only the dense `set_block_selection` writes; the K ragged writes run in a plain Python loop.
2. **Latency-bound, not bytes.** The tdigest payload for the whole shard is only tens of MB; the cost is the object *count* × per-object round-trip, not throughput.

Net: **for a t-digest (ragged-dominated) product, `sharded: true` gives almost no write-consolidation benefit** — #108 bundles the small dense arrays but the payload field still emits K CSR groups.

## Proposed fixes (independent; either helps, both compose)

1. **Parallelize the K ragged writes.** They target disjoint subgroups, so they're embarrassingly parallel but currently serial. Fan them out (same `async.concurrency` budget, or a bounded thread pool) — should reclaim a large share of the ~100 s with **zero on-disk format change**. Lowest-risk first step.
2. **Consolidate the ragged field per shard.** Write one CSR group for the whole dispatch shard (concatenate the K inner chunks' payloads, offsetting `cell_ids` into the shard-wide cell space) instead of K groups — mirroring what `inner_K1` already produces, but decoupled from dense chunking so we keep the fine 64×64 dense read granularity *and* a single tdigest write. Requires a reader-side change (the `{group}/{field}/{shard_key}` layout in `readers/tdigest_tensor.py` already keys by shard at K=1; K>1 currently keys by block).

## Not a bug in the codec
`grids/base.py::sharded_array_spec` is configured correctly (bytes-only inner codecs, crc32c index, `index_location: end`, one object per dispatch shard). No change needed there.

## Refs
- #108 (ShardingCodec eval) — this quantifies the limit of the current sharding for ragged fields.
- #48 (ragged/CSR fields), #133 (shard_order object split).
- Surfaced during the #138 memory/perf investigation (o9 write-cost profiling).

espg:
Closes #142.

## What / approach

Option (1) from the [issue discussion](https://github.com/englacial/zagg/issues/142#issuecomment-4853709495), per @espg's decisions ([here](https://github.com/englacial/zagg/issues/142#issuecomment-4858891191): ship (1) only, reuse 128): on the sharded path, the dispatch shard's K ragged (CSR) subgroup writes ran in a **serial** Python loop in `write_shard_to_zarr`, which dominated a t-digest shard's write time (~K×3 tiny S3 objects, latency- not bytes-bound — ~110 s on a K=256 o9 shard). Those subgroups target **disjoint** prefixes (`{group_path}/{field}/{block_key}/…`), so they are write-independent — this fans them out over a bounded `ThreadPoolExecutor` instead. **Zero on-disk format change** (still one CSR group per inner chunk; only the write scheduling changes).

### Why the sharded path only (not the regular/streaming path)

The evidence table lists both `sharded` (453 s) and the regular `inner` (505 s) at K=256, but only the **sharded** path is safe to parallelize: it already accumulates all K inner-chunk carriers in memory (it bundles the dense side into one shard object), so fanning out its ragged writes adds **no** memory cost. The regular/streaming path deliberately writes-then-frees each chunk to bound peak memory (#91/#124) — it never holds the K carriers at once, so parallelizing it would break that memory bound. So this change is scoped to `write_shard_to_zarr` and the streaming path stays serial (noted in the helper docstring).

### Concurrency bound

Reuses 128 (the dense path's `async.concurrency` budget, per @espg), but via a `ThreadPoolExecutor` since `write_csr` uses the **sync** `zarr.open_array` API (not the async-concurrency knob). The cap is `min(128, len(writes), fd_safe_max_workers())` so this per-worker inner fan-out respects the same open-file ceiling the Lambda dispatcher already guards (`zagg.concurrency.fd_safe_max_workers`) — each in-flight write holds a store socket. Concurrent writes to disjoint keys of one store are already zagg's model (the dispatcher fans cells over threads onto a shared store). Failures are surfaced (not swallowed): the first exception is re-raised after the pool drains, tagged with how many of the K writes failed.

## Changes

- `src/zagg/processing/write.py` — collect the per-inner-chunk ragged writes in `write_shard_to_zarr` and route them through a new `_write_ragged_fanout(...)` helper (companion writes + dense slab writes unchanged; ragged still runs before the dense loop, so store contents are order-independent and byte-identical). New `_RAGGED_WRITE_CONCURRENCY = 128` constant.
- `tests/test_processing.py` — new `TestRaggedWriteFanout`: all-subgroups-written past the 128 cap (real concurrent writes), **byte-identical parallel-vs-serial** (proves no format change + thread-safety), failure-surfacing, empty no-op, and a wiring test that `write_shard_to_zarr` invokes the fan-out exactly once per shard.

## How tested

- `uv run pytest tests/test_processing.py::TestRaggedWriteFanout -q` → **5 passed** (2.9 s).
- `uv run pytest tests/test_processing.py::TestWriteShardToZarr tests/test_processing.py::TestShardOrderObjectSplit tests/test_processing.py::TestRaggedCsrWrite -q` → **12 passed** (existing sharded/ragged coverage, no regression).
- `ruff check --select=E,F,W,I --ignore=E501` clean; `ruff format --check` clean on both files. `mypy src/zagg/processing/write.py` adds no new errors (only pre-existing missing-stub imports).

## Questions for review

- **Local-backend fan-out depth (the acknowledged unknown).** The 128 cap bounds the inner fan-out *per worker*. On Lambda each worker is its own isolated container, so 128 concurrent S3 sockets is well under the ~1024 ulimit — safe. On the **local** backend a worker runs inside the orchestrator's `ThreadPoolExecutor`, so the theoretical peak is `outer_cells × 128` threads/sockets; the `fd_safe_max_workers()` floor guards the process-wide FD ceiling, but doesn't coordinate across the outer pool. This matches the tradeoff you flagged. It's a non-issue for the Lambda production path; if you'd want a smaller local cap, that's a one-line change to `_RAGGED_WRITE_CONCURRENCY` (or a backend-aware cap) — flagging rather than guessing.

## Deferred (out of scope, per @espg)

Option (2) — consolidating the ragged field to one CSR group per dispatch shard — is deferred. Confirmed from the codebase (in the issue thread) that it's *not* a cheap re-key: a dispatch shard is K 64×64 coverage chunks (not one), so it needs re-offsetting per-chunk `cell_ids` into shard-wide space + generalizing the reader's fixed 64×64/`divmod` assumption. It composes with this PR when scoped.

---
_Generated by [Claude Code](https://claude.ai/code/session_017wNcBZd6KvDfe7mSeQfo49)_

### DIFF
--- a/src/zagg/processing/write.py
+++ b/src/zagg/processing/write.py
@@ -7,11 +7,14 @@ Assembles the per-shard output carrier and writes it to the Zarr template
 stays acyclic.
 """
 
+from concurrent.futures import ThreadPoolExecutor, as_completed
+
 import numpy as np
 import pandas as pd
 from zarr import config, open_array
 from zarr.abc.store import Store
 
+from zagg.concurrency import fd_safe_max_workers
 from zagg.config import (
     PipelineConfig,
     get_agg_fields,
@@ -19,6 +22,13 @@ from zagg.config import (
 )
 from zagg.csr import write_csr
 from zagg.grids.morton import is_morton_array, morton_words
+
+# The sharded path's K ragged (CSR) subgroup writes fan out over a bounded thread
+# pool (issue #142): each inner chunk emits an independent CSR group at a disjoint
+# prefix, so they are write-independent and latency-bound (~K×3 tiny objects), and
+# a serial loop dominated a t-digest shard's write time. 128 mirrors the dense
+# path's ``async.concurrency`` budget (issue #108).
+_RAGGED_WRITE_CONCURRENCY = 128
 
 
 def _arrow_column(block: np.ndarray, sig: dict):
@@ -299,16 +309,20 @@ def write_shard_to_zarr(
     # accumulated and written independently; preserve encounter order for stable,
     # deterministic writes. Companions + ragged stay per inner chunk (unsharded).
     objects: dict = {}
+    ragged_writes: list = []
     for block_index, carrier, ragged in chunk_results:
         if not _carrier_empty(carrier):
             objects.setdefault(_object_block(block_index), []).append((block_index, carrier))
-        # Companions + ragged for this inner chunk are not sharded: write them
-        # straight through (companion array is one block per chunk; ragged is CSR).
+        # Companions for this inner chunk are not sharded: write them straight
+        # through (one block per chunk). Ragged (CSR) subgroups are collected and
+        # fanned out below (issue #142) -- they target disjoint prefixes, so the
+        # per-chunk serial loop needlessly dominated a t-digest shard's write time.
         if chunk_res_fields:
             _write_companion_columns(carrier, store, grid, block_index, chunk_res_fields)
-        write_ragged_to_zarr(
-            ragged, store, grid=grid, shard_key=_block_index_key(block_index, grid)
-        )
+        if ragged:
+            ragged_writes.append((ragged, _block_index_key(block_index, grid)))
+
+    _write_ragged_fanout(ragged_writes, store, grid=grid)
 
     # One accumulate→write→free pass per sharding object: each holds at most one
     # object's slab resident at a time (the phase-8 memory bound).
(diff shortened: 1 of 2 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the diff. **No blocking findings** — the change is correct and safe:

- **Disjoint keys / no race.** Each `chunk_results` entry is a distinct inner chunk → distinct `block_index` → `_block_index_key` is injective, so the K CSR subgroups target disjoint store prefixes with no RMW hazard (unlike the dense shard object, which is why *that* stays a single block-selection write).
- **Thread-safety holds.** The fan-out does **not** enter the `async.concurrency` config context (only the dense slab loop does), so there's no zarr global-config race. `write_csr` → `zarr.open_array` routes through zarr v3's `sync()` bridge, which submits cross-thread onto one dedicated IO loop via `run_coroutine_threadsafe` — the supported pattern. `get_agg_fields`/`get_output_signature` are read-only; MemoryStore's dict `__setitem__` on disjoint keys is GIL-atomic. The n=200 test exercises real 128-way-eligible concurrency.
- **Error handling correct.** `as_completed` drains the whole pool (no writes abandoned), collects every exception, re-raises the first chained with an accurate failed-count.
- **`if ragged:` filter is behavior-equivalent** to the old unconditional call (which early-returned on falsy). The streaming/non-sharded path staying serial for the #91 memory bound is accurate (`runner.py:404` `_write_chunk` write-then-free).

**Folded two non-blocking test-strengthening findings** (commit `b7cd304`):
1. `test_fanout_byte_identical_parallel_vs_serial` could silently degenerate to serial-vs-serial on a low-`ulimit` host (fd ceiling clamps workers to 1) — now pins `fd_safe_max_workers → 128` on the parallel side so it deterministically takes the pool branch.
2. Added `test_fanout_cap_respects_fd_ceiling` — a direct test that the pool is sized `min(128, len, fd_safe_max_workers())` and that an fd ceiling of 1 forces the serial branch (no pool constructed).

Nits left standing (non-diff-scoped): the `_RAGGED_WRITE_CONCURRENCY` constant is a thread-count knob distinct from zarr's `async.concurrency` (they just share the value 128) — noted for clarity, not a change.

---

espg:
🤖 *from Claude*

Implemented option (1) per your relabel to `implement` — parallelized the sharded path's K serial ragged (CSR) writes via a bounded `ThreadPoolExecutor` (`b7cd304`). Fresh-context self-review came back with no blocking findings (disjoint-key correctness + zarr `sync()` cross-thread safety verified); folded its two non-blocking test-strengthening items (pin the parallel branch deterministically; add a direct fd-cap test).

Green locally: `TestRaggedWriteFanout` 6 passed; full `test_processing.py` **176 passed**; the existing sharded/ragged coverage (`TestWriteShardToZarr`/`TestShardOrderObjectSplit`/`TestRaggedCsrWrite` + rectilinear `TestSharded`) unchanged; `ruff` clean, `mypy` no new errors. CI running on the PR.

Single-phase change, all done and folded. Marking `waiting` (kept `implement`) — the ball is with you for review/merge. One non-blocking **Question for review** stands in the PR body: whether you want a smaller *local-backend* fan-out cap (on Lambda each worker is isolated so 128 is safe; locally the theoretical peak is `outer_cells × 128`, guarded by the `fd_safe_max_workers()` floor but not coordinated across the outer pool). Staying a **draft**. Option (2) consolidation remains deferred as scoped.

---

Issue #149, opened 2026-07-01
compiled h5coro extension


espg:
Enabling #148 will require squeezing out more performance than what we're currently seeing. The place with the most gains to be realized is on the read side: h5coro as used today is pure python implementation. There exists a c++ bindings in the https://github.com/SlideRuleEarth/sliderule repo, specifically here: https://github.com/SlideRuleEarth/sliderule/tree/main/packages/h5coro

The h5coro c++ bindings are heavily integrated in sliderule , with lots of cross dependencies. We need to get those compiled routines, however, the best approach *how* to do this is unclear. A few options:

1. compiled extension specifically for zagg ; deploys on lambda arm64
2. generic c++ bindings for h5coro (the python package); added as PR upstream
3. rust port, packaged either separately as it's own package, or as an extension to h5coro 

Option 2 means supporting multiple platforms (i.e., windows, macos), which will run into the linux specific cross libs that sliderule is built around. Option 1 has it's own dangers-- we **must** have our solution fit on an AWS lambda, and only have ~100mb of headroom for new code deps to distribute. A straight ingest and compilation of the code in 1 may run into these limits.

Option 3 is appealing-- but it needs to be performance competitive with the c++ implementation-- which means we'd need to benchmark it locally against a local build of the sliderule package on our test dataset.

## Local notes

You have local clones of both h5coro and sliderule at ~/software/h5coro and ~/software/sliderule respectively; you should read both of these codebases. You also have a local directory with test granules at ~/ignore/zagg_neon_atl03_test_shard ; you should read the readme there.

For the benchmarks, we have podman, and we're running on macos arm64 -- this means that you can run a podman linux container with arm64 natively (no qemu or similar is needed).

We'll want to benchmark our current workflow that we have in tests for the test shard (defined at orders o10 and o9 ), and then compare that with what improvement we see when switching to sliderule compiled code (so we have a baseline). Since the ~/ignore files are local, this will give us a delta on the readtime (de)serialization improvements -- this ignores the io portion of s3 streaming, but is still *very* useful in isolating what that portion of the read ingest costs, and what improvements we can expect.

Give me your thoughts , concerns , ideas on the above (posted here); then I'll respond and we can move towards 'plan'

espg:
🤖 *from Claude*

Read both trees (`~/software/h5coro`, `~/software/sliderule/packages/h5coro`) plus zagg's read path, and the test-shard README. What's actually in each codebase first, then thoughts per option, then a proposed order.

[...]

Issue #148, opened 2026-07-01
add 88 south stress test shard + benchmark


espg:
We'd like to have zagg work on the atl03 catalog globally; that means planning and testing for the worse case scenario. That scenario is reached near the 88S 'pole hole' , a donut shaped region that follows the 88 degree south line of latitude where all of the ICESat-2 orbits converge. The coverage becomes dense around this region, and is where we should stress test our pipeline. 

Ideally, we'd be able to run tdigest aggregation at o9 ; given that this region is particularly demanding, we should start an order up at o10 first to see if we can handle that. Handle here means being able to run on a 2GB lambda worker, without running out of memory or out of time. To make this somewhat easier, we should preemptively bump the timeout from 720 to 900 seconds while we test this.

Picking o9 and o10 shards and querying the catalog + building the shard maps for the 2018 to Jan 2026 will quickly surface how big of a lift this is given the total number of granuals that need to be inspected. We'll want to compare two approaches:

1. What we have now (i.e., hierarchical mapping from the geolocation segments to retrieve the index ranges and chunks)
2. Processing when we know the chunk offsets ahead of time

## Cached chunk boundries

Option 2 doesn't exist yet. It assumes the following:

- At some point, the CMR geometry metadata object is changed such that we can infer chunk boundaries from it
- This would mean instead of a polygon, we would have 6 linestrings / polylines , one per beam
- The linestrings would encode the lat/lon of the first / last locations of the photons on the chunk boundaries

This needs some thought and design. We'd ideally not like to encode both start and stop ends per linestring, since that doubles the size of the geometry and is redundant. Instead, we'd like to encode the start lat/lon of each photon at the chunk boundry, and then infer the end given the start of the next chunk photon recorded. This would give N+1 points per beam, where N is the number of chunks, and the +1 account to the fact that we need to record the last photon to be consistent with metadata requirements that we represent the whole line path and don't truncate the geometry. This is all future refinement-- we can start with recording start/end pairs to prototype for correctness.

We'll need our own lambda function for above; there will be thousands of granules that need to be read for o10 , and it will take took long to run in serial to grab this data. The lambda functions are simple: for each granule, read the datafile chunk by chunk, grab the first+last lat / lon per beam in the heights group, and then accumulate those per beam pairs to write out as a parquet file on an s3 directory. 

[...]

PR #150, opened 2026-07-01
h5coro read-path benchmark: profile + 4-variant comparison (issue #149)

espg:
Refs #149.

Benchmark investigation for the compiled-h5coro question: measures where zagg's granule-read time actually goes and what compiled alternatives buy, on the exact `readDatasets` workload the worker issues (captured once, replayed by every variant), with a hard byte-equality correctness gate. Decode cost is isolated from S3 latency by reading the locally cached NEON test-shard granules with `FileDriver` — per the plan on the issue ([plan comment](https://github.com/englacial/zagg/issues/149#issuecomment-4860097668)).

Everything lands under `bench/h5coro/` — no `src/zagg/` behavior changes. Granule `.h5` files are NASA EOSDIS data and are not committed; only captured request lists (dataset paths + index ranges), checksums, and results are.

## Findings so far

**Phase 1 (decomposition).** The pure-Python `shuffleChunk` de-shuffle is **68–70% of read wall time** (o10: 64.2 s of 94.6 s; o9: 97.1 s of 139.6 s, cProfile self-time). The big float coordinate datasets aren't shuffled at all — but `signal_conf_ph` (the confidence filter) is, and at `n_photons × 5` int8 it's the largest dataset by element count, de-shuffled element-wise in Python. The C zlib inflate floor is only ~16%. Read replay is also ~93% of total worker wall (o10: 85.5 s replay vs 91.9 s full worker incl. aggregation), so the read path is the right lever. Full decomposition in `bench/h5coro/results/profile_*.json`.

**Final canonical rows (linux/arm64 podman, serial, read-window timing, byte-equality gate passing on every row) — full analysis in `bench/h5coro/results/REPORT.md`:**

| variant | o10 read wall | o9 read wall | peak RSS (o10/o9) |
|---|---|---|---|
| h5coro 1.0.4 (baseline) | 93.5 s | 129.0 s | 248 / 282 MB |
| h5coro + numpy shuffle patch (`patches/`) | **24.5 s (3.8×)** | **34.0 s (3.8×)** | 255 / 286 MB |
| hidefix 0.12.0 (Rust, incl. per-granule index build) | **15.5 s (6.0×)** | **20.8 s (6.2×)** | 155 / 161 MB |
| sliderule C++ H5Coro (pybind shim) | **10.3 s (9.1×)** | **14.2 s (9.1×)** | 187 / 208 MB |

End-to-end worker check: `process_shard` on the o10 shard drops **91.9 s → 23.5 s (3.9×)** with only the numpy patch swapped in, identical output. hidefix's serialized chunk index (~0.2 s / ~600 KB per ~1.9 GB granule) carries exactly the `(addr, size, offset)` payload #148's offset cache needs; with an upstream save/load surface its cached replay (~5–8 s) reaches shim territory. Recommendation (REPORT.md): (1) upstream the numpy patch now; (2) pursue hidefix + a small upstream index-save/load PR as the compiled step *and* #148's cache mechanism; (3) keep the C++ shim as the measured 9× ceiling, not the product path.

## Phases

- [x] Phase 1 — request capture + baseline profile: `capture_requests.py` (wraps `H5Coro.readDatasets` during a real `process_shard` run on the densest o10/o9 benchmark shards), `bench_replay.py` (replay + wall/CPU/RSS + sha256 correctness gate + cProfile bucketing), baseline rows (macOS host + podman linux/arm64) and decode-time decomposition.
- [x] Phase 2 — numpy-vectorized h5coro comparable: patch against the h5coro clone (staged under `patches/` for upstream hand-off), re-measured as a separate row (unit-verified byte-identical vs the original loop across type sizes and ragged windows).
- [x] Phase 3 — pybind11 shim over sliderule's C++ `H5Coro::read` (`shim/`): real minimal sliderule build succeeded (core + streaming + h5coro, one consolidated container-side patch, reproducible via `shim/build.sh all`); footprint ~2.2 MiB read path / ~3–5 MiB bundled.
- [x] Phase 4 — hidefix spike (ran in parallel): full coverage incl. 2-D shuffled int8, native hyperslices, serialized chunk-index cost measured (feeds #148's offset-cache design). Spike-only dependency; not added to zagg.
- [x] Phase 5 — `results/REPORT.md`: canonical variant × {o9, o10} table, decomposition, Lambda footprint per route, recommendation + follow-ups; summary posted on #149.

Also in this PR: a narrow `trailing-whitespace` pre-commit exclusion for `\.patch$` files — the hook otherwise strips diff-context whitespace from the staged patches (`bench/h5coro/patches/`, `shim/sliderule-minimal-build.patch`).

## How tested

The harness is the test: every variant is gated on byte-identical arrays vs the h5coro-1.0.4 baseline before it gets a benchmark row (`--baseline` exits non-zero on mismatch). Capture fidelity checked against the known benchmark truth: the captured o10 worker run reproduces `obs=939217` on shard `5347405532456026122` exactly (cf. the #138 history table). `ruff check` / `ruff format --check` clean; `pytest` unaffected (no package changes).

## Questions for review

- The numpy vectorization patch (Phase 2) targets upstream h5coro; submitting it to SlideRuleEarth/h5coro is outside this repo's push scope, so it's staged under `bench/h5coro/patches/` for hand-off.


bench/h5coro/bench_replay.py
```diff
@@ -0,0 +1,234 @@
...
+            continue
+        path = granule_dir / g["resource"]
+        g0 = time.perf_counter()
+        out = read_granule(str(path), g["calls"])
+        per_granule.append(round(time.perf_counter() - g0, 3))
+        for (ci, ds), arr in out.items():
+            key = f"{g['resource']}:{ci}:{ds}"
+            digest = checksum(arr)
+            checksums[key] = digest
+            if baseline is not None and baseline.get(key) != digest:
```
espg, line 210: 🤖 *from Claude (review)*

The correctness gate only checks keys present in the *current* run — arrays in the baseline that the variant never returned are not detected. Scenario: a phase-3/4 adapter that fails to resolve one dataset and omits it from `out` (or a future adapter bug that skips a call) sails through with `correctness: "pass"` and a smaller `n_arrays`; nothing gates on coverage.

Suggested fix: after the loop, treat missing keys as mismatches:

```python
if baseline is not None:
    mismatches += sorted(set(baseline) - set(checksums))
```

bench/h5coro/bench_replay.py
```diff
@@ -0,0 +1,234 @@
...
+        for ci, entries in enumerate(calls):
+            request = [
+                e["dataset"]
+                if e["hyperslice"] is None
+                else {"dataset": e["dataset"], "hyperslice": [tuple(h) for h in e["hyperslice"]]}
+                for e in entries
+            ]
+            promise = h5.readDatasets(request, block=True)
+            for e in entries:
+                out[(ci, e["dataset"])] = np.asarray(promise[e["dataset"]])
```
espg: 🤖 *from Claude (review)*

If any dataset read fails inside h5coro, this silently checksums garbage instead of failing. h5coro 1.0.4's `datasetThread` catches `RuntimeError`, logs a warning, and returns a *null* `H5Dataset` whose `.values` is `None`; `np.asarray(None)` is a 0-d object array and `checksum()`'s `.tobytes()` on it returns raw *pointer* bytes — nondeterministic per process. With `--write-baseline` that bakes an unreproducible digest into the reference; on gated runs it surfaces as a baffling mismatch with no hint the read failed.

Suggested fix: fail loudly in the adapter:

```python
values = promise[e["dataset"]]
if values is None:
    raise RuntimeError(f"h5coro returned no data for {path}:{e['dataset']}")
```

bench/h5coro/bench_replay.py
```diff
@@ -0,0 +1,234 @@
...
+    ap.add_argument("--adapter", default="h5coro", choices=sorted(ADAPTERS))
+    ap.add_argument("--baseline", default=None, help="checksums JSON to gate against")
+    ap.add_argument("--write-baseline", action="store_true", help="record reference checksums")
+    ap.add_argument("--profile", action="store_true", help="cProfile + bucketed decomposition")
+    args = ap.parse_args()
+
+    if args.profile:
+        prof = cProfile.Profile()
+        prof.enable()
+        run(args)
```
espg: 🤖 *from Claude (review)*

`--profile` runs `run(args)` under cProfile, and `run()` unconditionally writes `replay_{label}_{variant}.json` — so a profile run overwrites the clean benchmark row with instrumented timings at the identical path. The committed artifacts show this already happened: `results/replay_o10_h5coro-1.0.4.json` has `wall_s: 94.596` while `results/profile_o10_h5coro-1.0.4.json` has `total_s: 94.6` (same run; o9 likewise, 139.61 vs 139.616). The macOS "reference" rows are therefore cProfile-instrumented while the linux-arm64 rows are not — the cross-platform comparison in the README is apples-to-oranges.

Suggested fix: in profile mode, skip writing the replay row (or write `replay_..._profiled.json`), and re-record the darwin rows without `--profile`.

bench/h5coro/bench_replay.py
```diff
@@ -0,0 +1,234 @@
...
+        out = read_granule(str(path), g["calls"])
+        per_granule.append(round(time.perf_counter() - g0, 3))
+        for (ci, ds), arr in out.items():
+            key = f"{g['resource']}:{ci}:{ds}"
+            digest = checksum(arr)
+            checksums[key] = digest
+            if baseline is not None and baseline.get(key) != digest:
+                mismatches.append(key)
+        del out
+    wall, cpu = time.perf_counter() - wall0, time.process_time() - cpu0
```
espg: 🤖 *from Claude (review)*

The headline `wall_s`/`cpu_s` include per-array sha256 + baseline-dict work done between granules, while `per_granule_wall_s` excludes it — e.g. the committed linux-arm64 o10 row has `wall_s: 99.272` vs sum(per-granule) 98.36, ~0.9 s of verification cost inside the timed window. That cost is proportional to bytes returned (identical across variants), so it's a constant additive term that systematically understates a fast backend's speedup: a 10x-faster variant reads in ~10 s but reports ~11 s. Under `--profile` the same work also lands in the buckets (`ascontiguousarray` matches `np_convert`; sha256 inflates `other`).

Suggested fix: report `sum(per_granule_wall_s)` as the headline wall (or checksum outside the `wall0` window), so the row measures only the read path.

bench/h5coro/bench_replay.py
```diff
@@ -0,0 +1,234 @@
...
+    """Peak RSS of this process in MB (ru_maxrss is bytes on macOS, KB on Linux)."""
+    rss = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+    return rss / 2**20 if sys.platform == "darwin" else rss / 2**10
+
+
+def checksum(arr: np.ndarray) -> str:
+    h = hashlib.sha256()
+    h.update(arr.dtype.str.encode())
+    h.update(str(arr.shape).encode())
+    h.update(np.ascontiguousarray(arr).tobytes())
```
espg: 🤖 *from Claude (review)*

The digest is over `dtype.str` + raw bytes with no byte-order canonicalization, so the gate tests *representation* equality, not value equality. Scenario for phases 3/4: h5coro materializes arrays in the file's byte order, while a compiled backend (hidefix, or a shim that converts) may hand back native-order arrays — identical values, different `dtype.str` and bytes → hard CORRECTNESS FAILURE with no row. Fine for ATL03 (little-endian files on little-endian hosts), but the harness is meant to gate heterogeneous backends.

Suggested fix: normalize before hashing, e.g. `arr = arr.astype(arr.dtype.newbyteorder("="), copy=False)`.

bench/h5coro/capture_requests.py
```diff
@@ -0,0 +1,151 @@
...
+        },
+        "capture_run": {
+            "wall_s": round(wall, 3),
+            "cpu_s": round(cpu, 3),
+            "max_rss_mb": round(max_rss_mb(), 1),
+            "total_obs": int(meta["total_obs"]),
+            "cells_with_data": int(meta["cells_with_data"]),
+        },
+        "granules": [
+            {"resource": name, "calls": calls_by_resource.get(name, [])} for name in names
```
espg, line 147: 🤖 *from Claude (review)*

Calls recorded under a resource key that doesn't match a shardmap basename are silently dropped here (`.get(name, [])`), and `bench_replay.py` silently skips granules with empty `calls` — so a capture-side naming drift (e.g. re-capturing with a driver whose URL rewrite changes `H5Coro.resource`, or an h5coro version that stores it differently) shrinks the frozen workload to nothing while the script still prints "captured N calls" from the recorder's own count. Today's committed captures are consistent (meta `n_calls` == sum over granules), but nothing guards the invariant.

Suggested fix: assert no unmatched keys before writing, e.g. `assert not (set(calls_by_resource) - set(names))`.

espg:
🤖 *from Claude*

Folded all six phase-1 self-review findings in b88586b ("fold review: harness gate/timing fixes + re-measured baselines"):

1. **Omission hole in the gate** — replay now appends `missing:<key>` mismatches for every baseline key absent from the run, so a partially-failing adapter can't score `pass`.
2. **`None` reads** — the h5coro adapter raises on a null dataset result instead of digesting `np.asarray(None)` garbage.
3. **`--profile` overwrote clean rows** *(confirmed: the originally committed macOS rows were instrumented)* — profiled runs no longer write replay rows, and every timed row in `results/` was re-measured clean (the linux/arm64 numbers moved ≤6%: o10 99.3→93.5 s, o9 134.2→129.0 s, which also reflects the timing fix below).
4. **Checksum cost inside the timed window** — only the read windows are timed now; gate cost is reported separately as `gate_s`.
5. **Endianness canonicalization** — digests are computed over little-endian-canonical bytes (no-op for the existing references; matters for phase-3/4 backends).
6. **Capture/replay invariants** — capture now fails if a recorded resource doesn't match a shardmap basename and reports zero-call granules; replay rows carry `skipped_granules`.

One review note was out of PR scope and stays standing for you: the reviewer traced `src/zagg/processing/read.py` L651/L659 passing dicts *without* a `"hyperslice"` key to `readDatasets`, which `H5Promise.__init__` in h5coro 1.0.4 indexes unconditionally — if that coarse-filter path is exercised it would `KeyError` and be swallowed by the broad `except` in `worker.py`. Worth a quick check / its own issue if it reproduces.

bench/h5coro/patches/0001-vectorize-shuffleChunk-as-a-numpy-byte-matrix-transp.patch
```diff
@@ -0,0 +1,41 @@
...
+         shuffle_block_size = int(len(input) / type_size)
+         elements_to_shuffle = int(output_size / type_size)
+         start_element = int(output_offset / type_size)
+-        for element_index in range(start_element, start_element + elements_to_shuffle):
+-            for val_index in range(0, type_size):
+-                src_index = (val_index * shuffle_block_size) + element_index
+-                output[dst_index] = input[src_index]
+-                dst_index += 1
++        blocks = numpy.frombuffer(input, dtype=numpy.uint8, count=type_size * shuffle_block_size)
++        window = blocks.reshape(type_size, shuffle_block_size)[:, start_element:start_element + elements_to_shuffle]
```
espg, line 37: 🤖 *from Claude (review)*

Bug-compatibility divergence in the out-of-range case. The original loop indexes `input[val_index * shuffle_block_size + element_index]` directly, so a window extending past `shuffle_block_size` (e.g. `inflateChunk` returning fewer bytes than `dataChunkBufferSize` while the caller still requests `chunk_bytes` for the full chunk) raises `IndexError` — a loud failure. The numpy slice `[:, start_element:start_element + elements_to_shuffle]` silently **clamps**, and the tail-padding branch below then fills the shortfall with zeros — the same condition now returns silently corrupted (zero-filled) data.

Verified by simulation against the original loop: byte-identical for all in-range cases (ragged `output_size`, unaligned `output_offset`, `len(input)` not divisible by `type_size`, zero elements — the zero-padding tail is bug-compatible with `bytearray(output_size)`), diverging only on over-range windows:

```
orig(bytes(range(10)), 0, 16, 2)  ->  IndexError
new (bytes(range(10)), 0, 16, 2)  ->  00050106020703080409000000000000  (silent zero tail)
```

Fix: guard the invariant before slicing, e.g. `if start_element + elements_to_shuffle > shuffle_block_size: raise FatalError(...)` (matching the function's existing `errorChecking` style), and mirror it in the ~/software/h5coro `numpy-vectorize` branch so the .patch stays in sync.

bench/h5coro/hidefix_spike.py
```diff
@@ -0,0 +1,204 @@
...
+        per_granule.append(
+            {
+                "resource": g["resource"],
+                "index_build_s": round(build_s, 4),
+                "read_s": round(gwall - build_s, 3),
+                "wall_s": round(gwall, 3),
+            }
+        )
+        del out, idx
+    wall, cpu = time.perf_counter() - wall0, time.process_time() - cpu0
```
espg: 🤖 *from Claude (review)*

Timed-window inconsistency with `bench_replay.py` post-fold (b88586b): here `wall_s`/`cpu_s` span the whole loop **including the checksum gate**, while `bench_replay.py` now sums only the per-granule read windows and reports `gate_s` separately. Committed rows show the gap: o10 `wall_s` 15.981 vs `sum(per_granule_wall_s)` 15.15 (~0.83 s of gate inside the window); o9 23.979 vs ~22.9 (~1.05 s). So the hidefix headline `wall_s` is not the same quantity as the h5coro rows it is compared against (~5% at these magnitudes), and the row also lacks the `gate_s`/`skipped_granules` fields the other replay rows now carry.

Fix: hoist the checksum loop out of the timed window and accumulate `gate_s` exactly as `bench_replay.py` does (headline `wall` = sum of per-granule read windows), or failing that, state the convention difference in the results row and notes so nobody builds a cross-variant table from mismatched windows.

bench/h5coro/results/hidefix_spike_notes.md
```diff
@@ -0,0 +1,161 @@
...
+## Replay results (linux-arm64 container = primary rows)
+
+| variant | requests | wall_s | cpu_s | max_rss_mb | correctness |
+|---|---|---|---|---|---|
+| h5coro-1.0.4-linux-arm64 (baseline) | o10 | 99.3 | 103.9 | 256 | pass |
+| **hidefix-0.12.0-linux-arm64** | o10 | **16.0** | 16.4 | 155 | pass |
+| h5coro-1.0.4-linux-arm64 (baseline) | o9 | 134.2 | 138.9 | 301 | pass |
+| **hidefix-0.12.0-linux-arm64** | o9 | **24.0** | 23.1 | 162 | pass |
+
+≈ **6.2× (o10) / 5.6× (o9)** faster than h5coro 1.0.4, at ~55% of the RSS.
```
espg: 🤖 *from Claude (review)*

This table cites the **superseded phase-1 baselines**, not the re-measured rows committed in b88586b. Committed `replay_o10_h5coro-1.0.4-linux-arm64.json` now says wall 93.467 / cpu 98.182 / rss 248.1 — the table's 99.3 / 103.9 / 256 are exactly the phase-1 values that commit replaced. Same for o9 (committed 129.024 / 133.615 / 282.5 vs table 134.2 / 138.9 / 301), and the macOS reference figures below ("94.6 s / 139.6 s") are the phase-1 darwin walls (committed rows: 84.755 / 116.544).

Against the committed post-fold rows the speedups are ≈**5.8×** (o10) and ≈**5.4×** (o9), not 6.2× / 5.6× — and the gap widens slightly further because the hidefix walls still include ~0.8–1.0 s of checksum-gate time that the h5coro walls now exclude (see comment on `hidefix_spike.py`). Fix: recompute the table and the derived speedup/RSS claims from the rows actually committed on this branch.

bench/h5coro/Containerfile.hidefix
```diff
@@ -0,0 +1,33 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal --default-toolchain 1.96.0
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# python extension (maturin build backend resolved via pip build isolation)
+RUN pip install --no-cache-dir --no-binary hidefix "hidefix==0.12.0"
+
+# index-serialization CLI (not part of the wheel; needed for #148 numbers)
+RUN cargo install hidefix --version 0.12.0 --features clap,bincode,flexbuffers \
```
espg: 🤖 *from Claude (review)*

`cargo install hidefix --version 0.12.0` without `--locked` resolves transitive crate versions at image-build time, so rebuilding this container later yields a different `hfxidx` binary. The notes already observed the symptom — ≈1.1% bincode index-size drift for the same granule between the macOS and linux builds (`hidefix_spike_notes.md`, "Curiosity" bullet) — and themselves recommend `--locked`; the Containerfile should apply it. The `pip install --no-binary hidefix` step above has the same class of drift (build isolation pulls the latest maturin at build time); a build-backend pin (e.g. via `PIP_CONSTRAINT`) is worth it if the wheel build is ever treated as reproducible.

espg:
🤖 *from Claude*

Folded the four phase-2/4 review findings in d58ac1d ("fold review: shuffle clamp guard + spike gate/timing parity"):

1. **Out-of-range shuffle window silently clamped** — the numpy `shuffleChunk` now raises `FatalError` when the requested window exceeds `shuffle_block_size`, staying loud like the `IndexError` of the loop it replaces (clone commit amended; `patches/0001-*` regenerated).
2. **Spike timing window mismatch** — `hidefix_spike.py` now times only index-build + reads and reports `gate_s` separately, matching `bench_replay.py`; it also gained the missing-key gate and `skipped_granules` for full row parity.
3. **Stale baselines in the spike notes** — table recomputed against the post-fold baselines (93.5/129.0 s): hidefix is **5.8× (o10) / 5.4× (o9)**, not 6.2×/5.6×. The hidefix rows themselves still carry ~1 s of gate cost and will be re-measured once with the aligned window in the phase-5 canonical pass.
4. **Non-reproducible builds** — `Containerfile.hidefix` now uses `cargo install --locked` (the notes' own 1.1% bincode drift was the symptom) and pins `maturin==1.14.1` with `--no-build-isolation` so the pin actually applies to the sdist build.

bench/h5coro/shim/sliderule-minimal-build.patch
```diff
@@ -0,0 +1,136 @@
+sliderule minimal-build workarounds for the zagg h5coro benchmark shim
+(issue #149 phase 3). Applies with `patch -p1` at the sliderule source root;
+generated against sliderule v5.4.3 (ce4be309, 2026-06-16).
+
+Benchmark-only change:
+  * Asset.h — the Asset constructor and attributes_t sit in the classs single
```
espg: 🤖 *from Claude (review)*

`classs` here trips codespell (`classs ==> class`), and the codespell pre-commit hook only skips `fixture,*.ipynb,archive` — it does not share the trailing-whitespace hook's new `\.patch$` exclude. Verified with codespell 2.4.1: this line is flagged, so `pre-commit run --all-files` (the local CI mirror required by CLAUDE.md §4) fails on this commit.

Fix: this is free-text header prose, not diff content, so it's safe to edit — change to `class's` (which also fixes the missing apostrophe). No need to widen the codespell skip list.

bench/h5coro/Containerfile.hidefix
```diff
@@ -0,0 +1,38 @@
...
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential cmake curl pkg-config \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --profile minimal --default-toolchain 1.96.0
+ENV PATH="/root/.cargo/bin:${PATH}"
+
+# python extension (maturin build backend resolved via pip build isolation;
+# pinned so the sdist build is reproducible)
+RUN pip install --no-cache-dir "maturin==1.14.1" \
```
espg: 🤖 *from Claude (review)*

The pin contradicts the sdist's declared build contract: hidefix 0.12.0's `pyproject.toml` has `build-system.requires = ["maturin>=1.3,<1.4"]`, and `--no-build-isolation` makes this `maturin==1.14.1` the one that runs — a version the package explicitly excludes. I verified the build does still succeed (maturin 1.x doesn't enforce the bound, unlike 0.x major mismatches), so this isn't broken today, but "pinned so the sdist build is reproducible" is achieved with a maturin the upstream never tested; a future maturin that resumes enforcing `build-system.requires` (or changes wheel behavior) breaks the image silently.

Fix: pin inside the declared range instead — `maturin==1.3.2` gives the same reproducibility while staying in-contract (matches what pip's build isolation would have resolved).

bench/h5coro/shim/build.sh
```diff
@@ -0,0 +1,144 @@
...
+#   bash bench/h5coro/shim/build.sh replay     # gated rows for o10 + o9
+#
+# Env overrides: SLIDERULE_SRC (default ~/software/sliderule), GRANULE_DIR
+# (default ~/ignore/zagg_neon_atl03_test_shard/granules), SHIM_BUILD_DIR
+# (writable scratch for the out-of-tree build; default $TMPDIR/zagg-shim-build).
+set -euo pipefail
+
+HERE="$(cd "$(dirname "$0")" && pwd)"     # bench/h5coro/shim
+REPO="$(cd "$HERE/../../.." && pwd)"      # zagg repo root
+SLIDERULE_SRC="${SLIDERULE_SRC:-$HOME/software/sliderule}"
```
espg, line 19: 🤖 *from Claude (review)*

Reproducibility gap (document-level, no code defect): nothing checks that `$SLIDERULE_SRC` is actually at the commit `sliderule-minimal-build.patch` was generated against (v5.4.3 / `ce4be309` per its header), and the `tar --exclude=.git` copy in `cmd_build` removes any in-container way to tell. On a drifted clone, `patch -p1` can apply with fuzz/offsets and still exit 0, so `build.sh replay` would emit rows silently attributed to a sliderule version that was never built. (I verified the patch applies hunk-clean at `ce4be309`.)

Related: `cmd_image`/`build.sh all` assumes the `zagg-bench-h5coro` base image already exists (built from `bench/h5coro/Containerfile`); on a fresh machine podman fails resolving the `FROM` with no hint where it comes from.

Fix: a host-side guard before `podman run`, e.g. `[ "$(git -C "$SLIDERULE_SRC" rev-parse --short=8 HEAD)" = ce4be309 ] || echo "warning: sliderule clone is not at ce4be309 (patch base)" >&2`, plus one header line noting the base image must be built first (`podman build -t zagg-bench-h5coro -f bench/h5coro/Containerfile bench/h5coro`).

espg:
🤖 *from Claude*

Folded the three phase-3 review findings in 42ca79c:

1. **`classs` typo in the shim patch header** — fixed (header prose only; `git apply --check` re-verified against the sliderule clone). This was also the one thing tripping codespell.
2. **maturin pin out of the sdist's build contract** — repinned to `maturin==1.3.2`, inside hidefix 0.12.0's declared `maturin>=1.3,<1.4`.
3. **`build.sh` reproducibility guards** — it now refuses to run if the mounted sliderule clone isn't at the patch's base rev (`ce4be309`, overridable via `SLIDERULE_REV=`) and fails with a clear message if the `zagg-bench-h5coro` base image is missing.

All phases are complete, all review findings across the three passes folded. Marking ready for review once CI is green.

espg:
merge to bench to continue track (and close)

Issue #151, opened 2026-07-01
Warm start lambda and memory freeing are *still* a problem


espg:
This is getting super annoying. Most recent workflow push on main failed: https://github.com/englacial/zagg/actions/runs/28549136504/job/84642556847

This was (again) on a push that was 'supposed' to fix this for us. Related issues / PRs so far:

 #137 / #146 / #141 / #145 / #144 / #143 / #140 / #139 

...probably others too that I've missing (check past week of new issues)

Can we do a deep dive on this is figure out a real solution so we don't keep failing in the same place?

espg:
🤖 *from Claude*

**Recommendation up front:** I don't think the recurring benchmark failures are memory at all anymore. The evidence points to the **synchronous Lambda invoke connection from the GitHub runner being severed at ~4 minutes idle** (Azure SNAT idle timeout on hosted runners; botocore doesn't enable TCP keepalive). Every failing target dies with `Connection was closed before we received a valid response` at ~250–290 s after dispatch; every target that finishes in ≲245 s passes — regardless of memory. The memory work (#139/#140/#143/#144) was real but is a separate track, and the #144 fix likely never reached the deployed function anyway (details below).

## What the run history actually shows

I pulled the logs for every Lambda Benchmark run of the current 6-target matrix (Lambda `runtime_s` in parens for passes):

| run (merge) | dispatch | o11_sharded | o11_inner | o10_sharded | o10_inner | o9_sharded | o9_inner |
|---|---|---|---|---|---|---|---|
| [28472721816](https://github.com/englacial/zagg/actions/runs/28472721816) (#101) | sequential | 792 MB ok | 1391 MB ok | 2013 MB ok (247 s) | fail | fail | fail |
| [28494574307](https://github.com/englacial/zagg/actions/runs/28494574307) (#136) | sequential | 802 MB ok | 1373 MB ok | fail | fail | fail | fail |
| [28502909333](https://github.com/englacial/zagg/actions/runs/28502909333) (#139/#140 in) | sequential | 808 MB ok | 1354 MB ok | fail | fail | fail | fail |
| [28542318475](https://github.com/englacial/zagg/actions/runs/28542318475) (#143/#144 in) | sequential | 798 MB ok | 1364 MB ok | fail | fail | fail | fail |
| [28544792919](https://github.com/englacial/zagg/actions/runs/28544792919) (#142) | sequential | 795 MB ok | 1378 MB ok | 1968 MB ok (244 s) | fail | fail | fail |
[...]

PR #152, opened 2026-07-01
88S stress shard: pins, 900s timeout, chunk-boundary extraction, a-priori read plan, streaming tdigest (issue #148)

espg:
Refs #148. PR #150 merged 2026-07-02 (into `bench`; benchmark-only, no `src/zagg` changes — the issue decisions hold the production reader constant), unblocking phase 3, which is now complete. Reconciled with `main` twice since: the issue #151 async-result channel and the issue #12 temporal pipeline (PR #70) — both merged in with the extract/a-priori behavior unchanged. Phase 5 remains gated on deployment (yours).

88S stress-test infrastructure: pinned worst-case shards over the ICESat-2 turning-latitude ring, the 720→900 s timeout bump, chunk-boundary geometry extraction, the faked-geometry a-priori read plan (arm 2a), and the streaming buffered tdigest merge — per the [plan](https://github.com/englacial/zagg/issues/148#issuecomment-4860223344) and the [review decisions](https://github.com/englacial/zagg/issues/148#issuecomment-4860294338) on it: **one pinned test location** (extraction at o9, nesting o10), **NEON temporal window** (no per-target temporal override), **δ adaptive** (512 → 128 on OOM), **byte offsets deferred** pending the cost estimate + inferability findings (posted on #148), and **#149 assumed to land before any timed runs**.

## Phases

Phase numbering follows the plan; phase 1 landed first because phase 0&#39;s shardmap build runs against live CMR (~20 min fetch of 35,639 granules).

- [x] **Phase 1 — timeout 720→900**: `template.yaml` `Timeout` default (900 is the Lambda ceiling; no 720/900 variants per the decisions), `runner._DEFAULT_FUNCTION_TIMEOUT_S` mirror, doc references. Review fold: dispatch client `read_timeout` 900→960 so a shard running to the new ceiling surfaces as the Lambda&#39;s own timeout, not a retryable client-side read timeout ([thread](https://github.com/englacial/zagg/pull/152#discussion_r3509377406)). The **baseline timed run** of approach (1) is *not* here: the 900 s timeout isn&#39;t deployed (deploys are yours).
- [x] **Phase 0 — 88S catalog, shardmaps, pinned stress shard.**
  - Catalog: ATL03 v007, 2018-10-13 → 2025-06-01 (NEON window), ring −88° … −87.7°: **35,639 granules**.
  - **o9 pin `11530494877603201033` = 5,620 granules** (ring: 576 shards, 2.04M pairs; median shard 3,192) — ~95× NEON o9&#39;s 59.
  - **o10 pin `11530494877603201034` = 4,605 granules**, the densest o10 shard **nested inside the o9 pin** (`nested_in` key; global o10 max is 4,642 at another longitude) so one o9 extraction pass covers both orders per the decisions. The drift test derives the same nested quantity when `nested_in` is set (`_containing_shard`).
  - **Pruned shardmaps**: the full ring maps are ~0.7 GB (o9) / ~1.7 GB (o10) of JSON, so the committed maps keep only the pinned shard (~1.9/1.6 MB, provenance in `metadata.pruned`). The benchmark only ever dispatches the pinned shard. Committing them was confirmed in [review](https://github.com/englacial/zagg/pull/152#issuecomment-4880908345).
  - **AOI geojson fix**: the shipped `antarctic_88s.geojson` (one lat/lon rectangle spanning −180…180) collapses to an antimeridian sliver under mortie&#39;s spherical polygon fill — 10 cells instead of 576 at o9. Replaced with a MultiPolygon of eight 45° sectors (vertices every 1° of longitude); README documents the pattern.
  - Targets `tdigest_healpix_o{9,10}_88s` are **provisional** (PR-tree-only, explicit `--target`): an expected OOM/timeout must not turn the every-merge benchmark matrix red now that #145&#39;s fail-on-empty is live. `test_targets_manifest_consistent` now covers provisional pins; the drift test&#39;s config lookup searches provisional targets too.
- [x] **Phase 2 — chunk-boundary extraction.**
  - `src/zagg/catalog/extract.py`: per-beam chunk-boundary (first/last photon) lat/lon extraction streaming in 16-chunk blocks (memory flat ~25 MB), per-granule parquet via fastparquet (layer-safe, no pyarrow) with provenance/cost metadata, `run_extraction` batch body with per-granule failure isolation and `wall_s` cost capture.
  - Lambda: **`mode=&#34;extract&#34;` on the existing function AND a dedicated `ExtractFn` twin** in `template.yaml` — per [review](https://github.com/englacial/zagg/pull/152#issuecomment-4880908345) (&#34;Can we do both?&#34;): the mode rides the existing deployment for incremental updates; the twin gives full-archive extraction its own concurrency/throttling pool. Same code zip/layer/role/memory and the shared 900 s `Timeout`, plus a lockstep `EventInvokeConfig` (retries 0, event age 60); template tests pin both twins against `ProcessFn`.
  - Schema verification + inferability + cost findings posted on #148 (chunk grids aligned; byte offsets **not** inferable; ~$75–120 extrapolated for the full 500k-granule catalog, ~$1 for the pinned shard&#39;s 5,620 granules).
  - Found and worked around an **h5coro B-tree off-by-one**: any hyperslice starting exactly at `k * chunk_size` (k&gt;0) fails the whole dataset read (`h5dataset.py` start-edge intersection uses `&lt;` where the half-open node slice needs `&lt;=`). The extractor&#39;s block reads start one element early; details on #148 — this can also silently drop granules on today&#39;s planned-read path and is relevant to #149.
  - Tests: extraction stub emulates the h5coro surface *including* the aligned-start bug (regression-guards the workaround), parquet round-trip, alignment guard, handler event contract (creds mapping, partial-failure → 500, `block_chunks` forwarding). Validated against real local granules (values verified against a full lat/lon read).
- [x] **Phase 3 — faked-geometry a-priori read plan (arm 2a).** Boundary parquet → per-beam chunk polylines → chunk-aligned `ReadPlan`, no geolocation-rate read and no segment→shard mask.
  - `src/zagg/processing/apriori.py`: loads the granule&#39;s phase-2 parquet (local dir or `s3://` prefix), interpolates each chunk&#39;s boundary photons on the unit sphere (pole/antimeridian-safe, unlike lat/lon lerp) and applies the **same mortie test the photon path uses** (`grid.shards_of(grid.assign(...)) == shard_key`); matched chunks feed the existing `plan_read` as coarse parents, so run merging, `pad` (whole chunks, default 1), and the `full_read_threshold` selectivity fallback are shared with the planned path — and the slices are chunk-aligned by construction.
  - Correctness does not rest on the faked geometry being exact: the plan only bounds IO, and the shared executor applies the exact photon-level shard mask + filters, so **output is bit-identical** as long as no touched chunk is missed (interpolated samples + one-chunk pad are the conservative margin).
  - Seam: `data_source.read_plan.chunk_boundaries: {prefix, samples_per_chunk}` routes `_read_group` to the a-priori path (a **presence** check with precedence over `spatial_index`, so one benchmark config selects the arm with a single key, and an empty/misconfigured block fails loudly instead of silently running another arm). The shared back half of `_planned_read_group` is refactored into `_execute_plan_group` (pure relocation) — both arms run identical post-plan semantics. **Flag off is byte-identical**: the worker passes the new `granule_url` kwarg only when the feature is on, so the production call and every monkeypatched `_read_group` fake keep their existing signature.
  - The phase-2 **h5coro aligned-start bug bites this arm by construction** (every a-priori slice starts exactly on a chunk boundary): the arm&#39;s read callback starts one element early for any nonzero start (unconditional — no chunk-size metadata dependence) and drops it, mirroring the extractor&#39;s workaround.
  - Tests (`tests/test_apriori.py`): **bit-identity against both production paths** (planned and full) on a NEON-shaped local shard — the stub *enforces* the aligned-start bug, so identical output proves the workaround; the shard band sits strictly inside one chunk with both boundary photons outside it, so it also proves interior-crossing detection. Plus: IO bounding (zero geolocation reads; hyperslices exactly the padded chunks), antimeridian sampling safety, plan alignment/merge/pad/truncated-final-chunk, selectivity fallback parity, empty-block/missing-parquet/URL/prefix failure modes, the `samples_per_chunk &gt;= 2` guard, s3 staging, and both worker-seam directions. Real granules aren&#39;t in the tree, so the stub mirrors the NEON test-shard layout exactly as the phase-2 extraction tests do.
- [x] **Phase 4 — streaming buffered tdigest merge.** `aggregation.streaming: {buffer_granules: B}` (default 50) gates a bounded-buffer worker path (`src/zagg/processing/streaming.py`): reads accumulate for B granules, are grouped by the same `_concat_and_group`, folded into running per-cell state (`len` by summation — exact; tdigest via `merge_tdigests` under the field&#39;s own δ), and released — peak memory is one buffer + digest state, independent of granule count. Validated up front: only mergeable reducers may stream (expressions, vectors, `resolution: chunk`, `chunk_precompute` raise); the real benchmark tdigest configs validate as-is. **When the shard fits in one buffer the output is byte-identical to pooled** (asserted in tests). δ/flush study (1M-obs bimodal+tail synthetic, max abs error over q ∈ {.01,.05,.25,.5,.75,.95,.99}): δ=512 one-shot 0.026 → 0.027–0.066 at 4–200 flushes; δ=128 one-shot 0.21 → 0.22–0.35. So merge-round count barely moves accuracy — B should be sized for merge CPU (~10 µs/touched cell/flush) and buffer memory, and δ per the adaptive 512→128 plan.
- [ ] **Phase 5 — 88S benchmark matrix + go/no-go** — gated on **deployment of the 900 s timeout** (deploys are yours; #150 merged 2026-07-02, so the reader-held-constant precondition is met). Matrix per the plan: {(1), (2a)} × {pooled, streaming} on the pinned o10 then o9 shard, plus the cheap (2a)+on-the-fly-offsets arm noted on #148.

## Review folds

- Phase-3 fresh-context review ([review](https://github.com/englacial/zagg/pull/152#pullrequestreview-4629111169)) folded in df44f60: presence-gated arm-2a dispatch (empty block fails loudly), `ExtractFnAsyncConfig` completing the twin, `samples_per_chunk >= 2` guard, unconditional aligned-start workaround (no chunk-size metadata dependence) with the single-range assumption pinned, and the extract × `result_url` seam test.
- espg review comment ([2026-07-04](https://github.com/englacial/zagg/pull/152#issuecomment-4880908345)): dedicated `ExtractFn` twin added alongside `mode="extract"` (&#34;Can we do both?&#34; — yes, done); drift rebuild now supports a **build-once catalog snapshot** (`catalog_parquet` per-entry key: the drift test rebuilds from the committed stac-geoparquet instead of re-fetching CMR — see Questions (1) for the one remaining step); pruned shardmaps stay committed as confirmed.
- All ten phase-0/2/4 adversarial-review findings folded earlier (b1075b6, e81a923, c62ea8e) — every thread has a fold commit + reply.

## Merge state

- 5b752e1 reconciled the issue #151 async-result channel (extract-mode dispatch placed before the `result_url` mirror).
- e4e0874 reconciled the issue #12 temporal pipeline (PR #70): the mode dispatch now orders setup → finalize → **extract (returns directly)** → process_event/process (both feeding the shared `result_url` mirror). Extract stays outside the mirror by design — `test_extract_mode_ignores_result_url` pins it against the restructured block. `main`&#39;s issue #157 h5coro full-read-form fix in `_read_group_full` auto-merged cleanly around the `_execute_plan_group` refactor (the a-priori/planned paths use the string full-read form, unaffected). Extract/a-priori behavior unchanged.

## Testing

- `pytest -v` fully green locally after the temporal reconcile (1,286 passed, 26 skipped — includes the merged-in `test_temporal.py`; this environment&#39;s `python3` is 3.12, so the earlier local-only `test_function_build_succeeds` failure doesn&#39;t reproduce).
- ruff at the pinned pre-commit rev (0.14.10): clean on all touched files (a newer local ruff 0.15 false-flags several untouched files; CI and pre-commit pin 0.14.10). One pre-existing `N818` in `src/zagg/registry.py` (`UnknownCapability`), untouched here — the CI lint bot runs `--select=E,F,W,I` so it has never fired there; renaming a public exception is out of scope for this PR.
- Phase 3: `tests/test_apriori.py` (20 tests) — bit-identity vs planned AND full paths under a bug-enforcing stub, IO bounding, worker-seam both directions; full `tests/test_processing.py` unchanged-path suite green after the `_execute_plan_group` refactor.
- Extraction validated on 8 real granules (0.6–2.5 GB): boundary values match a full lat/lon read exactly; 0.54–2.46 s/granule locally.

## Questions for review

1. **88S catalog snapshot (build-once, per your comment).** The `catalog_parquet` mechanism is in (drift test + README): point the 88S shardmap entries at a committed stac-geoparquet snapshot and the weekly job stops re-fetching CMR. Generating the snapshot itself needs one CMR pass (~20 min, 35,639 granules), which this environment can&#39;t reach. Options: (a) you run the README&#39;s `python -m zagg.catalog ... --catalog-out tests/data/benchmark/catalogs/cat_88s.parquet` once and commit it (or attach it for me to commit), or (b) a network-enabled routine run builds and commits it to this branch next pass. Size should be O(10 MB); if it lands much larger, release-asset instead of committing.
2. **Arm-2a defaults**: `samples_per_chunk=64` (≈0.3–1.1 km sample spacing at the ATL03 100k chunk size, vs ~6.4 km o10 cells) and `pad=1` whole chunk. Both config-tunable; say the word if you want different defaults for the 88S matrix.


src/zagg/runner.py
```diff
@@ -998,8 +998,9 @@ def _log_concurrency_report(report: ConcurrencyReport, max_workers: int) -> None
 
 # Function Timeout fallback when get_function_configuration can't be read
 # (permission denied, etc.). Mirrors the CloudFormation default in
-# deployment/aws/template.yaml (Timeout Default: 720).
-_DEFAULT_FUNCTION_TIMEOUT_S = 720
+# deployment/aws/template.yaml (Timeout Default: 900, the Lambda hard
+# ceiling — bumped from 720 for the 88S stress shards, issue #148).
+_DEFAULT_FUNCTION_TIMEOUT_S = 900
```
espg, line 1614: 🤖 *from Claude (review)*

The dispatch boto client is built with `read_timeout=900` (`runner.py` ~line 696, unchanged by this PR):

```python
boto_config = Config(
    read_timeout=900,
    connect_timeout=10,
    retries={"max_attempts": 0},
    ...
)
```

That 900 was sized with **180 s of headroom over the old 720 s function timeout** — a timed-out invoke returned its `"Task timed out"` `FunctionError` at ~720 s, well inside the client's socket-read window. With the function timeout now also 900 s, the two race, and the client side will typically lose: a shard that runs to the Lambda ceiling — exactly the 88S stress case this bump exists for — trips a botocore read-timeout exception instead of ever seeing the `FunctionError`.

Two consequences in `_invoke_lambda_cell`:

1. The exception message matches `"Read timeout"` in the retryable list (~line 1205), so a **deterministic** timeout gets re-invoked up to `max_retries` times, each billed for another full ~900 s run — the exact behavior the #119 comment above the `FunctionError` branch says timeouts must not do.
2. The result is never tagged `timeout=True`, so benchmark/summary accounting (`worker_pct_timeout`, timeout counts per issue #100/#119) misclassifies the stress-shard failure mode as a generic client error.

Suggest bumping `read_timeout` to keep headroom above the function timeout (e.g. 930–960 s) in this same phase, since the constant here explicitly mirrors the deployed timeout.

espg: 🤖 *from Claude*

Folded in b1075b6: `read_timeout` 900→960 with a comment tying the headroom to the function `Timeout` ceiling, so a shard that runs to the 900 s Lambda cap surfaces as the Lambda's own "Task timed out" `FunctionError` (non-retryable accounting intact) instead of a retryable client-side read timeout.

src/zagg/catalog/extract.py
```diff
@@ -0,0 +1,291 @@
...
+    h5coro's B-tree intersection test is off by one at the *start* edge, so a
+    hyperslice beginning at ``k * chunk_size`` (k > 0) spuriously matches the
+    preceding chunk and the whole read fails with ``invalid location to read
+    chunk``. Starting one element early (and skipping it) sidesteps the bug at
+    the cost of re-decompressing one chunk per block.
+    """
+    n, chunk = heights_chunk_grid(h5obj, beam)
+    cols = ("beam", "chunk", "start_lat", "start_lon", "end_lat", "end_lon", "n_photons")
+    if n == 0:
+        return pd.DataFrame({c: [] for c in cols})
```
espg: 🤖 *from Claude (review)*

**Empty-beam granules produce a drifting parquet schema.** `pd.DataFrame({c: [] for c in cols})` carries no dtypes — under pandas 2.2.3 every column comes out `float64` — so:

- a granule where *any* beam has 0 photons: `pd.concat` at line 158 promotes `chunk`/`n_photons` to **float64** (granules with all-non-empty beams write `int64`);
- an all-empty granule: `beam` is written as a **DOUBLE** parquet column instead of a string column.

Verified against this branch with the test stub:

```python
extract_granule_boundaries(_StubH5({"gt1l": _beam(0), "gt2r": _beam(200)}), "g.h5")[0].dtypes
# chunk: float64, n_photons: float64   (int64 when no beam is empty)
extract_granule_boundaries(_StubH5({"gt1l": _beam(0)}), "g.h5")[0].dtypes
# beam: float64 -> fastparquet writes a DOUBLE `beam` column
```

Weak beams with zero photons do occur in real ATL03, so the phase-3 consumer reading these parquets back will hit per-file schema mismatches and float chunk indices. Fix: build the empty frame with explicit dtypes (a typed empty template shared by the two call sites here and at line 158, e.g. `.astype({"beam": str, "chunk": "int64", "n_photons": "int64", ...})`), and add a test that writes an empty-beam granule's parquet and asserts dtypes.

src/zagg/catalog/extract.py
```diff
@@ -0,0 +1,291 @@
...
+        try:
+            h5obj = h5coro.H5Coro(
+                rewrite(url), h5coro_driver, credentials=credentials, errorChecking=True
+            )
+            df, meta = extract_granule_boundaries(h5obj, granule_id, block_chunks=block_chunks)
+            name = f"{granule_id.removesuffix('.h5')}.boundaries.parquet"
+            local = os.path.join(scratch_dir, name)
+            write_boundaries_parquet(df, meta, local)
+            dest = _put_parquet(local, output_prefix, name)
+            os.remove(local)
```
espg: 🤖 *from Claude (review)*

`os.remove(local)` only runs on the success path — if `_put_parquet` raises (S3 throttle, wrong prefix, missing PutObject perms), the `except` records the failure and moves on, leaving the scratch parquet behind. Lambda's `/tmp` (512 MB default) persists across warm invocations, so a systematic upload failure during the 5,620-granule fan-out accumulates orphaned parquets until later batches start failing with ENOSPC — one misconfiguration cascading into unrelated per-granule errors. Wrapping write→upload in `try/finally` (unlink in the `finally`, guarded by `os.path.exists`) keeps the scratch dir clean on every outcome.

deployment/aws/lambda_handler.py
```diff
@@ -230,16 +241,75 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
...
+        error_msg = f"Missing required parameters: {', '.join(missing)}"
+        logger.error(error_msg)
+        return {"statusCode": 400, "body": json.dumps({"error": error_msg, "mode": "extract"})}
+
+    driver = event.get("driver", "s3")
+    s3_creds = event["s3_credentials"]
+    if driver == "https":
+        credentials = s3_creds.get("edl_token", s3_creds)
+    else:
+        credentials = {
```
espg, line 376: 🤖 *from Claude (review)*

**Extract mode skips the per-key credential validation process mode performs, and the resulting failure is expensive and opaque instead of a fast 400.** Process mode rejects an event whose `s3_credentials` is missing `accessKeyId`/`secretAccessKey`/`sessionToken` with a 400 (lines 410–414). Here `s3_credentials` need only *exist*: missing keys become `None` values, and because the snake_case keys are then still *present* in the mapped dict, h5coro's `create_session` takes its `"aws_access_key_id" in credentials` branch and builds `boto3.Session(aws_access_key_id=None, ...)` — which silently falls through to the default chain, i.e. the execution role. Every NSIDC read then 403s, so the caller pays for the whole invocation and gets a 500 with N per-granule errors instead of an immediate 400 naming the missing keys.

Suggest mirroring process mode's `required_cred_keys` check for the s3 driver. Minor related note: `worker.py` accepts snake_case fallbacks in this same mapping (`s3_credentials.get("accessKeyId") or s3_credentials.get("aws_access_key_id")`, worker.py lines 206–212) while this one is camelCase-only — fine if the validated contract is camelCase, but the module docstring's "same shape as process mode" is currently enforced weaker than process mode itself.

tests/test_benchmark_shardmap.py
```diff
@@ -52,9 +52,32 @@ def resolve_aoi_temporal_cmr(sm_meta: dict) -> tuple[dict, dict, dict]:
...
+    HEALPix cells nest, so a finer cell's center maps unambiguously into its
+    containing coarser cell; routing through ``assign``/``shards_of`` keeps this
+    on the same mortie machinery the shard maps themselves are built with.
+    """
+    import numpy as np
+    from mortie import mort2geo
+
+    lat, lon = mort2geo(np.array([shard_key], dtype=np.uint64))
+    leaf = parent_grid.assign(np.atleast_1d(lat), np.atleast_1d(lon))
+    return int(parent_grid.shards_of(leaf)[0])
```
espg, line 73: 🤖 *from Claude (review)*

The `nested_in` machinery — this helper, the manifest key, and the invariant that the o10 pin actually sits inside the o9 pin — is only ever exercised under `ZAGG_BENCHMARK_DRIFT=1` in the weekly network job; the fast suite never runs it. The core invariant is checkable offline in milliseconds (verified on this branch):

```python
o9 = from_config(load_config("tests/data/benchmark/configs/atl03_tdigest_healpix_o9.yaml"))
_containing_shard(o9, 11530494877603201034)  # == 11530494877603201033
```

A fast-suite test (e.g. next to `test_targets_manifest_consistent`) asserting, for every shardmap entry carrying `nested_in`: (1) the referenced parent entry exists in `manifest["shardmaps"]`, and (2) `_containing_shard(parent_grid, shard_key) == parent shard_key` — would catch a `nested_in` typo, a mismatched re-pin, or a regression in `_containing_shard` itself on every PR, instead of surfacing a week later in the drift job where it would read as CMR drift rather than a wiring bug. (Nothing in the fast suite currently validates the `nested_in` reference at all.)

src/zagg/catalog/extract.py
espg: 🤖 *from Claude*

Folded in e81a923: `_BOUNDARY_DTYPES` is now the single schema source — `_empty_boundaries()` builds typed zero-row frames and the populated path casts through the same map, so empty-beam/all-empty granules emit the identical parquet schema. Regression test: `test_empty_beam_schema_matches_populated` compares fastparquet schema elements across populated vs empty granules.

src/zagg/catalog/extract.py
espg: 🤖 *from Claude*

Folded in e81a923: write+upload wrapped in try/finally with an existence-guarded remove, so a failed `_put_parquet` can no longer leak scratch parquets into the warm container's /tmp. Test: `test_run_extraction_cleans_scratch_on_upload_failure` asserts the scratch dir is empty after a failed delivery.

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Folded in e81a923: `_handle_extract` now mirrors process mode's `required_cred_keys` gate for the s3 driver — missing keys are a fast 400 (`test_missing_cred_keys_rejected_fast`) instead of a whole-batch 403 burn under the execution role.

tests/test_benchmark_shardmap.py
espg: 🤖 *from Claude*

Folded in e81a923: `test_88s_nested_pin_invariant` (fast suite, offline) validates every `nested_in` reference resolves and that each nested pin's containing parent shard equals the referenced entry's pin, via the same `_containing_shard` the drift job uses — so the nesting invariant is guarded on every PR, not just the weekly CMR job.

src/zagg/processing/streaming.py
```diff
@@ -0,0 +1,207 @@
...
+        self.grid = grid
+        self.handoff = handoff
+        self.buffer_granules = buffer_granules
+        agg_fields = get_agg_fields(config)
+        self._count_fields: list[str] = []
+        self._digest_fields: dict[str, tuple[str, int]] = {}  # name -> (source, delta)
+        for name, meta in agg_fields.items():
+            if get_output_signature(meta)["kind"] == "ragged":
+                delta = int((meta.get("params") or {}).get("delta", _DEFAULT_DELTA))
+                self._digest_fields[name] = (meta["source"], delta)
```
espg: 🤖 *from Claude (review)*

`meta["source"]` raises a bare `KeyError: 'source'` for a tdigest field that omits `source`, but the pooled path accepts that config — `calculate_cell_statistics` defaults it (`source = meta.get("source") or value_col`, aggregate.py L420, with `value_col="h_li"` from the worker). Verified:

```
validate_streaming(cfg)          # passes
StreamingAggregator(cfg, ...)    # KeyError: 'source'
```

So a config that runs pooled dies at aggregator construction with an opaque KeyError when streaming is turned on. Either mirror the pooled default (`meta.get("source") or "h_li"`) or reject source-less ragged fields in `validate_streaming` with a named-field message like the other checks.

src/zagg/processing/streaming.py
```diff
@@ -0,0 +1,207 @@
...
+    problems: list[str] = []
+    if get_chunk_precompute(config):
+        problems.append("chunk_precompute is chunk-scoped and cannot stream")
+    for name, meta in get_agg_fields(config).items():
+        sig = get_output_signature(meta)
+        if "expression" in meta:
+            problems.append(f"field '{name}': expression fields cannot stream")
+        elif sig["resolution"] != "cell":
+            problems.append(f"field '{name}': resolution '{sig['resolution']}' cannot stream")
+        elif sig["kind"] == "ragged":
```
espg, line 81: 🤖 *from Claude (review)*

The buffered path never runs `_coerce_ragged_value`, so a tdigest field's declared `inner_shape` is unvalidated end-to-end. Verified divergence with `inner_shape: [3]`:

- pooled: `ValueError: ragged field produced inner shape (2,), expected (3,)` (aggregate.py L558)
- streaming: `validate_streaming` passes, `flush` stores raw `(k, 2)` digests, and `write_csr` accepts them (it only checks payload-to-payload consistency, csr.py L91-96) — the store's `values` array silently disagrees with the config's declared `inner_shape` that readers/`output_field_signature` key on.

Since `build_tdigest` always emits `(k, 2)`, a cheap fix is one more check here: `sig["inner_shape"] != (2,)` → problem. (Declared `dtype` is fine — `write_csr` casts on write.)

src/zagg/processing/streaming.py
```diff
@@ -0,0 +1,207 @@
...
+            if meta.get("function") != _TDIGEST_FUNCTION:
+                problems.append(
+                    f"field '{name}': ragged function {meta.get('function')!r} has no "
+                    f"merge law (only {_TDIGEST_FUNCTION})"
+                )
+        elif sig["kind"] == "scalar":
+            if meta.get("function") != "len":
```
espg: 🤖 *from Claude (review)*

The pooled path treats `count` as an alias of `len` (`if func_name in ("len", "count")`, aggregate.py L456) and it merges by summation exactly the same way, but this check rejects it: a working pooled config using `function: count` can't turn streaming on. Accepting `("len", "count")` here (and in `__init__`'s count/digest split, which keys on kind only, so it already handles it) closes the gap.

src/zagg/processing/worker.py
```diff
@@ -284,7 +304,10 @@ def process_shard(
     if profile:
         phase_timings["read"] = time.time() - _read_t0
 
-    if not all_reads:
+    if buffered is not None:
+        buffered.flush()  # drain the tail buffer (< buffer_granules granules)
```

Minor, profile-only: under `profile=True` the streaming path's merge cost is misattributed. Intermediate flushes run inside the read loop (`granule_done` → `flush`), so their group+digest CPU is charged to `phase_timings["read"]`; this tail flush runs *after* the `read` stamp (L305) but *before* `_index_t0` (L363), so its cost lands in no phase at all, and `index` then measures ~nothing (the buffered branch at L376 just reads counters). A pooled-vs-streaming profile comparison per issue 148 would show streaming with a fat `read` and near-zero `index`/`aggregate` even when merge CPU dominates. Cheapest fix: move this flush below `_index_t0` (it is the last grouping step, i.e. index work), or note the attribution in the docstring.

deployment/aws/lambda_handler.py
```diff
@@ -230,16 +241,87 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
...
+    missing = [p for p in ("granule_urls", "output_prefix", "s3_credentials") if p not in event]
+    if missing:
+        error_msg = f"Missing required parameters: {', '.join(missing)}"
+        logger.error(error_msg)
+        return {"statusCode": 400, "body": json.dumps({"error": error_msg, "mode": "extract"})}
+
+    driver = event.get("driver", "s3")
+    s3_creds = event["s3_credentials"]
+    if driver == "https":
+        credentials = s3_creds.get("edl_token", s3_creds)
```

The new fail-fast gate only covers the s3 branch; the `https` branch still has the same slow-burn failure mode the gate is meant to prevent: with `driver="https"` and no `edl_token`, `s3_creds.get("edl_token", s3_creds)` passes the whole creds *dict* through as the bearer token, so every granule in the batch 401/403s downstream instead of failing fast with a 400. Process mode never hits this (its gate at L423-428 runs unconditionally before any driver handling), so mirroring it here would mean also rejecting `driver="https"` events missing `edl_token`. One extra `if "edl_token" not in s3_creds: return 400` completes the gate.

src/zagg/processing/streaming.py
espg: 🤖 *from Claude*

Folded in c62ea8e: the aggregator mirrors the pooled default (`meta.get("source") or "h_li"`, matching aggregate.py's `value_col` fallback) so a source-less tdigest field constructs identically under streaming. Test: `test_sourceless_tdigest_field_defaults_like_pooled`.

src/zagg/processing/streaming.py
espg: 🤖 *from Claude*

Folded in c62ea8e: `validate_streaming` now rejects any tdigest field whose declared `inner_shape` is not `[2]`, failing at validation instead of letting the on-disk CSR silently diverge from the schema readers key on. Test: `test_mis_declared_inner_shape_rejected`.

src/zagg/processing/streaming.py
espg: 🤖 *from Claude*

Folded in c62ea8e: `("len", "count")` both accepted, matching the pooled alias in aggregate.py (the count/digest split already keyed on kind, so no other change needed). Test: `test_count_alias_of_len_accepted`.

src/zagg/processing/worker.py

Folded in c62ea8e: the tail flush now runs *before* the `read` stamp, so under profiling the streaming path deliberately charges all group+merge cost (intermediate and tail flushes alike) to `read` — none of it vanishes between phases, and the comment documents that a pooled-vs-streaming profile comparison should read it that way. Test: `test_profile_charges_merge_to_read_phase`.

deployment/aws/lambda_handler.py

Folded in c62ea8e: the https branch now fail-fasts too — `edl_token` absent is a 400 (`Missing s3_credentials keys: edl_token (required for driver='https')`) and the dict-as-bearer-token fallback is gone (`credentials = s3_creds["edl_token"]`). Test: `test_https_driver_missing_edl_token_rejected_fast`.

espg:
🤖 *from Claude*

Status: phases 0, 1, 2, and 4 are complete with all adversarial-review findings folded (10 inline findings across three review passes — every thread has a fold commit + reply); phases 3 and 5 are **Blocked by #150** (the a-priori read plan hooks into the read path #150 rewrites, and the issue decisions hold that reader constant), plus phase 5 additionally needs the 900 s timeout deployed. Pins, schema findings, and the extraction cost estimate are posted on #148 ([comment](https://github.com/englacial/zagg/issues/148#issuecomment-4860577012)). Three non-blocking questions for review sit in the PR body (extract-as-mode vs separate CFN function, 88S drift-job cost, committed shardmap size) — defaults are implemented, redirect at will.

espg:
> Drift-test cost for the 88S maps. The weekly drift rebuild now includes two 88S entries, each re-fetching the 35,639-granule catalog (~20 min) and rebuilding the ring (~3.5 min) — ~50 min added to the weekly job. Options: (a) keep as-is (it's the only guard on the pins), (b) mark the 88S entries drift-skipped and re-pin manually when needed. Default: (a), can flip to (b) in a follow-up commit.

...this seems antithetical to our catalog design process. We can build the granule catalog once, and then save that parquet file. Am I missing something, why aren't we doing this?

> Extract as a mode, not a second function. mode="extract" on the existing process-shard function (zero new CFN resources, shared 900 s timeout/layer/role) — reading "cleanest is just to bump it, rather than any special casing" as pointing this way. If you'd rather have an isolated function (separate concurrency/memory knobs, independent throttling during a full-catalog run), it's a small template.yaml addition — say the word.

I think this makes sense-- we'd want to bundle it as part of the current process-shard function for incremental updates. The case for separate would be as it's own lambda to process the full archive -- this is also useful though. Can we do both? Have it as a mode and a seperate function?

> Committed shardmap size. The pruned 88S maps are ~1.9 MB + 1.6 MB (vs 65–235 kB for NEON maps) because the pinned shards carry 5,620/4,605 granule URL records. Committing them keeps the benchmark self-contained like every other pin; if that's too heavy for the tree, they could move to a release asset + download step instead.

We should commit them, not too large.

Note that there are now branch conflicts that need to be resolved.

espg:
🤖 *from Claude*

All four points from the [review comment](https://github.com/englacial/zagg/pull/152#issuecomment-4880908345) are addressed (844ba9b + the merge commit), and phase 3 is in (fbb3fad):

1. **Catalog built once — agreed, and now supported.** The drift test honors a per-entry `catalog_parquet` key: when set, the weekly rebuild loads the committed stac-geoparquet snapshot (`Catalog.from_geoparquet`) instead of re-fetching CMR — the 88S drift check becomes an offline, deterministic guard on the shardmap build + pin, and ~50 min drops out of the weekly job. (Why it wasn't already: the phase-0 build predated this question and committed only the pruned maps, not the catalog.) The one missing piece is the snapshot file itself — it takes one CMR pass this environment can't make. Two one-pass options are in the PR body under "Questions for review" (1): you run the README's `python -m zagg.catalog ... --catalog-out` command once and commit it, or a network-enabled routine run builds and commits it to this branch next pass.
2. **Extract as a mode AND a separate function — done.** `ExtractFn` twin added to `template.yaml`: same code zip, layer, role, memory, and the shared 900 s `Timeout`, with its own concurrency/throttling pool so a full-archive fan-out can't starve per-cell processing; `mode="extract"` stays on the process function for incremental updates. `test_extract_fn_mirrors_process_fn` pins the twin in lockstep with `ProcessFn`.
3. **Shardmaps stay committed** — unchanged, as confirmed.
4. **Branch conflicts resolved** — merged `origin/main` (the issue #151 async-result channel). The extract-mode dispatch now sits before the `result_url` mirror, which remains process-mode-only per the #151 design (`test_setup_mode_ignores_result_url` guards the non-process modes).

**Phase 3 (arm 2a) is complete**: boundary parquet → per-beam chunk polylines → chunk-aligned `ReadPlan` (`src/zagg/processing/apriori.py`). Chunk selection samples each chunk's boundary photons on the unit sphere and applies the same mortie test as the photon path; matched chunks feed the existing `plan_read` as coarse parents, and the shared `_execute_plan_group` (refactored out of `_planned_read_group`, pure relocation) applies the exact photon mask + filters — so the output is **bit-identical**, asserted against both the planned and full production paths under a stub that *enforces* the h5coro aligned-start B-tree bug, which this arm trips by construction (every a-priori slice starts on a chunk boundary; the arm's read callback mirrors the extractor's start-one-early workaround). With `read_plan.chunk_boundaries` unset, behavior is byte-identical to today's paths. Full description + test inventory in the PR body; 1,156 tests green locally.

Remaining: **phase 5** needs the 900 s timeout deployed (yours) — with PR #150 merged that is the only gate. Applying `waiting` for the snapshot decision (1) and the deploy.

---

src/zagg/processing/read.py
```diff
@@ -563,6 +610,12 @@ def _read_group(h5obj, group: str, data_source: dict, shard_key: int, grid, arro
     rp = data_source.get("read_plan")
     levels = data_source.get("levels")
     base_level = data_source.get("base_level")
+    if isinstance(rp, dict) and rp.get("chunk_boundaries"):
```
espg: 🤖 *from Claude (review)*

**should-fix — `chunk_boundaries: {}` silently runs the wrong arm.** This gate uses truthiness (`rp.get("chunk_boundaries")`), so an empty dict — e.g. `prefix` commented out while tuning a config — falls through to `spatial_index`/full read with *correct output* and no signal. Verified against the branch:

```python
ds["read_plan"]["chunk_boundaries"] = {}
df = _read_group(g, "gt1l", ds, 0, _BandGrid(*BAND))
# rows: 51, geolocation read (arm 1 ran): True — silently measured arm 1
```

That's exactly the failure mode two blocks down warns about for `levels` ("Truthy-checking … would route an empty `{}` … back to the full-read path silently"), and `_load_boundaries`'s own docstring says a silent fallback to another read path "would corrupt the comparison". For a benchmark arm the corrupted artifact is the timing matrix itself.

Suggest a presence check — `"chunk_boundaries" in rp` — here **and** in the worker's mirror gate (`worker.py:223`, `bool(_rp.get("chunk_boundaries"))`) so the two stay in lockstep; `_apriori_read_group`'s existing `prefix` validation then fails loudly on `{}`/`None`.

---

deployment/aws/template.yaml
```diff
@@ -190,10 +193,38 @@ Resources:
       MaximumRetryAttempts: 0
       MaximumEventAgeInSeconds: 60
 
+  # Dedicated chunk-boundary extraction function (issue #148). Extraction is
+  # ALSO a mode of ProcessFn (mode="extract" — the incremental-update path);
+  # this twin exists for full-archive extraction runs, giving them their own
+  # concurrency/throttling pool so a catalog-wide fan-out cannot starve
+  # per-cell processing. Same code zip, layer, role, memory, and the shared
+  # 900 s Timeout — only the function name differs.
+  ExtractFn:
```
espg, line 202: 🤖 *from Claude (review)*

**should-fix — `ExtractFn` has no `EventInvokeConfig` twin.** `ProcessFn` pins `MaximumRetryAttempts: 0` / `MaximumEventAgeInSeconds: 60` (`ProcessFnAsyncConfig`, issue #151); `ExtractFn` gets the AWS async defaults: **2 retries and 6 h max event age**. The twin's stated purpose is exactly the full-archive async fan-out, so a deterministically failing batch (bad granule → the handler's partial-failure 500 doesn't apply to a timeout/OOM, which *does* trigger async retry) re-runs up to 3× at 900 s each. No correctness hazard — parquet writes are idempotent per granule, unlike process mode's zombie-write-into-finalized-store risk — but on a ~500k-granule run it's a real cost multiplier, and the #151 rationale ("re-failing deterministic failures at extra cost") applies verbatim.

Either add an `ExtractFnAsyncConfig` mirroring the pin, or state the intentional divergence in the block comment. Note `test_extract_fn_mirrors_process_fn` compares only the two functions' `Properties`, so this is precisely the drift the lockstep test can't see — worth pinning whichever way you decide.

(Otherwise the twin checks out: same handler/code/layer/role/memory/timeout/env, no new IAM or other resources, and the two Outputs additions are read-only.)

---

src/zagg/processing/apriori.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def _xyz(lat_deg, lon_deg):
+        lat, lon = np.radians(lat_deg), np.radians(lon_deg)
+        return np.stack(
+            [np.cos(lat) * np.cos(lon), np.cos(lat) * np.sin(lon), np.sin(lat)], axis=-1
+        )
+
+    p0 = _xyz(bdf["start_lat"].to_numpy(), bdf["start_lon"].to_numpy())
+    p1 = _xyz(bdf["end_lat"].to_numpy(), bdf["end_lon"].to_numpy())
+    t = np.linspace(0.0, 1.0, samples_per_chunk)[None, :, None]
```
espg, line 109: 🤖 *from Claude (review)*

**nit — guard `samples_per_chunk < 2`.** This is config-reachable (`int(cfg.get("samples_per_chunk", ...))`), and degenerate values fail silently in the dangerous direction:

- `1` → `np.linspace(0, 1, 1) == [0.0]`: only the *start* boundary photon is tested, contradicting the docstring's "Endpoints are included" and halving the guarantee the pad margin is sized against.
- `0` → an empty sample axis → `.any(axis=1)` over zero columns is all-`False` → every chunk unmatched → `_apriori_read_group` returns `None` and the granule is **silently dropped** from the shard.

A one-liner in `_apriori_read_group` (or here) — `if samples_per_chunk < 2: raise ValueError(...)` — turns both typos loud.

---

src/zagg/processing/apriori.py
```diff
@@ -0,0 +1,236 @@
...
+    start one element early and drop it, costing one re-decompressed chunk per
+    run. Valid for every heights dataset because the chunk grid is aligned
+    (verified at extraction time by ``heights_chunk_grid``).
+    """
+
+    def _read_fn(path, hyperslice=None):
+        if hyperslice is None:
+            return h5obj.readDatasets([path])[path]
+        start, end = hyperslice[0]
+        r0 = start - 1 if start > 0 and start % chunk_size == 0 else start
```
espg: 🤖 *from Claude (review)*

**nit — the workaround could be unconditional, dropping the `chunk_size` dependence.** Starting one element early and slicing it off is correct for *any* `start > 0`, aligned or not (when unaligned, `start-1` is in the same chunk, so it costs nothing). Applying it unconditionally would remove the `start % chunk_size` test and with it the `meta.get("chunk_size") or bdf["n_photons"].max()` fallback heuristic in `_apriori_read_group` — one less way for a foreign/legacy parquet without `chunk_size` metadata to miss the trigger and hit the h5coro failure. (I traced the fallback: `max(n_photons)` is in fact always safe — multi-chunk beams have a full chunk, single-chunk beams only ever produce `start == 0` — so this is simplification, not a bug.)

Related: `start, end = hyperslice[0]` assumes single-range hyperslice lists. True for every `plan_read` output today (`chunk_lists` entries are always `[(start, end)]`), but the default `_read_fn` in `_execute_plan_group` forwards the whole list, so if the plan shape ever grows a second range this callback would silently drop it. A `len(hyperslice) == 1` assert would pin the assumption.

---

tests/test_lambda_handler.py
```diff
@@ -914,6 +914,124 @@ def test_sampler_unavailable_off_linux_returns_none(self, handler_mod, monkeypat
         assert s._thread is None  # never spawned
 
 
+class TestExtractMode:
```
espg, line 1155: 🤖 *from Claude (review)*

**nit — pin the merge's one behavioral seam: `result_url` + `mode="extract"`.** The merge resolution places extract dispatch before the issue #151 `result_url` mirror (correct per the design — the mirror is process-mode-only), but the guard test from `main` (`test_setup_mode_ignores_result_url`) only exercises `setup`. Since this branch is what *adds* a new mode on the other side of that seam, one test asserting an extract event with `result_url` set performs no write would pin the resolution — otherwise a future refactor that moves the mirror above the dispatch (mirroring extract's 500s to a poller that then treats a partial batch failure as a shard failure) passes today's suite.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of today's three commits only (merge 5b752e1, fold 844ba9b, phase 3 fbb3fad); earlier commits were reviewed in prior rounds and are not re-litigated.

**Findings: 2 should-fix, 3 nits** (inline). None block the next phase.

1. should-fix — `chunk_boundaries: {}` silently routes to arm 1/full read (reproduced against the branch); contradicts the file's own posture on truthy-checking configs (`read.py`, with the worker's mirror gate).
2. should-fix — `ExtractFn` lacks the `EventInvokeConfig` twin: AWS default 2 retries / 6 h event age vs `ProcessFn`'s pinned 0 / 60 s; cost multiplier on deterministic failures in the full-archive fan-out, and outside what `test_extract_fn_mirrors_process_fn` pins (`template.yaml`).
3. nit — `samples_per_chunk` 0/1 degrade silently (0 silently drops the granule) (`apriori.py`).
4. nit — boundary read_fn: workaround could be unconditional (removes the `chunk_size` fallback dependence); single-range hyperslice assumption unpinned (`apriori.py`).
5. nit — the merge's one behavioral seam (`result_url` × `mode="extract"`) has no test; the existing guard only covers `setup` (`test_lambda_handler.py`).

**Verified clean:**

- **Merge resolution** (5b752e1): diffed the merged branch against `main` 8bece8b for both conflicted files — the branch is exactly `main` + the extract additions; nothing from the issue #151 async-result channel was dropped or doubled, `_write_result` intact, extract dispatch before the process-only mirror as described.
- **`_execute_plan_group` refactor is a pure relocation**: mechanically diffed old `_planned_read_group` vs new split — only deltas are the moved `coordinates`/`variables` bindings and `levels`/`base_level` via `.get` (identical for the planned caller, which validates both upstream).
- **Flag-off byte identity**: `granule_url` kwarg defaults to `None` and the worker builds the kwarg dict only when the feature is on; the flag-off call is literally the pre-existing signature (and `test_flag_off_call_signature_unchanged` pins it with a fake that would `TypeError`).
- **Bit-identity test gates what it claims**: `pd.testing.assert_frame_equal(..., check_exact=True)` against *both* production paths, under a stub that fails any chunk-aligned-start read (so identity proves the workaround), with the shard band strictly interior to one chunk (so identity also proves interpolated-sample detection, not endpoint luck); IO bounding asserted (zero geolocation reads, all hyperslices `(499, 800)`).
- **Plan math**: `plan_read`'s `full_read` plans always carry `parent_runs`, so the `not plan.parent_runs` → `full_read` check order in `_apriori_read_group` is safe; truncated-final-chunk offsets come from `cumsum(n_photons)` not `chunk * chunk_size`; unit-sphere lerp is well-conditioned for intra-chunk spans and the antimeridian test pins the failure mode of naive lon lerp.
- **Conventions**: `apriori.py` 236 lines, `read.py` 814 (under the ~1000 cap); no new dependencies (boto3/fastparquet/numpy all existing); no lint suppressions added; drift-test `catalog_parquet` path is mechanism-only (no manifest entry points at a missing snapshot, matching Questions for review (1)).
- Ran the 18 new/changed tests (`test_apriori.py` + `test_extract_fn_mirrors_process_fn`) from a clean checkout of the branch head: all pass. CI on fbb3fad was still pending at review time.

---

espg:
🤖 *from Claude*

All five findings from the [phase-3 review](https://github.com/englacial/zagg/pull/152#pullrequestreview-4629111169) are folded in df44f60:

1. **Presence-gated dispatch** — `_read_group` and the worker's mirror gate now check `"chunk_boundaries" in rp` instead of truthiness, so a present-but-empty/null block raises `read_plan.chunk_boundaries requires a 'prefix'` inside the a-priori path instead of silently running the planned arm. Pinned by `test_empty_block_fails_loudly_not_falls_back` (parametrized over `{}` and `None`, on a config that also carries `spatial_index` so a truthiness gate would have quietly fallen back).
2. **`ExtractFnAsyncConfig` added** — mirrors `ProcessFnAsyncConfig` (`MaximumRetryAttempts: 0`, `MaximumEventAgeInSeconds: 60`), completing the twin so an async extraction fan-out can't re-bill a deterministic failure at up to 900 s per service retry. `test_extract_fn_async_config_mirrors_process_fn` pins the EventInvokeConfig in lockstep (Properties identical apart from the `FunctionName` Ref).
3. **`samples_per_chunk >= 2` enforced** — `_chunk_shard_mask` raises on fewer (with 1, `np.linspace(0, 1, 1)` would silently test only the start photon, breaking the documented both-endpoints contract). Test: `test_samples_per_chunk_below_two_rejected`.
4. **Unconditional aligned-start workaround** — the boundary read_fn now shifts *any* nonzero start one element early, removing the `chunk_size` metadata dependence entirely (`_make_boundary_read_fn(h5obj)`; the meta fallback in `_apriori_read_group` is gone). The single-range `hyperslice[0]` assumption is pinned with an assert (plan_read emits one `[start, end)` pair per slice). The bit-identity test still passes under the bug-enforcing stub, and its slice assertion (`(499, 800)`) is unchanged since a-priori starts are chunk-aligned anyway.
5. **Extract × `result_url` seam test** — `test_extract_mode_ignores_result_url` added next to the setup-mode guard: extract dispatches before the mirror, which stays process-mode-only.

Gates: full suite green (1,161 passed, 26 skipped), ruff check + format clean at the pinned 0.14.10 (the pre-existing `N818` aside).

---

espg:
🤖 *from Claude*

Reconciled with `main` after the PR #70 (issue #12 temporal pipeline) merge: e4e0874. One conflicted file (`deployment/aws/lambda_handler.py` — module docstring + mode dispatch); resolution keeps main's canonical `process_event`/mirror structure with extract dispatched before it, returning directly — extract stays outside the `result_url` mirror, pinned by `test_extract_mode_ignores_result_url` against the restructured block. Extract/a-priori behavior unchanged; full suite green post-merge (1,286 passed, 26 skipped).

---

espg:
🤖 *from Claude*

Question (1) is closed: the 88S catalog snapshot is committed in adbe1ce.

1. **Built once from CMR** (35,639 granules, one ~23 min pass, run from espg's workstation): `--catalog-out` per the README recipe; the same build's shardmap output reproduced the 576-shard ring exactly.
2. **Size**: 31 MB as written (snappy) → **16.6 MB committed** (lossless zstd-15 recompression). Above the O(10 MB) guess for a structural reason: `geometry` is 25 of the 31 MB — full CMR swath polygons, which cannot be simplified without changing shard assignment and invalidating the very pins the drift test guards. Judged in-tree-acceptable as a one-time immutable artifact per the build-once direction; say the word to flip to a release asset instead.
3. **Wired and proven**: both 88S `targets.json` entries now carry `catalog_parquet: catalogs/cat_88s.parquet`, and the gated drift check (`ZAGG_BENCHMARK_DRIFT=1`) rebuilds both rings from the snapshot and reproduces the pins — **2 passed in 6m09s, fully offline**. The weekly job's ~23 min CMR refetch per 88S entry drops out; what remains is the deterministic rebuild.

Issue #101, opened 2026-06-24
Optional strict-AOI polygon clip for shardmap output (sub-shard clipping)


espg:
### Problem

zagg's output currently includes **all data in any grid cell touching the AOI**, not data strictly inside the AOI polygon. The shard universe is `all_shards = grid.coverage(region)` — the `parent_order` cells that *overlap* the region — and each worker then aggregates every photon in any child cell of its shard. There is no final clip to the region polygon, so the effective collected area overhangs the AOI by ~one shard-cell all around.

Because that overhang scales with shard size, the reported obs count depends on `parent_order` even for the same AOI. Measured on the NEON SERC bbox (`[-76.6233, 38.8430, -76.5061, 38.9346]`, see #100):

| parent_order | shard cell | obs total | obs inside bbox | obs outside bbox |
|---|---|---|---|---|
| 10 | ~6 km | 10,130,471 | 2,160,665 | 7,969,806 |
| 13 | ~0.8 km | 2,548,582 | 2,123,810 | 424,772 |

The in-AOI data agrees (~2.1M); the rest is out-of-AOI spillover. This is working as currently designed, but there's no way to ask for a strict-AOI result.

### Proposal

Add an opt-in **strict-AOI clip**: after aggregation (or before, as a per-photon mask), drop observations whose cell center falls outside the supplied region polygon. Open questions for design:

- Clip granularity: per-photon (exact) vs per-leaf-cell center (cheaper, ~10 m slop) vs keep-whole-cell-if-center-in.
- Where to apply: in the worker before reduce (saves write volume) vs a post-pass over the store.
- Interaction with `pad` and the existing shard mask (`grid.shards_of(leaf)==shard_key`).
- Default off (preserve current cell-aligned behavior) with a config/CLI flag, or a `region`-derived mask carried in the shardmap.

### Context

Spun out of #100, where the obs-count divergence between order-11 and order-13 runs turned out to be this overhang effect, not data loss. Refs #100.

espg:
> The in-AOI data agrees (~2.1M); the rest is out-of-AOI spillover. This is working as currently designed

yes it is.

>  ...but there's no way to ask for a strict-AOI result.

I feel like the clean solution here is an AOI_MASK boolean array that gets packaged... this would be an optional argument to add to the agg template.

espg:
🤖 *from Claude*

[...]

PR #153, opened 2026-07-01
async Lambda dispatch with S3-polled worker results (issue #151)

### LINKED ISSUE #151: Warm start lambda and memory freeing are *still* a problem

espg:
Closes #151.

Implements the durable fix directed on the issue ([espg's comment](https://github.com/englacial/zagg/issues/151#issuecomment-4860468469), item (3) of [the analysis](https://github.com/englacial/zagg/issues/151#issuecomment-4860426861)): stop holding a synchronous Lambda connection open for minutes. The lambda backend now dispatches each cell with `InvocationType="Event"` and polls for a per-shard result object the worker writes next to the output store, so no connection sits idle while a shard runs — the ~4-minute NAT/SNAT idle kill on GitHub-hosted runners (the cause of every o10_inner/o9 benchmark failure) can no longer sever a result, and the 6 MB synchronous response cap no longer applies.

## Mechanism confirmed in CloudWatch before implementing

`/aws/lambda/process-shard-test`, failing window of run [28549136504](https://github.com/englacial/zagg/actions/runs/28549136504) (Jul 1 ~21:36–21:50 UTC): every "failed" target **completed cleanly server-side, with data, far under both caps**:

| REPORT | duration | max memory | worker `processing_complete` |
|---|---|---|---|
| `a0863fb3` (o10_inner) | 258.0 s | 985 MB | shard 5347405532456026122, **obs=939,217** |
| `f4e9c58f` (o9_sharded) | 330.8 s | 1290 MB | shard 5347395636851376137, **obs=3,112,138** |
| `2125ddc9` (o9_inner) | 475.5 s | 1250 MB | shard 5347395636851376137, **obs=3,112,138** |

No `Task timed out`, no `Runtime.OutOfMemory`. The client recorded `obs=0` for shards whose data was on S3 — transport, not memory. (Bonus: o9's true worst-case duration is ~476 s, comfortably inside the 720 s function timeout.)

## What the change does

**Worker** (`deployment/aws/lambda_handler.py`): new optional `result_url` event key. When present, `lambda_handler` mirrors the process-mode response envelope (`{statusCode, body}`) as JSON to that object — `s3://…/out.zarr.status/<run_id>/<shard_key>.json` — using the same credential/endpoint resolution as the output store (`_output_store_kwargs`). Covers the 200/400/500 branches (it wraps `_handle_process`); never raises. Absent key → byte-identical behavior.

**Runner** (`src/zagg/runner.py`): new `agg(..., invocation="async")` kwarg (default `"async"`):
- `_run_lambda` mints a per-run status prefix `<store>.status/<uuid>` (the run id keeps reruns into the same store from reading stale results) and threads `result_url` / a lazily-built `result_fetch` closure / `poll_timeout_s` into each cell invoke.
- `_invoke_lambda_cell` invokes with `InvocationType="Event"`; `_poll_lambda_result` polls the result object (5 s + sub-second jitter) and maps the envelope into the **same result-dict shape** as the sync path, so the accumulator, cost math, and summary are unchanged. Transient dispatch faults (throttle) keep the existing retry/backoff; a result missing at the deadline (function `Timeout` + 90 s margin) is recorded as the shard's failure and **not retried** (deterministic, mirroring the #119 FunctionError rule).
- The function `Timeout` read moves into `preflight` (it now feeds the poll deadline as well as the existing `worker_pct_timeout` summary field) — still exactly one `GetFunctionConfiguration` call.
- `invocation="sync"` keeps the legacy `RequestResponse` path byte-identical — the escape hatch for older deployed workers or per-cell events above the 256 KB async payload cap.

**Store** (`src/zagg/store.py`): `open_object_store()` — the raw obstore store (same path/credential handling as `open_store`, no Zarr wrapper) used by both sides of the channel.

**Infra riding in this PR for you to deploy (nothing was deployed from this session):**
- `deployment/aws/template.yaml`: `AWS::Lambda::EventInvokeConfig` with `MaximumRetryAttempts: 0` / `MaximumEventAgeInSeconds: 900`. Without it, Lambda's async defaults re-run a timed-out/OOM'd shard twice with delays — violating the benchmark's "measure one clean invocation" contract (#119) and risking late zombie writes into a store the caller has moved on from.
- `.github/scripts/deploy_lambda.sh`: applies the same event-invoke config on the in-place deploy path (which is how `process-shard-test` actually gets updated — the tier-2 deploy never re-runs the CFN stack). Warn-don't-fail until the deploy roles gain the new IAM action.
- `deployment/aws/benchmark_cicd.yaml`: `BenchmarkInvokeRole` gains `s3:GetObject` on the benchmark store bucket (the orchestrator must now *read* the result objects; previously it needed only `lambda:InvokeFunction`); Deploy/Release roles gain `lambda:PutFunctionEventInvokeConfig`.

## Phases

- [x] Phase 1 — worker-side result channel: `open_object_store` + handler `result_url` write, tests
- [x] Phase 2 — runner async dispatch (default) + poll + `invocation="sync"` fallback, tests
- [x] Phase 3 — infra: EventInvokeConfig (template + deploy script), benchmark-role IAM, template/deploy tests

## How it was tested

- `pytest` green locally: 1089 passed. One pre-existing environment failure (`test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`: the local `pip` resolves py3.10 wheels; fails identically on a clean `origin/main` checkout — unrelated).
- `pre-commit run` (pinned ruff check/format, codespell, mypy) clean on all touched files; the mypy failures it reports are pre-existing in untouched modules under local tool versions.
- New tests: handler mirrors the envelope on 200/400 paths, skips absent-key/setup-mode, never raises on write failure; runner async event carries `result_url`, envelope→result-dict mapping, poll-until-lands, deadline-expiry-not-retried, throttle-retry-then-accept; `agg` threads the channel by default, omits it for `invocation="sync"`, rejects unknown values; `open_object_store` local roundtrip + S3 credential branches.

## Questions for review

1. **Deployment ordering / default choice.** The default flipped to async, so a new runner against a not-yet-redeployed worker records every shard as `no worker result within …s`. The benchmark is safe (tier-2 redeploys `process-shard-test` from the PR tree before invoking), but production `process-shard` needs a redeploy before anyone upgrades the client — or callers pass `invocation="sync"` in the interim. Option: flip the default to `"sync"` and let the benchmark opt in, if you'd rather stage it.
2. **IAM/stack updates needed before this PR's benchmark can pass**: apply the `benchmark_cicd.yaml` changes. The poll hard-requires the benchmark role's `s3:GetObject` grant (unless `sliderule-public` is anonymously readable — the explicit grant is correct either way); the event-invoke config is warn-only until the deploy roles gain `PutFunctionEventInvokeConfig`.
3. **Async payload cap**: `InvocationType="Event"` caps the request at 256 KB (vs 6 MB sync). Benchmark shard events are well under it; an extreme granule-dense shard with a large `aoi_payload` could exceed it and will surface as a per-cell dispatch error. Documented in the `agg` docstring with `invocation="sync"` as the workaround; a pointer-to-S3 event scheme is a possible follow-up if it ever bites.
4. `runner.py` was already past the ~1000-line note in CLAUDE.md §4 (1222 lines before this PR, ~1360 after). Happy to split the Lambda invoke/poll helpers into their own module as a follow-up if you want.
5. Status objects are left in place after the run (`<store>.status/<run_id>/`) as a debugging artifact rather than deleted — cheap, and the run id prevents stale reads. Say the word if you'd rather the runner best-effort delete them.

### DIFF
--- a/deployment/aws/template.yaml
+++ b/deployment/aws/template.yaml
@@ -172,18 +172,23 @@ Resources:
         S3Key: !Ref FunctionS3Key
 
   # Async-invoke hygiene (issue #151): the runner dispatches per-cell events
-  # with InvocationType=Event and polls for the worker-written result object.
-  # Lambda's async defaults would re-run a timed-out/OOM'd shard twice with
-  # delays -- re-failing deterministic failures at extra cost and risking late
-  # writes into a store the caller has moved on from -- so pin retries to 0 and
-  # bound queued-event age to the runner's poll window (Timeout + margin).
+  # with InvocationType=Event and polls for the worker-written result object
+  # until Timeout + a 90 s margin. Lambda's async defaults would re-run a
+  # timed-out/OOM'd shard twice with delays -- re-failing deterministic
+  # failures at extra cost and risking late writes into a store the caller has
+  # moved on from -- so pin retries to 0. Event age must stay UNDER the poll
+  # margin so no first delivery can start after the runner stops listening (a
+  # late run would write into the store post-finalize); 60 is the API minimum
+  # and leaves >=30 s of the margin for the result write. A delivery delayed
+  # past it is dropped and the runner records the shard failed at its
+  # deadline -- strictly better than a zombie write.
   ProcessFnAsyncConfig:
     Type: AWS::Lambda::EventInvokeConfig
     Properties:
       FunctionName: !Ref ProcessFn
       Qualifier: "$LATEST"
       MaximumRetryAttempts: 0
-      MaximumEventAgeInSeconds: 900
+      MaximumEventAgeInSeconds: 60
 
 Outputs:
   FunctionArn:
--- a/deployment/aws/template.yaml
+++ b/deployment/aws/template.yaml
@@ -171,6 +171,20 @@ Resources:
         S3Bucket: !Ref ArtifactBucket
         S3Key: !Ref FunctionS3Key
 
+  # Async-invoke hygiene (issue #151): the runner dispatches per-cell events
+  # with InvocationType=Event and polls for the worker-written result object.
+  # Lambda's async defaults would re-run a timed-out/OOM'd shard twice with
+  # delays -- re-failing deterministic failures at extra cost and risking late
+  # writes into a store the caller has moved on from -- so pin retries to 0 and
+  # bound queued-event age to the runner's poll window (Timeout + margin).
+  ProcessFnAsyncConfig:
+    Type: AWS::Lambda::EventInvokeConfig
+    Properties:
+      FunctionName: !Ref ProcessFn
+      Qualifier: "$LATEST"
+      MaximumRetryAttempts: 0
+      MaximumEventAgeInSeconds: 900
+
 Outputs:
   FunctionArn:
     Description: ARN of the deployed Lambda function.
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -16,6 +16,7 @@ import logging
 import os
 import statistics
 import time
+import uuid
 import warnings
 from concurrent.futures import ThreadPoolExecutor
 
@@ -55,7 +56,7 @@ from zagg.processing import (
     write_shard_to_zarr,
 )
 from zagg.processing.write import _block_index_key
-from zagg.store import open_store
+from zagg.store import open_object_store, open_store
 
 logger = logging.getLogger(__name__)
 
@@ -90,6 +91,7 @@ def agg(
     handoff: str | None = None,
     profile: bool = False,
     max_retries: int = 3,
+    invocation: str = "async",
 ) -> dict:
     """Run the aggregation pipeline.
 
@@ -158,6 +160,20 @@ def agg(
         exception) are never retried regardless. Default ``3``. Ignored by the
         ``"local"`` backend. Set to ``1`` (e.g. the CI benchmark) to measure one
         clean invocation and record a failure as a failure.
+    invocation : str
+        Lambda-only (issue #151). ``"async"`` (default) dispatches each cell
+        with ``InvocationType="Event"`` and polls a per-shard result object the
+        worker writes next to the output store
+        (``<store>.status/<run_id>/<shard_key>.json``), so no synchronous
+        connection sits idle while the shard runs -- shards longer than a NAT
+        idle window (~4 min on GitHub-hosted runners) complete reliably, and
+        the 6 MB synchronous response cap no longer applies. Requires (a) a
+        deployed worker that honors the ``result_url`` event key and (b) the
+        caller to have read access to the output bucket for the poll. Note the
+        async request payload cap is 256 KB (vs 6 MB synchronous); pass
+        ``"sync"`` for older deployed workers or extreme per-cell payloads
+        (granule-dense shards with large AOI masks). Ignored by the
+        ``"local"`` backend.
 
     Returns
     -------
@@ -227,6 +243,8 @@ def agg(
         max_workers = min(max_workers, n_cells)
         if not store_path.startswith("s3://"):
             raise ValueError(f"Lambda backend requires s3:// store path, got: {store_path}")
+        if invocation not in ("async", "sync"):
+            raise ValueError(f"Unknown invocation: {invocation!r} (expected 'async' or 'sync')")
         if function_name is None:
             function_name = os.environ.get("ZAGG_LAMBDA_FUNCTION_NAME", "process-shard")
         return _run_lambda(
@@ -246,6 +264,7 @@ def agg(
             handoff=resolved_handoff,
             profile=profile,
             max_retries=max_retries,
+            invocation=invocation,
         )
     else:
         raise ValueError(f"Unknown backend: {backend!r} (expected 'local' or 'lambda')")
@@ -603,6 +622,7 @@ def _run_lambda(
     handoff="arrow",
     profile=False,
     max_retries=3,
+    invocation="async",
 ):
     """Run processing via AWS Lambda invocation.
 
@@ -664,6 +684,19 @@ def _run_lambda(
         region,
     )
 
+    # Async result channel (issue #151): a per-run unique status prefix next to
+    # the output store. Each worker mirrors its response envelope to
+    # <prefix>/<shard_key>.json and the dispatch threads poll for it instead of
+    # holding a synchronous invoke connection open -- GitHub-hosted runners sit
+    # behind a ~4 min NAT idle timeout that severed every >250 s benchmark
+    # target. The run_id keeps reruns into the same store from reading stale
+    # results. ``invocation="sync"`` skips the channel (legacy RequestResponse).
+    result_prefix = None
+    result_box: dict = {}
+    if invocation == "async":
+        result_prefix = f"{store_path.rstrip('/')}.status/{uuid.uuid4().hex}"
+        logger.info(f"Async worker results at {result_prefix}")
+
     # The dispatch lambda_client is built inside preflight() (once the probe
     # has clamped the worker count, which sizes its connection pool), so the
     # per-cell / finalize closures read it from this holder rather than closing
@@ -704,6 +737,9 @@ def _run_lambda(
             region_name=region,
             config=boto_config,
         )
+        # Read the function Timeout once, pre-fan-out: the async poll deadline
+        # is keyed to it (issue #151), and the summary reports it (#100).
+        state["function_timeout_s"] = _get_function_timeout_s(state["lambda_client"], function_name)
         return PreflightReport(workers=clamped, detail=concurrency_report)
 
     # Per-cell invoke, bound to everything but the (shard_key, records) pair so
@@ -717,6 +753,17 @@ def _run_lambda(
         extra = {}
         if aoi_by_shard:
             extra["aoi_payload"] = aoi_by_shard.get(int(shard_key))
+        # Async dispatch (issue #151): where the worker writes this shard's
+        # result, how to poll for it, and how long before giving up (function
+        # timeout + queue/write margin). Sync runs pass none of these, keeping
+        # the invoke byte-identical to the legacy path.
+        if result_prefix is not None:
+            key = f"{int(shard_key)}.json"
+            extra["result_url"] = f"{result_prefix}/{key}"
+            extra["result_fetch"] = _result_fetcher(
+                result_box, result_prefix, output_creds_event, region, key
+            )
+            extra["poll_timeout_s"] = state["function_timeout_s"] + _ASYNC_POLL_MARGIN_S
         return _invoke_lambda_cell(
             state["lambda_client"],
             grid.block_index(int(shard_key)),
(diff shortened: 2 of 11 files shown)


src/zagg/runner.py
```diff
@@ -1001,6 +1048,110 @@ def _log_concurrency_report(report: ConcurrencyReport, max_workers: int) -> None
...
+    accumulator and summary are unchanged. A result missing at the deadline
+    means the worker timed out, was OOM-killed, or crashed before its write --
+    deterministic outcomes, so (like sync FunctionErrors, #119) it is recorded
+    as the shard's failure rather than re-invoked.
+    """
+    if poll_timeout_s is None:
+        poll_timeout_s = _DEFAULT_FUNCTION_TIMEOUT_S + _ASYNC_POLL_MARGIN_S
+    deadline = wall_start + poll_timeout_s
+    while True:
+        result = result_fetch()
```

**Fetch faults escape into the invoke retry classifier — a transient S3 error re-dispatches a still-running shard.** `_poll_lambda_result` runs inside `_invoke_lambda_cell`'s `try:` (the `return _poll_lambda_result(...)` at runner.py:1334), and `_fetch_result` swallows only `NotFoundError`/`FileNotFoundError`. Any other exception from a poll GET — S3 503 SlowDown, connection reset, a GET read timeout, AccessDenied/expired STS creds, even a `JSONDecodeError` — propagates into the generic `except Exception` retry handler. If its message matches a retryable substring (`"Read timeout"` / `"timed out"` are exactly what a throttled or slow S3 GET produces), the whole shard is **re-invoked as a second Event** while the first worker is still running: two concurrent workers writing the same shard's chunks, double cost, and a violation of both this function's "missing result is recorded, not re-invoked" contract and the #119 one-clean-invocation rule. A non-retryable fetch fault instead records the shard failed immediately while the first worker is still running (zombie write into a store the accumulator has moved past).

Fetch faults should be contained inside the poll loop: treat them as a miss and keep polling until the deadline, and surface the persistent cause (e.g. missing `s3:GetObject`) in the deadline error instead of the misleading "worker timeout, OOM, or crash" message. There is no test covering a raising `result_fetch`; worth one asserting `client.invoke.call_count == 1`.

deployment/aws/template.yaml
```diff
@@ -171,6 +171,20 @@ Resources:
...
+  # delays -- re-failing deterministic failures at extra cost and risking late
+  # writes into a store the caller has moved on from -- so pin retries to 0 and
+  # bound queued-event age to the runner's poll window (Timeout + margin).
+  ProcessFnAsyncConfig:
+    Type: AWS::Lambda::EventInvokeConfig
+    Properties:
+      FunctionName: !Ref ProcessFn
+      Qualifier: "$LATEST"
+      MaximumRetryAttempts: 0
+      MaximumEventAgeInSeconds: 900
```

**`MaximumEventAgeInSeconds: 900` does not match the poll window the comment above claims it mirrors — a delayed delivery becomes a false failure plus exactly the late zombie write this config is meant to prevent.** The runner gives up at function `Timeout` + 90 s (`_ASYNC_POLL_MARGIN_S`), i.e. 810 s with the default `Timeout: 720`, budgeting 90 s for async queue latency. But this setting lets Lambda hold a queued event for up to 900 s before its *first* delivery (delivery is delayed under throttling — e.g. a concurrent run or other tenants consuming account concurrency after the preflight clamp took its snapshot). Any event delivered later than the 90 s margin starts a worker after the runner has already recorded `no worker result within 810s` for that shard: a false failure, and a worker writing into the store as late as ~900+720 s after dispatch — after `executor.finalize()` has consolidated metadata and possibly after the caller has moved on. `MaximumRetryAttempts: 0` stops *re-runs* but not *late first runs*.

To actually bound the hazard, either (a) drop `MaximumEventAgeInSeconds` to ≤ the poll margin (e.g. 60), so an event that cannot be delivered inside the runner's window is discarded instead of run late, or (b) keep 900 and grow the runner's poll deadline to `Timeout + MaximumEventAgeInSeconds + margin`. Either way, align the comment on line 179 ("bound queued-event age to the runner's poll window (Timeout + margin)" — today 900 > 810) and the copy of both the value and the comment in `.github/scripts/deploy_lambda.sh`.

src/zagg/runner.py

Folded in 461ffdd: `_poll_lambda_result` now wraps `result_fetch()` in its own try/except -- any fetch fault is treated as a miss (keep polling to the deadline), so it can never reach the invoke retry classifier and re-dispatch a running shard. A persistent fault (e.g. missing `s3:GetObject`) is surfaced in the deadline error (`result fetch failing: ...`) instead of masquerading as a worker crash. Pinned by `test_fetch_fault_is_contained_not_reinvoked` (raising fetch -> exactly one invoke, cause in the error).

deployment/aws/template.yaml
espg: 🤖 *from Claude*

Folded in 461ffdd, taking the clamp option: `MaximumEventAgeInSeconds: 60` (the API minimum), which is strictly under the 90 s poll margin -- no first delivery can start after the runner stops listening, and a delivery that would have been that late is dropped and recorded as the shard's failure at the deadline (strictly better than a zombie post-finalize write). Growing the poll deadline to Timeout + event age + margin was the alternative, but it makes every real worker crash cost up to ~28 min of benchmark wall time. Template comment, deploy script, and both tests aligned (the template test now also asserts age < `_ASYNC_POLL_MARGIN_S`).

espg:
> Deployment ordering / default choice

We'll cut and deploy a new zagg release today, so async is fine (the prod client will match soon)

> IAM/stack updates needed before this PR's benchmark can pass: apply the benchmark_cicd.yaml changes. The poll hard-requires the benchmark role's s3:GetObject grant (unless sliderule-public is anonymously readable — the explicit grant is correct either way); the event-invoke config is warn-only until the deploy roles gain PutFunctionEventInvokeConfig.

remind me what the cloudformation invokation command for this is?

My initial read here was that we should switch the output bucket to sliderule-public-cors , which is publicly readable... pros / cons of this vs the grant?

> Async payload cap: InvocationType="Event" caps the request at 256 KB (vs 6 MB sync). Benchmark shard events are well under it; an extreme granule-dense shard with a large aoi_payload could exceed it and will surface as a per-cell dispatch error. Documented in the agg docstring with invocation="sync" as the workaround; a pointer-to-S3 event scheme is a possible follow-up if it ever bites.

hesitant on this point... we may have large payloads. We'll know soon enough with #148 

> runner.py was already past the ~1000-line note in CLAUDE.md §4 (1222 lines before this PR, ~1360 after). Happy to split the Lambda invoke/poll helpers into their own module as a follow-up if you want.

defer for now

> Status objects are left in place after the run (<store>.status/<run_id>/) as a debugging artifact rather than deleted — cheap, and the run id prevents stale reads. Say the word if you'd rather the runner best-effort delete them.

if these are left in the sliderule-public bucket, they're deleted automatically after 30 days anyways

espg:
🤖 *from Claude*

Recommendations on the five open questions, plus two cross-PR notes.

**(1) Default async vs sync — keep `"async"` as the default, but make the stale-worker failure self-diagnosing.** The failure being fixed (SNAT-severed results recorded as `obs=0`) was *silent* wrongness; the failure mode of async-against-an-old-worker is loud and total (every shard `no worker result within …s`), which is the better default posture — provided the message names the remedy. Concretely: append to the deadline-expiry error text something like "worker may predate result_url support — redeploy the function, or pass `invocation=\"sync\"`". One string change; then deployment ordering is a non-issue for anyone who can read an error. Flipping the default to `"sync"` would keep the silent-corruption path as the default for one more cycle, which is backwards.

**(2) IAM/stack ordering** — yours to run, sequence: apply `benchmark_cicd.yaml` (role grants) first, then a function redeploy picks up the EventInvokeConfig (warn-only until the roles land, as built), then a benchmark run validates end-to-end. The poll's `s3:GetObject` grant is the hard prerequisite; everything else degrades gracefully.

**(3) 256 KB async cap** — documented workaround is fine. One cheap hardening worth a line of code: check the serialized event size client-side before `Invoke` and fail with the `invocation="sync"` pointer, rather than letting Lambda's `RequestEntityTooLargeException` surface raw. The realistic overflow is a strict-AOI `aoi_payload` MOC (#101) on a complex polygon, so it will bite someone eventually, and the pre-flight check turns a confusing dispatch error into an actionable one. Pointer-to-S3 events stay a follow-up only if that ever actually triggers.

**(4) `runner.py` at ~1360 lines** — split, but as its own mechanical follow-up PR (move the Lambda invoke/poll/result-channel helpers to a `src/zagg/dispatch.py` or similar, no behavior change), not inside this diff. Keeping this PR reviewable as-is beats restructuring it now; the §4 note is satisfied by having raised it here — it just needs your go-ahead for the follow-up.

**(5) Status-object retention** — keep them, don't make the runner delete. Runner-side cleanup adds a failure mode to the happy path and destroys exactly the artifact you want when a run went sideways. If accumulation ever matters, an S3 lifecycle rule on the `.status/` prefix (e.g. expire after 30 days) is the right mechanism — zero code, no new IAM for the runner.

**Cross-PR notes:**
- PR #152 (88S, issue #148) also touches `deployment/aws/template.yaml` (`Timeout` 720→900) and the dispatch client config (`read_timeout` 900→960 — a sync-path setting that matters much less once this PR's async default is live, but is still correct for the `invocation="sync"` fallback). No textual conflict expected, but whichever merges second should rebase and re-run the template tests. Pleasant synergy: the poll deadline here derives from `GetFunctionConfiguration`'s `Timeout`, so #152's 900 s bump flows into the poll margin automatically with no edit.
- The issue thread's short-term item (`tcp_keepalive` on the dispatch `Config`) is now only relevant to the sync fallback path. Worth either a two-line addition here (harmless, helps `invocation="sync"` users behind NAT) or an explicit "won't do — async default obsoletes it" note on #151 so the item doesn't dangle.

CI is fully green on this head (both test matrices, both lambda-build arches, ruff). With (1)'s one-line message tweak folded, this looks ready for your review pass.

espg:
🤖 *from Claude*

Both items from [the assessment](https://github.com/englacial/zagg/pull/153#issuecomment-4860825024) are folded in a9d4158:

1. **Deadline error is now self-diagnosing** — the no-result path reads: `no worker result within Ns (worker timed out, was OOM-killed, or crashed before writing its result (check CloudWatch) -- or the deployed worker predates result_url support: redeploy the function, or pass invocation="sync")`. With that remedy inline, the async default stays (per the assessment's answer to Questions-for-review (1)).
2. **256 KB async payload pre-flight** — `_invoke_lambda_cell` serializes the event once and, on the Event path only, raises with the `invocation="sync"` remedy at >250 KiB (`_ASYNC_PAYLOAD_CAP_BYTES`) instead of letting `RequestEntityTooLargeException` surface raw. Pinned by `test_oversized_async_payload_raises_before_dispatch` (oversized synthetic `aoi_payload`, invoke never called) and `test_oversized_payload_allowed_on_sync_path` (same event passes synchronously, so the remedy is real).

Issue #158, opened 2026-07-02
88S offsets extraction: per-chunk byte offsets + boundaries for the pinned benchmark shard (arm 2b of #148)


espg:
🤖 *from Claude*

Sub-issue of #148, opened at espg's direction. PR #152 built the boundary-*geometry* arm; this tracks the byte/chunk-*offset* side it deliberately deferred, for the pinned 88S benchmark shard.

## What #152 already delivered (and decided)

- `src/zagg/catalog/extract.py` + `mode="extract"` on the existing Lambda (PR #152 phase 2): per-granule, per-beam chunk-boundary lat/lon → parquet, scoped to the pinned 88S o9 shard's granule list. Parquet schema is geometry-only per espg's decision on the thread.
- The offset deferral's two prerequisites were then delivered there: **(a) cost** — extraction 0.54–2.46 s/granule locally (~$1 for the pinned shard, ~$75–120 for the full ~500k-granule catalog); **(b) inferability** — per-chunk byte offsets **cannot** be inferred from boundary geometry (deflate produces a ~3× compressed-chunk-size spread and chunks interleave across datasets), so offsets must be read from each file's B-tree.
- Status caveat: the extract mode has only run against locally cached NEON granules; nothing has executed against 88S yet (PR #152 is blocked by #150, and the full-shard run also waits on the 900 s deploy).

## Scope of this issue

Produce per-chunk `(byte_offset, nbytes)` + element ranges for every heights dataset of every granule in the pinned 88S shards (o9 pin `11530494877603201033`, 5,620 granules; o10 pin `11530494877603201034` nested inside it), as the a-priori read input for #148's arm (2b) — workers reading chunks with zero HDF5 metadata I/O.

## Mechanism options

1. **h5coro B-tree walk (pure Python, deployable today)** — extend the #152 extract mode to also emit `(dataset, chunk_idx, elem_start, elem_end, byte_offset, nbytes)` rows. Zero new dependencies; the deployed layer already runs h5coro; same Lambda fan-out as the geometry extraction.
2. **hidefix serialized index** (#155) — ~0.2 s / ~600 KB per granule, natively consumable by the intended (2b) reader; but it can't run in Lambda until the SlideRuleEarth binding crate ships (#155), and the Python binding's save/load surface doesn't exist yet.
[...]

PR #159, opened 2026-07-02
88S offsets extraction: per-chunk byte offsets for the pinned benchmark shard (issue #158)

espg:
Refs #158, Refs #148.

Per-chunk byte-offset extraction for ATL03 granules — the offset side PR #152 deliberately deferred (its parquet is geometry-only per the decision on the #148 thread). Emits one parquet row per HDF5 chunk of every heights dataset plus the geolocation link arrays:

```
(granule, beam, dataset, chunk_idx, elem_start, elem_end, byte_offset, nbytes, filter_mask)
```

Self-contained under `bench/offsets/` (base branch `bench`, matching PR #150's placement; no imports from unmerged PR #150/#152 code — the parquet conventions of `src/zagg/catalog/extract.py` on the #152 branch are mirrored, not imported).

## Approach

Two independent extraction routes, hard-gated against each other **and** against hidefix's index (option 3 "both, cheaply" from #158):

- **route (a) h5py** — `dset.id.chunk_iter` / `get_chunk_info`: libhdf5's own answer, reference only (h5py is in the dev venv, not the Lambda layer).
- **route (b) h5coro** — pure-Python chunk B-tree walk driven through `H5Dataset` internals (`metaOnly` metadata parse, then `readField`/`readBTreeNodeV1` over the v1 B-tree). This is the Lambda-deployable route the #152 `mode="extract"` handler would grow. The walk enumerates every leaf entry, so the h5coro hyperslice off-by-one found in #148 is not on this path.

Neither route decompresses any chunk — offsets extraction is metadata-only I/O (~30–80× cheaper than the #152 boundary-geometry scan, which must inflate every lat/lon chunk).

## Phases

- [x] Phase 1 — extractor, both routes + parquet schema + CLI (`bench/offsets/extract_offsets.py`)
- [x] Phase 2 — hidefix cross-validation gate (`bench/offsets/crosscheck_hidefix.py`): **61 NEON granules / 184,616 chunks, zero mismatches** across h5py vs h5coro vs hidefix (report: `bench/offsets/results/crosscheck_neon.json`)
- [x] Phase 3 — 88S sample (`bench/offsets/fetch_88s_sample.py`): 3 granules from the pinned o9 shard (first/middle/last of the 5,620-granule list — 2018/2022/2025), HTTPS/EDL from a workstation. **15,472 chunks, all gates green**, including HTTPS-streamed vs local route (b). Streamed parquets + report committed under `bench/offsets/results/`.
- [x] Phase 4 — artifacts + README: combined NEON offsets parquet (184,616 rows, 1.3 MB gzip), per-granule timings, manifest layout proposal `<prefix>/offsets/<granule_id>.offsets.parquet`, hidefix-index convergence plan (now referencing the `espg/h5coro-hidefix` binding @ `93d02ba`: `Index.save/load`, `chunks()`, `read_plan()`, `read_from_buffers()`)

## Key findings

- **Streaming offsets extraction costs one 4 MiB ranged GET per granule.** NSIDC's repacked files keep object headers + chunk B-trees in the first 4 MiB, inside h5coro's first cache line — route (b) over HTTPS runs in ~4–5 s wall (TLS/EDL-dominated), never touching the multi-GB granule. Standalone-extraction cost model: **~$0.000034/granule** → pinned o9 shard ≈ $0.19, full ~500k catalog ≈ $17 (vs #152's $75–120 for the geometry scan; piggybacked on it, offsets are ~free).
- Local extraction: 0.02–0.03 s/granule/route (mean; max 0.06 s) on the 61-granule NEON cache — vs 0.54–2.46 s/granule for the #152 geometry pass.
- hidefix chunk records decode as little-endian u64 `(addr, size, offset[D])`; verified byte-identical to libhdf5's `get_chunk_info` on every chunk checked (199,000+ total incl. 88S).

## 88S wall-clock note

The three-granule 88S run finished all downloads and gates (~25 min: 458/848/206 s per download over the workstation link) just as the abort-for-wall-clock call arrived, so the completed N=3 results are kept as committed. The download leg existed only so h5py + hidefix could arbitrate the streamed output; the at-scale validation belongs to the in-region Lambda fan-out (S3 fetch in seconds), gated on the #152/#153 deploys. EDL auth + pinned-shard URL resolution are proven either way. Downloads were temp-dir only and deleted on exit (EOSDIS no-redistribute); only offsets parquets and reports are committed — verified zero granule data left on disk.

## Testing

The cross-validation gate is the test: three independent implementations (libhdf5 via h5py, pure-Python B-tree walk via h5coro internals, hidefix's Rust indexer) agree chunk-for-chunk on `(elem_start, elem_end, byte_offset, nbytes)` over 64 granules (61 NEON + 3 88S), plus streamed-vs-local equality; `filter_mask` is additionally compared between the two extractor routes on the 61 NEON granules (the committed 88S report predates that strict compare — the script runs it now, and `filter_mask` is 0 on all 15,472 committed 88S chunks; see the README honesty note). `pre-commit run --files bench/offsets/*` green (mypy scopes to src|tests, so bench/ is ruff+codespell).

## Questions for review

- `filter_mask` is an extra column beyond the #158 schema sketch — kept because both h5py and the B-tree carry it for free and a nonzero mask (filter skipped on a chunk) would silently corrupt a raw ranged-GET reader; drop it if you want the schema minimal.
- The cross-check tooling parses `hfxidx`'s flexbuffers dump on the host, which needs the pure-Python `flatbuffers` package on `PYTHONPATH` (bench-tool-only, not a zagg dependency — same posture as PR #150's podman-contained deps). Flagging per the dependency rule even though nothing in `pyproject.toml` changes.


bench/offsets/extract_offsets.py
```diff
@@ -0,0 +1,335 @@
...
+    for beam in beams:
+        for dataset in datasets:
+            path = f"/{beam}/{dataset}"
+            try:
+                ds = H5Dataset(h5obj, path, earlyExit=True, metaOnly=True, enableAttributes=False)
+            except Exception:
+                if dataset == datasets[0]:
+                    missing.append(beam)
+                    break
+                raise
```
espg: 🤖 *from Claude (review)*

**`missing_beams` detection on this route is dead code, and the broad `except Exception` misclassifies real errors as missing beams.**

`H5Dataset(..., metaOnly=True)` never raises for a nonexistent path: the symbol-table/link traversal simply finds no matching name and returns, and the sanity checks that would raise (`missing data type information for ...` etc., `h5dataset.py` ~lines 177–182) sit *after* the `if self.metaOnly: return` early-exit. The constructor comes back with a default `H5Metadata` (`dimensions=[]`, `typeSize=0`), so a missing beam group falls through to `dims == ()` and is silently `continue`d for all eight datasets — `missing_beams` stays `[]`. That diverges from the h5py route, which records the beam in `missing` (line 132–134) and raises `KeyError` for a missing dataset inside an existing beam.

Conversely, the only things this `except Exception` *can* catch are real failures (corrupt-metadata `FatalError`, driver I/O errors) — and when they happen on `datasets[0]` they get swallowed and relabeled "missing beam".

Suggest detecting missing from the metadata after construction (e.g. `ds.meta.typeSize == 0` / `ds.meta.layout is None` → first dataset ⇒ missing beam, otherwise a missing-dataset error matching route (a)'s `KeyError`), and letting exceptions propagate. Note `ds.datasetFound` is *not* a reliable probe here: it's only set on the new-style link-message path (`h5dataset.py` line 1044), not the old-style symbol-table traversal ATL03 files use.

bench/offsets/extract_offsets.py
```diff
@@ -0,0 +1,335 @@
...
+            if not dims or 0 in dims:
+                continue
+            if ds.meta.layout == H5Dataset.CHUNKED_LAYOUT:
+                if ds.meta.address == INVALID_VALUE[h5obj.offsetSize]:
+                    continue  # chunked but no allocated chunks
+                chunk_dims = tuple(int(x) for x in ds.meta.chunkDimensions)
+                rows.extend(_chunk_rows(beam, dataset, dims, chunk_dims, _walk_chunk_btree(ds)))
+            else:  # contiguous: one pseudo-chunk, no filters
+                nbytes = ds.meta.typeSize * math.prod(dims)
+                rows.append((beam, dataset, 0, 0, dims[0], ds.meta.address, nbytes, 0))
```
espg: 🤖 *from Claude (review)*

**The contiguous branch is missing the unallocated-storage guard that route (a) has (line 139, `off is None`) and the chunked branch has (line 245).**

For a contiguous dataset with deferred/unallocated storage, `ds.meta.address` is `INVALID_VALUE[offsetSize]` (2^64−1) — `metaOnly=True` returns before h5coro's own `invalid data address` sanity check, so it reaches this branch, and the resulting `byte_offset` then makes `_finalize`'s `astype("int64")` raise `OverflowError` instead of the row being skipped like route (a) does.

Two smaller asymmetries on the same `else`:

1. Compact-layout datasets also land here (`layout != CHUNKED_LAYOUT`) — route (a) skips them (`get_offset()` returns `None` for compact) while this emits a row, which would trip the phase-2 `df.equals` gate if one ever appears. Matching on `H5Dataset.CONTIGUOUS_LAYOUT` explicitly would keep the routes aligned.
2. `ds.meta.size` is the contiguous layout message's own storage size — same value as `typeSize * prod(dims)` but the direct answer, mirroring route (a)'s `get_storage_size()`.

bench/offsets/extract_offsets.py
```diff
@@ -0,0 +1,335 @@
...
+    routes = ("h5py", "h5coro") if args.route == "both" else (args.route,)
+    out_dir = Path(args.out_dir)
+    timings: list[dict] = []
+    for path in paths:
+        gid = path.name
+        for route in routes:
+            if route == "h5py":
+                df, meta = extract_offsets_h5py(str(path), gid)
+            else:
+                h5obj = _open_h5coro(str(path))
```
espg, line 327: 🤖 *from Claude (review)*

Minor: one fd leaks per granule — the `H5Coro` object (FileDriver holds the file open) is never closed, and `h5coro.H5Coro` has a `.close()`. Harmless for the phase-3 2–3-granule sample, and the 61-granule crosscheck stays under default soft limits, but a full-shard local run (5,620 granules) would exhaust them (cf. the fd-cap handling from issue #142). A `try/finally: h5obj.close()` around the extract call covers it.

bench/offsets/extract_offsets.py
espg: 🤖 *from Claude*

Folded in 4b2ac2b. Replaced the dead try/except with an explicit `ds.meta.typeSize == 0` absent-path check (metaOnly leaves default metadata for a path the traversal never finds): a missing first dataset records the beam in `missing_beams` and breaks; a missing dataset inside a present beam raises `KeyError`, matching route (a). Verified: fake beam -> `missing_beams=["gtXX"]`, fake dataset -> `KeyError("/gt1l/heights/nope")`, routes still `df.equals` on a real granule. Agreed on `datasetFound` being unsafe for symbol-table groups — not used.

bench/offsets/extract_offsets.py
espg: 🤖 *from Claude*

Folded in 4b2ac2b. The `INVALID_VALUE` unallocated-storage guard now sits ahead of the layout branch (mirrors route (a)'s `get_offset() is None` skip), the contiguous row uses `ds.meta.size` (the layout message's storage size), and COMPACT layout is explicitly skipped with a comment (data lives in the object header; h5py skips it too), so the phase-2 `df.equals` gate can't diverge there.

bench/offsets/extract_offsets.py
espg: 🤖 *from Claude*

Folded in 4b2ac2b: the CLI closes each granule's `H5Coro` in a `finally` (FileDriver holds the fd), so a full 5,620-granule run can't exhaust the soft fd limit.

bench/offsets/crosscheck_hidefix.py
```diff
@@ -0,0 +1,199 @@
...
+        df_b, _ = read_offsets_parquet(offsets_dir / "h5coro" / pq.name)
+        fx = fx_dir / (gid.removesuffix(".h5") + ".fx")
+        if not fx.exists():
+            if not args.make_index:
+                raise SystemExit(f"missing {fx} (pass --make-index to generate)")
+            make_index(Path(args.granule_dir).expanduser() / gid, fx, args.image)
+        # hidefix indexes every dataset under a beam (bckgrd_atlas/ etc.);
+        # the gate covers the read-path datasets the extractor scopes to.
+        df_h = parse_hidefix_fx(fx)
+        df_h = df_h[df_h["dataset"].isin(set(df_a["dataset"]))]
```
espg: 🤖 *from Claude (review)*

**Gate weakening: filtering `df_h` to `set(df_a["dataset"])` blinds the outer join's `right_only` arm to any dataset our extractor missed entirely.**

The outer join + `indicator` in `compare` is what upgrades this from "offsets match where we produced rows" to "coverage matches chunk-for-chunk". But pre-filtering hidefix's frame to the datasets *observed* in `df_a` means a target dataset that route (a) dropped wholesale — via a skip path in `extract_offsets.py` (`get_offset() is None`, `not dims or 0 in dims`) or a future regression — vanishes from both hidefix comparisons instead of surfacing as `right_only`. Degenerate case: an empty `df_a` (zero-chunk parquet) filters `df_h` to empty and all three comparisons pass trivially (`n_chunks=0`, `ok=true`).

Filter by intent rather than observation: `df_h[df_h["dataset"].isin(DEFAULT_DATASETS)]` (importable from `extract_offsets`) still drops `bckgrd_atlas/*` etc. but lets the join flag a read-path dataset only hidefix saw. Same class of weakness one level up at line 155: the sweep is driven solely by the `h5py/*.offsets.parquet` glob, so a granule whose extraction crashed or never ran is silently outside the gate — the report cannot distinguish "61 = all cached granules" from "61 = all that produced parquets". Enumerating `--granule-dir` and requiring a parquet per `.h5` would close that.

bench/offsets/crosscheck_hidefix.py
```diff
@@ -0,0 +1,199 @@
...
+from pathlib import Path
+
+import pandas as pd
+from extract_offsets import OFFSETS_DTYPES, _chunk_rows, read_offsets_parquet
+
+PODMAN_IMAGE = "zagg-bench-hidefix"
+
+#: Join key + compared values (everything the arm-(2b) reader consumes).
+KEY_COLS = ["beam", "dataset", "chunk_idx"]
+VALUE_COLS = ["elem_start", "elem_end", "byte_offset", "nbytes"]
```
espg, line 45: 🤖 *from Claude (review)*

**`filter_mask` is never validated by the gate: `VALUE_COLS` excludes it, so the h5py-vs-h5coro comparison skips the one column those two routes both carry.**

Excluding it is right for the two hidefix comparisons (its index does not store the mask — line 103 hardcodes `0`), but `compare(df_a, df_b, "h5py-vs-h5coro")` uses the same `VALUE_COLS`, so route (b)'s parse of the B-tree key's mask (`curr["filter_mask"]` in `_walk_chunk_btree`) is unchecked across the 61-granule sweep — the only coverage is the single-granule `df.equals` smoke from phase 1. The PR body's own rationale for keeping the column ("a nonzero mask … would silently corrupt a raw ranged-GET reader") is exactly why the a-vs-b pair should compare it. Cheap fix: give `compare` a `value_cols` parameter and pass `VALUE_COLS + ["filter_mask"]` for the a-vs-b call only.

bench/offsets/crosscheck_hidefix.py
```diff
@@ -0,0 +1,199 @@
...
+            f"{fx_path.parent}:/out",
+            image,
+            "hfxidx",
+            f"/data/{granule.name}",
+            f"/out/{fx_path.name}",
+            "--out-type",
+            "flexbuffers",
+        ],
+        check=True,
+        capture_output=True,
```
espg, line 72: 🤖 *from Claude (review)*

**`check=True` + `capture_output=True` swallows hfxidx's stderr on failure.** `CalledProcessError.__str__` reports only the command and exit status; the captured stderr sits unprinted on the exception object, so a failing `podman run` (image not built, volume mount denied, corrupt granule, hfxidx panic) kills the sweep with no diagnostic. Either drop `capture_output=True` — a gate tool being noisy on failure is fine — or wrap it:

```python
try:
    subprocess.run([...], check=True, capture_output=True)
except subprocess.CalledProcessError as e:
    raise RuntimeError(f"hfxidx failed for {granule.name}: {e.stderr.decode(errors=\"replace\")}") from e
```

bench/offsets/crosscheck_hidefix.py
```diff
@@ -0,0 +1,199 @@
...
+    bad = m["_merge"] != "both"
+    for c in VALUE_COLS:
+        bad |= m[f"{c}_a"] != m[f"{c}_b"]
+    mismatches = m[bad]
+    return {
+        "comparison": label,
+        "n_chunks": int(len(m)),
+        "n_mismatch": int(len(mismatches)),
+        "ok": len(mismatches) == 0,
+        "mismatch_sample": mismatches.head(20).to_dict(orient="records"),
```
espg: 🤖 *from Claude (review)*

**On the failure path, `mismatch_sample` can make the committed report invalid JSON.** A `left_only`/`right_only` join miss leaves the other side's value columns as float64 `NaN`; `to_dict(orient="records")` preserves them and `json.dumps` (default `allow_nan=True`) emits a bare `NaN` token, which strict JSON parsers reject (`jq`, most non-Python readers, `json.loads` with a strict `parse_constant`). Reproduced with pandas 2.2.3:

```
"elem_start_b": NaN,
```

So precisely when there are mismatches to document, `--report-out` may be unloadable by anything but Python. Cheap fix: `json.loads(mismatches.head(20).to_json(orient="records"))` — pandas' `to_json` serializes `NaN` as `null` (and handles the `_merge` categorical).

bench/offsets/crosscheck_hidefix.py
espg: 🤖 *from Claude*

Folded in b075d6f: the hidefix frame is now filtered by intent (`DEFAULT_DATASETS`) so a dataset our side dropped surfaces as `right_only`, and coverage is judged against the `--granule-dir` cache — any cached granule without an offsets parquet is listed under `uncovered` and fails the run. Regenerated report: 61/61 covered, still 0 mismatches.

bench/offsets/crosscheck_hidefix.py
espg: 🤖 *from Claude*

Folded in b075d6f: `compare()` takes `value_cols`, and the h5py-vs-h5coro pair now compares `STRICT_VALUE_COLS` (incl. `filter_mask`); hidefix pairs stay on `VALUE_COLS` since its index does not store the mask. Full 61-granule gate re-run green with the mask compared.

bench/offsets/crosscheck_hidefix.py
espg: 🤖 *from Claude*

Folded in b075d6f: `make_index` now runs with `check=False` and raises a `RuntimeError` carrying hfxidx/podman stderr on a non-zero exit.

bench/offsets/crosscheck_hidefix.py
espg: 🤖 *from Claude*

Folded in b075d6f: `mismatch_sample` goes through a `to_json(orient="records")` round-trip so join-miss NaNs serialize as `null` — the failure-path report is strict-JSON now. Thanks for reproducing it on the repo pandas pin.

bench/offsets/fetch_88s_sample.py
```diff
@@ -0,0 +1,171 @@
...
+                rec["local_wall_s"] = {"h5py": meta_a["wall_s"], "h5coro": meta_b["wall_s"]}
+                results = [
+                    compare(df_a, df_b, "h5py-vs-h5coro"),
+                    compare(df_s, df_b, "https-vs-local-h5coro"),
+                ]
+                if not args.skip_hidefix:
+                    fx = Path(tmp) / f"{gid.removesuffix('.h5')}.fx"
+                    make_index(local, fx)
+                    df_h = parse_hidefix_fx(fx)
+                    df_h = df_h[df_h["dataset"].isin(set(df_a["dataset"]))]
```
espg: 🤖 *from Claude (review)*

**The 88S gate is weaker than the NEON gate on the two exact points b075d6f hardened in `crosscheck_hidefix.py`** ("fold review: intent-scoped gate, filter_mask in a-vs-b"):

1. **`filter_mask` is never compared for the 88S granules.** Lines 148–149 call `compare(...)` with the default `value_cols=VALUE_COLS`; `crosscheck_hidefix.main` passes `value_cols=STRICT_VALUE_COLS` for the h5py-vs-h5coro pair. So the PR body's "plus `filter_mask` between the two extractor routes … over 64 granules" actually only held for the 61 NEON granules, and the README's "the HTTPS-*streamed* route-(b) frame identical to the local one" is equality on `(elem_start, elem_end, byte_offset, nbytes)` only.
2. **The hidefix frame is observation-scoped, not intent-scoped.** `df_h[df_h["dataset"].isin(set(df_a["dataset"]))]` filters by whatever route (a) emitted — the pattern the fold replaced with `df_h["dataset"].isin(DEFAULT_DATASETS)` precisely so a dataset dropped by both routes surfaces as `right_only` instead of vanishing from the join.

Both are one-line fixes (import `DEFAULT_DATASETS` and `STRICT_VALUE_COLS` from the crosscheck module). Note the committed `report_88s.json` was produced by the weaker gate, so strictly the strict-gate claim for 88S needs a re-run (stream leg is ~4 s/granule; the download leg is the expensive part — flagging, not demanding).

bench/offsets/fetch_88s_sample.py
```diff
@@ -0,0 +1,171 @@
...
+    return df, meta, transfer
+
+
+def download(url: str, token: str, dest_dir: Path) -> tuple[Path, float]:
+    """Fetch one granule over HTTPS with the EDL token; returns (path, wall_s)."""
+    import requests
+
+    dest = dest_dir / url.rsplit("/", 1)[-1]
+    t0 = time.time()
+    with requests.get(url, headers={"Authorization": f"Bearer {token}"}, stream=True) as r:
```
espg: 🤖 *from Claude (review)*

Two robustness points on the download leg:

1. **No `timeout` on `requests.get`** — a stalled connection hangs the run forever (requests defaults to no timeout). The stream side inherits h5coro's `HTTPDriver` `timeout=30`; this leg has nothing. A `timeout=(10, 60)` (connect, per-read — `iter_content` resets it per chunk) keeps the multi-minute transfers working while bounding a dead socket.
2. **A hard kill leaves the granule on disk.** `TemporaryDirectory` cleans up on normal exit and exceptions, but not on SIGKILL (or unhandled SIGTERM) — a mid-download abort strands a multi-GB EOSDIS granule under `$TMPDIR`. Given the docstring and README lean on "downloads are temp-dir only and deleted on exit" for the no-redistribute posture, worth a one-line caveat here (the PR-body note that this run was verified clean covers the run, not the mechanism).

On the token: passing the EDL bearer per-request is fine — requests strips `Authorization` on the cross-host redirect to the presigned S3 URL (`Session.rebuild_auth`), which is also why the presigned leg works at all. No leak; maybe worth a comment since the correctness of this leg silently depends on that stripping behavior.

bench/offsets/fetch_88s_sample.py
```diff
@@ -0,0 +1,171 @@
...
+    line — where NSIDC's repacked files keep most of the metadata) is
+    counted."""
+
+    class Counting(base):
+        def read(self, pos, size):
+            stats["n_gets"] += 1
+            stats["bytes"] += size
+            return super().read(pos, size)
+
+    return Counting
```
espg, line 77: 🤖 *from Claude (review)*

Minor / latent: the counting subclass is not **copy-safe**. h5coro's `HTTPDriver.copy()` hardcodes the base class (`new_driver = HTTPDriver(self.resource, self.cached_credentials, ...)` in `webdriver.py`), and h5coro's multiprocess `BTreeReader` re-drives all reads through `driver.copy(max_connections=5)` (`h5dataset.py`). Today that path is unreachable here — `H5Coro` defaults `multiProcess=False` and the metaOnly B-tree walk never calls `readDataset` — so the reported `n_gets=1` / `mb_transferred=4.0` are accurate. But if either assumption changes, GETs silently escape the counter and the headline "one 4 MiB GET" claim would under-report with no error. Suggest either overriding `copy()` in `Counting` to return another `Counting`, or a one-line comment pinning the single-process/metaOnly assumption.

(Related nit: `stats["bytes"] += size` counts *requested* bytes, not received — same value here since the 4 MiB range is fully inside a multi-GB object, but `len(result)` would be the honest metric.)

bench/offsets/README.md
```diff
@@ -0,0 +1,134 @@
...
+- `<granule_id>` is the `.h5` basename without extension, mirroring #152's
+  `<granule_id>.boundaries.parquet` convention; the two caches join on
+  `(granule, beam, chunk_idx)` — boundaries answer *which* chunks a shard
+  needs, offsets answer *where the bytes are*.
+- Per-granule parquet metadata (key `zagg:offsets_meta`) carries provenance
+  (`route`, `wall_s`, `schema_version`, `missing_beams`).
+- Per-granule files (~50–90 KB) keep the Lambda fan-out write pattern of
+  #152 (one worker → one object, no coordination); a per-shard rollup is a
+  cheap post-pass if GET count ever matters. The committed
+  `results/neon_offsets.parquet` is such a rollup (61 granules, 184,616
```
espg, line 118: 🤖 *from Claude (review)*

Reproducibility gap: two committed artifacts have no generating command anywhere in the Workflow section.

1. **`results/neon_offsets.parquet`** — its custom metadata (`{"source": ..., "n_granules": 61, "n_chunks": 184616, "schema_version": 1}`) is not what `write_offsets_parquet` emits (different keys entirely), so the rollup came from an ad-hoc uncommitted snippet. A one-liner here (or a `--rollup-out` flag on `extract_offsets.py`) would make it regenerable. (Content itself checks out: 184,616 rows, 61 granules, gzip codec, offsets-only columns.)
2. **`results/offsets_88s_sample/*.parquet`** — step 3 writes them to `<scratch>/out_88s/h5coro-https/`; the copy into `results/` is undocumented.

Related, since this section proposes the provenance keys: the committed 88S parquets record `route: "h5coro"` with nothing distinguishing HTTPS-streamed from local extraction — the "streamed" claim is currently only verifiable by matching `wall_s` against `report_88s.json` (it does match: 0.061/0.064/0.048). A `driver` (or `transport`) field in `zagg:offsets_meta` would pin it.

bench/offsets/README.md
```diff
@@ -0,0 +1,134 @@
...
+# 3. 88S sample (EDL creds via ~/.netrc; downloads are temp-dir only and deleted)
+PYTHONPATH=<dir with flatbuffers> python bench/offsets/fetch_88s_sample.py \
+    --out-dir <scratch>/out_88s --report-out bench/offsets/results/report_88s.json
+```
+
+## Results
+
+### Cross-validation gate (NEON cache, 61 granules)
+
+**61 granules × 184,616 chunks, zero mismatches** — `(elem_start, elem_end,
```
espg: 🤖 *from Claude (review)*

Wording nit in a headline claim: "**61 granules × 184,616 chunks**" reads as a product (~11.3M). It's 184,616 chunks *total across* 61 granules — the PR body's "61 NEON granules / 184,616 chunks" phrasing is the right one. (Numbers themselves verify: `timings_neon.json` has 61 h5coro entries summing to exactly 184,616 chunks, and the mean/median/max table below matches the committed timings to the millisecond.)

bench/offsets/fetch_88s_sample.py
espg: 🤖 *from Claude*

Folded in a432bab: both compares now pass `STRICT_VALUE_COLS` (incl. `filter_mask`) and the hidefix frame is intent-scoped to `DEFAULT_DATASETS`, matching the b075d6f hardening in `crosscheck_hidefix.main`. On the committed report: it predates this fix and its compares covered `(elem_start, elem_end, byte_offset, nbytes)` only — re-running would mean re-downloading ~5 GB, so instead the README carries an explicit honesty note, plus the verifiable fact that `filter_mask == 0` on all 15,472 chunks of the committed streamed parquets. The strict gate is what the script runs from now on.

bench/offsets/fetch_88s_sample.py
espg: 🤖 *from Claude*

Folded in a432bab: `requests.get` now has `timeout=(10, 60)` (connect / per-read), and the docstring documents both the SIGKILL-strands-tempdir caveat (with the check to run after any hard abort) and the cross-host `Authorization`-strip behavior that the presigned-S3 redirect silently relies on.

bench/offsets/fetch_88s_sample.py
espg: 🤖 *from Claude*

Folded in a432bab: the subclass now overrides `copy()` to return `type(self)(...)` so reads cannot escape the counter on any driver-copying path, and the byte counter is annotated as counting *requested* sizes (== received for a 206 range hit). Agreed it is latent-only today (metaOnly, single-process), hence fix-with-comment rather than restructure.

bench/offsets/README.md
espg: 🤖 *from Claude*

Folded in a432bab: the manifest section now documents how `neon_offsets.parquet` is built (route-(b) frames concatenated, fastparquet gzip, meta under `zagg:offsets_meta`) and the provenance chain for the 88S sample parquets (streamed leg; `stream.extract_wall_s` in `report_88s.json` equals each parquet's meta `wall_s`). A `transport` meta field is noted as the refinement if this graduates from bench tooling — kept out now to avoid a third schema variant across already-committed artifacts.

bench/offsets/README.md
espg: 🤖 *from Claude*

Folded in a432bab: reworded to "61 granules, 184,616 chunks total".

espg:
🤖 *from Claude*

Status: all four phases complete, CI green on `a432bab`, all 12 adversarial-review findings across three passes folded (each thread has a fold commit + reply). Marking ready for review and applying `waiting`.

Run summary: extractor routes (a) h5py + (b) pure-Python h5coro B-tree walk; three-way gate green on 61 NEON granules / 184,616 chunks and on 3 pinned-88S granules / 15,472 chunks (incl. HTTPS-streamed vs local); headline finding — streamed offsets extraction costs **one 4 MiB ranged GET** per granule (~4–5 s over EDL), so a standalone offsets pass over the full ~500k-granule catalog models at ~$17. Manifest proposal + hidefix-index convergence plan (now referencing the `espg/h5coro-hidefix` binding) in `bench/offsets/README.md`. The 88S workstation downloads (arbitration only) completed just before the abort call, are deleted, and the at-scale validation is noted as belonging to the in-region Lambda fan-out gated on the #152/#153 deploys.

PR #162, opened 2026-07-04
Fix coarse-level filter KeyError on missing hyperslice key (issue #157)

### LINKED ISSUE #157: Coarse-level filter reads KeyError on missing 'hyperslice' key and are silently swallowed (0-obs shard)
🤖 *from Claude*

Surfaced by the PR #150 phase-1 adversarial review while tracing capture fidelity; verified against both codebases before filing. Opened at espg's direction (PR #150 follow-ups).

## The bug

The coarse-level structured-filter path passes dicts **without a `"hyperslice"` key** to `readDatasets`:

- `src/zagg/processing/read.py:651` — `h5obj.readDatasets([{"dataset": flag_path}])`
- `src/zagg/processing/read.py:659-664` — `h5obj.readDatasets([{"dataset": ibeg_path}, {"dataset": cnt_path}])`

h5coro 1.0.4 normalizes only *string* entries to `{"dataset": ..., "hyperslice": []}`; dict entries pass through unchanged (`h5coro/h5coro.py:161-167`). `H5Promise.__init__` then indexes unconditionally — `dataset["hyperslice"]` at `h5coro/h5promise.py:80` — so each of these calls raises **`KeyError: 'hyperslice'`**.

## Why it's silent

The `KeyError` propagates out of `_read_group` and is caught by the per-group `except Exception` in `worker.py` (~L250), which logs `Error reading track <g>` at WARNING and continues. A config that exercises this path reads **zero observations for every beam** while the run itself "succeeds" — the same failure signature class as #116 (the dem_h broadcast failure that hid behind logging).

## Why nothing has hit it yet

The path is the issue #43 Phase-B coarse-level filter expansion; it only runs when a filter's `level` resolves to a non-base level (`coarse_structured` non-empty). The benchmark/production configs to date filter at base level (`/{group}/heights/signal_conf_ph` photon-rate), so the branch is dormant — and unit tests that mock `readDatasets` never execute h5coro's real normalization, which is the test gap that let it in.

## Fix

Two-sided, both trivial:

1. **zagg (this issue):** add `"hyperslice": []` to the three dict literals — `[]` matches h5coro's own normalization of a full read (`None` is *not* what string entries normalize to). Plus a regression test that runs the coarse-filter path against the real `readDatasets` entry normalization (a tiny fixture file or a monkeypatch-free unit around `H5Promise` input handling), so mocked reads can't hide it again.
2. **h5coro (optional hardening, upstream):** `dataset.setdefault("hyperslice", [])` in the dict branch of `readDatasets` — one line; could ride alongside the pending shuffle PR (SlideRuleEarth/h5coro#44) or go separately.

Sibling upstream item for awareness, not this issue: PR #152's phase 2 found a *different* h5coro hyperslice defect (chunk-aligned hyperslice start drops the dataset read — off-by-one), worked around and regression-tested there; both belong in the same upstream conversation.

Refs: PR #150 (review note in the fold summary), #43 (the phase-B path), #116 (the silent-zero-read precedent), SlideRuleEarth/h5coro#44.

*(Left unlabeled for espg to triage — `small-fix` shaped.)*

espg:
Closes #157.

## Small-fix checklist

- [x] #157 — coarse-level filter reads `KeyError` on missing `'hyperslice'` key, silently swallowed into a 0-obs shard.

## What / approach

h5coro 1.0.4's `readDatasets` normalizes only *string* entries to `{"dataset": ..., "hyperslice": []}`; dict entries pass through unchanged, and `H5Promise.__init__` indexes `dataset["hyperslice"]` unconditionally. The coarse-level structured-filter branch of `_read_group_full` passed three dicts without the key — the coarse flag read and the two link-array reads in `src/zagg/processing/read.py` — so any config filtering at a non-base level raised `KeyError: 'hyperslice'`, which the worker's per-group `except Exception` swallowed into a WARNING and a silent zero-observation shard (same failure signature class as #116).

Fix, per the issue's proposal (1): add `"hyperslice": []` to the three dict literals — `[]` is h5coro's own normalization of a full read:

```python
coarse_data = h5obj.readDatasets([{"dataset": flag_path, "hyperslice": []}])
...
link_data = h5obj.readDatasets(
    [
        {"dataset": ibeg_path, "hyperslice": []},
        {"dataset": cnt_path, "hyperslice": []},
    ]
)
```

These were the only bare-dict `readDatasets` call sites in `src/` (all other sites pass strings or dicts that already carry `hyperslice`).

## Closing the test gap

The bug survived because the test stubs used a tolerant `d.get("hyperslice")`, so mocked reads never exercised h5coro's real entry normalization. This PR makes the stubs replicate the strict contract (verified against installed h5coro 1.0.4 source):

- `_FakeH5.readDatasets`, `_serve_datasets`, and `_ViewBackedH5.readDatasets` in `tests/test_processing.py` now index `d["hyperslice"]` unconditionally on dict entries (`KeyError` on a missing key, `[]` = full read), mirroring `H5Promise`.
- With the strict stubs, the existing `TestReadGroupCrossLevel` cases become live regression guards for this fix — before the src change they fail with `KeyError: 'hyperslice'` on exactly the three reads the issue identified (reproduced locally).
- New `TestCoarseFilterHypersliceKey` pins the stubs' strictness (dict-without-key raises for both stub bodies; `"hyperslice": []` is a full read), so a tolerant `.get` can't creep back in and re-hide the bug.

The optional upstream hardening (`setdefault("hyperslice", [])` in h5coro's `readDatasets`) is out of scope here per the issue — it stays a candidate one-liner for a future SlideRuleEarth/h5coro PR.

## How tested

- Reproduced first: with the strict stubs and the src fix reverted, `TestReadGroupCrossLevel` fails with `KeyError: 'hyperslice'` (3 failures, matching the issue's read sites); green with the fix.
- Full suite: `uv run pytest -v` — **1100 passed, 24 skipped, 0 failed** (Python 3.12, h5coro 1.0.4).
- `uv run ruff check --select=E,F,W,I --ignore=E501 src tests` (the PR lint-bot gate) — clean.
- `uv run ruff format --check` — both touched files clean. Pre-existing on `main` and untouched here: one `N818` in `src/zagg/registry.py` under the full local select (already flagged in PR #152), and format drift in 6 unrelated files (identical result with this diff stashed).

## Questions for review

- None blocking. One judgment call: the stub strictness change also applies to `_ViewBackedH5` (the issue #66 cache-eviction test), not just the coarse-path stubs — done for a uniform stub contract so no `readDatasets` stub in the file tolerates a missing key.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Pvu3umLej7woxRE8uKUu2a)_

### DIFF
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -2989,7 +2989,11 @@ class TestVectorChunkResolutionCompanion:
 class _FakeH5:
     """Stub h5coro object: ``readDatasets`` returns canned arrays by path.
 
-    Honors the ``hyperslice`` bound so sliced reads mirror the real driver.
+    Honors the ``hyperslice`` bound so sliced reads mirror the real driver,
+    including h5coro's strict dict contract: only *string* entries are
+    normalized to ``{"dataset": ..., "hyperslice": []}``; dict entries must
+    carry ``"hyperslice"`` themselves (``H5Promise`` indexes it
+    unconditionally — issue #157), and ``[]`` means a full read.
     """
 
     def __init__(self, arrays):
@@ -3003,8 +3007,8 @@ class _FakeH5:
                 continue
             path = d["dataset"]
             arr = self._arrays[path]
-            hs = d.get("hyperslice")
-            if hs is not None:
+            hs = d["hyperslice"]  # KeyError on missing key, as in real h5coro
+            if hs:
                 lo, hi = hs[0]
                 arr = arr[lo:hi]
             out[path] = arr
@@ -3482,6 +3486,36 @@ class TestReadGroupCrossLevel:
         }
         df = _read_group(h5, "gt1l", ds, 0, _ShardGrid())
         assert df["h"].tolist() == [1.0, 3.0]
+
+
+class TestCoarseFilterHypersliceKey:
+    """Issue #157: coarse-filter reads must pass dict entries WITH ``"hyperslice"``.
+
+    h5coro 1.0.4 normalizes only *string* entries to ``{"dataset": ...,
+    "hyperslice": []}``; dict entries pass through unchanged and ``H5Promise``
+    indexes ``dataset["hyperslice"]`` unconditionally, so a dict without the
+    key raises ``KeyError`` (swallowed by the worker's per-group except into a
+    silent 0-obs shard). The stubs here replicate that strict contract, so the
+    :class:`TestReadGroupCrossLevel` cases above regression-guard the fix;
+    these tests pin the stubs' strictness so a tolerant ``.get`` can't creep
+    back in and re-hide the bug behind mocked reads.
+    """
+
+    def test_fake_h5_rejects_dict_without_hyperslice(self):
+        h5 = _FakeH5({"/x": np.arange(3.0)})
+        with pytest.raises(KeyError, match="hyperslice"):
+            h5.readDatasets([{"dataset": "/x"}])
+
+    def test_serve_datasets_rejects_dict_without_hyperslice(self):
+        with pytest.raises(KeyError, match="hyperslice"):
+            _serve_datasets({"/x": np.arange(3.0)}, [{"dataset": "/x"}])
+
+    def test_empty_hyperslice_is_full_read(self):
+        # "hyperslice": [] is h5coro's own normalization of a full read —
+        # exactly what the coarse flag + link-array reads now pass.
+        h5 = _FakeH5({"/x": np.arange(3.0)})
+        out = h5.readDatasets([{"dataset": "/x", "hyperslice": []}])
+        np.testing.assert_array_equal(out["/x"], np.arange(3.0))
 
 
 # ---------------------------------------------------------------------------
@@ -4688,7 +4722,8 @@ def _release_cfg():
 
 def _serve_datasets(arrays, datasets):
     """Shared ``readDatasets`` body: honor the same path/hyperslice contract as
-    :class:`_FakeH5` (mirrors the real h5coro driver)."""
+    :class:`_FakeH5` (mirrors the real h5coro driver, including the strict
+    dict-entry ``"hyperslice"`` requirement — issue #157)."""
     out = {}
     for d in datasets:
         if isinstance(d, str):
@@ -4696,8 +4731,8 @@ def _serve_datasets(arrays, datasets):
             continue
         path = d["dataset"]
         arr = arrays[path]
-        hs = d.get("hyperslice")
-        if hs is not None:
+        hs = d["hyperslice"]  # KeyError on missing key, as in real h5coro
+        if hs:
             lo, hi = hs[0]
             arr = arr[lo:hi]
         out[path] = arr
@@ -4830,7 +4865,7 @@ class TestProcessShardCacheRelease:
                     path = d if isinstance(d, str) else d["dataset"]
                     buf = self._buffers[path]
                     arr = np.frombuffer(buf, dtype=self._dtypes[path])  # view into buffer
-                    if not isinstance(d, str) and d.get("hyperslice") is not None:
+                    if not isinstance(d, str) and d["hyperslice"]:
                         lo, hi = d["hyperslice"][0]
                         arr = arr[lo:hi]
                     out[path] = arr
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -1,3 +1,5 @@
+import gc
+
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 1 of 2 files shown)


tests/test_processing.py
```diff
@@ -3484,6 +3488,36 @@ def test_flat_form_unchanged(self):
         assert df["h"].tolist() == [1.0, 3.0]
 
 
+class TestCoarseFilterHypersliceKey:
```
espg, line 3493: 🤖 *from Claude (review)*

**should-fix (non-blocking):** issue #157's test criterion was a regression test "against the real `readDatasets` entry normalization (a tiny fixture file or a monkeypatch-free unit around `H5Promise` input handling)". This class pins the hand-replicated *stub* contract instead. That does close the mocks-hide-the-bug gap, but nothing in the suite touches installed h5coro, so the stubs can silently drift from the driver they claim to mirror (e.g. if upstream lands the `setdefault("hyperslice", [])` hardening the issue proposes, these stubs keep enforcing a contract the real driver no longer has).

The monkeypatch-free unit the issue named is a verified no-I/O one-liner against h5coro 1.0.4 — the `KeyError` is raised synchronously while `H5Promise.__init__` builds its futures list (`h5promise.py:80`), before any resource access, so `resourceObject=None` works and no fixture file is needed:

```python
from h5coro.h5promise import H5Promise

def test_real_h5coro_rejects_dict_without_hyperslice(self):
    with pytest.raises(KeyError, match="hyperslice"):
        H5Promise(None, {"x": {"dataset": "x"}}, True,
                  earlyExit=True, metaOnly=False, enableAttributes=False)
```

Confirmed locally against the installed 1.0.4: raises `KeyError: 'hyperslice'`, no I/O attempted. Adding it alongside the stub-pinning tests ties the stubs' contract to the real driver — and it would also cover the third strict stub (`_ViewBackedH5`, method-local, so unpinnable by these tests) by pinning the contract at the source instead of per-stub. If reaching into `H5Promise` feels too private, note it's the exact surface the issue proposed pinning.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the issue #157 fix (commit c7a61e4), checked against installed h5coro 1.0.4 source and CLAUDE.md §§2–4. The core fix is correct and every claim in the PR body checked out under independent reproduction. One non-blocking finding posted inline.

**Verified (reproduced, not taken on faith):**

- **Fix semantics** — `readDatasets` normalizes only *string* entries to `{"dataset": ..., "hyperslice": []}` and passes dict entries through unchanged (site-packages `h5coro/h5coro.py`, `readDatasets`); `H5Promise.__init__` indexes `dataset["hyperslice"]` unconditionally (`h5promise.py:80`). So `"hyperslice": []` is exactly the driver's own full-read form. Also checked the `None` edge: `H5Dataset` maps `hyperslice=None` to `[]` (`h5dataset.py:144`), so the stubs' `if hs:` falsy check is faithful to the driver on both counts, not just convenient.
- **Call-site coverage** — grepped `readDatasets` across `src/` at the branch head: the three fixed literals were the only bare-dict entries; every other call passes strings or dicts already carrying `hyperslice`. The planned-read path (`src/zagg/processing/read.py:395`) is untouched, and the coarse-filter branch has no behavioral change beyond the added key.
- **Repro claim** — ran this branch's tests against `main`'s src: exactly 3 failures, all `KeyError: 'hyperslice'`, all in `TestReadGroupCrossLevel` — matching the PR body's "3 failures, matching the issue's read sites". Against the branch's src: green.
- **Suite and lint** — full suite green in this environment (1013 passed / 18 skipped; `test_viz.py` and `test_shardmap.py` fail *collection* here on a missing optional `pyarrow` — environmental, unrelated to this diff). `ruff check --select=E,F,W,I --ignore=E501 src tests` clean; `ruff format --check` clean on both touched files.
- **Conventions** — no new dependencies, no infra touched, `read.py` at 762 lines (under the §4 guidance), one title-only commit per §3, diff terse and scoped to the issue.

**Finding (inline, should-fix, non-blocking):** the issue asked for a regression test against *real* h5coro entry normalization; the PR pins hand-replicated stub strictness only. A verified no-I/O one-liner against `H5Promise` closes that gap — see the inline comment on `TestCoarseFilterHypersliceKey`.

**Note for triage, not attributable to this PR:** `tests/test_processing.py` is now 4,955 lines — roughly 5x the §4 ~1000-line module guidance, and pre-existing on `main`. Flagging rather than fixing per §4; splitting it is a discussion for a separate issue.

---

espg:
🤖 *from Claude*

Folded the review's should-fix finding in 2d67c10 (review: https://github.com/englacial/zagg/pull/162#pullrequestreview-4629090610): added a monkeypatch-free regression test against the **installed** h5coro, `TestCoarseFilterHypersliceKey::test_real_h5promise_requires_hyperslice_on_dict_entries`. It constructs `H5Promise` directly with a `None` resource — safe because the `dataset["hyperslice"]` indexing (1.0.4, h5promise.py:80) happens synchronously, before any resource access — and asserts:

- a bare dict entry (`{"dataset": "x"}`, the pre-fix zagg shape) raises `KeyError: 'hyperslice'`;
- the fixed shape (`{"dataset": "x", "hyperslice": []}`) passes entry handling (the `None` resource then fails downstream, never with the missing-key `KeyError`).

Guarded with `pytest.importorskip("h5coro.h5promise")`. One wrinkle handled: h5coro's half-constructed `H5Dataset` raises in `__del__` when the failed promise is collected, so the test `gc.collect()`s it in place under a `filterwarnings` mark for `PytestUnraisableExceptionWarning`, keeping the unraisable confined to (and silenced in) this test rather than floating into whichever test GC lands on. This closes the issue #157 acceptance criterion of a regression test around h5coro's real entry handling — the stub tests pin the mocked contract, this one pins the driver itself.

Gates: `pytest tests/test_processing.py` 181 passed (targeted class 8/8); PR-bot ruff gate clean; touched files `ruff format`-clean.

---

PR #163, opened 2026-07-04
Virtual chunk-index backends for the read path (issue #160)

### LINKED ISSUE #160: Virtual index entry point: pluggable chunk-index backends for the read path
🤖 *from Claude*

Opened at espg's direction, from the design discussion following PR #159's results. This is the piece that ties #148 arm (2b), #155, and #97 together: a **pluggable chunk-index abstraction for the read path**, where today's hierarchical read becomes the default backend rather than the thing being replaced.

## Motivation

PR #159 established that a granule's element→byte chunk map is metadata — one ~4 MiB ranged GET, ~0.02 s, ~$17 for the full catalog — while PR #152 established that chunk-boundary *geometry* requires one decompression pass, ever. Together they enable a read path that answers "which chunks cross my shard" and "what bytes do I fetch" without per-run geolocation reads or B-tree walks (#148 arm (2b); consumption machinery in `espg/h5coro-hidefix`: `read_plan`/`read_from_buffers`). The question this issue settles is **where that capability plugs in** — cleanly, without forking the pipeline or adding dependencies to zagg core.

The answer follows a pattern zagg already uses three times: grid backends (`src/zagg/grids/`, `from_config` dispatch), catalog intersection backends (mortie / spherely), and — the closest precedent — h5coro registering itself as an xarray engine via a setuptools entry point, gaining xarray a backend without xarray depending on h5coro.

## The protocol

A virtual index answers exactly two questions per (granule, shard, config):

1. **Selection** — which chunks / element ranges of the configured datasets intersect this shard?
2. **Addressing + decode** — produce the column arrays for those ranges (byte-range GETs + buffer decode, or h5coro hyperslice reads — the backend's choice).

It deliberately does **not** own filtering (predicates like `signal_conf_ph` are data-dependent and stay downstream, unchanged) or aggregation. The seam is thin: `_read_group` asks the index for columns; everything after is backend-invariant. That makes PR #150's byte-equality gate the natural **conformance test** for any backend — same captured workload, sha256-identical output required.

## Backends

| backend | ships in | mechanism |
|---|---|---|
| `hierarchical` (default) | zagg | today's path: geolocation read + `plan_read` + h5coro hyperslices — pure refactor behind the protocol, byte-identical |
| `inline` | zagg | build the chunk map on the fly (PR #159 route (b): pure-Python B-tree walk, ~1 GET + 20 ms/granule); optional write-back to the sidecar store (self-populating cache) |
| `sidecar` | `h5coro-hidefix` | granule-keyed sidecar objects on S3 (bincode index ~600 KB + chunk⋈geometry parquet ~40 KB per granule); selection from geometry columns, decode via `read_from_buffers` |
| future: `cmr-geometry`, icechunk/VirtualiZarr consumers | external | slot in via the same registry if/when upstream metadata or archive re-publication materializes (#97) |

Registration via an entry-point group (e.g. `zagg.index_backends`): **zagg core never imports hidefix** — the §4 dependency question reduces to "does the Lambda layer include the wheel," and the `sidecar` backend simply isn't discoverable where it isn't installed.

## Config surface

```yaml
data_source:
  reader: h5coro
  index:
    backend: hierarchical   # default when the block is absent — zero-change upgrade
    # backend: sidecar | inline
    store: s3://<bucket>/zagg-index/ATL03/007/   # backend-specific
    on_miss: fallback       # fallback | error | build
```

`on_miss` makes rollout incremental rather than flag-day: `fallback` uses the hierarchical path for uncovered granules; `build` is `inline` + write-back.

## What the sidecar store is (and is not)

A flat, immutable, **granule-keyed** namespace — `<prefix>/<product>/<version>/<granule_id>.{idx,parquet}` — not a store you open or query. The lookup key (granule id) is already in hand at every consumer (shardmap granule lists, worker URLs), which is why no time-axis/zarr/icechunk indirection is warranted internally (fuller argument on #97). Versioning is free: granule ids carry the product version, so reprocessing changes the key. The index is grid- and AOI-independent — built once per granule, reused by every config forever. Dispatch events carry no index data (workers fetch by convention — relevant to PR #153's 256 KB async cap). A shard-merged materialization (`<shardmap>.index/<shard_key>.parquet`, one GET per worker instead of thousands) is a v2 optimization the flat namespace makes trivial.

## Infrastructure (complete list)

(1) an S3 prefix + naming convention; (2) IAM: `PutObject` for the extraction role, `GetObject` for workers; (3) the population fan-out = PR #152's `mode="extract"` + PR #159's offset columns pointed at the prefix (~$0.19 for the pinned 88S shard, ~$17 catalog-wide); (4) one zagg PR: the protocol + registry + `hierarchical`/`inline` backends (mostly refactoring the existing path behind the seam); (5) the `sidecar` provider in `h5coro-hidefix` (surface already exists). No new stack, database, or service.

## Phasing sketch

1. Protocol + registry + `hierarchical` backend (pure refactor, byte-identical, gated on PR #150's captured-workload equality).
2. `inline` backend (route (b) extractor already written in PR #159's `bench/offsets/`).
3. `sidecar` provider in h5coro-hidefix + the store convention + extraction-to-prefix.
4. 88S benchmark matrix (#148 phase 5) exercises `hierarchical` vs `inline` vs `sidecar` as the (1)/(2a′)/(2b) arms — one mechanism, three policies, same conformance gate.

## Open questions

1. Protocol granularity: per-granule (`plan(granule, shard) → columns`) vs per-(granule, dataset) — the former matches `_read_group`'s shape; recommend it.
2. Should `inline` write-back be on by default (self-populating fleet) or opt-in? Recommend opt-in until the IAM story for worker `PutObject` is settled.
3. Sidecar pair (bincode + parquet) vs single artifact — consolidating on parquet-only needs an `Index.from_chunks()` constructor in the binding (small follow-up there); fine to ship the pair first.
4. Where does the entry-point group name live — `zagg.index_backends` vs something reader-scoped — and does `reader:` eventually fold into the same registry?

Refs: #148 (parent architecture; arm (2b)), #155 (binding crate + dependency discussion), #97 (virtualization reframe — this issue is its "mechanism" answer), #158 / PR #159 (offsets + inline route), PR #152 (geometry + extraction), PR #150 (conformance gate).

*(Left unlabeled for espg to triage — `discuss`/`plan` shaped.)*

espg:
Closes #160.

Pluggable chunk-index backends for the read path, per the design settled on the issue thread ([protocol + backends](https://github.com/englacial/zagg/issues/160), [corrected config semantics + Q1–Q4 decisions](https://github.com/englacial/zagg/issues/160#issuecomment-4880966830), [store decision](https://github.com/englacial/zagg/issues/160#issuecomment-4880988678)). Today's hierarchical read becomes the **default backend** rather than the thing being replaced; the seam matches `_read_group`'s shape (issue Q1, decided).

## What / approach

- **`src/zagg/index/`** — new package: the `VirtualIndex` protocol (`read_group` per (granule, group) + a `finish_granule` per-granule hook for side effects like write-back), a backend registry (builtins in a static dict merged with `zagg.index_backends` entry points — issue Q4; mirrors the h5coro-as-xarray-engine precedent, so zagg core never imports an external backend's deps), config validation, and `index_from_config` resolution.
- **`hierarchical` backend** (`src/zagg/index/hierarchical.py`) — pure delegation to `zagg.processing._read_group`, resolved through the package namespace at call time so every existing `monkeypatch.setattr("zagg.processing._read_group", ...)` test still intercepts reads. Byte-identical by construction; an absent `data_source.index` block resolves to it (zero-change upgrade).
- **Config surface** (`src/zagg/config.py`) — `data_source.index: {backend: ..., backend-specific keys}`. Per the corrected semantics on the thread: `store`/`on_miss`/`write_back` are backend-specific parameters, and a key the named backend does not accept is a config **error**, not ignored. Wired into `validate_config`; `index_from_config` re-validates so dict-built Lambda payloads that skip `validate_config` still fail loudly.
- **Worker seam** (`src/zagg/processing/worker.py`) — `process_shard` resolves the backend once up front, routes group reads through `backend.read_group(...)`, and calls `backend.finish_granule(h5obj, url)` after each granule's groups (wrapped: a write-back failure can never fail the read).
- **`inline` backend** (`src/zagg/index/inline.py`) — per-dataset chunk maps built at read time by the pure-Python h5coro v1 B-tree walk (port of PR #159's `bench/offsets/extract_offsets.py` route (b), which was cross-validated there against h5py and hidefix on 61 granules / 184,616 chunks with zero mismatches), plugged into `_planned_read_group`/`_execute_plan_group` via the `read_fn` addressing seam. Planned ranges are issued as-is (h5coro already inflates exactly the covering chunks); the chunk map detects a range starting exactly on an interior chunk boundary — which h5coro 1.0.4's B-tree start-edge intersection **drops entirely** (the PR #152 off-by-one; empirically pinned against the test fixture) — and shifts one element early with a trim, so output is row-identical to `hierarchical` and survives shards the plain read fails on.
- **`inline` write-back (opt-in, issue Q2)** — `write_back: true` + `store`: `finish_granule` persists the granule-keyed manifest `store/granule_id.parquet` (granule id = URL stem, so it carries product+version and reprocessing changes the key). Schema = PR #159's offsets columns (`dataset, chunk_idx, elem_start, elem_end, byte_offset, nbytes, filter_mask`) + the `Index.from_chunks` decode-metadata columns (`chunk_offset`/`shape`/`chunk_shape` as JSON lists; `dtype` as the **byte-order-explicit** `np.dtype(...).str` form; `gzip`/`shuffle` as **booleans** — both pinned by test per the contract decisions relayed on this thread). Coverage per visited group is deterministic (chunk maps pre-built metadata-only for every dataset the config can touch), so concurrent shards of one granule write identical manifests. Writes route through `zagg.store.open_object_store` — local dir or `s3://` with ambient (execution-role) credentials, never the granule-read credentials; a failed write is logged and the read continues.
- **IAM grant** (`deployment/aws/template.yaml` + `execution_role.yaml`) — the shared execution role gains `s3:GetObject` + `s3:PutObject` scoped to exactly `arn:aws:s3:::sliderule-public-cors/zagg-index/*` (both copies of the role, pinned by `test_execution_role_grants_zagg_index_store`). Landed per espg's approval on this thread's relay; **no stack update is run from here** — the change deploys with the pending `Timeout=900` update (espg's, single deploy).
- **Docs** — `docs/virtual_index.md` (config, backend table, manifest schema, external-backend entry-point recipe) wired into the mkdocs nav.
- **Merged onto post-#152 `main`** (`merge main: reconcile index seam onto #152 a-priori read path`, no force-push): main's `_execute_plan_group` refactor already carries a `read_fn` parameter, so the index seam now *forwards* through it instead of duplicating the split; the worker routes the a-priori `granule_url` kwarg through `backend.read_group` (presence-gated exactly as main does, so monkeypatched `_read_group` fakes keep their signature). The a-priori arm (2a) and index backends coexist: `chunk_boundaries` routes inside `_read_group` through the default backend, and `inline` + `chunk_boundaries` together is rejected as a config error (the arm would silently bypass the backend's addressing).

## Phases

- [x] **Phase 1 — protocol + registry + config surface + `hierarchical` backend + worker seam.** Pure refactor of today's path behind the protocol; default path byte-identical (asserted in tests).
- [x] **Phase 2 — `inline` backend** + the committed HDF5 test fixture (`tests/data/index/`, ~50 KB; generation script alongside, needs h5py offline only — **espg approved keeping the binary**) + fixture-gated equality tests.
- [x] **Phase 3 — `inline` write-back** (opt-in): granule-keyed parquet manifest to the store via `finish_granule`.
- [x] **Phase 4 — docs.**
- [x] **Post-review folds** (see the review comment): shared planned-config gate, memoized entry-point discovery, boundary-only shift, deterministic write-back coverage, BytesIO manifest write, ValueError at the config-validation surface; then the from_chunks contract pins (byte-order-explicit dtypes, boolean gzip).
- [x] **IAM grant for the zagg-index store** (espg-approved; deploys bundled with the pending Timeout=900 stack update).

Not in this PR: the `sidecar` backend itself ships in `h5coro-hidefix` (external, discovered via the entry-point group); the 88S benchmark matrix is #148 phase 5; full-file (~all-datasets) manifest coverage rides PR #152's `mode="extract"` fan-out per the coverage decision — with the noted follow-up that the fan-out should assert every data-dictionary dataset is indexed or skip-listed per granule.

## How tested

`pytest -v` on the merged branch: **1338 passed, 26 skipped** (full suite; includes all of #152's a-priori/streaming/extract tests plus this PR's 51 in `tests/test_index.py` and the role-grant pin in `tests/test_lambda_build.py`):

- **Registry/protocol**: builtin resolution, entry-point discovery + memoization (one scan per interpreter, failed lookups retried), builtin-shadowing rejection, broken-plugin isolation.
- **Config**: irrelevant-keys-are-errors, required keys, `validate_config` wiring, `inline` cross-field rules (`write_back` bool, `store` iff `write_back`, `read_plan.spatial_index` required, `chunk_boundaries` mutually exclusive).
- **Byte-equality gates**: `hierarchical` backend vs direct `_read_group`; explicit `hierarchical` block vs no block through `process_shard`; `inline` vs `hierarchical` row/byte-identical per group (pandas and arrow carriers) on the committed fixture through both `read_group` and end-to-end `process_shard` — the PR #150 conformance philosophy at unit scale.
- **Chunk map**: counts/bounds vs the known fixture layout (10×256-photon chunks incl. partial last, 2-D trailing-dim collapse, contiguous pseudo-chunk, missing-path `KeyError`, boundary detection).
- **The h5coro off-by-one**: a shard whose plan starts exactly on photon 256 makes the *plain* hyperslice read fail (regression-pinned), while `inline`'s boundary-shifted read succeeds and matches numpy ground truth.
- **Write-back**: local-store round trip, deterministic coverage incl. empty groups, byte-order-explicit dtype pin (`<f8`/`|i1`, regex over every row), opt-in default writes nothing, store failure logged and the shard still returns.
- **IAM**: `test_execution_role_grants_zagg_index_store` pins the statement (exact prefix ARN, exactly Get+Put) in both `template.yaml` and `execution_role.yaml`.

Lint: `ruff check src tests` clean except the **pre-existing** `N818` in `src/zagg/registry.py` (on `main`, out of scope). `ruff format --check`: all touched files clean; 7 untouched files have pre-existing drift on `main` — left alone per the don't-fix-unrelated rule.

## Questions for review

All four earlier questions are resolved on the thread: **(1)** fixture binary kept (espg), **(2)** manifest schema pinned — boolean `gzip` (binding maps presence in h5coro-hidefix 0.1.1) + byte-order-explicit dtype strings, **(3)** coverage stays lazy/per-config with full-file coverage on the extraction fan-out, **(4)** IAM grant landed here per espg's approval. One residual note rather than a question: the grant was added to **both** role definitions (`template.yaml` inline role and `execution_role.yaml` for IAM-constrained accounts) to keep them in lockstep — flag if the admin-created-role variant should instead wait.


espg:
🤖 *from Claude (review)*

Adversarial self-review of the four phases (8 finder angles over `origin/main...HEAD`). Findings, each with disposition — folds landed in 1e56890:

**Folded**

1. `src/zagg/index/inline.py` (read seam) — the chunk-cover read started one element early *unconditionally* (`lo = cs - 1 if cs > 0`), inflating one extra stored chunk per planned run vs `hierarchical`, and the alignment itself bought nothing (h5coro already inflates exactly the covering chunks for any `[s, e)`). → Reads now issue the planned range as-is; the chunk map is used only to *detect* a start that lands exactly on an interior chunk boundary (the h5coro drop case) and shift one element early just there. `aligned_cover` removed; `starts_on_boundary` added.
2. `src/zagg/index/inline.py` / `src/zagg/processing/read.py` — the planned-route completeness gate (`levels`/`base_level`) was duplicated verbatim between `_read_group` and `InlineIndex.read_group`. → Extracted `_validate_planned_config`, used by both.
3. `src/zagg/index/__init__.py:available_index_backends` — `importlib.metadata.entry_points()` re-scanned on every call (twice per `index_from_config`), unlike `zagg.registry`'s memoized discovery. → Scan memoized per interpreter; failed lookups are not cached (retry semantics match `registry.py`).
4. `src/zagg/index/inline.py:write_manifest` — parquet round-tripped through a `TemporaryDirectory` (disk write + read-back per granule). → Serialized to `io.BytesIO` (verified fastparquet supports it).
5. **Write-back coverage was nondeterministic** (the strongest finding): the `full_read` selectivity fallback returns via `_read_group_full` *before* the `read_fn` seam, and an empty-shard group returns `None` before any base-rate read — either way those datasets were missing from the manifest, so concurrent shards of one granule could overwrite `G.parquet` with different partial subsets. → Write-back mode now pre-builds the chunk maps for every dataset the config can touch (base coords/variables/filter datasets + the spatial-index level's coord/link arrays) at the top of each group's `read_group`, metadata-only. Coverage per visited group is deterministic, so last-writer-wins between same-config shards is idempotent. Covered by `test_coverage_deterministic_for_empty_groups`.
6. `src/zagg/index/__init__.py:validate_index_config` — an unknown backend name escaped `validate_config` as `UnknownCapability` (a `KeyError`), inconsistent with the `ValueError` contract of every other config-validation failure. → Converted to `ValueError` at the validation surface; `get_index_backend` (runtime resolution) still raises `UnknownCapability`, matching `zagg.registry`.

**Declined, with reasons (standing for review)**

7. "Reuse `zagg.registry.Registry[T]` instead of a parallel discovery loop" — deliberate per the issue's Q4 decision ([permalink](https://github.com/englacial/zagg/issues/160#issuecomment-4880966830)): a dedicated `zagg.index_backends` group whose entry points resolve *directly to backend classes* (the `xarray.backends` precedent), with builtins in a static dict so core needs no entry-point machinery for its own paths. `Registry`'s `zagg.plugins` contract (zero-arg `register()` callables, shared `_DISCOVERED` sweep) is a different registration model; `UnknownCapability` is reused where it fits.
8. "Double validation: `validate_config` + `index_from_config`" — deliberate defense for dict-built Lambda payloads that never pass through `validate_config`; after fold (3) the repeat costs dict lookups once per shard.
9. "The h5coro start-edge workaround should live in `execute_read_plan` so `hierarchical` is covered too" — the default path *cannot* detect chunk boundaries without a chunk map (that is the point of the backend), and the exposure is pre-existing on `main` (found in PR #152, discussed on issue #148; the failure mode is a loud dropped read, not silent corruption — confirmed empirically against the fixture). The right global fix is upstream in h5coro or adoption of `inline`; changing the production default's read behavior is out of this PR's byte-identical mandate.
10. "Patching `zagg.processing._read_group` doesn't intercept `inline` reads" — by design: the monkeypatch contract is preserved for the default path (and asserted in tests); a config that opts into `inline` opts into real chunk-map reads.
11. "A group whose *data* read raises mid-granule still gets manifest coverage" — the manifest describes the file's metadata (offsets from the B-tree), which is valid regardless of whether this run's photon read succeeded; after fold (5) coverage is intentionally independent of read success.

Full suite after folds: 1146 passed, 24 skipped; `ruff check` clean apart from the pre-existing `N818` noted in the PR body.

---

espg:
🤖 *from Claude*

Status: all four phases are complete (protocol/registry/config/`hierarchical` seam → `inline` backend → opt-in write-back → docs), the adversarial-review findings are folded (see the review comment above; folds in 1e56890 + 002356d), and the suite is green locally (1146 passed, 24 skipped; ruff clean apart from the pre-existing `N818` on `main`). Leaving the PR in draft with `waiting`: the four **Questions for review** in the PR body need calls before this should merge. Condensed, each answerable in one pass:

1. **Fixture binary** — keep the committed ~50 KB `tests/data/index/atl03_mini.h5` (+ its offline h5py generation script), **(a)**, or add h5py to the `test` extra and generate in CI, **(b)**? Default as shipped: (a).
2. **Manifest schema** — the write-back parquet emits PR #159's offsets columns + `chunk_offset`/`dtype`/`shape`/`chunk_shape`/`gzip`/`shuffle` (JSON-encoded tuples, numpy dtype strings, boolean filter flags; deflate *level* isn't exposed by h5coro's metadata parse). Does this match the `Index.from_chunks` contract in the binding, **(a)** yes, or **(b)** name the deltas and I'll match them?
3. **Manifest coverage** — per-config lazy coverage now (deterministic per visited group), full ~200-dataset file coverage as a follow-up, **(a)**, or grow the full-file walk in this PR, **(b)**? Recommend (a): the dedicated extraction fan-out is PR #152's `mode="extract"` surface anyway.
4. **IAM grant** — extraction-role `s3:PutObject` on `arn:aws:s3:::sliderule-public-cors/zagg-index/*`: ride **(a)** this PR or **(b)** PR #152 (which owns the `mode="extract"` write side)? Recommend (b); `deployment/aws/` is untouched here.

---

espg:
🤖 *from Claude*

Answers to the three PR-body questions, per espg (decisions, not suggestions — act on them):

1. **Manifest schema — (b), two deltas, and only one lands on you.** Checked against the shipped `Index.from_chunks` contract (espg/h5coro-hidefix#1 / PR #2): everything matches except (i) `gzip` — you emit a boolean (h5coro's metadata parse can't see the deflate level) while `from_chunks` currently rejects bool. **Resolution is binding-side**: h5coro-hidefix 0.1.1 will accept a bool and map it to filter presence (deflate level is irrelevant for decode), so **keep emitting the boolean** — do not fabricate a level int. (ii) `dtype` strings must carry explicit byte order (`<f8`, `|i1`, not bare `f8`) — if you emit `np.dtype.str` you already do; confirm and pin it with a test either way. JSON-tuple encodings for shape/chunk_shape/chunk_offset are fine (conversion is caller-side by contract).
2. **Coverage — (a).** Write-back stays lazy/per-config (deterministic per visited group); full-file coverage belongs to PR #152's `mode="extract"` fan-out. One follow-up to note for the extraction side (not this PR): ATL03's dataset inventory is fixed by the product spec, so the fan-out should assert per granule that every data-dictionary dataset is either indexed or skip-listed with a reason (real-granule reality check: ~1,055 datasets after beam-group expansion, of which a handful are legitimately skipped — custom dtypes / 0-d).
3. **IAM — the (b) recommendation is stale: PR #152 merged *without* the grant.** The execution role's `s3:PutObject` is scoped to `${OutputBucketName}/*` only, which does not cover `sliderule-public-cors/zagg-index/*`. **Add the explicit statement in this PR** (`s3:PutObject` + `s3:GetObject` on `arn:aws:s3:::sliderule-public-cors/zagg-index/*` for the shared execution role) — in-scope, #160 names the grant. Note in the PR body that the stack update bundles with the pending `Timeout=900` one (single deploy, espg's).

espg:
> Fixture binary

keep the committed ~50 KB tests/data/index/atl03_mini.h5  -- this is fine

also, resolve branch conflicts

espg:
🤖 *from Claude*

Acted on the new thread input; branch is `87b941e`, four commits since the last status:

1. **Conflicts resolved** — `merge main: reconcile index seam onto #152 a-priori read path` (`ca94cb7`, a merge, no force-push). Main's `_execute_plan_group` refactor already carries a `read_fn` parameter, so the index seam now forwards through it rather than duplicating the split; `_planned_read_group` gained the forwarding kwarg, and the shared `_validate_planned_config` gate was re-applied around main's `chunk_boundaries` dispatch. The worker routes the a-priori `granule_url` kwarg through `backend.read_group` exactly as main presence-gates it, so monkeypatched `_read_group` fakes keep their signature. Coexistence is explicit: arm (2a) runs through the default backend, and `inline` + `chunk_boundaries` together is rejected at config validation (`test_inline_rejects_apriori_chunk_boundaries`) — the arm would otherwise silently bypass the backend's addressing.
2. **Schema contract pinned** — `fold review: manifest dtype byte-order pin (issue #160)` (`77d7e98`): `gzip` stays boolean per the binding decision; dtype strings were already `np.dtype(...).str`, now pinned by test (`<f8`/`|i1` exact values + a byte-order regex over every manifest row) and documented in the schema comment + docs table.
3. **Fixture** — kept as committed per espg's call; nothing to change.
4. **IAM grant landed** — `grant zagg-index write-back S3 access (issue #160)` (`87b941e`), per espg's approval relayed for this session: `s3:GetObject` + `s3:PutObject` on exactly `arn:aws:s3:::sliderule-public-cors/zagg-index/*`, added to the shared execution role in `template.yaml` **and** its admin-created mirror `execution_role.yaml` (kept in lockstep; flag if the mirror should wait), pinned by `test_execution_role_grants_zagg_index_store`. No stack update is run from here — it deploys bundled with the pending `Timeout=900` update.

Gates on the merged branch: full suite **1338 passed, 26 skipped**; `ruff check`/`format --check` clean on all touched files (pre-existing `N818` + format drift on `main` untouched, per the PR body). Staying in draft with `waiting` for review.

---

espg:
🤖 *from Claude*

Two deployment-readiness notes from espg's #149 direction (verifying `h5coro-hidefix` is ready for the architecture-test deployment):

1. **The wheel is proven on the deployment target.** Inside `public.ecr.aws/lambda/python:3.12` (arm64), PyPI `h5coro-hidefix==0.1.1` installs cleanly and the full sidecar worker flow runs byte-identical against h5py: `from_chunks` with **boolean `gzip`** (the schema answer above is live in 0.1.1), then `read_plan → obstore.get_ranges → read_from_buffers`, with obstore at this repo's own pin (`>=0.8.2`). The consumption chain this PR designs is validated end-to-end at the binding boundary.
2. **Two gaps to close for the deployment test, one on each side.** (a) *Binding side (in progress)*: the `sidecar = "h5coro_hidefix.zagg_backend:SidecarIndex"` entry point this PR's docs specify doesn't exist in 0.1.1 — the provider module + declaration are being built now against this PR's protocol (pinned to its current head; if the protocol shifts under review, flag it and the provider tracks it), landing as 0.2.0. (b) *This PR*: nothing in the diff adds the wheel to the Lambda layer — the diff has no `pyproject.toml` change, and the layer build installs from the `lambda` extra pins. For the architecture test the layer must carry the wheel: add `h5coro-hidefix==0.2.0` (once tagged) to the `lambda` extra — or state explicitly in the PR body how the test deployment's layer acquires it if a different mechanism is intended (size impact is negligible: 2.2 MB abi3 wheel, numpy-only runtime deps).

Issue #89, opened 2026-06-23
Reuse shardmaps across configs with different aggregators (relax signature guard to spatial-only)


espg:
🤖 *from Claude*

## Problem

A ShardMap is a **spatial** artifact — `shard_keys` + granule→shard assignment — depending only on the spatial grid (type, indexing, `parent_order`) and the AOI/product/time query. It does **not** depend on what you aggregate. Yet you currently can't reuse one shardmap across two configs that share the spatial grid but declare different aggregation fields.

`runner._check_signature` (`runner.py:287`) does an **exact dict-equality** check between the run config's `grid.signature()` and the shardmap's stored `grid_signature`, and `HealpixGrid.signature()` (`grids/healpix.py`) — plus the rectilinear equivalent — include `output_fields = output_field_signature(config)`. So two configs with identical spatial grids but different agg fields produce different signatures and the runner raises:

```
ValueError: ShardMap was built for a different grid than this run config.
```

…even though the shardmap content is byte-identical for both.

Concretely: the new `atl03_tdigest_healpix.yaml` and `atl03_gain_bias_healpix.yaml` (PR #86) share parent_order 11 / chunk_inner 13 / child_order 19, so one NEON shardmap maps both — but today you must build (and CMR-fetch) two.

## Why `output_fields` is in the signature

`signature()` serves two distinct jobs:
1. **The shardmap-reuse guard** (`_check_signature`) — purely spatial; `output_fields` is irrelevant here.
2. **Co-aggregation nesting** (`nests_with`, #29) — where co-aggregated grids genuinely must share the output schema; `output_fields` belongs here.

Conflating them over-constrains (1).

## Proposed fix

- Split the spatial/structural fingerprint from the output-schema one: add a `spatial_signature()` (HEALPix: type, indexing, `parent_order`, `child_order`, `layout`; rectilinear: crs/resolution/bounds/chunk_shape) and have `_check_signature` compare **that**, while `nests_with` keeps using the full `signature()` (still incl. `output_fields`).
- ShardMap stores the spatial signature (`grid_signature` → spatial-only, or add a `spatial_signature` field). **Back-compat:** existing maps carry the full signature — the check should accept a spatial-subset match against an old full signature, so previously-built maps keep validating.
- Net: one shardmap serves any config sharing the spatial grid; #29's schema/nesting guarantees are untouched.

## Files
- `src/zagg/runner.py` (`_check_signature` ≈ 287).
[...]

PR #164, opened 2026-07-04
Typed morton_index Arrow surface over the arro3 carrier

### LINKED ISSUE #135: new from / to arrow surface for morton indices
https://github.com/espg/mortie/pull/86 recently merged in our upstream mortie dep ; what does this change / enable for us?

espg:
Closes #135

Implements the plan approved in https://github.com/englacial/zagg/issues/135#issuecomment-4881083363 with the amendments from https://github.com/englacial/zagg/issues/135#issuecomment-4881103631: carry mortie's `morton_index` Arrow extension type through the arro3-core worker carrier instead of the deliberate uint64 strip, with the on-disk Zarr output staying plain `uint64` byte-for-byte (the extension metadata lives at the interchange layer only; the #71 storage guarantee is unchanged).

## What / approach

mortie 0.8.4 (espg/mortie#94) ships a pyarrow-free PyCapsule C Data Interface surface: `MortonIndexArray.__arrow_c_array__` / `from_arrow` and `mortie.arrow.export_c_array` / `import_c_array`, with `ARROW:extension:name == "mortie.morton_index"` on the schema and null↔sentinel via a real validity bitmap. zagg adopts it at three boundaries:

1. **Worker carrier** (`src/zagg/processing/write.py`): `_build_output` routes the morton coordinate through `morton_to_arrow` (typed extension column) instead of the uint64 strip; `_iter_carrier_columns` detects the extension column via field metadata and extracts the packed words with `import_c_array` for the Zarr write. Store bytes proven identical to the pre-change strip and to the pandas carrier — including for a sentinel/null-bearing morton column (`tests/test_processing.py::TestTypedMortonCarrier`). The pandas carrier path is untouched.
2. **Boundary adapter** (`src/zagg/grids/morton.py`): `morton_to_arrow(values)`, `morton_from_arrow(col)`, `is_morton_arrow(col)`, plus a pinned `MORTON_EXTENSION_NAME` constant — all type logic stays in the one module the #71 migration established.
3. **Catalog side** (`src/zagg/catalog/shardmap.py`): `ShardMap.to_parquet` / `from_parquet` — the Arrow-native sibling of the JSON manifest, with `shard_keys` as a typed `morton_index` pyarrow extension column (mortie registers the type on import). pyarrow imports stay lazy inside the methods, so the worker path remains pyarrow-free. Degraded manifests (extension type stripped, plain uint64) still load; foreign/truncated files are rejected cleanly. Dispatch stays JSON ([confirmed](https://github.com/englacial/zagg/pull/164#issuecomment-4883419266)).

Per the amendment on question (3): `cell_ids` stays NESTED HEALPix `uint64` by default, and a new validated flag `output.grid.cell_ids_encoding: morton` emits the packed morton words as `cell_ids` instead (HEALPix-only; rejected on rectilinear; value validated both at config load and at grid construction). Default (key absent, explicit null, or `nested`) is byte-identical — asserted by tests. The store's `dggs.indexing_scheme` attr records the active encoding ([confirmed](https://github.com/englacial/zagg/pull/164#issuecomment-4883419266): morton will be upstreamed to the DGGS convention later); the legacy descriptive `output.grid.indexing_scheme` config key is now validated (`nested` only, error pointing at the real knob). Per the [land-b decision](https://github.com/englacial/zagg/pull/164#issuecomment-4883509497), mixed `cell_ids_encoding` grids are **explicitly incompatible for co-aggregation**: `nests_with` compares the encoding and `signature()` records it, while `spatial_signature()` (shard-map reuse, #89) deliberately stays encoding-free (option a1).

Docs: new [Typed morton boundary](../blob/claude/135-morton-arrow-surface/docs/morton_arrow.md) design page (wired into mkdocs nav), updated `grids/morton.py` module docstring (it previously documented the strip as the design), and carrier notes in `write.py`.

## Interop gate result (phase 1)

All four gates **pass** on arro3-core 0.8.1 / mortie 0.8.4 — no fallback needed, no mortie-side change needed (`tests/test_morton_arrow.py`):

- (a) `MortonIndexArray` → `arro3.core.Array.from_arrow` carries `ARROW:extension:name == "mortie.morton_index"`;
- (b) the metadata survives `Table.from_pydict` → `column(name)` → `combine_chunks()` (the exact carrier ops);
- (c) `mortie.arrow.import_c_array(col)` returns the words byte-equal, all-zero sentinel ↔ Arrow null intact, bit-63 southern words non-negative as `uint64`;
- (d) the whole path runs in a subprocess with pyarrow made unimportable (mortie registers its pyarrow extension type eagerly at import when pyarrow is present, so the check must block before import).

## Phases

- [x] Phase 1 — `mortie>=0.8.4` floor bump (`pyproject.toml` only; no other dep or pin changes) + interop gate tests (`tests/test_morton_arrow.py`)
- [x] Phase 2 — boundary adapter in `src/zagg/grids/morton.py` (`morton_to_arrow`, `morton_from_arrow`, `is_morton_arrow`) + round-trip unit tests (`tests/test_grids.py::TestMortonArrowAdapter`)
- [x] Phase 3 — carrier integration in `src/zagg/processing/write.py`; store bytes identical across typed-arrow / pandas / pre-change strip; pandas path untouched (`tests/test_processing.py::TestTypedMortonCarrier`)
- [x] Phase 4 — `output.grid.cell_ids_encoding` flag (default `nested`, byte-identical) + validation and tests (`tests/test_config.py::TestCellIdsEncoding`, `tests/test_grids.py::TestCellIdsEncoding`)
- [x] Phase 5 — catalog/shardmap: `ShardMap.to_parquet`/`from_parquet` with a typed morton `shard_keys` column (`tests/test_shardmap.py::TestParquetIO`)
- [x] Phase 6 — docs: `docs/morton_arrow.md`, `grids/morton.py` module docstring, carrier notes in `write.py`
- [x] Adversarial self-review folded (`c299500`): present-but-null `cell_ids_encoding` no longer bypasses the default; encoding value validated at grid construction; legacy `indexing_scheme` key validated; `from_parquet` clean rejection on missing `shard_keys`; hot-path dict-copy and loop-local import removed; sentinel-bearing store-byte test added
- [x] Phase 7 (`f5d7efa`, per the [land-b decision](https://github.com/englacial/zagg/pull/164#issuecomment-4883509497)) — `cell_ids_encoding` in the co-aggregation gate: `nests_with` rejects mixed encodings, `signature()` records the key; `spatial_signature()` unchanged (a1). Tests: `test_mixed_encodings_do_not_nest`, `test_signature_records_encoding`

## How tested

- `uv run pytest -v` green at every phase: 1097 passed / 24 skipped at baseline → **1136 passed / 24 skipped** with all phases + folds (39 new tests).
- `uv run ruff check` / `ruff format --check` clean on every file this PR touches; `pre-commit` mypy adds **no new errors** over the `main` baseline.
- Byte-identity is asserted directly: the typed-arrow store, the pandas-carrier store, and a store written through a reproduction of the pre-change uint64-strip carrier compare equal key-by-key, byte-for-byte — for both a normal and a sentinel/null-bearing morton column.
- Verified in-env on mortie 0.8.4: `import_c_array` returns correct contiguous `uint64` words for plain (extension-stripped) columns, multi-chunk ChunkedArrays, and null-bearing extension columns.
- `main` moved under the branch (issues #148/#12 landed); re-verified with `git merge-tree`: merges clean, no conflicts.

## Questions for review

All design questions are resolved ([espg's answers](https://github.com/englacial/zagg/pull/164#issuecomment-4883419266), [land-b + a1](https://github.com/englacial/zagg/pull/164#issuecomment-4883509497)): `dggs.indexing_scheme: "morton"` stays; shardmap dispatch stays JSON; `spatial_signature()` stays encoding-free (a1); the co-aggregation gate records the encoding (b, landed in phase 7).

Remaining notes for the reviewer:

- Pre-existing (untouched by this PR, flagged per §4): `ruff check src tests` fails on `main` with one `N818` (`src/zagg/registry.py:64` `UnknownCapability`); `ruff format --check` would reformat 6 files on `main`; codespell flags 2 words (`grids/rectilinear.py:270`, `grids/aoi.py:68`); mypy reports ~39 errors on `main`. All left alone here.

### DIFF
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -584,5 +584,64 @@ class ShardMap:
             d.get("aoi_mask"),
         )
 
+    # Schema-metadata key for the manifest's non-columnar payload (parquet form).
+    _PARQUET_META_KEY = b"zagg:shardmap_meta"
+
+    def to_parquet(self, path: str) -> None:
+        """Write the manifest as parquet with a TYPED morton ``shard_keys`` column.
+
+        The Arrow-native sibling of :meth:`to_json` (issue #135): ``shard_keys``
+        carries mortie's ``morton_index`` pyarrow extension type (registered by
+        mortie on import), so any Arrow-aware consumer sees morton words, not
+        anonymous ints. ``granules`` (and ``aoi_mask`` when present) ride as
+        per-shard JSON strings — the same self-contained payloads the JSON form
+        stores — and ``metadata``/``grid_signature`` live in the schema metadata,
+        mirroring the ``Catalog`` geoparquet convention (``sources.py``).
+
+        Requires pyarrow (the off-Lambda ``catalog`` extra); the worker path
+        never calls this — the runner dispatches from the JSON manifest.
+        """
+        import pyarrow as pa
+        import pyarrow.parquet as pq
+        from mortie.arrow import from_morton_index
+
+        words = np.asarray(self.shard_keys, dtype=np.uint64)
+        columns: dict = {
+            "shard_keys": from_morton_index(words),
+            "granules": pa.array([json.dumps(g) for g in self.granules]),
+        }
+        if self.aoi_mask is not None:
+            columns["aoi_mask"] = pa.array([json.dumps(m) for m in self.aoi_mask])
+        meta = json.dumps({"metadata": self.metadata, "grid_signature": self.grid_signature})
+        table = pa.table(columns).replace_schema_metadata({self._PARQUET_META_KEY: meta.encode()})
+        pq.write_table(table, path)
+
+    @classmethod
+    def from_parquet(cls, path: str) -> "ShardMap":
+        """Load a manifest from the parquet form written by :meth:`to_parquet`.
+
+        Importing :mod:`mortie.arrow` first registers the ``morton_index``
+        extension type, so the ``shard_keys`` column rehydrates typed; the words
+        are pulled over the C Data Interface (``import_c_array``) regardless.
+        """
+        import pyarrow.parquet as pq
+        from mortie.arrow import import_c_array
+
+        table = pq.read_table(path)
+        raw = (table.schema.metadata or {}).get(cls._PARQUET_META_KEY)
+        if raw is None or "granules" not in table.column_names:
+            raise ValueError(f"{path}: not a zagg ShardMap parquet manifest")
+        d = json.loads(raw)
+        if "grid_signature" not in d:
+            raise ValueError(f"{path}: missing required key 'grid_signature'")
+        shard_keys = [int(w) for w in import_c_array(table.column("shard_keys"))]
+        granules = [json.loads(g) for g in table.column("granules").to_pylist()]
+        aoi_mask = (
+            [json.loads(m) for m in table.column("aoi_mask").to_pylist()]
+            if "aoi_mask" in table.column_names
+            else None
+        )
+        return cls(d["grid_signature"], shard_keys, granules, d.get("metadata", {}), aoi_mask)
+
 
 __all__ = ["ShardMap"]
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -629,7 +629,7 @@ class ShardMap:
 
         table = pq.read_table(path)
         raw = (table.schema.metadata or {}).get(cls._PARQUET_META_KEY)
-        if raw is None or "granules" not in table.column_names:
+        if raw is None or not {"shard_keys", "granules"}.issubset(table.column_names):
             raise ValueError(f"{path}: not a zagg ShardMap parquet manifest")
         d = json.loads(raw)
         if "grid_signature" not in d:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -222,6 +222,17 @@ def validate_config(config: PipelineConfig) -> None:
                     "output.grid.cell_ids_encoding only applies to healpix grids "
                     f"(grid type is {grid['type']!r})"
                 )
+        # The legacy output.grid.indexing_scheme key is descriptive only (the
+        # shipped configs carry "nested"); it does NOT select the cell_ids
+        # encoding. Reject any other value so a user reaching for it lands on the
+        # real knob instead of a silently-NESTED store.
+        legacy_scheme = grid.get("indexing_scheme")
+        if legacy_scheme is not None and legacy_scheme != "nested":
+            raise ValueError(
+                f"output.grid.indexing_scheme is descriptive and must be 'nested' "
+                f"(got {legacy_scheme!r}); to emit morton words as cell_ids set "
+                f"output.grid.cell_ids_encoding: morton"
+            )
 
     # Validate the optional per-cell carrier (issue #132). Mirrors the worker's
     # ``{"pandas", "arrow"}`` guard (worker.py) so a typo in the aggregation YAML
@@ -1256,8 +1267,8 @@ def get_cell_ids_encoding(config: PipelineConfig) -> str:
     ``"nested"`` (default) stores the standardized NESTED HEALPix cell IDs.
     ``"morton"`` stores the packed morton words instead — the same ``uint64``
     words the ``morton`` coordinate carries — opening test/prototype flows that
-    index by morton directly. Default behavior (key absent or ``"nested"``) is
-    byte-identical to a pre-flag run.
+    index by morton directly. Default behavior (key absent, explicit ``null``,
+    or ``"nested"``) is byte-identical to a pre-flag run.
 
     Parameters
     ----------
@@ -1268,7 +1279,9 @@ def get_cell_ids_encoding(config: PipelineConfig) -> str:
     str
         ``"nested"`` (default) or ``"morton"``.
     """
-    return config.output.get("grid", {}).get("cell_ids_encoding", "nested")
+    # A present-but-null key (YAML ``cell_ids_encoding:``) must fall back to the
+    # default too — the same treatment ``from_config`` gives a null ``layout``.
+    return config.output.get("grid", {}).get("cell_ids_encoding") or "nested"
 
 
 def get_store_path(config: PipelineConfig) -> str | None:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -207,6 +207,21 @@ def validate_config(config: PipelineConfig) -> None:
         layout = grid.get("layout")
         if layout is not None and layout not in ("dense", "fullsphere"):
             raise ValueError(f"output.grid.layout must be 'dense' or 'fullsphere' (got {layout!r})")
+        # Optional cell_ids encoding (issue #135): "nested" (default, the DGGS
+        # standard) or "morton" (emit the packed morton words as cell_ids — a
+        # test/prototype capability). HEALPix-only: rectilinear grids have no
+        # cell_ids coordinate, so the knob would silently do nothing there.
+        encoding = grid.get("cell_ids_encoding")
+        if encoding is not None:
+            if encoding not in ("nested", "morton"):
+                raise ValueError(
+                    f"output.grid.cell_ids_encoding must be 'nested' or 'morton' (got {encoding!r})"
+                )
+            if grid["type"] != "healpix":
+                raise ValueError(
+                    "output.grid.cell_ids_encoding only applies to healpix grids "
+                    f"(grid type is {grid['type']!r})"
+                )
 
     # Validate the optional per-cell carrier (issue #132). Mirrors the worker's
     # ``{"pandas", "arrow"}`` guard (worker.py) so a typo in the aggregation YAML
@@ -1233,6 +1248,27 @@ def get_shard_order(config: PipelineConfig) -> int | None:
     """
     val = config.output.get("grid", {}).get("shard_order")
     return None if val is None else int(val)
+
+
+def get_cell_ids_encoding(config: PipelineConfig) -> str:
+    """Return the HEALPix ``cell_ids`` coordinate encoding (issue #135).
+
+    ``"nested"`` (default) stores the standardized NESTED HEALPix cell IDs.
+    ``"morton"`` stores the packed morton words instead — the same ``uint64``
+    words the ``morton`` coordinate carries — opening test/prototype flows that
+    index by morton directly. Default behavior (key absent or ``"nested"``) is
+    byte-identical to a pre-flag run.
+
+    Parameters
+    ----------
+    config : PipelineConfig
+
+    Returns
+    -------
+    str
+        ``"nested"`` (default) or ``"morton"``.
+    """
+    return config.output.get("grid", {}).get("cell_ids_encoding", "nested")
 
 
 def get_store_path(config: PipelineConfig) -> str | None:
--- a/src/zagg/grids/healpix.py
+++ b/src/zagg/grids/healpix.py
@@ -172,7 +172,16 @@ class HealpixGrid:
         # cell_ids coordinate encoding (issue #135): "nested" (default, the DGGS
         # standard) or "morton" (emit the packed morton words as cell_ids — a
         # test/prototype capability). Default is byte-identical to a pre-flag run.
+        # Re-validated here (not only in validate_config) because both coords_of
+        # (the cell_ids values) and _dggs_attrs (the recorded indexing_scheme)
+        # interpret this string: an unvalidated third value would write NESTED
+        # values while recording a different scheme — a mis-decode for consumers.
         self.cell_ids_encoding = get_cell_ids_encoding(self.config)
+        if self.cell_ids_encoding not in ("nested", "morton"):
+            raise ValueError(
+                f"Unknown cell_ids_encoding: {self.cell_ids_encoding!r} "
+                "(expected 'nested' or 'morton')"
+            )
         self._position_map: dict[int, int] | None = None
         if populated_shards is not None:
             self.set_populated_shards(populated_shards)
--- a/src/zagg/grids/healpix.py
+++ b/src/zagg/grids/healpix.py
@@ -524,12 +524,15 @@ class HealpixGrid:
         """Canonical fingerprint of the grid's defining parameters.
 
         The full fingerprint: the spatial layout (:meth:`spatial_signature`)
-        plus the Option-B output-field set. ``nests_with`` (#29) keys on the
-        latter; the shard-map reuse guard keys on the former (#89).
+        plus the Option-B output-field set and the ``cell_ids_encoding``
+        (issue #135 — mixed encodings must never co-aggregate; ``nests_with``
+        keys on both). The shard-map reuse guard keys on the spatial part
+        only (#89).
         """
         return {
             **self.spatial_signature(),
             "output_fields": output_field_signature(self.config),
+            "cell_ids_encoding": self.cell_ids_encoding,
         }
 
     def nests_with(self, other) -> bool:
@@ -538,9 +541,14 @@ class HealpixGrid:
         Any two HEALPix grids nest (the nested hierarchy subdivides 4-for-1 at
         every order), provided they declare the same Option-B output-field set
         (issue #29) — co-aggregated grids must produce the same scalar/vector
-        schema. Cross-family (e.g. rectilinear) never nests.
+        schema — and the same ``cell_ids_encoding`` (issue #135): NESTED ids
+        and morton words are different id spaces, so a consumer joining
+        co-aggregated products on ``cell_ids`` would silently mismatch.
+        Cross-family (e.g. rectilinear) never nests.
         """
         if not isinstance(other, HealpixGrid):
+            return False
+        if self.cell_ids_encoding != other.cell_ids_encoding:
             return False
         return output_field_signature(self.config) == output_field_signature(other.config)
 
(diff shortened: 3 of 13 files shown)


src/zagg/config.py
```diff
@@ -1235,6 +1250,27 @@ def get_shard_order(config: PipelineConfig) -> int | None:
...
+    Parameters
+    ----------
+    config : PipelineConfig
+
+    Returns
+    -------
+    str
+        ``"nested"`` (default) or ``"morton"``.
+    """
+    return config.output.get("grid", {}).get("cell_ids_encoding", "nested")
```

**Present-but-null key bypasses the default and regresses the store attrs.** A YAML `cell_ids_encoding:` (explicit null) makes this `.get(..., "nested")` return `None` — the key exists, so the default never fires — and `validate_config`'s `if encoding is not None` guard skips it too. `HealpixGrid._dggs_attrs` then writes `"indexing_scheme": None` where pre-flag code hard-coded `"nested"`, breaking the byte-identical-default guarantee, and the declared `-> str` return type is violated. Treat `None` as `"nested"` here (the same pattern `from_config` uses for `layout`: `explicit_layout or "fullsphere"`).

---

src/zagg/grids/healpix.py
```diff
@@ -168,6 +169,10 @@ def __init__(
         self.shard_objects_per_shard = 4 ** (shard_obj_order - parent_order)
         self.layout = layout
         self.config = config or default_config("atl06")
+        # cell_ids coordinate encoding (issue #135): "nested" (default, the DGGS
+        # standard) or "morton" (emit the packed morton words as cell_ids — a
+        # test/prototype capability). Default is byte-identical to a pre-flag run.
+        self.cell_ids_encoding = get_cell_ids_encoding(self.config)
```
espg, line 179: 🤖 *from Claude (review)*

**No single owner for the encoding value.** `coords_of` interprets the flag with a binary `== "morton"` (anything else falls to nested cell_ids) while `_dggs_attrs` passes the raw string through as `indexing_scheme`. A `HealpixGrid` built directly from an unvalidated config (the construction path the new tests themselves use) with a third value — say `"ring"` — would write NESTED cell_ids but record `indexing_scheme: "ring"`: the recorded scheme no longer matches the values on disk, and a consumer decoding by the attr mis-places every cell. Validate the value here in `__init__` (mirroring the existing `layout` check) so both consumers can only ever see the two supported encodings.

---

src/zagg/config.py
```diff
@@ -207,6 +207,21 @@ def validate_config(config: PipelineConfig) -> None:
         layout = grid.get("layout")
         if layout is not None and layout not in ("dense", "fullsphere"):
             raise ValueError(f"output.grid.layout must be 'dense' or 'fullsphere' (got {layout!r})")
+        # Optional cell_ids encoding (issue #135): "nested" (default, the DGGS
+        # standard) or "morton" (emit the packed morton words as cell_ids — a
+        # test/prototype capability). HEALPix-only: rectilinear grids have no
+        # cell_ids coordinate, so the knob would silently do nothing there.
+        encoding = grid.get("cell_ids_encoding")
```
espg, line 214: 🤖 *from Claude (review)*

**The legacy `output.grid.indexing_scheme` key is now a decoy.** Every shipped healpix YAML carries `indexing_scheme: nested` (e.g. `configs/atl06.yaml:77`), but no code reads it — and this PR makes the store's `dggs.indexing_scheme` attr derive solely from the *new* `cell_ids_encoding` key. A user who edits the existing key to `indexing_scheme: morton` (the natural read — it literally names the store attr) gets no error and a NESTED store. At minimum, validate the legacy key when present (`nested` only, with a message pointing at `cell_ids_encoding`) so the trap fails loudly.

---

src/zagg/catalog/shardmap.py
```diff
@@ -584,5 +584,64 @@ def from_json(cls, path: str) -> "ShardMap":
...
+        from mortie.arrow import import_c_array
+
+        table = pq.read_table(path)
+        raw = (table.schema.metadata or {}).get(cls._PARQUET_META_KEY)
+        if raw is None or "granules" not in table.column_names:
+            raise ValueError(f"{path}: not a zagg ShardMap parquet manifest")
+        d = json.loads(raw)
+        if "grid_signature" not in d:
+            raise ValueError(f"{path}: missing required key 'grid_signature'")
+        shard_keys = [int(w) for w in import_c_array(table.column("shard_keys"))]
```
espg, line 637: 🤖 *from Claude (review)*

**Missing `shard_keys` column escapes the clean-rejection guard.** The guard on line 632 checks the metadata key and `granules` but not `shard_keys`, so a truncated/foreign file that happens to carry both raises a bare pyarrow `KeyError('shard_keys')` here instead of the deliberate `ValueError(... not a zagg ShardMap parquet manifest)`. Add `"shard_keys" not in table.column_names` to the guard.

(Verified clean, for the record: `mortie.arrow.import_c_array` on a **plain uint64** column — extension type lost — and on a **multi-chunk** ChunkedArray both return the correct words on mortie 0.8.4, so the "regardless" claim in the docstring holds and no `combine_chunks()` is needed.)

---

src/zagg/grids/morton.py
```diff
@@ -55,4 +67,56 @@ def to_morton_array(words):
...
+    """True if ``col`` is an Arrow array/column carrying the morton extension type.
+
+    Reads the ``ARROW:extension:name`` field metadata (present on both an
+    ``arro3.core.Array`` and a table column's ``ChunkedArray``); anything
+    without field metadata is not a typed morton column.
+    """
+    field = getattr(col, "field", None)
+    if field is None:
+        return False
+    return dict(field.metadata_str).get(_EXTENSION_NAME_KEY) == MORTON_EXTENSION_NAME
```

Nit (hot path): `field.metadata_str` is already a plain `dict` on arro3, so the `dict(...)` here is a throwaway copy made once per column per shard write. `field.metadata_str.get(_EXTENSION_NAME_KEY)` does the same job without the copy.

---

src/zagg/processing/write.py
```diff
@@ -715,6 +730,16 @@ def _iter_carrier_columns(carrier):
     n_rows = carrier.num_rows
     for name in carrier.column_names:
         col = carrier.column(name).combine_chunks()
+        # A typed morton column (mortie's ``morton_index`` extension type; issue
+        # #135) is the arro3 mirror of the pandas MortonIndexArray branch above:
+        # pull the packed uint64 words over the C Data Interface for the on-disk
+        # write (Arrow nulls -> the all-zero sentinel), keeping the stored dtype
+        # plain uint64 (#71).
+        if is_morton_arrow(col):
+            from mortie.arrow import import_c_array
```

Nit: the `from mortie.arrow import import_c_array` sits inside the per-column loop. Hoist it next to `from arro3.core import list_flatten` at the top of the arro3 branch — matches the surrounding pattern and keeps import machinery out of the loop body.

---

tests/test_processing.py
```diff
@@ -2502,6 +2502,105 @@ class _OneChunkGrid:
...
+        assert (words.view(np.int64) < 0).any()
+
+    @staticmethod
+    def _store_bytes(carrier, grid, parent):
+        store = MemoryStore()
+        grid.emit_template(store)
+        write_dataframe_to_zarr(carrier, store, grid=grid, chunk_idx=grid.block_index(parent))
+        return {k: v.to_bytes() for k, v in store._store_dict.items()}
+
+    def test_store_bytes_identical_across_carriers(self):
```
espg, line 2564: 🤖 *from Claude (review)*

**Byte-identity is untested for a null-bearing morton column.** This test builds its carrier from `grid.children(parent)` — all valid words, no all-zero sentinel — so the sentinel→Arrow-null→sentinel leg of the new typed path is never exercised at the *store-byte* level (only at the array level in `test_sentinel_null_roundtrip`). The pre-#135 strip wrote a literal `0` with no validity bitmap; the typed path routes it through an Arrow null and back via `import_c_array`. I verified on mortie 0.8.4 that `import_c_array` returns a contiguous `uint64` with `0` at null positions, so the bytes should match — but a store-level test with a sentinel-bearing morton column would pin the guarantee where it matters.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial self-review of all six phases (fresh context, five independent finder angles + verification against mortie 0.8.4 / arro3-core 0.8.1 in-env). Seven inline findings posted: two correctness (present-but-null `cell_ids_encoding` regressing `dggs.indexing_scheme` to `None`; no single owner for the encoding value between `coords_of` and `_dggs_attrs`), two robustness (legacy `output.grid.indexing_scheme` key now a silent decoy; `from_parquet` bare `KeyError` on a missing `shard_keys` column), two hot-path nits (dict copy in `is_morton_arrow`; loop-local import in `_iter_carrier_columns`), one test gap (store-byte identity unverified for a sentinel/null-bearing morton column).

Verified clean, for the record:

- `import_c_array` on a plain (extension-stripped) uint64 column, on a multi-chunk ChunkedArray, and on a null-bearing extension column all return the correct contiguous `uint64` words on mortie 0.8.4 — the `from_parquet` "regardless" claim holds, no `combine_chunks()` needed.
- The non-morton coord path in `_build_output` (`cell_ids`, `aoi_mask`) is byte-for-byte the old coercion (`Array.from_numpy(np.ascontiguousarray(np.asarray(vals)))`).
- Rectilinear carriers never hit the typed branch (`coords_of` returns no `morton` column).
- `validate_config` ordering is safe: `grid["type"]` in the new check runs only after the existing `"type"`-required check.
- `spatial_signature()` staying `"nested"` under `cell_ids_encoding: morton` is intentional (shard-map reuse is spatial; tested) — but note the store attr and manifest signature can now disagree for the same run; worth one docs sentence.
- Redundant morton decode on the sharded+companion path (`_write_companion_columns` decodes then discards) mirrors the pre-existing double `to_numpy()` pattern — noted, not folded.
- `MORTON_EXTENSION_NAME` mirroring mortie's constant is pinned by a test; `to_parquet`'s JSON-string granule columns match the `Catalog` geoparquet convention; no new deps, no `noqa`/`type: ignore`, no infra files touched.

---

espg:
🤖 *from Claude*

Folded all seven self-review findings in `c299500b74522ff1f6f2c45214918e13908ff43e`:

- `get_cell_ids_encoding` now treats a present-but-null key as `"nested"` (no more `indexing_scheme: None` in the store attrs; return type honest) — test `test_present_but_null_key_falls_back_to_nested`.
- `HealpixGrid.__init__` validates the encoding value, closing the unvalidated-config hole where `coords_of` and `_dggs_attrs` could disagree on a third value — test `test_unknown_encoding_rejected_at_grid_construction`.
- The legacy `output.grid.indexing_scheme` key is now validated (`nested` only, error pointing at `cell_ids_encoding`) so it can't act as a silent decoy — test `test_legacy_indexing_scheme_key_is_descriptive_only`; shipped configs unaffected.
- `ShardMap.from_parquet` rejects a missing `shard_keys` column with the clean manifest error instead of a bare `KeyError` — test `test_missing_shard_keys_column_rejected_cleanly`; plus a regression test pinning that an extension-stripped (plain uint64) manifest still loads (`test_extension_stripped_column_still_loads`, behavior verified on mortie 0.8.4).
- `is_morton_arrow` drops the per-column `dict()` copy; the `import_c_array` import is hoisted out of the per-column loop in `_iter_carrier_columns`.
- New store-byte test `test_store_bytes_identical_with_sentinel_nulls`: a sentinel/null-bearing morton column writes bytes identical across the typed-arrow, pandas, and pre-#135-strip carriers.
- One docs sentence recording the intentional `spatial_signature()`-stays-`nested` behavior under `cell_ids_encoding: morton`.

Full suite green locally after the fold: 1134 passed, 24 skipped; ruff/format clean on touched files; no new mypy errors over the `main` baseline.

---

espg:
> dggs.indexing_scheme under cell_ids_encoding: morton: the store attr now records "morton" instead of "nested" so the metadata never misdescribes the coordinate — but "morton" is outside the DGGS convention's standard schemes.

this is great -- exactly what we want, we'll upstream morton to dggs at some point in the future.

> Relatedly, spatial_signature() (recorded in ShardMap manifests) deliberately stays "nested" under the flag so shard maps remain reusable (https://github.com/englacial/zagg/issues/89, tested) — the store attr and the manifest signature can therefore disagree for a morton-encoded run; documented in docs/morton_arrow.md.

this is less clear to me...

> Phase 5 scoping: ShardMap had no parquet form (JSON only), so the typed-morton parquet surface landed as new to_parquet/from_parquet methods mirroring to_json/from_json (granules/aoi as per-shard JSON strings, provenance in schema metadata — the Catalog geoparquet convention from sources.py). The runner still dispatches from JSON; nothing on the worker path changed. Flag if you wanted a different shape (e.g. runner-consumable parquet manifests).

yes, correct-- shardmap stays json

> grid.signature() (the full fingerprint) does not include cell_ids_encoding. If you want the full signature to record the encoding for co-aggregation checks, say the word.

can you explain what this means? what would "the full signature to record the encoding for co-aggregation checks" do / mean? I'm not understanding the scope or behavior changes here

espg:
🤖 *from Claude*

Acknowledged on the two confirmed items: `dggs.indexing_scheme: "morton"` stays as landed, and shardmap dispatch stays JSON (`to_parquet`/`from_parquet` remain the Arrow-native side surface only). Answers to the two questions:

### (a) Why `spatial_signature()` stays `"nested"` under `cell_ids_encoding: morton`

`spatial_signature()` is the **shard-map reuse fingerprint**. It's what a `ShardMap` records at build time (`grid_signature`), and what [`runner._check_signature`](https://github.com/englacial/zagg/blob/c299500b74522ff1f6f2c45214918e13908ff43e/src/zagg/runner.py#L370-L379) compares before dispatch — its whole job (#89) is "is this map's *granule→shard assignment* valid for this run's grid":

```python
return {
    "type": "healpix",
    "indexing_scheme": "nested",
    "parent_order": self.parent_order,
    "child_order": self.child_order,
    "layout": self.layout,
}
```

The `cell_ids_encoding` flag changes only what's **written into the `cell_ids` coordinate column** — same `parent_order`/`child_order`, same shard keys, same granule intersection, same chunk layout. So a shardmap built before the flag existed (or under `nested`) is still spatially valid for a `morton`-encoded run; keying the signature on the encoding would invalidate every cached shardmap for no spatial reason. That's the rationale, and `test_spatial_signature_unchanged_by_encoding` pins it.

The honest tradeoff: the store's `dggs.indexing_scheme` attr and the manifest's `grid_signature.indexing_scheme` can now disagree for a morton-encoded run (documented in [docs/morton_arrow.md](https://github.com/englacial/zagg/blob/c299500b74522ff1f6f2c45214918e13908ff43e/docs/morton_arrow.md)). They answer different questions — the manifest field describes the *grid hierarchy* the map was built against, the store attr describes the *written coordinate* — but they share a key name, which is the confusing part. If you prefer strictness: **(a1)** keep as-is (my lean), or **(a2)** record `cell_ids_encoding` as a *separate* field in `spatial_signature()`, with `_check_signature`'s projection treating a missing key as `nested` so old maps stay valid — at the cost that maps then stop being reusable *across* encodings. Your call.

### (b) What "record the encoding in the full signature" means

[`signature()`](https://github.com/englacial/zagg/blob/c299500b74522ff1f6f2c45214918e13908ff43e/src/zagg/grids/healpix.py#L523-L533) = `spatial_signature()` + the output-field set. Its consumer is the co-aggregation gate: `grids.validate_compatible` calls [`nests_with`](https://github.com/englacial/zagg/blob/c299500b74522ff1f6f2c45214918e13908ff43e/src/zagg/grids/healpix.py#L535-L545) on every grid pair to decide whether two products can nest into mutually aligned data cubes, and prints `signature()` in the rejection message:

```python
def nests_with(self, other) -> bool:
    if not isinstance(other, HealpixGrid):
        return False
    return output_field_signature(self.config) == output_field_signature(other.config)
```

**The gap:** two runs differing *only* in `cell_ids_encoding` have identical `signature()` today and `nests_with` says they're compatible — but their `cell_ids` columns live in different id spaces (NESTED ids vs packed morton words). A consumer joining the co-aggregated products on `cell_ids` would silently mismatch every cell. Mixed-encoding co-aggregation is never right.

**Scope if recorded:** two lines + a test, no behavior change for anyone not using the flag —
- `nests_with`: also require `self.cell_ids_encoding == other.cell_ids_encoding`;
- `signature()`: add `"cell_ids_encoding": self.cell_ids_encoding` so the `validate_compatible` error names the actual mismatch.

`spatial_signature()` / shardmap reuse is untouched by this (separate mechanism, question (a)). My recommendation is to land it; happy to push it as a follow-up commit on this branch on your go-ahead — say "land (b)" (and pick (a1)/(a2) above) and I'll fold both in one commit.

---

espg:
> My recommendation is to land it; happy to push it as a follow-up commit on this branch on your go-ahead

Land b ; agree on keeping a1

espg:
🤖 *from Claude*

(b) landed in `f5d7efaea230924a50776575d3f6855cb144a7d5` (phase 7), (a1) kept:

- `HealpixGrid.nests_with` now rejects grids that differ in `cell_ids_encoding` (NESTED ids vs morton words never co-aggregate), and `signature()` records `"cell_ids_encoding"` so the `validate_compatible` rejection message names the mismatch.
- `spatial_signature()` is untouched per (a1) — shard-map reuse stays encoding-free; `test_healpix_excludes_output_fields` updated to pin that the encoding lives in the full signature only.
- New tests: `test_mixed_encodings_do_not_nest` (nested vs morton → False both directions; nested vs default and morton vs morton → True, so nothing changes for flag-off runs) and `test_signature_records_encoding`.

Gates: full suite 1136 passed / 24 skipped locally; ruff/format clean on touched files. Also re-verified against the moved `main` (issues #148/#12 landed since this branch was cut): `git merge-tree` reports a clean merge, no conflicts. All phases complete — awaiting review.

---

PR #167, opened 2026-07-05
Wire h5coro-hidefix (0.2.0) into core deps and the Lambda layer

espg:
Refs #149.

> **⛔ Gated on the h5coro-hidefix 0.2.0 release — CI will be red on dependency resolution until it publishes; re-run CI after the release.** Per the direction on this PR ([comment](https://github.com/englacial/zagg/pull/167#issuecomment-4887569700)), the floor bump to the entry-point release is folded into this PR and the target version is **0.2.0**, which is not on PyPI yet. `uv sync`/`uv lock` fail with "requirements are unsatisfiable" until upstream cuts it.

## What

Wires `h5coro-hidefix` (MIT) into zagg's dependency tiers and the Lambda layer, per the direction on the #149 thread ("it's not an extra -- it needs to be in the lambda ... either that or in core *and* lambda") — that thread comment is both the dependency sign-off and the authorization for the `deployment/aws/build_layer.sh` edit:

- `pyproject.toml` core `dependencies`: `h5coro-hidefix>=0.2.0` (floor, like `h5coro`/`mortie`).
- `pyproject.toml` `[lambda]` extra: `h5coro-hidefix==0.2.0` exact pin, mirroring the numpy/pandas/h5coro pins and carrying the "keep in sync with build_layer.sh" comment.
- `deployment/aws/build_layer.sh`: installed on the existing `--no-deps` line alongside `h5coro==1.0.4` and `mortie` — its only runtime dep is `numpy>=1.21` (as of 0.1.1), already installed earlier in the script. The published 0.1.1 wheels are abi3 `manylinux_2_28` on both x86_64 and aarch64 (~2.2 MB each), so one install line covers both arches and no wheel-side changes are needed for 0.2.0; expected layer growth is **~+2.2 MB per arch** against the 250 MB gate (`lambda-build.yml` enforces it on this PR).
- Layer-size numbers from `lambda-build.yml`: **placeholder — to be copied here from the first green run after the 0.2.0 release publishes** (per the review answer, the exact per-arch unzipped/zipped sizes go in this description).

## Version state: why 0.2.0, and what 0.1.1 lacks

The currently published release is **0.1.1**, and it ships **only the compiled reader** — `h5coro_hidefix/__init__.py` + `h5coro_hidefix.abi3.so`, verified from the installed wheel in a synced env:

```
$ python -c "from importlib import metadata; print(list(metadata.entry_points(group='zagg.index_backends')))"
[]
$ python -c "import zagg.index as z; print(sorted(z.available_index_backends()))"
['hierarchical', 'inline']
```

There is **no `zagg_backend` module and no `zagg.index_backends` entry point** in 0.1.1, so the `sidecar` backend from `docs/virtual_index.md` does not register from it. **0.2.0** is the upstream release expected to carry `h5coro_hidefix/zagg_backend.py::SidecarIndex` plus the entry-point declaration ([version per this comment](https://github.com/englacial/zagg/pull/167#issuecomment-4887569700)); this PR pins directly at it so `sidecar` lights up with zero further zagg-side change the moment it publishes.

## Checklist

- [x] Core dep + `[lambda]` pin + `build_layer.sh` install, in lockstep at 0.2.0
- [x] Lockstep tests (`tests/test_lambda_build.py`): `[lambda]` pin == `build_layer.sh` pin, and pin >= core floor — same pattern as the existing numpy/pandas/h5coro tests
- [x] Import smoke test: `pytest.importorskip("h5coro_hidefix")` — deliberately skip-not-fail so the suite passes in envs that cannot install the unpublished 0.2.0 yet
- [x] Discovery test (`tests/test_index.py::TestBackendRegistry::test_real_environment_discovery`): scans the *real* installed entry points (memo reset, lookup not faked); asserts `hierarchical`/`inline` resolve, and is tolerant of `sidecar` appearing — once 0.2.0 registers it, the test asserts it is a `VirtualIndex` subclass instead of failing CI
- [x] Future-proofed `test_unknown_backend_rejected`: it used `sidecar` as its unknown name against real entry points, so the 0.2.0 release would have broken it spuriously; it now pins entry points empty via the existing `_patch_entry_points` helper
- [ ] After the 0.2.0 release: re-run CI (should go green with no code change) and copy the `lambda-build.yml` layer sizes into this description

## How tested

- With the published 0.1.1 (`uv sync --extra test` before the re-pin): full suite **1431 passed, 26 skipped**; `import h5coro_hidefix` clean; entry-point scan empty as expected.
- After the 0.2.0 re-pin: `uv lock` fails ("unsatisfiable") as expected — 0.2.0 is unpublished; no lock regeneration was possible (the repo does not track `uv.lock`, so nothing stale is committed). `ruff check` + `ruff format --check` clean on touched files; `tests/test_lambda_build.py` + `tests/test_index.py` re-run against the existing env: **81 passed** (lockstep tests parse `pyproject.toml`/`build_layer.sh` text, so they verify the 0.2.0 pins without installing them).

Pre-existing, unrelated: `ruff check src tests` reports one `N818` on `src/zagg/registry.py` (`UnknownCapability` naming) on `main` as well — present since the issue #160 merge; not touched here since the PR lint bot runs `--select=E,F,W,I` (no `N`). Flagging rather than fixing per convention.

## Questions for review

Both earlier questions were answered ([here](https://github.com/englacial/zagg/pull/167#issuecomment-4887569700)): the floor bump is folded into this PR (done — pins now target 0.2.0 directly), and the layer-size numbers will be copied into this description from the first `lambda-build.yml` run after the release. Remaining: none — the ball is upstream (cut the 0.2.0 release with the `zagg_backend` entry point), then CI re-runs green here.

https://claude.ai/code/session_01Pvu3umLej7woxRE8uKUu2a

### DIFF
--- a/deployment/aws/build_layer.sh
+++ b/deployment/aws/build_layer.sh
@@ -70,8 +70,12 @@ $PIP install \
     -t "$OUTPUT_DIR/python" \
     --no-cache-dir
 
-echo "Installing h5coro and mortie (--no-deps)..."
-$PIP install "h5coro==1.0.4" mortie --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
+# h5coro-hidefix (issue #149): compiled hidefix chunk reader backing the
+# `sidecar` index backend. abi3 manylinux_2_28 wheel (~2.2 MB) on both arches;
+# its only runtime dep is numpy (already installed above). Keep the pin in
+# sync with the `lambda` extra in pyproject.toml.
+echo "Installing h5coro, h5coro-hidefix and mortie (--no-deps)..."
+$PIP install "h5coro==1.0.4" "h5coro-hidefix==0.1.1" mortie --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
 
 # Verify numpy stayed < 2.3
 NUMPY_VERSION=$(ls "$OUTPUT_DIR/python" | grep -E "^numpy-" | head -1)
--- a/deployment/aws/build_layer.sh
+++ b/deployment/aws/build_layer.sh
@@ -75,7 +75,7 @@ $PIP install \
 # its only runtime dep is numpy (already installed above). Keep the pin in
 # sync with the `lambda` extra in pyproject.toml.
 echo "Installing h5coro, h5coro-hidefix and mortie (--no-deps)..."
-$PIP install "h5coro==1.0.4" "h5coro-hidefix==0.1.1" mortie --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
+$PIP install "h5coro==1.0.4" "h5coro-hidefix==0.2.0" mortie --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
 
 # Verify numpy stayed < 2.3
 NUMPY_VERSION=$(ls "$OUTPUT_DIR/python" | grep -E "^numpy-" | head -1)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,11 +29,12 @@ dependencies = [
     "pandas>=2.2",
     "h5coro>=1.0.3",
     # Compiled hidefix chunk reader for the `sidecar` index backend (issue #149;
-    # espg sign-off on the #149 thread). NOTE: 0.1.1 ships only the compiled
-    # reader — no zagg_backend module / zagg.index_backends entry point yet, so
-    # `sidecar` does not register from this version. Bump the floor when the
-    # upstream release with the entry point lands.
-    "h5coro-hidefix>=0.1.1",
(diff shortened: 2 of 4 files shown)


espg:
> Layer headroom numbers: lambda-build.yml will print the exact post-change layer sizes on this PR; want them copied into this description once the run finishes, or is the CI log enough?

sure, add them.

> Floor-bump follow-up: when upstream ships the entry-point release, should the bump ride this same branch/PR pattern (small follow-up PR under https://github.com/englacial/zagg/issues/149), or fold into whatever PR wires sidecar into the https://github.com/englacial/zagg/issues/148 benchmark matrix?

fold it in here. assume version 0.2.0

PR #173, opened 2026-07-06
Compiled read path by default: inline via hidefix, read_workers pool, cached/uncached benchmark columns (issue #170)

### LINKED ISSUE #170: Compiled (hidefix) read path is never exercised by default, and its fetch/decode is serial
🤖 *from Claude*

## Summary

zagg 0.14.0 ships the compiled hidefix reader (h5coro-hidefix 0.2.0, a core dep since PR #167) in the Lambda layer — and then, in the default configuration, never calls it. Worse, on the one path that *does* reach it (`sidecar`), both the chunk fetches and the decode are serial, forfeiting most of the measured 16–19× read speedup from the PR #150 benchmark. This issue tracks making the Rust read path the one zagg actually uses, in general — not just in benchmark configs.

## Evidence

**1. The default read path never exercises the compiled decoder.**

- With no `data_source.index:` block, `index_from_config` returns `HierarchicalIndex` (`src/zagg/index/__init__.py:216-219`) — pure-Python h5coro 1.0.4, including its byte-by-byte Python `shuffleChunk` (the 68–70% hotspot from the issue #149 profile; the numpy patch is merged upstream as SlideRuleEarth/h5coro#44 but unreleased, see #154).
- Every config under `tests/data/benchmark/configs/` lacks an `index:` block, so the standing Lambda benchmarks measure the pure-Python path exclusively. This is why the 0.14.0 deploy produced no benchmark movement.
- `inline` doesn't reach the Rust decoder either: its read fn decodes via `h5obj.readDatasets(...)` (`src/zagg/index/inline.py:479,488`). Only `sidecar` (in h5coro-hidefix's `zagg_backend.py`) calls `read_from_buffers`.

**2. The `sidecar` hot path is serial at both stages** (`h5coro-hidefix/python/h5coro_hidefix/zagg_backend.py`, `_read_fn_for`):

```python
buffers = [
    h5obj.ioRequest(int(a), int(s), caching=False)   # one GET per chunk, serial
    for a, s in zip(addrs, sizes)
]
return vidx.read_from_buffers(path, buffers, start, end)  # serial CacheReader decode
```

- Chunk ranges are fetched one blocking GET at a time, uncoalesced.
- `read_from_buffers` decodes through hidefix's serial `CacheReader` (`h5coro-hidefix/src/lib.rs` `assemble_typed`); hidefix 0.12.0's rayon path (`as_par_reader`/`values_to_par`) is file-path-only and its `decode_chunk` is `pub(crate)`, so a Rust-side parallel buffer decode needs an upstream change. But `read_from_buffers` already releases the GIL, so **Python-level threading across read calls gets concurrent fetch + multi-core decode today, with no hidefix release**.

**3. Measured today on the SERC 9-shard fleet (cold containers, per-worker `phase_timings`):** read = **74% of worker wall** under the pure-Python path (median 166 s read / 60 s aggregate / 52 s write; worker median 226 s). Against the PR #150 table (h5coro 93.5 s vs hidefix-with-prebuilt-index ~4.9 s on the o10 shard), a working compiled path should cut ~2–3 minutes off every dense-shard worker — the reduction expected (and not observed) after the 0.14.0 deploy.

## What "fixed" looks like

- `inline` (and therefore `sidecar`'s `on_miss: build` delegate, which wraps `InlineIndex`) decodes through the compiled reader — it already builds the exact chunk map that `Index.from_chunks` consumes (~15 ms rebuild). h5coro-hidefix is a core dep, so this is available unconditionally.
- Chunk-range fetches and buffer decodes run concurrently (ThreadPoolExecutor; the GIL is released during decode).
- The shipped configs and benchmark configs select the compiled path so the weekly numbers track it.
- `hierarchical` remains the fallback/`on_miss: fallback` arm, unchanged.

Implementation plan to follow as a comment; implementation directed by espg (2026-07-06).

Refs: #148, #149, #154, #160, PR #163, PR #167, PR #150 (bench branch, `bench/h5coro/`).

espg:
Closes #170.

## What

Makes the compiled (hidefix) read path the one zagg actually uses, per the plan on the issue ([plan](https://github.com/englacial/zagg/issues/170#issuecomment-4897072816), [espg's answers](https://github.com/englacial/zagg/issues/170#issuecomment-4897166543), [non-ATL03 design note](https://github.com/englacial/zagg/issues/170#issuecomment-4897624120) — **Option B chosen by espg (session, 2026-07-06): one unconditional default flip after the full-read seam, no conditional interim**).

## Phases

- [x] **Phase 1 — inline decodes through the compiled reader** (planned route). Per-dataset single-dataset `Index` (`from_chunks` ~ms), reads via `read_plan` → `ioRequest(caching=False)` → `read_from_buffers` (GIL released); per-dataset degrade to h5coro (warning, never aborts a shard); transient I/O (`None` buffers) surfaces as `OSError`, never misclassified as decode failure. Review findings folded + independently verified closed.
- [x] **Phase 2 — compiled full-read route.** `_read_group_full` (and `_read_segment_broadcasts`) take the same optional `read_fn` seam as `_planned_read_group`; `read_fn=None` keeps the batched h5coro reads byte-identical. `InlineIndex` now serves **read-plan-less (flat) data sources** through it — the shape most non-ATL03 targets have — and its `spatial_index` config requirement is dropped (`chunk_boundaries` mutual exclusion stays). Bonus: the flat compiled route is immune to h5coro's PR #152 start-edge off-by-one (the new parity test's window lands exactly on a chunk boundary; the hierarchical reference cannot serve it).
- [x] **Phase 3 — unconditional default flip**: `index: None` → `inline` for every data source (flat and planned alike); `hierarchical` stays selectable (the uncached benchmark baseline). Sole exception: a-priori `read_plan.chunk_boundaries` sources keep hierarchical (mutually exclusive with inline's addressing). Flat-route chunk-map build failures degrade per dataset (pre-#170 flat reads never required a map); the planned route keeps its re-raise (the map is what makes the PR #152 boundary workaround possible). Worker-logic tests that stub `_read_group` now pin the hierarchical seam explicitly.
- [x] **Phase 4 — `read_workers` thread pool** (default 8, `data_source.read_workers`): the per-dataset reads in `_execute_plan_group` (coord pair + variable/filter fan-out) and `_read_group_full`'s seamed sites run through a bounded pool. Engages ONLY when a compiled `read_fn` is in play — the hierarchical baseline keeps its exact serial batched reads (pinned benchmark series). Thread safety verified: h5coro FileDriver locks reads, S3Driver rides boto3's thread-safe client, `ioRequest` caching has per-cache-line locks; results keyed by path so completion order can't leak into output (byte-identity test at width 1 vs 8).
- [x] **Phase 5 — benchmark columns**: all 11 configs pinned to `{backend: hierarchical}` (series continuity across the default flip), three `*_cached` companion configs+targets (tdigest HEALPix o9/o10/o11, `sidecar` + `on_miss: build` over `zagg-index/ATL03/007` — the same prefix the sponsor example populates, so the caches are shared). Matrix-pinning tests extended for the new read axis.
- [x] **Phase 6 — docs**: README read-backend section (inline default / sidecar / hierarchical baseline + `read_workers`), inline.py module docstring updated for the compiled decode and dual routes.

## How it was tested

- Phase 1: `TestInlineReadGroup` parity suite (row/byte-identical to hierarchical incl. arrow carrier + boundary case vs numpy ground truth); engagement pinned via spy **and** no-fallback-warnings caplog; poison-dataset isolation (`test_bad_dataset_degrades_alone`); I/O-error surfacing (`test_none_buffer_surfaces_as_io_error`).
- Phase 2: flat-source parity vs numpy ground truth (`test_inline_serves_flat_sources`), flat-route engagement (`test_flat_source_engages_compiled_decoder`); `read_fn=None` default byte-identity covered by the untouched hierarchical/planned suites.
- Suites: test_index 84, test_read_plan (in the 84 run), test_processing 194 — all green; ruff clean.

Fleet A/B: PR #169 [benchmark comment](https://github.com/englacial/zagg/pull/169#issuecomment-4897221283) is the pre-change baseline. The deployed fleet runs released zagg, so branch-code fleet numbers require the next release+deploy; a **local** SERC measurement (inline `read_workers` 1 vs 8: wall + peak RSS) lands here instead before ready-for-review, and the deployed A/B rides the release (espg-run, per convention).

## Questions for review

- Phase 1/2 make reads build chunk maps where they previously went straight to `readDatasets` — under `write_back` this widens manifest coverage to full-read datasets (deliberate: fewer sidecar misses later). Flagging in case wider manifests are unwanted.

### DIFF
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -477,28 +477,42 @@ class InlineIndex(VirtualIndex):
         datasets are disjoint from every other group's, so nothing is
         rebuilt or leaked across granules. Under ``write_back`` the cache is
         the instance's pending dict instead, accumulating the granule's maps
-        across groups for ``finish_granule`` to persist. The in-memory
-        ``Index`` is rebuilt (~ms) whenever a new dataset's map lands.
+        across groups for ``finish_granule`` to persist. Each dataset gets
+        its own single-dataset in-memory ``Index`` (~ms), so a spec hidefix
+        rejects degrades that dataset alone.
         """
         maps: dict[str, ChunkMap] = self._pending if self.write_back else {}
-        state: dict = {"vidx": None, "n_maps": -1}
+        # One single-dataset Index per path (~ms each): a dataset whose spec
+        # hidefix rejects (nonzero filter_mask, non-tiling chunk table) then
+        # pins only ITSELF to the fallback — a shared Index rebuilt from all
+        # of ``maps`` would fail every subsequent dataset's reconstruction
+        # and degrade innocent paths with it (review finding, PR #173).
+        indices: dict = {}
         direct: set[str] = set()  # datasets pinned to the h5coro decoder
 
-        def _vidx():
-            if state["n_maps"] != len(maps):
+        def _vidx_for(path):
+            vidx = indices.get(path)
+            if vidx is None:
                 from h5coro_hidefix import Index
                 from h5coro_hidefix.manifest import datasets_from_manifest
 
-                state["vidx"] = Index.from_chunks(
-                    "inline", datasets_from_manifest(granule_manifest(maps))
+                vidx = indices[path] = Index.from_chunks(
+                    "inline", datasets_from_manifest(granule_manifest({path: maps[path]}))
                 )
-                state["n_maps"] = len(maps)
-            return state["vidx"]
+            return vidx
 
         def _compiled(path, start, end):
-            vidx = _vidx()
+            vidx = _vidx_for(path)
             addrs, sizes, _ = vidx.read_plan(path, start, end)
-            buffers = [h5obj.ioRequest(int(a), int(s), caching=False) for a, s in zip(addrs, sizes)]
+            buffers = []
+            for a, s in zip(addrs, sizes):
+                buf = h5obj.ioRequest(int(a), int(s), caching=False)
+                if buf is None:
+                    # h5coro's S3 driver swallows exceptions and returns None;
+                    # surface it as I/O so it is never misclassified as a
+                    # decode failure (review finding, PR #173).
+                    raise OSError(f"ranged read failed for {path} at {int(a)}+{int(s)}")
+                buffers.append(buf)
             return vidx.read_from_buffers(path, buffers, start, end)
 
         def _h5coro_read(path, hyperslice, cm):
@@ -529,6 +543,8 @@ class InlineIndex(VirtualIndex):
                         return _compiled(path, None, None)
                     parts = [_compiled(path, s, e) for s, e in hyperslice]
                     return parts[0] if len(parts) == 1 else np.concatenate(parts)
+                except OSError:
+                    raise  # transient I/O, not a decode problem: fail the read loudly
                 except Exception as exc:
                     direct.add(path)
                     logger.warning(
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -444,7 +444,7 @@ class InlineIndex(VirtualIndex):
             # selectivity fallback and empty-shard early returns — still
             # contribute this group's datasets to the granule manifest.
             self._prebuild_group_maps(h5obj, group, data_source)
-        read_fn = self._chunk_aligned_read_fn(h5obj)
+        read_fn = self._chunk_aligned_read_fn(h5obj, planned=planned)
         if planned:
             return _planned_read_group(
                 h5obj, group, data_source, shard_key, grid, arrow=arrow, read_fn=read_fn
@@ -453,7 +453,7 @@ class InlineIndex(VirtualIndex):
             h5obj, group, data_source, shard_key, grid, arrow=arrow, read_fn=read_fn
         )
 
-    def _chunk_aligned_read_fn(self, h5obj):
+    def _chunk_aligned_read_fn(self, h5obj, *, planned=True):
         """Build the addressing seam: planned ranges, compiled decode.
 
         The chunk maps this backend already builds are exactly what
@@ -522,7 +522,7 @@ class InlineIndex(VirtualIndex):
             parts = []
             for s, e in hyperslice:
                 lo = s
-                if s > 0 and len(cm) and cm.starts_on_boundary(s):
+                if s > 0 and cm is not None and len(cm) and cm.starts_on_boundary(s):
                     lo = s - 1  # h5coro start-edge workaround (see docstring)
                 arr = h5obj.readDatasets([{"dataset": path, "hyperslice": [(lo, e)]}])[path]
                 parts.append(arr[s - lo :])
@@ -534,8 +534,11 @@ class InlineIndex(VirtualIndex):
                 try:
                     cm = maps[path] = build_chunk_map(h5obj, path)
                 except Exception:
-                    if hyperslice is not None:
-                        raise  # planned reads required the map before #170 too
+                    if planned and hyperslice is not None:
+                        # The planned route required the map before #170 too:
+                        # without it the boundary workaround can't run, and a
+                        # plain hyperslice read may trip the PR #152 edge.
+                        raise
                     direct.add(path)
                     logger.warning(f"  no chunk map for {path}; reading through h5coro")
             if path not in direct:
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -7,11 +7,17 @@ metadata-only, ~1 ranged GET + tens of ms per granule (the B-trees live in
 the front-of-file metadata block NSIDC keeps inside h5coro's first cache
 line; measured on PR #159's ``bench/offsets`` route (b), cross-validated
 there against h5py and hidefix chunk-for-chunk over 61 granules with zero
-mismatches). Planned reads are issued as-is — the h5coro decoder already
-inflates exactly the covering chunks — except that the chunk map lets reads
-that start exactly on an interior chunk boundary be detected and shifted one
-element early (see below), keeping output row-identical to ``hierarchical``
-while surviving inputs the plain hyperslice read drops.
+mismatches). Decode goes through the compiled h5coro-hidefix reader (issue
+#170): the chunk map reconstructs a per-dataset in-memory ``Index`` whose
+``read_from_buffers`` inflates exactly the covering chunks, byte-identical
+to h5py/h5coro, with the GIL released. Both read routes are served — the
+planned (chunk-aligned hyperslice) route for sources with
+``read_plan.spatial_index``, and the full-read route for read-plan-less
+(flat) sources — so this backend is the package default for every data
+source. Datasets the compiled reader cannot serve degrade to the h5coro
+decoder per dataset, where the chunk map still lets reads that start
+exactly on an interior chunk boundary be detected and shifted one element
+early (see below).
 
 The chunk map is also the write-back payload: with ``write_back: true``
 (opt-in — issue #160 Q2) plus a ``store``, every granule's accumulated chunk
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -453,33 +453,57 @@ class InlineIndex(VirtualIndex):
         )
 
     def _chunk_aligned_read_fn(self, h5obj):
-        """Build the addressing seam: planned ranges, boundary-safe.
+        """Build the addressing seam: planned ranges, compiled decode.
 
-        The plain planned read is already chunk-minimal with the h5coro
-        decoder (it inflates exactly the covering chunks), so ranges are
-        issued as-is — except when a range starts exactly on an interior
-        chunk boundary, which h5coro's B-tree start-edge intersection drops
-        entirely (PR #152 off-by-one, issue #148 thread). The chunk map makes
-        those starts detectable: shift one element early and trim, paying one
-        extra chunk inflate only in that case (the bench extractor's
-        workaround, applied surgically).
+        The chunk maps this backend already builds are exactly what
+        ``h5coro_hidefix.Index.from_chunks`` consumes (the same
+        ``granule_manifest`` columns ``write_back`` persists), so decode goes
+        through the compiled reader (issue #170): plan the covering chunks,
+        fetch their byte ranges through the worker's h5coro driver
+        (``ioRequest`` — no second credential path), and inflate with
+        ``read_from_buffers`` (byte-identical to h5py/h5coro, GIL released).
+        Datasets the compiled route cannot serve — undecodable dtypes
+        (strings, compounds), empty chunk maps, reconstruction failures —
+        degrade to the h5coro decoder per dataset with a warning, never
+        aborting the shard. The h5coro fallback keeps the PR #152 start-edge
+        workaround: a range starting exactly on an interior chunk boundary
+        is shifted one element early and trimmed (h5coro's B-tree start-edge
+        intersection drops the chunk entirely); the compiled reader has
+        exact ``[start, end)`` semantics and needs no workaround.
 
-        Chunk maps are built lazily per dataset (first planned read of that
-        path) and cached for the life of the returned callable — i.e. one
+        Chunk maps are built lazily per dataset (first read of that path)
+        and cached for the life of the returned callable — i.e. one
         ``read_group`` call, which is exactly one (granule, group): a group's
         datasets are disjoint from every other group's, so nothing is
         rebuilt or leaked across granules. Under ``write_back`` the cache is
         the instance's pending dict instead, accumulating the granule's maps
-        across groups for ``finish_granule`` to persist.
+        across groups for ``finish_granule`` to persist. The in-memory
+        ``Index`` is rebuilt (~ms) whenever a new dataset's map lands.
         """
         maps: dict[str, ChunkMap] = self._pending if self.write_back else {}
+        state: dict = {"vidx": None, "n_maps": -1}
+        direct: set[str] = set()  # datasets pinned to the h5coro decoder
 
-        def read_fn(path, hyperslice=None):
+        def _vidx():
+            if state["n_maps"] != len(maps):
+                from h5coro_hidefix import Index
+                from h5coro_hidefix.manifest import datasets_from_manifest
+
+                state["vidx"] = Index.from_chunks(
+                    "inline", datasets_from_manifest(granule_manifest(maps))
+                )
+                state["n_maps"] = len(maps)
+            return state["vidx"]
+
+        def _compiled(path, start, end):
+            vidx = _vidx()
+            addrs, sizes, _ = vidx.read_plan(path, start, end)
+            buffers = [h5obj.ioRequest(int(a), int(s), caching=False) for a, s in zip(addrs, sizes)]
+            return vidx.read_from_buffers(path, buffers, start, end)
+
+        def _h5coro_read(path, hyperslice, cm):
             if hyperslice is None:
                 return h5obj.readDatasets([path])[path]
-            cm = maps.get(path)
-            if cm is None:
-                cm = maps[path] = build_chunk_map(h5obj, path)
             parts = []
             for s, e in hyperslice:
                 lo = s
@@ -488,5 +512,28 @@ class InlineIndex(VirtualIndex):
                 arr = h5obj.readDatasets([{"dataset": path, "hyperslice": [(lo, e)]}])[path]
                 parts.append(arr[s - lo :])
             return parts[0] if len(parts) == 1 else np.concatenate(parts)
+
+        def read_fn(path, hyperslice=None):
+            cm = maps.get(path)
+            if cm is None and path not in direct:
+                try:
+                    cm = maps[path] = build_chunk_map(h5obj, path)
+                except Exception:
+                    if hyperslice is not None:
+                        raise  # planned reads required the map before #170 too
+                    direct.add(path)
+                    logger.warning(f"  no chunk map for {path}; reading through h5coro")
+            if path not in direct:
+                try:
+                    if hyperslice is None:
+                        return _compiled(path, None, None)
+                    parts = [_compiled(path, s, e) for s, e in hyperslice]
+                    return parts[0] if len(parts) == 1 else np.concatenate(parts)
+                except Exception as exc:
+                    direct.add(path)
+                    logger.warning(
+                        f"  compiled decode unavailable for {path} ({exc}); reading through h5coro"
+                    )
+            return _h5coro_read(path, hyperslice, cm)
 
         return read_fn
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -347,16 +347,12 @@ class InlineIndex(VirtualIndex):
                 "index.store is only meaningful for backend 'inline' with "
                 "write_back: true (inline never reads the store)"
             )
-        # The chunk map drives *addressing*; selection still needs the coarse
-        # spatial index, so the hierarchical read_plan surface is required.
+        # Both read routes accept this backend (issue #170 phase 2): sources
+        # with read_plan.spatial_index take the planned route, read-plan-less
+        # (flat) sources the full-read route -- same compiled addressing seam.
         if data_source is not None:
             rp = data_source.get("read_plan")
-            if not (isinstance(rp, dict) and rp.get("spatial_index")):
-                raise ValueError(
-                    "index backend 'inline' requires data_source.read_plan.spatial_index "
-                    "(chunk-aligned addressing plugs into the planned read path)"
-                )
-            if "chunk_boundaries" in rp:
+            if isinstance(rp, dict) and "chunk_boundaries" in rp:
                 # The a-priori arm (issue #148 arm 2a) takes precedence over
                 # spatial_index inside _read_group, which would silently bypass
                 # this backend's chunk-map addressing -- reject the combination.
@@ -426,30 +422,35 @@ class InlineIndex(VirtualIndex):
         logger.info(f"  inline write-back: {len(maps)} dataset(s) -> {self.store}/{key}")
 
     def read_group(self, h5obj, group, data_source, shard_key, grid, arrow=False, granule_url=None):
-        from zagg.processing.read import _planned_read_group, _validate_planned_config
+        from zagg.processing.read import (
+            _planned_read_group,
+            _read_group_full,
+            _validate_planned_config,
+        )
 
         rp = data_source.get("read_plan")
-        # inline is planned-route only: without a spatial index it would have
-        # to silently fall back to the full read (ignoring the chunk map),
-        # so reject rather than degrade. Completeness of the planned config
-        # itself is the shared gate from the read module.
-        if not (isinstance(rp, dict) and rp.get("spatial_index")):
-            raise ValueError("index backend 'inline' requires data_source.read_plan.spatial_index")
-        _validate_planned_config(data_source)
+        # Two routes, one addressing seam (issue #170 phase 2): sources with a
+        # spatial index take the planned (chunk-aligned hyperslice) route;
+        # read-plan-less (flat) sources take the full-read route — both decode
+        # through the compiled read_fn, so non-ATL03-shaped products get the
+        # fast path too. Completeness of the planned config is the shared
+        # gate from the read module.
+        planned = isinstance(rp, dict) and bool(rp.get("spatial_index"))
+        if planned:
+            _validate_planned_config(data_source)
         if self.write_back:
             # Deterministic write-back coverage (metadata-only, ~ms): built up
             # front so routes that bypass the read seam — the ``full_read``
             # selectivity fallback and empty-shard early returns — still
             # contribute this group's datasets to the granule manifest.
             self._prebuild_group_maps(h5obj, group, data_source)
-        return _planned_read_group(
-            h5obj,
-            group,
-            data_source,
-            shard_key,
-            grid,
-            arrow=arrow,
-            read_fn=self._chunk_aligned_read_fn(h5obj),
+        read_fn = self._chunk_aligned_read_fn(h5obj)
+        if planned:
+            return _planned_read_group(
+                h5obj, group, data_source, shard_key, grid, arrow=arrow, read_fn=read_fn
+            )
+        return _read_group_full(
+            h5obj, group, data_source, shard_key, grid, arrow=arrow, read_fn=read_fn
         )
 
     def _chunk_aligned_read_fn(self, h5obj):
--- a/src/zagg/processing/read.py
+++ b/src/zagg/processing/read.py
@@ -193,7 +193,7 @@ def _segment_level_variables(data_source: dict) -> dict[str, dict[str, str]]:
 
 
 def _read_segment_broadcasts(
-    h5obj, group: str, data_source: dict, levels: dict, n_base: int
+    h5obj, group: str, data_source: dict, levels: dict, n_base: int, read_fn=None
 ) -> dict[str, np.ndarray]:
     """Read each segment-level variable and broadcast it to a base-rate column (issue #30).
 
@@ -215,9 +215,13 @@ def _read_segment_broadcasts(
         index_base = int(link.get("index_base", 0))
         ibeg_path = link["index_beg"].format(group=group)
         cnt_path = link["count"].format(group=group)
-        link_data = h5obj.readDatasets([ibeg_path, cnt_path])
-        ibeg_arr = link_data[ibeg_path]
-        cnt_arr = link_data[cnt_path]
+        if read_fn is None:
+            link_data = h5obj.readDatasets([ibeg_path, cnt_path])
+            ibeg_arr = link_data[ibeg_path]
+            cnt_arr = link_data[cnt_path]
+        else:
+            ibeg_arr = read_fn(ibeg_path)
+            cnt_arr = read_fn(cnt_path)
         for col_name, tmpl in mapping.items():
             if col_name in base_cols:
                 raise ValueError(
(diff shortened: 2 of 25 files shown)


src/zagg/index/inline.py
```diff
@@ -453,33 +453,57 @@ def read_group(self, h5obj, group, data_source, shard_key, grid, arrow=False, gr
...
+        direct: set[str] = set()  # datasets pinned to the h5coro decoder
 
-        def read_fn(path, hyperslice=None):
+        def _vidx():
+            if state["n_maps"] != len(maps):
+                from h5coro_hidefix import Index
+                from h5coro_hidefix.manifest import datasets_from_manifest
+
+                state["vidx"] = Index.from_chunks(
+                    "inline", datasets_from_manifest(granule_manifest(maps))
```
espg: 🤖 *from Claude (review)*

**One undecodable dataset silently disables the compiled route for the whole group — and, under `write_back`, for the rest of the granule.**

`_vidx()` always rebuilds from **all** of `maps`, so a single dataset whose spec `Index.from_chunks` rejects poisons every rebuild: the first failure pins that path to `direct`, but its `ChunkMap` stays in `maps`, so every *subsequent* dataset's first `_compiled()` call re-fails at `from_chunks` and gets pinned too — with a warning that names the innocent path but carries the poison dataset's error. Under `write_back`, `maps` is `self._pending`, which persists across the granule's groups, so one bad dataset in gt1l degrades gt2l/gt3l as well. The docstring's "degrade to the h5coro decoder **per dataset**" doesn't hold for reconstruction failures.

Concrete per-dataset triggers for a `from_chunks` rejection (h5coro-hidefix `parse_spec`/`build_dataset_d`): a nonzero per-chunk `filter_mask` (HDF5 sets it when an optional filter — deflate on an incompressible chunk — is skipped; hidefix's chunk model can't represent it, so `parse_spec` raises), or a chunk table that doesn't tile the shape (partially allocated dataset).

Verified empirically on this branch: injecting one `ChunkMap` with `filter_mask=1` into `_pending` before `read_group` on the fixture makes **every** dataset warn and fall back:

```
compiled decode unavailable for /gt1l/heights/lat_ph (dataset /gt1l/heights/poison: chunk 0 has nonzero filter_mask 1; ...); reading through h5coro
compiled decode unavailable for /gt1l/heights/lon_ph (dataset /gt1l/heights/poison: ...); reading through h5coro
compiled decode unavailable for /gt1l/heights/h_ph (dataset /gt1l/heights/poison: ...); reading through h5coro
compiled decode unavailable for /gt1l/heights/signal_conf_ph (dataset /gt1l/heights/poison: ...); reading through h5coro
```

Rows stay correct (the h5coro fallback is row-identical), so this is a silent performance/observability defect: the compiled path this PR benchmarks quietly turns itself off group-wide on the first bad dataset.

Suggested fix — exclude pinned paths from the rebuild and key the cache on both sets:

```python
def _vidx():
    key = (len(maps), len(direct))
    if state["key"] != key:
        live = {p: m for p, m in maps.items() if p not in direct}
        state["vidx"] = Index.from_chunks(
            "inline", datasets_from_manifest(granule_manifest(live))
        )
        state["key"] = key
    return state["vidx"]
```

After the bad dataset is pinned, the next rebuild excludes it and the remaining datasets keep compiled decode. `maps` and `direct` only ever grow (nothing is deleted or replaced in either), so the tuple key can't go stale.

src/zagg/index/inline.py
```diff
@@ -453,33 +453,57 @@ def read_group(self, h5obj, group, data_source, shard_key, grid, arrow=False, gr
...
+                )
+                state["n_maps"] = len(maps)
+            return state["vidx"]
+
+        def _compiled(path, start, end):
+            vidx = _vidx()
+            addrs, sizes, _ = vidx.read_plan(path, start, end)
+            buffers = [h5obj.ioRequest(int(a), int(s), caching=False) for a, s in zip(addrs, sizes)]
```
espg: 🤖 *from Claude (review)*

**A transient S3 failure is misclassified as "compiled decode unavailable" and pins the dataset to the slow path for the rest of the group.**

`ioRequest(..., caching=False)` is a bare `driver.read()`, and h5coro's `S3Driver.read` (s3driver.py, h5coro 1.0.4) swallows **every** exception — throttling, timeout, connection reset — and returns `None`:

```python
except Exception as e:
    logger.error(f"Error reading {self.key} from {self.bucket_name}: {e}")
    return None
```

That `None` lands in `buffers`, and `read_from_buffers` fails inside hidefix's `extract_buffers` with `buffer N is not bytes-like: cannot make memory view...` — nothing in the message says an S3 GET failed. The broad `except Exception` in `read_fn` then adds the path to `direct` and logs a warning blaming the decoder rather than the network.

Consequences: (a) the warning is not actionable — an operator hunting a slowdown sees decode-capability errors, not S3 throttling; (b) the h5coro fallback refetches and usually succeeds, so a throttling episode silently switches shards off the compiled path, polluting exactly the cached/uncached benchmark columns this PR adds; (c) pre-#170, a planned-read I/O failure surfaced as a read error, not a silent per-dataset downgrade.

Suggested fix — treat a failed fetch as an I/O error, not a decode-capability failure, and let it propagate past the fallback catch:

```python
buffers = [h5obj.ioRequest(int(a), int(s), caching=False) for a, s in zip(addrs, sizes)]
if any(b is None for b in buffers):
    raise OSError(f"{path}: ranged GET failed for a planned chunk")
```

paired with narrowing `read_fn`'s handler (e.g. `except (ValueError, KeyError)` — the two shapes the compiled seam actually produces for capability failures: hidefix spec/buffer rejections and dataset-not-in-index) or an explicit `except OSError: raise` ahead of it.

tests/test_index.py
```diff
@@ -591,6 +591,44 @@ def test_chunk_boundary_start_succeeds_where_plain_hyperslice_fails(self):
...
+        def spy(columns):
+            calls.append(1)
+            return orig(columns)
+
+        monkeypatch.setattr(hh_manifest, "datasets_from_manifest", spy)
+        df = InlineIndex().read_group(
+            _open_fixture(), "gt1l", _fixture_data_source(), 1, _LeafSetGrid(_UNALIGNED_LEAVES)
+        )
+        assert df is not None and len(df) > 0
+        assert calls, "compiled decode was never engaged"
```
espg, line 628: 🤖 *from Claude (review)*

**This test still passes when the compiled decode is completely broken, so it doesn't pin what its comment claims.**

The spy fires inside `_vidx()`, i.e. it proves a compiled *attempt* was made — not that the compiled route served the read. If `Index.from_chunks` or `read_from_buffers` then fails, `read_fn` degrades per dataset and rows stay byte-identical, so both assertions here still hold. Verified on this branch: with `h5coro_hidefix.Index.read_from_buffers` monkeypatched to raise unconditionally, this test body passes as-is — which is exactly the "silent per-dataset fallback" the comment says byte-parity can't distinguish. A regression that breaks compiled decode outright would leave the whole suite green.

Suggested fix — also assert that no fallback fired:

```python
with caplog.at_level(logging.WARNING, logger="zagg.index.inline"):
    df = InlineIndex().read_group(
        _open_fixture(), "gt1l", _fixture_data_source(), 1, _LeafSetGrid(_UNALIGNED_LEAVES)
    )
assert df is not None and len(df) > 0
assert calls, "compiled decode was never engaged"
assert "compiled decode unavailable" not in caplog.text
assert "no chunk map for" not in caplog.text
```

(or spy on `Index.read_from_buffers` and assert it returned successfully, which pins engagement directly).

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Folded in 12c299f: replaced the shared rebuilt-from-all-maps Index with one single-dataset Index per path — a rejected spec now pins only itself; test_bad_dataset_degrades_alone pins exactly one warning naming the offending dataset while the rest of the group stays compiled and row-identical.

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Folded in 12c299f: _compiled now raises OSError on a None ioRequest buffer, and read_fn re-raises OSError explicitly before the broad except — transient I/O propagates instead of silently pinning the dataset; test_none_buffer_surfaces_as_io_error exercises the seam directly (a read_group-wide ioRequest patch breaks h5coro's own selection reads first, which is itself a nice confirmation of your finding's mechanism).

tests/test_index.py
espg: 🤖 *from Claude*

Folded in 12c299f: the engagement test now wraps the read in caplog and additionally asserts no 'compiled decode unavailable' / 'no chunk map' warnings fired — your read_from_buffers-raises repro now fails the test as intended.

src/zagg/index/inline.py
espg: 🤖 *from Claude (review)*

Resolved by 12c299f. Re-ran the original probe against the fold: a `filter_mask=1` ChunkMap injected into `_pending` now produces **zero** "compiled decode unavailable" warnings — every real dataset stays on the compiled route (the poison map is simply never consulted, since each path gets its own singleton `Index`). Also verified the new `test_bad_dataset_degrades_alone` pins exactly one warning naming the offending dataset, and that the per-dataset build changes no `read_plan`/`read_from_buffers` semantics: `chunk_idx` is computed per dataset from its own grid, and both calls are single-dataset lookups with no cross-dataset state — the byte-parity tests confirm. The singleton rebuild is also strictly cheaper than the old shared rebuild-on-growth (O(n) singleton builds vs O(n²) cumulative spec work), and the 1000+-dataset full-coverage case belongs to the sidecar backend, which is unchanged (one Index per granule from the parquet).

src/zagg/index/inline.py
espg: 🤖 *from Claude (review)*

Resolved by 12c299f. Verified beyond the new test: with a flaky `ioRequest` (returns `None` once for `caching=False`), `read_fn` raises `OSError` **and does not pin the path** — a retry through the same closure with healthy I/O goes compiled again and returns the rows, so a transient blip costs one read error, not the compiled route. The `except OSError: raise` sits before `direct.add`, and worker-side the error lands in the per-group handler (`read_errors += 1`, warning) rather than a silent downgrade. Also checked the `write_back` interaction: `finish_granule` still runs after a mid-granule read error, persisting the maps built so far — correct, since each ChunkMap is complete whole-dataset metadata from a successful B-tree walk (independent of the failed data GET) and the manifest contract is explicitly lazy-coverage; the sidecar consumer handles absent datasets via `on_miss`.

tests/test_index.py
espg: 🤖 *from Claude (review)*

Resolved by 12c299f. Re-ran the original adversarial probe against the fold: with `Index.read_from_buffers` monkeypatched to raise unconditionally, the strengthened test's `assert "compiled decode unavailable" not in caplog.text` now trips — a silently-broken compiled decode can no longer leave the suite green.

espg:
🤖 *from Claude (review)*

**Phase 2 adversarial pass (854efe4): no defects found.** Everything below was verified against the actual code in a detached worktree, with `pytest tests/test_index.py tests/test_processing.py` green (251 passed).

What was probed:

1. **Seam completeness in `_read_group_full`** — all five read sites route through `read_fn` when set (coords, coarse flags, link arrays, windowed variables/filters, and both sites inside `_read_segment_broadcasts`). The remaining unseamed `h5obj.readDatasets` calls are the *planned* route's selection/coarse reads (read.py lines 357, 514, and the unseamed `_read_segment_broadcasts` call at 533) — the phase-1 "base-rate addressing only" scope, unchanged and byte-identical either way.
2. **Coarse-flag and link-array parity through the seam** — built a flat-with-levels source (no `read_plan`, cross-level filter on `reference_photon_lat`, segment-level `variables` broadcast): inline output is `assert_frame_equal` to hierarchical with **zero** fallback warnings — including the first-ever compiled decode of the *contiguous-layout* link arrays (`ph_index_beg`/`segment_ph_cnt` pseudo-chunk). 2-D flag shape parity holds (hidefix never squeezes; h5coro returns shaped arrays).
3. **`hyperslice: []` vs `read_fn(path)`** — equivalent full-read semantics; the fallback's plain-path `readDatasets([path])` form is the same full read as the dict-with-empty-hyperslice form.
4. **Mixed full+window reads of one path** — a coordinate also declared as a variable (full read then windowed read through one closure) is row-identical to hierarchical, no fallback.
5. **Flat + `write_back`** — `_prebuild_group_maps` with `si_lvl` None is a clean no-op branch; the persisted manifest covers coords + variables + base filters (prebuild) **plus** the seam-read coarse datasets (flag, both link arrays) — a coverage superset of the planned route, consistent with the documented lazy-coverage contract.
6. **The parity test's PR #152 claim is real, not an artifact** — on `_UNALIGNED_LEAVES` the flat-route window starts at photon 512 = 2×256, an interior chunk boundary; the hierarchical flat read doesn't just drop rows, it hard-fails (`H5Coro encountered error reading gt1l/heights/h_ph: invalid location to read chunk: 1024, 2048` → null dataset → `TypeError` in `_read_group_full`), while inline matches numpy ground truth byte-for-byte on the same window. Gating that test against ground truth instead of the hierarchical reference is the right call.

Two non-blocking observations (no action required to land this phase):

- The ten `test_processing` harness comments say "the issue #170 default otherwise resolves to inline", but at this commit `index_from_config` still defaults to `HierarchicalIndex` (the flip is explicitly deferred per the `index/__init__.py` docstring) — the pins are currently no-ops placed ahead of the flip. Fine as forward-defense; just be aware the comments are premature until the flip commit lands, and misleading if it doesn't.
- Hardening nit for a future pass: now that the flat route sends *both* call shapes through one `read_fn` closure, a path whose `build_chunk_map` raises on a **full** read (pinned to `direct` with no cached map) and is later read with a hyperslice would hit `len(cm)` with `cm=None` in `_h5coro_read` (`TypeError`). Reaching it needs a contrived file+config (e.g. a compact-layout dataset declared as both coordinate and variable), so not a defect today — a `cm is not None and len(cm)` guard would close it for free.

tests/data/benchmark/targets.json
```diff
@@ -75,6 +75,16 @@
       "codec": "inner",
       "sharded": false
     },
+    "tdigest_healpix_o11_cached": {
+      "config": "configs/atl03_tdigest_healpix_o11_cached.yaml",
+      "shardmap": "healpix_o11",
+      "aggregator": "tdigest",
+      "grid_type": "healpix",
+      "grid_size": "o11",
+      "codec": "inner",
```
espg, line 84: 🤖 *from Claude (review)*

**The `*_cached` targets collide with the real `inner` column in the codec figure: `plot_series._codec_layout` slots by `(grid_size, codec)`, and these targets carry `codec: "inner"`.**

`run_benchmark.py` records `codec=target.get("codec")` into the series row but never reads the new `read` key, so a cached target's row is `(grid_size="o9", codec="inner")` — the exact slot key of `tdigest_healpix_o9_inner`. `_codec_layout` (plot_series.py) fills `by_slot[(grid_size, codec)] = target` in row order, last write wins: on the first merge after this lands, each order's "inner" panel in the 2×3 codec figure will silently show **either** the real inner column **or** the cached column, depending on row order in the series parquet.

Verified empirically against this branch:

```python
hist = pd.DataFrame([
    {"target": "tdigest_healpix_o9_inner",  "grid_size": "o9", "codec": "inner"},
    {"target": "tdigest_healpix_o9_cached", "grid_size": "o9", "codec": "inner"},
])
grid, _, _ = plot_series._codec_layout(hist)
assert grid[0][1] == "tdigest_healpix_o9_cached"   # displaced the real inner panel
```

(No parquet **schema** drift — `read: "cached"` is simply ignored by run_benchmark/bench_metrics — but that's the problem: the read axis exists only in the target *name*, invisible to the layout.)

Note the trap doors on the obvious dodges: omitting `codec` from the cached targets routes their rows into the **frozen** historical series instead (`codec.isna` split in `plot_series._codec_mask`), and `test_forward_matrix_is_tdigest_healpix_arrow_codec_ab` currently pins `codec == "inner"` for them. Suggested fix: record the read axis into the series (ctx `read=target.get("read")` + a `read` column in `bench_metrics.SERIES_COLUMNS`) and make `_codec_layout`/`_codec_history` read-aware — either exclude `read == "cached"` rows from the 2×3 grid or extend `CODEC_COLS` with a third `cached` column. (`.github/scripts` is CI tooling, so if touching it is out of scope for this PR, the cached targets should move to `provisional_targets` until the renderer can represent them — as merge-matrix members they corrupt the figure on the first merge.)

One adjacent thing worth confirming before the first run (can't be verified from the repo): `on_miss: build` needs `s3:PutObject` on `s3://sliderule-public-cors/zagg-index/ATL03/007` from the **benchmark Lambda role**. If the write is denied, `finish_granule` swallows the failure (worker logs a warning and continues), the store never populates, and the "cached" column silently measures inline-uncached forever while labeled cached.

src/zagg/processing/read.py
```diff
@@ -659,8 +674,41 @@ def _validate_planned_config(data_source: dict) -> None:
...
+    ``read_fn`` (each is one blocking ranged fetch + a GIL-released decode,
+    so threads overlap S3 latency and use both Lambda vCPUs); the
+    hierarchical h5coro path keeps its serial batched reads regardless — it
+    is the pinned uncached benchmark baseline. Default 8. Peak RSS grows
+    with width (each in-flight read holds its compressed buffers + decoded
+    output), so dense-shard configs can dial it down; ``1`` is serial.
+    """
+    w = data_source.get("read_workers", 8)
+    if isinstance(w, bool) or not isinstance(w, int) or w < 1:
+        raise ValueError(f"data_source.read_workers must be an integer >= 1 (got {w!r})")
```
espg, line 690: 🤖 *from Claude (review)*

**`read_workers` is never validated at config load — a bad value passes `validate_config` and then turns every group read into a swallowed per-group error: silent empty shards instead of a fast failure.**

`validate_config` (zagg/config.py) has no `read_workers` check (it validates `handoff`, `aoi_mask`, filters, links, bounds... at load), so `read_workers: 0` / `"8"` / `2.5` in a YAML sails through submission. The `ValueError` then fires here, inside `_execute_plan_group`/`_read_group_full`, which sits inside the worker's per-group `except Exception` (worker.py `read_errors += 1; logger.warning(...); continue`) — every group of every granule fails identically, the shard completes "successfully" with zero rows and only a `read_errors` count in metadata. On a fleet run that's a config typo becoming silent data loss across every shard, discovered from an empty store rather than a rejected submission. Verified empirically on this branch:

```python
cfg = PipelineConfig(data_source={..., "read_workers": 0}, ...)
validate_config(cfg)                      # passes
_read_workers(cfg.data_source)            # ValueError -- but only at read time, per group
```

Two adjacent asymmetries worth noting: (a) with an explicit `index: {backend: hierarchical}` pin, `read_fn` is None so `_read_workers` is never called and the bad value is silently *ignored*; (b) `test_read_workers_validation` pins the helper's behavior but nothing pins the load-time path, which is where the operator needs the failure.

Suggested fix: validate in `validate_config`'s data_source section (the same one-liner this helper already implements — e.g. call `_read_workers(config.data_source)` or duplicate the check), mirroring how `aggregation.handoff` is validated at load "so a typo in the YAML fails at load, not deep in a worker" (config.py's own words for that key).

tests/data/benchmark/configs/atl03_tdigest_healpix_o9.yaml
```diff
@@ -23,6 +23,11 @@
 # Confidence filter, multi-level form and read_plan match atl03.yaml (the planned
 # segment->photon IO of issue #43 applies here too); only the aggregation differs.
 data_source:
+  # Pinned read backend (issue #170): the UNCACHED pure-h5coro baseline
+  # column. The package default flipped to the compiled `inline` path; an
+  # unpinned config would silently move this series onto the fast path.
+  index:
+    backend: hierarchical
```
espg, line 30: 🤖 *from Claude (review)*

**This pin also freezes the two 88S provisional stress probes on pure h5coro, stranding their documented purpose.**

`provisional_targets.tdigest_healpix_o9_88s` / `o10_88s` (targets.json) reference **this same config** (and o10's), and their `_comment` says they are "the issue #148 stress probes: on-demand runs against the pinned worst 88S ring shard, **expected to OOM/timeout until the streaming/cached-read work lands**". The cached-read work is landing in this very PR — but with the shared base config now pinned `hierarchical`, the 88s probes can never see it: they'll OOM/timeout on the h5coro baseline forever, and the one arm built to demonstrate the #148 fix can't demonstrate it.

Keeping the *matrix* baseline honest is right (that's what this pin is for); the gap is only that the 88s probes inherited the pin through the shared config. Suggested fix, either: (a) add `tdigest_healpix_o9_88s_cached` / `o10_88s_cached` provisional companions pointing at the new `*_cached.yaml` configs with the existing `healpix_*_88s` shardmaps — on-demand `--target` runs that can actually answer "does the compiled/cached read survive the worst shard?"; or (b) if that's deliberately deferred, update the provisional `_comment` so it stops promising an outcome the pinned config now precludes.

tests/data/benchmark/targets.json
espg: 🤖 *from Claude*

Folded in 659382c: run_benchmark now records read=target.get('read') into the series context, and _codec_layout keys cached targets into their own third column (CODEC_COLS grew to sharded/inner/cached; slot label is read-aware). Layout tests updated for the 3x3 grid. Note this touches .github/scripts/ — in-scope here because the issue's phase 5 (two benchmark columns, espg-directed) is unrepresentable in the series without it; flagging per the CLAUDE.md infra rule for review. Your IAM caveat is real and stands for espg: the benchmark Lambda role needs PutObject on sliderule-public-cors/zagg-index/* for on_miss:build self-population (the execution-role grant landed today for the sponsor stack — worth verifying it covers the benchmark function too, else the cached column silently measures inline-uncached until manifests exist).

src/zagg/processing/read.py
espg: 🤖 *from Claude*

Folded in 659382c: validate_config now rejects a bad data_source.read_workers at submission (mirrors the handoff guard; same message as _read_workers). The silently-ignored-under-hierarchical-pin case remains by design — the key is documented as compiled-path-only — but it now at least fails loudly when malformed anywhere.

tests/data/benchmark/configs/atl03_tdigest_healpix_o9.yaml
espg: 🤖 *from Claude*

Folded in 659382c: added tdigest_healpix_{o9,o10}_88s_cached provisional companions (same pinned 88S shardmaps, *_cached configs, read:'cached') and updated the provisional _comment — the base _88s probes stay pinned hierarchical as the honest issue #148 arm-1 baseline, and the pair now measures exactly the cached-vs-uncached delta at the pole.

espg:
do (a)

espg:
🤖 *from Claude*

Re "do (a)": already landed in 659382c (crossed with your comment) — `tdigest_healpix_o9_88s_cached` / `tdigest_healpix_o10_88s_cached` provisional companions on the `*_cached.yaml` configs + the existing `healpix_*_88s` shardmaps, with the provisional `_comment` updated to describe the cached-vs-uncached pairing at the pole. On-demand `--target` runs as before; nothing joins the every-merge matrix.

tests/data/benchmark/targets.json
espg: 🤖 *from Claude (review)*

**Not resolved — the read axis never reaches the series parquet, so the collision is intact.** 659382c threads `read` into run_benchmark's ctx and makes `_codec_layout` read-aware, but the pipe between them was not extended:

- `bench_metrics.build_record` builds the record from an explicit key list that has no `"read"` — the ctx value is dropped there;
- even if it weren't, `update_series.records_to_frame` reindexes to `bench_metrics.RECORD_COLUMNS` (no `"read"`), which by design drops any column not in the schema.

Verified end-to-end on 659382c with exactly run_benchmark's ctx for a cached target:

```
record carries read: False
frame has read column: False
o9 row [sharded, inner, cached]: [None, 'tdigest_healpix_o9_cached', None]
```

The cached target still lands in (and here displaces) the **inner** slot; the new third column stays blank forever. Fix is the two missing pieces: append `"read"` to `RECORD_COLUMNS` (stable-schema rule: last position) and add `"read": context.get("read")` to `build_record`. Frozen safety holds in both worlds: a pre-existing parquet has no `read` column and `Series.get("read", "")` returns the default; once the column exists, legacy rows read back NaN and `str(nan) != "cached"`, so they keep keying on codec — I verified the layout tests' frozen paths pass either way.

Coverage note (why the suite stayed green through this): both `_codec_layout` tests assert the third column is `None` on histories with **no cached rows** — nothing drives a cached row through `records_to_frame` → `_codec_layout`. A one-row addition to `_full_codec_matrix` (a cached record built through `build_record`) would have caught this and will pin the fix.

src/zagg/processing/read.py
espg: 🤖 *from Claude (review)*

Resolved by 659382c. Verified empirically: `validate_config` now rejects `read_workers` 0 / -1 / `True` / `"eight"` / 2.5 at submission (each with the `read_workers` message) and accepts 4 and absent; the guard mirrors `_read_workers`' semantics including the bool-is-int trap. A config typo now fails at load instead of surfacing as per-group swallowed read errors and silently-empty shards.

tests/data/benchmark/configs/atl03_tdigest_healpix_o9.yaml
espg: 🤖 *from Claude (review)*

Resolved by 659382c (option (a) as confirmed on the thread). Verified: `tdigest_healpix_o9_88s_cached` / `o10_88s_cached` point at the `*_cached.yaml` configs with the existing `healpix_*_88s` shardmaps, stay provisional (on-demand `--target` only, never the merge matrix), and the `_comment` now states the pairing correctly — base `_88s` pinned hierarchical as the honest uncached reference, `_cached` companions measuring the compiled/sidecar read on the same worst-case shard.

tests/data/benchmark/targets.json
espg: 🤖 *from Claude*

Right — the first fold wired both ends and dropped the middle. Fixed in the follow-up commit: `read` appended to `RECORD_COLUMNS` (stable-schema: newest last; the order test now pins read last, codec second-to-last) and `build_record` threads `context.get("read")`. Coverage closed the way you suggested: `_cached_row` builds through the real `bench_metrics.build_record`, and the layout test now asserts the full ctx→record→frame→layout pipe lands each cached target in the third column with the true inner target keeping its slot. Your exact repro (o9 row displacement) is what the new assertion fails on if the pipe breaks again.

espg:
🤖 *from Claude*

Marking ready for review per espg (merge call on the session thread). State: all six phases complete; eight adversarial-review findings across four passes folded and verified closed (the last — the read-column record-schema gap — in adb2764). CI running on adb2764 (ruff green; the commit touches only bench_metrics + its tests). The local `read_workers` 1-vs-8 A/B (wall + peak RSS, dense SERC shard over https) is still running and will be posted here / on #170 when it lands — espg's call to merge ahead of it, with the definitive fleet A/B riding the 0.15.0 release (espg-run):

- deploy order suggestion: release → layer/function deploy → SERC fleet A/B (validates the new default + `read_workers` on-fleet) → the 88s `*_cached` provisional probes (the issue #148 arm this PR unblocks).
- release-blockers on the ops side (from the PR #172 thread): the benchmark role needs `lambda:UpdateFunctionConfiguration` (or run_benchmark pins `force_cold=False`), and Get/Put on `sliderule-public-cors/zagg-index/*` for the cached column's self-population.

espg:
🤖 *from Claude*

Correction to the release-blockers note two comments up: the `force_cold`/benchmark-role item is **void**. PR #172's final form (phases A–D on the thread) flips `force_cold` back to default `False` (explicit certification-run tool only) and adds worker-side **self-recycle**: `ZAGG_RECYCLE_MAX_INVOCATIONS=1` on the function environment forces a fresh sandbox for every invocation with **no caller IAM at all** (the worker mirrors its result to S3 first, then `os._exit(0)`; `MaximumRetryAttempts: 0` keeps billing clean), and `ZAGG_RECYCLE_RSS_MB` bounds the ratchet surgically within runs. espg's no-grant decision is recorded on the #172 thread.

The one remaining ops item for the 0.15.0 cached benchmark column: Get/Put on `sliderule-public-cors/zagg-index/*` for the **benchmark stack's** execution role (the sponsor stack got the `/*` fix today; the benchmark function's role is separate).

espg:
🤖 *from Claude*

## Deferred `read_workers` A/B (posted post-merge per espg)

Local run, densest SERC shard (5347404432944398346: 50 granules, 1.54M obs), `inline` backend, **https driver over a residential link**, single process, identical config except `data_source.read_workers`:

| `read_workers` | wall | peak RSS (`ru_maxrss`) | total_obs | cells_error |
|---|---|---|---|---|
| 1 | 2823.5 s | 361.8 MB | 1,538,493 | 0 |
| 8 | 2080.5 s | **351.8 MB** | 1,538,493 | 0 |

Takeaways:

- **Memory: the default-8 concern does not materialize.** Peak RSS is flat (−10 MB, noise) at width 8 on the worst shard — in-flight reads hold one dataset's window buffers each, and only ~5 datasets are ever eligible per group, so the pool's memory ceiling is naturally bounded well below the width. The `read_workers: 8` default stands validated on the axis that mattered (#169 context).
- **Output identity holds under the pool**: byte-count-identical obs, zero errors, matching the fleet's dense-shard count from the PR #169 benchmark.
- **Wall: 1.36×, exactly the bounded speedup predicted.** The phase-4 pool overlaps reads *across datasets* (~5 wide/group); each dataset's chunk fetches remain serial inside the binding's `read_fn`, groups/granules remain sequential, and per-request EDL/https overhead dominates this link. The remaining headroom lives in the **h5coro-hidefix binding PR** (chunk-level fetch pool + concurrent hyperslice runs — the tracked follow-up on #170), and the honest wall numbers come from the post-deploy fleet A/B (in-region S3, sidecar path), which rides the 0.15.0 release.

Raw JSONs retained locally; store outputs were local scratch zarrs (nothing published).

Issue #154, opened 2026-07-01
Hand-off: upstream numpy-vectorized shuffleChunk to h5coro


espg:
🤖 *from Claude*

Tracking issue opened at espg's direction on #149 ([comment](https://github.com/englacial/zagg/issues/149#issuecomment-4861112544)).

## What

The #149 investigation (PR #150) found that h5coro's pure-Python shuffle-filter decode (`shuffleChunk`, `h5coro/h5dataset.py`) is **68–70% of zagg's granule-read wall time** — `signal_conf_ph` is the only shuffled dataset zagg reads, but at `n_photons × 5` int8 it's the largest by element count. Rewriting the element-wise loop as a numpy byte-matrix transpose (~10 lines) gives **3.8× reads / 3.9× end-to-end worker**, byte-identical output (sha256 gate over the full captured workload + randomized property test incl. ragged tails; the out-of-range window case raises `FatalError` where the old loop raised `IndexError`).

## Where the change lives

- Staged patch: [`bench/h5coro/patches/0001-vectorize-shuffleChunk-as-a-numpy-byte-matrix-transp.patch`](https://github.com/englacial/zagg/blob/claude/149-read-bench/bench/h5coro/patches/0001-vectorize-shuffleChunk-as-a-numpy-byte-matrix-transp.patch) (PR #150)
- Local clone: `~/software/h5coro`, branch `numpy-vectorize` — note it sits on top of the clone's local `main`, which carries **2 unpublished commits** (memory handling / close-context-manager); the upstream PR must be rebased onto `SlideRuleEarth/h5coro` `main` so it doesn't drag those along.

## Upstream PR checklist

- [x] Rebase the commit onto upstream `main` (exclude the 2 unpublished local commits)
- [x] Upstream test suite green locally (note env-dependent skips); add a shuffle byte-equivalence unit test if it fits upstream's test layout
- [x] Branch pushed to the espg fork; PR opened against SlideRuleEarth/h5coro with the benchmark table from PR #150's `REPORT.md`
- [x] Merged upstream + released
- [ ] zagg pin bump (`h5coro>=<release>`) — separate `small-fix` once released

Refs #149; benchmark evidence in PR #150 (`bench/h5coro/results/REPORT.md`).

*(Left unlabeled for espg to triage.)*

espg:
🤖 *from Claude*

[...]

PR #181, opened 2026-07-07
Layer build still pins h5coro 1.0.4 — sync to 1.0.5 (issue #154)

espg:
Refs #154, #170.

The 0.16.0 pyproject bump (core `h5coro>=1.0.5`, lambda extra `==1.0.5`) missed the third pin site: `deployment/aws/build_layer.sh:78` still installs `h5coro==1.0.4` into the layer — verified in the tier-2 layer `:50` built from that merge (contains `h5coro-1.0.4.dist-info` alongside h5coro-hidefix 0.3.0). Without this, the 0.16.0 production deploy ships a layer whose h5coro contradicts the package floor, and the fleet never gets the 3.8x numpy `shuffleChunk` decode (SlideRuleEarth/h5coro#44).

One line: `h5coro==1.0.4` → `==1.0.5` in the layer install. The benchmark tier-2 auto-deploy of this PR's merge is also what unblocks arm (b) of the o9 read matrix (#170 thread).

Same trap as always (§7: 'keep these pins in sync with the layer') — a follow-up worth considering: a test that asserts the build_layer.sh pins match the `lambda` extra, so this can't drift silently again.

### DIFF
--- a/deployment/aws/build_layer.sh
+++ b/deployment/aws/build_layer.sh
@@ -75,7 +75,7 @@ $PIP install \
 # its only runtime dep is numpy (already installed above). Keep the pin in
 # sync with the `lambda` extra in pyproject.toml.
 echo "Installing h5coro, h5coro-hidefix and mortie (--no-deps)..."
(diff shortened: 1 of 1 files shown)


espg:
Folding directly into main per espg (staged locally with the release push); no PR.

PR #182, opened 2026-07-07
Small fixes: recycle generation counting + stand_up.sh artifact defaults

### LINKED ISSUE #174: stand_up.sh artifact defaults are stale (source.coop → sliderule-public-cors migration): default run fails NoSuchKey on a phantom version
🤖 *from Claude*

Filed at espg's direction after tonight's failed `zagg-backend` stack update (forensics below): `stand_up.sh`'s artifact-location defaults predate the source.coop → `sliderule-public-cors` migration, so a default-env run references a bucket the release pipeline no longer stages to — and a version derivation that can point at a never-staged minor.

## What happened (2026-07-06, `zagg-backend`)

1. **06:56 UTC update — succeeded**: ExecutionRole gained the `zagg-index/*` grant (issue #160), DepsLayer `:16` built from `s3://sliderule-public-cors/0.14/lambda_layer_arm64.zip`, `ExtractFn` created. This is the state the stack is in now.
2. **22:49 UTC update — failed and auto-rolled-back** (`UPDATE_ROLLBACK_COMPLETE`, restoring the 06:56 state): `DepsLayer` `UPDATE_FAILED` with S3 `NoSuchKey`. The failed event's properties show it referenced:

   ```
   bucket: us-west-2.opendata.source.coop
   key:    englacial/zagg/lambda/0.15/lambda_layer_arm64.zip
   ```

   Wrong bucket (the retired pre-0.14 hosting — every stack event before July 6 shows `us-west-2.opendata.source.coop/englacial/zagg/lambda/0.N/...`; the 0.14.0 release's `distribute` step stages to `sliderule-public-cors/<minor>/` instead) **and** a phantom version (`0.15` has never been staged anywhere — plausibly derived from post-tag `main` being ahead of `0.14.0`).

Not a permissions problem, and no manual rollback was needed — but the run was a confusing failure for something that should be routine, and the same defaults will bite any fresh standup or any operator who reaches for `stand_up.sh` without overriding env.

## Scope

1. Point `stand_up.sh`'s default artifact bucket/prefix at the current release-pipeline layout (`sliderule-public-cors/<minor>/…`, per `publish.yml`'s `distribute` job / `distribute_zips.sh`), removing the source.coop defaults.
2. Make the version derivation refuse to guess: resolve to an EXISTING staged minor (verify the layer key with a HEAD request before calling CloudFormation, failing with an actionable message listing available prefixes) rather than deriving an unstaged version from git state.
3. Consider echoing the resolved bucket/key/version and requiring a confirm (or a `--yes`) before the stack call — tonight's failure would have been a 2-second catch at that prompt.
4. Update `deployment/aws/` README/docs where the old locations are referenced.

Interim workaround (documented in the stack forensics on the thread): bypass the script — `aws cloudformation deploy --stack-name zagg-backend --template-file deployment/aws/template.yaml --capabilities CAPABILITY_IAM` reuses the previous (correct) artifact parameters and picks up new template parameter defaults.

*(Left unlabeled for espg to triage — touches `deployment/aws/`, so per repo convention the implementing PR needs this issue as its named authorization.)*

### LINKED ISSUE #177: Self-recycle generation counts the sync setup invocation: MAX_INVOCATIONS=1 yields generation-2 workers
🤖 *from Claude*

## Summary

`ZAGG_RECYCLE_MAX_INVOCATIONS=1` does not deliver all-cold *workers*: the runner's synchronous **setup** invocation (zarr template creation, ~0.5 s) warms a fresh container first, and self-recycle is async-path-only by design (a sync response would be lost on `os._exit`). The worker then lands on that warm container at **generation 2**, and recycle fires only after it.

Observed on the 2026-07-07 00:52 UTC manual benchmark run (0.15.0 fleet, `ZAGG_RECYCLE_MAX_INVOCATIONS=1` set):

- worker envelopes: `container_cold: false, container_generation: 2`
- CloudWatch: `ZAGG_SELF_RECYCLE rss_mb=747 generation=2 threshold=1` — fired *after* the worker, not before it.

Bounded and mostly harmless (each container serves exactly one heavy invocation before dying), but the RSS baseline the worker inherits from setup (~215 MB container HWM) muddies per-invocation memory telemetry, and "max invocations 1" reading as "every worker cold" is a reasonable-but-wrong expectation.

## Options

1. **Count only recycle-eligible (async) invocations toward the recycle generation** — setup (sync) doesn't increment the counter; telemetry keeps reporting the true container generation separately. Worker then recycles at its own threshold count. Small handler change.
2. Dispatch setup with an event flag the handler uses to skip counting (equivalent effect, more plumbing).
3. Document the semantics and do nothing (the setup invocation's ~215 MB footprint is small; the ratchet protection still works).

(1) looks like the small-fix. Refs #171, PR #172, the #169 warm-ratchet forensics.

espg:
Closes #177
Closes #174

Bundled `small-fix` PR (one commit per issue, plus one review-fold commit). The `deployment/aws/` touch is authorized by the issue naming it and by the maintainer comment on the thread (https://github.com/englacial/zagg/issues/174#issuecomment-4900055257).

## Checklist

- [x] #177 — count only recycle-eligible (async) invocations toward the self-recycle budget (`deployment/aws/lambda_handler.py`)
- [x] #174 — `stand_up.sh` artifact defaults: current release layout, refuse-to-guess version verification, confirm prompt, docs (`deployment/aws/stand_up.sh` + docs)
- [x] fold adversarial-review findings (see "Review fold" below)

## Issue #177 — recycle budget vs. container generation

Option (1) from the issue. A new module global `_ASYNC_INVOCATIONS_SERVED` is billed at the dispatcher, on every async (`result_url`) invocation, and `ZAGG_RECYCLE_MAX_INVOCATIONS` now compares against it instead of the raw generation:

```python
elif gen_limit > 0 and async_served >= gen_limit:
    threshold = gen_limit
```

So the runner's synchronous setup invocation (the one observed warming a container in the 2026-07-07 00:52 UTC benchmark run) no longer consumes the worker budget: `MAX_INVOCATIONS=1` now means "one heavy async invocation per container". Telemetry is unchanged — `_INVOCATIONS_SERVED` / `container_generation` still count every invocation (the existing `test_setup_counts_toward_generation_but_body_unchanged` pins this). The `ZAGG_SELF_RECYCLE` log line now carries both counters (`async_served=<n> generation=<n>`); the metric filters in `template.yaml` key only on the `ZAGG_SELF_RECYCLE` prefix, so they are unaffected. `docs/deployment/lambda.md` updated to match.

## Issue #174 — stand_up.sh stale artifact defaults

Per the issue's four scope items:

1. **Current release layout only.** The `MIRROR_BUCKET`/`MIRROR_PREFIX` (`us-west-2.opendata.source.coop/englacial/zagg/lambda`) defaults and the silent fallback to them are removed. The single source is `s3://sliderule-public-cors/<minor>/<zip>` + root `versions.json` — matching `publish.yml`'s `distribute` job / `.github/scripts/distribute_zips.sh`. Self-hosting still works via `DIST_BUCKET`/`DIST_PREFIX`/`DIST_REGION`.
2. **Refuse to guess.** After the minor resolves (explicit / `latest` / git tag / installed zagg), both the layer and function keys are verified with `aws s3api head-object ... --no-sign-request` before any CloudFormation call. An unstaged minor (the failed run's phantom `0.15`) fails fast with the `aws` stderr surfaced, the staged minors read from `versions.json`, and a pointer to `LAMBDA_VERSION`.
3. **Echo + confirm.** The resolved version/layer/function/stack are echoed and a `Proceed with deploy? [y/N]` prompt gates the stack call; `--yes` skips it. EOF on stdin (unattended without `--yes`) aborts safely with the actionable message.
4. **Docs.** `docs/deployment/standup.md` (steps, env-var table, release-population paragraph), `deployment/LAMBDA_DEPLOYMENT.md`, `README.md`, and the `docs/deployment/benchmark-cicd.md` transition section updated; stale `source.coop` artifact references removed (the source.coop mentions left in `docs/deployment/lambda.md` are about *output* stores, not artifacts, and stand alone).

Nothing was run against AWS — all changes verified statically and via the stubbed-CLI tests below.

## Review fold (commit `fold review: bill recycle budget on failed mirror + stand_up HEAD/EOF fixes`)

All four diff-scoped findings from the adversarial review folded:

- **Failed-mirror billing (correctness):** the budget increment moved out of `_maybe_self_recycle()` (now a pure check) to the dispatcher's `result_url` branch, before the mirror-success gate — a failed-mirror async invocation is served and now burns budget, matching pre-#177 behavior. New `test_failed_mirror_still_bills_recycle_budget` (mirror fails on invoke 1, succeeds on invoke 2, cap 2 → recycle fires on invoke 2).
- **HEAD both keys:** chose to verify **both** the layer and function zips (a two-iteration loop; a partially staged minor now also fails fast) rather than scoping the header down — the header/echo claims and `docs/deployment/standup.md` now match the code. New `test_partially_staged_minor_fails_before_any_stack_call`.
- **HEAD error detail:** the `head-object` stderr is captured and printed in the error block, so a network/credential failure of the check reads differently from a 404 (message says so explicitly).
- **EOF prompt abort:** `read ... || REPLY=""` so an unattended run without `--yes` prints "Aborted (pass --yes ...)" instead of dying silently under `set -e`; `test_no_input_and_no_yes_aborts` now asserts the message.

## How tested

- **#177:** `tests/test_lambda_handler.py` — new `test_sync_setup_does_not_consume_recycle_budget` (cap 2: sync setup + first worker stays below the cap — fails on pre-fix code — second worker recycles), `test_recycle_log_reports_true_generation` (cap 1 after setup: recycles on the first async worker with `async_served=1 generation=2`, envelope `container_generation == 2`), and `test_failed_mirror_still_bills_recycle_budget`; the two existing log-format assertions updated. 80/80 pass.
- **#174:** new `tests/test_stand_up.py` runs the real script with a stub `aws` on `PATH` (no network, nothing deployed), mirroring the existing `tests/test_distribute_zips.py` harness: unstaged and partially staged minors fail before any `cloudformation` call with staged minors listed; staged minor deploys with `ArtifactBucket=sliderule-public-cors` / `LayerS3Key=<minor>/lambda_layer_arm64.zip`; `LAMBDA_VERSION=latest` resolves via `versions.json`; prompt aborts on `n`/EOF (with message) and proceeds on `y`/`--yes`; unknown flags rejected. 8/8 pass, plus `bash -n`.
- Full suite: `pytest -q` → **1480 passed, 26 skipped** (pre-fold; folded files re-run green at 88/88); `ruff check` and `ruff format --check` clean on every touched file.

## Questions for review

- **`deployment/aws/lambda_handler.py` is now ~1035 lines**, past the ~1000-line convention ceiling (it was already over at base, 1012). Standing note from the review: a decision on whether/how to split (or bless the size) belongs on a follow-up issue rather than this small-fix PR.
- **Pre-existing lint drift on `main` (not touched here):** `ruff check src tests` fails with one `N818` (`src/zagg/registry.py:64`, `UnknownCapability`), and `ruff format --check` would reformat 7 files (`src/zagg/__init__.py`, `src/zagg/__main__.py`, `src/zagg/catalog/sources.py`, `src/zagg/viz/crs.py`, `tests/test_catalog.py`, `tests/test_extract.py`, `tests/test_integration.py`). Both reproduce on a clean `origin/main` checkout, so they are flagged rather than fixed per repo convention.
- **`template.yaml` `RecycleMaxInvocations` description** now reads slightly stale ("recycles after every invocation" → after every *async* invocation). Issue #177 names only `lambda_handler.py`, so the template was left untouched per the deployment-files convention — happy to fold the one-line description update in if that's preferred.
- **`publish_mirror.sh`** (the legacy source.coop publisher) still exists; the benchmark-cicd doc previously said to retire it "in a follow-up". This PR only removes the *consumption* side in `stand_up.sh`; deleting the script itself is left for a separate decision.
- **stand_up.sh behavior change:** older minors that only ever lived on the source.coop mirror now require `DIST_BUCKET`/`DIST_PREFIX` overrides instead of falling back silently. This reads as intended by the issue ("removing the source.coop defaults") but is a hard cut for pre-0.14 standups.

https://claude.ai/code/session_01Wp83fdkH2qhtEvoLgyTWU1

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -150,8 +150,8 @@ logger.setLevel(logging.INFO)
 _CONTAINER_INIT_TS = time.time()
 _INVOCATIONS_SERVED = 0
 # Recycle budget (issue #177), counted separately from the true generation
-# above: only recycle-eligible (async, result_url-mirrored) invocations are
-# billed against ``ZAGG_RECYCLE_MAX_INVOCATIONS``. A synchronous setup/
+# above: only recycle-eligible (async ``result_url``) invocations are billed
+# against ``ZAGG_RECYCLE_MAX_INVOCATIONS``. A synchronous setup/
 # finalize invoke still warms the sandbox (the generation keeps counting it,
 # and telemetry keeps reporting it) but must not consume the worker budget --
 # MAX_INVOCATIONS=1 means "one heavy async invocation per container", not
@@ -260,9 +260,11 @@ def _maybe_self_recycle() -> None:
     exiting so dashboards can split intentional recycles from real crashes
     (metric-filter note in docs/deployment/lambda.md). The line carries both
     the async budget spent and the true container generation.
+
+    Pure check: the async budget itself is billed at the dispatcher (every
+    ``result_url`` invocation, mirror success or not), so a failed mirror --
+    which skips this call -- still burns the budget (issue #177 review fold).
     """
-    global _ASYNC_INVOCATIONS_SERVED
-    _ASYNC_INVOCATIONS_SERVED += 1
     rss_limit = _recycle_limit("ZAGG_RECYCLE_RSS_MB")
     gen_limit = _recycle_limit("ZAGG_RECYCLE_MAX_INVOCATIONS")
     kib = _read_vmrss_kib()
@@ -462,6 +464,13 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     # result_url for it to poll. Covers every branch (200 / 400 / 500) of both
     # per-unit handlers (spatial process, temporal process_event -- #12 Phase 8).
     if event.get("result_url"):
+        # Bill this async invocation against the recycle budget (issue #177)
+        # BEFORE the mirror-success gate: the invocation was served either
+        # way, so a failed mirror must not stretch the sandbox's budget (the
+        # pre-#177 generation counter counted it too). Only the recycle
+        # itself stays gated on the mirror landing.
+        global _ASYNC_INVOCATIONS_SERVED
+        _ASYNC_INVOCATIONS_SERVED += 1
         mirrored = _write_result(event["result_url"], response, event)
         # Self-recycle strictly AFTER a successful result mirror (issue #171):
         # the orchestrator polls the result object, not the Lambda response
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -149,6 +149,14 @@ logger.setLevel(logging.INFO)
 # requiring CloudWatch forensics.
 _CONTAINER_INIT_TS = time.time()
 _INVOCATIONS_SERVED = 0
+# Recycle budget (issue #177), counted separately from the true generation
+# above: only recycle-eligible (async, result_url-mirrored) invocations are
+# billed against ``ZAGG_RECYCLE_MAX_INVOCATIONS``. A synchronous setup/
+# finalize invoke still warms the sandbox (the generation keeps counting it,
+# and telemetry keeps reporting it) but must not consume the worker budget --
+# MAX_INVOCATIONS=1 means "one heavy async invocation per container", not
+# "one invocation of any kind".
+_ASYNC_INVOCATIONS_SERVED = 0
 
 
 def _container_telemetry() -> Dict[str, Any]:
@@ -239,30 +247,38 @@ def _maybe_self_recycle() -> None:
       against a 2047 MB cap, so the template's 1400 catches a dirty sandbox
       after roughly one heavy retention while leaving the triggering
       invocation ~650 MB of headroom to complete first.
-    - ``ZAGG_RECYCLE_MAX_INVOCATIONS`` -- generation cap (template default
-      1: recycle after every invocation, the cold-every-time posture; raise
-      it for a belt-and-suspenders cap over retention modes the RSS read
-      misses -- and the only check that fires off-Linux, where RSS reads
-      are None).
+    - ``ZAGG_RECYCLE_MAX_INVOCATIONS`` -- cap on recycle-eligible (async)
+      invocations served, NOT the raw container generation (issue #177: the
+      runner's synchronous setup invoke warms a sandbox first, and counting
+      it made MAX_INVOCATIONS=1 deliver generation-2 workers while telemetry
+      read as if recycling had failed). Template default 1: recycle after
+      every async invocation, the cold-every-time posture; raise it for a
+      belt-and-suspenders cap over retention modes the RSS read misses --
+      and the only check that fires off-Linux, where RSS reads are None.
 
     Emits one CloudWatch-searchable line (``ZAGG_SELF_RECYCLE ...``) before
     exiting so dashboards can split intentional recycles from real crashes
-    (metric-filter note in docs/deployment/lambda.md).
+    (metric-filter note in docs/deployment/lambda.md). The line carries both
+    the async budget spent and the true container generation.
     """
+    global _ASYNC_INVOCATIONS_SERVED
+    _ASYNC_INVOCATIONS_SERVED += 1
     rss_limit = _recycle_limit("ZAGG_RECYCLE_RSS_MB")
     gen_limit = _recycle_limit("ZAGG_RECYCLE_MAX_INVOCATIONS")
     kib = _read_vmrss_kib()
     rss_mb = kib / 1024.0 if kib is not None else None
+    async_served = _ASYNC_INVOCATIONS_SERVED
     generation = _INVOCATIONS_SERVED
     if rss_limit > 0 and rss_mb is not None and rss_mb >= rss_limit:
         threshold = rss_limit
-    elif gen_limit > 0 and generation >= gen_limit:
+    elif gen_limit > 0 and async_served >= gen_limit:
         threshold = gen_limit
     else:
         return
     rss_repr = f"{rss_mb:.0f}" if rss_mb is not None else "n/a"
     logger.info(
-        f"ZAGG_SELF_RECYCLE rss_mb={rss_repr} generation={generation} threshold={threshold:g}"
+        f"ZAGG_SELF_RECYCLE rss_mb={rss_repr} async_served={async_served} "
+        f"generation={generation} threshold={threshold:g}"
     )
     _exit(0)
 
@@ -421,7 +437,8 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     # Count EVERY invocation toward the sandbox's generation (issue #171): a
     # setup/finalize/extract invoke warms the container just like a shard does,
     # so the next shard on this sandbox is genuinely generation N+1. Snapshot
-    # start RSS here, before any per-unit work inflates it.
+    # start RSS here, before any per-unit work inflates it. (The recycle
+    # budget is billed separately, per async invocation -- issue #177.)
     telemetry = _container_telemetry()
     mode = event.get("mode", "process")
     if mode == "setup":
--- a/deployment/aws/stand_up.sh
+++ b/deployment/aws/stand_up.sh
@@ -112,20 +112,26 @@ fi
 echo "zagg lambda artifacts: $MINOR ($ARCH), region $REGION"
 
 # Refuse to guess (issue #174): the resolved minor must actually be staged on
-# the distribution bucket. A HEAD on the layer key catches a phantom version
-# (e.g. one derived from post-tag main) here, with an actionable message,
-# instead of as a CloudFormation NoSuchKey mid-update.
+# the distribution bucket. A HEAD on both the layer and function keys catches
+# a phantom version (e.g. one derived from post-tag main) here, with an
+# actionable message, instead of as a CloudFormation NoSuchKey mid-update.
+# The aws stderr is surfaced so a 404 (minor not staged) reads differently
+# from a network/credentials failure of the check itself.
 SRC_BUCKET="$DIST_BUCKET"; SRC_REGION="$DIST_REGION"
 SRC_LAYER_KEY="$(dist_key "$MINOR" "$LAYER_ZIP")"; SRC_FUNC_KEY="$(dist_key "$MINOR" "$FUNC_ZIP")"
-if ! aws s3api head-object --bucket "$DIST_BUCKET" --key "$SRC_LAYER_KEY" --region "$DIST_REGION" --no-sign-request >/dev/null 2>&1; then
-    echo "ERROR: minor $MINOR is not staged: s3://$DIST_BUCKET/$SRC_LAYER_KEY does not exist."
-    STAGED="$(aws s3 cp "s3://$DIST_BUCKET/$(dist_root versions.json)" - --region "$DIST_REGION" --no-sign-request 2>/dev/null \
-        | python3 -c 'import json,sys; print(" ".join(json.load(sys.stdin).get("minors", [])))' 2>/dev/null || true)"
-    [ -n "$STAGED" ] && echo "       Staged minors on s3://$DIST_BUCKET: $STAGED"
-    echo "       Set LAMBDA_VERSION to a staged minor, or LAMBDA_VERSION=latest for the"
-    echo "       newest published one (versions.json)."
-    exit 1
-fi
+for KEY in "$SRC_LAYER_KEY" "$SRC_FUNC_KEY"; do
+    if ! HEAD_ERR="$(aws s3api head-object --bucket "$DIST_BUCKET" --key "$KEY" --region "$DIST_REGION" --no-sign-request 2>&1 >/dev/null)"; then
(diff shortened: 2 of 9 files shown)


deployment/aws/lambda_handler.py
```diff
@@ -239,30 +247,38 @@ def _maybe_self_recycle() -> None:
...
+      and the only check that fires off-Linux, where RSS reads are None.
 
     Emits one CloudWatch-searchable line (``ZAGG_SELF_RECYCLE ...``) before
     exiting so dashboards can split intentional recycles from real crashes
-    (metric-filter note in docs/deployment/lambda.md).
+    (metric-filter note in docs/deployment/lambda.md). The line carries both
+    the async budget spent and the true container generation.
     """
+    global _ASYNC_INVOCATIONS_SERVED
+    _ASYNC_INVOCATIONS_SERVED += 1
```

**should-fix (correctness edge):** the counter name says "async invocations served", but incrementing here — inside `_maybe_self_recycle()` — actually counts *successfully mirrored* async invocations, because the only call site is gated on `if mirrored:` (line ~475). An async invocation whose `result_url` mirror fails is served (full heavy work, full RSS growth) but never billed against the budget.

Pre-fix this case still counted: `_INVOCATIONS_SERVED` increments on every invocation, so a failed-mirror invocation advanced the generation and the next successful one recycled on schedule. Post-fix, with cap N, a container serves N + (failed mirrors) heavy invocations, and each failed mirror's RSS stays resident — precisely the flaky-S3 scenario the failed-mirror recycle-skip comment worries about. Issue #177's option (1) says count *async invocations*; a failed-mirror async invocation is one.

Suggest: increment on the async path in `lambda_handler` (inside `if event.get("result_url"):`, before the `mirrored` check) and keep `_maybe_self_recycle()` a pure check, preserving the recycle-skip on failed mirror. That also removes the side effect from a function whose name says "maybe" — a second call site added later would silently double-bill. Worth a companion test (mirror fails on invocation 1, succeeds on invocation 2, cap 2 → recycle fires on invocation 2).

---

deployment/aws/stand_up.sh
```diff
@@ -100,14 +111,34 @@ else
...
 
-# Pick the source that actually holds this minor: prefer the CORS bucket; if the
-# layer isn't there (an older minor), fall back to the source.coop mirror.
+# Refuse to guess (issue #174): the resolved minor must actually be staged on
+# the distribution bucket. A HEAD on the layer key catches a phantom version
+# (e.g. one derived from post-tag main) here, with an actionable message,
+# instead of as a CloudFormation NoSuchKey mid-update.
 SRC_BUCKET="$DIST_BUCKET"; SRC_REGION="$DIST_REGION"
 SRC_LAYER_KEY="$(dist_key "$MINOR" "$LAYER_ZIP")"; SRC_FUNC_KEY="$(dist_key "$MINOR" "$FUNC_ZIP")"
 if ! aws s3api head-object --bucket "$DIST_BUCKET" --key "$SRC_LAYER_KEY" --region "$DIST_REGION" --no-sign-request >/dev/null 2>&1; then
```
espg: 🤖 *from Claude (review)*

**should-fix:** only the layer key is HEAD-checked, but the header comment (lines 18–21) claims "the layer/function keys are verified to EXIST on the bucket before any stack call". A partially staged minor — layer present, function zip missing — passes this gate and dies as a CloudFormation `NoSuchKey` mid-update, the exact failure mode issue #174 targets (and `SRC_FUNC_KEY` is echoed in the confirm block as if verified). Issue #174's scope item (2) literally says "verify the layer key", so the *code* matches the issue — but then the header overclaims. Cheapest correct fix is a second `head-object` on `$SRC_FUNC_KEY` (a small loop over both keys keeps it terse); otherwise scope the header comment (and the echo) down to the layer key, matching `docs/deployment/standup.md`'s accurate "layer key is HEAD-checked" wording.

**nit (same line):** `>/dev/null 2>&1` conflates "key missing" with every other HEAD failure — no network, proxy/credential trouble, wrong `DIST_REGION` — and line 121 then asserts categorically "does not exist". A transient fault masquerading as "not staged" is the same class of misleading failure the issue set out to kill. Consider capturing stderr and printing it in the error block, or wording the message as "not staged (or not reachable)".

---

deployment/aws/stand_up.sh
```diff
@@ -100,14 +111,34 @@ else
...
+echo "  version:  $MINOR ($ARCH)"
+echo "  layer:    s3://$SRC_BUCKET/$SRC_LAYER_KEY"
+echo "  function: s3://$SRC_BUCKET/$SRC_FUNC_KEY"
+echo "  stack:    $STACK_NAME ($REGION)"
+if [ "$ASSUME_YES" != "true" ]; then
+    read -r -p "Proceed with deploy? [y/N] " REPLY
+    case "$REPLY" in
+        [Yy]|[Yy][Ee][Ss]) ;;
+        *) echo "Aborted (pass --yes to skip this prompt)."; exit 1 ;;
+    esac
```
espg, line 149: 🤖 *from Claude (review)*

**should-fix (minor):** under `set -euo pipefail`, `read` returns nonzero on EOF and kills the script *before* the case statement — so in the exact scenario this message targets (unattended run without `--yes`), "Aborted (pass --yes to skip this prompt)." never prints; the script dies silently with exit 1. Verified empirically:

```
$ ./stand_up.sh </dev/null   # (reduced repro)
exit=1, no output
```

`test_no_input_and_no_yes_aborts` passes because it asserts only the exit code and the absence of a `cloudformation` call, not the message. Fix is one token:

```bash
read -r -p "Proceed with deploy? [y/N] " REPLY || REPLY=""
```

which keeps the fail-safe default and lets the actionable message print. Strengthening the test to assert `"Aborted"`/`"--yes"` in stdout would pin it. (The PR body's "EOF on stdin ... aborts safely" is true on the safety half only.)

---

deployment/aws/lambda_handler.py
espg, line 1: 🤖 *from Claude (review)*

**nit (convention, standing note):** this module is now 1029 lines — past CLAUDE.md §4's ~1000-line threshold. It was already over at base (1012), so not caused here, but the convention asks that growth past the line be raised rather than continued silently; this PR adds 17 net lines without flagging it. Leaving this as a standing note: a follow-up decision on whether/how to split (or to bless the size) belongs on an issue.

---

espg (commented):
🤖 *from Claude (review)*

**Verdict:** solid small-fix pair — both issues' acceptance criteria are substantively met, tests are real (stubbed-CLI harness for the shell script, pre-fix-failing handler tests), commits are title-only, no lint suppressions, and the `deployment/aws/` touch is authorized by issue #174 plus the maintainer thread comment. I independently ran the branch's `tests/test_stand_up.py` + `tests/test_lambda_handler.py` (86/86 pass) and `ruff check`/`format --check` on the touched files (clean), and confirmed the `template.yaml` metric filter is a term match (`"ZAGG_SELF_RECYCLE"`), so the extra log field is safe. Findings below are foldable; nothing blocking.

**Findings (by severity):**

1. **should-fix (correctness edge)** — `deployment/aws/lambda_handler.py:264`: the recycle budget increments inside `_maybe_self_recycle()`, which only runs after a *successful* mirror — a failed-mirror async invocation is served but never billed, so with cap N a container can serve N + (failed mirrors) heavy invocations and keep their RSS. Pre-fix behavior counted these. Increment on the async path instead; add a failed-mirror test.
2. **should-fix** — `deployment/aws/stand_up.sh:120`: only the layer key is HEAD-checked, but the script header (lines 18–21) claims layer *and* function keys are verified; a partially staged minor still dies as CloudFormation `NoSuchKey`. HEAD both keys, or scope the header down.
3. **should-fix (minor)** — `deployment/aws/stand_up.sh:137`: on EOF, `read` under `set -e` exits before the "Aborted (pass --yes ...)" message prints (verified). `|| REPLY=""` fixes it; test only pins the exit code.
4. **nit** — `deployment/aws/stand_up.sh:120-121`: HEAD failure from network/credential/region trouble is reported categorically as "does not exist"; surface stderr or soften the wording.
5. **nit (convention)** — `lambda_handler.py` is now 1029 lines, past the ~1000-line threshold (1012 at base); growth should be raised, not continued silently.

**On the PR body's open questions:** the one-line `template.yaml` `RecycleMaxInvocations` description staleness is real and its text now contradicts the shipped semantics; whether that one-liner rides here is a scope call left standing for the maintainer. The hard cut for pre-0.14 source.coop-only minors matches the issue's "removing the source.coop defaults" and is properly surfaced in the body and in `docs/deployment/benchmark-cicd.md`. The flagged pre-existing lint drift on `main` reproduces and is correctly not "fixed" here.

---

espg:
🤖 *from Claude*

State: complete and green. Both issues implemented (one commit each), all four diff-scoped adversarial-review findings folded (`5da54de`), CI fully green (ruff, x86_64/arm64 builds, tests on 3.12 and 3.13), ruff bot posted nothing.

One process note: CLAUDE.md §2 has me flip a finished PR from draft → ready-for-review myself at exactly this point, but this routine's run configuration says drafts only / never mark ready — a genuine conflict, so per the stop-and-explain rule I'm leaving it in draft and flagging it here instead. Everything §2 requires for the flip is satisfied; it just needs your click (or a config change for future runs). Applying `waiting` — the standing questions in the PR body (module-size note, `template.yaml` description staleness, `publish_mirror.sh` retirement) are yours.

---

Issue #186, opened 2026-07-07
Sharded-path parallel CSR writes race under real CPU parallelism (fails 3/3 on 6 GB workers, largest shard)


espg:
🤖 *from Claude*

## Summary

The parallel ragged (CSR) writer fails non-deterministically on the **sharded** output path when the worker has real CPU parallelism: on a 6144 MB (~3.47 vCPU) test worker, the largest NEON o9 shard (5.8 M obs, 256 subgroups) failed its write in **3 of 3 independent runs** with `ragged (CSR) write failed for N of 256 subgroup(s) on the sharded path` — while reads and aggregation completed normally (~52 s). The same shard passes at 2048 MB (many runs, incl. the K-sweep) and passed once at 4096 MB.

## Evidence (2026-07-07 ~08:32 UTC, process-shard-test, zagg 0.18 + h5coro-hidefix 0.3.1)

- K=8 run: 1/256 failed, first failing shard_key **244151133**
- K=10 run: 1/256 failed, first failing **244151132**
- K=12 run: 2/256 failed, first failing **244151182**

Different subgroups each attempt → race/limit signature, not data. Only the largest shard (most subgroups × biggest CSR payloads) trips it. `granule_workers` (read-side, #183) is likely irrelevant except as noise — the failure is in the post-aggregation write stage (`_write_ragged_parallel`, `src/zagg/processing/write.py:411-439`, the #147 pool bounded by `_RAGGED_WRITE_CONCURRENCY`/`fd_safe_max_workers`).

## Hypothesis

Under `output.grid.sharded: true`, concurrent `write_ragged_to_zarr` calls for different subgroups are not fully independent at the store level (shared shard objects / metadata under the ShardingCodec layout?) — a read-modify-write race whose hit probability scales with true CPU parallelism. At 2 GB (~1.16 vCPU) the GIL serializes enough to hide it; at ~3.5 vCPU the writes genuinely overlap. Needs confirmation: the underlying exception **is** chained (`raise ... from first_exc`) but never surfaces — the worker envelope and CloudWatch both record only the summary string.

## Asks

1. **Observability first (small)**: log the chained `first_exc` (type + message + traceback) in the worker's write-failure path so the next reproduction tells us the actual S3/zarr error. Currently invisible.
2. Root-cause + fix the sharded-path concurrency (serialize per-shard-object writes, or a store-level lock, or verify the ShardingCodec write contract for concurrent writers).
[...]

Issue #184, opened 2026-07-07
Follow-ons to #183: dispatcher-side shard-aware granule_workers, and function-size variants gated on utilization


espg:
🤖 *from Claude*

Follow-up to PR #183 (granule-level read concurrency, Closes #180 / #179), filed at espg's direction. Two items were deliberately left out of that PR; this issue captures them so they don't live only in the PR's "Questions for review". Both are **gated on the fleet A/B at K∈{1,2,4,6}** (issue #180 phase 3, espg-side): the per-invocation `cpu_seconds` telemetry that PR #183 adds is what produces the utilization data these decisions key on.

## 1. Dispatcher-side shard-aware K (per-shard `granule_workers` override)

PR #183 ships `data_source.granule_workers` as a config-level knob only: one K for the whole run. The spec (issue #180 phase 2) sketched the finer version:

> Shard-aware K (free 80% of "shardmap-scoped workers"): the dispatcher can thread a per-shard `granule_workers` override into the cell event from the shardmap's granule counts — precedent: the `aoi_payload` per-cell extra in `runner.py`'s `_cell_work`.

The worker side already accepts event-injected values (the plumbing seam exists in PR #183); nothing sets it yet. The point: shardmaps record granule counts per shard, so the dispatcher can clamp K to `min(K_config, n_granules)` — a 5-granule shard shouldn't spin up K=6 threads — and could go further (scale K down on the densest shards where RSS headroom at 2 GB is the binding constraint, per issue #117).

**Scope sketch:** `runner.py` `_cell_work` threads `granule_workers` into the cell event alongside `aoi_payload`; simplest policy first (`min(K, n_granules)`), anything cleverer only if the A/B shows shard heterogeneity matters. Tests: dispatcher unit test on the event payload; worker already validates event values.

**Gate:** worth doing at all only if the A/B lands on a default K &gt; 1.

## 2. Function-size variants (different-size Lambdas) — separate, harder-gated decision

From issue #180's follow-on section, espg's criterion verbatim:

> Lambda bills GB-seconds → bigger workers (vCPU ∝ memory, ~1/1769 MB, max 6) are cost-neutral whenever wall drops proportionally. Scale up only once K-in-flight work demonstrably saturates 2 vCPUs ("I don't want a 16-core lambda unless we hammer the CPUs" — espg). If shard heterogeneity warrants, function-size variants (2/4/7 GB) + dispatch routing by granule count.

[...]

PR #187, opened 2026-07-07
Inline selection-read heuristic + granule_workers default 4

### LINKED ISSUE #185: 0.18 inline regression: #179's selection routing costs inline ~41% at default granule_workers=1
🤖 *from Claude*

## Summary

0.18's #179 change (route the remaining planned-read seams — the segment-rate *selection* reads — through the index backend's `read_fn`) regressed the **inline** backend by ~41% at the shipped default `granule_workers: 1`: worker median 323 → 455 s, read median ~146 → ~276 s on the o9 NEON AOI (isolated runs, prod 0.18 fleet, codec-independent read comparison). Full evidence: the K-sweep tables on [PR #183](https://github.com/englacial/zagg/pull/183#issuecomment-4901386362).

## Mechanism

For `sidecar`, #179 is a pure win (selection datasets are manifest-covered: no metadata walks, coalesced compiled fetches — its K=1 improved −25%). For `inline`, routing selection through the compiled seam means **building per-granule chunk maps (B-tree walks) for the segment-rate datasets** and fetching them through the compiled path — datasets that h5coro 1.0.5 reads faster via its 4 MiB cache lines (the #170 matrix measured exactly this trade). The cost is per (granule × group × selection dataset) and lands squarely on the shipped default path (`index: None` → inline, K=1).

`granule_workers ≥ 2` hides the cost in worker totals (latency-bound work the #180 pool overlaps: K=4 inline is 296.9 s, nominally ahead of 0.17) — but the ~2× read gap vs sidecar persists at every K, and K=1 is the default.

## Fix menu (pick one or combine)

1. **Conditional routing**: only send selection reads through `read_fn` when the backend can serve them without building addressing state — i.e., `sidecar` (manifest-covered) yes, `inline` no (falls back to the h5coro selection read it used through 0.17). Smallest change, restores 0.17 inline exactly, keeps sidecar's win.
2. **Inline-side heuristic**: inline's `read_fn` serves selection datasets via `h5obj.readDatasets` directly (skip chunk-map build for datasets it hasn't already mapped for data reads). Similar effect, contained in inline.
3. **Default `granule_workers` bump** (e.g. 2–4): masks the regression and banks the #180 win for all users — worth doing on the PR #183 evidence regardless, but doesn't remove the underlying 2× read gap, so pair with 1 or 2.

Fallback context measured on the same fleet: wholesale `hierarchical` + sharded is 164.5 s at K=1 / 118.0 s at K=4 — only ~16%/51% behind sidecar — so the h5coro selection path being fast is not hypothetical.

Refs: #179, #180, PR #183, #170 (the 1.0.5-vs-compiled fetch-pattern evidence).

espg:
Closes #185. Refs #184 (the per-cell clamp below is item 1's simple `min()` case — the rest of #184 stays open). Refs #186 (ask (1) only — root-cause of the sharded write race stays on #186).

Implements option (2) from the issue thread — the **inline-internal heuristic** — plus the default `granule_workers` bump, per the approval on [issue #185](https://github.com/englacial/zagg/issues/185#issuecomment-4901692895) ("yup, sounds like (2) it is (plus the default granule workers bump)"). The mechanism and evidence are the [K-sweep tables on PR #183](https://github.com/englacial/zagg/pull/183#issuecomment-4901386362): #179's selection routing costs inline ~41% at K=1 (B-tree chunk-map walks for the segment-rate selection datasets) while being a pure win for sidecar.

## Phases

- [x] Phase 1 — inline selection-read heuristic (`src/zagg/index/inline.py`)
- [x] Phase 2 — `granule_workers` default 1 → 4 (`src/zagg/processing/worker.py`, README)
- [x] Phase 3 — dispatcher-side per-cell `min(K, n_granules)` clamp (`src/zagg/runner.py`)
- [x] Phase 4 — surface the chained CSR write failure (issue #186 ask (1), `src/zagg/processing/write.py`)
- [x] Fold adversarial-review findings ([review](https://github.com/englacial/zagg/pull/187#pullrequestreview-4643501393)): clamp on resolved-URL count; bounded per-failure write summary
- [x] Fold second-round review findings ([review](https://github.com/englacial/zagg/pull/187#pullrequestreview-4644501334)): header-first failure roll call + single-failure guard; named `_RAGGED_FAILURE_LOG_CAP`; `fd_safe_max_workers` pins in the fanout failure tests; lockstep comment on the local clamp resolve

## What / approach

**Phase 1 — the heuristic, contained in `InlineIndex._chunk_aligned_read_fn`.** `read.py`'s call sites are untouched; the seam stays uniform. The discriminator is the call shape: on the planned route, data reads arrive as hyperslices (`execute_read_plan`) and selection reads (coarse coords + link arrays, cross-level filter flags, segment-level broadcasts) arrive as full-dataset reads (`hyperslice is None`). The change is one early exit:

```python
if selection_direct and hyperslice is None:   # selection_direct = planned and not write_back
    return h5obj.readDatasets([path])[path]
```

- **Default path** (`planned=True`, `write_back` off — the shipped `index: None` default where the regression lives): unmapped full-dataset reads are served whole through h5coro's `readDatasets` (its 4 MiB cache-line path, the fast side of the #170 matrix for these datasets) and build **no chunk map** — restoring the 0.17 selection behavior.
- **`write_back` ON**: unchanged byte-for-byte. The walk's product *is* the manifest payload (`_pending_for`/`finish_granule`), so selection datasets stay mapped and served compiled — freshly built manifests keep covering the selection datasets, preserving sidecar's −25% K=1 win.
- **`planned=False`** (flat sources): unchanged in all modes — the compiled full-dataset decode there is the point of issue #170.
- A path already mapped this `read_group` (e.g. both a data and a selection read) keeps the compiled route — the map is already paid for. Full `readDatasets` reads don't need the PR #152 start-edge workaround (that's a ranged-read quirk), so no boundary handling is added.

Call-shape note beyond the issue-thread sketch: the `plan.full_read` selectivity fallback (`_planned_read_group` → `_read_group_full` with the planned `read_fn`, routed through the seam since #179) also issues full-dataset reads — the base-rate *coordinate* reads. The heuristic serves those direct too, which matches the pre-#179 fallback (it bypassed the seam entirely, per `inline.py`'s own docstring); the fallback's ranged data reads stay compiled. Pinned by `test_full_read_fallback_data_reads_stay_compiled`.

A/B watch-list (review finding 2, no code change): direct `readDatasets` reads retain their 4 MiB cache lines until granule close. On the `full_read` fallback that can be ~245 MB/beam of base-rate volume, × K=4 granules in flight at the new default — no fleet datapoint yet; watch as an issue #117 memory axis in the next sweep.

**Phase 2 — default `granule_workers: 4`.** Directed on the PR #183 K-sweep evidence (K=4 inline 296.9 s vs 0.17's 323 s median, o9 NEON AOI) via the same [approval comment](https://github.com/englacial/zagg/issues/185#issuecomment-4901692895). Docs updated honestly: output remains byte-identical to serial by construction of the ordered fold — the *default* just no longer equals serial execution (`granule_workers: 1` restores the serial loop, which still constructs no pool).

**Phase 3 — dispatcher-side clamp (Refs #184, item 1's simple case).** `_clamped_data_source(data_source, n_granules)` in `runner.py` returns a shallow-copied `data_source` carrying `min(K, n_granules)` only when that lowers K, else `None`. Both `_cell_work` sites use it: `_run_local` via `dataclasses.replace(config, data_source=...)`, `_run_lambda` via a per-cell `{**config_dict, "data_source": ...}`. Unclamped cells pass the shared config through untouched, keeping their event payloads byte-identical. Per the review fold, the clamp counts the **resolved** URLs (`_resolve_urls` output — what the worker actually reads), not raw records, so a partially-resolvable cell no longer under-clamps. **Deviation from the plan sketch:** there is no dedicated worker event key for this — the worker reads the width from the event's `config` (`_granule_workers(event["config"]["data_source"])`, PR #183), so the clamp rides the per-cell config copy rather than a new top-level key (no `lambda_handler` change needed).

**Phase 4 — CSR write-failure observability (issue #186 ask (1) only).** `_write_ragged_fanout`'s failure path now (a) logs the first failing subgroup's chained exception with its **full traceback** (`logger.error(..., exc_info=first_exc)`), (b) when more than one subgroup failed, follows that header with a bounded one-line-per-failure roll call (`subgroup_key: ExcType: message`, capped at `_RAGGED_FAILURE_LOG_CAP = 20` lines) so the race signature across subgroups is visible, and (c) carries `TypeName: message` inside the raised summary string, so the worker envelope (`lambda_handler`'s `Failed to write zarr: {e}`) shows the actual S3/zarr error even without log access. Envelope schema unchanged; no behavior change on the write path itself — the root-cause/race fix stays on #186.

## How tested

- **Phase 1** (`tests/test_index.py`): `test_default_selection_reads_skip_chunk_maps` spies the map-build seam (`build_chunk_map`) and a `readDatasets` recorder — selection paths are never walked, are served as whole-dataset `readDatasets` calls, data paths still map + compile, rows identical to the hierarchical baseline. `test_write_back_selection_reads_stay_compiled` pins the ON mode (`readDatasets` never touched, pending maps cover the selection datasets, rows identical). `test_manifest_byte_identical_to_direct_chunk_maps` pins the sidecar-build product: the persisted parquet is byte-identical to one assembled directly from `build_chunk_map` over the deterministic coverage. Segment-variable, cross-level-filter, full-read-fallback, and per-dataset-degrade tests updated to the new mode split.
- **Phase 2** (`tests/test_processing.py`): default asserted as 4; `test_k1_never_touches_the_pool` keeps the serial guarantee on the explicit escape hatch; `test_default_output_identical_to_serial` pins default-vs-serial frame equality under out-of-order completions.
- **Phase 3** (`tests/test_runner.py`): `TestClampedDataSource` unit-tests the `min()` policy (incl. default-K and the 0-granule floor); dispatcher tests for both backends assert per-shard event/config values (2-granule cell → 2, 1-granule cell → 1, 6-granule cell → untouched, and a 6-record cell with only 3 s3-resolvable hrefs → 3) and that unclamped local cells receive the shared config object as-is.
- **Phase 4** (`tests/test_processing.py`): the fanout failure test asserts the raised summary matches `first failing shard_key 3: OSError: disk full`, that the log record carries `exc_info` whose formatted traceback names the raising frame, and that a single failure logs no roll call (the header alone); `test_fanout_logs_each_failure_bounded` pins the multi-failure roll call (header first, then 25 failures → `_RAGGED_FAILURE_LOG_CAP` per-key lines + "and 5 more", exactly one traceback record). Both tests pin `fd_safe_max_workers` so a low-ulimit host can't silently take the serial branch.
- Full local gate: `pytest -q` green — 1516 passed, 26 skipped (3.12, `uv sync --extra test`) on the first-fold state; the second fold re-ran `tests/test_index.py tests/test_processing.py tests/test_runner.py` (425 passed) plus the two changed modules in full (352 passed); `ruff check` + `ruff format --check` clean on every touched file.

## Questions for review

1. **Concurrent `readDatasets` on one `h5obj`:** selection reads now reach `readDatasets` through the `read_workers` pool (`_read_paths_pooled`), so up to four single-path calls can be in flight on one granule handle. **Answered safe** by the [adversarial review](https://github.com/englacial/zagg/pull/187#pullrequestreview-4643501393)'s h5coro source inspection: `h5promise.py` already fans one thread per dataset over the shared resource, cache-line access is locked per line, and the S3 client is documented thread-safe.
2. **Pre-existing lint drift (flagged, not fixed, per the §4 convention):** on clean `main`, `ruff check src tests` fails N818 (`UnknownCapability`, `src/zagg/registry.py:64`) and `ruff format --check` would reformat 7 untouched files (`src/zagg/__init__.py`, `src/zagg/__main__.py`, `src/zagg/catalog/sources.py`, `src/zagg/viz/crs.py`, `tests/test_catalog.py`, `tests/test_extract.py`, `tests/test_integration.py`). The PR lint bot selects only `E,F,W,I`, so PRs won't see these.
3. The K=4 default reaches **direct `process_shard` callers** (no dispatcher) unclamped — a 1-granule hand-rolled call now constructs a 4-wide pool (threads spawn lazily, so the cost is one idle pool object). Worth a worker-side `min(K, len(granule_urls))` too, or keep the clamp dispatcher-only as agreed on #184?

### DIFF
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -529,7 +529,27 @@ class InlineIndex(VirtualIndex):
         ``finish_granule`` to persist. Each dataset gets
         its own single-dataset in-memory ``Index`` (~ms), so a spec hidefix
         rejects degrades that dataset alone.
+
+        Selection-shaped reads skip the walk (issue #185): on the planned
+        route a full-dataset read (``hyperslice is None``) is a segment-rate
+        selection read — data reads arrive as hyperslices — and h5coro's
+        ``readDatasets`` (4 MiB cache lines) serves those faster than the
+        compiled route once the B-tree walk is priced in (the 0.18 inline
+        regression, ~41% at ``granule_workers: 1``; the #170 matrix measured
+        the same trade). So with ``write_back`` off, an unmapped full-dataset
+        read on the planned route goes straight through ``readDatasets`` and
+        builds no chunk map. A path already mapped (e.g. one that is both a
+        data and a selection read) keeps the compiled route — the map is paid
+        for. With ``write_back`` ON nothing changes: the walk's product IS
+        the manifest payload, so selection datasets stay mapped (and served
+        compiled) to keep manifests covering them for sidecar consumers. The
+        full-read route (``planned=False``) is untouched — its compiled
+        full-dataset decode is the point of issue #170.
         """
+        # Issue #185: with write_back off there is no manifest to enrich, so
+        # the planned route's full-dataset (selection) reads bypass the
+        # chunk-map build entirely -- see read_fn below.
+        selection_direct = planned and not self.write_back
         maps: dict[str, ChunkMap] = self._pending_for(h5obj) if self.write_back else {}
         # One single-dataset Index per path (~ms each): a dataset whose spec
         # hidefix rejects (nonzero filter_mask, non-tiling chunk table) then
@@ -579,6 +599,13 @@ class InlineIndex(VirtualIndex):
         def read_fn(path, hyperslice=None):
             cm = maps.get(path)
             if cm is None and path not in direct:
+                if selection_direct and hyperslice is None:
+                    # Selection-shaped read (issue #185): serve through
+                    # h5coro's cache-line path, build no chunk map. Full
+                    # readDatasets reads never trip the PR #152 start-edge
+                    # off-by-one (that is a ranged-read quirk), so no
+                    # boundary workaround is needed here.
+                    return h5obj.readDatasets([path])[path]
                 try:
                     cm = maps[path] = build_chunk_map(h5obj, path)
                 except Exception:
--- a/src/zagg/processing/write.py
+++ b/src/zagg/processing/write.py
@@ -31,6 +31,11 @@ from zagg.grids.morton import is_morton_array, is_morton_arrow, morton_to_arrow,
 # path's ``async.concurrency`` budget (issue #108).
 _RAGGED_WRITE_CONCURRENCY = 128
 
+# Cap on the per-failure roll call logged when sharded CSR writes fail
+# (issue #186): enough to show the race signature across subgroups without
+# flooding CloudWatch when many of the K writes fail at once.
+_RAGGED_FAILURE_LOG_CAP = 20
+
 logger = logging.getLogger(__name__)
 
 
@@ -439,18 +444,20 @@ def _write_ragged_fanout(ragged_writes: list, store: Store, *, grid) -> None:
         # the operator — the worker envelope and CloudWatch both record only
         # this summary string — so log the first failure's full traceback
         # here, and carry its type + message in the summary text too. A
-        # bounded one-line-per-failure roll call makes the race signature
-        # ACROSS subgroups visible (which keys, same or different errors)
-        # without flooding the log when many of the K writes fail at once.
-        for key, exc in errors[:20]:
-            logger.error(f"  ragged (CSR) subgroup {key}: {type(exc).__name__}: {exc}")
-        if len(errors) > 20:
-            logger.error(f"  ... and {len(errors) - 20} more subgroup write failure(s)")
+        # bounded one-line-per-failure roll call (after the header, so the
+        # indented lines read as its continuation) makes the race signature
+        # ACROSS subgroups visible (which keys, same or different errors).
         logger.error(
             f"ragged (CSR) subgroup write failed (shard_key {first_key}): "
             f"{type(first_exc).__name__}: {first_exc}",
             exc_info=first_exc,
         )
+        if len(errors) > 1:
+            for key, exc in errors[:_RAGGED_FAILURE_LOG_CAP]:
+                logger.error(f"  ragged (CSR) subgroup {key}: {type(exc).__name__}: {exc}")
+            if len(errors) > _RAGGED_FAILURE_LOG_CAP:
+                n_more = len(errors) - _RAGGED_FAILURE_LOG_CAP
+                logger.error(f"  ... and {n_more} more subgroup write failure(s)")
         raise RuntimeError(
             f"ragged (CSR) write failed for {len(errors)} of {len(ragged_writes)} "
             f"subgroup(s) on the sharded path (first failing shard_key {first_key}: "
--- a/src/zagg/processing/write.py
+++ b/src/zagg/processing/write.py
@@ -7,6 +7,7 @@ Assembles the per-shard output carrier and writes it to the Zarr template
 stays acyclic.
 """
 
+import logging
 from concurrent.futures import ThreadPoolExecutor, as_completed
 
 import numpy as np
@@ -29,6 +30,8 @@ from zagg.grids.morton import is_morton_array, is_morton_arrow, morton_to_arrow,
 # a serial loop dominated a t-digest shard's write time. 128 mirrors the dense
 # path's ``async.concurrency`` budget (issue #108).
 _RAGGED_WRITE_CONCURRENCY = 128
+
+logger = logging.getLogger(__name__)
 
 
 def _arrow_column(block: np.ndarray, sig: dict):
@@ -432,9 +435,19 @@ def _write_ragged_fanout(ragged_writes: list, store: Store, *, grid) -> None:
                 errors.append((futures[fut], exc))
     if errors:
         first_key, first_exc = errors[0]
+        # Observability (issue #186 ask 1): the chained cause never reaches
+        # the operator — the worker envelope and CloudWatch both record only
+        # this summary string — so log the first failure's full traceback
+        # here, and carry its type + message in the summary text too.
+        logger.error(
+            f"ragged (CSR) subgroup write failed (shard_key {first_key}): "
+            f"{type(first_exc).__name__}: {first_exc}",
+            exc_info=first_exc,
+        )
         raise RuntimeError(
             f"ragged (CSR) write failed for {len(errors)} of {len(ragged_writes)} "
-            f"subgroup(s) on the sharded path (first failing shard_key {first_key})"
+            f"subgroup(s) on the sharded path (first failing shard_key {first_key}: "
+            f"{type(first_exc).__name__}: {first_exc})"
         ) from first_exc
 
 
--- a/src/zagg/processing/write.py
+++ b/src/zagg/processing/write.py
@@ -438,7 +438,14 @@ def _write_ragged_fanout(ragged_writes: list, store: Store, *, grid) -> None:
         # Observability (issue #186 ask 1): the chained cause never reaches
         # the operator — the worker envelope and CloudWatch both record only
         # this summary string — so log the first failure's full traceback
-        # here, and carry its type + message in the summary text too.
+        # here, and carry its type + message in the summary text too. A
+        # bounded one-line-per-failure roll call makes the race signature
+        # ACROSS subgroups visible (which keys, same or different errors)
+        # without flooding the log when many of the K writes fail at once.
+        for key, exc in errors[:20]:
+            logger.error(f"  ragged (CSR) subgroup {key}: {type(exc).__name__}: {exc}")
+        if len(errors) > 20:
+            logger.error(f"  ... and {len(errors) - 20} more subgroup write failure(s)")
         logger.error(
             f"ragged (CSR) subgroup write failed (shard_key {first_key}): "
             f"{type(first_exc).__name__}: {first_exc}",
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -1151,8 +1151,11 @@ def _run_local(
             extra["aoi_payload"] = aoi_by_shard.get(int(shard_key))
         # Per-cell granule_workers clamp (issue #184): min(K, n_granules), so
         # a small cell doesn't spin idle reader threads; unclamped cells pass
-        # the shared config through untouched.
-        ds = _clamped_data_source(config.data_source, len(records))
+        # the shared config through untouched. Count the RESOLVED urls — what
+        # the worker actually reads — not the raw records: _resolve_urls
+        # drops href-less records, so len(records) would under-clamp a
+        # partially-resolvable cell (review finding, PR #187).
+        ds = _clamped_data_source(config.data_source, len(_resolve_urls(records, driver)))
         cell_config = replace(config, data_source=ds) if ds is not None else config
         try:
             meta = _process_and_write(
@@ -1398,7 +1401,11 @@ def _run_lambda(
         # Per-cell granule_workers clamp (issue #184): the worker reads the
         # width from the event's config, so the clamp rides a per-cell copy
         # of it; unclamped cells send the shared config_dict byte-identical.
-        ds = _clamped_data_source(config.data_source, len(records))
+        # Clamp on the RESOLVED url count — what the worker actually reads —
+        # since _resolve_urls drops href-less records (review finding,
+        # PR #187).
+        granule_urls = _resolve_urls(records, "s3")
+        ds = _clamped_data_source(config.data_source, len(granule_urls))
         cell_config_dict = {**config_dict, "data_source": ds} if ds is not None else config_dict
         return _invoke_lambda_cell(
             state["lambda_client"],
@@ -1406,7 +1413,7 @@ def _run_lambda(
             int(shard_key),
             parent_order,
             child_order,
-            _resolve_urls(records, "s3"),
+            granule_urls,
             store_path,
             s3_creds,
             function_name=function_name,
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -1154,7 +1154,9 @@ def _run_local(
         # the shared config through untouched. Count the RESOLVED urls — what
         # the worker actually reads — not the raw records: _resolve_urls
         # drops href-less records, so len(records) would under-clamp a
-        # partially-resolvable cell (review finding, PR #187).
+        # partially-resolvable cell (review finding, PR #187). This resolve
+        # must stay in lockstep with _process_and_write's own
+        # _resolve_urls(records, driver) — same inputs, so same count.
         ds = _clamped_data_source(config.data_source, len(_resolve_urls(records, driver)))
         cell_config = replace(config, data_source=ds) if ds is not None else config
         try:
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -19,6 +19,7 @@ import time
 import uuid
 import warnings
 from concurrent.futures import ThreadPoolExecutor
+from dataclasses import replace
 
 from zarr import consolidate_metadata
 
@@ -964,6 +965,26 @@ def _resolve_urls(records: list, driver: str | None) -> list[str]:
     return [r[key] for r in records if r.get(key)]
 
 
+def _clamped_data_source(data_source: dict, n_granules: int) -> dict | None:
+    """Per-cell ``granule_workers`` clamp: ``min(K, n_granules)`` (issue #184).
+
+    The shardmap gives the dispatcher each cell's granule count up front, so a
+    2-granule cell never asks the worker for a 4-wide pool (the issue #185
+    default). Returns a shallow-copied ``data_source`` carrying the clamped
+    width when it is below the configured/default K, else ``None`` — the
+    caller then passes the shared config through untouched, keeping unclamped
+    cell payloads byte-identical. Just the simple ``min()`` policy; the
+    worker still resolves the value through its ``_granule_workers`` guard.
+    """
+    from zagg.processing.worker import _granule_workers
+
+    k = _granule_workers(data_source)
+    clamped = min(k, max(int(n_granules), 1))
+    if clamped == k:
+        return None
+    return {**data_source, "granule_workers": clamped}
+
+
 def _check_signature(grid, catalog_data: dict) -> None:
     """Refuse a ShardMap built for a different *spatial* grid than the run config.
 
@@ -1128,6 +1149,11 @@ def _run_local(
         extra = {}
         if aoi_by_shard:
             extra["aoi_payload"] = aoi_by_shard.get(int(shard_key))
+        # Per-cell granule_workers clamp (issue #184): min(K, n_granules), so
+        # a small cell doesn't spin idle reader threads; unclamped cells pass
+        # the shared config through untouched.
+        ds = _clamped_data_source(config.data_source, len(records))
+        cell_config = replace(config, data_source=ds) if ds is not None else config
         try:
             meta = _process_and_write(
                 shard_key,
@@ -1136,7 +1162,7 @@ def _run_local(
                 grid,
                 s3_creds,
                 zarr_store,
-                config,
+                cell_config,
                 driver=driver,
                 handoff=handoff,
                 **extra,
@@ -1369,6 +1395,11 @@ def _run_lambda(
                 result_box, result_prefix, output_creds_event, region, key
             )
             extra["poll_timeout_s"] = state["function_timeout_s"] + _ASYNC_POLL_MARGIN_S
+        # Per-cell granule_workers clamp (issue #184): the worker reads the
+        # width from the event's config, so the clamp rides a per-cell copy
+        # of it; unclamped cells send the shared config_dict byte-identical.
+        ds = _clamped_data_source(config.data_source, len(records))
+        cell_config_dict = {**config_dict, "data_source": ds} if ds is not None else config_dict
         return _invoke_lambda_cell(
             state["lambda_client"],
             grid.block_index(int(shard_key)),
@@ -1379,7 +1410,7 @@ def _run_lambda(
             store_path,
             s3_creds,
             function_name=function_name,
-            config_dict=config_dict,
+            config_dict=cell_config_dict,
             output_creds_event=output_creds_event,
             max_retries=max_retries,
             max_workers=state["workers"],
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -1291,6 +1291,35 @@ class TestRaggedWriteFanout:
         assert "OSError: disk full" in tb_text
         assert "in boom" in tb_text
 
+    def test_fanout_logs_each_failure_bounded(self, monkeypatch, caplog):
+        """Multiple subgroup failures each get a one-line ``key: Type: msg``
+        summary (issue #186: the race signature ACROSS subgroups), capped at
+        20 lines, while exactly one record — the first completed failure —
+        carries the full traceback."""
+        import logging
+
+        import zagg.processing.write as wmod
+        from zagg.processing.write import _write_ragged_fanout
+
+        grid, writes = self._grid_and_writes(25)
+
+        def boom(ragged, store, *, grid, shard_key):
+            raise OSError(f"reset {shard_key}")
+
+        monkeypatch.setattr(wmod, "write_ragged_to_zarr", boom)
+        with caplog.at_level(logging.ERROR, logger="zagg.processing.write"):
+            with pytest.raises(RuntimeError, match=r"failed for 25 of 25"):
+                _write_ragged_fanout([(r, k) for r, k in writes], MemoryStore(), grid=grid)
+        lines = [r.message for r in caplog.records]
+        roll_call = [m for m in lines if m.startswith("  ragged (CSR) subgroup ")]
+        assert len(roll_call) == 20  # bounded roll call
+        assert any("and 5 more subgroup write failure(s)" in m for m in lines)
+        # Each roll-call line names its own subgroup key and its own error.
+        for m in roll_call:
+            key = m.split("subgroup ")[1].split(":")[0]
+            assert f"reset {key}" in m
+        assert sum(1 for r in caplog.records if r.exc_info) == 1
+
     def test_fanout_empty_is_noop(self):
         """No ragged writes -> nothing written, no pool spun up."""
         from zagg.processing.write import _write_ragged_fanout
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -1257,8 +1257,15 @@ class TestRaggedWriteFanout:
         _write_ragged_fanout([(r, k) for r, k in writes], MemoryStore(), grid=grid)
         assert "workers" not in recorded
 
-    def test_fanout_surfaces_write_failure(self, monkeypatch):
-        """A failure in any subgroup write is re-raised (not silently swallowed)."""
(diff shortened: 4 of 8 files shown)


src/zagg/index/inline.py
```diff
@@ -579,6 +599,13 @@ def _h5coro_read(path, hyperslice, cm):
         def read_fn(path, hyperslice=None):
             cm = maps.get(path)
             if cm is None and path not in direct:
+                if selection_direct and hyperslice is None:
+                    # Selection-shaped read (issue #185): serve through
+                    # h5coro's cache-line path, build no chunk map. Full
+                    # readDatasets reads never trip the PR #152 start-edge
+                    # off-by-one (that is a ranged-read quirk), so no
+                    # boundary workaround is needed here.
+                    return h5obj.readDatasets([path])[path]
```
espg, line 608: 🤖 *from Claude (review)*

**Answering the PR's Question 1 concretely (concurrent `readDatasets` on one `h5obj`): no new hazard class — no serialization needed.** Traced through h5coro 1.0.5's source: a single batched `readDatasets([p1..pK])` call already fans **one thread per dataset** over the shared resource object (`h5promise.py:79-80`, `H5Promise.__init__` builds a `ThreadPoolExecutor(max_workers=len(datasetTable))` and each `datasetThread` runs its own `H5Dataset` traversal against the same `resourceObject`). So K pooled single-path calls from `_read_paths_pooled` are structurally the same concurrency the pre-#179 batched selection read produced internally, just packaged as K one-thread promises instead of one K-thread promise. The shared state is guarded the same way in both shapes: the byte cache uses per-cache-line double-checked locking (`h5coro.py:126` `self.cache_locks = defaultdict(threading.Lock)`; `ioRequest`'s check-lock-recheck), `metadataTable`/`pathAddresses` writes are GIL-atomic idempotent dict inserts, and the S3 driver's own comment pins the client contract (`s3driver.py:49-50`: "Thread-safe client: one per instance, reused across all threads"). The only deltas are cosmetic: K short-lived executors instead of one, and concurrent sibling traversals may redundantly walk the group's object headers before `pathAddresses` warms — which the batched call's internal threads do too. I'd resolve Question 1 as "safe as-is."

One small related nit, not worth blocking: because the direct path never records the path in `maps`/`direct`, a selection path read twice in one `read_group` (e.g. the link arrays read in `_planned_read_group` and again in `_read_segment_broadcasts` when segment variables are declared) issues two full `readDatasets` decodes. The second hits the warmed byte cache so there are no repeat GETs, and the pre-#185 compiled route re-fetched on repeat reads anyway (`caching=False`), so this is at worst decode-time parity with an extra decode — fine to leave.

---

src/zagg/index/inline.py
```diff
@@ -529,7 +529,27 @@ def _chunk_aligned_read_fn(self, h5obj, *, planned=True):
...
+        for. With ``write_back`` ON nothing changes: the walk's product IS
+        the manifest payload, so selection datasets stay mapped (and served
+        compiled) to keep manifests covering them for sidecar consumers. The
+        full-read route (``planned=False``) is untouched — its compiled
+        full-dataset decode is the point of issue #170.
         """
+        # Issue #185: with write_back off there is no manifest to enrich, so
+        # the planned route's full-dataset (selection) reads bypass the
+        # chunk-map build entirely -- see read_fn below.
+        selection_direct = planned and not self.write_back
```
espg, line 552: 🤖 *from Claude (review)*

**RSS observation for the fleet A/B (should-watch, not blocking): the direct path swaps freed-after-decode buffers for cache lines retained until granule close — and the K=4 default multiplies that by in-flight granules.** The compiled route fetched with `ioRequest(..., caching=False)` (buffers dropped after `read_from_buffers`); `readDatasets` populates the h5obj's 4 MiB cache lines, which stay resident until the worker's per-granule `close()` (the issue #66 handling in `worker.py`). For the segment-rate selection datasets that's small and clearly worth it. But the heuristic also serves the `plan.full_read` fallback's **base-rate coordinate full reads** direct — up to ~245 MB/beam per `_planned_read_group`'s own docstring — so a fallback-heavy dense shard now retains those bytes in cache for the rest of the granule, times up to K=4 granules in flight under the new phase-2 default. 0.17 had the same cached reads but at K=1; the K-sweep numbers on PR #183 were measured before this heuristic existed, so the (cached fallback reads x K=4) combination has no fleet datapoint yet. This lands exactly on the 2 GB marginality axis of issue #117. Suggest adding it to "Questions for review" / the A/B watch list (RSS on dense shards with `full_read_threshold` trips) rather than changing code now.

---

src/zagg/runner.py
```diff
@@ -964,6 +965,26 @@ def _resolve_urls(records: list, driver: str | None) -> list[str]:
...
+    default). Returns a shallow-copied ``data_source`` carrying the clamped
+    width when it is below the configured/default K, else ``None`` — the
+    caller then passes the shared config through untouched, keeping unclamped
+    cell payloads byte-identical. Just the simple ``min()`` policy; the
+    worker still resolves the value through its ``_granule_workers`` guard.
+    """
+    from zagg.processing.worker import _granule_workers
+
+    k = _granule_workers(data_source)
+    clamped = min(k, max(int(n_granules), 1))
```
espg, line 982: 🤖 *from Claude (review)*

**Should-fix (minor): the clamp counts records, but the worker receives resolved URLs.** Both `_cell_work` sites feed `len(records)` into the clamp, while the payload the worker actually loops over is `_resolve_urls(records, driver)` (`runner.py:958-965`), which **drops records missing the driver's href key** (`if r.get(key)`). A cell with 6 records of which only 2 carry an `s3` href is not clamped (6 >= K=4), so the worker builds a 4-wide pool for 2 granules — precisely the idle-pool case the clamp exists to prevent. Severity is low (an idle `ThreadPoolExecutor` spawns threads lazily), but the fix is cheap: at the lambda site `_resolve_urls(records, "s3")` is already computed one expression below — hoist it and pass its `len`; the local site can resolve the same way. Alternatively, this folds naturally into the PR's own Question 3 (a worker-side `min(K, len(granule_urls))`), which would make the dispatcher-side count precision moot and also cover direct `process_shard` callers — one answer can settle both.

---

src/zagg/processing/write.py
```diff
@@ -432,9 +435,19 @@ def _write_ragged_fanout(ragged_writes: list, store: Store, *, grid) -> None:
                 errors.append((futures[fut], exc))
     if errors:
         first_key, first_exc = errors[0]
+        # Observability (issue #186 ask 1): the chained cause never reaches
+        # the operator — the worker envelope and CloudWatch both record only
+        # this summary string — so log the first failure's full traceback
+        # here, and carry its type + message in the summary text too.
+        logger.error(
+            f"ragged (CSR) subgroup write failed (shard_key {first_key}): "
```
espg, line 451: 🤖 *from Claude (review)*

**Nit (take or leave): only the first failure's cause is surfaced, and "first" is `as_completed` order.** Issue #186's three reproductions each failed 1-2 of 256 subgroups, and for a race-signature bug the *distribution* of causes is diagnostic (all the same exception type/message, or mixed?). Since `errors` already holds every `(key, exc)`, a one-line-per-failure `type(exc).__name__: exc` summary (full traceback still only for the first, to stay clear of the issue #175 metric-filter concern the neighboring `finish_granule` handler documents) would cost ~3 lines and make the next 6 GB reproduction answer the "same race everywhere?" question in one shot. Also note `errors[0]` is completion-ordered, so "first failing shard_key" is nondeterministic across runs — pre-existing wording, but now that the summary string is the operator's primary signal it may be worth saying "e.g." rather than "first". Both fine to decline; the ask (1) contract — type + message + traceback visible in log and envelope — is met, and the test pins the traceback frame (`in boom`), not just the repr.

---

espg (commented):
🤖 *from Claude (review)*

**Verdict: sound.** No blocking-correctness findings across all four phases. The scary claims were verified empirically, not just by reading: I checked out the branch in a scratch clone, ran the touched test files in full — `tests/test_index.py` + `tests/test_processing.py` + `tests/test_runner.py`, **424 passed, 0 failed** (which also means the pre-existing default-config `process_shard` tests survived silently running K=4 pooled) — and byte-compared a `write_back` manifest produced by base-commit `inline.py` (dfb7593) against the branch: **byte-identical**, so the sidecar-build product (the reason option 2 beat option 1 on issue #185) is preserved beyond what the in-tree test asserts.

Verified by inspection: (a) `write_back` ON is untouched (`selection_direct = planned and not self.write_back` gates the early exit off; `maps` still aliases `_pending_for`); (b) `planned=False` unreachable by the heuristic; (c) both orders of a path that is both data- and selection-read resolve correctly (`maps.get` runs first, the direct branch fires only for unmapped no-hyperslice reads); (d) the discriminator holds — under inline, `_execute_plan_group` only runs with `full_read=False`, so every `hyperslice=None` call on the planned route really is a selection/coordinate full read, and full `readDatasets` reads don't need the PR #152 start-edge workaround (a ranged-read quirk); (e) dtype/row parity is pinned against the hierarchical baseline by `assert_frame_equal` in every mode split; (f) phase 2's default lives in exactly two places (`_granule_workers`, README) and both moved — `validate_config` never encoded the default; (g) phase 3's per-cell copy only diverges on `granule_workers` (unclamped cells ship the shared `config_dict` by reference), payload growth is one small dict per clamped cell; (h) phase 4 changes no behavior on the write path, envelope schema unchanged, and the test asserts the traceback frame. Commits are title-only per convention; no new deps; no lint suppressions.

Findings, by severity:

**Should-fix**
1. `src/zagg/runner.py:982` (both `_cell_work` sites) — the clamp counts `len(records)` but the worker receives `_resolve_urls(records, driver)`, which drops href-less records; partially-resolvable cells under-clamp. Cheap fix, or fold into your own Question 3 (worker-side clamp covers it and direct callers at once).
2. `src/zagg/index/inline.py:552` — direct reads retain h5obj cache lines until granule close where the compiled route freed buffers after decode; on the `full_read` fallback that's base-rate coordinate volumes, multiplied by the new K=4 in-flight default. No fleet datapoint covers this combination yet — add to the A/B RSS watch list (issue #117 axis).

**Answered (your Question 1)**
3. `src/zagg/index/inline.py:608` — concurrent `readDatasets` on one handle is structurally identical to the concurrency a single batched call already creates internally (`H5Promise` fans one thread per dataset; per-cache-line locks; thread-safe S3 client per h5coro's own contract). Safe as-is; details inline.

**Nit**
4. `src/zagg/processing/write.py:443` — consider a one-line-per-failure type/message summary (traceback still first-only), and "first failing" is `as_completed`-ordered, i.e. nondeterministic.

On your Question 3: a worker-side `min(K, len(granule_urls))` would subsume finding (1) and costs one line inside `process_shard`'s already-resolved guard — but it changes the worker/dispatcher contract split agreed on issue #184 item 1, so that call stays with the repo owner.

---

espg:
🤖 *from Claude*

Adversarial-review findings folded in c65a7c8 (`fold review: clamp on resolved urls + per-failure write summary`):

- **Clamp counts resolved URLs** ([finding](https://github.com/englacial/zagg/pull/187#discussion_r3535114440)): both `_cell_work` sites now clamp on `len(_resolve_urls(records, ...))` — the granules the worker actually reads — with the lambda site hoisting the already-computed list. The dispatcher test's catalog gained a 6-record cell with only 3 s3-resolvable hrefs, asserted to clamp to 3.
- **Per-failure write summary** ([finding](https://github.com/englacial/zagg/pull/187#discussion_r3535115900)): `_write_ragged_fanout` now logs a bounded one-line-per-failure roll call (`subgroup_key: ExcType: message`, capped at 20 lines + a "N more" tail), keeping the single full traceback for the first-completed failure (the issue #175 metric-filter concern). New test pins 25 failures → 20 lines + tail + exactly one `exc_info` record. Declined only the "first → e.g." wording tweak; the roll call now shows every key, so the summary string's example key is no longer the only signal.
- **Question 1 (pooled `readDatasets`)** marked answered-safe in the body, citing the [review's h5coro source inspection](https://github.com/englacial/zagg/pull/187#discussion_r3535111351); the [RSS watch-list note](https://github.com/englacial/zagg/pull/187#discussion_r3535112915) (direct reads retain 4 MiB cache lines until granule close; full_read fallback × K=4 has no fleet datapoint, issue #117 axis) added to the phase 1 section — no code change.

Local gate after the fold: `pytest -q` 1516 passed / 26 skipped; ruff check + format clean on touched files.

---

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Folded: Question 1 in the PR body now records this answered-safe verdict (with the h5promise/cache-lock/S3-client trace). The repeat-decode nit is left as-is per the finding's own "fine to leave" — the warmed byte cache means no repeat GETs. Resolving.

---

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Folded as a no-code-change watch item: the PR body's "A/B watch-list" note now records the retained-cache-lines × K=4 axis (full_read fallback, ~245 MB/beam, issue #117 memory territory) for the next sweep. Resolving; the watch item itself stays visible in the body.

---

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed in c65a7c8: both `_cell_work` sites now clamp on the resolved-URL count (the lambda site hoists the already-computed `_resolve_urls(records, "s3")`; the local site resolves the same way), and `tests/test_runner.py` pins the partially-resolvable case (6 records / 3 s3 hrefs → clamp 3). The worker-side `min(K, len(granule_urls))` alternative stays as Question 3 in the PR body for the reviewer. Resolving.

---

src/zagg/processing/write.py
espg: 🤖 *from Claude*

Partially folded in c65a7c8: the per-failure roll call landed (`subgroup_key: ExcType: message`, capped at 20 lines + "and N more", one traceback for the first failure; pinned by `test_fanout_logs_each_failure_bounded`). The "first → e.g." wording change is declined as offered — the summary string keeps "first failing" and the completion-order caveat stands. Leaving this thread open for that declined wording point.

---

src/zagg/runner.py
```diff
@@ -1128,6 +1149,14 @@ def _cell_work(payload):
         extra = {}
         if aoi_by_shard:
             extra["aoi_payload"] = aoi_by_shard.get(int(shard_key))
+        # Per-cell granule_workers clamp (issue #184): min(K, n_granules), so
+        # a small cell doesn't spin idle reader threads; unclamped cells pass
+        # the shared config through untouched. Count the RESOLVED urls — what
+        # the worker actually reads — not the raw records: _resolve_urls
+        # drops href-less records, so len(records) would under-clamp a
+        # partially-resolvable cell (review finding, PR #187).
+        ds = _clamped_data_source(config.data_source, len(_resolve_urls(records, driver)))
```
espg, line 1160: 🤖 *from Claude (review)*

**Nit (consistency/duplication):** the local path now calls `_resolve_urls(records, driver)` twice per cell — once here for the clamp and again inside `_process_and_write` (line 1060) — while the fold deduped the lambda path into a single `granule_urls = _resolve_urls(records, "s3")` that feeds both the clamp and the invoke. The cost is a trivial list comprehension, but the asymmetry invites drift: if the two resolutions ever diverge (e.g. a future filter added to one call site), the clamp would count something different from what the worker reads — the exact bug this fold fixes. Either hoist `urls = _resolve_urls(records, driver)` here and thread it through (signature ripple, so maybe not worth it), or add a one-line comment noting the duplicate resolution is intentional and must stay in lockstep with line 1060.

---

src/zagg/processing/write.py
```diff
@@ -432,9 +435,26 @@ def _write_ragged_fanout(ragged_writes: list, store: Store, *, grid) -> None:
...
+        # the operator — the worker envelope and CloudWatch both record only
+        # this summary string — so log the first failure's full traceback
+        # here, and carry its type + message in the summary text too. A
+        # bounded one-line-per-failure roll call makes the race signature
+        # ACROSS subgroups visible (which keys, same or different errors)
+        # without flooding the log when many of the K writes fail at once.
+        for key, exc in errors[:20]:
+            logger.error(f"  ragged (CSR) subgroup {key}: {type(exc).__name__}: {exc}")
+        if len(errors) > 20:
+            logger.error(f"  ... and {len(errors) - 20} more subgroup write failure(s)")
```

**Nit (log ordering):** the roll-call lines are indented with two leading spaces — visually continuation lines — but they're emitted *before* the header record (`"ragged (CSR) subgroup write failed (shard_key ...)"`, line 449). In a raw CloudWatch/console stream the operator sees orphaned `  ragged (CSR) subgroup ...` lines with no preceding context, then the header last. Consider emitting the header (or a short lead-in like `f"ragged (CSR): {len(errors)} of {len(ragged_writes)} subgroup writes failed:"`) first, then the roll call.

Related micro-nit: in the common single-failure case the roll call now duplicates the header's content (`subgroup {key}: {Type}: {msg}` appears twice, three times counting the raised summary). Guarding with `if len(errors) > 1:` would keep the one-failure log identical to the pre-fold output.

**Nit (magic number):** `20` appears as a bare literal twice here (and once in the test). A named module constant next to `_RAGGED_WRITE_CONCURRENCY` (e.g. `_RAGGED_FAILURE_ROLL_CALL = 20`) would pin code and test to the same value.

Cap logic itself is correct: `errors[:20]` + `len(errors) > 20` gives exactly min(N, 20) lines and the "and N-20 more" tail only when needed; `exc_info` stays on the single header record.

---

tests/test_processing.py
```diff
@@ -1257,20 +1257,68 @@ def spy_tpe(max_workers):
...
+        """Multiple subgroup failures each get a one-line ``key: Type: msg``
+        summary (issue #186: the race signature ACROSS subgroups), capped at
+        20 lines, while exactly one record — the first completed failure —
+        carries the full traceback."""
+        import logging
+
+        import zagg.processing.write as wmod
+        from zagg.processing.write import _write_ragged_fanout
+
+        grid, writes = self._grid_and_writes(25)
```
espg, line 1309: 🤖 *from Claude (review)*

**Nit (environment-sensitive test):** this test relies on the pool branch being taken (`workers > 1`), but doesn't pin `fd_safe_max_workers` — on a low-ulimit host `min(128, 25, fd_safe_max_workers())` can clamp to 1, the serial branch raises the bare `OSError` from the first write, and the `pytest.raises(RuntimeError, ...)` plus the roll-call assertions all fail. The neighboring `test_fanout_byte_identical_parallel_vs_serial` (line 1221) monkeypatches `fd_safe_max_workers` for exactly this reason, with a comment naming the hazard. One line fixes it:

```python
monkeypatch.setattr(wmod, "fd_safe_max_workers", lambda: 128)
```

(Pre-existing `test_fanout_surfaces_write_failure` shares this fragility; out of this fold's scope, but the same one-liner would harden it.)

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of fold commit c65a7c8 ("fold review: clamp on resolved urls + per-failure write summary") — the delta since the earlier review of 6ad62ea..448e815.

**Verdict: the fold is correct; no must-fix findings.** Both prior findings are properly addressed:

- **Resolved-URL clamp:** both dispatch paths now clamp on `len(_resolve_urls(...))` — what the worker actually reads. The lambda path dedupes the resolution into `granule_urls` reused for the invoke, guaranteeing clamp and payload can't diverge. Edge cases check out: zero resolvable URLs hits the pre-existing `max(n, 1)` guard (unit-pinned at tests/test_runner.py line 1141), K-already-lower and at/above-K pass-through are unchanged, and non-mutation of the shared `data_source` is pinned (line 1132). The shared-config byte-identity for unclamped cells is preserved (`ds is None` → shared object/dict as-is, asserted in both integration tests).
- **Bounded roll call:** exactly min(N, 20) one-line summaries + "and N−20 more", traceback (`exc_info`) on the single header record only, raised summary string unchanged, no behavior change to the write path itself. The new test pins the cap (25 failures → 20 lines + "and 5 more"), per-key error attribution, and the single-traceback invariant.

Three nits inline (all diff-scoped, none blocking): duplicate URL resolution on the local path vs the deduped lambda path; roll-call log lines emitted before the header they visually continue (plus single-failure duplication); bare `20` literal vs a named constant; and the new fanout test not pinning `fd_safe_max_workers` the way its neighbor deliberately does.

Checked/ran: `git show c65a7c8` against the full PR diff; `uv run pytest tests/test_runner.py tests/test_processing.py` — 352 passed; `ruff check` + `ruff format --check` on the four touched files — clean. Touched src modules within limits (write.py 789 lines; runner.py's pre-existing size was covered in the earlier review and the fold adds ~11 net lines).

---

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded in c2fecab via the offered lockstep comment — the clamp's resolve now carries a note that it must stay in step with `_process_and_write`'s own `_resolve_urls(records, driver)` call. The signature ripple wasn't worth it: `_process_and_write` has direct test callers that pass `records`. Resolving.

---

src/zagg/processing/write.py
espg: 🤖 *from Claude*

All three folded in c2fecab: header (with traceback) now logs first with the roll call as its continuation; the roll call is guarded by `len(errors) > 1` so a single failure logs exactly the pre-fold output; and the cap is the named `_RAGGED_FAILURE_LOG_CAP = 20` beside `_RAGGED_WRITE_CONCURRENCY`, referenced by code and test. Ordering and the single-failure guard are pinned in `test_fanout_logs_each_failure_bounded` / `test_fanout_surfaces_write_failure`. Resolving.

---

tests/test_processing.py
espg: 🤖 *from Claude*

Folded in c2fecab: `test_fanout_logs_each_failure_bounded` pins `fd_safe_max_workers` to 128 with the hazard comment, and since the single-failure guard meant touching `test_fanout_surfaces_write_failure` anyway, it got the same one-line pin. Resolving.

---

espg:
🤖 *from Claude*

**Status: all phases complete, both review rounds folded, CI green on c2fecab — awaiting espg review.** This run:

- Ran the fresh-context adversarial review of the first fold commit c65a7c8 ([review](https://github.com/englacial/zagg/pull/187#pullrequestreview-4644501334)): verdict sound, no must-fix findings, four nits.
- Folded all four in c2fecab: header-first failure logging with a `len(errors) > 1` guard (single failure logs exactly the pre-fold output), the cap named as `_RAGGED_FAILURE_LOG_CAP`, `fd_safe_max_workers` pins in both fanout failure tests, and a lockstep comment on the local clamp's duplicate `_resolve_urls` (the signature ripple wasn't worth it — `_process_and_write` has direct test callers). Replies on each inline thread record the dispositions; the one open thread is the declined "first → e.g." wording nit, left standing per convention.
- Verified locally: `ruff check` + `ruff format --check` clean; `tests/test_processing.py` + `tests/test_runner.py` full modules 352 passed on c2fecab (earlier full `pytest -q` on the first-fold state: 1516 passed, 26 skipped).

**Process note — draft state:** repo conventions (CLAUDE.md §2) say a PR in this state — all phases done, findings folded, CI green, not waiting on a question — should be flipped from draft to ready-for-review by the routine itself. This run's routine configuration explicitly forbids marking PRs ready-for-review, and CLAUDE.md's own preamble says to surface such a conflict rather than guess. So the PR stays **draft** with `waiting` applied; it is review-ready in substance, and the draft→ready flip (one click, reversible) is left to espg. The open items for review remain Questions 2 (pre-existing lint drift, flagged only) and 3 (worker-side clamp for direct `process_shard` callers) in the body.

---

PR #188, opened 2026-07-07
Paced S3 retry backoff for 503 SlowDown

### LINKED ISSUE #186: Sharded-path parallel CSR writes race under real CPU parallelism (fails 3/3 on 6 GB workers, largest shard)

espg:
Closes #186. Implements fix-menu item (1) from the [root-cause comment](https://github.com/englacial/zagg/issues/186#issuecomment-4906248940) — the paced retry policy. Fix-menu item (2) (fewer metadata round trips) was a design fork, **decided as option (b)** (espg direction, in-session): this PR stays retry-only and the op storm + upstream zarr ask is tracked on #189 with the full measurements and an escalation trigger.

## Phases

- [x] Phase 1 — paced S3 retry defaults (`src/zagg/store.py`)
- [x] Fold adversarial-review findings ([review](https://github.com/englacial/zagg/pull/188#pullrequestreview-4647421702)): short `_POLL_RETRY_CONFIG` for the issue #151 result poller; `retry_config=None` → zagg default; deepcopy at the injection site; `server_close()` in the 503 test; docstring + comment clarifications
- [x] ~~Phase 2 — cut CSR-writer metadata round trips~~ → **decided (b): out of scope here, tracked on #189**
- [x] Phase 3 — read-only retry carve-out (espg direction, in-session); [second review](https://github.com/englacial/zagg/pull/188#pullrequestreview-4648803153) findings folded (None-edge test, which-bound-binds comment, population wording)
- [x] Phase 4 — close the coverage gaps the second review found (espg direction, in-session: "both"): route the example notebooks + `temporal.open_dataset`'s NetCDF branch through the read-only policy; anonymous (`skip_signature`) branch in `open_store`

## What / approach

**The defaults were the amplifier, empirically confirmed.** Reproduced the trapped fleet signature against a local always-503 `SlowDown` endpoint: obstore's default policy (10 retries, 100 ms init backoff, base 2, uniform-from-zero jitter) burns its entire retry budget in **2–4 s** — gaps of 100–500 ms — because the jitter draws uniformly from zero over the nominal exponent. That matches the fleet failures' `after 10 retries` in 1.8–3.1 s exactly: near-immediate retries that hammer the already-throttled prefix, then surface the error. Probe (default config): `11 requests in 3.3s gaps=[0.10, 0.18, 0.23, 0.23, 0.14, 0.20, 0.29, 0.35, 0.53, 1.06]`.

**Phase 1** sets a module-level `_S3_RETRY_CONFIG` in `store.py` — `max_retries: 12`, `init_backoff: 1 s`, `max_backoff: 30 s`, `base: 2`, `retry_timeout: 180 s` (obstore's default, kept below the 5-minute credential-validity bound its docs warn about; since the nominal sleep sum of 12 paced retries exceeds 180 s, the timeout is the effective bound under a long burst) — injected in `_s3_object_store`, the **single `S3Store` construction site** in the package, so every zagg S3 client inherits it: output store, sidecar manifests (`open_object_store` via `index/inline.py`), side-channel result JSON, temporal reads. A caller-supplied `retry_config` wins; `None` means the zagg default (nothing reconstructs the fast-burn policy by accident), and each store gets a deep copy so none alias the module constant. Expected pacing under a sustained burst: jittered ~0.5/1/2/4/8/15/15/… s sleeps, ≈1–2 min of headroom — the backoff S3's `Please reduce your request rate` asks for. The policy only engages on retryable failures (5xx / dropped connections / read timeouts) — healthy requests are untouched, and 4xx still fails immediately. (h5coro's read-side S3 client is its own and out of scope; every trapped failure was on the output store.)

**Poller carve-out (review fold):** the issue #151 async result poller already retries at the loop level (a fetch every 5 s until `poll_timeout_s`, deadline checked only *between* fetches), so `_result_fetcher` opens its store with a short `_POLL_RETRY_CONFIG` (2 retries, 15 s ceiling) — a 5xx during one fetch can no longer block ~3 min and silently overrun the poll deadline.

**Phase 3 — read-only carve-out.** obstore's retry policy is per store client, and `open_store`'s existing `read_only` flag cleanly separates interactive readers from fleet writers. `read_only=True` stores default to `_S3_READONLY_RETRY_CONFIG` (4 retries, 30 s ceiling, 1→8 s backoff; nominal sleep sum 15 s, so `max_retries` is the effective bound): a genuinely dead endpoint surfaces a clear error in ~15 s instead of feeling hung for ~2–3 min, while still riding a short throttle burst (deliberately not a sustained one). Same precedence rules. Fleet workers open read-write, so the #186 fix is unaffected.

**Phase 4 — actually cover the interactive population.** The second review found the carve-out's intended audience mostly bypassed it: the example notebooks constructed `S3Store` + `ObjectStore(read_only=True)` directly (keeping obstore's fast-burn default — the exact #186 behavior), and `temporal.open_dataset`'s NetCDF branch read through `open_object_store` on the 180 s write policy. Per espg ("both"): the three notebooks (`cryocloud_example`, `jupyterhub_example`, `rasterized_zarr`) now call `open_store(f"s3://{bucket}/{prefix}", read_only=True, skip_signature=True)` — `rasterized_zarr`'s bespoke `client_options`/`retry_config` ride through kwargs unchanged — and the temporal NetCDF branch passes `retry_config=_S3_READONLY_RETRY_CONFIG` explicitly. Enabling this required an **anonymous branch** in `_s3_object_store`: with `skip_signature=True` and no explicit credentials, no `Boto3CredentialProvider` is constructed (its `__init__` raises `ValueError` without ambient AWS credentials — `open_store` was previously unusable on binder). The anonymous construction is parameter-identical to what the notebooks did directly.

## How tested

- `tests/test_store.py::TestS3RetryConfig`: the default lands on both credential branches; a caller-supplied `retry_config` wins (read-write and read-only); explicit `None` gets the zagg default on both paths; non-aliasing (nested `backoff` included, both constants); `skip_signature=True` skips the credential provider entirely (provider never constructed); and an end-to-end test drives `open_object_store` against a local always-503 HTTP server asserting exactly `1 + max_retries` requests arrive.
- `tests/test_runner.py::TestResultFetcher::test_poller_store_uses_short_retry_config` pins the poller carve-out; `tests/test_temporal.py::test_open_dataset_s3_netcdf_fetches_bytes` now also pins the NetCDF branch's read policy.
- Notebooks: nbformat-4 valid, every edited cell `ast.parse`s, no remaining direct `S3Store`/`ObjectStore` construction, stable cell ids.
- Full local gate: `ruff check` + `ruff format --check` clean on all touched files; full `pytest -q` green on the phase-1 commit — **1520 passed, 26 skipped** (3.12); later commits re-ran `tests/test_store.py` + `tests/test_runner.py` + `tests/test_temporal.py` (26 + 75 passed). CI green through 7f20641; phase-4 CI pending at last update.

## Questions for review

1. ~~Phase 2 design fork~~ — **decided (b)** (espg direction, in-session): retry-only here; op-storm measurements, the internals-writer prototype, the upstream zarr ask, and the escalation trigger all live on #189.
2. ~~Retry-value calibration~~ — **decided** (espg direction, in-session): uniform paced default for read-write stores + the phase-3 read-only carve-out, extended to the notebooks and the temporal NetCDF branch in phase 4.


tests/test_store.py
```diff
@@ -105,6 +105,96 @@ def test_endpoint_only_no_creds(self, mock_s3):
...
+                        "init_backoff": timedelta(milliseconds=400),
+                        "max_backoff": timedelta(seconds=2),
+                        "base": 2,
+                    },
+                },
+            )
+            with pytest.raises(Exception, match="SlowDown|503|Server"):
+                obstore.put(store, "zarr.json", b"{}")
+        finally:
+            srv.shutdown()
```
espg, line 269: 🤖 *from Claude (review)*

**should-fix:** `srv.shutdown()` stops the `serve_forever` loop but never closes the listening socket — add `srv.server_close()` after it in the `finally`. `ThreadingHTTPServer` holds the fd otherwise, which leaks into the rest of the pytest session and can surface as a `ResourceWarning`; `daemon=True` only covers interpreter exit, not the in-session leak. Runs regardless of pass/fail either way, so no hang risk — just the fd.

---

src/zagg/store.py
```diff
@@ -1,10 +1,31 @@
...
+# ``retry_config`` through ``**kwargs`` to override.
+_S3_RETRY_CONFIG = {
+    "max_retries": 12,
+    "retry_timeout": timedelta(seconds=180),
+    "backoff": {
+        "init_backoff": timedelta(seconds=1),
+        "max_backoff": timedelta(seconds=30),
+        "base": 2,
+    },
+}
```
espg, line 31: 🤖 *from Claude (review)*

Two non-blocking observations on the config block:

- **nit (shared mutable dict):** `kwargs.setdefault` aliases this exact dict (and its nested `backoff` dict) into every store construction. obstore converts it to a Rust config at construction time, so nothing mutates it today, but any future caller that tweaks `kwargs["retry_config"]["max_retries"]` before construction would silently edit the global default for every later store. A defensive `copy.deepcopy` at the setdefault site — or one warning line in this comment — closes that off cheaply.
- **note (max_retries vs retry_timeout):** worst-case nominal sleep sum is 1+2+4+8+16+30×7 = 241 s, above the 180 s `retry_timeout`, so unlucky jitter draws can hit the timeout before the 12th retry — the effective bound is the timeout, not the count. That's consistent with the "~2 min of headroom" framing, but a clause here would stop someone later "fixing" `max_retries` expecting it to always be exhausted.

---

src/zagg/store.py
```diff
@@ -109,6 +130,7 @@ def _s3_object_store(
 
     bucket, prefix = parse_s3_path(path)
     region = kwargs.pop("region", "us-west-2")
+    kwargs.setdefault("retry_config", _S3_RETRY_CONFIG)
```
espg: 🤖 *from Claude (review)*

**nit:** two small gaps at the override seam —

- An explicit `retry_config=None` sails past `setdefault` (the key exists), so `S3Store` gets `None` → obstore's built-in policy, i.e. the exact 2–4 s burn this PR removes. Defensible as an escape hatch, but it's neither documented nor tested — `test_caller_override_wins` only covers a truthy override. Worth a one-line test pinning whichever semantics is intended.
- `open_store`'s docstring still documents `**kwargs` as only `region`, while the module comment above now advertises `retry_config` as a supported override — list it there (and in `open_object_store`) so callers don't have to find the private constant's comment.

---

src/zagg/store.py
```diff
@@ -1,10 +1,31 @@
...
+# S3 retry pacing (issue #186). obstore's default policy retries 5xx/connection
+# errors up to 10 times with jittered exponential backoff from 100 ms — under a
+# sustained 503 SlowDown burst the jitter draws small sleeps and the whole
+# budget is spent in ~2-4 s of near-immediate retries, hammering the throttled
+# prefix and then surfacing the error (the trapped fleet failures burned all 10
+# retries in 1.8-3.1 s). These defaults pace retries seconds apart with ~2 min
+# of headroom, which is what S3's "Please reduce your request rate" asks for.
+# ``retry_timeout`` stays at obstore's 180 s default, below the 5-minute
+# credential-validity bound its docs warn about. Callers can pass their own
+# ``retry_config`` through ``**kwargs`` to override.
```
espg: 🤖 *from Claude (review)*

**note (blast radius on the issue #151 result poller):** `_poll_lambda_result` in `src/zagg/runner.py` already implements its own outer retry — a fetch fault is "treated as a miss" and re-polled every ~5 s until `poll_timeout_s`. Under this default, each `_fetch_result` GET during a sustained 5xx now blocks up to `retry_timeout` (180 s) *inside* a single poll iteration, and the deadline is only checked between fetches. Consequences: a shard can overrun `poll_timeout_s` by up to ~3 min, the `no worker result within {poll_timeout_s:.0f}s` message understates true wall time, and `_ASYNC_POLL_MARGIN_S` (90 s) is now smaller than one fetch's worst case. Since that call site paces itself already, consider a short `retry_config` override in `_result_fetcher` — the override plumbing this PR adds makes it a two-line change. Fine as a phase-2 item or an entry under Questions for review rather than a blocker here.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context review of phase 1 (commit 3138fc6). What I checked and ran:

- **Both construction branches covered:** `kwargs.setdefault("retry_config", ...)` lands before the credential branch and both `S3Store` calls splat `**kwargs`, so credentialed, custom-endpoint, and ambient-provider paths all inherit the default — confirmed by the two branch tests. No call path passes `retry_config` positionally (`_s3_object_store` is the single construction site; all callers forward keyword-only).
- **obstore compatibility:** the dict's keys and `timedelta` types match the installed obstore 0.11 `RetryConfig`/`BackoffConfig` stubs, and `1 + max_retries` total requests is object_store's documented contract (the e2e test pins it). One unverified point: the pyproject floor is `obstore>=0.8.2` and I could not confirm from this worktree exactly when `retry_config` gained these fields — almost certainly long before 0.8, so flagging as unverified rather than as a finding.
- **Ran:** `pytest tests/test_store.py -q` (19 passed, ~1.7 s), `TestS3RetryConfig` three consecutive times (no flakes), and `ruff check` + `ruff format --check` on both touched files (clean).

Findings are inline: one should-fix (the e2e test never calls `srv.server_close()`, leaking the listening socket), one note on the runner poller's timeout accounting now that a single fetch can block up to 180 s, and two nits (shared mutable config dict; the untested `retry_config=None` seam plus docstring drift). Nothing blocks phase 1. The PR body's Question 2 (whether interactive reads should get a shorter policy) is a genuine design call and correctly left for a human — this review takes no position that would preempt it.

---

tests/test_store.py
espg: 🤖 *from Claude*

Fixed in d165c83: `srv.server_close()` added after `shutdown()` in the `finally`. Resolving.

---

src/zagg/store.py
espg: 🤖 *from Claude*

Both folded in d165c83: the injection site now `copy.deepcopy`s the default (pinned by `test_default_not_aliased`, nested `backoff` included), and the config comment carries the retry_timeout-is-the-effective-bound clause. Resolving.

---

src/zagg/store.py
espg: 🤖 *from Claude*

Folded in d165c83, with the semantics decided the other way: `retry_config=None` now means the zagg default (the `setdefault` became an explicit `is None` check + deepcopy), so nothing can reconstruct the fast-burn policy by accident — obstore's built-in behavior is reachable only by passing an explicit dict. Pinned by `test_explicit_none_gets_default`; `open_store`'s docstring now lists `retry_config` under `**kwargs`. Resolving.

---

src/zagg/store.py
espg: 🤖 *from Claude*

Taken now rather than deferred (d165c83): `_result_fetcher` passes a short `_POLL_RETRY_CONFIG` (2 retries, 15 s ceiling, 0.5→2 s backoff) since the poll loop owns retrying at its 5 s cadence — one fetch can no longer overrun the deadline by minutes. Pinned by `test_poller_store_uses_short_retry_config`. Resolving.

---

src/zagg/store.py
```diff
@@ -1,10 +1,52 @@
...
+    },
+}
+
+# Read-only carve-out: stores opened via ``open_store(read_only=True)`` are the
+# interactive population (notebooks, readers, read-back analysis) — for them a
+# genuinely failing endpoint should surface a clear error in ~30 s, not feel
+# hung for the full write policy above. Still paced (rides a typical throttle
+# burst), and reads are far harder to throttle anyway (S3's per-prefix GET
+# budget is ~5,500/s vs ~3,500 for PUT). Fleet workers open read-write, so the
+# issue #186 fix is unaffected.
```
espg: 🤖 *from Claude (review)*

**note (comment overstates the covered population).** The claim that `open_store(read_only=True)` is "the interactive population (notebooks, readers, ...)" doesn't match the tree:

- The three shipped example notebooks (`notebooks/cryocloud_example.ipynb`, `notebooks/jupyterhub_example.ipynb`, `notebooks/rasterized_zarr.ipynb`) construct `S3Store` + `ObjectStore(store=s3, read_only=True)` **directly**, bypassing `open_store` entirely. cryocloud/jupyterhub therefore keep obstore's fast-burn default (10 retries in ~2-4 s — the exact behavior issue #186 diagnosed); rasterized_zarr sets its own. Only the `bench/*.ipynb` notebooks route through `open_store(read_only=True)`.
- The "readers" half is split: `temporal.open_dataset` (src/zagg/temporal.py:516) gets this policy for `.zarr` URIs, but the NetCDF/HDF5 branch of the *same function* (temporal.py:531) goes through `open_object_store` — which has no `read_only` parameter — so a pure GET of a static input keeps the 180 s **write** policy. Same interactive caller, ~30 s vs ~3 min failure latency depending on file suffix.

Diff-scoped fix: temper this comment to what the carve-out actually reaches (`open_store(read_only=True)` callers: `temporal.open_dataset` zarr branch, bench notebooks, external users of the public API). Whether to route the example notebooks through `open_store` and/or give `open_object_store` a read-policy knob is a scope decision — leaving that standing for espg rather than treating it as part of this commit.

---

src/zagg/store.py
```diff
@@ -1,10 +1,52 @@
...
+_S3_READONLY_RETRY_CONFIG = {
+    "max_retries": 4,
+    "retry_timeout": timedelta(seconds=30),
+    "backoff": {
+        "init_backoff": timedelta(seconds=1),
+        "max_backoff": timedelta(seconds=8),
+        "base": 2,
+    },
+}
```
espg, line 51: 🤖 *from Claude (review)*

**nit (which bound binds — comment vs numbers).** Nominal sleep sum for 4 retries at 1→8 s base-2 is 1+2+4+8 = 15 s, and obstore's jitter draws uniform from zero (as the `test_default_paces_retries` comment notes), so the expected backoff total is ~7.5 s: on a dead endpoint `max_retries` exhausts well before the 30 s `retry_timeout`, and the error typically surfaces in under ~15 s, not "~30 s". The write-policy comment above is careful to say which bound is effective ("the timeout, not max_retries"); this one implies the opposite relationship than actually holds. Suggest either stating that here too (30 s is a ceiling that only binds when individual requests are slow; max_retries is the effective bound) or softening "~30 s" to "well under a minute / ~15 s". "Rides a typical throttle burst" is fair for a short burst, but at ~7.5 s expected pacing it will *not* outlast a sustained SlowDown episode — which is presumably the intended tradeoff for interactive reads, so worth saying explicitly.

---

tests/test_store.py
```diff
@@ -105,6 +105,136 @@ def test_endpoint_only_no_creds(self, mock_s3):
...
+        _, kwargs = s3_cls.call_args
+        assert kwargs["retry_config"] == _S3_READONLY_RETRY_CONFIG
+        assert kwargs["retry_config"] is not _S3_READONLY_RETRY_CONFIG  # no aliasing
+
+    def test_read_only_override_wins(self, mock_s3):
+        s3_cls, _ = mock_s3
+        custom = {"max_retries": 7}
+        open_store("s3://bucket/prefix.zarr", read_only=True, retry_config=custom)
+        _, kwargs = s3_cls.call_args
+        assert kwargs["retry_config"] == custom
```
espg, line 177: 🤖 *from Claude (review)*

**should-fix (test gap: explicit `retry_config=None` + `read_only=True`).** The write path pins this exact edge with `test_explicit_none_gets_default` (guarding against a `setdefault`/`not in kwargs` regression letting `None` through to obstore). The read-only branch has the same `kwargs.get("retry_config") is None` shape at `_open_s3_store` (src/zagg/store.py:142) but only the absent-key case is tested — if the guard ever regressed to `"retry_config" not in kwargs`, `open_store(path, read_only=True, retry_config=None)` would silently fall through to the **write** default with no test failing. One test mirroring `test_explicit_none_gets_default` closes it:

```python
def test_read_only_explicit_none_gets_short_policy(self, mock_s3):
    s3_cls, _ = mock_s3
    open_store("s3://bucket/prefix.zarr", read_only=True, retry_config=None)
    _, kwargs = s3_cls.call_args
    assert kwargs["retry_config"] == _S3_READONLY_RETRY_CONFIG
```

Minor while there: `test_read_only_gets_short_policy` checks top-level non-aliasing only; its write counterpart `test_default_not_aliased` also asserts the nested `backoff` dict isn't aliased. The `deepcopy` makes it safe today, but the symmetric assertion is one line.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of commit 18ca5be only (read-only stores get a short retry policy); earlier commits 3138fc6/d165c83 were previously reviewed and are out of scope.

**Verified correct:**
- Precedence chain in `_open_s3_store`: truthy caller `retry_config` passes through both `is None` guards and wins; explicit `None` (and absent key) on a read-only store gets the read-only default via `deepcopy`; a read-write store is untouched by the carve-out and gets the paced write default in `_s3_object_store`. Consistent with the write path's `is None` semantics from the prior fold.
- No interaction with the poller carve-out: `_POLL_RETRY_CONFIG` (runner.py) is passed explicitly through `open_object_store`, which `_open_s3_store` doesn't gate. The zarr `ObjectStore(read_only=...)` wrapper is unchanged.
- Values: 4 retries / 30 s / 1→8 s base-2 is internally coherent (nominal sleep sum 15 s < 30 s timeout), though the comment's "~30 s" framing is off — see inline nit.

**Findings (3 inline):**
1. should-fix — tests: no case for `read_only=True` + explicit `retry_config=None`; the write path pins this edge, the read-only branch doesn't.
2. note — the constant's comment claims the carve-out covers "notebooks, readers", but the shipped example notebooks bypass `open_store` (keeping obstore's fast-burn default), and the NetCDF branch of `temporal.open_dataset` reads through `open_object_store` on the write policy.
3. nit — comment timing math: `max_retries`, not the 30 s timeout, is the effective bound; typical failure surfaces in well under 15 s.

**Ran:** `pytest tests/test_store.py -q` (23 passed), `ruff check` + `ruff format --check` on the two changed files (clean). Module sizes fine (store.py 223 lines).

---

tests/test_store.py
espg: 🤖 *from Claude*

Both folded in 7f20641: `test_read_only_explicit_none_gets_short_policy` mirrors the write-path None edge, and `test_read_only_gets_short_policy` now asserts nested `backoff` non-aliasing symmetrically. Resolving.

---

src/zagg/store.py
espg: 🤖 *from Claude*

Folded in 7f20641: the comment now states the 15 s nominal sum, that `max_retries` is the effective bound, and that 30 s is a ceiling — mirroring the write-policy comment's which-bound-binds framing. The sustained-burst tradeoff (interactive reads deliberately won't outlast a long SlowDown episode) is the intent. Resolving.

---

src/zagg/store.py
espg: 🤖 *from Claude*

Diff-scoped part folded in 7f20641: the comment no longer claims the notebooks and now names the exclusions explicitly (`open_object_store` paths and direct-client notebook code). The standing scope question — routing the example notebooks through `open_store` (cryocloud/jupyterhub currently keep obstore's fast-burn default, the exact issue #186 behavior) and/or a read-policy knob for `open_object_store` (the temporal NetCDF branch keeps the 180 s write policy on a pure GET) — has been raised with espg in-session; leaving this thread open until that call is made.

---

espg: 🤖 *from Claude*

Scope question answered by espg (in-session): both. Landed in 1d31623 — the three notebooks now route through `open_store(..., read_only=True, skip_signature=True)` (which required a new anonymous branch in `_s3_object_store`: `Boto3CredentialProvider` raises without ambient credentials, so binder could never have used `open_store` before), and `temporal.open_dataset`'s NetCDF branch passes `retry_config=_S3_READONLY_RETRY_CONFIG` explicitly. Resolving.

---

src/zagg/store.py
```diff
@@ -127,6 +187,11 @@ def _s3_object_store(
         if endpoint_url:
             opts["endpoint"] = endpoint_url
         s3 = S3Store(**opts, **kwargs)
+    elif kwargs.get("skip_signature"):
+        # Anonymous read of a public bucket: no credential provider —
+        # Boto3CredentialProvider raises without ambient AWS credentials,
+        # which anonymous environments (e.g. binder) lack by definition.
+        s3 = S3Store(bucket, prefix=prefix, region=region, **kwargs)
```
espg, line 200: 🤖 *from Claude (review)*

**note (addressing-style asymmetry):** The credentialed/endpoint branch above pins `"virtual_hosted_style_request": False` (its comment: "required for dotted bucket names (TLS)"), but this anonymous branch relies on obstore's path-style *default*. I verified the construction is parameter-identical to what the notebooks did directly (positional bucket, `prefix`, `region`, `skip_signature`, no addressing override), so no behavior change is introduced — the dotted `us-west-2.opendata.source.coop` bucket works today exactly as it did. But the anonymous path now serves that dotted bucket through zagg's own code, and it's the only S3 branch that leaves addressing to the upstream default: if obstore ever flips to virtual-hosted-style, anonymous reads break on TLS while the credentialed branch keeps working. Worth either pinning `virtual_hosted_style_request=False` here too (byte-identical today) or a one-line comment saying the default is deliberately inherited for parity with the old direct construction.

Precedence check (fine as-is): with `skip_signature=True` **plus** explicit `credentials`/`endpoint_url`, the first branch wins and `skip_signature` still rides through `**kwargs` into `S3Store`, so requests stay unsigned with the keys unused — sane, no kwarg collision, matches what passing both to `S3Store` directly would do.

---

src/zagg/temporal.py
```diff
@@ -525,11 +525,17 @@ def open_dataset(uri, *, credentials=None, endpoint_url=None, region="us-west-2"
...
     bucket, key = parse_s3_path(uri)
+    # Pure read of a static input: same failure-latency policy as the .zarr
+    # branch above (open_object_store has no read_only concept, issue #186).
     store = open_object_store(
-        f"s3://{bucket}", credentials=credentials, endpoint_url=endpoint_url, region=region
+        f"s3://{bucket}",
+        credentials=credentials,
+        endpoint_url=endpoint_url,
+        region=region,
+        retry_config=_S3_READONLY_RETRY_CONFIG,
```
espg, line 538: 🤖 *from Claude (review)*

**nit (aliasing consistency):** `_S3_READONLY_RETRY_CONFIG` is passed by reference — no `copy.deepcopy`. This matches the `runner._POLL_RETRY_CONFIG` precedent (also passed uncopied at its construction site), and obstore consumes the dict at construction, so there's no live hazard. But `_s3_object_store` deep-copies its defaults precisely so "no store's kwargs alias the module default", and `test_default_not_aliased` pins that invariant with `is not` assertions — this call is now the one site where the module constant itself crosses into a store construction. A `copy.deepcopy(...)` here (or accepting the runner precedent as the rule and noting it) would keep the invariant uniform; as written, a future mutation inside the store path would corrupt the constant that `_open_s3_store` copies *from*.

---

tests/test_store.py
```diff
@@ -105,6 +105,160 @@ def test_endpoint_only_no_creds(self, mock_s3):
...
+    def test_anonymous_read_skips_credential_provider(self, mock_s3):
+        """``skip_signature=True`` (public bucket, e.g. the example notebooks
+        on binder) must not construct ``Boto3CredentialProvider`` — it raises
+        without ambient AWS credentials, which anonymous environments lack."""
+        s3_cls, prov_cls = mock_s3
+        open_store("s3://bucket/public.zarr", read_only=True, skip_signature=True)
+        _, kwargs = s3_cls.call_args
+        assert kwargs["skip_signature"] is True
+        assert "credential_provider" not in kwargs
+        prov_cls.assert_not_called()
```
espg, line 188: 🤖 *from Claude (review)*

**nit (coverage gap):** This test does pin the ValueError hazard — `prov_cls` is patched at the source module (`obstore.auth.boto3`), so a regression that reintroduces the function-local import + construction would trip `prov_cls.assert_not_called()`, and `"credential_provider" not in kwargs` catches it independently. Two adjacent cases are untested, though:

1. **`skip_signature=True` + explicit `credentials`** — which branch wins is load-bearing precedence (`credentials or endpoint_url` takes priority; `skip_signature` rides through kwargs). A one-liner asserting the provider is skipped *and* `access_key_id` is set would pin it.
2. The anonymous branch inheriting the read-only retry default — `assert kwargs["retry_config"] == _S3_READONLY_RETRY_CONFIG` here would confirm the new branch flows through the same `retry_config` plumbing the other branches are tested on (it does today, since the injection happens before the branch, but the assertion is nearly free).

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of commit 1d31623 only (prior commits 3138fc6/d165c83/18ca5be/7f20641 already reviewed and folded). No must-fix findings — the commit does what it says. Three inline comments: one note (anonymous branch inherits obstore's addressing default while the credentialed branch pins path-style — no behavior change today, but the asymmetry is one obstore default-flip away from mattering for the dotted bucket), two nits (uncopied `_S3_READONLY_RETRY_CONFIG` in `temporal.open_dataset` vs the deepcopy invariant `_s3_object_store` documents and tests; untested `skip_signature`+`credentials` precedence).

What I checked:

- **Anonymous branch equivalence** — traced `open_store(read_only=True, skip_signature=True)` → `_open_s3_store` → `_s3_object_store`: construction is parameter-identical to the notebooks' old direct `S3Store(...)` (positional bucket, `prefix`, `region` default `us-west-2` matches the old explicit value, `skip_signature=True`) plus the intended `retry_config` injection, and `Boto3CredentialProvider` is never imported on this path — so the cells run without AWS credentials, which is the binder requirement. `.binder/postBuild` installs zagg from the checked-out repo, so the new `from zagg import open_store` import is binder-resolvable.
- **Notebooks** — all three parse as valid nbformat 4 JSON; cryocloud/jupyterhub cell ids stable and unique (rasterized_zarr is nbformat 4.4, id-less throughout — pre-existing, and the edit correctly added none); no remaining references to the removed `s3`/`s3_store` variables or `obstore.store` imports (cryocloud's `icechunk.s3_store(...)` hit is an unrelated icechunk API); all three notebooks are stored output-free, so no stale outputs were invalidated; `rasterized_zarr`'s bespoke `client_options`/`retry_config` ride through kwargs and win over the read-only default per the `retry_config is None` guard; trailing-newline EOF fixes are correct. One micro-nit not worth an inline thread: cryocloud cell `40f1f450` rebuilds `f's3://{OUTPUT_BUCKET}/{OUTPUT_KEY}'` when cell `47a35a59` already defines `OUTPUT_URL` as exactly that string.
- **temporal.py** — `retry_config` threads through `open_object_store` → `_s3_object_store` untouched (not `None`, so the write-policy default doesn't overwrite it); explicit-pass is consistent with the `runner._POLL_RETRY_CONFIG` precedent; the updated test lambda now captures kwargs correctly and returning `None` as the fake store is harmless since `obstore.get` is also patched.
- **Tests/CLAUDE.md** — `pytest tests/test_store.py tests/test_temporal.py -q`: 75 passed; `ruff check` + `ruff format --check` clean on the four touched Python files; no new dependencies; module sizes well under the 1000-line bound (store.py 233, temporal.py 578); comments are terse and grounded in issue #186.

---

src/zagg/store.py
espg: 🤖 *from Claude*

Folded in 8532b0e as the comment option: the anonymous branch now states the addressing default is deliberately inherited for parity with the notebooks' old direct construction (pinning `virtual_hosted_style_request=False` was declined — it would deviate from the construction that's proven against the dotted source.coop bucket, trading a hypothetical obstore default flip for a real behavioral change today). Precedence check noted, and now pinned by `test_skip_signature_with_credentials_takes_credentialed_branch`. Resolving.

---

src/zagg/temporal.py
espg: 🤖 *from Claude*

Folded in 8532b0e, one level down: `_s3_object_store` now deep-copies `retry_config` unconditionally at the construction seam, so the invariant holds for every path a config arrives through — this call site, the read-only branch, and the pre-existing `runner._POLL_RETRY_CONFIG` precedent — without per-caller copies. The redundant deepcopy in `_open_s3_store` was dropped accordingly. Resolving.

---

tests/test_store.py
espg: 🤖 *from Claude*

Both added in 8532b0e: `test_skip_signature_with_credentials_takes_credentialed_branch` pins the precedence (credentialed branch wins, `skip_signature` rides through, provider never constructed), and the anonymous test now asserts the read-only retry policy flows through. The cryocloud `OUTPUT_URL` micro-nit from the summary is folded too. Resolving.

---

Issue #191, opened 2026-07-07
Skip zarr metadata consolidation by default: ~70s/run finalize tax no reader uses


espg:
🤖 *from Claude*

## Summary

The finalize step (`zarr.consolidate_metadata`, run as a dedicated Lambda invoke after every AOI dispatch) costs **~70 s per run** — now ~half of total AOI wall time — and produces a consolidated-metadata blob that **nothing in zagg's read path actually uses**. Recommend skipping it by default (opt-in flag), and if ever kept, parallelizing its metadata reads.

## Measurements (o9 NEON AOI, 4-shard sharded store, 2026-07-07)

- Finalize (`finalize_s`) = **69–73 s**, ~constant across K / memory / cached-vs-uncached / hierarchical-vs-sidecar.
- The store has **4,102 `zarr.json` metadata objects** (≈1,025/shard: 256 ragged/CSR subgroups × their arrays + the main arrays). `consolidate_metadata` reads all 4,102 **serially** (~17 ms/GET) → the ~70 s.
- As workers got ~8× faster this week, finalize went from a rounding error to **~47% of total wall** ([#180 breakdown](https://github.com/englacial/zagg/issues/180#issuecomment-4909625641)). It's cheap in dollars (one short invoke) but dominates the wall a user waits on.

## Why we (probably) don't need it

- **No reader depends on it.** Nothing in zagg opens with `use_consolidated=True` / `open_consolidated`; readers navigate to specific paths (`zarr.open_group(store)["19"]["h_tdigest"][shard]`), ~5 GETs to target, never the full 4,102. The write path already passes `consolidated=False` everywhere (`src/zagg/processing/write.py`).
- **It's optional in the zarr v3 spec** — a non-core extension. v3 readers do lazy metadata reads natively; stores without it read fine.
- The only beneficiary would be a *full-hierarchy browse* (iterate every group at once), which no zagg reader or notebook does.
- **Sharding irony**: sharding cut data-object count, but the ragged/CSR subgroups still emit ~1,025 metadata objects/shard, so this cost scales with (K inner chunks × ragged fields), not shard count — and grows with finer grids.

## Proposal

1. **Skip consolidation by default.** Make the finalize invoke conditional on an opt-in `output.consolidate_metadata: true` (default `false`). Reclaims ~70 s/run (~half the AOI wall) at zero read-side cost for our access pattern.
[...]

PR #192, opened 2026-07-08
Add total AOI wall + finalize to benchmark columns (issue #180)

espg:
Refs #180. Adds the wall-time breakdown to the benchmark record + PR comment table, per the request on the issue thread.

## What

`agg` already returns `wall_time_s` + `setup_s`/`fanout_s`/`finalize_s` in every summary; this surfaces them:

- **`RECORD_COLUMNS`** gains `total_wall_s`, `setup_s`, `fanout_s`, `finalize_s` (appended last, stable-schema rule — old parquet rows reindex to null, no migration).
- **`build_record`** threads them straight from the summary.
- **The rendered table** (PR comment / `latest.md` / PNG) gains two columns: **`wall (s)`** (total end-to-end AOI dispatch) and **`finalize (s)`** (the metadata-consolidation invoke). The existing `runtime (s)` (single-worker compute) is unchanged.

## Why

`runtime (s)` is per-shard worker compute; it badly understates what the orchestrator actually waits. On the o9 AOI, median worker ~56 s but **total wall ~154 s** — the gap is fanout/poll (≈ slowest worker) + a **~70 s fixed finalize** (`consolidate_metadata`) that's now ~half the wall since workers got 8× faster ([breakdown](https://github.com/englacial/zagg/issues/180#issuecomment-4909625641)). Surfacing `wall (s)` + `finalize (s)` makes both trackable run-over-run — and gives the finalize-cost regression (#191) a column to watch.

The CI runs one shard/target, so its `wall (s)` ≈ single-shard worker + finalize; still the right metric to track the finalize component.

## Testing

`tests/test_benchmark.py`: new `test_wall_breakdown_columns_threaded_and_rendered` (record threading, null-safe on older summaries, table cells), and the schema-order test updated for the four appended columns. Full file green; consumers verified (`update_series` reindexes → old rows null; `plot_series` untouched).

### DIFF
--- a/.github/scripts/bench_metrics.py
+++ b/.github/scripts/bench_metrics.py
@@ -63,6 +63,16 @@ RECORD_COLUMNS = [
     # targets, None otherwise. Keeps the cached column out of the inner
     # codec slot in the renderer; legacy rows read back NaN and key on codec.
     "read",
+    # Wall-time breakdown (issue #180): end-to-end AOI dispatch wall and its
+    # three phases. ``runtime_s`` above is the single worker's compute;
+    # ``total_wall_s`` is what the orchestrator actually waits (setup +
+    # fanout/poll + finalize). ``finalize_s`` is the zarr metadata
+    # consolidation invoke -- a ~fixed cost worth tracking for regression
+    # (issue #191). Null on rows recorded before these were surfaced.
+    "total_wall_s",
+    "setup_s",
+    "fanout_s",
+    "finalize_s",
 ]
 
 
@@ -168,6 +178,12 @@ def build_record(
         # the target. None on targets that don't set it (legacy/frozen rows).
         "codec": context.get("codec"),
         "read": context.get("read"),
+        # Wall-time breakdown (issue #180): total end-to-end AOI dispatch wall
+        # plus its phases, straight from the agg summary.
+        "total_wall_s": summary.get("wall_time_s"),
+        "setup_s": summary.get("setup_s"),
+        "fanout_s": summary.get("fanout_s"),
+        "finalize_s": summary.get("finalize_s"),
     }
     return record
 
@@ -188,6 +204,8 @@ TABLE_HEADERS = [
     "target",
     "obs",
     "runtime (s)",
+    "wall (s)",
+    "finalize (s)",
     "cost/shard",
     "cost/100 km²",
     "% timeout",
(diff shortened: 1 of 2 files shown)


espg:
🤖 *from Claude (review)*

Adversarial fresh-context pass on the wall-time columns — **clean, no blocking findings.** Verified each risk end-to-end:

1. **Summary key names match exactly.** `build_record` reads `wall_time_s`/`setup_s`/`fanout_s`/`finalize_s`; the spatial `_run_lambda` summary (`src/zagg/runner.py:1563-1575`) emits those exact keys. No `wall_s`-style typo.
2. **CI populates the columns.** `run_benchmark.py` calls `agg(backend="lambda", morton_cell=…)` on the spatial point-cloud targets → `_run_lambda`, which sets all four. `--dry-run` emits `summary={}` → all four null via `.get()` (by design, same as the other metrics there).
3. **Backward-compat confirmed empirically.** Simulated an old parquet series (without the four columns) + a new record through `update_series.append_records`: old rows reindex to `NaN`, new columns land last, `latest_markdown` and the PNG `_render_table` (10 cols) both render — legacy `NaN` → `n/a`, no crash. `plot_series` is entirely name-keyed (no positional column assumptions); `% cap`/codec masks resolve by name.
4. **`_fmt` handles both `None` and `NaN`** (`bench_metrics.py:192-197`) → `n/a`.
5. **Tests are honest.** `test_wall_breakdown_columns_threaded_and_rendered` exercises threading (210/5/140/65), null-safety on a bare summary, and the rendered cells; the schema-order test's `cols[-4:]` correctly pins append order; `set(rec)==set(RECORD_COLUMNS)` (line 117) and `list(df.columns)==RECORD_COLUMNS` (lines 869/874) stay consistent. No stale hardcoded column-count assertion. Full `tests/test_benchmark.py` green (61 passed).
6. **No golden/schema-drift file** to break — `test_benchmark_shardmap.py` is CMR shard drift, unrelated; column validation is dynamic against `RECORD_COLUMNS`.

One minor, non-blocking accuracy note (PR body, not code): the phase keys `setup_s`/`fanout_s`/`finalize_s` are set **only** on the spatial-lambda path, not the events/temporal-lambda summary (`runner.py:800`) or the two local summaries (`525`, `1220`) — so "agg already returns … in every summary" is slightly overstated. Harmless here (CI is spatial-lambda, and `.get()` nulls are the documented "row omits them" semantics), but worth knowing if `build_record` is ever reused on a temporal summary.

PR #193, opened 2026-07-08
Skip zarr metadata consolidation by default (opt-in)

### LINKED ISSUE #191: Skip zarr metadata consolidation by default: ~70s/run finalize tax no reader uses

espg:
Closes #191

## What / approach

Metadata consolidation (`zarr.consolidate_metadata`, run as a finalize step after every AOI dispatch) costs ~70 s/run and produces a consolidated-metadata blob that nothing in zagg's read path consults — readers navigate to specific paths in a few GETs, and the write path already passes `consolidated=False` everywhere. This makes consolidation **opt-in** via a new `output.consolidate_metadata` flag (default `false`), gated dispatcher-side so the finalize invoke is skipped entirely when off.

- **Config** (`src/zagg/config.py`): added `output.consolidate_metadata` (default `false`) with a boolean validator (mirrors `output.aoi_mask`) and a `get_consolidate_metadata()` accessor (mirrors `get_aoi_mask()`).
- **Local backend** (`src/zagg/runner.py` `_run_local`): the `consolidate_metadata(zarr_store, zarr_format=3)` call now runs only when the flag is true.
- **Lambda backend** (`_run_lambda`): the `mode: "finalize"` invoke is gated dispatcher-side — when the flag is off the executor gets a no-op `finalize_fn` (mirroring the temporal path `_run_lambda_events`, which already skips finalize because tabular output has no metadata to consolidate), so no finalize Lambda is dispatched. When on, `_invoke_lambda_finalize` runs byte-identically to before. The `mode: "finalize"` handler in `deployment/aws/lambda_handler.py` is unchanged — it still consolidates when invoked (which now only happens on the opt-in path), so it needed no edit.

Behavior when the flag **is** true is identical to the pre-change path (still consolidates, both backends).

### Notebooks

The example notebooks read the published `atl06_cycle22_fullsphere.zarr` store with `xr.open_dataset(..., consolidated=True)`. Since that store will be regenerated without consolidation (opt-out, no backward-compat needed), those reads are flipped to `consolidated=False`. `consolidated=False` is a safe v3 read regardless (readers do lazy metadata reads natively). Edited via a surgical JSON round-trip (`indent=1, ensure_ascii=False` — verified identical to the repo's on-disk formatting), so each notebook has a **one-line diff** and cell ids/outputs are untouched.

## Phases

- [x] **Phase 1** — Default-skip gate + `output.consolidate_metadata` config flag (config accessor/validator; local + Lambda dispatcher-side gate).
- [x] **Phase 2** — Notebooks flipped to `consolidated=False` + read-parity test.
- [x] **Phase 3** — Parallelize the opt-in consolidation path. **Out of scope here; tracked in #194** (see Questions for review 2).
- [x] Fold adversarial-review finding ([review](https://github.com/englacial/zagg/pull/193#pullrequestreview-4649930199)): document in the parity fixture that the leaf ShardingCodec is deliberately omitted (orthogonal to the metadata-tree consolidation path).

## How it was tested

- `uv run pytest -q` (excluding two pre-existing `pyarrow`-gated modules — see below): **1443 passed, 20 skipped** in 401 s.
- Touched suites (`test_readers test_config test_runner test_output`): **419 passed**.
- `uv run ruff check` and `uv run ruff format --check` on every touched file: clean.

New tests:
- `tests/test_config.py::TestConsolidateMetadataFlag` — accessor default off, accessor true, validator rejects non-bool.
- `tests/test_runner.py::TestConsolidationGate` — local runner skips `consolidate_metadata` by default and calls it when enabled; Lambda dispatcher skips `_invoke_lambda_finalize` by default and calls it when enabled.
- `tests/test_readers.py::TestReadParityWithoutConsolidation` — builds a two-shard located t-digest CSR store with **no** consolidated metadata (the default now), asserts `read_tensors` / `read_raw_values` / `read_locations` reach the data, then consolidates the same store and asserts every reader returns byte-identical results. Also pins that a freshly written store carries no consolidated-metadata blob (`root.metadata.consolidated_metadata is None`) yet is fully navigable.

## Questions for review

1. **Third notebook beyond the two named in the issue.** `notebooks/rasterized_zarr.ipynb` reads the *same* regenerated store with `consolidated=True`, so it's flipped too (a one-line kwarg change), otherwise it would break against the non-consolidated store. Call out if you'd rather scope this PR to only the two the issue enumerated.
2. **Phase 3 deferred → tracked in #194.** `zarr==3.2.1`'s `consolidate_metadata(store, path, zarr_format)` exposes **no** concurrency parameter; the only lever is the global `zarr.config['async.concurrency']` (default 10), whose effect on the ~4,102-object serial-GET path can't be verified without the production store the issue measured against. Per the issue's guidance ("if it doesn't expose concurrency cleanly, don't fight it — note the finding and defer; do not add a dependency"), it's deferred to #194. The opt-in path is off by default, so it's not a default foot-gun.
3. **Pre-existing lint drift (not fixed, per conventions).** `ruff check src tests` reports 1 pre-existing `N818` in `src/zagg/registry.py` (`UnknownCapability`), and `ruff format --check src tests` would reformat 7 files I did not touch (`src/zagg/__init__.py`, `src/zagg/__main__.py`, `src/zagg/catalog/sources.py`, `src/zagg/viz/crs.py`, `tests/test_catalog.py`, `tests/test_extract.py`, `tests/test_integration.py`). All of my touched files are clean.
4. **Pre-existing test-collection gap.** `tests/test_shardmap.py` and `tests/test_viz.py` fail to collect under `uv sync --extra test` because they `import pyarrow`, which the `test` extra doesn't install (present on `origin/main` too). Excluded from the run above; unrelated to this change.

_Note on the earlier #188 merge-order caveat: PR #188 has since merged to `main`, and this branch is based on `5004df4` (post-merge), so the notebook cells already carry both the `open_store` rewire and `consolidated=False` — no rebase needed._

### DIFF
--- a/tests/test_readers.py
+++ b/tests/test_readers.py
@@ -374,7 +374,13 @@ class TestReadParityWithoutConsolidation:
 
     def _build_store(self):
         """A two-shard located t-digest field, written like the worker's CSR
-        path (per-parent-morton subgroups), with no consolidated metadata."""
+        path (per-parent-morton subgroups), with no consolidated metadata.
+
+        The leaf ShardingCodec is deliberately omitted: ``consolidate_metadata``
+        is a pure metadata-tree op (it gathers each group/array ``zarr.json``),
+        orthogonal to the leaf codec, and the per-morton CSR subgroups are
+        exactly the objects that dominate the finalize cost this PR skips — so
+        the parity claim holds without the sharding codec in the fixture."""
         store = MemoryStore()
         # Two coverage chunks (parent mortons 100 and 205), distinct values so
         # read_raw_values is lossless (every centroid weight 1).
--- a/tests/test_readers.py
+++ b/tests/test_readers.py
@@ -1,6 +1,7 @@
 """Tests for the t-digest → tensor read helpers — issue #79."""
 
 import math
+import warnings
 
 import numpy as np
 import pytest
@@ -356,3 +357,92 @@ class TestReadLocations:
         _write_chunk(store, "f", 7, {0: np.array([1.0, 2.0])})
         with pytest.raises(ValueError, match="has no locations array"):
             list(read_locations(store, "f"))
+
+
+class TestReadParityWithoutConsolidation:
+    """Issue #191: consolidation is now opt-out, so published stores are read
+    without a consolidated-metadata blob. Pin that every reader navigates a
+    non-consolidated store to the same bytes it would read from a consolidated
+    one — readers reach paths directly (``zarr.open_group``/``open_array`` per
+    subgroup), never the consolidated blob."""
+
+    @staticmethod
+    def _point_words(n, seed):
+        from conftest import point_words
+
+        return point_words(n, seed)
+
+    def _build_store(self):
+        """A two-shard located t-digest field, written like the worker's CSR
+        path (per-parent-morton subgroups), with no consolidated metadata."""
+        store = MemoryStore()
+        # Two coverage chunks (parent mortons 100 and 205), distinct values so
+        # read_raw_values is lossless (every centroid weight 1).
+        shards = {
+            100: {0: np.array([10.0, 11.0, 12.0]), 5: np.array([13.0, 14.0])},
+            205: {2: np.array([20.0, 21.0]), 63: np.array([22.0, 23.0, 24.0])},
+        }
+        seed = 1
+        for morton, cell_to_vals in shards.items():
+            cell_ids = sorted(cell_to_vals)
+            pairs = [
+                build_tdigest(
+                    cell_to_vals[c],
+                    delta=512,
+                    locations=self._point_words(len(cell_to_vals[c]), seed),
+                )
+                for c in cell_ids
+            ]
+            write_csr(
+                store,
+                f"h_tdigest/{morton}",
+                [p[0] for p in pairs],
+                cell_ids,
+                locations_list=[p[1] for p in pairs],
+            )
+            seed += 1
+        return store
+
+    @staticmethod
+    def _read_all(store):
+        from zagg.readers.tdigest_tensor import read_locations
+
+        tensors = {m: t for t, m in read_tensors(store, "h_tdigest", n_bins=64, resolution=0.5)}
+        raw = {(m, c): v for m, c, v in read_raw_values(store, "h_tdigest")}
+        locs = {(m, c): loc for m, c, loc in read_locations(store, "h_tdigest")}
+        return tensors, raw, locs
+
+    def test_non_consolidated_is_navigable_and_reads_parity(self):
+        store = self._build_store()
+
+        # A freshly written store carries NO consolidated-metadata blob — the
+        # default now (issue #191). Readers must still navigate it.
+        root = zarr.open_group(store, mode="r", zarr_format=3)
+        assert root.metadata.consolidated_metadata is None
+
+        tensors_plain, raw_plain, locs_plain = self._read_all(store)
+        # Sanity: the readers actually reached the data.
+        assert set(tensors_plain) == {100, 205}
+        assert (100, 0) in raw_plain and (205, 63) in raw_plain
+
+        # Consolidate the SAME store and re-read: consolidation only adds a
+        # metadata blob no reader consults, so every byte must match.
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            zarr.consolidate_metadata(store, zarr_format=3)
+        assert (
+            zarr.open_group(store, mode="r", zarr_format=3).metadata.consolidated_metadata
+            is not None
+        )
+
+        tensors_cons, raw_cons, locs_cons = self._read_all(store)
+
+        assert set(tensors_cons) == set(tensors_plain)
+        for m in tensors_plain:
+            np.testing.assert_array_equal(tensors_plain[m], tensors_cons[m])
+        assert set(raw_cons) == set(raw_plain)
+        for key in raw_plain:
+            np.testing.assert_array_equal(raw_plain[key], raw_cons[key])
+        assert set(locs_cons) == set(locs_plain)
+        for key in locs_plain:
+            np.testing.assert_array_equal(locs_plain[key], locs_cons[key])
--- a/notebooks/cryocloud_example.ipynb
+++ b/notebooks/cryocloud_example.ipynb
@@ -465,7 +465,7 @@
     "# open_store applies zagg's read-only retry policy (issue #186).\n",
     "store = open_store(OUTPUT_URL, read_only=True, skip_signature=True)\n",
     "ds = xr.open_dataset(\n",
-    "    store, engine='zarr', consolidated=True, zarr_format=3,\n",
+    "    store, engine='zarr', consolidated=False, zarr_format=3,\n",
     "    group=str(CHILD_ORDER),\n",
     ")\n",
     "\n",
--- a/notebooks/jupyterhub_example.ipynb
+++ b/notebooks/jupyterhub_example.ipynb
@@ -236,7 +236,7 @@
     "pub = xr.open_dataset(\n",
     "    open_store(f\"s3://{PUBLIC_BUCKET}/{PUBLIC_PREFIX}\", read_only=True, skip_signature=True),\n",
     "    engine=\"zarr\",\n",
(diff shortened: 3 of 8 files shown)


tests/test_readers.py
```diff
@@ -356,3 +357,92 @@ def test_value_only_field_raises_clearly(self):
...
+
+    @staticmethod
+    def _point_words(n, seed):
+        from conftest import point_words
+
+        return point_words(n, seed)
+
+    def _build_store(self):
+        """A two-shard located t-digest field, written like the worker's CSR
+        path (per-parent-morton subgroups), with no consolidated metadata."""
```
espg: 🤖 *from Claude (review)*

note (not blocking): this store is a per-parent-morton CSR **subgroup** store, not a `ShardingCodec`-sharded store, whereas issue #191's acceptance is worded around "a non-consolidated **sharded** store." The good news is that the distinction doesn't weaken the test's claim: `consolidate_metadata` is a pure metadata-tree operation (it gathers the `zarr.json` objects for every group/array), entirely orthogonal to whether the leaf arrays use a sharding codec — and it's precisely the ~1,025 CSR/ragged subgroup `zarr.json` objects per shard (per the issue's measurement) that this store reproduces and that dominate the finalize cost. So the metadata-navigation parity that matters is faithfully exercised. Worth a one-line comment here saying the sharding codec is deliberately omitted as irrelevant to the consolidation path, so a future reader doesn't mistake it for an oversight.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the opt-in-consolidation change. Verdict: **clean** — no must-fix / should-fix. One low-severity inline note on the read-parity test; everything else checked out.

**What I verified (independently, not on faith):**

1. **Lambda gate correctness — clean.** Traced `_finalize_fn` → `LambdaExecutor.finalize()` (`dispatch.py:390`): `finalize()` calls `self._finalize_fn()`, discards its return, and returns a fresh `RunReport()` that the caller also discards. So the no-op path dispatches **no** `mode:"finalize"` invoke, and nothing downstream keys off a real finalize — the `finalize_s` bracket (`runner.py:1524`) still runs (~0 s), and cost/summary accounting reads `report.cost` / per-cell results, never the finalize result. This exactly mirrors the temporal path's `finalize_fn=lambda: None` (`runner.py:722`). Local path: workers already write with `consolidated=False`, so gating the standalone `consolidate_metadata` call still yields a valid store. Default `false` confirmed end-to-end (config accessor → both backends).

2. **Store validity + reader parity — clean.** `git grep` across `src/` + notebooks: no reader opens with `consolidated=True` / `use_consolidated`; readers navigate to paths directly. The parity test asserts `consolidated_metadata is None` on a fresh store, then byte-identical reads after consolidating the same store. See the one inline note: the store is a CSR-subgroup store rather than a literal `ShardingCodec` store — faithful for the consolidation claim (codec-independent), worth a clarifying comment.

3. **Third notebook — correct.** `rasterized_zarr.ipynb` reads the same regenerated published store; one-kwarg diff, valid JSON, `nbformat_minor 4` (no cell-id requirement — pre-existing, untouched). I also checked a **fourth** `consolidated=True` reader, `bench/layout_access.ipynb`: correctly left alone — its stores come from `bench/layout_materialize.py:141`, which always calls `consolidate_metadata` itself, so that pair is self-consistent and unaffected by the config gate.

4. **Config validator — clean.** Exact mirror of `output.aoi_mask` (validator placed adjacent, `bool` accessor with `default=False`). Non-bool rejected and tested. `output` has no allowed-keys whitelist to register the new key in, and no other `output.*` reader needs to honor it.

5. **Test quality — solid.** `TestConsolidationGate` drives the real `_run_local` / `_run_lambda` and counts real `consolidate_metadata` / `_invoke_lambda_finalize` calls (0 off, 1 on) across **both** backends × both flag states — a dropped gate or re-enabled default would fail them; not tautological (both branches genuinely reach `finalize()`).

6. **Conventions — clean.** Title-only commits (`phase 1/2 of issue #191`); Phase 3 deferred honestly with rationale; pre-existing N818/format drift + pyarrow collection gap correctly flagged-not-fixed. `runner.py` is well over ~1000 lines but that's pre-existing (+23 here), not this PR's remit.

**What I ran:** `uv run pytest tests/test_config.py::TestConsolidateMetadataFlag tests/test_runner.py::TestConsolidationGate tests/test_readers.py::TestReadParityWithoutConsolidation -q` → **8 passed**. Notebook JSON validated for all three (nbformat 4, one-line diffs, cell structure intact).

No side-effecting directives raised — the PR body's #188 merge-order note is already moot since the `open_store` rewire is present on `main` (base `5004df4`); whichever the author lands, keep both the `open_store` call and `consolidated=False`.

---

tests/test_readers.py
espg: 🤖 *from Claude*

Folded in fed67b3: `_build_store`'s docstring now states the leaf ShardingCodec is deliberately omitted — `consolidate_metadata` is a pure metadata-tree op orthogonal to the leaf codec, and the per-morton CSR subgroups are exactly the `zarr.json` objects that dominate the finalize cost. Resolving.

---

Issue #213, opened 2026-07-14
Event interchange + catalog for the temporal pipeline (antarctic_AR_dataset integration)


espg:
🤖 *from Claude*

Tracking issue for making [jbbutler/antarctic_AR_dataset](https://github.com/jbbutler/antarctic_AR_dataset) consume zagg's temporal pipeline for its heavy per-storm aggregations. The engine itself landed via #70 (with #73 registries and #76 dispatch); this issue covers the remaining work: the **event interchange/catalog layer** plus a handful of capability gaps found by auditing every `attribute_utils.py` kernel against `zagg.temporal` built-ins. Supersedes the remaining scope of #12. The prior downstream attempt, jbbutler/antarctic_AR_dataset#1, was closed unmerged — its `aggregation_registry.py` design already lives in `src/zagg/temporal.py`, and its catalog/auth ideas are folded into the plan below.

## Attribute coverage audit

All 14 field-based attributes (the ones that stream MERRA-2) expressed as temporal specs, vs. what current `main` supports. The ~10 mask-geometry attributes (max/mean area, duration, start/end dates, southward extent, cumulative space-time, landfall region, trajectory, elevation gradient) need no MERRA-2 reads and **stay client-side in artools** (decided; trajectory is per-timestep and doesn't fit the tabular row model anyway).

| output | variable / collection | spatial_func × reducer | mask | needs |
|---|---|---|---|---|
| max_T2m_ais | T2M / slv | max × max | ais | works today |
| max_T2M_anomaly_ais | T2M / slv | max × max (anomaly) | ais | works today |
| max_IWV_ais | TQV / slv | max × max | ais | works today |
| max_IWV_anomaly_ais | TQV / slv | max × max (anomaly) | ais | works today |
| avg_vIVT_ais | VFLXQV / flx | weighted_mean × weighted_mean (negate) | ais | gap 3 |
| max_vIVT_ais | VFLXQV / flx | max × max (negate) | ais | gap 3 |
| max_vIVT | VFLXQV / flx | max × max (negate) | full | gap 3 |
| min_SLP | SLP / slv | min × first_landfall | ocean | gap 2 |
| max_SLP_gradient | SLP / slv | max_gradient × first_landfall | ocean | gap 2 |
| max_landfalling_v850hPa | V850 / v850 | max × first_landfall (negate) | ocean | gaps 2, 3 |
[...]

PR #214, opened 2026-07-14
Temporal event triggers, reader features, and event contract (issue #213)

espg:
Refs #213 (zagg-side phases; the issue also tracks the downstream antarctic_AR_dataset PR).

Adds the temporal-pipeline capabilities and reader/config features needed for the antarctic_AR_dataset integration, per the design resolution on the issue thread (no third-party code on Lambda workers — everything lands as zagg built-ins or declarative config; see https://github.com/englacial/zagg/issues/213#issuecomment-4965682167).

## Phases

- [x] Phase 1 — `first_intersection` event trigger (aliased `first_landfall`): first timestep where the event mask overlaps the static mask named by `trigger_mask` (default `ais_mask`). `EVENT_TRIGGERS` previously shipped no built-ins, so `trigger:` gating in `process_event` was unreachable from config; without gating, the `first_landfall` *reducer* captures the storm's first timestep rather than first landfall. Never-intersecting events return an empty trigger set, so gated specs finalize to NaN (mirrors the downstream `is_landfalling` filter).
- [x] Phase 2 — `mean_of_ratios` streaming reducer (parity with downstream `compute_average`: per-timestep weighted ratio, unweighted mean over timesteps). Resolves parity question (1) on the issue.
- [x] Phase 3 — declarative reader features: `data_source.collections` mapping form with per-collection options (`variables`, `time_offset`, `resample {freq, how, scale}`, `derived` numpy-expression variables), multi-URI collections concatenated along time, and `data_source.reader` resolved through the registry (worker + `merra2_storm.yaml` example updated). Same semantics applied on the local backend via `prepare_collection`.
- [x] Phase 4 — `gesdisc` credential provider (NSIDC helper generalized to `get_daac_s3_credentials`), both registered in `CREDENTIAL_PROVIDERS`; temporal Lambda fan-out fetches once from `data_source.credentials_provider` and attaches to events lacking per-event `s3_credentials`.
- [x] Phase 5 — event payload contract documented (`docs/temporal_events.md`: event shapes for both backends, collections config, spec keys, capability-resolution policy) + spec `params:` pass-through so registered capabilities receive tuning knobs verbatim.
- [x] Phase 6 — provider-selected credentials unified across both pipelines (maintainer direction on the phase-4 review question): `_resolve_source_credentials` (registry lookup, NSIDC default) now backs the spatial local + lambda paths too; validation hoisted ahead of the pipeline-type branch; `credentials_provider` declared in `DataSourceDict`.

## Testing

Full suite in the worktree venv: 1826 passed, 13 skipped, 1 failed — `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, which fails identically on a clean `main` checkout (environment/build tooling, unrelated). New coverage per phase: trigger unit + `process_event` gating/batching tests; `MeanOfRatiosAccumulator` unit tests + an end-to-end divergence test pinning mean-of-ratios (15.0) apart from pooled weighted-mean (18.0) on the same event; `prepare_collection`/multi-URI/collection-options tests plus registry-resolved-reader tests in the Lambda handler and a local-backend derived-variable run; auth provider registration + DAAC parameterization tests, payload credential-injection precedence tests; `params` pass-through to a registered capability; `TestResolveSourceCredentials` for the unified provider lookup.

## Adversarial review status

Per-phase fresh-context reviews posted inline (🤖 *from Claude (review)*). All findings folded: phase-1 (trigger_mask guard, alignment contract, value pass-through test), phase-3 (parse-validation of `time_offset`/`resample.freq` at config load), phase-4 (TypedDict declaration; the temporal-only scope question resolved as phase 6 by maintainer direction). Phase-2 review returned no findings.

## Questions for review

- Pre-existing, not addressed here (flagging per convention): `ruff check` on full `src` trips N818 on `registry.py:64` (`UnknownCapability`); `ruff format --check` would reformat 7 files untouched by this PR; and the lambda-build test failure above — all present on clean `main`.

### DIFF
--- a/docs/temporal_events.md
+++ b/docs/temporal_events.md
@@ -1,0 +1,129 @@
+# Temporal events: the consumer contract
+
+The temporal pipeline (`pipeline.type: temporal`, issues #12/#213) aggregates
+*events* — spatiotemporal objects like storms — into one tabular row of scalar
+attributes each. This page is the stable interface for libraries that feed
+events into `zagg.agg()`: the event shapes both backends accept, the config
+blocks that describe collections and specs, and the capability-resolution
+policy that governs what runs where.
+
+## The event model
+
+An event is **one binary mask** with dims `(time, lat, lon)`: where and when
+the object exists. `process_event` streams the mask's timesteps, applies a
+per-timestep `spatial_func` under a `mask` provider, and reduces across time
+with a `temporal_reducer` — one scalar per configured attribute.
+
+There is deliberately no second mask channel: an attribute set needing a
+different footprint (e.g. a precipitation lookahead window) is a *different
+event set*. Run `agg()` once per mask family and join the tabular outputs on
+`event_key`.
+
+## Passing events
+
+`zagg.agg(config, events=..., backend=...)` accepts events in two shapes,
+chosen by the backend:
+
+**Local backend** — in-memory tuples, one per event:
+
+```python
+(event_key, event_mask, collections, static_data)
+```
+
+- `event_key`: hashable identifier, echoed into the output row.
+- `event_mask`: `xr.DataArray`, dims `(time, lat, lon)`.
+- `collections`: `{name: xr.Dataset}` — the source fields, already opened.
+  The runner applies the config's per-collection options (below) to each
+  event's collections, so local and Lambda semantics match.
+- `static_data`: `{name: DataArray | Dataset}` — e.g. `ais_mask`,
+  `cell_areas`, `climatology`.
+
+**Lambda backend** — URI dicts, one per event; the worker loads its own
+inputs, keeping the async payload far under Lambda's 256 KB Event cap:
+
+```python
+{
+    "event_key": "storm_00042",
+    "event_mask_uri": "s3://.../masks/storm_00042.nc",
+    "collection_uris": {"merra2_slv": ["s3://.../day1.nc4", "s3://.../day2.nc4"]},
+    "static_uris": {"ais_mask": "s3://.../ais_mask.nc", "cell_areas": "s3://.../areas.nc"},
+    "s3_credentials": {...},   # optional, per-event read credentials
+}
+```
+
+- `event_mask_uri` / `static_uris` values: a single URI each (local path,
+  `s3://`; `.zarr` opens as a store, anything else is fetched as bytes and
+  opened in-memory). A single-variable file becomes a `DataArray`.
+- `collection_uris` values: one URI **or a list** — a multi-granule event
+  (e.g. a storm spanning several MERRA-2 daily files) is opened per-URI,
+  concatenated along `time`, and time-sorted.
+- `s3_credentials`: optional. Events without it receive shared credentials
+  fetched once from the `data_source.credentials_provider` registry name when
+  the config sets one (`nsidc` and `gesdisc` ship as built-ins).
+
+The full worker payload (mode, config, `return_results`, `result_url`) is
+assembled by the runner — consumers supply only the dicts above.
+
+## Describing collections
+
+`data_source.collections` accepts a list of names, or a mapping carrying
+declarative per-collection reader options:
+
+```yaml
+data_source:
+  reader: xarray_s3
+  collections:
+    merra2_slv:                 # no options needed
+    merra2_precip:
+      variables: [PRECCU, PRECLS, PRECSN]      # subset before anything else
+      time_offset: "-30min"                     # shift stamps onto the hour
+      resample: {freq: "3h", how: sum, scale: 3600}  # rates -> totals
+      derived:
+        rainfall: "PRECCU + PRECLS"             # numpy expression
+      doi: "10.5067/Q5GVUVUIVGO7"               # unknown keys pass through
+```
+
+Options apply in the order listed above (`prepare_collection`). `derived`
+expressions evaluate in the same restricted namespace as the spatial
+pipeline's `expression` fields: numpy plus the collection's variables, no
+builtins. Unknown keys (like `doi`) are ignored by the reader — they are
+metadata for catalog tooling. Validated option values fail at config load,
+not per-worker.
+
+## Spec keys
+
+Each `aggregation.variables` entry:
+
+| key | required | meaning |
+|---|---|---|
+| `variable` | yes | variable name in the collection (may be `derived`) |
+| `collection` | yes | collection name |
+| `spatial_func` | yes | per-timestep reduction (registry name) |
+| `temporal_reducer` | yes | cross-timestep accumulator (registry name) |
+| `mask` | no (`ais`) | mask provider: `full` / `ais` / `ocean` / registered |
+| `anomaly` | no | sugar for `transform: monthly_anomaly` |
+| `transform` | no | field-transform name applied per timestep |
+| `negate` | no | negate the field (southward-positive fluxes) |
+| `trigger` | no | event-trigger name gating which timesteps update (e.g. `first_landfall`) |
+| `trigger_mask` | no | static field the trigger tests against (default `ais_mask`) |
+| `params` | no | free-form mapping passed to capabilities verbatim |
+
+## Capability resolution and the worker policy
+
+Every name above resolves through `zagg.registry` at run time. The Lambda
+payload is **pure data** — capability names, numpy expressions, URIs — and a
+name only resolves against what is installed in the worker's layer. There is
+no mechanism for shipping third-party code to the fleet, by design (issue
+#213): on `backend="lambda"`, worker-side capabilities are zagg built-ins and
+declarative config, full stop.
+
+The registries stay open on `backend="local"`: register a custom trigger,
+reducer, mask, transform, or reader (directly via `zagg.registry.register_*`
+or a `zagg.plugins` entry point) and run it on your own machine. To make a
+capability Lambda-eligible, upstream it into zagg as a built-in — prototyping
+locally as a plugin and promoting what proves out is the intended
+contribution funnel.
+
+Credential providers run orchestrator-side, so they carry no such
+restriction; `data_source.credentials_provider` may name a plugin provider on
+either backend.
--- a/docs/temporal_events.md
+++ b/docs/temporal_events.md
@@ -52,8 +52,9 @@ inputs, keeping the async payload far under Lambda's 256 KB Event cap:
 ```
 
 - `event_mask_uri` / `static_uris` values: a single URI each (local path,
-  `s3://`; `.zarr` opens as a store, anything else is fetched as bytes and
-  opened in-memory). A single-variable file becomes a `DataArray`.
+  `s3://`; `.zarr` opens as a store, a local non-`.zarr` path opens directly,
+  and an `s3://` non-`.zarr` object is fetched as bytes and opened
+  in-memory). A single-variable file becomes a `DataArray`.
 - `collection_uris` values: one URI **or a list** — a multi-granule event
   (e.g. a storm spanning several MERRA-2 daily files) is opened per-URI,
   concatenated along `time`, and time-sorted.
@@ -106,7 +107,7 @@ Each `aggregation.variables` entry:
 | `negate` | no | negate the field (southward-positive fluxes) |
 | `trigger` | no | event-trigger name gating which timesteps update (e.g. `first_landfall`) |
 | `trigger_mask` | no | static field the trigger tests against (default `ais_mask`) |
-| `params` | no | free-form mapping passed to capabilities verbatim |
+| `params` | no | free-form mapping threaded into mask providers, field transforms, and event triggers (not spatial_funcs or reducers) |
 
 ## Capability resolution and the worker policy
 
@@ -127,5 +128,6 @@ contribution funnel.
 Credential providers run orchestrator-side, so they carry no such
 restriction; `data_source.credentials_provider` may name a plugin provider on
 either backend — including non-NASA S3-compatible sources. The spatial
-pipeline honors the same key for its source reads, defaulting to `nsidc` when
-the key is absent.
+pipeline honors the same key for its S3-driver source reads, defaulting to
+`nsidc` when the key is absent; the HTTPS driver instead authenticates with an
+EDL bearer token and does not consult credential providers.
--- a/docs/temporal_events.md
+++ b/docs/temporal_events.md
@@ -126,4 +126,6 @@ contribution funnel.
 
 Credential providers run orchestrator-side, so they carry no such
 restriction; `data_source.credentials_provider` may name a plugin provider on
-either backend.
+either backend — including non-NASA S3-compatible sources. The spatial
+pipeline honors the same key for its source reads, defaulting to `nsidc` when
+the key is absent.
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -73,6 +73,11 @@ class DataSourceDict(TypedDict):
     # keys are backend-specific and validated against the backend's declared
     # ``config_keys`` — irrelevant keys are config errors, not ignored.
     index: NotRequired[dict]
+    # Credential-provider registry name for source-data S3 reads (issue #213
+    # Phase 4/6): built-ins ``nsidc``/``gesdisc``; plugins may register others.
+    # Absent → the spatial default (NSIDC); temporal events may also carry
+    # per-event ``s3_credentials``, which win.
+    credentials_provider: NotRequired[str]
 
 
 # Structured-predicate comparison operators (issue #43). ``in``/``not_in`` take a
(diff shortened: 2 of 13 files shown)


src/zagg/temporal.py
```diff
@@ -303,6 +303,48 @@ def monthly_anomaly(var_t, static_data, spec):
...
+
+def first_intersection(event_mask, static_data, spec):
+    """Earliest timestep where the event overlaps a named static mask.
+
+    The mask is ``static_data[spec['trigger_mask']]`` (default ``ais_mask``),
+    so the built-in expresses "first landfall on the ice sheet" without any
+    domain knowledge: the *meaning* of the mask is config. Returns an empty
+    array when the event never intersects it.
+    """
+    static_mask = static_data[spec.get("trigger_mask") or "ais_mask"].astype(bool)
```
espg: 🤖 *from Claude (review)*

**Error quality (should-fix):** a typo'd or missing `trigger_mask` surfaces as a bare `KeyError` from this dict lookup — e.g. `trigger_mask: shelf_maskk` in config gives `KeyError: 'shelf_maskk'` mid-worker, with no mention of the spec, that it's a trigger mask, or which static fields exist. Verified:

```
>>> process_event(..., specs=[{..., "trigger_mask": "shelf_mask"}], static_data={"ais_mask": ais})
KeyError: 'shelf_mask'
```

This is the first spec key that names an arbitrary static field (the `ais`/`ocean` mask providers hardwire `ais_mask`), so config typos are likely. `process_event` already sets the precedent with its `cell_areas` guard (`ValueError: area-weighted spatial funcs ... require a 'cell_areas' static field`), and the registry design doc emphasizes diagnosable misses (`UnknownCapability` carries the available names). Suggest:

```python
name = spec.get("trigger_mask") or "ais_mask"
if name not in static_data:
    raise ValueError(
        f"trigger_mask {name!r} not in static_data; available: {sorted(static_data)}"
    )
```

src/zagg/temporal.py
```diff
@@ -303,6 +303,48 @@ def monthly_anomaly(var_t, static_data, spec):
...
+    space_dims = [d for d in event_mask.dims if d != "time"]
+    hits = (event_mask.where(static_mask, 0) > 0).any(space_dims)
```
espg, line 368: 🤖 *from Claude (review)*

**Alignment requirement undocumented (nit, question):** `DataArray.where(cond, other)` with an explicit `other` aligns with `join="exact"`, so this line raises an opaque `AlignmentError` whenever `static_mask` is not already on the event grid. Verified with xarray in this repo's venv:

```
>>> event_mask.where(full_grid_mask, 0)   # mask on 4x4 superset of the 2x2 event
AlignmentError: cannot align objects with join='exact' ... along these coordinates (dimensions): 'lat'
```

Through `process_event` this is fine because `static_sub` pre-subsets via `.sel(lat=lats, lon=lons)` (I verified the full-grid-mask path end-to-end). But two callers hit it: (a) anyone invoking the trigger directly, per the docstring's "callable" framing; (b) `process_event` itself when the `.sel` in the static-subset loop fails (e.g. float-coord drift between the event grid and the mask grid raises `KeyError`, the *full* mask falls through, and the trigger then dies with the `AlignmentError` above rather than a message naming the mask).

Options, cheapest first: (1) add one docstring line stating the mask must share the event's lat/lon index (matching `mask_ais`'s implicit contract); (2) also subset defensively here with `static_mask.sel(lat=event_mask["lat"].values, lon=event_mask["lon"].values)` inside a try, falling back to the current path. (1) seems sufficient for phase 1 since the hazard is pre-existing for the mask providers too.

tests/test_temporal.py
```diff
@@ -218,6 +218,7 @@ def test_keys_present(self):
             "negate",
             "transform",
             "trigger",
+            "trigger_mask",
```
espg, line 259: 🤖 *from Claude (review)*

**Test gap (nit):** `trigger_mask` pass-through in `specs_from_config` is only covered by this key-presence assertion, and the `_temporal_config()` fixture never sets it — so its value is always `None` in the config-bridge tests. A wrong value source (e.g. `"trigger_mask": meta.get("trigger")`) would still pass this test and every trigger test (those call `first_intersection`/`process_event` with hand-built specs, bypassing the config bridge). A one-line value assertion pins it, e.g. add `"trigger": "first_landfall", "trigger_mask": "shelf_mask"` to one fixture variable and assert `specs["..."]["trigger_mask"] == "shelf_mask"` in `test_flag_defaults_and_overrides`.

src/zagg/temporal.py
espg: 🤖 *from Claude*

Addressed in 24d2d95: `first_intersection` now raises `ValueError` naming the missing field and listing `sorted(static_data)`, mirroring the `cell_areas` guard; covered by `test_missing_static_field_raises`.

src/zagg/temporal.py
espg: 🤖 *from Claude*

Went with the docstring option in 24d2d95 (matches the pre-existing mask-provider contract): the static mask must already share the event grid; `process_event` guarantees this via its pre-subset, and direct callers now have the contract stated where they'll read it. No defensive subset added — it would mask genuine coordinate drift.

tests/test_temporal.py
espg: 🤖 *from Claude*

Addressed in 24d2d95: `_temporal_config` now sets `trigger: first_intersection` / `trigger_mask: grounded_mask` on one variable and `test_flag_defaults_and_overrides` asserts the configured *values* survive `specs_from_config` (plus `None` defaults on an unconfigured spec).

src/zagg/config.py
```diff
@@ -559,6 +559,94 @@ def _validate_temporal_config(config: PipelineConfig) -> None:
...
+    ``time_offset`` (a pandas-parseable offset string), ``resample``
+    (``{freq, how: sum|mean, scale}``), and ``derived`` (name -> numpy
+    expression string). Unknown option keys (e.g. ``doi``) pass through for
+    catalog tooling. Fails at load so a config typo doesn't surface as a
+    per-event worker error after the Lambda spend.
+    """
+    colls = (config.data_source or {}).get("collections")
+    if colls is None or isinstance(colls, list):
+        return
+    if not isinstance(colls, dict):
```
espg, line 603: 🤖 *from Claude (review)*

This docstring promises "Fails at load so a config typo doesn't surface as a per-event worker error after the Lambda spend," but the validation only checks the *type* of `time_offset`/`resample.freq`, not their parseability — so the exact failure it claims to prevent still slips through to the worker.

Verified against this diff:

```
# both pass validate_config() at load:
time_offset: "gibberish"          # -> pd.to_timedelta("gibberish") ValueError in prepare_collection
resample: {freq: "notafreq"}      # -> ds.resample(time="notafreq")  ValueError in prepare_collection
```

Confirmed both raise only inside `prepare_collection` at worker runtime (i.e. after the Lambda spend), not at load. `resample.freq` is additionally never type-checked (only presence is), so a non-string freq also reaches runtime.

A related silent-failure mode from the "unknown keys pass through" rule: a typo'd *known* option like `time_offest: "-30min"` passes validation and is silently ignored by `prepare_collection` — the FLX half-hour shift just never happens, producing a wrong-but-error-free result rather than a load-time failure.

Options:
1. Validate parseability at load — `pd.to_timedelta(offset)` and `to_offset(freq)` inside `_validate_collection_options`, surfacing the error with the collection name. Closes the offset/freq hole (not the typo'd-key hole).
2. Also warn/reject unknown keys under a known namespace to catch `time_offest` — but this conflicts with the intentional `doi` passthrough, so likely a soft warning at most.
3. Leave as-is and soften the docstring claim to "validates option shape" rather than implying it catches all config typos.

espg: 🤖 *from Claude*

Implemented option (parse-validate at load) in `_validate_collection_options`:

- `time_offset`: after the `isinstance(str)` check, it now parses via `pd.to_timedelta(offset)` in a try/except and re-raises as `ValueError` naming the collection and bad value. So `time_offset: "gibberish"` fails at load instead of on the worker.
- `resample.freq`: now required to be a string and parsed via `pd.tseries.frequencies.to_offset(freq)` in a try/except → `ValueError` naming collection and value. Catches the previously-unchecked type and `{freq: "notafreq"}`.
- pandas is imported lazily inside the function (matching `temporal.py`), so `config.py` stays importable without pandas at module import.
- Unknown option keys (e.g. `doi`) still pass through by design for catalog tooling metadata — no strict key whitelist added — and the docstring now states both the genuine fail-at-load behavior for the validated keys and the intentional pass-through for unknown keys.

Tests added to `TestCollectionOptions`: unparseable `time_offset`, non-string `resample.freq`, and unparseable `resample.freq`; the existing mapping-form test already exercises valid `"-30min"` / `{freq: "3h"}`. `ruff check`, `ruff format`, and `pytest tests/test_config.py` all green (235 passed). Commit will follow shortly.

src/zagg/config.py
```diff
@@ -559,6 +559,100 @@ def _validate_temporal_config(config: PipelineConfig) -> None:
                 f"temporal variable '{name}' is missing required key(s): "
                 f"{', '.join(missing)} (need {', '.join(_TEMPORAL_SPEC_KEYS)})"
             )
+    _validate_collection_options(config)
+    provider = (config.data_source or {}).get("credentials_provider")
```

Low (type/docs consistency): phase 4 adds `credentials_provider` as a recognized `data_source` key, but the `DataSourceDict` TypedDict that enumerates `data_source` keys (config.py:56) is not updated to declare it — every other optional key there (`quality_filter`, `filters`, `levels`, `base_level`, `index`) is listed with `NotRequired`. Consider adding `credentials_provider: NotRequired[str]` so the key is documented alongside the others and stays consistent. Not a runtime bug (configs load from plain dicts, so mypy never checks a literal here), purely consistency.

Low (type/docs consistency): phase 4 adds `credentials_provider` as a recognized `data_source` key, but the `DataSourceDict` TypedDict that enumerates `data_source` keys (config.py:56) is not updated to declare it — every other optional key there (`quality_filter`, `filters`, `levels`, `base_level`, `index`) is listed with `NotRequired`. Consider adding `credentials_provider: NotRequired[str]` so the key is documented alongside the others. Not a runtime bug (configs load from plain dicts, so mypy never checks a literal here), purely consistency.

src/zagg/runner.py
```diff
@@ -640,6 +653,20 @@ def _run_lambda_events(
     if function_name is None:
         function_name = os.environ.get("ZAGG_LAMBDA_FUNCTION_NAME", "process-shard")
 
+    # Orchestrator-side credential fetch (issue #213, Phase 4): a config-named
+    # provider is resolved through the registry and called ONCE (DAAC creds
+    # last ~1 h, far past the fan-out), then attached to every event that does
+    # not carry its own per-event s3_credentials.
+    provider_name = (config.data_source or {}).get("credentials_provider")
```
espg, line 660: 🤖 *from Claude (review)*

Question (scope boundary): the config-selected `credentials_provider` is honored only on the temporal lambda path (here) and validated only in `_validate_temporal_config`. The spatial lambda path still hardwires `get_nsidc_s3_credentials()` (runner.py:1213 and 1446). So a *spatial* config that sets `data_source.credentials_provider: gesdisc` is silently ignored — NSIDC creds are used and no validation fires, which is a quiet footgun. Is temporal-only the intended boundary for this phase?

Options: (a) leave as-is but note "temporal-only" on the config key / docstring so the silent-ignore is not surprising; (b) reject the key on a spatial config so it fails loudly; (c) route the spatial path through the same registry lookup in a follow-up. If this phase is deliberately temporal-scoped, (a) or (b) seem like the low-effort ways to avoid the silent divergence.

src/zagg/config.py

Addressed in f060331: `credentials_provider` is now declared in `DataSourceDict` as `NotRequired[str]` with a comment covering the built-ins, plugin providers, and the default/precedence rules.

src/zagg/runner.py
espg: 🤖 *from Claude*

Resolved as option (c) by maintainer direction (robust provider support across both pipelines, any provider incl. non-NASA): phase 6 (f060331) adds `_resolve_source_credentials` — registry lookup with the historical NSIDC default when the key is absent — and routes both spatial call sites (`_run_local`, `_run_lambda`) through it. The string-shape validation moved ahead of the pipeline-type branch in `validate_config`, so it now fires for spatial configs too. Temporal keeps its no-default injection semantics (per-event `s3_credentials` win). Covered by `TestResolveSourceCredentials`.

docs/temporal_events.md
```diff
@@ -0,0 +1,131 @@
...
+| `collection` | yes | collection name |
+| `spatial_func` | yes | per-timestep reduction (registry name) |
+| `temporal_reducer` | yes | cross-timestep accumulator (registry name) |
+| `mask` | no (`ais`) | mask provider: `full` / `ais` / `ocean` / registered |
+| `anomaly` | no | sugar for `transform: monthly_anomaly` |
+| `transform` | no | field-transform name applied per timestep |
+| `negate` | no | negate the field (southward-positive fluxes) |
+| `trigger` | no | event-trigger name gating which timesteps update (e.g. `first_landfall`) |
+| `trigger_mask` | no | static field the trigger tests against (default `ais_mask`) |
+| `params` | no | free-form mapping passed to capabilities verbatim |
```
espg: 🤖 *from Claude (review)*

Low (contract precision): the table says `params` is "passed to capabilities verbatim", but `process_event` only hands the spec dict to mask/transform/trigger providers. `spatial_func` is called `(var_t, mask_t, cell_areas)` and the reducer `()` (temporal.py:477 and :518), so neither receives the spec — a custom `spatial_func` or `temporal_reducer` cannot read `spec["params"]`. The `specs_from_config` docstring already scopes this correctly ("masks/transforms/triggers read their tuning knobs from `spec['params']`"). Mirror that scoping in the table so the consumer contract does not over-promise? Options: (a) reword to "passed verbatim to mask/transform/trigger providers"; (b) leave generic if a later change threads the spec into spatial_func/reducer.

docs/temporal_events.md
```diff
@@ -0,0 +1,131 @@
...
+    "event_key": "storm_00042",
+    "event_mask_uri": "s3://.../masks/storm_00042.nc",
+    "collection_uris": {"merra2_slv": ["s3://.../day1.nc4", "s3://.../day2.nc4"]},
+    "static_uris": {"ais_mask": "s3://.../ais_mask.nc", "cell_areas": "s3://.../areas.nc"},
+    "s3_credentials": {...},   # optional, per-event read credentials
+}
+```
+
+- `event_mask_uri` / `static_uris` values: a single URI each (local path,
+  `s3://`; `.zarr` opens as a store, anything else is fetched as bytes and
```
espg: 🤖 *from Claude (review)*

Low (doc accuracy): "anything else is fetched as bytes and opened in-memory" holds only for `s3://` non-`.zarr` URIs. For a **local** non-`.zarr` path — which this same parenthetical lists as supported — `open_dataset` hits the `if not uri.startswith("s3://"): return xr.open_dataset(uri)` branch (temporal.py:613-614) and opens the file directly rather than fetching bytes. Suggest qualifying the byte-fetch to the `s3://` case.

docs/temporal_events.md
```diff
@@ -0,0 +1,131 @@
...
+reducer, mask, transform, or reader (directly via `zagg.registry.register_*`
+or a `zagg.plugins` entry point) and run it on your own machine. To make a
+capability Lambda-eligible, upstream it into zagg as a built-in — prototyping
+locally as a plugin and promoting what proves out is the intended
+contribution funnel.
+
+Credential providers run orchestrator-side, so they carry no such
+restriction; `data_source.credentials_provider` may name a plugin provider on
+either backend — including non-NASA S3-compatible sources. The spatial
+pipeline honors the same key for its source reads, defaulting to `nsidc` when
```
espg: 🤖 *from Claude (review)*

Low: this states the spatial pipeline "honors the same key for its source reads" unconditionally, but `_run_local` only routes through `_resolve_source_credentials(config)` on the non-`https` branch — with `driver: https` it uses `{"edl_token": get_edl_token()}` and `credentials_provider` is silently ignored (runner.py:1228-1231; `_run_lambda` has no such branch). Since https reads use an EDL bearer token rather than S3 creds this may be intentional, but a config setting both `credentials_provider` and `driver: https` gets no signal the provider is a no-op. Intended? If so, a one-line caveat here would help; if not, a load-time warning when both are set. (Flagging because the newly-hoisted validation now accepts `credentials_provider` on spatial configs, so this combination is newly reachable.)

docs/temporal_events.md
espg: 🤖 *from Claude*

Fixed the `params` table row. It now reads: "free-form mapping threaded into mask providers, field transforms, and event triggers (not spatial_funcs or reducers)". Verified against `process_event`: the spec dict is passed to `get_mask_provider(...)(storm_t, static_sub, spec)`, `get_field_transform(...)(var_t, static_sub, spec)`, and `get_event_trigger(...)(event_mask, static_sub, spec)`, but reducers are constructed with no args (`get_reducer(...)()`) and spatial_funcs are called `(var_t, mask_t, cell_areas)`. Doc-only fix; commit to follow.

docs/temporal_events.md
espg: 🤖 *from Claude*

Fixed. The URI sentence now distinguishes both non-`.zarr` cases: "a local non-`.zarr` path opens directly, and an `s3://` non-`.zarr` object is fetched as bytes and opened in-memory". Verified against `open_dataset` in `src/zagg/temporal.py`: `.zarr` → `open_store`; `not uri.startswith("s3://")` → `xr.open_dataset(uri)` directly; only the `s3://` non-`.zarr` path does the obstore byte-fetch into `BytesIO`. Doc-only fix; commit to follow.

docs/temporal_events.md
espg: 🤖 *from Claude*

Decision: code unchanged (EDL bearer tokens are a distinct credential type), doc made precise. The credentials paragraph now reads: the spatial pipeline honors the key for its S3-driver source reads (default `nsidc`); "the HTTPS driver instead authenticates with an EDL bearer token and does not consult credential providers." Verified against `_run_local` in `src/zagg/runner.py`: `if driver == "https": s3_creds = {"edl_token": get_edl_token()}` else `_resolve_source_credentials(config)` (which consults `credentials_provider`, default `nsidc`). Doc-only fix; commit to follow.

Issue #218, opened 2026-07-14
Sentinel-2 ingest: generic STAC source + pull-NN raster sampling onto HEALPix grids


espg:
🤖 *from Claude*

## Summary

Add Sentinel-2 L2A ingest as a template + example: a generic STAC catalog source (Earth Search) and a **raster reader that samples COGs at HEALPix cell centers** (pull-based nearest-neighbor), emitting onto the same morton grids the ICESat-2 products use. Goal is datafusion: ICESat-2 ATL03/ATL06 configs already aggregate at child order 19 (~12.4 m), so Sentinel-2 stored at the same child order makes coincident extraction a direct cell-ID match — and provides ancillary values in cells with no ICESat-2 coverage (sharpening).

This is the first dense-raster (vs. sparse point) dataset in zagg; the design decisions below were ratified in the planning discussion (2026-07-14).

## Access mechanics (verified live 2026-07-14)

- NASA CMR does **not** carry raw Sentinel-2 (closest is HLS S30 at 30 m — too coarse), so this bypasses `earthaccess`/CMR-STAC.
- **Earth Search STAC** (`https://earth-search.aws.element84.com/v1`, Element 84, no auth): use collections **`sentinel-2-c1-l2a`** + **`sentinel-2-pre-c1-l2a`** (same item/asset schema; the pre-c1 collection fills the archive where ESA Collection-1 reprocessing is incomplete). ~5 h item latency observed.
- Assets are COGs in **`s3://e84-earth-search-sentinel-data`, us-west-2, anonymous** (verified: unsigned Range GET → HTTP 206). Same region as Earthdata Cloud. Internals (gdalinfo): 1024×1024 DEFLATE blocks (predictor 2), uint16, nodata=0, AVERAGE overviews; 10 m bands ≈ 210–245 MB/tile.
- Items carry `proj:transform`/`proj:shape`/`proj:epsg`/`raster:bands`/`eo:cloud_cover`/`grid:code` per asset — windows are plannable without touching file headers.
- **Radiometry**: c1 stores raw ESA DNs; reflectance = DN×1e-4 **− 0.1**, uniformly (embedded in GeoTIFF tags and STAC, mutually consistent). The *old* `sentinel-2-l2a` collection (`s3://sentinel-cogs`) pre-subtracted the offset but its STAC `raster:bands` still advertises `offset: -0.1` (wrong by exactly 0.1 if applied) — one more reason to standardize on c1. Verified at pixel level against both buckets.
[...]

PR #219, opened 2026-07-14
Sentinel-2 ingest: STAC source, pull-NN raster reader, template + fusion example (issue #218)

espg:
Refs #218 (closes when all phases land).

Sentinel-2 L2A ingest as a template + example: a generic STAC catalog source (Earth Search) and a raster reader that pull-NN samples COGs at HEALPix cell centers, writing morton-aligned Zarr in the same cell space as the ICESat-2 products (child order 19 → cell-ID joins for datafusion). All design decisions were ratified on #218 — see the issue body for the verified access mechanics, the pull-vs-push measurements, and the resolved questions (per-acquisition `(time, cell)` layout; nearest-tile-center cross-zone ownership; `grid.sample()` as a grid-level primitive; async-tiff as the single decode engine, fleet-ready from day 1).

## Phases

- [x] **Phase 1 — generic STAC source + per-band assets** (`catalog/sources.py`, `catalog/shardmap.py`)
- [x] **Phase 2 — `grid.sample()` primitive + raster reader built-in** (async-tiff engine; adds the dependency per the §4 discussion on #218)
- [x] **Phase 3 — config surface + MGRS ownership** (`(time, cell)` layout; same-zone clip + nearest-tile-center cross-zone)
- [x] **Phase 4 — template config** (`sentinel2_l2a.yaml`; uint16 DN + scale/offset attrs; o19-vs-o20 compressed-size measurement)
- [x] **Phase 5 — fusion example notebook** (binder-runnable; S2 × ICESat-2 same-order join)

## Phase 1: what changed

- **`STACSource`** (`src/zagg/catalog/sources.py`): generic STAC API client — configurable root, POST item-search over multiple collections, optional `eo:cloud_cover` filter (query extension), optional asset-key keep-list (`assets=["red", "nir", "scl"]`). Generic APIs use stable per-collection asset keys, so unlike CMR there is no canonical-key normalization — `_subset_assets` only prunes (and raises loudly if an item has none of the requested keys).
- **Shared pager `_page_search`**: the `rel=next` loop previously private to `CMRSource._search`, now handling GET-next, POST-next, and the STAC API `merge` form (Earth Search's stac-server returns POST next links with a body to merge). `CMRSource` behavior is unchanged — its `_search` is now a thin wrapper.
- **`Catalog.granule_records()`**: multi-asset items (no canonical `data`/`data_s3`) now carry `assets` (`{key: href}`) and `datetime` (ISO acquisition time, needed for the `(time, cell)` layout in Phase 3). Canonical single-asset records keep their exact prior shape — the existing `test_option_c_self_contained` byte-shape pin still passes, so **existing CMR shardmaps are unchanged**.
- **`ShardMap` granule entries** (`catalog/shardmap.py`): entry construction factored into `_granule_entry`, which forwards `assets`/`datetime` when present. Both manifest forms round-trip them (JSON natively; parquet stores granules as JSON strings).

## Testing

New `tests/test_sources.py` (19 tests, no network — scripted `requests` fake): pager GET/POST/merge/termination cases, `STACSource.fetch` body construction + provenance, cloud-cover filter, asset subsetting (incl. no-match raise and non-mutation), `granule_records` multi-asset + datetime + CMR-shape regression, geoparquet round-trip, `_granule_entry` shape pins, ShardMap JSON + parquet round-trips with per-band assets.

Local: `pytest tests/test_sources.py tests/test_shardmap.py tests/test_catalog.py tests/test_benchmark_shardmap.py tests/test_viz.py` green (105 passed); `ruff check` + `ruff format --check` clean on touched files.

## Questions for review

- **Pre-existing lint failure, not addressed here**: `ruff check src tests` with the repo's pyproject config (`select = [E, F, W, I, N]`) fails on `src/zagg/registry.py:64` — `N818 Exception name UnknownCapability should be named with an Error suffix`. Present on the branch base (verified via stash); flagging per the don't-fix-unrelated rule. Renaming is a public-API change, so leaving the call to a human.

## Phase 2: what changed

- **`grid.sample()`** on the `OutputGrid` protocol + both grids, backed by shared `sample_nearest` in `grids/base.py`: cell centers → raster CRS (pyproj) → general 2×2 affine inversion (rotated/skewed transforms included; 6- or 9-element STAC `proj:transform`) → floor to containing pixel + bounds mask. HEALPix gets `cell_centers()` (`mort2geo`); rectilinear reuses its existing one.
- **`processing/raster.py`**: the async-tiff read path. Georeferencing comes from the GeoTIFF IFD itself (geo keys + pixel scale/tiepoint — no STAC `proj:*` carriage through shardmaps); `sample_asset` fetches **only** the tiles the sampled pixels touch (concurrent fan-out), decodes, and gathers one value per cell. Store resolution handles `s3://`, virtual-hosted S3 HTTPS (Earth Search hrefs), plain HTTPS, and local paths; anonymous by default. Nodata masking stays with the caller (Phase 3).
- **Dependency (espg-ratified on #218 §4)**: `async-tiff>=0.7.2` in core deps, `==0.7.2` pinned in the `lambda` extra (~4 MB abi3 wheels per arch, `obspec`-only transitive).

Testing: `tests/test_raster.py` (22 tests) writes synthetic tiled GeoTIFFs with a minimal in-test writer (pure struct/numpy — no GDAL, no network) and reads them back through async-tiff LocalStore: affine math incl. rotated/9-element/degenerate cases; **dense-by-construction property at the hostile sites** (equator + base-cell corner, the ones with measured push-hash holes on #218); multi-tile exact value round-trip (pins tile x/y orientation); OOB fill; empty-valid dtype path; HEALPix end-to-end vs manual projection. Full suite: 1864 passed / 22 skipped.

## Phase 3: what changed

The raster pipeline core, per the [time-layout resolution](https://github.com/englacial/zagg/issues/218#issuecomment-4974953050) — true `(time, cells)` arrays via a lean raster-specific template + slab writer that **bypasses** (never modifies) the aggregation write machinery:

- **Config surface** (`config.py`): `reader: raster` validation branch — `data_source.bands` (field → `{asset, dtype, fill_value, scale, offset}`; dtype required, DNs stored exact with CF `scale_factor`/`add_offset` as attrs), `nodata`, no aggregation section (pull-NN has nothing to reduce). Sharded output and rectilinear `(time, y, x)` templates are explicit not-yet errors.
- **Acquisition grouping** (`sources.py`/`shardmap.py`): `STACSource(time_key=...)` records the grouping property (`s2:datatake_id` for S2) in catalog metadata; `granule_records()`/`_granule_entry` carry it per raster record. `raster_time_index()` builds the global timestep index (group key → t, coordinate = earliest member datetime) — adjacent MGRS tiles of one datatake are one timestep, not two.
- **Worker path** (`processing/raster.py`): `sample_item_async` (all configured bands of one item, concurrent, nodata-combined validity, returns the tile center) and `process_raster_shard` (per timestep: sample covering items → **nearest-tile-center ownership** on unit-sphere chord distance where items overlap — comparable across UTM zones, unlike per-zone projected distances). One dense slab per timestep × shard, fill where no valid source.
- **Lean template + writer**: `raster_group_spec`/`emit_raster_template` — per band `(T, n_cells)` arrays, chunks `(1, cells_per_chunk)` (one object per timestep-chunk → per-date rewrites exact), zstd(3), real `time` coordinate (int64 µs, CF attrs), `cell_ids` coordinate; `write_raster_slab` is a chunk-aligned region assignment, `write_raster_coords` once per shard. `grid.cell_lonlat()` added to both grids (grid-agnostic WGS84 centers for the ownership rule).

Testing: `tests/test_raster_pipeline.py` (18 tests): validation matrix; datatake grouping (adjacent tiles seconds apart → one timestep, earliest-datetime coordinate); multi-band sampling + nodata; **ownership property test** (two overlapping synthetic tiles, values split exactly at the center midline); end-to-end healpix — template → process → slab writes → read-back of two timesteps with exact DN values, fill outside footprint, time and cell_ids round-trips.

Deferred to Phase 4 (wiring): runner orchestration for `reader: raster` (template emission + resize-on-append + shard fan-out), Lambda handler branch, `sentinel2_l2a.yaml`, and the o19-vs-o20 compressed-size measurement.

## Phase 4: what changed

- **`RasterStrategy`** in `runner.py`, registered in `_STRATEGIES` and selected when a spatial-kind config declares `reader: raster`: loads the ShardMap, builds the grid, computes the global datatake time index, emits the `(time, cells)` template (runner owns template + future resize-on-append per the single-writer decision), and fans shards out on a thread pool — `process_raster_shard` → slab + coord writes per shard, per-shard error isolation. `backend="lambda"` raises `NotImplementedError` for now: the handler branch lives under `deployment/aws/`, which repo conventions keep out of scope unless named — flagged under Questions for review.
- **`src/zagg/configs/sentinel2_l2a.yaml`**: the shipped template — red/green/blue/nir (uint16, scale 1e-4 / offset −0.1 as CF attrs) + scl (uint8, categorical), nodata 0, anonymous, parent 11 / child 19, with the catalog/shardmap recipe in the header comment.
- **Measurement + first live validation** ([issue comment](https://github.com/englacial/zagg/issues/218#issuecomment-4975088224)): against the production c1 bucket, 8 real order-11 shards, 2.6M cells, 100% valid. **Compressed o20/o19 = 3.56×** (raw 4×) — the duplication compresses far less than hoped, so the o19 default stands with the trade now quantified.

Testing: `tests/test_raster_runner.py` — `agg(backend="local")` end-to-end over a real ShardMap manifest (summary counts, exact DN read-back through `_shard_cell_range`, CF attrs, ascending time coordinate), dry-run path, lambda-not-yet guard, and shipped-template load+validate.

## Questions for review

- (carried) Pre-existing `ruff check` N818 on `src/zagg/registry.py` under the pyproject config — not touched here.
- ~~Lambda handler branch scope~~ — **confirmed by espg** ("scope confirmed, we want this functional on lambda"); landed as Phase 6.

## Phase 5: what changed

**`notebooks/sentinel2_fusion.ipynb`** — Binder-runnable (badge included), every byte anonymous: Earth Search STAC metadata, the public c1 COG bucket, and the source.coop ATL06 cycle-22 store. Flow: catalog (`STACSource`, datatake `time_key`) → `ShardMap.build` over a Larsen B AOI → `agg(backend="local")` ingest of 2 austral-summer datatakes (45 shards, ~100 s locally, 0 errors, 2.9M order-19 cells) → read the `(time, cells)` product → **fusion**: order-19 cells roll up to their order-12 ATL06 parents by integer division, and 82 parents come back with coincident cycle-22 heights — joined with zero geometry at read time. Validated by executing the full cell sequence against the live endpoints before commit.

Note: the intended same-order (19↔19) direct-join target, `neon_tdigest.zarr`/`neon_gainbias.zarr` on source.coop, are currently **empty templates** (metadata only, no data chunks) — the notebook therefore demonstrates the cross-order join against the populated ATL06 cycle-22 store and mentions the direct join in prose. If the NEON stores get populated, a same-order section can be added verbatim.

- [x] **Phase 6 — Lambda fleet execution** (espg-confirmed scope)

## Phase 6: what changed

- **`deployment/aws/lambda_handler.py`**: new `mode="process_raster"` branch (`_handle_process_raster`) routed through the per-unit envelope path (container telemetry + `result_url` mirroring apply). Event: shard_key + the shard's ShardMap entries + config dict + store_path + the shard's slice of the time index. No `s3_credentials` block — the raster source is anonymous; output uses `output_credentials` or the execution role like every other mode. Dense layout is a 400 (its block indexing needs populated-shard state a lone worker can't reconstruct; fullsphere — the default — keys by nested id).
- **`RasterStrategy` lambda backend** (`runner.py`): the runner still owns the template + global time index (single-writer design), then fans out one **synchronous** invoke per shard with transient-only retries, per-shard isolation, and the same all-error raise as local. Deliberately the simple transport for the first functional cut: the issue-151 async result-object channel and the preflight concurrency probe are follow-ups (raster shards are seconds of COG windows, well inside NAT idle limits — noted in the strategy docstring).

Testing: `tests/test_lambda_raster_handler.py` (4 tests — real handler loaded by path: end-to-end slab write through `lambda_handler` with exact DN + cell_ids readback, missing-params 400, dense-layout 400, worker-failure 500) and `TestRasterLambdaBackend` in `tests/test_raster_runner.py` (scripted boto3 client: event structure incl. per-shard time-index slice, summary accumulation, all-shards-error raise). Full runner + handler suites: 275 passed.


src/zagg/catalog/sources.py
```diff
@@ -259,30 +397,52 @@ def granule_records(self) -> list[dict]:
...
                 "https": data.get("href"),
                 "s3": data_s3.get("href"),
                 "lats": np.asarray(y),
                 "lons": np.asarray(x),
-            })
+            }
+            extra = {
+                k: (a or {}).get("href")
+                for k, a in (asset_map or {}).items()
+                if k not in ("data", "data_s3", "metadata") and (a or {}).get("href")
```
espg: 🤖 *from Claude (review)*

**Breaks the "existing CMR shardmaps are byte-identical" invariant for `preserve_thumbnails=True` catalogs.** This `extra` filter only excludes `data`/`data_s3`/`metadata`, but `_normalize_assets(..., preserve_thumbnails=True)` keeps `thumbnail_*`/`browse` assets under their own keys. Those fall through here, so a CMR record grown a `thumbnail`/`browse` entry gets `rec["assets"]` **and** `rec["datetime"]` — a shape change vs. pre-#218.

Repro (normalized keys `data`, `data_s3`, `browse`):
```
record keys: ['assets', 'datetime', 'https', 'id', 'lats', 'lons', 's3']
has assets? True {'browse': 'https://x/thumb.jpg'}
```
The default (`preserve_thumbnails=False`) path is unaffected and `test_h5_record_shape_unchanged` only pins that path, so the regression is untested. Either add `thumbnail`/`browse` to the excluded set (they are not per-band data), or key the multi-asset branch off the source kind rather than "any non-canonical asset present".

src/zagg/catalog/sources.py
```diff
@@ -102,6 +136,56 @@ def _normalize_assets(item: dict, *, preserve_thumbnails: bool) -> dict:
...
 
 
+def _subset_assets(item: dict, keep: list[str]) -> dict:
+    """Keep only the ``keep`` asset keys (stable per-collection keys, #218).
+
+    Raises if an item has none of the requested keys -- a silent empty asset
+    map would surface much later as an unreadable granule.
+    """
+    have = item.get("assets", {})
+    out = {k: have[k] for k in keep if k in have}
```
espg: 🤖 *from Claude (review)*

**Partial-missing asset set passes silently — inconsistent with the stated "raise loudly" contract.** The guard only fires when an item has *none* of the requested keys. An item missing *some* (but not all) requested bands keeps a partial map and sails through:
```
requested red,nir,scl; kept: ['nir', 'red']
```
For the raster reader (Phase 2), a granule silently missing a band is exactly the "unreadable granule surfacing much later" this function claims to prevent. If all keep-list keys are meant to be mandatory, check `set(keep) - set(have)` and raise on any missing; if partial is intentionally allowed, the docstring should say so. Untested either way — `test_no_match_raises` only covers the all-missing case.

src/zagg/catalog/sources.py
```diff
@@ -168,33 +251,88 @@ def fetch(
...
+            "bbox": list(bbox),
+            "datetime": datetime,
+            "limit": limit,
+        }
+        if query.max_cloud_cover is not None:
+            body["query"] = {"eo:cloud_cover": {"lt": query.max_cloud_cover}}
+
+        items = _page_search(f"{self.root}/search", body=body, timeout=self.timeout)
+        if self.assets is not None:
+            items = [_subset_assets(it, self.assets) for it in items]
```
espg: 🤖 *from Claude (review)*

**One anomalous item aborts a fully-paged crawl.** `_subset_assets` raises inside this list-comp over *all* already-fetched items, so a single foreign/empty-asset item (plausible in a `sentinel-2-c1-l2a` + `sentinel-2-pre-c1-l2a` union where asset schemas differ) discards a completed multi-thousand-item page crawl. Consider collecting the offenders and skipping-with-report (or raising only after the crawl, with counts), rather than letting item N kill items 0..N-1. Flagging as a design call for @espg — reasonable to defer, but worth a decision before this runs at fleet scale.

src/zagg/catalog/sources.py
```diff
@@ -102,6 +136,56 @@ def _normalize_assets(item: dict, *, preserve_thumbnails: bool) -> dict:
...
+        else:
+            resp = requests.get(url, params=params, timeout=timeout)
+        resp.raise_for_status()
+        doc = resp.json()
+        feats = doc.get("features", [])
+        items.extend(feats)
+        nxt = next((ln for ln in doc.get("links", []) if ln.get("rel") == "next"), None)
+        if not nxt or not feats:
+            break
+        url = nxt.get("href", url)
```
espg: 🤖 *from Claude (review)*

**`nxt.get("href", url)` can spin forever; no page/item cap.** The pre-refactor code used `nxt["href"]` (a href-less next link raised `KeyError` — loud). With the fallback to the current `url`, a next link that omits `href` but returns non-empty `features` re-requests the same URL/body on every iteration → infinite loop (the `not feats` guard does not trip). Unlikely against Earth Search/CMR, but there is otherwise no max-page or max-item bound on the crawl. Minor: keep the `KeyError` on a missing href, or add a loop bound.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of Phase 1 (generic STAC source). Overall the refactor is clean and the CMR GET-flow really is behavior-identical to the pre-refactor `_search` (default `nxt.get("method","GET")`, `not feats` termination, GET→POST switch all preserved), POST body construction (collections/bbox/datetime/`eo:cloud_cover` `lt`) and provenance are correct, and the JSON/parquet round-trips for multi-asset entries are sound. Four findings, none blocking Phase 2 but two worth folding here.

**Ranked:**

1. **[Medium] Shape-change regression for `preserve_thumbnails=True` catalogs** (`sources.py:435`). The `extra` filter excludes only `data`/`data_s3`/`metadata`, so a preserved `browse`/`thumbnail` asset makes a *CMR* record grow `assets` + `datetime` — breaking the "existing CMR shardmaps byte-identical" invariant for that path. Verified by repro. Default path (`preserve_thumbnails=False`) is fine and is the only one pinned by `test_h5_record_shape_unchanged`, so this is untested. Fix: add `thumbnail`/`browse` to the excluded set, or gate the multi-asset branch on source kind rather than "any non-canonical asset present".

2. **[Medium] `_subset_assets` accepts partial asset sets silently** (`sources.py:146`). The raise only fires when *all* requested keys are missing; an item missing *some* bands keeps a partial map and passes — exactly the "unreadable granule surfacing later" the raise claims to prevent. Untested (`test_no_match_raises` covers only all-missing). Decide whether keep-list keys are mandatory (`set(keep)-set(have)` → raise) or partial is allowed (document it).

3. **[Medium / design, @espg call] One bad item aborts a completed crawl** (`sources.py:316`). `_subset_assets` raises inside a list-comp over all paged items, so a single foreign/empty item (plausible in the c1 + pre-c1 collection union) discards a multi-thousand-item crawl. Consider skip-with-report or raise-after-crawl. Reasonable to defer, but worth a decision before fleet scale.

4. **[Low] `nxt.get("href", url)` fallback + no crawl bound** (`sources.py:178`). A href-less next link with non-empty features would re-request the same URL forever (pre-refactor `nxt["href"]` raised loudly instead); no max-page/item cap otherwise. Unlikely against Earth Search/CMR.

**Test-coverage gaps** (beyond the untested cases above): datetime round-trip is only exercised with a `Z` suffix — a numeric offset (`+00:00`/`-05:00`) is untested; and there is no coverage of a heterogeneous collection-union asset schema. Nothing in the pager/body/round-trip logic I could fault beyond the four items above.

src/zagg/catalog/sources.py
espg: 🤖 *from Claude*

Fixed in 94606e5: the multi-asset `assets`/`datetime` extras are now gated on the record having *no* canonical data asset (`data`/`data_s3` both absent), so every CMR record -- including `preserve_thumbnails=True` -- stays byte-identical to pre-#218. Added `test_preserved_thumbnail_keeps_pre218_shape` pinning that path, and updated the `granule_records` docstring.

src/zagg/catalog/sources.py
espg: 🤖 *from Claude*

Fixed in 42bfe25: `_subset_assets` is now per-item strict -- it returns the subset only when the item carries *every* requested key, else `None` (no partial map). A partially-assetted item would break the Phase 2 readers per-band reads, so it is skipped rather than kept. The all-mandatory semantics is the authors call and stands for espg review. New coverage: `test_missing_some_returns_none` and the fetch-level `test_partial_missing_item_skipped_and_recorded`.

src/zagg/catalog/sources.py
espg: 🤖 *from Claude*

Fixed in 42bfe25: `STACSource.fetch` now partitions the paged items instead of raising mid-crawl -- items with all requested assets are kept, items missing any are skipped with a single `logging.warning` (count + up to 5 example ids), and the skip is recorded in catalog metadata under `skipped_items`. A `ValueError` is raised only when *no* item survives (message names the requested keys and an example available-key set). One anomalous item no longer discards a multi-thousand-item crawl; loud-skip + provenance is the middle ground, and this semantics decision is the authors call and stands for espg review. Covered by `test_all_skipped_raises_with_keys` and `test_heterogeneous_union_no_corruption`.

src/zagg/catalog/sources.py
espg: 🤖 *from Claude*

Fixed in 33f69c6: restored the pre-refactor `url = nxt["href"]` so a next link without an href raises `KeyError` loudly instead of re-requesting the current url forever. Updated `test_post_flow_next_without_method_stays_post` (and `test_post_flow_merges_next_body`, which also had an href-less next link) to include an href.

src/zagg/processing/raster.py
```diff
@@ -0,0 +1,157 @@
...
+        gathered[m] = arr[vr[m] - trow * th, vc[m] - tcol * tw]
+
+    values = np.full(rows.shape, fill, dtype=gathered.dtype)
+    values[valid] = gathered
+    return values, valid
+
+
+def sample_asset(grid, cells, href: str, **kwargs):
+    """Sync facade over :func:`sample_asset_async` (worker call sites are sync)."""
+    return asyncio.run(sample_asset_async(grid, cells, href, **kwargs))
```

`sample_asset` is the documented **sync** entry ("worker call sites are sync") and wraps `asyncio.run`. `asyncio.run` raises `RuntimeError: asyncio.run() cannot be called from a running event loop` whenever an event loop is already running — which is exactly the case in Jupyter/IPython. Phase 5 of #218 is an explicit **Binder-runnable fusion notebook**, so the first cell that calls `sample_asset(...)` there will hard-fail.

Verified:
```
RuntimeError: asyncio.run() cannot be called from a running event loop
```
Suggest: detect a running loop (`try: asyncio.get_running_loop()`) and fall back to running the coroutine on a worker thread (or expose `sample_asset_async` as the notebook path). The Lambda/worker sync path is unaffected either way.

src/zagg/processing/raster.py
```diff
@@ -0,0 +1,157 @@
...
+        dtype = _DTYPES[(int(ifd.sample_format[0]), int(ifd.bits_per_sample[0]))]
+        return np.full(rows.shape, fill, dtype=dtype), valid
+
+    pairs = np.unique(np.stack([tr, tc], axis=1), axis=0)
+    tiles = await asyncio.gather(*[tiff.fetch_tile(int(c), int(r), 0) for r, c in pairs])
+    decoded = await asyncio.gather(*[t.decode() for t in tiles])
+
+    gathered = None
+    for (trow, tcol), dec in zip(pairs, decoded):
+        arr = np.asarray(dec)[:, :, 0]
```
espg, line 173: 🤖 *from Claude (review)*

`np.asarray(dec)[:, :, 0]` silently keeps only band 0. For the Sentinel-2 per-band single-band COGs this is correct, but #218 places this behind the general `grid.sample`/raster→grid primitive ("shared by future raster sources and rectilinear targets"). A multi-band COG (e.g. an RGB or a stacked-band asset) would be read as band 0 with no error and no signal to the caller. Suggest either asserting single-band here (`SamplesPerPixel == 1`, i.e. `len(ifd.bits_per_sample) == 1`) so a multi-band asset fails loudly, or documenting the single-band contract on `sample_asset`.

src/zagg/processing/raster.py
```diff
@@ -0,0 +1,157 @@
...
+        if gathered is None:
+            gathered = np.full(vr.shape, fill, dtype=arr.dtype)
```

`np.full(..., fill, dtype=arr.dtype)` raises rather than wraps when `fill` does not fit the asset dtype — e.g. a caller passing a `-1` / `NaN` sentinel into a `uint16` band:
```
>>> np.full((3,), -1, dtype=np.uint16)
OverflowError: Python integer -1 out of bounds for uint16
```
(numpy 2.2.6, this repo). The default `fill=0` is safe for the S2 nodata-DN-0 convention, but a non-fitting `fill` fails deep in the gather with an opaque message (same at lines 133 and 147). Minor: either validate `fill` against the asset dtype up front with a clear error, or document that `fill` must be representable in the asset dtype.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of phase 2 (commit 173d587) — the pull-NN raster path. Ran `pytest tests/test_raster.py` (22 pass) and repro-verified each item below.

**The core is solid.** Affine inversion in `sample_nearest` is correct for north-up, rotated, and cross-CRS cases; the pixel-corner floor convention is right (and, because a point is always within 0.5 px of its containing pixel's center, floor-of-containing == nearest-center — the primitive really does implement most-central selection); 9-element `proj:transform` truncation, tiepoint `x - i*sx` / `y + j*sy`, and `fetch_tile(x=col, y=row, z)` argument order all check out against the async-tiff stubs and the asymmetric-value multi-tile test. Points that fail to project to `inf` are correctly masked (`+inf`→int64 max is `< 0`-false but `>= h`-true, `-inf`→min is `< 0` — both land `valid=False`), so bad projections do not poison the gather. Geographic-source rasters (`EPSG:4326` affine in lon/lat) are handled correctly under `always_xy`. `SampleFormat` enum ints (Uint=1/Int=2/Float=3) match the `_DTYPES` keys, so the empty-shard dtype fallback is right. EPSG 32767 / 0 raise `CRSError` (fail-loud, not silent misgeoref). Adding `sample` to the `@runtime_checkable` `OutputGrid` Protocol is safe — there are **no** `isinstance(x, OutputGrid)` checks anywhere in src/tests, so the many fake grids do not break.

**Findings (ranked):**

1. **[Medium] `sample_asset` breaks inside a running event loop (Jupyter/Binder).** `asyncio.run` raises `RuntimeError` when a loop is already running — precisely the phase-5 notebook environment. `sample_asset` is the documented sync entry. (inline @ raster.py:154)

2. **[Low-Med] Multi-band assets silently reduced to band 0** via `[:, :, 0]`. Fine for S2 single-band COGs, but this is billed as the general raster→grid primitive; a multi-band COG is read as band 0 with no signal. (inline @ raster.py:141)

3. **[Low] `fill` outside the asset dtype raises `OverflowError`** in `np.full` (e.g. `-1`/`NaN` into `uint16`), deep in the gather with an opaque message. Default `fill=0` is safe. (inline @ raster.py:143)

**Noted, not blocking:** path-style S3 HTTPS (`https://s3.<region>.amazonaws.com/bucket/key`) does not match `_S3_VHOST` and falls through to a plain `HTTPStore` (region/signing dropped) — harmless for anonymous Earth Search data, which uses virtual-hosted or `s3://` hrefs, but worth a comment if path-style/dualstack/China endpoints are ever in scope. The unsuppressed `RuntimeWarning: invalid value encountered in cast` on `inf` points is cosmetic (repo does not set `filterwarnings=error`).

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Fixed in e3a5244. `sample_asset` now detects a running loop via `asyncio.get_running_loop()`: on `RuntimeError` (no loop) it takes the plain `asyncio.run` path as before; when a loop is already running (Jupyter/Binder) it runs the coroutine to completion on a one-shot `ThreadPoolExecutor(max_workers=1)` and returns `.result()` synchronously. The Lambda/worker sync path is unchanged. Docstring notes the running-loop safety and points async callers at `sample_asset_async`. New test `test_sync_facade_under_running_loop` calls the sync entry from inside `asyncio.run(...)` against the synthetic TIFF and asserts the exact value round-trip.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Fixed in 2ec7e16. `sample_asset_async` now raises `ValueError` right after opening the IFD when `len(ifd.bits_per_sample) != 1` (message: single-band rasters only; Sentinel-2 distributes one COG per band; reports the samples-per-pixel found). The `[:, :, 0]` gather stays but is now guarded, so a multi-band COG fails loudly instead of silently reducing to band 0. `_write_tiff` gained an optional `bands` param that writes a valid pixel-interleaved multi-band TIFF (tags 258/277/339 sized to `bands`, band-replicated tile data); the `bands == 1` byte layout is unchanged so existing tests are unaffected. New test `test_multiband_asset_raises` writes a 2-band raster and asserts the guard fires.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Fixed in fbb9731. `sample_asset_async` now derives the asset dtype from the IFD tags up front and validates `fill` with `np.can_cast(np.min_scalar_type(fill), dtype)`, raising a clear `ValueError` naming the fill value and asset dtype (e.g. `fill=-1 is not representable in the asset dtype uint16`) before any output array is built — no more opaque `OverflowError` deep in the gather. The precomputed `dtype` is now reused for all three `np.full` sites (no-valid, gathered, values), so the returned dtype is exactly the validated one. Docstring documents the fill-representability contract. New test `test_fill_not_representable_in_dtype_raises` passes `fill=-1` against the uint16 synthetic raster and asserts the message.

src/zagg/processing/raster.py
```diff
@@ -0,0 +1,577 @@
...
+    for e in granules:
+        if not e.get("assets"):
+            skipped += 1
+            continue
+        groups.setdefault(e.get("time_key") or e["datetime"], []).append(e)
+
+    lonlat = None  # computed once, only if some timestep has overlapping items
+    slabs: dict = {}
+    for key, items in groups.items():
+        t = time_index[key]
```
espg, line 433: 🤖 *from Claude (review)*

`t = time_index[key]` trusts a dict built elsewhere and fails with a bare `KeyError` if a group key is absent. Within this commit the index and the per-shard groups derive the key identically (`e.get("time_key") or e["datetime"]`), so it is consistent today. But this is the seam the runner phase will lean on: the ratified append flow resizes the time axis for new datatakes, and if a worker is dispatched a granule whose key is not in the index it was handed (stale index during an append, or an index built from a different shardmap subset), the worker dies with an opaque KeyError deep in a gather. Suggest a guarded lookup that names the offending key/datatake, or at least a docstring invariant that the index MUST be built from the same granule set that is dispatched.

src/zagg/processing/raster.py
```diff
@@ -0,0 +1,577 @@
...
+            for f in fields
+        ]
+    )
+    values = {f: r[0] for f, r in zip(fields, results)}
+    valid = results[0][1].copy()
+    for _v, mask, _c in results[1:]:
+        valid &= mask
+    if nodata is not None:
+        for f in fields:
+            valid &= values[f] != nodata
```
espg, line 290: 🤖 *from Claude (review)*

The nodata mask ANDs one global `nodata` across every band, so a cell is dropped if it equals nodata in ANY band. For the ratified S2 config that includes `scl`, this means a cell with valid `red` (>0) but `scl == 0` is masked out entirely. `scl == 0` is S2 NO_DATA, so this is defensible as footprint masking — but note the coupling: there is no per-band nodata, so any band that legitimately carries a 0 sentinel with different semantics than "no data" would silently drop otherwise-valid cells across all bands. Worth a one-line docstring note that `nodata` is a single scene-wide sentinel and only bands that share it should be co-declared. (Not a blocker; flagging the semantics.)

src/zagg/processing/raster.py
```diff
@@ -0,0 +1,577 @@
...
+                    grid,
+                    cells,
+                    e["assets"],
+                    bands,
+                    nodata=nodata,
+                    region=region,
+                    anonymous=anonymous,
+                )
+            )
+            for e in items
```

Perf note for the runner phase (correctness is fine): timesteps here run fully sequentially, and within each timestep every item is sampled via a separate `_run_sync` — so under a Lambda worker (no running loop) each item spins up and tears down its own `asyncio.run` event loop, and nothing overlaps across timesteps. The per-item band fan-out inside `sample_item_async` is concurrent, but a dense multi-year S2 shard with many datatakes will fetch its timesteps one-after-another against S3. That is network-bound serial work under a 15-min Lambda ceiling. Consider gathering across a timestep's items (and possibly across timesteps) in one event loop when the runner wires this up.

src/zagg/config.py
```diff
@@ -581,6 +589,79 @@ def _validate_temporal_config(config: PipelineConfig) -> None:
...
+    if not bands or not isinstance(bands, dict):
+        raise ValueError("raster pipeline requires data_source.bands (field -> {asset, dtype})")
+    for name, meta in bands.items():
+        if not isinstance(meta, dict):
+            raise ValueError(f"band '{name}' must be a mapping (got {meta!r})")
+        for key in ("asset", "dtype"):
+            if not isinstance(meta.get(key), str) or not meta.get(key):
+                raise ValueError(f"band '{name}' requires a string '{key}'")
+        for key in ("fill_value", "scale", "offset"):
+            if key in meta and not isinstance(meta[key], (int, float)):
```

Nit: `isinstance(meta[key], (int, float))` accepts `bool` (`isinstance(True, int)` is `True`), so `fill_value: true` / `scale: false` pass validation and later flow through as `1`/`0`. Harmless in practice but silently wrong-typed; a `not isinstance(..., bool)` guard would reject it at config time. Same applies to the `nodata` check below.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **phase 3** (commit `ea47680`): raster config validation + `get_raster_bands`, `time_key` carriage, `cell_lonlat`, and the pull-NN sampling / ownership / `(time, cell)` template+slab path in `processing/raster.py`. All 18 new tests pass locally.

**Bottom line: the core logic is correct.** I verified the load-bearing pieces with repros beyond the test suite:

- `_shard_cell_range` — confirmed `block_index()[0] * cells_per_shard` is the exact `[start, stop)` extent for **fullsphere** (nested parent id × cells_per_shard; every child nested id falls inside the range) as well as dense (populated-shard position). Slab and `cell_ids` are co-written in the same `children()` order to the same block, so the product is self-consistent regardless of morton-vs-nested ordering within a shard.
- `_combine_by_ownership` — 3-item overlap resolves to nearest tile center; dtype preserved through `np.stack`/fancy-index; the all-invalid row (`argmin` over all-`inf`) is fully covered by the caller's `out[~valid] = fill`. Ties break deterministically by granule order.
- `_chord2` — unit-sphere Cartesian, correct across the antimeridian and poles.
- `raster_time_index` — earliest-datetime coordinate, deterministic `(time, key)` ordering; `int(dt.timestamp()*1e6)` was a suspected float-truncation hazard but is **exact** for this epoch (values ~1.8e15 < 2^52) — verified over 200k microsecond samples, zero mismatch.
- Template — `n_time=0` emits cleanly; uint16 fill, the `bytes`+`zstd` codec chain, and `scale_factor`/`add_offset` attrs all round-trip through `to_zarr` → `open_array` (exercised by the end-to-end test).

**Findings (all non-blocking, ranked):**

1. *(low-med, robustness)* `process_raster_shard`: bare `time_index[key]` KeyError — trusts an externally-built index; opaque failure if the runner's index and dispatched granules ever diverge (append / stale index). Inline.
2. *(low-med, perf)* `process_raster_shard`: timesteps + items run fully sequentially, one `asyncio.run` per item — network-bound serial work under the Lambda ceiling for dense multi-year shards. Inline; a runner-phase concern.
3. *(low, semantics)* `sample_item_async`: one global `nodata` ANDed across all bands drops a cell valid in `red` but `scl==0`. Defensible for S2 (0 = NO_DATA) but worth documenting. Inline.
4. *(nit)* `_validate_raster_config`: `isinstance(True, int)` lets `fill_value: true` pass. Inline.

**Test gaps** (behavior I confirmed correct by hand but which the suite does not lock in): fullsphere slab range, 3-item ownership + all-invalid combine, `n_time=0` template, and the `time` array CF-attr round-trip. Adding a fullsphere end-to-end case and a 3-item ownership case would be the highest-value additions.

No correctness defects found — the flagged items are robustness/perf/semantics hardening for the runner phase.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Guarded the lookup: `process_raster_shard` now checks every acquisition-group key against the passed `time_index` before sampling and raises `ValueError` naming the missing key (with the invariant that the index must be built from the same manifest the shards were dispatched from — see `raster_time_index`), rather than dying with an opaque `KeyError` deep in the gather. Added the invariant to the docstring and a test (`test_missing_time_key_in_index_raises`). Commit 200d034.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Restructured to one event loop per shard: an async inner now `asyncio.gather`s `sample_item_async` across each timestep group's items *and* across timestep groups, run once via `_run_sync`, so a dense multi-datatake shard fans its S3 fetches out concurrently instead of one `asyncio.run` per item. Combine semantics (single-item fast path / `_combine_by_ownership` for overlaps) and the returned slab structure are unchanged; existing tests pass, plus a golden `test_two_timesteps_two_items_concurrent` (2 datatakes x 2 overlapping tiles) asserting the per-timestep output matches the sequential result. Commit 7a9160e.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Documented the semantics on `sample_item_async` (no behavior change): `nodata` is a single scene-wide sentinel, not per-band — for S2 a DN of 0 means the pixel is outside the scene footprint in every band, so a cell valid in `red` but reading `scl == 0` is dropped intentionally (footprint masking). Added the note that only bands sharing this sentinel's meaning should be co-declared, since a band legitimately carrying a 0 with different semantics would drop otherwise-valid cells. Commit 84a812c.

src/zagg/config.py

Added a `not isinstance(..., bool)` guard so `fill_value: true` / `scale: false` / `offset: true` now fail config validation instead of silently flowing through as `1`/`0`; applied the same guard to the `nodata` check below per your note. Test `test_bool_fill_value_rejected` covers it. Commit 58e3b8e.

espg:
🤖 *from Claude*

Heads-up per the crossover note on #220: PR #221 (xarray/h5netcdf/h5py into the Lambda layer) is CI-green and ready for review, and it edits the same `lambda` extra block in `pyproject.toml` that this PR's async-tiff pin touches — whichever lands second has a trivial rebase there. Measured budget after #221: **x86_64 178 MB / 250 MB, arm64 176 MB / 250 MB combined** — ~72–74 MB of headroom remains, so async-tiff's ~4 MB/arch fits comfortably.

src/zagg/runner.py
```diff
@@ -556,14 +561,144 @@ def _accumulate(report, i, outcome):
...
+                except Exception as e:  # noqa: BLE001 - per-shard isolation, run continues
+                    errors += 1
+                    logger.warning(f"raster shard {label} failed: {e}")
+                    continue
+                if meta["timesteps"]:
+                    shards_with_data += 1
+                    timesteps_written += meta["timesteps"]
+
+        wall_time = time.time() - t0
+        summary = {
```
espg, line 719: 🤖 *from Claude (review)*

Two summary-contract nits on this dict:

1. **`total_obs` is not an obs count for raster.** It is `timesteps_written` — the sum over shards of *how many timesteps each shard wrote* (i.e. a shard×timestep tally), not a per-cell/per-pixel observation count. The `agg` docstring advertises `total_obs` as the same key the spatial/temporal paths populate, so a consumer that sums `total_obs` across pipeline kinds gets a metric that means something different here. It reads more like a `slabs_written` counter. Consider renaming (or documenting the raster meaning) so cross-pipeline dashboards do not silently mix units.

2. **A run where every shard errors returns a success-shaped summary.** With per-shard isolation (line 666), a misconfiguration that makes *every* shard raise — e.g. a config band whose `asset` is absent from the granules’ `assets` map, which `sample_item_async` rejects with `ValueError` per item — yields `cells_error == total_cells`, `cells_with_data == 0`, `total_obs == 0`, and **no exception**: `agg` returns normally and the store is all-fill. The failures are logged as warnings and surfaced in the counts, but a caller that does not inspect `cells_error` (the notebook just prints `summary`) cannot tell an empty-AOI run from a fully-broken one. Worth considering an upfront band/asset presence check (the ShardMap carries `assets` per granule) or raising when `cells_error == total_cells and total_cells > 0`.

tests/test_raster_runner.py
```diff
@@ -0,0 +1,133 @@
...
+    _write_tiff(tmp_path / "t0.tif", data)
+    _write_tiff(tmp_path / "t1.tif", np.full((96, 96), 555, dtype=np.uint16))
+    cfg = _cfg(tmp_path)
+    shard = _shard_for_raster()
+    grid = from_config(cfg, populated_shards=[shard])
+    entries = [
+        _entry("g0", str(tmp_path / "t0.tif"), T0, "dt-1"),
+        _entry("g1", str(tmp_path / "t1.tif"), T1, "dt-2"),
+    ]
+    sm = ShardMap(grid.spatial_signature(), [shard], [entries], {"collection": "s2-test"})
```
espg, line 87: 🤖 *from Claude (review)*

The end-to-end fixture is **single-shard** (`[shard]`, `[entries]`), so the parts of `RasterStrategy` that only exist because there is more than one shard go untested:

- **Multi-shard fan-out + disjoint slab writes.** The whole point of the thread pool is concurrent `write_raster_slab`/`write_raster_coords` into non-overlapping `_shard_cell_range` blocks. A `block_index`/`cells_per_shard` off-by-one (wrong `start`, or two shards colliding on a chunk) would still pass every test here because there is only one block. A 2-shard manifest with distinct DN patterns per shard, asserting each shard reads back its own values at its own range, would actually exercise the fan-out. (Chunk-object disjointness itself holds — `cells_per_shard` is always a multiple of `cells_per_chunk` — but nothing pins it.)
- **`_check_signature` mismatch path** (build the ShardMap under one grid, run `agg` under a different `parent_order`/`child_order`) — the raise is untested.
- **Rerun semantics.** `overwrite=False` on an existing store raises `ContainsGroupError` (verified locally), and `overwrite=True` re-emits the full template. Neither is covered, yet rerun/overwrite is the only way to add timesteps today (there is no resize-on-append despite the notebook’s cell-14 claim). A test pinning "overwrite=False over an existing store raises" and "overwrite=True with a changed time index rewrites cleanly" would lock the contract.

notebooks/sentinel2_fusion.ipynb
```diff
@@ -0,0 +1,355 @@
...
+   "source": [
+    "## 6. Where this goes\n",
+    "\n",
+    "* **Training loaders**: the band arrays are chunked `(1, 65536)` -- one\n",
+    "  storage object per (timestep, shard). A PyTorch `DataLoader` can range-read\n",
+    "  chunk `k` from every aligned array (`red`, `nir`, `scl`, the ICESat-2\n",
+    "  aggregates) and get co-registered tensors with no geospatial code in the\n",
+    "  loop: `np.asarray` of a Zarr block, straight to `torch.from_numpy`.\n",
+    "  The scale/offset decode is one fused multiply-add on the GPU.\n",
+    "* **More timesteps**: appends are the standard Zarr resize-then-write-slab\n",
```
espg: 🤖 *from Claude (review)*

Minor accuracy check on this "More timesteps" bullet: it states appends "are the standard Zarr resize-then-write-slab pattern (the runner owns the resize)" in the present tense, but `RasterStrategy.run` has **no resize path** — it re-emits the full template every run, and `overwrite=False` over an existing store raises `ContainsGroupError`. Today the only way to add a datatake is a full `overwrite=True` re-run. The runner docstring correctly frames the resize as future work ("and, later, the single-writer resize on append"); this bullet is under a "Where this goes" heading so the aspirational tense is defensible, but a reader could take it as a currently-working feature. Consider "appends **will** follow the standard resize-then-write-slab pattern" to match the implementation state.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **phases 4 (a7ca9e3) and 5 (cbb6bf3)**. No blocking defects — the pipeline is correct where it counts. Verified locally: `pytest tests/test_raster_runner.py` green (4 passed); reran `agg` twice to confirm rerun behavior; confirmed chunk-object disjointness (`cells_per_shard % cells_per_chunk == 0` always, so concurrent shard slab writes never collide on a chunk — thread-pool writes are safe); notebook imports load top-to-bottom and every dep is core or in `[analysis,catalog,viz]` (async-tiff/obstore/mortie are core, xarray/matplotlib in analysis).

**Things I checked that are fine (no action):**
- `**_ignored` in `RasterStrategy.run` swallows `driver`, `handoff`, `function_name`, `profile`, `invocation`, `force_cold`, `events` — all legitimately N/A to raster (no HDF5 driver, no carrier, Lambda raises `NotImplementedError`). `overwrite` **is** honored (passed to `emit_raster_template`).
- `max_workers = min(max_workers or 4, len(cells)) or 1` — `len(cells)==0` degrades to 1 worker with nothing submitted (no crash); empty-granule shards after notebook subsetting no-op cleanly.
- `overwrite=False` rerun raises `ContainsGroupError` rather than silently reusing a stale time axis — safe, not a corruption footgun.
- Strategy selection `kind == "spatial" and reader == "raster"`; `_maybe_warn_dense` intentionally does **not** fire for raster (dense-by-construction is the design, not a warning-worthy accident).
- `_check_signature` is spatial-only, correct for a ShardMap (a band mismatch is caught later per-item by `sample_item_async`, not here).
- Notebook: fully anonymous (Earth Search + public c1 COGs + source.coop ATL06, `skip_signature=True`); cross-order fusion math (`clip2order(12,…)` + `searchsorted`) aligns cell order between `cells_morton` and the gathered slabs; `times[0]` is safe with ≥1 datatake.

**Findings (all non-blocking, inline):**
1. *(low-med, runner.py:675)* A run where **every** shard errors (e.g. a config band absent from the granules’ `assets`) returns a success-shaped summary and an all-fill store — surfaced only via `cells_error`, which callers/the notebook do not inspect. Consider an upfront asset-presence check or raising when all shards error.
2. *(low, runner.py:679)* `total_obs` here is a shard×timestep tally, not an observation count — cross-pipeline consumers of the documented `total_obs` key mix units.
3. *(medium, tests)* Coverage is single-shard only: multi-shard fan-out / disjoint slab writes, `_check_signature` mismatch, and overwrite/rerun semantics are all untested — a cross-shard `block_index` off-by-one would pass today.
4. *(low, notebook cell 14)* The "appends are the standard resize-then-write-slab pattern" bullet is present-tense but no resize/append path is implemented yet.

None of these gate the phase. (2)/(4) are wording; (1)/(3) are worth a follow-up commit if cheap.

tests/test_raster_runner.py
espg: 🤖 *from Claude*

Folded in a7f22c7. Three tests added to `tests/test_raster_runner.py`:

- `test_multi_shard_disjoint_slabs` — two rasters ~7 km apart feed two *distinct* order-10 shards in one `agg` run; each shard is read back at its own `_shard_cell_range` with its own DN pattern (index raster vs constant 777), so a cross-shard `block_index`/`cells_per_shard` collision surfaces one shard's DNs in the other's range and fails deterministically.
- `test_signature_mismatch_raises` — ShardMap built under a parent 9 / child 15 grid against a parent 10 / child 16 run config; asserts `_check_signature` raises.
- `test_overwrite_and_rerun` — two `overwrite=True` runs rewrite cleanly, then `overwrite=False` over the existing (2-timestep) store with a *changed* one-datatake manifest raises `ContainsGroupError`, and `overwrite=True` with that changed index rewrites to 1 timestep.

One accuracy note on the rerun contract: `overwrite=False` raises `ContainsGroupError` only when the new template *differs* from the existing store — pydantic-zarr's `to_zarr(overwrite=False)` is idempotent for a byte-identical re-emit (verified locally: an identical re-emit does *not* raise). So the test drives a changed time index to exercise the genuine refuse-to-clobber path rather than an identical rerun.

src/zagg/runner.py
espg: 🤖 *from Claude*

Both findings on this dict folded.

(1) All-shards-error — 3dd2dec: `RasterStrategy.run` now captures the last caught exception in the accumulation loop and, after the pool completes, raises `RuntimeError` (naming the failed-shard count and the last error message) when `cells` is non-empty and `errors == len(cells)`. Covered by `test_all_shards_error_raises` — a config band whose `asset` is absent from every granule's `assets` map makes every shard raise, and the run now fails loudly instead of returning an all-fill success-shaped summary.

(2) `total_obs` — bdfc936: no rename, so the summary schema stays shared across strategies. Documented in the `RasterStrategy` docstring and in a comment at the summary dict that on the raster path `total_obs` counts shard×timestep slabs written (the raster analogue of an observation tally), with `timesteps` separately carrying the datatake count; a dashboard summing `total_obs` across pipeline kinds mixes units.

notebooks/sentinel2_fusion.ipynb
espg: 🤖 *from Claude*

Folded in 6838679. The "More timesteps" bullet (cell 14) is now future tense: appends *will* land as the standard Zarr resize-then-write-slab pattern (a runner-owned resize, tracked on issue #218), and it notes that today a changed time range is a full re-run with `overwrite=True` — matching the implementation state (no resize/append path yet).

src/zagg/runner.py
```diff
@@ -556,14 +561,318 @@ def _accumulate(report, i, outcome):
...
+            ev = {
+                "mode": "process_raster",
+                "shard_key": int(shard_key),
+                "granules": granules,
+                "config": config_dict,
+                "store_path": store_path,
+                "time_index": {k: time_index[k] for k in keys},
+            }
+            if output_credentials:
+                ev["output_credentials"] = output_credentials
```

**Output credentials are passed raw here, diverging from the spatial/temporal lambda envelope.** `_run_lambda_shards` sets `ev["output_credentials"] = output_credentials` unmodified, whereas the spatial path (this file, ~L346/L360) routes through `_build_output_creds_event(output_credentials, resolved_endpoint, region)` which (a) normalizes snake_case / STS PascalCase spellings to the camelCase shape the handler requires and (b) threads the resolved output endpoint into the event block.

Two concrete consequences:

1. **Cred-spelling regression.** The handler `_output_store_kwargs` (lambda_handler.py L425) hard-requires camelCase `accessKeyId`/`secretAccessKey` and raises otherwise. A caller who passes `aws_access_key_id`/`AccessKeyId` — accepted on every other lambda mode via `normalize_output_credentials` — gets every raster shard failing with `output_credentials missing keys`, i.e. `errors == len(cells)` → the all-shards-failed `RuntimeError`.

2. **Custom endpoint dropped.** `resolved_endpoint = output_endpoint_url or get_output_endpoint_url(config)` (L641) is used to open the *template* store, but is never forwarded to `_run_lambda_shards`, so workers writing to an R2/MinIO `s3://` store hit real AWS S3 — template and slabs land in different stores. The raw `output_credentials` dict here typically does not carry `endpointUrl` (the endpoint comes from the param/config, not the creds), so raster is strictly worse than spatial, which does thread it.

Suggest: thread `resolved_endpoint` into `_run_lambda_shards` and build the event block with `_build_output_creds_event(output_credentials, resolved_endpoint, region)` so the raster lambda envelope matches the other modes.

src/zagg/runner.py
```diff
@@ -556,14 +561,318 @@ def _accumulate(report, i, outcome):
...
+            return ev
+
+        transient_markers = (
```
espg, line 807: 🤖 *from Claude (review)*

**Transient-retry classification diverges from the established `_invoke_lambda_cell` set (#119) and could misclassify.** That function retries on `["TooManyRequestsException", "Rate exceeded", "Read timeout", "timed out", "UNEXPECTED_EOF"]`; here the markers are `("TooManyRequests", "ServiceException", "throttl", "Connection", "Timeout")`. Substring match is case-sensitive, so:

- A botocore `ReadTimeoutError` renders as `Read timeout occurred` — matched by `_invoke_lambda_cell` but **not** here (`Timeout` != `timeout`).
- `throttl` (lowercase) will not match a message containing `Throttling`.

Lambda `Invoke` throttles surface as `TooManyRequestsException` so the common case is covered, but the divergence is a latent gap worth closing (or a comment justifying the different set). Separately, the backoff at L822 (`time.sleep(min(2**attempt, 8))`) has no jitter, unlike `_invoke_lambda_cell` which adds `time.time() % 1` — on a wide synchronous fan-out that invites synchronized retry waves. Suggest reusing the same marker list + jittered sleep.

tests/test_raster_runner.py
```diff
@@ -0,0 +1,326 @@
...
+        self.events = []
+
+    def invoke(self, **kwargs):
+        import io
+
+        event = json.loads(kwargs["Payload"])
+        self.events.append(event)
+        resp = self.responder(event)
+        out = {"Payload": io.BytesIO(json.dumps(resp).encode())}
+        if resp.get("__function_error__"):
```
espg, line 254: 🤖 *from Claude (review)*

**`_FakeLambdaClient` grows a `__function_error__` hook that no test exercises, and the retry/backoff loop is entirely untested.** The two `TestRasterLambdaBackend` cases cover the happy path (200) and the all-non-200 raise. Untested branches of `_run_lambda_shards._one`:

- **`FunctionError` path** — the `if resp.get("FunctionError")` branch that returns `Lambda error: <raw[:150]>` with no retry. The fake already supports it via `__function_error__`; a responder returning `{"__function_error__": True, ...}` would drive it.
- **Transient retry + backoff** — a responder raising a `TooManyRequests`-marked exception on the first call then succeeding would prove the retry loop, the marker match, and that a shard recovers. As-is nothing covers `max_retries`, the marker set, or the sleep.
- **`output_credentials` pass-through** — no test asserts the event carries creds; this is exactly where the raw-passthrough divergence flagged on runner.py L783 hides (a normalization test comparing spelling variants would catch it).
- **datetime-only keying** — the manifest entries all carry `time_key`, so the `_event` `.get("datetime")` fallback and its `time_index` slice are never exercised with a mixed/datetime-only granule.

Suggest adding at least the FunctionError and transient-retry cases (the fake is already scripted for both).

deployment/aws/lambda_handler.py
```diff
@@ -821,6 +823,92 @@ def _handle_process_event(event: Dict[str, Any]) -> Dict[str, Any]:
...
+    """
+    start_time = time.time()
+    try:
+        required = ["shard_key", "granules", "config", "store_path", "time_index"]
+        missing = [p for p in required if p not in event]
+        if missing:
+            error_msg = f"Missing required parameters: {', '.join(missing)}"
+            logger.error(error_msg)
+            return {"statusCode": 400, "body": json.dumps({"error": error_msg})}
+
```
espg, line 848: 🤖 *from Claude (review)*

**Minor: the dense-layout rejection is per-worker, so a dense-config lambda run fires N invokes before failing.** The guard is correct and reachable (`get_layout` defaults to `fullsphere`; only an explicit `layout: dense` trips it, covered by `test_dense_layout_400`). But because the runner emits the template and fans out *before* any worker reports, a dense config means every shard returns 400 → `errors == len(cells)` → the all-shards-failed `RuntimeError` — after paying for the whole fan-out. A cheap fail-fast `get_layout(config) == "dense"` check in `RasterStrategy.run` before dispatch would reject it once, up front. Non-blocking; the worker-side guard is a fine backstop.

Separately (also non-blocking): a worker reaching `write_raster_slab` before the orchestrator template exists surfaces the raw zarr open error as a generic 500. The single-writer design makes that an orchestration bug rather than a worker one, so raw is defensible, but a one-line "template missing — orchestrator must emit before fan-out" wrap would read better in logs.

espg (commented):
🤖 *from Claude (review)*

Adversarial fresh-context review of phase 6 (commit 34ceda1) — the `mode="process_raster"` handler branch and the `RasterStrategy._run_lambda_shards` sync-invoke backend. Tests pass locally (`14 passed`). The core is sound: the JSON round-trip is handled (`int()` casts on `shard_key` and every `time_index` value), the worker slices only its own global-index entries so slabs land at correct global `t` positions, `Payload` is read exactly once per attempt (no double-read), the `total_obs`/`timesteps` summary keys stay consistent with the local backend (`len(time_index) == len(times_us)`), telemetry + the `result_url` mirror ride the shared envelope, and `open_store` tolerates the `region` kwarg on a local path. Findings, ranked:

1. **(Medium) `output_credentials` passed raw, diverging from the spatial/temporal envelope** — runner.py L783. No `normalize_output_credentials` (so snake_case/PascalCase creds that work on every other lambda mode fail the handler's camelCase check → all-shards-failed), and the resolved output `endpoint_url` used for the template store (L641) is never threaded to the workers (custom R2/MinIO output → workers hit real AWS S3). Route through `_build_output_creds_event(output_credentials, resolved_endpoint, region)` and pass `resolved_endpoint` into `_run_lambda_shards`.

2. **(Low–Medium) Untested lambda branches** — test_raster_runner.py. `_FakeLambdaClient` has a `__function_error__` hook no test uses; the FunctionError path, the transient-marker retry/backoff loop, `output_credentials` pass-through, and datetime-only `_event` keying are all uncovered. The creds gap in (1) lives exactly in the untested path.

3. **(Low) Transient-marker set diverges from `_invoke_lambda_cell` (#119)** — runner.py L786. Case-sensitive substrings miss `Read timeout`/`Timeout` casing and `Throttling`; backoff at L822 has no jitter (retry-wave risk on wide sync fan-out). Align the marker list + add jitter, or comment the intentional difference.

4. **(Low, non-blocking) Dense layout rejected per-worker** — lambda_handler.py. A dense config pays the full fan-out before the all-shards-failed raise; a fail-fast `get_layout=="dense"` guard in `RasterStrategy.run` would reject once up front. Also, a worker hitting a missing template surfaces a raw zarr 500 (defensible under single-writer, but an opaque message).

None are blockers on their own; (1) is the one I'd fold before ready-for-review, since it silently narrows which credential shapes and output endpoints the raster lambda backend accepts relative to every other mode.

src/zagg/runner.py

Fixed. `RasterStrategy.run` now threads `resolved_endpoint` into `_run_lambda_shards`, which builds the worker block via `_build_output_creds_event(output_credentials, output_endpoint_url, region)` and attaches it (only when non-None) as `event["output_credentials"]` — same envelope as the spatial/temporal paths. This restores snake_case/STS-PascalCase cred leniency and threads the custom (R2/MinIO) output endpoint to workers. Covered by `test_output_credentials_normalized_and_endpoint_threaded` (snake_case creds + custom endpoint → camelCase keys + endpointUrl in the worker event). 335514a; test also in d9d15b2.

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed. Transient classification now matches `last.lower()` against lowercase markers (`toomanyrequests`, `throttling`, `serviceexception`, `connection`, `timeout`), so `Read timeout occurred` and `Throttling` both match now; a comment notes the parity intent with `_invoke_lambda_cell` (#119). The backoff gains jitter — `min(2**attempt, 8) * (0.5 + random.random() / 2)` (`random` already imported) — so a wide synchronous fan-out no longer retries in a synchronized wave. Covered by `test_lambda_transient_retry_then_success` (transient fault on attempt 1 → retry → success, two recorded invokes). 7a4cfb3; test in d9d15b2.

Issue #222, opened 2026-07-14
Chunk-range temporal reads via hidefix (drop whole-granule fetch, eventually h5py)


espg:
🤖 *from Claude*

Follow-up to #220/#221 (which put `xarray` + `h5netcdf` + `h5py` in the Lambda layer to unblock the temporal worker): replace the temporal reader's **whole-granule fetch** with **chunk-range reads**, most likely via hidefix — which would also let `h5py`/libhdf5 leave the layer again.

## Why

`zagg.temporal.open_dataset`'s non-`.zarr` S3 branch GETs the entire object into memory before opening (`obstore.get(...).bytes()` → `xr.open_dataset(BytesIO)`). Correct, but each temporal worker pulls full MERRA-2 daily granules (hundreds of MB) to read one storm's bbox sliver. The pre-zagg orchestrator (jbbutler/antarctic_AR_dataset#1) averaged ~7 MB/storm with range reads; the first post-#221 slice run will measure what whole-file fetching costs by comparison — that number sizes the payoff here.

## What hidefix already provides (verified against the local checkout, v0.13.0)

- **`src/reader/s3.rs`**: async chunk-range S3 reader — ranged GETs, up to 16 concurrent, adjacent chunks coalesced into single requests, any S3-compatible endpoint with caller-supplied credentials. Exactly the read model needed.
- **Codecs**: gzip/deflate + shuffle — covers MERRA-2 `.nc4`.
- **Serializable chunk indexes** — the same machinery zagg's spatial `sidecar` index backend already consumes through `h5coro-hidefix`.

## The gaps (all in the Python/xarray glue, `python/hidefix/xarray.py`)

1. The xarray entrypoint is **local-path only**; the S3 reader is unreachable from `engine="hidefix"`.
2. **No prebuilt-index injection** — upstream has the exact TODO in place: `# TODO: allow take an existing index, maybe from a serialized object.`
3. `HidefixDataStore` opens **`netCDF4.Dataset`** alongside the hidefix index for metadata/attributes — adopting the engine as shipped would put libnetcdf **and** libhdf5 back in the layer (heavier than the h5py it replaces).

## Two routes (design fork — hence `discuss`)

**A. Upstream the glue** (three well-scoped PRs to gauteh/hidefix): accept `(s3 url, serialized index)` in the engine; implement the serialized-index TODO; derive metadata from hidefix's own index instead of netCDF4-python. Cleanest long-term (one engine, everyone benefits); waits on upstream review.

[...]

PR #226, opened 2026-07-15
Bound temporal worker memory: subset + load at the event extent (issue #225)

### LINKED ISSUE #225: Temporal worker OOMs: whole-granule opens held for all collections at once
🤖 *from Claude*

Third field failure of the temporal fan-out, and the first that is a genuine capacity bug rather than a config/credential gap: **every slice-run worker died `Runtime.OutOfMemory`** (CloudWatch: Max Memory Used 4095–4096 MB of 4096 MB, killed 26–128 s in). An OOM kill takes the runtime down before the async result mirror is written, so the orchestrator polls into its 990 s timeout with no signal — the failure is also *silent*, which compounds it.

**Mechanism:** `read_temporal_inputs` opens every collection's full daily granules (whole-object `obstore.get(...).bytes()` → `xr.open_dataset(BytesIO)`) and holds all of them — compressed bytes + lazy datasets — for the life of `process_event`. For the AR workload that is four collections × 2–3 days each, including `M2I3NPASM` (3-D OMEGA, ~1 GB/day), so peak resident memory scales with storm duration × collection sizes and clears 4 GB before the first timestep computes. #222 (chunk-range reads via hidefix) is the structural fix; this issue is the stopgap that keeps the 4 GB fleet viable — and remains correct afterwards.

**Fix (single phase): subset-and-load at the event extent inside the reader.**
- `_handle_process_event` opens the mask first, so the event's lat/lon extent is known before any collection read: pass it to the reader (in-process call, so arrays are fine) as an optional `extent=(lats, lons)`.
- In `read_temporal_inputs`, per granule: open lazily, apply the collection's `variables` subset (already available via `collection_options`) and the extent `.sel`, `.load()` the tiny result, then close the dataset and drop the byte buffer before opening the next granule.
- Peak memory becomes ~one granule's bytes + the loaded event-extent slivers (a few MB), independent of storm length or collection count.
- No behavioral change to selection semantics — the same exact-match `.sel(lat=lats, lon=lons)` that `process_event` performs later simply happens earlier; `prepare_collection` (offset/resample/derived) then operates on small in-memory arrays, which also removes the eager-resample-over-full-file hazard for the precip branch.
- Local backend unaffected (in-memory tuples never call the reader).

Also worth an operator note (not part of this issue's code): the async-poll path cannot distinguish OOM/crash from slow workers — the 990 s silence per event. A DLQ or destinations-on-failure hook for Event invokes would make this failure class loud; deferring unless it recurs.

Refs #213 (contract), #220/#223 (prior fleet failures), #222 (structural successor). Needs a patch release + redeploy after merge; the downstream slice re-fires unchanged (espg/antarctic_AR_dataset#1 already fails fast on systemic 0/N runs now).

espg:
Closes #225. Third fleet failure of the temporal fan-out: every slice worker died `Runtime.OutOfMemory` (CloudWatch: 4095–4096 MB used of 4096 MB, killed 26–128 s in), and OOM kills the runtime before the async result mirror writes — so the orchestrator saw 990 s of silence instead of errors. Mechanism: the reader held every collection's whole daily granules (bytes + lazy datasets) simultaneously for the life of `process_event` — four collections × 2–3 days including ~1 GB/day 3-D OMEGA clears 4 GB before the first timestep computes.

One commit:
- `read_temporal_inputs` gains `extent=(lats, lons)`: each granule is opened, subset to the collection's `variables` + the event extent, **loaded**, and closed (buffer released) before the next opens. Peak memory ≈ one granule's bytes + the loaded event-extent slivers, independent of storm length and collection count. Without `extent` the behavior is byte-identical to before (lazy, full-file).
- `_handle_process_event` passes the extent from the already-opened event mask (it reads the mask before any collection).
- Selection semantics are unchanged — the same exact-match `.sel(lat=…, lon=…)` `process_event` performs later simply happens earlier; `prepare_collection` (offset/resample/derived) now operates on small in-memory arrays, which also defuses the eager-resample-over-whole-file hazard in the precip branch.
- Tests: extent path subsets/loads/frees (close spy via `set_close`, multi-URI concat under extent, variables pre-subset), no-extent path unchanged, handler forwards the mask's coordinates.

This is the stopgap that keeps the 4 GB fleet viable; #222 (hidefix chunk-range reads) is the structural successor and this change stays correct under it. Needs a patch release + redeploy; the downstream slice re-fires unchanged.

## Testing
`pytest tests/test_temporal.py tests/test_lambda_handler.py tests/test_runner.py`: 343 passed locally (pre-fix), 172 in the two directly-affected files post-fix; full suite via CI.

### DIFF
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -979,11 +979,14 @@ class TestTemporalReader:
         with pytest.raises(ValueError, match="input_credentials"):
             _input_channel("anonymous")
 
-    def test_read_temporal_inputs_extent_subsets_loads_and_frees(self, monkeypatch):
+    def test_read_temporal_inputs_extent_subsets_loads_and_closes(self, monkeypatch):
         # With an event extent, each granule is subset to it (and the
         # collection's variables), loaded, and closed before the next opens --
         # peak memory ~one granule instead of every collection whole
-        # (issue #225).
+        # (issue #225). This test asserts close() is invoked once per granule;
+        # the actual buffer release was verified empirically in the PR-226
+        # review (weakref dies after the loop). The structural guarantee here is
+        # that close() runs before the next granule opens.
         xr = pytest.importorskip("xarray")
         import zagg.temporal as temporal
         from zagg.temporal import read_temporal_inputs
@@ -1021,7 +1024,10 @@ class TestTemporalReader:
         # concat + sortby: day 1 before day 2, values loaded as plain numpy
         assert out["T2M"].values[0, 0, 0] == 1.0
         assert isinstance(out["T2M"].data, np.ndarray)
-        assert len(closed) == 2  # each granule's buffer released after load
+        # close() invoked once per granule (spy replaces the h5netcdf closer);
+        # buffer release itself verified empirically in the PR-226 review. The
+        # structural guarantee is that close runs before the next granule opens.
+        assert len(closed) == 2
 
     def test_read_temporal_inputs_no_extent_stays_lazy_shaped(self, monkeypatch):
         # Without an extent the reader returns the opened datasets untouched
--- a/tests/test_temporal.py
+++ b/tests/test_temporal.py
@@ -979,6 +979,62 @@ class TestTemporalReader:
         with pytest.raises(ValueError, match="input_credentials"):
             _input_channel("anonymous")
 
+    def test_read_temporal_inputs_extent_subsets_loads_and_frees(self, monkeypatch):
+        # With an event extent, each granule is subset to it (and the
+        # collection's variables), loaded, and closed before the next opens --
+        # peak memory ~one granule instead of every collection whole
+        # (issue #225).
+        xr = pytest.importorskip("xarray")
+        import zagg.temporal as temporal
+        from zagg.temporal import read_temporal_inputs
+
+        lat = np.arange(-80.0, -60.0, 0.5)
+        lon = np.arange(-30.0, 30.0, 0.625)
+        closed = []
+
+        def _wide(uri, **kwargs):
+            day = int(uri[-4])
+            time = np.array([f"1980-01-0{day}T00", f"1980-01-0{day}T03"], dtype="datetime64[ns]")
+            ds = xr.Dataset(
+                {
+                    "T2M": (("time", "lat", "lon"), np.full((2, lat.size, lon.size), float(day))),
+                    "SLP": (("time", "lat", "lon"), np.zeros((2, lat.size, lon.size))),
+                },
+                coords={"time": time, "lat": lat, "lon": lon},
+            )
+            ds.set_close(lambda u=uri: closed.append(u))
+            return ds
+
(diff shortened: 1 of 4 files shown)


tests/test_temporal.py
```diff
@@ -979,6 +979,62 @@ def test_input_channel_forms(self):
...
+            extent=(ev_lats, ev_lons),
+        )
+        out = collections["merra2"]
+        assert list(out.data_vars) == ["T2M"]
+        np.testing.assert_array_equal(out["lat"].values, ev_lats)
+        np.testing.assert_array_equal(out["lon"].values, ev_lons)
+        # concat + sortby: day 1 before day 2, values loaded as plain numpy
+        assert out["T2M"].values[0, 0, 0] == 1.0
+        assert isinstance(out["T2M"].data, np.ndarray)
+        assert len(closed) == 2  # each granule's buffer released after load
```
espg: 🤖 *from Claude (review)*

Minor / test-honesty (non-blocking): `set_close` **replaces** the datasetʼs close callback with this spy, so `closed.append(u)` runs *instead of* the h5netcdf backendʼs real closer. The assertion therefore verifies only that `.close()` is **invoked** per granule — not that the backing `BytesIO` buffer is actually released. The comment here ("each granuleʼs buffer released after load") and the `_and_frees` in the test name both read as if release itself is exercised, which it isnʼt.

I confirmed empirically (outside this suite) that the real path *does* free the buffer: after `part = ds[keep].sel(...).load(); ds.close(); ds = part`, a `weakref` to the opened `io.BytesIO` is dead post-`gc.collect()` (buffer collectable — the fix genuinely works), and peak RSS during `.load()` of a 3-point extent subset from a 151 MB / 3-var file was +23 MB, i.e. chunk-scale not variable-scale. So the code is correct; only the testʼs framing overstates its own coverage.

Options: (1) rename to `_subsets_loads_and_closes` and reword the comment to "close() called per granule (buffer release verified separately)"; (2) drop the spy and instead assert release directly via `weakref.ref` on the buffer (needs `open_dataset` to expose it / a real in-memory netcdf, heavier); (3) leave as-is if the intent is only to pin the close-per-granule contract. I lean (1) — cheapest, and stops the name implying an RSS assertion the unit test cannot make.

espg: 🤖 *from Claude*

Good catch — the spy replaces the real h5netcdf closer, so the assertion proves `close()` is invoked once per granule, not that the byte buffer is released. Folding this:

- Renamed the test `test_read_temporal_inputs_extent_subsets_loads_and_frees` → `..._loads_and_closes`.
- Reworded the inline comment on `assert len(closed) == 2` to state precisely what is asserted (close() invoked once per granule; buffer release verified empirically in the PR-226 review; structural guarantee = close runs before the next granule opens).
- Updated the top-of-test comment to match so it no longer overstates coverage.

`pytest tests/test_temporal.py -q` green (81 passed), ruff clean. Commit follows shortly.

PR #230, opened 2026-07-15
Declarative coord_round for dirty source-grid coordinates (issue #229)

### LINKED ISSUE #229: Temporal reader: declarative coord_round for source grids with dirty coordinates
🤖 *from Claude*

Fifth fleet finding, and the root cause predates every pipeline involved: **MERRA-2 granules carry float dirt in their own coordinate arrays** — each file has exactly one dirty longitude (`-5.920304394294029e-13` where `0.0` should be) and one dirty latitude (equator). Verified against a live granule; the AR storm masks inherited the dirt from MERRA-2 via the Wille catalog, and the downstream repo's own reference pipeline rounds granule coordinates everywhere for precisely this reason (a code comment there says 'avoid any issues with -0 not matching 0').

Consequence on the fleet: exported masks are 5dp-rounded (espg/antarctic_AR_dataset#1), the statics are rounded, but the temporal reader selects against the granule's **native** coords — so the extent `.sel` raises `"not all values found in index 'lon'"` for any event whose bbox crosses the prime meridian (live failure: storm_00003 on the slice run, post-#225 fleet). The local backend never sees it because `local_event_tuples` rounds granule coords on open — which is exactly why the 84/84 parity run was green.

**Fix (single phase):** a declarative per-collection option `coord_round: <decimals>` —
- validated in `_validate_collection_options` (non-negative int),
- applied in `read_temporal_inputs` immediately after each granule opens (before the `variables`/extent subset, so the extent `.sel` compares like with like),
- applied as the first step of `prepare_collection` (local path parity; idempotent when the caller already rounded).
Downstream configs then declare `coord_round: 5` on the MERRA-2 collections — the same convention the reference pipeline hard-codes, expressed as config per the no-user-code-on-Lambda policy (#213).

Default off: silently rounding arbitrary source grids is a data mutation zagg must not assume.

Needs a patch release + redeploy. Refs #213, #225/#226 (the extent sel this surfaces through), espg/antarctic_AR_dataset#1.

espg:
Closes #229. Fifth fleet finding, root cause upstream of every pipeline involved: MERRA-2 granules carry float dirt in their own coordinate arrays (one dirty lon — `-5.92e-13` for 0.0 — and one dirty lat per file, verified live). Rounded event masks then fail the reader's exact-match extent `.sel` for any event crossing the prime meridian; the local backend never sees it because `local_event_tuples` rounds granule coords, which is why the 84/84 parity run stayed green while the fleet failed.

One commit: per-collection `coord_round: <decimals>` — validated in `_validate_collection_options` (non-negative int, bools rejected), applied in `read_temporal_inputs` immediately after each granule opens (before the variables/extent subset, so the extent sel compares like with like), and as the first step of `prepare_collection` (local-path parity; idempotent when the caller already rounded). Default off — silently rounding arbitrary source grids is a data mutation zagg must not assume; the downstream configs opt in with `coord_round: 5`, expressing declaratively the same convention the reference pipeline hard-codes ('avoid any issues with -0 not matching 0').

Tests: dirty-grid round in `prepare_collection`; reader-level dirty-granule + rounded-extent case (KeyError without the option, clean subset with it); config validation. 411 passing across the three affected files.

Needs a patch release + fleet redeploy; the downstream configs gain `coord_round: 5` in the same window (espg/antarctic_AR_dataset#1).

### DIFF
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -702,6 +702,14 @@ def _validate_collection_options(config: PipelineConfig) -> None:
             raise ValueError(
                 f"data_source.collections[{cname!r}].variables must be a list of names"
             )
+        coord_round = opts.get("coord_round")
+        if coord_round is not None and (
+            isinstance(coord_round, bool) or not isinstance(coord_round, int) or coord_round < 0
+        ):
+            raise ValueError(
+                f"data_source.collections[{cname!r}].coord_round must be a "
+                f"non-negative integer (got {coord_round!r})"
+            )
         offset = opts.get("time_offset")
         if offset is not None:
             if not isinstance(offset, str):
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -672,7 +672,9 @@ def _validate_collection_options(config: PipelineConfig) -> None:
 
     The block accepts a list of collection names (no options) or a mapping of
     name -> per-collection reader options consumed by
-    :func:`zagg.temporal.prepare_collection`: ``variables`` (list of names),
+    :func:`zagg.temporal.prepare_collection`: ``coord_round`` (non-negative
+    int, decimals to round lat/lon coords to -- for source grids whose own
+    coordinate arrays carry float dirt), ``variables`` (list of names),
     ``time_offset`` (a pandas-parseable offset string), ``resample``
     (``{freq, how: sum|mean, scale}``), and ``derived`` (name -> numpy
     expression string). ``time_offset`` and ``resample.freq`` are parsed with
--- a/src/zagg/temporal.py
+++ b/src/zagg/temporal.py
@@ -697,6 +697,7 @@ def prepare_collection(ds, options):
     """
     if not options:
         return ds
+    ds = _round_coords(ds, options.get("coord_round"))
     variables = options.get("variables")
     if variables:
         ds = ds[list(variables)]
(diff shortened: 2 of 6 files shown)


src/zagg/config.py
```diff
@@ -702,6 +702,14 @@ def _validate_collection_options(config: PipelineConfig) -> None:
             raise ValueError(
                 f"data_source.collections[{cname!r}].variables must be a list of names"
             )
+        coord_round = opts.get("coord_round")
```
espg, line 707: 🤖 *from Claude (review)*

Doc gap (low): this same function's docstring above (lines 674–678) enumerates the recognized option set as a closed list — `variables`, `time_offset`, `resample`, and `derived` — but the new `coord_round` you validate here isn't added to it. Since the enumeration reads as exhaustive (the `and` before `derived`), a reader will not learn `coord_round` is accepted. Worth a one-line addition to keep the docstring in sync with what the function now validates.

src/zagg/temporal.py
```diff
@@ -716,6 +717,20 @@ def prepare_collection(ds, options):
     return ds
 
 
+def _round_coords(ds, decimals):
```
espg, line 720: 🤖 *from Claude (review)*

Doc/example gap (low): `coord_round` is not reflected in the user-facing docs. `docs/temporal_events.md` (the reader-options YAML block + the "Options apply in the order listed above" prose) enumerates `variables`/`time_offset`/`resample`/`derived` but not `coord_round`, which now runs *first* (ahead of `variables`) — so the documented order is also slightly stale. And `src/zagg/configs/merra2_storm.yaml` is the canonical MERRA-2 example, i.e. exactly the collection that hits the fleet failure #229 fixes, yet it doesn't demonstrate the option. Two options: (a) add a `coord_round: 5` line to the merra2 collection(s) in `merra2_storm.yaml` plus a docs line (surfaces the fix where it matters); or (b) leave both, treating it as an advanced default-off knob and keep the example minimal. Neither blocks — flagging for your call.

src/zagg/config.py
espg: 🤖 *from Claude*

Folded. Added `coord_round` to the `_validate_collection_options` docstring option enumeration in `src/zagg/config.py` — listed first (matching its apply-order in `prepare_collection`) as a non-negative int, decimals to round lat/lon coords to, for source grids whose own coordinate arrays carry float dirt. Commit follows shortly.

src/zagg/temporal.py
espg: 🤖 *from Claude*

Folded. Added `coord_round` to the user-facing docs: in `docs/temporal_events.md` the collections YAML example now shows `coord_round: 5` with a comment, and the options-ordering prose now states `coord_round` runs first (before `variables`/`time_offset`/`resample`/`derived`) — verified against `prepare_collection` and the reader loop in `src/zagg/temporal.py`, where `_round_coords` is applied ahead of the other options. Also added `coord_round: 5` to both MERRA-2 collection entries in `src/zagg/configs/merra2_storm.yaml` with a one-line "float dirt (issue #229)" comment. Commit follows shortly.

Issue #228, opened 2026-07-15
Pre-flight feasibility check for Lambda runs (memory/timeout/capability vs shard geometry)


espg:
🤖 *from Claude*

## Summary

Add a **pre-flight feasibility check for Lambda runs**: given a shardmap + config + deployed function, predict whether the fan-out will *work*, *fail*, or is *on the border* — and warn/error **before** dispatching (and before the user pays for a doomed fan-out). Requested by espg after today's Sentinel-2 empirical-test session (issue #218 follow-up).

## Motivation (fresh empirics)

Two failure classes surfaced while standing up the SERC S2 time-series test, both discoverable only *after* invoking the fleet:

1. **Worker memory vs shard geometry.** The raster worker accumulates one slab per timestep × band before writing, so per-shard peak is
   `n_timesteps × Σ(band dtype bytes) × 4^(child_order − parent_order)` (+ decode/pool overhead).
   Concrete numbers from today, SERC AOP box at child 19 on a 2048 MB function:
   - o9 shards × 2025-only (85 datatakes × 5 bands) ≈ **~900 MB — fits, but on the border**;
   - o9 × all-time (~600 datatakes) ≈ **~6 GB — guaranteed OOM**, discovered only as a fleet failure;
   - o11 × 2025 ≈ ~55 MB — trivially green.
2. **Deployment capability mismatch.** The live function may simply not speak the request: the 0.27.0 layer shipped without `async-tiff` (#227), and an earlier probe found a stale copy that predates the `shard_key` event rename. A one-invoke capability probe (cheap, e.g. a missing-params `mode="process_raster"` event — the error shape identifies the handler generation) would have caught both instantly.

## Proposal

**Phase 1 — a function** (e.g. `zagg.preflight.check_lambda_run(config, shardmap, function_name)`), called by `RasterStrategy`/`SpatialStrategy` before fan-out and importable standalone from notebooks:

- **Memory model**: per-shard peak estimate from shard geometry + timestep count + band dtypes (raster), or the existing per-shard heuristics (spatial); compare against `GetFunctionConfiguration.MemorySize`.
- **Wall-clock model**: rough per-shard duration estimate vs the function timeout.
- **Capability probe** (opt-in, one invoke): does the deployed handler speak the requested mode/event schema?
[...]

Issue #217, opened 2026-07-14
Chunk-scoped digest streaming: bound worker memory to one inner chunk (unlock coarse HEALPix orders)


espg:
🤖 *from Claude*

## Summary

The streaming aggregator bounds the **read pool** but still holds a running t-digest for **every occupied cell across the whole shard**. That per-cell digest state is the memory floor that makes coarse HEALPix orders (o8, o7) OOM at CONUS scale on the 4 GB Lambda — and `aggregation.streaming: {buffer_granules: N}` cannot touch it (measured: RSS plateaus as the buffer shrinks). Bounding the digest state to **one inner chunk at a time** (process → write → free per chunk) would decouple worker memory from shard area and unlock coarser, cheaper orders.

## Motivation — coarser orders are monotonically cheaper

NEON order sweep (0.24.0 sharded): o8 is ~40% cheaper per obs than o9 (fewer fixed-overhead payments + fewer redundant granule re-reads / less #65 over-assignment); o10 is 2.5× worse. But at CONUS scale o8 OOMs on ~20% of shards at 4 GB, and streaming rescues only the moderately-heavy ones — the densest-coverage tail has a per-cell-digest-state floor >4 GB. Full analysis: `docs/deployment/conus_estimate.md` §4c (PR #207, issue #202).

## Root cause

- The default pooled path holds the whole shard's photons (`worker.py` `all_reads` → `_concat_and_group`). `chunk_inner` only chunks the *output*, applied **after** pooling — it never bounds read/aggregation memory.
- `StreamingAggregator` (`src/zagg/processing/streaming.py`) bounds the read pool to `buffer_granules`, but its state — `self.counts` and `self.digests` (`streaming.py:150-151`) — is keyed by **cell** with **no chunk dimension**. `flush()` (`streaming.py:167-186`) merges into the global per-cell dict and frees nothing until the worker returns. So the digest floor equals the whole shard's occupied cells (∝ shard area). At o8 (4× o9's cell count) the densest shards exceed 4 GB regardless of buffer size.

Measured evidence (5 worst CONUS o8 shards, 4 GB, `buffer_granules` sweep): RSS plateaus as the buffer shrinks (a 120-granule shard sat at ~2,200 MB at both buffer=25 and buffer=12 — the read pool went to ~0, the digest state is the floor). The densest two (148 g, 155 g) OOM at *every* buffer.

## What's already in place

[...]

PR #232, opened 2026-07-15
Bound raster worker memory: cap cross-timestep sampling concurrency + stream slab writes

### LINKED ISSUE #231: Bound raster worker memory: cap cross-timestep sampling concurrency + stream slab writes
🤖 *from Claude*

## Summary

`process_raster_shard` gathers **all** acquisition groups concurrently in one event loop (a phase-3 review fold on PR #219 aimed at wall-clock under the 15-min ceiling). At coarse shard orders that concurrency is unbounded memory: every in-flight timestep holds its decoded COG tiles + per-band gather buffers simultaneously, on top of the accumulated output slabs.

## Empirical numbers (SERC o9 × 2025, first production fleet run, 0.29.0 @ 4096 MB)

- 85 timesteps × 5 bands, 1.05M cells/shard: **unbatched → Runtime.OutOfMemory**, every shard.
- Client-side batching (≤N datatakes per invoke, same worker): **batch 15 → 15/24 invokes OOM; batch 5 → 68/68 clean**, mean 37.5 s/invoke.
- Slabs alone are ~900 MB at 85 timesteps — the OOM margin is dominated by the concurrent sampling buffers, not the slabs.

## Fix (two halves)

1. **Bound the sampling concurrency**: `asyncio.Semaphore(K)` over acquisition groups in `process_raster_shard` (K ~ 4–8), keeping the wall-clock win at fine orders while capping peak at ~K timesteps of buffers.
2. **Stream slab writes in the Lambda handler** (`_handle_process_raster`): write + free each timestep's slab as it completes instead of accumulating all T before writing. Bounds the output side to ~1 timestep and is what makes **all-time** runs (~600 datatakes ≈ 6 GB accumulated at o9) feasible at any memory size. Local `RasterStrategy` already streams per shard; this brings the worker to parity.

With both, per-shard peak ≈ K × (tiles + gather) + 1 slab — o9 all-time drops from "impossible" to routine, and #228's pre-flight model gets a much friendlier red line.

Refs #218 (pipeline), #228 (pre-flight would have predicted this — these numbers are its first calibration points), #217 (same worker-memory-bounding theme for the aggregation path). The interim workaround (client-side datatake batching, `scripts_s2_serc_timeseries.py --step run-batched`) is exact — every (timestep, shard) slab is an independent object — and doubles as the temporal-partitioning pattern.

espg:
Closes #231. Refs #218 (pipeline), #228, #217.

## What / approach

`process_raster_shard` gathered **all** acquisition groups (timesteps) concurrently in one event loop, and `_handle_process_raster` accumulated all `T` slabs before writing. At coarse shard orders that is unbounded memory (SERC o9 × 2025: 85 timesteps × 5 bands OOM'd every shard at 4096 MB). This bounds both sides.

**Phase 1 — cap sampling concurrency.** An `asyncio.Semaphore(K)` over the acquisition groups in `process_raster_shard`, so at most `K` timesteps of decoded COG tiles + per-band gather buffers are in flight at once. Within a bounded group every item still fans out concurrently (adjacent MGRS tiles of one datatake are seconds apart), so the wall-clock overlap survives at fine orders.

`K` is config-surfaced as `data_source.sample_concurrency`, mirroring the existing `data_source.read_workers` / `data_source.granule_workers` worker knobs: default **4** (same as `granule_workers`), validated at submission in `_validate_raster_config` and re-checked in the worker helper `_sample_concurrency` with the same `int>=1` / bool-trap guard.

**Phase 2 — stream slab writes.** `process_raster_shard` gains an optional `on_slab(t_idx, slab)` callback; when set it builds each timestep's slab as its group completes (`asyncio.as_completed`) and hands it to the sink, which writes + frees it before the next arrives — bounding the output side to ~1 timestep. `_handle_process_raster` and the local `RasterStrategy` both pass it, so the whole sampling→write path streams (single write path across backends). Default (no callback) still accumulates and returns the `{t_idx: slab}` dict, so existing callers/tests are unchanged.

Peak per-shard then ≈ `K × (tiles + gather) + ~1 slab`.

## Phases

- [x] **Phase 1** — bound sampling concurrency with `asyncio.Semaphore(K)`; surface `K` as `data_source.sample_concurrency` (default 4); validation + worker-helper guard; concurrency-bound test.
- [x] **Phase 2** — stream slab writes via an `on_slab` callback in `process_raster_shard`; wire it in `_handle_process_raster` and local `RasterStrategy`; streaming/interleave tests.

## How tested

- `tests/test_raster_pipeline.py::TestSampleConcurrency` — `N=10` single-item groups under `Semaphore(K)` via an instrumented fake `sample_item_async`; parametrized over `k ∈ {1, 3, 10, 12}`, asserting observed peak concurrency `== min(k, N)` (serial at `k=1`, capped in the interior, and reaching `N` when `k≥N`).
- `TestOwnership::test_on_slab_streams_and_matches_dict` — streaming mode returns an empty `slabs` (nothing accumulated) yet the per-timestep slabs handed to the sink match the buffered dict-mode output exactly.
- `test_lambda_raster_handler.py::test_handler_streams_slabs_incrementally` — a fake worker drives the sink per timestep; asserts the handler writes one slab per timestep in stream order with never more than one slab alive (`live["max"] == 1`), and writes coords once after.
- Config: `sample_concurrency` default/override + rejection of `0/-1/True/2.0` at both `validate_config` and the worker helper.
- Existing raster suites stay green — the default (no-callback) path is byte-identical.

Local: `ruff check` / `ruff format --check` clean on changed files; `pytest` green on `test_raster_pipeline.py`, `test_raster_runner.py`, `test_lambda_raster_handler.py`, `test_config.py`, `test_runner.py`.

## Questions for review

1. **Default `K` value.** Chose **4** to mirror `granule_workers`. The issue suggests 4–8; the empirical clean point was client-side datatake batch 5 at 4096 MB (a different axis, but adjacent). Happy to bump to 6/8 if you'd rather trade a little more peak memory for overlap at coarse orders.
2. **Knob placement / name.** Put it on `data_source.sample_concurrency` alongside `read_workers`/`granule_workers`. Alternative: a dedicated worker-config field, or a `*_workers` name for consistency (`timestep_workers`). Preference?
3. **Streaming write in the loop.** The `on_slab` write runs synchronously inside the sampling event loop (single-writer, so a completed slab is written+freed before the next group is drained). This serializes each slab write against the other groups' in-flight COG fetches. That's the intended memory/throughput trade at coarse orders (writes are the ~1-slab tail); flag if you'd prefer offloading the write to an executor (which would relax the ~1-slab bound).
4. Pre-existing, not touched here: `ruff check src tests` flags `N818` on `src/zagg/registry.py`; and `config.py` is ~1978 lines (both predate this PR). Flagging per §4, not fixing.

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -897,10 +897,22 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
         # #231) — parity with the local RasterStrategy's per-shard streaming.
         store = open_store(event["store_path"], **_output_store_kwargs(event))
         wrote = False
+        # Opt-in phase timing (the issue #100 convention, raster flavor):
+        # ``write`` is the sink's accumulated wall and ``sample`` the
+        # remainder — the split the PR #232 double-buffer decision needs
+        # measured. Exact at write_buffer=1 (writes serialize in the loop);
+        # at write_buffer>1 writes overlap sampling, so the remainder is
+        # approximate — A/B on ``duration_s`` instead. Default (no
+        # ``profile`` key) emits nothing: the body stays byte-identical.
+        profile = bool(event.get("profile"))
+        write_s = 0.0
 
         def _write_slab(t_idx: int, slab: dict) -> None:
-            nonlocal wrote
+            nonlocal wrote, write_s
+            w0 = time.time() if profile else 0.0
             write_raster_slab(store, grid, shard_key, t_idx, slab)
+            if profile:
+                write_s += time.time() - w0
             wrote = True
 
         _slabs, meta = process_raster_shard(
@@ -927,6 +939,9 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             "total_obs": meta["timesteps"],
             "duration_s": time.time() - start_time,
         }
+        if profile:
+            total = time.time() - start_time
+            body["phase_timings"] = {"sample": total - write_s, "write": write_s}
         return {"statusCode": 200, "body": json.dumps(body)}
     except Exception as e:
         logger.error(f"raster worker failed: {e}", exc_info=True)
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -890,7 +890,19 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
         shard_key = int(event["shard_key"])
         time_index = {k: int(v) for k, v in event["time_index"].items()}
         source = config.data_source or {}
-        slabs, meta = process_raster_shard(
+
+        # Stream the slabs: open the store up front and write + free each
+        # timestep's slab as ``process_raster_shard`` completes its acquisition
+        # group, so peak output memory holds ~1 slab instead of all T (issue
+        # #231) — parity with the local RasterStrategy's per-shard streaming.
+        store = open_store(event["store_path"], **_output_store_kwargs(event))
+        wrote = {"any": False}
+
+        def _write_slab(t_idx: int, slab: dict) -> None:
+            write_raster_slab(store, grid, shard_key, t_idx, slab)
+            wrote["any"] = True
+
+        _slabs, meta = process_raster_shard(
             grid,
             shard_key,
             event["granules"],
@@ -898,12 +910,9 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             time_index,
             region=source.get("source_region"),
             anonymous=source.get("anonymous", True),
+            on_slab=_write_slab,
         )
-
-        store = open_store(event["store_path"], **_output_store_kwargs(event))
-        for t_idx, slab in slabs.items():
-            write_raster_slab(store, grid, shard_key, t_idx, slab)
-        if slabs:
+        if wrote["any"]:
             write_raster_coords(store, grid, shard_key)
 
         body = {
@@ -913,7 +922,7 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             "skipped": meta["skipped"],
             # The shared summary accumulators key on these two names: a raster
             # shard's observation tally is its shard x timestep slab count.
-            "cells_with_data": grid.cells_per_shard if slabs else 0,
+            "cells_with_data": grid.cells_per_shard if wrote["any"] else 0,
             "total_obs": meta["timesteps"],
             "duration_s": time.time() - start_time,
         }
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -896,11 +896,12 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
         # group, so peak output memory holds ~1 slab instead of all T (issue
         # #231) — parity with the local RasterStrategy's per-shard streaming.
         store = open_store(event["store_path"], **_output_store_kwargs(event))
-        wrote = {"any": False}
+        wrote = False
 
         def _write_slab(t_idx: int, slab: dict) -> None:
+            nonlocal wrote
             write_raster_slab(store, grid, shard_key, t_idx, slab)
-            wrote["any"] = True
+            wrote = True
 
         _slabs, meta = process_raster_shard(
             grid,
@@ -912,7 +913,7 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             anonymous=source.get("anonymous", True),
             on_slab=_write_slab,
         )
-        if wrote["any"]:
+        if wrote:
             write_raster_coords(store, grid, shard_key)
 
         body = {
@@ -922,7 +923,7 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             "skipped": meta["skipped"],
             # The shared summary accumulators key on these two names: a raster
             # shard's observation tally is its shard x timestep slab count.
-            "cells_with_data": grid.cells_per_shard if wrote["any"] else 0,
+            "cells_with_data": grid.cells_per_shard if wrote else 0,
             "total_obs": meta["timesteps"],
             "duration_s": time.time() - start_time,
         }
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -351,6 +351,29 @@ def _chord2(lons, lats, lon0: float, lat0: float) -> np.ndarray:
     return x * x + y * y + z * z
 
 
+# Default cap on how many acquisition groups sample concurrently per shard
+# (issue #231). Mirrors ``data_source.granule_workers``' default of 4: every
+# in-flight group holds one timestep's decoded COG tiles + per-band gather
+# buffers, so the cap bounds peak sampling memory to ~K timesteps instead of
+# all T at once, while still overlapping S3 fetches at fine orders.
+_DEFAULT_SAMPLE_CONCURRENCY = 4
+
+
+def _sample_concurrency(config) -> int:
+    """``data_source.sample_concurrency``: acquisition groups in flight per shard.
+
+    Bounds the :class:`asyncio.Semaphore` over timesteps in
+    :func:`process_raster_shard` (issue #231). Default 4; ``1`` samples one
+    timestep at a time. Re-checked here with the same int>=1 / bool-trap guard
+    ``validate_config`` applies at submission, so a hand-rolled worker payload
+    fails loudly rather than passing a bad width to ``Semaphore``.
+    """
+    k = (config.data_source or {}).get("sample_concurrency", _DEFAULT_SAMPLE_CONCURRENCY)
+    if isinstance(k, bool) or not isinstance(k, int) or k < 1:
+        raise ValueError(f"data_source.sample_concurrency must be an integer >= 1 (got {k!r})")
+    return k
+
+
 def process_raster_shard(
     grid,
     shard_key: int,
@@ -402,12 +425,21 @@ def process_raster_shard(
                 "the shards were dispatched from — see raster_time_index"
             )
 
-    # One event loop per shard: fan out across every item of every timestep
-    # concurrently (S3 fetches overlap), instead of one asyncio.run per item.
+    # One event loop per shard, but bound how many acquisition groups sample
+    # concurrently: an ``asyncio.Semaphore(K)`` over the timesteps caps peak
+    # memory at ~K in-flight timesteps of decoded COG tiles + per-band gather
+    # buffers instead of all T at once (issue #231). Within a bounded group
+    # every item still fans out concurrently — adjacent MGRS tiles of one
+    # datatake are seconds apart, so the wall-clock overlap survives at fine
+    # orders.
+    k = _sample_concurrency(config)
+
     async def _sample_all():
-        return await asyncio.gather(
-            *[
-                asyncio.gather(
+        sem = asyncio.Semaphore(k)
+
+        async def _sample_group(items):
+            async with sem:
+                return await asyncio.gather(
                     *[
                         sample_item_async(
                             grid,
@@ -421,9 +453,8 @@ def process_raster_shard(
                         for e in items
                     ]
                 )
-                for items in groups.values()
-            ]
-        )
+
+        return await asyncio.gather(*[_sample_group(items) for items in groups.values()])
 
     group_results = _run_sync(_sample_all())
 
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -352,15 +352,17 @@ def _chord2(lons, lats, lon0: float, lat0: float) -> np.ndarray:
 
 
 # Default cap on how many acquisition groups sample concurrently per shard
-# (issue #231). Mirrors ``data_source.granule_workers``' default of 4: every
+# (issue #231). One knob per pipeline family (issue #232): ``shard_workers`` is
+# "source units in flight per shard" — granules on the spatial path, acquisition
+# groups (timesteps) here — mirroring the spatial default of 4: every
 # in-flight group holds one timestep's decoded COG tiles + per-band gather
 # buffers, so the cap bounds peak sampling memory to ~K timesteps instead of
 # all T at once, while still overlapping S3 fetches at fine orders.
-_DEFAULT_SAMPLE_CONCURRENCY = 4
+_DEFAULT_SHARD_WORKERS = 4
 
 
-def _sample_concurrency(config) -> int:
-    """``data_source.sample_concurrency``: acquisition groups in flight per shard.
+def _shard_workers(config) -> int:
+    """``data_source.shard_workers``: acquisition groups in flight per shard.
 
     Bounds the :class:`asyncio.Semaphore` over timesteps in
     :func:`process_raster_shard` (issue #231). Default 4; ``1`` samples one
@@ -368,10 +370,31 @@ def _sample_concurrency(config) -> int:
     ``validate_config`` applies at submission, so a hand-rolled worker payload
     fails loudly rather than passing a bad width to ``Semaphore``.
     """
-    k = (config.data_source or {}).get("sample_concurrency", _DEFAULT_SAMPLE_CONCURRENCY)
+    k = (config.data_source or {}).get("shard_workers", _DEFAULT_SHARD_WORKERS)
     if isinstance(k, bool) or not isinstance(k, int) or k < 1:
-        raise ValueError(f"data_source.sample_concurrency must be an integer >= 1 (got {k!r})")
+        raise ValueError(f"data_source.shard_workers must be an integer >= 1 (got {k!r})")
     return k
+
+
+# Streamed-write slab budget (PR #232 review): ``1`` is the strict serial
+# bound — a completed slab is written and freed before the next group drains.
+# ``N`` allows N-1 writes in flight on worker threads while the next slab
+# builds, so peak output memory holds <= N slabs; write latency then overlaps
+# sampling instead of serializing against it.
+_DEFAULT_WRITE_BUFFER = 1
+
+
+def _write_buffer(config) -> int:
+    """``data_source.write_buffer``: max slabs alive under a streamed sink.
+
+    Only meaningful when ``process_raster_shard`` runs with ``on_slab``;
+    dict-mode accumulation ignores it. Same int>=1 / bool-trap guard as the
+    sibling worker knobs.
+    """
+    n = (config.data_source or {}).get("write_buffer", _DEFAULT_WRITE_BUFFER)
+    if isinstance(n, bool) or not isinstance(n, int) or n < 1:
+        raise ValueError(f"data_source.write_buffer must be an integer >= 1 (got {n!r})")
+    return n
 
 
 def process_raster_shard(
@@ -402,10 +425,13 @@ def process_raster_shard(
     on_slab : callable, optional
         ``on_slab(t_idx, slab)`` sink invoked once per timestep as its
         acquisition group completes (issue #231). When given, the slab is
-        handed off and dropped immediately, so peak output memory holds ~1
-        timestep instead of all ``T``; the returned ``slabs`` is then empty.
-        When ``None`` (the default) every slab accumulates into ``slabs`` and
-        is returned, as before.
+        handed off and dropped immediately, so peak output memory holds
+        ``data_source.write_buffer`` slabs (default 1: written + freed before
+        the next group drains; ``N`` runs up to ``N-1`` sink calls on worker
+        threads so write latency overlaps sampling — the PR #232
+        double-buffer); the returned ``slabs`` is then empty. When ``None``
+        (the default) every slab accumulates into ``slabs`` and is returned,
+        as before, and ``write_buffer`` is ignored.
 
     Returns
     -------
@@ -444,7 +470,8 @@ def process_raster_shard(
     # every item still fans out concurrently — adjacent MGRS tiles of one
     # datatake are seconds apart, so the wall-clock overlap survives at fine
     # orders.
-    k = _sample_concurrency(config)
+    k = _shard_workers(config)
+    wb = _write_buffer(config)
 
     async def _run_all():
         sem = asyncio.Semaphore(k)
@@ -469,28 +496,53 @@ def process_raster_shard(
 
         lonlat = None  # computed once, only if some timestep has overlapping items
         slabs: dict = {}
+        # Streamed-sink hand-off. At the default ``write_buffer`` of 1 the
+        # sink runs synchronously in the loop: a completed slab is written +
+        # freed before the next group drains (the strict issue #231 bound).
+        # At N>1 (the PR #232 double-buffer) up to N-1 sink calls run on
+        # worker threads while the next slab builds — <= N slabs alive, write
+        # latency overlapped with sampling. A sink error surfaces at most one
+        # group late, at the next hand-off (or the final drain below).
+        pending: list = []
+
+        async def _emit(t, slab):
+            if wb <= 1:
+                on_slab(t, slab)
+                return
+            while len(pending) >= wb - 1:
+                await pending.pop(0)
+            pending.append(asyncio.create_task(asyncio.to_thread(on_slab, t, slab)))
+
         # Drain groups as they finish (as_completed): build each timestep's slab
-        # and hand it to the sink, so a completed slab is written + freed before
-        # the next arrives — the output side stays ~1 timestep (issue #231).
-        for fut in asyncio.as_completed(
-            [_sample_group(key, items) for key, items in groups.items()]
-        ):
-            t, sampled = await fut
-            if len(sampled) == 1:
-                values, valid, _center = sampled[0]
-            else:
-                if lonlat is None:
-                    lonlat = grid.cell_lonlat(cells)
-                values, valid = _combine_by_ownership(sampled, lonlat, bands)
-            slab = {}
-            for f, v in values.items():
-                out = v.copy()  # keep the asset dtype (np.where would promote)
-                out[~valid] = bands[f]["fill_value"]
-                slab[f] = out
-            if on_slab is not None:
-                on_slab(t, slab)  # write + free this timestep before the next
-            else:
-                slabs[t] = slab
+        # and hand it to the sink — the output side stays ~write_buffer
+        # timesteps (issues #231/#232).
+        try:
+            for fut in asyncio.as_completed(
+                [_sample_group(key, items) for key, items in groups.items()]
+            ):
+                t, sampled = await fut
+                if len(sampled) == 1:
+                    values, valid, _center = sampled[0]
+                else:
+                    if lonlat is None:
+                        lonlat = grid.cell_lonlat(cells)
+                    values, valid = _combine_by_ownership(sampled, lonlat, bands)
+                slab = {}
+                for f, v in values.items():
+                    out = v.copy()  # keep the asset dtype (np.where would promote)
+                    out[~valid] = bands[f]["fill_value"]
+                    slab[f] = out
+                if on_slab is not None:
+                    await _emit(t, slab)
+                else:
+                    slabs[t] = slab
+        except BaseException:
+            # Reap in-flight writes before propagating the primary error, so
+            # no task is left un-awaited (their own errors are secondary here).
+            await asyncio.gather(*pending, return_exceptions=True)
+            raise
+        if pending:
+            await asyncio.gather(*pending)  # propagate any trailing write error
         return slabs
 
     slabs = _run_sync(_run_all())
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -383,6 +383,7 @@ def process_raster_shard(
     *,
     region: str | None = None,
     anonymous: bool = True,
+    on_slab=None,
 ):
     """Process one shard of a raster pipeline: every acquisition group -> slab.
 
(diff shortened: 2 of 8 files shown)


src/zagg/processing/raster.py
```diff
@@ -402,12 +425,21 @@ def process_raster_shard(
...
+    # every item still fans out concurrently — adjacent MGRS tiles of one
+    # datatake are seconds apart, so the wall-clock overlap survives at fine
+    # orders.
+    k = _sample_concurrency(config)
+
     async def _sample_all():
-        return await asyncio.gather(
-            *[
-                asyncio.gather(
+        sem = asyncio.Semaphore(k)
```
espg, line 477: 🤖 *from Claude (review)*

Correctness confirmed (positive note, no change needed): the `Semaphore(k)` is constructed *inside* `_sample_all`, i.e. inside the coroutine that `_run_sync` hands to `asyncio.run` (a fresh loop, possibly on a worker thread). That sidesteps the classic "Semaphore bound to a different loop" trap — good that it wasn't hoisted to module/`process_raster_shard` scope. The `async with sem` wraps the per-group inner `gather`, so at most `k` groups' inner fan-outs run at once while items within a group stay unbounded — exactly the cross-timestep bound Phase 1 asks for. Group ordering is preserved (`gather(*[_sample_group(items) for items in groups.values()])` is order-stable), so the downstream `zip(groups.items(), group_results)` alignment and `timesteps` metadata are unchanged.

---

src/zagg/processing/raster.py
```diff
@@ -421,9 +453,8 @@ async def _sample_all():
                         for e in items
                     ]
                 )
-                for items in groups.values()
-            ]
-        )
+
+        return await asyncio.gather(*[_sample_group(items) for items in groups.values()])
 
     group_results = _run_sync(_sample_all())
```

Scope note (informational, not a fault — flagging so the boundary with Phase 2 is explicit). The Semaphore bounds the *transient* sampling memory — each in-flight `sample_item_async` decodes its COG tiles/windows and frees them on return, and those decoded tiles are the dominant per-timestep buffer the issue calls out, so capping them to ~`k` is the real Phase 1 win and it is achieved here.

But `group_results = _run_sync(_sample_all())` still `gather`s across *all* T groups, so every group's *returned* `(values, valid, center)` — per-cell arrays, ~slab-sized — is held simultaneously before the slab-build loop runs. So the returned-buffer side is bounded to T, not `k`; it collapses into the same accumulation the returned `{t_idx: slab}` dict already has. That is precisely Phase 2's `on_slab`/`as_completed` remit (returned dict accumulation is called out as Phase 2 in the PR body), so nothing to change now — just confirming Phase 1 bounds the tiles, not the returned per-cell buffers, and the "~K × (tiles + gather)" peak claim holds for the tiles but the gather buffers are the T-accumulation Phase 2 removes.

---

tests/test_raster_pipeline.py
```diff
@@ -141,6 +158,58 @@ def test_missing_datetime_raises(self):
...
+        cfg = _raster_config(bands={"red": {"asset": "red", "dtype": "uint16"}}, nodata=None)
+        cfg.data_source["sample_concurrency"] = k
+        granules = [
+            _entry(f"g{i}", {"red": f"r{i}.tif"}, T0, time_key=f"dt-{i}") for i in range(n_groups)
+        ]
+        index, _ = raster_time_index([granules])
+        slabs, meta = process_raster_shard(_FakeGrid(n_cells), 0, granules, cfg, index)
+        assert meta["timesteps"] == n_groups
+        assert set(slabs) == set(range(n_groups))
+        assert state["max"] == k  # bounded by K, and the fan-out reaches K
```
espg: 🤖 *from Claude (review)*

The test genuinely proves the bound at `k=3` — `state["max"] == k` asserts both that the peak never exceeds `K` *and* that the fan-out actually reaches `K` (not serialized), which is the right double-sided check. Two suggestions to fold in (cheap, `pytest.mark.parametrize` over `(k, expected_peak)`):

1. **Boundary cases uncovered.** Only the interior `k=3 < n_groups=10` is exercised. Add `k=1` (asserts serial, peak `== 1`) and `k >= n_groups` (asserts peak `== n_groups`, i.e. the cap doesn't cap when it shouldn't). Those are the two ends the reviewer will worry about — an off-by-one in the semaphore width (e.g. `Semaphore(k-1)` or an accidental extra release) would pass the `k=3` case but fail `k=1`.

2. **Timing robustness (minor).** The exact `== k` on the *reach-K* side depends on ≥`k` fakes overlapping inside the 0.02 s `sleep`. It's sound here (the semaphore hard-caps the `<= k` side regardless, and 0.02 s is generous for 3 already-scheduled tasks), so not flaky in practice — but if the `k=1` case is added, note its peak is deterministic (no overlap needed), whereas the reach-`K` cases stay timing-dependent. Worth a one-line comment so a future reader doesn't shorten the sleep.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial self-review of Phase 1 (bound cross-timestep sampling concurrency). Phase 2 (streaming slab writes) is out of scope and correctly left as an unchecked checklist item — not faulted here.

**Verdict: Phase 1 is correct and does what the issue asks.** The `asyncio.Semaphore(k)` bounds concurrent *acquisition groups* (not just items within a group), is constructed inside the running loop, preserves group ordering, and leaves the ownership/slab-build path and `timesteps` metadata untouched. The config knob faithfully mirrors `read_workers`/`granule_workers`: submission-time validation in `_validate_raster_config` plus a worker-side re-check in `_sample_concurrency`, both with the same `int >= 1` / bool-trap guard; default `4` is sane. No new deps, no `noqa`/`type: ignore`, `raster.py` stays at ~605 lines.

Findings (all diff-scoped / fold into the Phase 2 commit — none block advancing):

1. **`raster.py` (semaphore block)** — positive confirmation, no change: Semaphore is created inside the `_run_sync` coroutine (fresh loop), so no cross-loop trap; ordering + metadata preserved.
2. **`raster.py` (`group_results = _run_sync(...)`)** — informational scope note: the Semaphore bounds the *transient decoded COG tiles* (the dominant sampling buffer) to ~`k`, which is the real Phase 1 win and is achieved; but the outer `gather` still materializes all T groups' *returned* per-cell buffers at once. That returned-buffer accumulation is Phase 2's `on_slab`/`as_completed` remit and is already called out in the PR body — confirming the boundary, not asking for a change now.
3. **`test_raster_pipeline.py` (`assert state["max"] == k`)** — the `k=3` test is a sound double-sided bound check, but add `k=1` (serial, peak `== 1`) and `k >= n_groups` (peak `== n_groups`) via `parametrize`: an off-by-one in the semaphore width would pass `k=3` yet fail `k=1`. Timing note on the sleep-based reach-`K` assertion inline.

Open items left standing for @espg (raised as questions in the PR body, no action from me): the default `K` value (4 vs 6/8) and the knob name/placement (`data_source.sample_concurrency` vs a `*_workers` name). Separately, pre-existing and not this PR's doing: `config.py` is ~1978 lines, well past the ~1000-line guideline — a standing item, not introduced here.

---

src/zagg/processing/raster.py
```diff
@@ -421,33 +465,40 @@ async def _sample_all():
...
+                if lonlat is None:
+                    lonlat = grid.cell_lonlat(cells)
+                values, valid = _combine_by_ownership(sampled, lonlat, bands)
+            slab = {}
+            for f, v in values.items():
+                out = v.copy()  # keep the asset dtype (np.where would promote)
+                out[~valid] = bands[f]["fill_value"]
+                slab[f] = out
+            if on_slab is not None:
+                on_slab(t, slab)  # write + free this timestep before the next
```

Confirming the streaming bound is real, with one nuance worth recording. `as_completed` eagerly schedules all N `_sample_group` tasks, but `Semaphore(k)` gates the heavy work (`async with sem:` wraps the per-item gather), so only K groups ever hold decoded tiles + gather buffers at once. Completed-but-undrained results (`sampled`) do **not** pile up to N: `as_completed`'s done-callback pops each task off `todo`/the done-queue as it finishes, and `t, sampled = await fut` rebinds the local each loop, so a completed task is freed once drained. Critically, `on_slab` is a *synchronous* zarr write that blocks this event loop, which halts further sampling while it runs — that back-pressure is what keeps the completed-undrained set bounded to ~K rather than unbounded.

Net worst-case live ≈ K (sampling) + up to ~K (completed, awaiting drain) + 1 (slab being written) = O(K), not "all T" — solves the OOM as intended. The only nuance: it's ~2K+1 timesteps of buffers in the worst micro-timing, not literally "~1 timestep on the sampling side"; the "~1 timestep" comment is accurate for the *output/dict* side (only the slab being written exists off the sampling path). No change requested — just noting the bound rests on the write being synchronous/blocking; if a future change made `on_slab` non-blocking, completed results could accumulate.

---

src/zagg/processing/raster.py
```diff
@@ -421,33 +465,40 @@ async def _sample_all():
...
+                slabs[t] = slab
+        return slabs
+
+    slabs = _run_sync(_run_all())
     metadata = {
         "shard_key": int(shard_key),
         "granule_count": len(granules),
         "skipped": skipped,
-        "timesteps": len(slabs),
+        "timesteps": len(groups),
```
espg, line 553: 🤖 *from Claude (review)*

`"timesteps": len(groups)` is correct. In stream mode `slabs` is empty so the old `len(slabs)` would report 0; `len(groups)` is the true per-shard slab count and is identical in both modes (every group builds exactly one slab in the drain loop, streamed or stored). Also confirming criterion (4): the slab build — including the lazy `grid.cell_lonlat(cells)` init and `_combine_by_ownership` — runs only in the single `_run_all` driver coroutine (the `as_completed` loop), never in the concurrent `_sample_group` tasks (those call `sample_item_async` only), so no race is introduced and the build stays single-threaded, same as the pre-Phase-2 post-gather section.

---

deployment/aws/lambda_handler.py
```diff
@@ -890,20 +890,29 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
...
         time_index = {k: int(v) for k, v in event["time_index"].items()}
         source = config.data_source or {}
-        slabs, meta = process_raster_shard(
+
+        # Stream the slabs: open the store up front and write + free each
+        # timestep's slab as ``process_raster_shard`` completes its acquisition
+        # group, so peak output memory holds ~1 slab instead of all T (issue
+        # #231) — parity with the local RasterStrategy's per-shard streaming.
+        store = open_store(event["store_path"], **_output_store_kwargs(event))
+        wrote = {"any": False}
```

Minor style: the handler uses a `wrote = {"any": False}` dict-box while the identical pattern in `runner.py` `_one` uses `nonlocal wrote` on a plain bool. A `nonlocal` here would match the runner and read a touch cleaner (the closure never needs the dict). Non-blocking.

Confirming the streaming safety otherwise: `on_slab` (`_write_slab` → `write_raster_slab`) runs blocking zarr v3 sync I/O from inside the `_run_all` asyncio driver, which is safe — zarr's `sync()` dispatches the coroutine to its own dedicated background loop thread (not the current `asyncio.run` loop), so it will not raise "cannot be called from a running event loop." This is the same sync-facade contract already relied on elsewhere in this module.

---

deployment/aws/lambda_handler.py
```diff
@@ -890,20 +890,29 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
...
             region=source.get("source_region"),
             anonymous=source.get("anonymous", True),
+            on_slab=_write_slab,
         )
-
-        store = open_store(event["store_path"], **_output_store_kwargs(event))
-        for t_idx, slab in slabs.items():
-            write_raster_slab(store, grid, shard_key, t_idx, slab)
-        if slabs:
+        if wrote["any"]:
```

Partial-write / coords-gating semantics look correct and acceptable. A group failing mid-stream now leaves earlier timesteps already written (vs the old all-or-nothing), but each `(t, shard)` slab is an independent, chunk-aligned object and the shard is retried/overwritten wholesale, so leftover slabs are harmless. `write_raster_coords` is gated on `wrote["any"]`, which is set only inside `_write_slab` — so on a mid-stream failure the exception short-circuits before this line and coords is skipped (rewritten on the successful retry), and a no-data shard (zero groups → `on_slab` never called → `wrote["any"]` stays False) also skips coords, matching prior behavior. `test_worker_failure_500` still exercises the single-group failure → 500 path through the streamed sink.

---

src/zagg/runner.py
```diff
@@ -684,12 +684,25 @@ def run(
...
+            # Stream: write + free each timestep's slab as it completes (issue
+            # #231), so a shard holds ~1 slab, not all T.
+            _slabs, meta = process_raster_shard(
+                grid,
+                int(shard_key),
+                granules,
+                config,
+                time_index,
+                on_slab=_write_slab,
+                **src_kwargs,
```
espg, line 703: 🤖 *from Claude (review)*

Confirming the local multi-shard case is safe. Each `_one` runs in a pool thread; `process_raster_shard` → `_run_sync` → `asyncio.run` gets that thread its own event loop, and `_write_slab` → `write_raster_slab(zarr_store, ...)` reaches zarr's shared background loop via `run_coroutine_threadsafe` (thread-safe). Concurrent shards write disjoint, chunk-aligned regions of the shared `zarr_store` (distinct `shard_key` → distinct cell range; distinct `t` index), so there's no read-modify-write contention. `nonlocal wrote` and the write both run on the same pool thread, so no cross-thread race on the flag. This is the same concurrency model as the pre-Phase-2 buffered local path — Phase 2 only moved the write inside the sampling loop.

---

tests/test_lambda_raster_handler.py
```diff
@@ -117,6 +117,52 @@ def test_end_to_end_slab_write(self, handler_mod, raster_event):
...
+        monkeypatch.setattr(
+            raster_mod, "write_raster_slab", lambda store, g, sk, t, slab: writes.append(t)
+        )
+        monkeypatch.setattr(
+            raster_mod, "write_raster_coords", lambda *a, **k: coords_calls.append(1)
+        )
+
+        resp = handler_mod.lambda_handler(event, MagicMock())
+        assert resp["statusCode"] == 200, resp
+        assert writes == [0, 1, 2]  # one write per timestep, in stream order
```
espg, line 158: 🤖 *from Claude (review)*

`live["max"] == 1` is somewhat an artifact of the fake's own shape: the fake does `live["cur"] += 1; on_slab(t); live["cur"] -= 1` sequentially, so the max is 1 by construction regardless of what the real driver does. What this test genuinely proves is the *handler's* streaming wiring — that it passes a real sink (`assert on_slab is not None`), writes once per timestep in order (`writes == [0,1,2]`), frees between writes, and writes coords exactly once after (`coords_calls == [1]`). That's the handler's Phase-2 contribution and is worth having.

The gap (non-blocking): the real `process_raster_shard` `as_completed` drain — the thing that actually bounds completed-but-unwritten slabs to ~K — has no direct memory assertion; it's only covered indirectly by `test_on_slab_streams_and_matches_dict` (data-equivalence + empty return) and by `TestSampleConcurrency` (the semaphore's sampling bound). A true "completed slabs don't pile up" assertion on the real driver is hard to write deterministically, so I'd treat this as an acceptable coverage gap rather than a required addition. Note `test_end_to_end_slab_write` now runs through the streamed path (the handler always passes `on_slab`) and still verifies written slab + cell_ids values, so the streamed path is exercised end-to-end.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial self-review of **Phase 2** (streaming slab writes) for issue #231. No correctness bug or regression found — the streaming refactor is sound and preserves the buffered path byte-for-byte. Notes below are all non-blocking (confirmations + one style nit + one test-coverage nuance); details inline.

**Correctness (criteria 1–4) — all check out:**
1. `on_slab` truly bounds memory to O(K), not all-T. `Semaphore(k)` gates the heavy per-item gather inside `_sample_group`; `as_completed` drains + frees each `sampled`/slab as it completes (done-queue pops the task, the local rebinds each loop). Because `on_slab` is a *synchronous* zarr write that blocks the event loop, sampling back-pressures and completed-but-undrained results stay bounded to ~K — worst-case live ≈ 2K+1 timesteps of buffers, still O(K). `"timesteps": len(groups)` is the right fix for the empty-`slabs` stream mode (identical to `len(slabs)` in buffered mode).
2. Blocking zarr v3 sync I/O inside the running `_run_all` loop is safe: zarr's `sync()` runs on its own background loop thread, so no "called from a running event loop." Local `RasterStrategy` streams concurrent shards to one shared `zarr_store` from the pool — disjoint chunk-aligned regions, no contention, same model as before Phase 2.
3. Partial-write on mid-stream failure is acceptable (independent `(t,shard)` objects, retried/overwritten). `write_raster_coords` correctly gated on the `wrote` flag in both handler and runner: skipped on mid-stream failure (exception short-circuits) and on no-data shards (zero groups), matching prior behavior.
4. No new race — slab build, lazy `grid.cell_lonlat`, and `_combine_by_ownership` all run single-threaded in the `_run_all` driver, not in the concurrent `_sample_group` tasks. Phase 1's semaphore + config knob remain correct under the Phase 2 refactor.

**Tests (criterion 5):** `test_on_slab_streams_and_matches_dict` proves data-equivalence + empty return; `test_end_to_end_slab_write` now exercises the streamed path end-to-end and still verifies written values. `test_handler_streams_slabs_incrementally` validates the *handler's* write-then-free wiring, but its `live["max"] == 1` is an artifact of the fake's structure, not a proof of the real driver's ~K bound — an acceptable, hard-to-close coverage gap (inline).

**Conventions (criterion 6):** raster.py ~615 lines (< 1000), no new deps, no broad `noqa`/`type: ignore`. One trivial style nit: handler's `wrote = {"any": False}` dict-box vs runner's `nonlocal wrote` bool for the identical purpose (inline).

---

espg:
> Default K value. Chose 4 to mirror granule_workers. The issue suggests 4–8; the empirical clean point was client-side datatake batch 5 at 4096 MB (a different axis, but adjacent). Happy to bump to 6/8 if you'd rather trade a little more peak memory for overlap at coarse orders.

seems great, also is close to empirical optimum

> Knob placement / name. Put it on data_source.sample_concurrency alongside read_workers/granule_workers. Alternative: a dedicated worker-config field, or a *_workers name for consistency (timestep_workers). Preference?

kinda seems like the same shape as read_workers/granule_workers ... can we make this more general? i.e., just shard_workers which hits for both the atl03 K threaded workers, and also here for K readers? Thoughts?

> Streaming write in the loop. The on_slab write runs synchronously inside the sampling event loop (single-writer, so a completed slab is written+freed before the next group is drained). This serializes each slab write against the other groups' in-flight COG fetches. That's the intended memory/throughput trade at coarse orders (writes are the ~1-slab tail); flag if you'd prefer offloading the write to an executor (which would relax the ~1-slab bound).

can you provide more context here? What's the trade space between the options-- any dangers? Do you have a lean-- and if so, why?

> Pre-existing, not touched here: ruff check src tests flags N818 on src/zagg/registry.py; and config.py is ~1978 lines (both predate this PR). Flagging per §4, not fixing.

don't care, this is fine

espg:
🤖 *from Claude*

Answering the three open notes:

**1. `shard_workers` generalization — yes, and your framing is the right one.** `granule_workers` (spatial: h5coro granules in flight per shard) and `sample_concurrency` (raster: acquisition groups in flight per shard) are the same knob semantically: *source units in flight per shard*. One name, per-pipeline unit. Two things to keep in mind folding it:
- **Back-compat**: `granule_workers` is in shipped configs and the #180 docs — accept `shard_workers` as the canonical key with the legacy keys as silent fallbacks (canonical wins when both present; no deprecation noise yet).
- **One number, two memory meanings**: a spatial "unit" is a granule read buffer; a raster "unit" is K timesteps × bands of decoded COG tiles. Same knob is fine — defaults already differ appropriately per path — but the docstring should say explicitly that the memory cost per unit is pipeline-dependent, so #228's pre-flight can price it per pipeline rather than assuming one profile.

**2. Streaming-write placement trade space.**
- **Sync in-loop (current)**: a completed slab is written + freed before the next group drains. Guarantees the hard **≤1 slab alive** bound and gives free failure semantics (a write error raises in-line, in-order, before more work is paid for). Cost: the event loop stalls for the write (~100–300 ms per slab in-region), serializing write latency against sampling *coordination*. Empirically negligible at current geometry: batch-5 workers averaged 37.5 s with writes a tiny tail; worst case at 85 timesteps is ~15–25 s of serialized writes per shard against minutes of sampling.
- **Executor offload**: overlaps write I/O with sampling — but the bound is only preserved with explicit backpressure (a bounded handoff), otherwise a throttled S3 lets the queue grow *exactly at the coarse-order geometry where the bound is the point*. It also moves write failures off the sampling path: they must be joined before return or a lost slab is silent. More machinery, and the failure mode lands in the memory-constrained case.
- **Danger summary**: offload's risks (unbounded queue under throttling, deferred error propagation) concentrate precisely where this PR's guarantee matters most.

**Lean: keep the synchronous in-loop write.** The ≤1-slab bound *is* the deliverable, the wall-clock cost is empirically small, and the middle ground exists if data ever says otherwise: a 1-deep double-buffer (sample group N+1 while writing N, cap = 2 slabs) recovers most of the overlap while staying bounded — worth doing only with profiling that shows writes dominating, which today's numbers do not.

(K=4 default: acknowledged as accepted, near the empirical clean point.)

espg:
> middle ground exists if data ever says otherwise: a 1-deep double-buffer (sample group N+1 while writing N, cap = 2 slabs) recovers most of the overlap while staying bounded 

...this sounds worthwhile, if it saves "~15–25 s" of per shard wall time (am I reading that correctly)?

espg:
🤖 *from Claude*

You're reading it right, with two qualifications that matter for the decision:

**Where the number comes from.** The ~15–25 s is (T−1) × per-slab write latency at the geometry #232 *enables* — a full-year o9 shard in ONE invoke (T=85, ~5 arrays × ~2 MB per slab, in-region puts estimated at ~200–300 ms per slab set). The double-buffer recovers essentially all of it except the final write. Against an estimated 3–6 min of sampling for that invoke, that's **~5–10% of per-shard wall (and GB-seconds)** — real, recurring on every raster fleet run, but not transformative. At today's operating point (batch-5 invokes), T=5 makes it ~1 s/invoke — noise. The win exists only where #232's bound lets T grow.

**The honest caveat: the write latency is estimated, not measured.** If per-slab writes are actually ~500 ms (5 sequential array puts), the saving doubles toward ~40 s (~15% of wall); if they're ~100 ms, it halves. The worker already has opt-in phase timing (#100) — one instrumented full-year-shard run after this PR merges gives the real write-tail number for free.

**Suggested disposition:** land #232 as-is (the bound is the deliverable), then one instrumented run; if the measured write tail is ≥ ~10% of invoke wall, add the 1-deep double-buffer as a small follow-up. The bounded form is genuinely tame — one pending `asyncio.to_thread` write future, awaited before the next write starts, so errors surface at most one group late and the slab bound moves from 1 to exactly 2 (the docstring's peak formula gains `+1 slab`). I'd skip straight to implementing it now only if you'd rather not spend the instrumented run — it's low-risk either way; the sequencing just keeps the memory-bound PR pure and lets the follow-up cite a measured number.

espg:
🤖 *from Claude*

All three review threads folded, in three commits:

- `dac52c5` — **`shard_workers` unification, spatial side**: `_granule_workers` now reads the canonical `shard_workers` key with the shipped `granule_workers` name as a silent legacy fallback (canonical wins when both set); precedence test added.
- `a000e19` — **raster phase timing**: `_handle_process_raster` gains the #100-style opt-in `profile` key, emitting `phase_timings: {sample, write}` (write = the streamed sink's accumulated wall). Exact at `write_buffer=1`; approximate above it (writes overlap sampling — A/B on `duration_s` there). Default payload byte-identical.
- `2173c92` — **`shard_workers` config surface + the `write_buffer` double-buffer** (espg-suggested, folded now rather than as a follow-up): `sample_concurrency` is renamed to `shard_workers` outright (it never shipped, so no alias needed on the raster side), and `data_source.write_buffer` lands with the proposed semantics — `1` (default) keeps this PR's strict write-and-free bound; `N` runs up to `N−1` sink calls on worker threads while the next slab builds, so peak output memory is ≤ N slabs and write latency overlaps sampling. Sink errors surface at most one group late (next hand-off or the final drain); on a sampling error in-flight writes are reaped before the primary exception propagates. Tests: bounded-overlap instrumentation at `wb ∈ {2,3}` (max sink concurrency ≤ wb−1, streamed output equals dict-mode exactly), validation matrix, sink-error propagation.

Measurement plan after merge + release: one full-year o9 shard, `profile: true`, A/B `write_buffer: 1` vs `2` — the write-tail number decides whether the template default ever moves off `1`.

espg:
🤖 *from Claude*

**Provenance record: the `write_buffer` double-buffer fold (commit `2173c92`).**

Decision chain, for future archaeology:

1. **Origin** — the [streaming-write trade-space reply](https://github.com/englacial/zagg/pull/232#issuecomment-4985233835) to espg's review notes ([notes](https://github.com/englacial/zagg/pull/232#issuecomment-4985019849)) named a "1-deep double-buffer (sample group N+1 while writing N, cap = 2 slabs)" as the bounded middle ground between the synchronous in-loop write (strict ≤1-slab bound, serialized write latency) and unbounded executor offload (overlap, but backpressure/error hazards).
2. **Sizing question** — espg [asked](https://github.com/englacial/zagg/pull/232#issuecomment-4985368861) whether the ~15–25 s/shard saving read correctly; the [answer](https://github.com/englacial/zagg/pull/232#issuecomment-4985377818): yes at the full-year o9 single-invoke geometry this PR enables (~5–10% of invoke wall/GB-s), with the caveat that per-slab write latency was estimated, not measured.
3. **Decision (espg, session direction, 2026-07-15)** — implement now rather than as a measured follow-up, *as a parameter*: "buffer=N where 1 is no buffer, and 2 is two slabs, etc; it sounds like a pretty small addition, and this will let us fold it in now and assess impact at test time." Accepted: parameterizing with default 1 keeps this PR's strict bound as the default behavior while making the A/B a flag flip instead of a release cycle.
4. **What landed** (`2173c92`) — `data_source.write_buffer` (default **1** = the unchanged issue-#231 bound; `N` = ≤N slabs alive, up to N−1 sink calls on `asyncio.to_thread` workers overlapping the sampling loop). Sink errors surface at most one group late (next hand-off or final drain); a sampling error reaps in-flight writes before propagating. Validated at submission and in the worker helper (int ≥ 1, bool-trap). Tests: bounded-overlap instrumentation at `wb ∈ {2,3}` (max sink concurrency ≤ wb−1; streamed output byte-equal to dict mode), validation matrix, sink-error propagation.
5. **Open gate** — the default stays `1` until the post-release measurement (one full-year o9 shard, `profile: true`, A/B `write_buffer: 1` vs `2` using `a000e19`'s `phase_timings` split) says the write tail clears ~10% of invoke wall; only then does any template default move.

espg:
🤖 *from Claude*

**Measurement gate result: `write_buffer` 1 vs 2 A/B on the fleet (0.30.0, layer :32, 4096 MB), NEON SERC AOI, full-year o9 shards — one invoke per shard (85 datatakes × 5 bands), `profile: true`.**

First, the headline that isn't the A/B: the geometry that motivated #231 now runs clean — 85-timestep single-invoke shards, 4/4 ok in both arms, where pre-#232 every one OOM'd.

| arm | mean wall | mean sample | mean write | write % of wall |
|---|---|---|---|---|
| `write_buffer: 1` | 578.5 s | 524.0 s | 54.5 s (serialized) | **9.4%** |
| `write_buffer: 2` | 555.6 s | 446.3 s | 109.3 s (overlapped, approx.) | — |

Per-shard wall deltas (wb1 → wb2): −11.0, −23.5, −31.1, −26.2 s — **consistent sign across all four shards, mean −22.9 s (−4.0%)** of wall and GB-seconds.

Reading:
- The measured write tail at wb=1 is **9.4%** — right at the ~10% bar, and per-slab latency is ~640 ms, about 2× the estimate in the [sizing discussion](https://github.com/englacial/zagg/pull/232#issuecomment-4985377818).
- wb=2 recovers **~40% of that tail** (≈23 of ≈55 s), not all of it: with one write in flight, writes queue when slabs complete faster than ~640 ms apart, and the overlapped write wall inflates under GIL contention (visible as the 109 s accumulated write at wb=2 — the documented approximate-split regime).
- Caveats: arm 1 ran first (cold containers may flatter arm 2 by a few seconds/shard); n=4 per arm.

**Disposition per the gate:** realized saving is 4.0% — under the 10% bar, so **template defaults stay at `write_buffer: 1`** (the strict bound). The knob is there for long-timeseries runs where 4% of a large fleet bill is worth taking; a `wb=3` arm might claw back a little more of the tail but with visibly diminishing returns and a 3-slab bound, untested here.

espg:
🤖 *from Claude*

**Extended A/B: the full `write_buffer ∈ {1..4}` curve + memory price** (same setup: 0.30.0 fleet @ 4096 MB, NEON SERC AOI, full-year o9 shards, one profiled invoke per shard; peak memory from CloudWatch REPORT lines):

| arm | mean wall (ok) | errors | write share | peak mem (median / max) |
|---|---|---|---|---|
| `wb=1` | 578.5 s | 0 | 9.4% (serialized) | 1428 / 1440 MB |
| `wb=2` | **555.6 s (−4.0%)** | 0 | overlapped | 1738 / 1743 MB (**+~300 MB**) |
| `wb=3` | 573.3 s (3 ok) | **1 × Sandbox.Timedout @ 900 s** | 31.9% (inflated) | 1426 / 1440 MB |
| `wb=4` | 658.6 s (**+13.8%**) | 0 (one at 802.8 s, near-ceiling) | 32.0% (inflated) | 1458 / 1460 MB |

**The curve inverts past 2.** The write walls balloon (55 s → 180–240 s accumulated) as concurrent zstd-compressing write threads contend with the sampling loop for the worker's 2 vCPUs — sampling itself slows (visible in the wb=4 sample walls), and the variance widens enough to hit the 900 s ceiling once at wb=3.

**Memory price**: wb=2 costs ~+300 MB peak (1.44 → 1.74 GB) — more than the naive 1-extra-slab (~11 MB at o9), because the loop keeps building ahead while writes hold compression buffers. wb=3/4 *don't* climb further (contention throttles the pipeline before memory can stack). Two couplings for the record: at the 4096 MB worker wb=2 is comfortable (~43%); on a hypothetical 2048 MB variant (issue #235) wb=1 already sits at ~70% and wb=2 would be ~85% — the #228 pre-flight model should carry both terms (baseline ~1.4 GB at o9/K=4, +~300 MB for the second slab-in-flight).

**Final disposition:** defaults stay `write_buffer: 1`; documented guidance is that `2` is the only value worth setting (−4% wall/GB-s for +300 MB peak), and `≥3` is actively harmful on the 2-vCPU worker. Timeout-risk note: long-timeseries single-invoke shards at any wb run 470–800 s against the 900 s ceiling — temporal batching stays the safety valve.

PR #233, opened 2026-07-15
Make sharded: true the default for HEALPix output

### LINKED ISSUE #215: Make sharded: true the default for HEALPix output (missing flag = silent ~250x object blow-up)
🤖 *from Claude*

## Proposal: make `sharded: true` the default for HEALPix output

`sharded` currently defaults to **False** (`src/zagg/grids/healpix.py:77`). When a config omits the flag, every output array is written **one object per inner chunk** (~K per shard) instead of collapsing to **one object per shard** via the `ShardingCodec`. At the benchmark geometry (o9, `chunk_inner: 13` → K=256), that's a **~250× object blow-up per array**, which inflates S3 PUT count/cost, write wall time, and `zarr.json` metadata — and it happens **silently**, from a *missing* line rather than an explicit choice.

## Why now — this session hit it three ways

1. **7 of the tdigest benchmark configs silently emit unsharded output** (missing the flag): `atl03_tdigest_healpix_o8.yaml`, `o11.yaml`, `o9_cached.yaml`, `o10_cached.yaml`, `o11_cached.yaml`, `atl03_tdigest_rect_3km.yaml`, `atl03_tdigest_rect_6km.yaml`. Only `o9.yaml` / `o10.yaml` carry it (from #193).
2. **It skewed the CONUS cost estimate (#202).** The CONUS regression uses `o9_cached.yaml`, so it ran unsharded — ~256 objects/shard/array — and missed the full #209/#211 write benefit. Re-running with the flag added is what finally measures the real number.
3. **It confounded a benchmark comparison.** The o8/o9/o10 order sweep ran **o8 unsharded (2,048 t-digest chunk objects)** while **o9/o10 ran sharded (4 objects)** — verified in the 0.24.0 output. So a head-to-head "does coarser sharding cost less" comparison was silently apples-to-oranges: o8 paid the unsharded write bloat that o9 didn't.

A missing flag should not cost a 250× object blow-up or invalidate a comparison. The safe state is sharded.

## Proposal

1. **Flip the default** to `sharded: true`. The existing guard (`healpix.py:127`, `if sharded and self.chunks_per_shard <= 1`) already makes it a **no-op when there's ≤1 chunk per shard**, so single-chunk grids are unaffected — the default only bites when there's something to collapse.
2. **Fix the 7 configs regardless** — add `sharded: true` so they match #193's intent (this can land immediately, independent of the default change).
3. Consider a one-release **warning** when a multi-chunk shard is written unsharded, so any remaining opt-out is explicit.

## Considerations

- **Reader compat:** #209/#211 updated the t-digest reader for the sharded vlen path; confirm the dense-array and rectilinear read paths are equally sharding-agnostic before flipping.
- **Existing unsharded stores** stay readable (zarr reads both); this only changes *new writes*.
- **Rectilinear:** the rect configs are also in the missing-flag set — confirm `sharded` is honored on the rect engine (or scope the default to HEALPix first).

Refs: #193 (turned sharding on for the matrix, but per-config), #209 / #211 (the ragged t-digest sharded vlen write), #202 (where the unsharded CONUS estimate + confounded o8 sweep surfaced). espg to steer.

espg:
Closes #215

## What / approach

`sharded` defaulted **False** for HEALPix output, so a config that omitted the flag wrote **one object per inner chunk** (~K per shard) instead of collapsing to **one object per shard** via the `ShardingCodec` — a silent ~K-fold object blow-up (e.g. ~250× at the o9 `chunk_inner: 13` geometry) driven by a *missing line*. This flips the safe default to sharded, scoped to **HEALPix only** (rectilinear default untouched — see Questions for review).

Changes:

- **`HealpixGrid.__init__`**: constructor default `sharded=False → True`.
- **The K==1 guard becomes a no-op instead of an error.** Previously `sharded=True` with `chunks_per_shard <= 1` raised; now that the default is True, a single-chunk grid (no `chunk_inner`) has nothing to bundle, so sharding is silently disabled there. This keeps single-chunk grids byte-identical to a pre-#215 unsharded write — the default "only bites when there's something to collapse."
- **`get_sharded(config, default=False)`** gains a `default` parameter. Its own default stays `False` (so a bare `get_sharded(config)` is unchanged), but `from_config` passes the HEALPix default in.
- **`from_config`** (HEALPix branch): the omitted-flag default is now `True` for **flat** layout. **Hive stays unsharded** (`store_layout: hive` → default False): each hive leaf is a vanilla zarr v3 store (D3) and `emit_shard_template` rejects sharded output, and `validate_config` already rejects an *explicit* `sharded: true` + hive. Without this carve-out a hive config that omitted the flag would validate but crash at write. Rectilinear branch left on the `False` default (scope: HEALPix).
- **Benchmark configs**: added an explicit `sharded: true` to the 6 HEALPix configs that still omitted it (see checklist). The new default already covers them, but the explicit line documents intent (matching #193) and doesn't rely on the default.

### Reader compatibility (confirmed before flipping)

- **Dense-array reads** go through plain zarr indexing; the `ShardingCodec` is recorded in the array metadata, so zarr reads sharded and regular arrays through the same path.
- **T-digest / ragged reads** (`readers/tdigest_tensor.py`) are explicitly sharding-agnostic: the read plan derives its span from `arr.shards or arr.chunks` (self-describing metadata), "so the sharded and per-inner-chunk layouts share this path" (#209/#211).
- Existing unsharded stores stay readable; only *new writes* change.

## Phases

- [x] **Phase 1** — flip the HEALPix default (constructor + `from_config`), make the K==1 guard a no-op, keep hive unsharded, add/update tests.
- [x] **Phase 2** — add explicit `sharded: true` to the 6 HEALPix benchmark configs that omitted it; confirm reader compatibility (above).

## Configs touched (Phase 2)

Added `sharded: true` (HEALPix, K>1, flat):

- `atl03_gain_bias_healpix_o10.yaml`
- `atl03_gain_bias_healpix_o11.yaml`
- `atl03_scalar_healpix_o11.yaml`
- `atl03_tdigest_healpix_o10_cached.yaml`
- `atl03_tdigest_healpix_o11.yaml`
- `atl03_tdigest_healpix_o11_cached.yaml`

Note on the issue's named list: `atl03_tdigest_healpix_o8.yaml`, `o9_cached.yaml`, and the plain `o9`/`o10` already carry `sharded: true` in the current tree (added in earlier work — the o8 one is even commented "issue #215"), so they needed no change. The two rectilinear configs the issue lists (`atl03_tdigest_rect_3km.yaml`, `_6km.yaml`) are intentionally left untouched (scope: HEALPix — see Questions).

## How tested

- `ruff check` / `ruff format --check` clean on all changed files.
- Full suite green locally: **1930 passed, 28 skipped** (`pytest -q`).
- New/updated tests in `tests/test_grids.py`:
  - K>1 omitted flag now defaults sharded; K==1 omitted flag is a no-op (unsharded); hive omitted flag stays unsharded; explicit `sharded: false` still opts out; explicit `sharded: true` + K==1 no-ops (was: raised); omitted-flag default template equals explicit `sharded: true`.
- `tests/test_processing.py`: the unsharded multi-chunk worker tests and the "regular"/hive-leaf reference grids now pin `sharded: false` explicitly (that path is the opt-out now), preserving their original coverage.

## Questions for review

1. **Rectilinear default (deliberately out of scope).** I confirmed the rect engine *does* honor `sharded` (full `ShardingCodec` path in `RectilinearGrid`), but per the issue ("scope the default to HEALPix first") I left the rect default and the two rect benchmark configs unchanged. Want the rect default flipped / the rect configs' flag added in a follow-up, or fold it here?
2. **Explicit `sharded: true` + K==1 now silently no-ops** rather than raising a `ValueError`. This is what made the True default safe for single-chunk grids, and matches the issue's "the guard makes it a no-op" framing — but it does drop a config-mistake signal for someone who explicitly asks for sharding on a K==1 grid. OK, or would you prefer to keep an error only for the *explicit* opt-in case (would need a tri-state default to distinguish explicit from defaulted)?
3. **One-release opt-out warning (issue item 3, optional).** I did **not** add a warning when a multi-chunk shard is written unsharded: with the default now True the only way to hit that is an explicit `sharded: false`, and hive leaves are legitimately unsharded at K>1 (`emit_shard_template`), so a blanket warning would fire on every hive leaf. Flagging as a follow-up rather than landing something noisy — want it as a narrower flat-layout-only warning instead?

---
_Generated by [Claude Code](https://claude.ai/code/session_011vwrYZY38grfwJqtCeuM65)_

### DIFF
--- a/src/zagg/grids/healpix.py
+++ b/src/zagg/grids/healpix.py
@@ -74,7 +74,7 @@ class HealpixGrid:
         config: PipelineConfig | None = None,
         populated_shards: list[int] | None = None,
         chunk_inner: int | None = None,
-        sharded: bool = False,
+        sharded: bool = True,
         shard_order: int | None = None,
     ):
         if child_order < parent_order:
@@ -121,17 +121,13 @@ class HealpixGrid:
         # Sharded storage (issue #108): bundle the shard's K inner chunks into one
         # zarr ShardingCodec shard object instead of K independent regular chunk
         # objects. Only meaningful when K > 1 (a finer ``chunk_inner`` gives the
-        # shard multiple inner chunks); sharding a K==1 shard is a no-op shard of
-        # one chunk, so reject it with a clear message (validated pre-deployment,
-        # matching the grid-mismatch errors). HEALPix lands first (issue #108).
-        if sharded and self.chunks_per_shard <= 1:
-            raise ValueError(
-                "sharded output requires chunk_inner to give more than one inner "
-                f"chunk per shard (K>1), but chunk_order ({chunk_order}) == "
-                f"parent_order ({parent_order}) gives K=1; set chunk_inner finer "
-                "than parent_order (e.g. chunk_inner=13 for the 64x64 read chunk) "
-                "or disable sharded"
-            )
+        # shard multiple inner chunks). ``sharded`` defaults True (issue #215 — a
+        # missing flag should not silently cost the ~K-fold object blow-up), so a
+        # K==1 shard (no ``chunk_inner``) has nothing to bundle: sharding is a no-op
+        # there and is silently disabled, leaving single-chunk grids byte-identical
+        # to a pre-#215 unsharded write. HEALPix lands first (issue #108).
+        if self.chunks_per_shard <= 1:
+            sharded = False
         self.sharded = bool(sharded)
         # shard_order (issue #133 phase 8): decouple the sharding OBJECT from the
         # dispatch shard. A ShardingCodec object normally spans the whole dispatch
--- a/src/zagg/grids/healpix.py
+++ b/src/zagg/grids/healpix.py
@@ -123,9 +123,10 @@ class HealpixGrid:
         # objects. Only meaningful when K > 1 (a finer ``chunk_inner`` gives the
         # shard multiple inner chunks). ``sharded`` defaults True (issue #215 — a
         # missing flag should not silently cost the ~K-fold object blow-up), so a
-        # K==1 shard (no ``chunk_inner``) has nothing to bundle: sharding is a no-op
-        # there and is silently disabled, leaving single-chunk grids byte-identical
-        # to a pre-#215 unsharded write. HEALPix lands first (issue #108).
+        # K==1 shard (``chunk_inner`` unset, or no finer than ``parent_order``) has
+        # nothing to bundle: sharding is a no-op there and is silently disabled,
+        # leaving single-chunk grids byte-identical to a pre-#215 unsharded write.
+        # HEALPix lands first (issue #108).
         if self.chunks_per_shard <= 1:
             sharded = False
         self.sharded = bool(sharded)
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1721,16 +1721,18 @@ def get_layout(config: PipelineConfig) -> str:
     return config.output.get("grid", {}).get("layout", "fullsphere")
 
 
-def get_sharded(config: PipelineConfig) -> bool:
+def get_sharded(config: PipelineConfig, default: bool = False) -> bool:
     """Return whether the output grid uses ShardingCodec storage (issue #108).
 
     The ``sharded`` knob lives on the grid/chunk block next to ``chunk_inner``
-    (mirroring its accessor), default ``False``. When ``True`` the grid bundles a
-    dispatch shard's K inner chunks into one zarr shard object instead of K
-    independent regular chunk objects; it is only valid when ``chunk_inner`` gives
-    K>1 (the grid raises otherwise, validated before deployment).
+    (mirroring its accessor). When ``True`` the grid bundles a dispatch shard's K
+    inner chunks into one zarr shard object instead of K independent regular chunk
+    objects; a K==1 grid has nothing to bundle, so the grid silently no-ops it
+    (issue #215). ``default`` is the value returned when the flag is omitted —
+    ``False`` here, but ``from_config`` passes ``True`` for HEALPix flat-layout
+    output (issue #215: a missing flag should not cost the ~K-fold object blow-up).
     """
-    return bool(config.output.get("grid", {}).get("sharded", False))
+    return bool(config.output.get("grid", {}).get("sharded", default))
 
 
 def get_shard_order(config: PipelineConfig) -> int | None:
--- a/src/zagg/grids/__init__.py
+++ b/src/zagg/grids/__init__.py
@@ -4,7 +4,13 @@ from __future__ import annotations
 
 import warnings
 
-from zagg.config import PipelineConfig, get_child_order, get_shard_order, get_sharded
+from zagg.config import (
+    PipelineConfig,
+    get_child_order,
+    get_shard_order,
+    get_sharded,
+    get_store_layout,
+)
 from zagg.grids.base import InconsistentShardError, OutputGrid, ShardKey
 from zagg.grids.healpix import HEALPIX_BASE_CELLS, HealpixGrid
 from zagg.grids.rectilinear import OOB_SENTINEL, RectilinearGrid
@@ -46,6 +52,13 @@ def from_config(
         resolved_parent = grid_cfg.get("parent_order", parent_order)
         if resolved_parent is None:
             raise ValueError("output.grid.parent_order is required for HEALPix grids")
+        # issue #215: HEALPix flat-layout output defaults to sharded — a missing
+        # flag should not silently cost the ~K-fold object blow-up (one object per
+        # inner chunk instead of one per shard). Hive is inherently unsharded (each
+        # leaf is a vanilla zarr v3 store — D3), so it keeps the False default; an
+        # explicit sharded:true with hive is rejected by ``validate_config``. The
+        # grid no-ops sharding when K==1, so single-chunk grids stay unaffected.
+        sharded_default = get_store_layout(config) != "hive"
         return HealpixGrid(
             parent_order=resolved_parent,
             child_order=get_child_order(config),
@@ -53,7 +66,7 @@ def from_config(
             config=config,
             populated_shards=populated_shards,
             chunk_inner=grid_cfg.get("chunk_inner"),
-            sharded=get_sharded(config),
+            sharded=get_sharded(config, default=sharded_default),
             shard_order=get_shard_order(config),
         )
     if grid_type == "rectilinear":
--- a/tests/data/benchmark/configs/atl03_gain_bias_healpix_o10.yaml
+++ b/tests/data/benchmark/configs/atl03_gain_bias_healpix_o10.yaml
@@ -127,4 +127,5 @@ output:
     indexing_scheme: nested
     parent_order: 10             # shard / dispatch unit: 4^(19-10) = 512x512 cells
     chunk_inner: 13              # inner Zarr chunk: 4^(19-13) = 64x64 cells (K=16 per shard)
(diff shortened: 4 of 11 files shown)


src/zagg/grids/healpix.py
```diff
@@ -121,17 +121,13 @@ def __init__(
...
-                "than parent_order (e.g. chunk_inner=13 for the 64x64 read chunk) "
-                "or disable sharded"
-            )
+        # shard multiple inner chunks). ``sharded`` defaults True (issue #215 — a
+        # missing flag should not silently cost the ~K-fold object blow-up), so a
+        # K==1 shard (no ``chunk_inner``) has nothing to bundle: sharding is a no-op
+        # there and is silently disabled, leaving single-chunk grids byte-identical
+        # to a pre-#215 unsharded write. HEALPix lands first (issue #108).
+        if self.chunks_per_shard <= 1:
+            sharded = False
```
espg, line 131: 🤖 *from Claude (review)*

The K==1 no-op is necessary and correct for the *default* path (flipping the constructor default to True would otherwise break every single-chunk grid). One behavioral regression is worth a conscious decision, though: this branch now also silently swallows an **explicit** `sharded: true` on a K==1 grid. Pre-#215 that case raised a pointed error ("set chunk_inner finer than parent_order (e.g. chunk_inner=13) …"), which caught the common "I asked for sharding but forgot `chunk_inner`" misconfig. Now the request is dropped with no signal.

The grid can't distinguish default-True from explicit-True because `from_config` resolves the default before constructing the grid (`get_sharded(config, default=sharded_default)` in `grids/__init__.py`), so preserving the error would mean threading the "was it explicit?" distinction down (or having `from_config` emit a warning when it downgrades an explicit `sharded: true` at K==1).

Not a blocker and possibly intended, but flagging for a decision: is silently ignoring an explicit `sharded: true` acceptable, or should the explicit-misconfig case still warn? A test pinning whichever behavior you choose would lock it in.

---

src/zagg/grids/healpix.py
```diff
@@ -121,17 +121,13 @@ def __init__(
...
-            raise ValueError(
-                "sharded output requires chunk_inner to give more than one inner "
-                f"chunk per shard (K>1), but chunk_order ({chunk_order}) == "
-                f"parent_order ({parent_order}) gives K=1; set chunk_inner finer "
-                "than parent_order (e.g. chunk_inner=13 for the 64x64 read chunk) "
-                "or disable sharded"
-            )
+        # shard multiple inner chunks). ``sharded`` defaults True (issue #215 — a
+        # missing flag should not silently cost the ~K-fold object blow-up), so a
+        # K==1 shard (no ``chunk_inner``) has nothing to bundle: sharding is a no-op
```

Doc nit: the parenthetical "a K==1 shard (no `chunk_inner`)" is slightly narrow — K==1 also arises when `chunk_inner` **is** set but equals `parent_order` (then `chunk_order == parent_order` and `chunks_per_shard == 1`). The guard itself (`self.chunks_per_shard <= 1`) correctly covers both; only the comment reads as if K==1 requires an absent `chunk_inner`. Consider "(no finer `chunk_inner`, i.e. chunk_order == parent_order)".

---

espg (commented):
🤖 *from Claude (review)*

Fresh-eyes adversarial review of the #215 "sharded default for HEALPix" change. Overall this is clean, correct, and well-tested — I found no correctness bug and scope stayed HEALPix-only. Verified locally: `ruff check`/`format --check` clean on the five changed source/test files (the one remaining ruff error, N818 in `registry.py`, is pre-existing and unrelated), and the affected suites are green (`test_grids.py::TestFromConfig`/`TestSharded`, `test_processing.py::TestMultiChunkWorker`, `test_readers.py`, `test_runner.py`, `test_hive.py`, `test_rectilinear.py`).

What I checked and where it holds up:
- **K==1 guard** (`healpix.py`) — the raise→no-op conversion is required by the default flip and correctly forces `sharded=False` before `self.sharded` is set; single-chunk grids stay byte-identical (covered by `test_flag_off_byte_identical_to_regular`, `test_sharded_default_matches_explicit_true`). One inline note on the explicit-`sharded:true`+K==1 case.
- **from_config hive carve-out** (`grids/__init__.py:61`) — `get_store_layout(config) != "hive"` correctly keeps hive at the False default, consistent with `validate_config`'s hive+sharded rejection (which reads the *raw* `grid.get("sharded")`, so the new resolved default never trips it). Covered by `test_sharded_default_off_for_hive`.
- **Scope stayed HEALPix-only** — the rectilinear `from_config` branch still calls `get_sharded(config)` with the False default and `RectilinearGrid` still *raises* on K==1 sharded; guarded by the existing `test_rectilinear.py::test_sharded_default_off`. See the asymmetry note below.
- **`get_sharded(config, default=...)`** — new param is the only signature change; the sole callers are the two `from_config` branches, and `validate_config` bypasses the accessor, so no stale-signature risk.
- **Reader compatibility** — dense-array and t-digest/ragged read paths are sharding-agnostic and already guarded by `test_readers.py::test_sharded_and_regular_layouts_read_identically`; `shard_order`/`emit_shard_template` are unaffected.
- **Benchmark configs** — all K>1 HEALPix configs now carry `sharded: true`; the 6 added here are exactly the ones that were still missing it (the rest were set under #193/#209/#202). All are flat + HEALPix + K=16/64, none hive or rectilinear. Consistent and complete.

Two non-blocking items, both inline:
1. `healpix.py` — the K==1 no-op now silently swallows an *explicit* `sharded: true` that pre-#215 raised a pointed error; worth a conscious decision on whether the explicit-misconfig case should still warn.
2. `healpix.py:126` — minor comment-precision nit (K==1 also occurs at `chunk_inner == parent_order`, not only when absent).

One repo-wide note (no inline anchor): HEALPix now silently no-ops K==1 sharding while rectilinear still raises on the same case. That divergence is correct under this issue's HEALPix-only scope, but it leaves the two grid families inconsistent — worth a tracking issue so the rect side is a deliberate follow-up rather than an accidental gap.

---

espg:
>  Want the rect default flipped / the rect configs' flag added in a follow-up, or fold it here?

fold it here

> Explicit sharded: true + K==1 now silently no-ops rather than raising a ValueError

this is fine, document behavior in user docs

>  I did not add a warning when a multi-chunk shard is written unsharded: with the default now True the only way to hit that is an explicit sharded: false, and hive leaves are legitimately unsharded at K>1 (emit_shard_template)

what do you mean "and hive leaves are legitimately unsharded at K>1 " ; in general, or with an explicit sharded: false?

espg:
ugh, why is this so complicated-- followup coming for hive / flat parity... apparently sharded as default doesn't make sense  or work for dense raster case...

espg:
🤖 *from Claude*

> what do you mean "and hive leaves are legitimately unsharded at K>1"; in general, or with an explicit `sharded: false`?

In general — pre-#236, unsharded is hive's *only* mode at any K, flag or no flag. Three layers enforce it: `from_config` resolves the hive default to `False` (the carve-out at `grids/__init__.py:61`), `validate_config` rejects explicit `sharded: true` + hive (`config.py:355`), and `emit_shard_template` raises on sharded leaf templates (`grids/healpix.py:620`). So at K>1 every hive run writes multi-chunk unsharded dense arrays, and that state carries no opt-out signal from the user — a "multi-chunk shard written unsharded" warning would have fired on 100% of hive runs while flagging nothing anyone chose. That's why I left the warning out rather than special-case hive.

#236 removes exactly this asymmetry: hive defaults sharded like flat, all three carve-outs above get deleted, and an explicit `sharded: false` becomes the same conscious opt-out on both layouts — at which point a warning, if still wanted, has a well-defined trigger. Implementation plan posted on #236.

PR #234, opened 2026-07-15
Wire per-release full-AOI NEON benchmark (issue #202 leg 1)

espg:
Refs #202 (leg 1). Wire the **per-release full-AOI NEON** benchmark that `docs/deployment/benchmark.md` documents but never actually ran — the "Per-release full-AOI NEON (issue #202 leg 1)" section stayed a skeleton across releases because three pieces were never connected: no release trigger, nothing invoked the harness, and the renderer was a `return False` stub. This closes all three.

The per-**merge** live matrix (one densest shard/target, isolating code deltas from data drift) already runs on `lambda-benchmark.yml`. This adds its sibling: the whole `AOP_NEON` box fanned over **every** shard, recorded once per **release** (`push: tags '*.*.*'`), for real dollar-cost truth across all shards.

## Phases

- [x] **Phase 1 — drop `finalize (s)` from the rendered tables** (`f4fe430`). It crowded `matrix_table.png` / `codec_table.png` for ~no signal. Removed from `TABLE_HEADERS` + `format_record_cells`; **`finalize_s` stays in the parquet series** (issue #191 regression column) — only the rendered table loses it.
- [x] **Phase 2 — full-AOI release series append** (`c9b21fc`). New `full_aoi_series.py`: a separate `full_aoi_series.parquet` on the `benchmarks` branch (the whole-AOI record schema differs materially from the single-shard series — `n_shards`, whole-AOI `cost_usd`, `total_wall_s`, write-throughput), retained on `event == "release"`, deduped `(commit, target)`.
- [x] **Phase 3 — render + wire** (`f525cfb`, guard fold `e5036be`). `make_full_aoi_release_figure` (was a stub): per-release 2×2 (inline/sidecar × mask/nomask), whole-AOI total cost **and** AOI-average cost/100 km² per panel, **release tag on the x-axis**, wall on the right axis. Generalized `_draw_panel`/`_render_panel_grid` with `runtime_col`/`label_col`/`label_len` (defaults keep the per-merge callers byte-identical). `benchmark.md` skeleton → live embeds.
- [x] **Phase 4 — pin NEON catalog + validate** (`63a96b3`). `tests/data/benchmark/catalogs/cat_neon.parquet` (2089 granules, the AOI-bbox+temporal subset of the local full ATL03 catalog — the `cat_88s.parquet` precedent) so the harness builds the whole-AOI shardmap **offline** in CI. Pinning it fixes the granule set: the per-release series then tracks **code** change over the whole AOI, not CMR drift. Validated end-to-end via `--dry-run` (4 shards, per-shard `[63,64,67,92]`, no billing).
- [x] **Phase 5 — tags-triggered workflow** (`5157935`). `lambda-benchmark-fullaoi.yml` on `push: tags '*.*.*'` + `workflow_dispatch`: run the harness against the stable production `process-shard`, append, render, push to `benchmarks`. Setup documented in `benchmark-cicd.md`.

## Grounding / notes

- **AOI-average vs worst-shard cost.** The new AOI-average cost/100 km² for the synthetic-checked NEON run is `$0.016 × 100 / (4 × 162 km²) ≈ $0.00247` — **below** the live matrix's `$0.003–0.004`, which is the *densest* (worst-case) shard. This is exactly the median-vs-worst gap: the live chart can only show the worst shard scaled to 4 GB; this new series is the honest whole-AOI average.
- **Full-AOI densest = 92 granules**, vs `targets.json`'s single CMR-pinned densest cell at 66. Expected: this is built from the pinned *local* catalog fanned over all 4 o9 shards — a distinct measurement from the single CMR-pinned probe. Determinism verified (identical `[63,64,67,92]` across repeated builds).
- **Reuses the per-merge config; no new vars, no hardcoded account.** The workflow reads the same `BENCHMARK_ROLE_ARN` / `BENCHMARK_AWS_REGION` / `BENCHMARK_STORE_PREFIX` / `BENCHMARK_FUNCTION_NAME` the per-merge run uses, so it runs on the next tag with **nothing extra to set**. It skips (green, no billing) until `BENCHMARK_ROLE_ARN` is set. The old `BENCHMARK_FULLAOI_*` override vars and the pinned `expect_account: 742127912612` were dropped (fork-hostile, and merge/release will never target different accounts); the account guard remains available as an **optional** opt-in (`dispatch.expect_account` in the manifest, no-op when absent).

## Testing

`ruff check` + `ruff format --check` clean under CI-pinned ruff 0.14.10 on all touched files (3 unrelated pre-existing format nits on `main` under newer ruff 0.15.x left untouched per convention). `pytest -m "not slow"` on the benchmark suite → **96 passed**; the slow `test_neon_catalog_builds_full_aoi_shardmap` (end-to-end offline shardmap build) → **1 passed (80s)**. Each phase self-reviewed fresh-context (Opus) — all five clean.

## Questions for review

- **Trigger is `push: tags '*.*.*'`** per your call — it fires **alongside** `publish.yml` (same tag event), benchmarking the currently-deployed stable function. Confirm that's the intended reference point.
- **Recurring AWS billing**: each version tag dispatches all NEON shards (4) to real Lambda. Flagged so it's a conscious choice; `workflow_dispatch` is there for the first validated run.
- **Cross-workflow push to `benchmarks`**: this workflow and `lambda-benchmark.yml` both push to the `benchmarks` branch (separate concurrency groups). A merge + a release landing at the exact same second could non-fast-forward-reject the second push (same no-retry pattern the merge job already has) — rare, visible (red job), re-runnable. Left matching the existing pattern; say if you want a rebase-retry.
- Standing @espg items unchanged: the `benchmarks`-branch series prune and the CLAUDE.md §8 `espg`→`englacial` scope reconciliation (the latter blocked the review subagents from posting to this repo; I posted their clean verdicts on the thread myself).


espg (commented):
🤖 *from Claude (review)*

Phase 1 (commit `f4fe430`, drop `finalize (s)` from rendered tables) reviewed adversarially — clean, nothing to flag.

- **Consumers stay in sync.** Both renderers iterate `TABLE_HEADERS` and index `cells[h]`: `_table_block` (bench_metrics.py:249-251) and `render_latest_table` (plot_series.py:211-213). Since `format_record_cells` drops the `"finalize (s)"` key in lockstep with `TABLE_HEADERS`, no `cells[h]` lookup can KeyError. `plot_series` only positionally indexes `"% cap"` (still present); no code hardcoded the `"finalize (s)"` column or its old index.
- **Series schema unchanged.** `finalize_s` is retained in `RECORD_COLUMNS` (bench_metrics.py:75) and still populated by `build_record` (bench_metrics.py:190 from `summary.get("finalize_s")`). No silent column drop in the parquet series.
- **No dead code / stale comments.** The only remaining `finalize` comment (bench_metrics.py:66-71) describes `finalize_s` as a retained *series/record* column tracked for regression — still accurate. No comment references the removed rendered column.
- **Test is non-vacuous.** `test_wall_breakdown_columns_threaded_and_rendered` now stops accessing `cells["finalize (s)"]` (which would KeyError post-change), still asserts `rec["finalize_s"] == 65.0`, and adds the correct positive/negative pair: `"finalize_s" in RECORD_COLUMNS` and `"finalize (s)" not in TABLE_HEADERS`. It genuinely pins the new behavior.

(pytest not run here — the review env lacks pandas — but the change is a static header/formatter drop verified by reading.)

espg:
🤖 *from Claude (review)*

**Phase 2 adversarial review (`c9b21fc`, full-AOI release series append) — clean.** Fresh-context review cross-checked `full_aoi_series.py` against the harness run record in `run_full_aoi_benchmark.py` (`run_target`'s `run` dict + `_write_throughput()`):

- **Schema faithful** — every scalar the harness emits maps to a `FULL_AOI_COLUMNS` entry; every column traces to a field the harness sets (none perpetually null). The three dropped keys (`temporal`, `per_shard_granules`, `apriori_estimate`) are exactly the non-scalar/planning-only fields.
- **`_WT_MAP`** keys match `_write_throughput()`'s emitted dict exactly.
- **Retention** `event=="release"` is faithful (harness copies `--event` verbatim); filtering before dedup means a non-release row can't evict a retained release point.
- **Dedup** `(commit, target)` keep-last is correct for release cadence — re-dispatch replaces, distinct target stays separate.
- **No `FutureWarning`** on pandas 2.2.3 across the empty/non-empty/all-NA paths; missing parquet → empty column-stable frame.
- Tests non-vacuous (mixed release/merge/"" events; missing-wt; non-list JSON). 7 pass.

Nothing to fold.

espg:
🤖 *from Claude (review)*

**Phase 3 (`f525cfb`) + Phase 4 (`e5036be`, `63a96b3`) adversarial review — clean.** Fresh-context (Opus).

**Phase 3 (render + wire):** the `_draw_panel`/`_render_panel_grid` generalization is backward-compatible — the three per-merge callers pass no new kwargs, so they take defaults equal to the old hardcoded literals (`str(cm)[:7] ≡ [:label_len]` at `label_len=7`); byte-identical. `_full_aoi_history` derives cost/100 km² with clean 0/null→NaN handling (`.where(area>0)`, no RuntimeWarning); `_matrix_layout` + `_aoi_axis` classify the `..._mask`/`..._nomask` target names correctly. Every column `_draw_panel` touches is in the canonical schema. `main` wiring skips cleanly on a missing/empty full-AOI series. One latent guard inconsistency was flagged and folded in `e5036be`.

**Phase 4 (pin NEON catalog):** `cat_neon.parquet` (2089 granules) loads via `Catalog.from_geoparquet` with every column the harness needs; the slow test builds 4 shardmaps end-to-end. **Determinism verified** — identical `per_shard_granules [63,64,67,92]` across repeated builds (order-stable `np.flatnonzero`+`take`, deterministic `ShardMap.build`), so the pinned catalog fixes the granule set across releases. Tests non-vacuous; `slow` marker registered; fixture not LFS/gitignored; content is public granule metadata (the `cat_88s` precedent). Guard fold is behavior-preserving on the canonical path.

Nothing to fold beyond the guard hardening already applied.

.github/workflows/lambda-benchmark-fullaoi.yml
```diff
@@ -0,0 +1,150 @@
...
+        working-directory: _series
+        run: |
+          set -euo pipefail
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add full_aoi_series.parquet site
+          if git diff --cached --quiet; then
+            echo "no full-AOI benchmark changes to commit"
+          else
+            git commit -m "full-aoi benchmark: ${{ github.ref_name }}"
```
espg: 🤖 *from Claude (review)*

Inconsistent with this workflow's own injection-hardening. The run step above (lines 91-93) documents that inputs "flow through env, never interpolated into the script body," and correctly routes `github.ref_name` through the `TAG` env var. But this commit step interpolates `${{ github.ref_name }}` **directly into the shell body**. Unlike the sibling `lambda-benchmark.yml` (which interpolates `github.sha` here — 40 hex chars, non-injectable), `github.ref_name` is the pushed tag, and git tag names are permitted to contain shell metacharacters (`$`, backticks, `;`, `(`, `)`). A tag like `` 1.0.`cmd`.0 `` still matches the `*.*.*` trigger and would execute inside `git commit -m "..."`.

Low severity — pushing a `*.*.*` tag already requires write access (and tag protection), so this is not a fresh privilege. But it contradicts the stated hardening and is trivially fixable: add `TAG: ${{ github.ref_name }}` to this step's `env:` and use `git commit -m "full-aoi benchmark: $TAG"`.

.github/workflows/lambda-benchmark-fullaoi.yml
```diff
@@ -0,0 +1,150 @@
...
+            git push origin benchmarks
```
espg, line 152: 🤖 *from Claude (review)*

Cross-workflow non-fast-forward push hazard on the `benchmarks` branch (noting, not blocking). This workflow and `lambda-benchmark.yml` both `git push origin benchmarks` with no fetch/rebase/retry, and they run under **different** concurrency groups (`lambda-benchmark-fullaoi` vs `lambda-benchmark`) — intentional, so a release and a merge don't cancel each other, but that also means nothing serializes their pushes. If a merge run pushes to `benchmarks` between this job's `checkout ref: benchmarks` and its `git push`, this push is rejected non-fast-forward: the run goes red and the just-billed release series point is lost (the benchmark dollars were already spent).

The sibling has the same no-retry push, but within its own group it is self-serialized; the new failure mode is specifically the two groups racing. Low-to-medium severity given releases are rare and merges skip on doc-only pushes, so the overlap window is small. A `git pull --rebase origin benchmarks` before push (both writers touch disjoint parquets plus a shared `site/`, so a rebase would usually clean-merge), or a shared advisory lock, would close it. Fine to defer, but worth capturing.

espg:
🤖 *from Claude (review)*

Phase 5 self-review folded:

- **Finding 1 (commit-message injection)** — fixed in `336300e`: the release tag now flows through a `TAG` env var in the commit step (`git commit -m "full-aoi benchmark: $TAG"`), matching the workflow's own hardening pattern instead of interpolating `${{ github.ref_name }}` into the shell body.
- **Finding 2 (cross-workflow push race to `benchmarks`)** — left as-is by design: it exactly matches the existing `lambda-benchmark.yml` push pattern (no fetch/rebase/retry), the overlap window is a merge and a release landing in the same second (rare), and it fails loud + re-runnable. Documented under "Questions for review" for @espg to decide if a rebase-retry is wanted.

espg:
🤖 *from Claude*

**Forward note: the hive-vs-flat comparison belongs on a CONUS-scale run, not NEON.**

Design thread with @espg concluded that the per-release full-AOI harness landed here will eventually be the right vehicle to measure the **hive** store layout (issue #199) against **flat** — but only at CONUS/global scale.

Why not NEON: hive's payoff is spreading PUTs across S3 prefix partitions under thousands of shards (S3 throttles ~3,500 PUT/s *per partition*). The NEON box is **4 shards**, so a hive arm here would validate output *parity + per-shard overhead* but show ~zero partition-spreading benefit. The cost/wall advantage only appears at CONUS scale.

Sequencing:
1. **#236** brings hive dense-array output to parity with flat (shard dense arrays in hive leaves — today they're K objects/leaf vs flat's 1). Until that lands, benchmarking hive measures the unintended K-object penalty, not the shipping layout.
2. Once #236 lands: add a **flat-vs-hive** arm (NEON first, for correctness/regression).
3. Then a **CONUS-scale** full-AOI run via this harness is where the partition-spreading / cost / wall advantage actually gets measured.

No change to this PR — recorded so the intent isn't lost.

Issue #237, opened 2026-07-15
Design: hive store layout for the raster pipeline (S2 adopts hive)


espg:
🤖 *from Claude*

Carved out of the 2026-07-15 pipeline-unification discussion. **Ratified up front (espg): yes — as hive becomes the production layout, the Sentinel-2 store adopts it.** This issue scopes the design; it does **not** block #236 or #233 (those cover the aggregation pipeline only).

## Where raster stands today

- `RasterStrategy` bypasses the aggregation write machinery entirely: `(time, cells)` arrays, chunks `(1, cells_per_chunk)`, per-timestep chunk-aligned slab writes (`src/zagg/runner.py:563-758`, `src/zagg/processing/raster.py:664-679`). No read-modify-write anywhere, by construction.
- It never consults `store_layout` — a raster config declaring `store_layout: hive` is **silently ignored** (the raster branch of `validate_config` returns early, before the `store_layout`/`coverage_moc` checks at `src/zagg/config.py:347-382`). A companion small-fix issue is filed alongside this one to make that an explicit rejection until this design lands.
- **Ratified (espg, same discussion): raster keeps its own pipeline default `sharded: false`, and sharding stays disabled on the raster path.** Per-timestep slab streaming over a ShardingCodec object spanning timesteps would read-modify-write that object once per slab; the codec also buys little here — raster object count is dominated by the time axis, which sharding can't collapse without cross-timestep buffering. The hive digit tree is orthogonal to the codec, so hive-raster = **hive tree + unsharded `(time, cells)` leaves**.

## Open design points

1. **D4 commit-stamp semantics under append.** The stamp (`docs/design/sparse_coverage.md` §2) assumes a shard is written once, then stamped complete. Raster shards gain new timesteps across runs (append is declared future work in the `RasterStrategy` docstring, `runner.py:570-571`). Options: (a) the stamp carries a high-water time index, updated per run ("complete through T"); (b) per-run stamp entries under the leaf root attrs; (c) time-partitioned leaves (one leaf per time window), preserving write-once semantics. This needs a decision before hive-raster ships — it's the only place the hive contract and the raster write model genuinely disagree.
[...]

PR #241, opened 2026-07-15
Shard dense arrays in hive leaves (flat-hive parity)

### LINKED ISSUE #236: Shard dense arrays in hive leaves (flat↔hive parity)
🤖 *from Claude*

## Summary

Bring **hive** store-layout output to parity with **flat** by writing dense arrays through the `ShardingCodec` (one object per shard) instead of streaming them as K per-inner-chunk objects. Today a hive leaf writes its dense arrays (e.g. `cell_ids`, `morton`, `count`) as K separate chunk objects, while flat + `sharded: true` collapses them to one object per shard. That K-fold PUT count lands on a **single leaf prefix** (same/adjacent S3 partition) — the opposite of what the hive layout exists to achieve — and diverges hive output from flat.

This is the follow-up carved out of #233 (making `sharded: true` the HEALPix flat default). #233 correctly leaves hive on the unsharded path and has `validate_config` reject `sharded: true` + hive; this issue removes that limitation.

## Why it matters

Hive (issue #199, `docs/design/sparse_coverage.md`) exists to spread PUTs across S3 prefix partitions at CONUS/global scale (~50k o9 shards × ~2k workers; S3 throttles ~3,500 PUT/s **per partition**). But *within* a shard, the dense arrays are still written as K objects to one leaf prefix — so we pay the object count **and** concentrate it on one partition. Sharding the dense arrays gives the intended **1 PUT/shard, morton-prefixed** — the actual goal.

It also unblocks making hive a viable production default with output identical to flat.

## The false premise to discard

The current rejection says *"each hive leaf is a vanilla zarr v3 store (D3); drop sharded."* That reasoning is wrong: **the `ShardingCodec` IS vanilla zarr v3** — a sharded array is spec-compliant and self-describing, and our flat sharded stores already are exactly that. Nothing in the hive design (D1–D5: digit tree, node invariant, D4 commit stamp, "zero metadata above the leaf") is affected by sharding — it's internal to the `{full_id}.zarr` leaf.

**Proof it's already compatible:** the ragged/t-digest field is *already* written sharded inside hive leaves today (one `ShardingCodec` object per leaf, issue #209) — see `src/zagg/grids/healpix.py:732` (the `elif leaf and self.chunks_per_shard > 1:` branch) and `src/zagg/hive.py:728-743`. So a hive leaf already contains a sharded array. Only the **dense** arrays were left on the streaming path.

## The real reason it's currently unsharded (and the tradeoff)

The hive worker `process_and_write_hive` (`src/zagg/hive.py:746`, `_write_chunk`) **streams each dense chunk write-then-free** → O(chunk) worker memory (issue #91). K chunk objects is the direct consequence of streaming. The ragged field is the one exception (accumulate + write once), and the `hive.py:728-743` comment documents *why they stopped there*: accumulation re-adds an O(shard-payload) memory term, accepted for the ragged field (~200 MB peak at o8 t-digest scale) but **deliberately not** for the dense side, to keep it O(chunk).

So the design axis is **PUT-count vs peak-RSS**, decided per field. This issue extends the "accumulate → one object" trade to dense arrays.

**Where dense memory actually matters:** the dense arrays in ICESat-2 aggregations are *small* (a tdigest shard's `cell_ids`+`morton`+`count` ≈ 20 MB at o9; o9 shards run <15% of the 4 GB cap). The o8 OOMs and the 88S stress are **ragged centroid state + read buffers**, NOT dense-array assembly. The genuinely dense payload is **Sentinel-2 raster** (issue #218: a value per cell, no ragged carrier). So for point aggregations this is a near-free win; the memory tradeoff only bites for dense raster at scale — that's the one open design point below.

## The machinery already exists (implementation sketch)

The flat path already has **both** modes, switched on `grid.sharded` — `src/zagg/runner.py:1512-1532`:

```python
chunk_results = [] if sharded else None
_df_out, metadata = process_shard(..., chunk_results=chunk_results,
                                  write_chunk=None if sharded else _write_chunk)
if sharded:
```

Bringing hive to parity is **reuse, not invention**:

1. **Array spec** — add a `leaf and chunks_per_shard > 1` sharding branch to the dense `_shard` in `HealpixGrid` (mirror the existing ragged branch, `healpix.py:732`), so the leaf template declares dense arrays sharded.
2. **Write path** — switch `process_and_write_hive` to the accumulate branch for dense (`chunk_results=[]`, `write_chunk=None`) and call the leaf equivalent of `write_shard_to_zarr` (the same pattern the ragged field already uses). **Note:** the spec change *requires* this — declaring dense sharded but keeping the per-chunk `_write_chunk` would read-modify-write the shard object K times (the failure the flat comment at `runner.py:1513` warns about). They are not separable.
3. **Drop the two rejections** — `validate_config` (`src/zagg/config.py:355`) and `HealpixGrid.emit_shard_template` (`src/zagg/grids/healpix.py:621`). Soften/remove the "does not support sharded output yet" message.

`process_and_write_hive` is a **shared** worker path (both the local runner's `_cell_work` and the Lambda hive branch call it — `runner.py:1643`), so the change is exercised by both dispatchers and needs its tests to cover both.

## Open design point (needs a decision)

Universal dense accumulation is fine for ICESat-2 (small dense), but for **Sentinel-2 raster** a whole-shard dense payload is a real O(shard) memory term. Options:
- **(a)** Always accumulate (simplest; matches flat-sharded, which already pays this) — accept the RSS, verify against the 4 GB cap at raster/o8 density.
- **(b)** Accumulate when the shard's dense payload fits a memory budget, else fall back to streaming (unsharded) — the same "escape valve" the ragged comment names (`hive.py:740-743`).

Recommend measuring (a) first (it's the flat-sharded profile we already run) and only building (b) if a raster/high-density shard blows the budget.

## Acceptance criteria

- A hive tdigest run produces **byte-parity** output with the flat sharded run for the same config (same object count per shard: dense collapsed to one sharded object, ragged already one object).
- The two rejections are gone; `sharded: true` + hive validates and writes.
- Worker RSS measured at o9 (expect <15%, matching flat) and at a dense/high-density case (o8 and/or a Sentinel-2 raster shard) to confirm the accumulate term stays under 4 GB — or (b) is implemented.
- Tests cover local + Lambda dispatch (shared `process_and_write_hive`), K==1 no-op parity, and the flat↔hive output-equality assertion.

## Benchmarking note (do NOT try to show the hive win on NEON)

Hive's payoff is **S3 partition spreading under thousands of shards** — a CONUS-scale throttle regime. The NEON benchmark box is **4 shards**; a hive arm there validates *parity + per-shard overhead* but shows ~zero partition benefit. So: add a flat-vs-hive **NEON** arm for correctness/regression once this lands, but the cost/wall **advantage** measurement belongs to a **CONUS-scale full-AOI** run (the per-release harness from #234). See the note left on #234.

## References

- #199 hive store layout · `docs/design/sparse_coverage.md` (D1–D5)
- #209 ragged field sharded-in-leaf (the precedent) · #211
- #233 / #215 sharded default for flat (this is the deferred hive half)
- #91 streaming per-chunk write (the O(chunk) memory bound)
- #218 Sentinel-2 raster (the dense workload that stresses the memory tradeoff)
- #234 per-release full-AOI benchmark (where the CONUS-scale hive-vs-flat comparison belongs)
- Key code: `runner.py:1477-1533` (flat both-modes), `hive.py:670-760` (`process_and_write_hive`), `healpix.py:611-625,679-753` (`_shard` / ragged branch / `emit_shard_template`), `config.py:349-360` (hive+sharded rejection)

### LINKED ISSUE #238: Demote or drop grid.shard_order (sub-shard sharding-object granularity)
🤖 *from Claude*

Carved out of the 2026-07-15 pipeline-unification discussion (context: #215 / PR #233 make `sharded: true` the aggregation default, #236 extends it to hive leaves, and the agreed end state is that an **empty config yields the production profile** — the user-facing surface collapses to roughly grid orders + fields).

## What it is

`grid.shard_order` sets the **sharding-object granularity** below the dispatch shard: constrained to `parent_order <= shard_order <= chunk_inner` (`src/zagg/grids/healpix.py:150-171`), default `None` = one ShardingCodec object per dispatch shard. Set finer, one worker writes several sharding objects — still one PUT each, still single-writer (`src/zagg/processing/write.py:295-393`).

## Why demote or drop

- No shipped config sets it; no benchmark arm exercises it; readers don't depend on it (they derive spans from `arr.shards or arr.chunks`).
- It is pure config surface with no current consumer, in exactly the place we're trying to collapse knobs.
- It is *not* dangerous (the ≤-one-dispatch-shard invariant preserves the single-writer, no-RMW guarantee) — this is about surface area, not correctness.

## Options

1. **Drop** — delete the key, the accessor (`src/zagg/config.py:1766`), and the healpix plumbing; re-add later if a consumer appears. Finer objects would only matter for cheaper sub-shard random access on dense arrays, and the sharding index already gives 2-GET single-cell reads (#209).
2. **Demote to expert-only** — keep the plumbing, remove it from docs/example configs, validate but don't advertise.

Lean: **(1) drop** — the random-access case is already served, and dead knobs rot.

espg:
Closes #236
Closes #238

Brings **hive** store-layout output to parity with **flat** by writing a leaf's dense arrays through the `ShardingCodec` — one object per array per leaf — instead of streaming them as K per-inner-chunk objects onto a single leaf prefix. Follows the ratified plan on the issue (always-accumulate, option (a): the dense accumulate term is ~1.3 MB at production geometry — parent 11 / child 19, 65,536 cells × ~20 B). Raster is out of scope (#237); rectilinear stays flat.

## Approach

- **New whole-leaf writer** `write_leaf_to_zarr` (`src/zagg/processing/write.py`): the hive counterpart of `write_shard_to_zarr` and the dense analog of `write_ragged_leaf_to_zarr`. Accumulates the shard's K `(block_index, carrier, ragged)` results and writes each dense array (and each ragged field, riding the same pass) in ONE `set_block_selection` at leaf block 0. Placement is `grid.shard_local_region`; `resolution: chunk` companions stay per inner chunk at leaf-LOCAL blocks. No issue #133 object split — a hive leaf is one object per array by construction.
- **Write switch** in `process_and_write_hive` (`src/zagg/hive.py`): mirrors the flat switch in `runner._process_and_write` — when `grid.sharded`, accumulate via `chunk_results` and write once after the stream; otherwise keep the per-chunk streaming path. The spec change and the switch are inseparable (a per-chunk loop against a sharded leaf array would read-modify-write the shard object K times). The leaf template stays lazily emitted, so a no-chunk shard still never creates the `.zarr/` prefix, and the D4 stamp remains the leaf's final write (dense+ragged → coverage sidecar → stamp).
- **Spec + carve-out deletions**: hive leaf templates emit `sharded_array_spec` for dense arrays (whole-leaf outer shard, `cells_per_shard`, never the object split); the `emit_shard_template` rejection (`grids/healpix.py`), the `validate_config` hive+sharded rejection (`config.py`), and the `from_config` hive default carve-out (`grids/__init__.py`, from #233/#215) are deleted. Hive now defaults sharded, same as flat; explicit `sharded: false` remains a legitimate opt-out; K==1 grids silently no-op sharding on both layouts. A new pointed rejection replaces the old one: `hive` + `grid.shard_order` is refused (the #133 object split is a flat-path memory bound; a leaf's arrays are one whole-leaf object each, so the knob would be silently ignored).

## Phases

- [x] Phase 1 — whole-leaf dense accumulate + single write (`write_leaf_to_zarr`) and the `process_and_write_hive` switch; unit tests for the new writer
- [x] Phase 2 — leaf spec emits sharded dense arrays; delete the three carve-outs (template rejection, `validate_config` rejection, hive `sharded` default carve-out)
- [x] Phase 3 — tests: flat↔hive parity (object counts + byte parity), both dispatchers (local `agg` end-to-end + Lambda handler) through the shared `process_and_write_hive`, K==1 no-op parity, stamp-ordering/sidecar unchanged
- [x] Phase 4 — docs: `docs/hive_layout.md` + `docs/sharding.md` lose the hive-unsharded caveats (and the stale "local backend only" line); the K==1 explicit-`sharded: true` no-op is documented on both layouts
- [x] Phase 5 — drop `grid.shard_order` entirely (issue #238, folded in per espg's ratification: https://github.com/englacial/zagg/issues/238#issuecomment-4986885796): the config key + `get_shard_order` accessor, the `HealpixGrid` param/attrs (`shard_obj_order`, `cells_per_shard_object`, `shard_objects_per_shard`) and the three `shard_object_*` methods, the issue #133 object-split branch in `write_shard_to_zarr` (collapsed to the single whole-shard-object pass — byte-identical at the former default, pinned by the existing sharded↔regular parity tests), the phase-2 hive+`shard_order` rejection, and all doc/config mentions (the hive manifest's `shard_order` FIELD — the dispatch/tree order — is a different concept and stays). The #133 object-split test class is salvaged as a knob-free mixed-field-kind gate (`TestShardedMixedFieldKinds`): sharded vs per-chunk parity for scalar/vector/companion/ragged plus one-object-per-array counts. Net −267 lines.

## Testing

- `TestWriteLeafToZarr` (`tests/test_processing.py`): one object per dense array and per ragged field at leaf block 0; byte-for-byte parity with `write_shard_to_zarr`'s shard region for the same `chunk_results`; companion (leaf-local blocks) and vector (trailing-dim) branches covered (review fold).
- `TestProcessAndWriteHiveSharded` (`tests/test_hive.py`): the accumulate switch (asserts `write_chunk=None` + `chunk_results` sink), single-object-per-array + flat parity through the real worker path, pinned write order (`leaf → sidecar → stamp`), error shard leaves no prefix, K==1 explicit-`sharded: true` no-op (file-set + byte identity modulo the stamp timestamp).
- Dispatchers: `TestRunnerWiring.test_local_hive_sharded_leaf_single_object` (real `agg(backend="local")` → real `process_and_write_hive`) and `TestProcessHive.test_hive_sharded_event_writes_single_object_leaf` (Lambda `_handle_process`) — both assert one `ShardingCodec` object per leaf array and a complete stamp.
- Template geometry: sharded leaf template shards at `(cells_per_shard,)`, including under a divergent `shard_order` (review fold); `validate_config` accepts hive+sharded, rejects hive+`shard_order`; hive default flips to sharded at K>1.
- Local: `ruff check` / `ruff format --check` clean on all touched files; full `pytest -q` (no `-m slow`): 1960 passed at phase 2, targeted suites green through the folds.

## Notes for review

- RSS at fleet scale is operator-run: the o9 RSS check rides the existing benchmark `max_memory_mb` metric (expected no visible change, +~1.3 MB accumulate term). No live-AWS runs from this PR.
- Pre-existing local failures NOT touched (unrelated to this change): `ruff check` N818 on `src/zagg/registry.py` (`UnknownCapability`), `ruff format --check` flags 5 files already unformatted on `main` (`src/zagg/__init__.py`, `src/zagg/viz/crs.py`, `tests/test_catalog.py`, `tests/test_extract.py`, `tests/test_integration.py`), and `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` fails on this machine because the build script's `pip` resolves against a Python 3.10 interpreter (cannot satisfy `zarr>=3.1.5`) — environmental, passes in CI's `lambda-build.yml`.

## Questions for review

- ~~The hive + `grid.shard_order` rejection is a NEW constraint this PR adds...~~ **Resolved by #238**: espg ratified cutting `grid.shard_order` entirely (https://github.com/englacial/zagg/issues/238#issuecomment-4986885796), folded here as phase 5 — the rejection is gone along with the knob. Rationale: one object per dispatch shard is the PUT-count optimum; the object split's only load-bearing corner (`sharded: true` at coarse orders vs the 2 GB write cap) is unoccupied, and the escape valve there is `sharded: false` (per-chunk objects) or re-adding the knob from the #238 record.
- The adversarial reviewer's phase-2 "layer gap" claim (whole-leaf spec vs `shard_slab_shape()` mismatch under a directly-constructed `shard_order` grid) was verified to be incorrect (`shard_slab_shape()` is already whole-shard); the rebuttal is on the review thread — no defense assert was added. Flag if you want one anyway.


src/zagg/processing/write.py
```diff
@@ -524,6 +524,98 @@ def write_ragged_leaf_to_zarr(ragged_chunks: list, store: Store, *, grid) -> Sto
...
+    slabs: dict = {}
+    ragged_slabs: dict = {}
+    for block_index, carrier, ragged in chunk_results:
+        region = grid.shard_local_region(block_index, shard_key)
+        # Companions for this inner chunk are not sharded: write them straight
+        # through at the chunk's leaf-LOCAL block (region start over the chunk
+        # extent — the ``hive.leaf_block_index`` derivation).
+        if chunk_res_fields:
+            local = tuple(int(s.start) // c for s, c in zip(region, inner_shape))
+            _write_companion_columns(carrier, store, grid, local, chunk_res_fields)
```
espg, line 536: 🤖 *from Claude (review)*

[non-blocking] Test gap on the one piece of logic that actually *differs* from `write_shard_to_zarr`. The flat path writes companions at the **global** `block_index`; here the leaf's companion arrays are leaf-local (`shard_spec` sizes them to `chunks_per_shard`), so this derives a leaf-local block `local = region.start // inner_shape`. That conversion (lines 568-571) is the genuinely new code in `write_leaf_to_zarr` — and it has zero coverage: `_ragged_cfg` (the `TestWriteLeafToZarr` fixture config) declares no `resolution: chunk` field, so `chunk_res_fields` is empty and this whole branch never executes. A parity test built on a chunk-resolution companion (there's already `_chunk_ragged_cfg` at tests/test_processing.py:1635) would guard the `region.start // inner_shape` mapping against an off-by-shard regression — the exact bug class that wouldn't surface until a hive+companion config runs at scale in phase 2+.

src/zagg/processing/write.py
```diff
@@ -524,6 +524,98 @@ def write_ragged_leaf_to_zarr(ragged_chunks: list, store: Store, *, grid) -> Sto
...
+                # trailing payload dim must be one whole chunk, or the write at
+                # trailing block 0 would silently cover only part of it.
+                target_trailing_chunks = array.chunks[len(slab_shape) :]
+                if target_trailing_chunks != trailing:
+                    raise ValueError(
+                        f"vector field {name!r}: trailing chunk "
+                        f"{target_trailing_chunks} must equal trailing shape "
+                        f"{trailing} (the payload dim must be one whole chunk)"
+                    )
+            array.set_block_selection((*leaf_block, *((0,) * len(trailing))), slab)
```
espg, line 579: 🤖 *from Claude (review)*

[non-blocking] The vector/trailing-dim branch (lines 602-613) plus the trailing-block placement `array.set_block_selection((*leaf_block, *((0,) * len(trailing))), slab)` are uncovered by `TestWriteLeafToZarr` — the fixture (`_ragged_cfg`) declares only a scalar (`h_min`) and a ragged (`h_raw`) field, so `trailing` is always empty and neither the single-trailing-chunk `ValueError` nor the trailing block-0 write ever runs. This is copied verbatim from `write_shard_to_zarr` (tested there via a vector field), but the leaf variant's `leaf_block = (0,)*len(slab_shape)` prefix is its own code path. Adding a `kind: vector` field to the flat↔hive parity test would exercise the trailing placement and close the gap in the same pass as the companion coverage.

src/zagg/grids/healpix.py
```diff
@@ -688,7 +691,12 @@ def _shard(arr):
             # 8): ``cells_per_shard_object`` cells, which is ``cells_per_shard`` (the
             # whole dispatch shard) at the default ``shard_order`` and SMALLER when a
             # finer ``shard_order`` is set. The inner read chunk is unchanged.
-            shard = (self.cells_per_shard_object, *inner[1:])
+            # A hive LEAF (issue #236) always shards across the whole leaf — one
+            # object per dense array, written at leaf block 0 — never the object
+            # split (``write_leaf_to_zarr`` assumes it; ``validate_config`` rejects
+            # hive + shard_order).
+            cells = self.cells_per_shard if leaf else self.cells_per_shard_object
```

[non-blocking] The new leaf-vs-object distinction here (`cells = self.cells_per_shard if leaf else self.cells_per_shard_object`) is guarded against a `shard_order` mismatch *only* at the config layer — `validate_config`'s new hive+`shard_order` rejection. Two gaps worth noting:

1. **Test blindspot.** `test_sharded_leaf_template_shards_whole_leaf` (tests/test_hive.py:270) builds the fixture with no `shard_order`, where `cells_per_shard == cells_per_shard_object` (parent 6 / child 10 → 256 == 256). So `assert grp[name].shards == (g.cells_per_shard,)` (line 280) would pass *equally* if this line wrongly read `cells_per_shard_object` — the assertion can't gate the very distinction it's testing. A direct-construction grid with `shard_order` (where the two diverge, e.g. `shard_order=8` → 16 vs 256) is what would actually catch a regression to the object-split size on a leaf.

2. **Layer gap.** `write_leaf_to_zarr` sizes its slab from `grid.shard_slab_shape()` == `(cells_per_shard_object,)`, while this leaf spec is sized `cells_per_shard`. They agree only when `shard_order` is unset. A directly-constructed `HealpixGrid(sharded=True, shard_order=…)` used as a hive leaf (bypassing `validate_config`) would thus emit a whole-leaf array but write only a partial-leaf slab at block 0 — a silent short write. Since the invariant is load-bearing ("write_leaf_to_zarr assumes it"), consider an assert `shard_order is None` in `shard_spec`/`emit_shard_template` as belt-and-suspenders rather than relying solely on the config gate.

src/zagg/grids/healpix.py
```diff
@@ -613,13 +613,12 @@ def emit_shard_template(self, store: Store, *, overwrite: bool = False) -> Store
         :meth:`emit_template` but with every dense array sized to a single
         shard (``cells_per_shard`` cells, K inner chunks) and a ROOT group so
         the D4 commit stamp is one attrs update on an object that exists
-        anyway. Writes go at leaf-LOCAL block indices (0..K-1).
+        anyway. Writes go at leaf-LOCAL block indices (0..K-1); when
+        ``sharded`` (issue #236) each dense array is instead ONE ShardingCodec
```
espg, line 538: 🤖 *from Claude (review)*

[non-blocking] Now that this commit deletes the `emit_shard_template` sharded rejection, the phase-1 test helper `TestWriteLeafToZarr._leaf_template` (tests/test_processing.py:1500-1508) is stale: its docstring still says *"the `emit_shard_template` seam still rejects a sharded grid until the spec phase of issue #236 deletes the carve-out"* and hand-builds the leaf template via `GroupSpec(members={grid.group_path: grid.shard_spec()})` to route around a rejection that no longer exists. Post-phase-2 that helper is exactly what `emit_shard_template` now does, so it can call `grid.emit_shard_template(store, overwrite=True)` directly (and drop the stale comment). Not wrong — just dead workaround left behind by this deletion; folding it in the tests phase keeps the phase-1 helper honest.

src/zagg/processing/write.py
espg: 🤖 *from Claude*

Addressed in 976babe: `TestWriteLeafToZarr.test_companion_and_vector_flat_leaf_parity` adds a `resolution: chunk` companion (`h_anchor`) to the fixture, so the `local = region.start // inner` branch now runs — the test pins the companion at leaf-LOCAL blocks 0..K-1 of the leaf-sized array and matches flat's value at the global block.

src/zagg/processing/write.py
espg: 🤖 *from Claude*

Addressed in 976babe: the same test adds a `kind: vector` field (`h_edges`) to the flat↔hive parity, so the `trailing` guard and `set_block_selection((*leaf_block, *((0,)*len(trailing))), slab)` placement now run — byte-for-byte flat parity plus a per-chunk value check on the trailing block write.

src/zagg/grids/healpix.py
espg: 🤖 *from Claude*

Folded in 3a041ed.

(1) Test blindspot — fixed. Added `test_sharded_leaf_template_shards_whole_leaf_under_shard_order`, a direct-construction `HealpixGrid(6, 10, chunk_inner=8, sharded=True, shard_order=7)` where `cells_per_shard` (256) and `cells_per_shard_object` (64) DIVERGE. It asserts `cells_per_shard != cells_per_shard_object` up front, then that every dense array's `.shards == (cells_per_shard,)` — so a regression to the object size in the leaf branch now fails.

(2) Layer gap — this claim is **incorrect**, so no defense-in-depth was added. The claim rests on `write_leaf_to_zarr` sizing its slab from `shard_slab_shape() == (cells_per_shard_object,)`, but `shard_slab_shape()` returns `(cells_per_shard,)` (healpix.py:285); it is `shard_object_slab_shape()` that returns `(cells_per_shard_object,)` (healpix.py:313), and that one is used by the flat object-split path, not the leaf path. `write_leaf_to_zarr` (write.py:560) reads `grid.shard_slab_shape()` == `(cells_per_shard,)`, matching the leaf spec's `cells_per_shard` (leaf=True). Both sides size to the whole leaf, so a direct-construction sharded grid with a finer `shard_order` used as a hive leaf emits a whole-leaf template AND writes a whole-leaf slab — no silent short write. An `assert shard_order is None` would be dead defense, so I left it out.

src/zagg/grids/healpix.py
espg: 🤖 *from Claude*

Folded in 3a041ed. Simplified `TestWriteLeafToZarr._leaf_template` to call `grid.emit_shard_template(MemoryStore(), overwrite=True)` directly and dropped the stale docstring/GroupSpec workaround — post phase 2 that seam is exactly what the helper hand-built. All three call sites (`test_one_object_per_array_at_leaf_block_zero`, the `_setup`-based test, and the newer `test_companion_and_vector_flat_leaf_parity` at 976babe) use the same helper and stay green (269 passed).

tests/test_hive.py
```diff
@@ -541,6 +559,251 @@ def wrapped(*a, **k):
...
+
+        grid = self._grid(cfg)
+        shard = _shard_word()
+        monkeypatch.setattr(processing, "process_shard", self._accumulate_fake(grid, error="boom"))
+        root = str(tmp_path / "store")
+        meta = hive.process_and_write_hive(
+            shard, ["s3://b/g1.h5"], grid, {}, root, cfg, store_kwargs={}
+        )
+        assert meta["error"] == "boom"
+        assert not os.path.exists(hive.shard_leaf_path(root, shard))
```
espg, line 758: 🤖 *from Claude (review)*

[non-blocking] Sharded path has no torn-write / debris-then-retry test, unlike the streaming suite. `test_error_shard_leaves_no_prefix` only pins the *clean-error* case (process_shard returns `error`, so the gated whole-leaf write is skipped and no prefix appears). But the sharded switch **reorders** every dense+ragged write into a single post-stream `write_leaf_to_zarr` → `_leaf()` (which emits the template, creating the prefix) → sidecar → stamp. A worker that dies *inside* `write_leaf_to_zarr` (some arrays written, others not) or *after* it but before `stamp_commit` leaves an UNSTAMPED prefix — a debris shape that does not exist on the streaming path (where debris begins at the first `_write_chunk`). The streaming analog `test_torn_write_leaves_debris_then_retry_succeeds` (line ~430) pins `read_commit is None` on the partial prefix and the `overwrite=True` wholesale wipe on retry; the sharded suite has no equivalent. The overwrite machinery is shared so risk is lower, but the deferred-write debris/retry contract is currently unpinned for `grid.sharded`. A sharded twin of the torn-write test (raise inside a patched `write_leaf_to_zarr`, assert debris + retry cleans) would close it.

tests/test_hive.py
```diff
@@ -604,6 +867,49 @@ def fake_hive_write(shard_key, granule_urls, grid, s3_creds, store_root, config,
         assert sorted(os.listdir(root)) == [hive.ROOT_COVERAGE_NAME, hive.MANIFEST_NAME]
         assert hive.read_manifest(root)["shard_order"] == 6
 
+    def test_local_hive_sharded_leaf_single_object(self, monkeypatch, cfg, tmp_path):
```
espg, line 932: 🤖 *from Claude (review)*

[non-blocking] Cross-class helper instantiation couples two test classes. `test_local_hive_sharded_leaf_single_object` does `helper = TestProcessAndWriteHiveSharded()` purely to borrow `_accumulate_fake`/`_chunk_carrier`/`_meta`. It works today only because those methods lean on nothing but `self` for static dispatch — no fixtures, no state. If `TestProcessAndWriteHiveSharded` ever grows an `__init__`, a class-scoped fixture, or a `_chunk_carrier` that reads instance state, this silently breaks (or worse, mis-behaves) in `TestRunnerWiring` with no signal that the two are entangled. The same accumulate fake is duplicated again inline in the lambda test (`tests/test_lambda_handler.py`), so there are now three copies/borrows of one contract. Hoisting the accumulate fake to a module-level helper (or a shared fixture) that both dispatcher tests call would decouple them and kill the duplication.

docs/hive_layout.md
```diff
@@ -30,8 +31,14 @@ round — wired to the **local backend only** (see [Status](#status)).
   inventories, unambiguous if moved. Each leaf is a vanilla zarr v3
   store: the same group/array template as the flat layout, sized to one shard
   (dense arrays hold `cells_per_shard` cells; `resolution: chunk` companions
-  hold the shard's K inner chunks; CSR ragged subgroups sit under the same
-  group path, named by the shard label at K==1). One recorded exception
+  hold the shard's K inner chunks). When `sharded` (the K > 1 **default**,
+  matching flat — [issue #236](https://github.com/englacial/zagg/issues/236))
+  each dense array is ONE `ShardingCodec` object spanning the whole leaf,
+  written at leaf block 0; a ragged field's vlen-bytes array is likewise one
```

[non-blocking] This ties the ragged vlen array's whole-leaf bundling to the dense `sharded` flag ("When `sharded` … a ragged field's vlen-bytes array is **likewise** one whole-leaf object"), but per the code it is unconditional on a K>1 leaf. `HealpixGrid._group_spec` sizes the ragged shard from `if leaf and self.chunks_per_shard > 1` (grids/healpix.py:738) — no dependence on `self.sharded` — and the comment right there reads "a hive LEAF shards the ragged field across the whole leaf **(sharded or not)** … issue #209". `process_and_write_hive` confirms it at runtime: on the unsharded (`sharded: false`) streaming path it still accumulates `ragged_chunks` and writes them in a single `write_ragged_leaf_to_zarr` (hive.py:805-806), one object. Combined with line 69 ("an explicit `sharded: false` opts back into the K streaming objects"), this reads as if the ragged field also reverts to K objects when sharding is off — it does not. Suggest pulling the ragged sentence out of the "When `sharded`" clause and noting the ragged bundling is always-on at K>1 (issue #209), independent of the dense toggle.

docs/hive_layout.md
```diff
@@ -274,5 +291,12 @@ under D9/O7). The §7 sweep remains the authoritative rebuilder.
   default on for hive) for the one-GET bootstrap, plus the `zagg.coverage`
   reader primitives (per-tier AOI intersection, O7 staleness lean, explicit
   refresh).
+- **Dense arrays shard inside the leaf**
+  ([issue #236](https://github.com/englacial/zagg/issues/236)): hive output is
+  byte-identical to the flat sharded layout — one `ShardingCodec` object per
+  dense array per leaf (plus one per ragged field,
+  [issue #209](https://github.com/englacial/zagg/issues/209)), the default at
+  K > 1, so a leaf costs one PUT per array instead of K per-inner-chunk PUTs
```

[non-blocking] "a leaf costs one PUT per array" overstates it — only the **dense** per-cell arrays (and the ragged field) collapse to one object; `resolution: chunk` companion arrays stay per inner chunk (K PUTs each). `write_leaf_to_zarr` writes companions straight through at leaf-LOCAL blocks inside the per-chunk loop (processing/write.py:565-568), never bundled — and the Layout section above (line 33) plus docs/sharding.md's storage-layout list both correctly say companions keep the regular one-block-per-chunk layout. Suggest "one PUT per **dense** array" so a leaf carrying a companion isn't implied to be a single PUT across all its arrays.

docs/hive_layout.md
```diff
@@ -55,8 +62,18 @@ output:
...
+default whenever `chunk_inner` gives K > 1 — same contract as flat
+([issue #236](https://github.com/englacial/zagg/issues/236)): each leaf's
+dense arrays collapse to one object apiece instead of K per-inner-chunk
+objects PUT onto a single leaf prefix. An explicit `sharded: false` opts back
+into the K streaming objects; an explicit `sharded: true` at K == 1 validates
+and is a no-op (nothing to bundle — the leaf is byte-identical either way).
+
 Validation rejects `hive` with a rectilinear grid (node names are morton
-digits), with `sharded: true` (a leaf is a vanilla zarr v3 store), and with
+digits), with `grid.shard_order` (the flat layout's ShardingCodec object
```

[question] Naming collision worth a disambiguating half-sentence: this rejects `grid.shard_order` (the issue #133 ShardingCodec object-split knob, a config input under `output.grid`), but the manifest a few lines down records `"shard_order": 9` (line 92) as a frozen key — and there it means the **dispatch / tree order** (== `parent_order` 9, `split_schedule` length 9), a different concept. The `grid.` prefix + #133 link is technically precise, but within one doc a reader hits "rejects … grid.shard_order" then "manifest shard_order: 9" and can conflate them. Verified in config.py: `get_shard_order` reads `output.grid.shard_order` (#133 split), while the manifest's `shard_order` is the dispatch order — genuinely distinct. Consider clarifying the rejected knob is the object-split order, not the manifest's dispatch shard_order (or leave as-is if the prefix is enough).

tests/test_hive.py
espg: 🤖 *from Claude*

Folded in ce52860. Added `TestProcessAndWriteHiveSharded::test_torn_write_leaves_debris_then_retry_succeeds`, the sharded twin of the streaming torn-write test. The first attempt monkeypatches `processing.write_leaf_to_zarr` to run the real leaf write (arrays land, so the `.zarr/` prefix exists) then raise `RuntimeError("worker died mid-shard")` before `stamp_commit` — pinning the deferred-write debris shape unique to `grid.sharded`. It asserts the prefix EXISTS but `hive.read_commit` is None (and `read_coverage` is None), plants both a stray `stale-debris` object AND a coverage sidecar in the debris, then a clean retry (real writer restored) overwrites the leaf WHOLESALE — both planted objects are gone — and stamps complete.

tests/test_hive.py
espg: 🤖 *from Claude*

Folded in ce52860. Hoisted the accumulate fake to a module-level `_sharded_accumulate_fake(grid, chunk_carrier, meta, ragged_by_local=None, occupied=None, error=None)` (keeps the `assert sink is not None and write_chunk is None` contract pin). `TestProcessAndWriteHiveSharded._accumulate_fake` now delegates to it, and `_meta` became a `@staticmethod` so it is callable class-level. `TestRunnerWiring.test_local_hive_sharded_leaf_single_object` no longer does `TestProcessAndWriteHiveSharded()` — it calls the helper directly with `TestProcessAndWriteHiveSharded._chunk_carrier` / `._meta` (statics), so an `__init__`/fixture on the sharded class can no longer silently entangle the two.

For the lambda test (`tests/test_lambda_handler.py`, different module) I left its inline fake with a short comment naming the shared `_sharded_accumulate_fake` contract, rather than importing it: that variant deliberately fills EVERY chunk (no every-4th-empty) so its whole-leaf `cell_ids` parity assertion holds, and it uses this module local chunk-level `_carrier_of` — importing the helper would change that behavior.

docs/hive_layout.md

Folded in 2822d5b. Pulled the ragged sentence out of the "When `sharded`" clause in the Layout section: the vlen-bytes array is now documented as one whole-leaf object whenever K > 1 — **sharded or not** (issue #209), independent of the dense `sharded` toggle — matching `healpix.py`'s `if leaf and self.chunks_per_shard > 1` (no `self.sharded` dependence). The Config section's `sharded: false` line is now scoped to the **dense** arrays opting back into K streaming objects, with an explicit note that the ragged vlen array stays one whole-leaf object regardless.

docs/hive_layout.md

Folded in 2822d5b. The Status bullet now reads "one PUT per **dense** array" — `resolution: chunk` companions stay per inner chunk (`write_leaf_to_zarr` writes them straight through at leaf-LOCAL blocks inside the per-chunk loop, processing/write.py:565-568, never bundled). The ragged mention is kept as-is.

docs/hive_layout.md

Folded in 2822d5b. Added a disambiguating parenthetical where the rejection is described: "the `grid.shard_order` config knob ... unrelated to the manifest's `shard_order` field below, which records the dispatch/tree order." Confirmed in config.py that `get_shard_order` reads `output.grid.shard_order` (#133 object split) while the manifest's `shard_order` is the dispatch/tree order (== parent_order) — genuinely distinct concepts.

espg:
🤖 *from Claude (review)*

Fresh-context adversarial review of **phase 5** (`e1c82bb`, drop `grid.shard_order` per issue #238). **No findings — clean removal.** What I verified:

**[verified] Byte-identity at the former default holds.** The collapsed `write_shard_to_zarr` is exactly the old `use_object_split=False` else-branch (`slab_shape = grid.shard_slab_shape()`, `shard_block = grid.block_index(shard_key)`, `region = grid.shard_local_region(...)`). Rectilinear had no `shard_object_*` methods, so it always took that else-branch → literally unchanged. HEALPix took the object-split branch, so I checked it reduces to the else-branch at the default (`shard_obj_order == parent_order`), and it does, algebraically:
- `cells_per_shard_object = 4^(child-parent) = cells_per_shard`, `shard_objects_per_shard = 1` → `shard_object_slab_shape() == shard_slab_shape()`.
- With `inner_block = shard_block·K + local` (`local < K`) and `cells_per_shard = K·cells_per_chunk`: `shard_object_block = cell_offset // cells_per_shard_object = shard_block = block_index(shard_key)`, and `shard_object_local_region` start `= local·cells_per_chunk = shard_local_region` start.

**[verified] Write ORDER preserved.** Companions are written unconditionally per chunk in `chunk_results` order in both versions (old: after the in-memory `objects.setdefault`; new: at loop top before the `continue`) — the append had no store effect, so the companion → dense-slab → ragged-slab store-write sequence is identical. Dense `slabs` insertion order (first-encounter column order) and the empty/ragged filter are equivalent (old `not empty or ragged` = new `not (empty and not ragged)`, De Morgan). At the default there was exactly ONE `obj_block`, so folding the per-object loop away changes nothing.

**[verified] Ragged spec truth table — no combination changed.** Since the knob is gone `cells_per_shard_object == cells_per_shard` always, so old `elif not leaf and self.sharded → (cells_per_shard_object,)` == new `(cells_per_shard,)`. Enumerated (leaf × sharded × K; K==1 forces sharded=False): (leaf,K>1)→`(cells_per_shard,)` both; (leaf,K==1)→None both; (¬leaf,sharded)→`(cells_per_shard,)` both; (¬leaf,¬sharded)→None both.

**[verified] Completeness.** `grep` for the removed symbols leaves only the manifest's `shard_order` (dispatch/tree order — `hive.py:150,210,322`, `coverage.py`, manifest test assertions, `docs/design/*` history) and the two deliberate historical mentions (`write.py:292` comment, salvaged-test docstring). Manifest field intact — nothing over-deleted.

**[verified] Test salvage.** `TestShardedMixedFieldKinds.test_sharded_reconstructs_all_field_kinds_in_one_object_each` keeps the mixed-kind (scalar/vector/companion/ragged) coverage through `write_shard_to_zarr` plus the one-object-each assertion. The dropped `test_default_byte_identical_to_explicit_parent_order` compared two now-identical grid constructions (it pinned nothing independent once the knob is gone); the standing sharded pins `TestWriteShardToZarr.test_sharded_matches_regular_byte_for_byte` + `test_one_shard_object_per_dispatch_shard` retain the value/object-count coverage.

**[verified] Docstrings.** `write_shard_to_zarr` Parameters ("Must be sharded with shard_slab_shape / shard_local_region") still accurate; no stale phase-8/object-split wording left except the one intentional historical note.

PR #242, opened 2026-07-16
Benchmark harness: record per-shard object counts (issue #240)

### LINKED ISSUE #236: Shard dense arrays in hive leaves (flat↔hive parity)

### LINKED ISSUE #240: Benchmark harness: record per-shard object counts + add NEON hive regression arm
🤖 *from Claude*

From the 2026-07-15 pipeline-unification discussion. The purpose of a NEON hive arm is **regression detection on the hive write path** — catching unexpected behavior (sharded writes silently bypassed, `shard_workers` doing something unexpected on that path) — not measuring hive's cost advantage; that measurement belongs to a CONUS-scale run per the note on #234.

## The gap

Both harnesses record cost, wall, RSS, and the phase-timing split (`.github/scripts/bench_metrics.py:44-90,136`) — but **no object-count metric**. The exact regression class the hive default needs a tripwire for (a leaf writing K per-chunk objects instead of one sharded object — the ~250× blow-up documented in #215) is invisible today except as second-order cost drift. Neither `run_benchmark.py` nor `run_full_aoi_benchmark.py` references `store_layout` at all.

## Proposed changes

1. **Object-count metric** *(independent of #236 — also guards the flat sharded path)*: after each run, LIST the output store per shard and record `objects_per_shard` (and store total) into `metrics.json` / the release-series parquet. Cheap at NEON scale (4 shards). **Assert** the expected count on sharded arms — per dense array one object, ragged one object, plus the fixed metadata objects — so a bypass fails the run instead of drifting.
2. **Hive regression arm** *(blocked by #236)*: add `store_layout: hive` target configs to both manifests — `tests/data/benchmark/targets.json` (per-merge, pinned densest shard) and `targets_full_aoi_neon.json` (per-release, all 4 shards) — as a layout axis alongside the existing 2×2 (`index_backend` × `aoi_mask`), plus a flat↔hive output-parity assertion on the full-AOI leg.
3. Existing coverage already handles the rest of the tripwire: `max_memory_mb` catches the #236 dense-accumulate term; the phase-timing split catches `shard_workers` wall-behavior anomalies. With (1)'s assertion these give the hive default the regression coverage it needs at NEON scale.

Item (1) can land first; item (2) waits on #236.

espg:
Closes #240

Closes #240 — item 1 (the object-count metric, phases 1–3) and item 2 (the hive regression arm + flat↔hive parity, phase 4, unblocked by #241 which closed #236/#238).

## What this does

Both benchmark harnesses record cost/wall/RSS/phase timings but no object-count metric, so a sharded-write bypass (a leaf writing K per-inner-chunk objects instead of one sharded object — the ~250× blow-up documented in #215) is invisible except as second-order cost drift. This PR LISTs the output store after each run, attributes objects per shard, compares against an expected-count model derived from the run's grid config, and:

- **per-merge** (`run_benchmark.py`): records `objects_total` / `objects_expected` into `metrics.json` + the retained series, and **hard-fails** the run on mismatch (the per-merge benchmark is opt-in via the `benchmark` label, so a false positive is cheap and a real bypass should block);
- **per-release full-AOI** (`run_full_aoi_benchmark.py`): **record-only** — the release run still lands its series data point; the columns join the release-series parquet and the rendered Pages panel so a regression is visible.

## Approach

New shared module `.github/scripts/bench_objects.py` (kept out of `bench_metrics.py`, which is documented as pure/no-IO):

- `expected_object_counts(grid, n_shards, store_layout, coverage_moc)` — the model is derived from the grid's **own template spec** (`grid.spec()` / `grid.shard_spec()`), never re-derived from config knobs, so it cannot drift from what the template emits. Per array: first-axis chunk extent vs the shard's span gives blocks-per-shard; a sharded (ShardingCodec) array is deterministic (1 object per shard object), an unsharded K>1 array is bounded 1..K (zarr's default `write_empty_chunks=False` omits all-fill chunks). Flat metadata = root `zarr.json` + group `zarr.json` + one per array. Hive = manifest + root `coverage.moc` at the store root; per leaf: root+group+per-array `zarr.json`, the in-leaf `coverage.moc` sidecar (depth > 0), one whole-leaf ragged object, 1..K per dense array. The live matrix config (`atl03_tdigest_healpix_o9.yaml`: flat, sharded, K=256) resolves to **exact**: 6 metadata objects + 4 objects/shard (7 + 5 on the mask arm).
- `store_object_counts(store_path, grid, shard_keys, store_layout)` — LISTs via `zagg.store.open_object_store` (identical for a local path and `s3://`, and its "/"-delimited prefix join keeps the sibling `<store>.status/` Lambda result channel out of the count). Flat attribution is block arithmetic (nested ids at a finer order tile contiguously under their parent, so `block * chunk0 // span` is the owning shard's block); hive attribution is the leaf prefix. Objects at undispatched blocks key as `block:<n>`; unclassifiable keys are counted and sampled.
- `object_count_mismatch(measured, expected)` — exact expectations assert total AND per-shard counts; bounded ones assert the total range; unrecognized keys always flag.

`index_backend` deliberately does **not** enter the model: the sidecar manifests are written to the separate index store (`data_source.index.store`), never under the output store prefix.

## Phases

- [x] Phase 1 — `bench_objects.py` (expected model + LIST/attribution + mismatch check) with fixture-backed tests: real local stores written through the production writers on both layouts (`tests/test_benchmark_objects.py`), including a sharded-bypass simulation that must read as a hard mismatch, and the committed benchmark configs pinned to their exact expected counts.
- [x] Phase 2 — per-merge wiring in `run_benchmark.py`: measure + record `objects_total`/`objects_expected` (series columns appended to `RECORD_COLUMNS`), hard-fail on mismatch after `metrics.json`/comment are written (the `--fail-on-empty` pattern).
- [x] Phase 3 — full-AOI record-only: run-record + per-shard-record columns in `run_full_aoi_benchmark.py`, `objects_total`/`objects_expected` appended to `FULL_AOI_COLUMNS`, and a per-release store-objects panel in the Pages render (`plot_series.FULL_AOI_FIGURES`).
- [x] Phase 4 — NEON hive regression arm (unblocked by #241): `tdigest_healpix_o9_hive` in the per-merge matrix and `full_aoi_neon_o9_hive` in the release manifest (new config `atl03_tdigest_healpix_o9_hive.yaml` = the o9 config + `store_layout: hive`); post-#236 the sharded hive leaf is ONE ShardingCodec object per array, so the hive model is **exact** (leaf root+group+per-array `zarr.json` + one data object per array + coverage sidecar; manifest + root MOC at the store root — hive defaults `coverage_moc: true`) and the per-merge hard-fail tripwire covers the hive write path. A new `store_layout` series column (both schemas) keys the 2×2 renderers on flat rows only — a hive row sharing `index_backend: inline` would otherwise silently overwrite the flat inline panel cell. Flat↔hive parity: `_flat_hive_parity` in the release harness, driven by `parity_with` on the hive target (the flat sibling dispatches first — manifest order), comparing every per-cell array's content per shard (NaN-aware dense, elementwise vlen bytes); verdict lands as the `parity_ok` series column, detail JSON-only. Review folds: parity covers the shards that completed WITH data (`_ok_shard_keys`, the `cells_with_data` predicate — an errored/empty shard writes no leaf and must not read as divergence; `shards_skipped` recorded), and is session-gated (`--target` subselection of just the hive arm records `parity_ok: null` + a `skipped` reason instead of comparing against a stale prior-release flat store).

## Testing

- `tests/test_benchmark_objects.py`: model pinned against real stores — flat sharded (exact, per-shard attribution), the unsharded-bypass regression sim (mismatch fires against the sharded model, passes against its own bounded model), stray-object flagging, and an end-to-end hive run through the local runner (manifest + root MOC + leaf metadata/dense/ragged/sidecar match the model object-for-object).
- `ruff check` / `ruff format --check` clean on the changed files; full `pytest -m "not slow"` green locally (2020 passed after phase 4 on the merged main); the slow offline dry-run test (`test_neon_catalog_builds_full_aoi_shardmap`) also passes. Phase-4 tests: the sharded hive leaf pinned object-for-object against the model through the real merged write path, parity clean/divergence/missing-leaf cases, manifest wiring (parity sibling order, hive config resolution), renderer flat-only gates, schema tails..

## Questions for review

1. ~~Phase 4 landing + failure semantics~~ **Resolved by espg** ([ruling](https://github.com/englacial/zagg/pull/242#issuecomment-4986921375)): phase 4 lands on this PR (done), and the full-AOI release leg must NEVER hard-fail on an object-count mismatch — record-only, permanently (a flaky-CMR release must not be blocked). Shipped accordingly: the release leg records object counts AND the flat↔hive parity verdict through never-raises helpers; the hard-fail exists only on the opt-in per-merge harness. **Parity placement:** the assertion lives in the release script as a record-only read-back (`parity_ok` column + JSON detail) because that is where both fleet-written stores exist side by side, with the comparator itself pinned hard by offline pytest against production-writer stores (`tests/test_full_aoi_benchmark.py::test_flat_hive_parity_*`) — so a parity bug fails CI, while a parity *finding* on a release is recorded, never blocking.
2. **Hive arm caveat:** the hive benchmark targets exercise the DEPLOYED worker — they need a post-#241 Lambda deploy to pass (an older worker would write the pre-#236 hive layout and trip the per-merge object-count hard-fail — which is exactly the tripwire working, but worth knowing on first run). Also, hive rows are excluded from the 2×2 cost panels for now (`store_layout` column); giving the layout axis its own panel row is a natural follow-up if you want hive cost charted.
3. Pre-existing, not touched: `ruff check src` flags `N818` on `zagg.registry.UnknownCapability` with the current ruff — present on `main` before this branch; flagging instead of fixing per the repo conventions.

_Adversarial self-review: three rounds posted inline (prefix `🤖 *from Claude (review)*`); all findings folded in c44ba02 and 29e0c87 — see the comment threads for dispositions._


.github/scripts/bench_objects.py
```diff
@@ -0,0 +1,258 @@
...
+    shard's count — a bypass regression writing K per-chunk objects (issue
+    #215) fails both. Bounded expectations (unsharded / hive data-dependent
+    counts) assert the total stays inside ``[total_min, total_max]``.
+    Unclassifiable keys are always a finding: the model claims to know every
+    object the run writes.
+    """
+    problems = []
+    total = measured["objects_total"]
+    if expected["exact"]:
+        if total != expected["total_max"]:
```
espg, line 320: 🤖 *from Claude (review)*

[non-blocking] `object_count_mismatch` computes and returns `objects_metadata` (line 223) but never asserts it against `expected["metadata"]`. The exact branch checks the grand total and the per-shard data counts; a metadata drift is only caught *indirectly*, insofar as it perturbs `objects_total`. Two consequences: (1) a regression that inflates metadata while under-writing data by the same amount passes silently (total unchanged); (2) more practically, the issue #215 CSR-subgroup blow-up writes group `zarr.json` objects, which the flat measured branch classifies as `metadata` (via the `endswith("/zarr.json")` rule, line ~193) — those are absorbed into the metadata bucket and never attributed to a shard, so the failure message reports a bare total mismatch rather than pointing at the ballooned metadata. Suggest adding an explicit check in the exact branch: `if measured["objects_metadata"] != expected["metadata"]: problems.append(f"metadata objects {measured[objects_metadata]} != expected {expected[metadata]}")`. It tightens the tripwire and makes the failure actionable.

.github/scripts/bench_objects.py
```diff
@@ -0,0 +1,258 @@
...
+
+    Rides ``zagg.store.open_object_store`` (the harness's own store factory),
+    whose prefix join is "/"-delimited, so sibling prefixes like the Lambda
+    ``<store>.status/`` result channel are never swept into the count.
+    """
+    import obstore
+
+    from zagg.store import open_object_store
+
+    store = open_object_store(store_path, **store_kwargs)
```
espg, line 179: 🤖 *from Claude (review)*

[non-blocking] `list_store_keys` is a read-only counting helper, but `open_object_store` on a local path runs `Path(path).resolve(); local.mkdir(parents=True, exist_ok=True)` (see `zagg/store.py`), so LISTing a non-existent / mistyped store path **creates** an empty directory and returns `[]` rather than surfacing "store not found". The harness then reports a confusing `total 0 != metadata N` mismatch (and leaves a stray empty dir on the runner) instead of a clear "no such store". Since phase 2 hard-fails on mismatch, a path typo would read as a spurious object-count regression. Consider guarding: check the path exists before opening (local), or list without the mkdir side effect, so an absent store fails loudly and distinctly from a real count mismatch.

.github/scripts/bench_objects.py
```diff
@@ -0,0 +1,258 @@
...
+    is a ragged (vlen-bytes) field.
+    """
+    spec = grid.shard_spec() if leaf else grid.spec()
+    members = []
+    for name, member in spec.members.items():
+        cg = member.chunk_grid
+        cfg = cg["configuration"] if isinstance(cg, dict) else cg.configuration
+        chunk0 = int(cfg["chunk_shape"][0])
+        dims = tuple(member.dimension_names or ())
+        span = grid.chunks_per_shard if dims and dims[0] == "chunks" else grid.cells_per_shard
```
espg, line 88: 🤖 *from Claude (review)*

[question] The flat model + attribution are HEALPix-fullsphere-only (per the module comment at line 44), but nothing guards it. On a rectilinear grid this path breaks in two incompatible ways: (1) `_member_layouts` accesses `grid.cells_per_shard`, which rect grids do not define (only `chunks_per_shard` exists) — so `expected_object_counts` / `store_object_counts` raise a bare `AttributeError` for any non-chunk member; (2) the `dims[0] == "chunks"` test here assumes the 1-D companion dim name, but rect companions use `("chunk_y","chunk_x")`, so a chunk-only rect config would *silently* pick `span = grid.cells_per_shard` (wrong) rather than crash — and flat attribution collapses the 2-D `block_index` to `[0]` and reads only `parts[3]` of a `c/i/j` key, mis-attributing every object. The benchmark manifests are HEALPix-only today so this is latent, but the mixed failure modes (loud AttributeError vs silent miscount) are a trap for a future caller. Worth an explicit guard at the top of both entry points (e.g. raise `NotImplementedError` unless the grid is fullsphere HEALPix) so the unsupported case fails clearly instead of miscounting. Is a rect/non-fullsphere caller in scope for the later phases, or should this be fenced off?

.github/scripts/run_benchmark.py
```diff
@@ -80,6 +87,31 @@ def _shardmap_with_mask(shardmap_path: Path, grid, aoi_file: Path) -> Path:
...
+
+    Returns the ``objects`` payload ``bench_metrics.build_record`` threads into
+    the record: measured total, the exact expectation (null when the layout's
+    count is data-dependent), the per-shard attribution, and the mismatch
+    description (null when clean) that ``main`` hard-fails on. Uses the same
+    store factory (and credentials) the dispatch just wrote through, so a LIST
+    failure is a real run failure, not a swallowed warning.
+    """
+    layout = get_store_layout(config)
+    expected = bench_objects.expected_object_counts(
```
espg: 🤖 *from Claude (review)*

[non-blocking] The object-count model is healpix-only, but `_measure_objects` runs it for **any** grid — a rect target aborts the whole benchmark.

`_measure_objects` fires unconditionally for every non-dry target with a store (`if store:` above), and `bench_objects.expected_object_counts` / `store_object_counts` both reach `grid.spec()` / `grid.shard_spec()` via `_member_layouts`. Only `HealpixGrid` exposes a public `spec()`; `RectilinearGrid` has `_spec()` only, so `grid.spec()` raises `AttributeError`. The harness ships rect configs and a rect grid backend, so the moment a rect target is (re)added to the manifest and run with `--store-prefix`, this crashes the fan-out with an opaque traceback instead of skipping the metric. Confirmed:

```
expected_object_counts(from_config(load_config('.../atl03_tdigest_rect_3km.yaml')), n_shards=1)
-> AttributeError: 'RectilinearGrid' object has no attribute 'spec'
```

No rect target is in the committed manifest today, so this is latent, not live. Suggest guarding: leave the object columns null (skip measurement) when the grid/layout isn't supported by the model, or raise a clear "object-count model supports healpix only" error — either beats crashing the run.

.github/scripts/run_benchmark.py
```diff
@@ -285,6 +325,15 @@ def main(argv: list[str] | None = None) -> int:
         "otherwise keep the job green (issue #145). Skipped under --dry-run, "
         "which emits empty metrics by design; --no-fail-on-empty opts out.",
     )
+    parser.add_argument(
+        "--fail-on-object-mismatch",
+        action=argparse.BooleanOptionalAction,
+        default=True,
```
espg, line 331: 🤖 *from Claude (review)*

[question] Default-on tripwire covers provisional targets whose expected model has no fixture assertion.

`--fail-on-object-mismatch` defaults True and applies to every non-dry target with a store. But `tests/test_benchmark_objects.py` only pins the model against the o9 flat-sharded config (`atl03_tdigest_healpix_o9.yaml` + `_aoimask`), and only validates a synthetic o6/o12 grid against a *real* store. The provisional targets run via `--target` have no such assertion — and some resolve to an **exact** model, e.g. `tdigest_healpix_o10_inline` -> `atl03_tdigest_healpix_o10.yaml` asserts `total==10` and `per_shard==4` exactly. The committed merge matrix (what actually gates a merge) is well covered and structurally identical to o9, so the false-fail risk there is low. But a model error on an unpinned provisional layout would hard-fail a legitimate experimental run by default, right when someone is iterating on it.

Two options: (a) add a real-store/model assertion for the other *exact* committed config (o10) so the strict path is covered, or (b) note in the flag help / provisional-targets doc that `--no-fail-on-object-mismatch` is the escape hatch for an unpinned provisional target. Preference?

.github/scripts/bench_objects.py
espg: 🤖 *from Claude*

Folded in c44ba02: `object_count_mismatch` now checks `objects_metadata` against `expected["metadata"]` unconditionally (it is a fixed scalar in both the exact and bounded branches), with `test_mismatch_flags_metadata_drift` pinning the compensated case (+1 metadata / −1 data, total unchanged).

.github/scripts/bench_objects.py
espg: 🤖 *from Claude*

Folded in c44ba02: `list_store_keys` now raises `FileNotFoundError` on an absent local path before touching `open_object_store` (so no stray directory is created), s3 paths unchanged. Guarded at this seam rather than in `zagg.store` — the factory's mkdir is load-bearing for its side-channel-JSON writers and out of this PR's scope. Test: `test_list_store_keys_absent_local_path_raises`.

.github/scripts/bench_objects.py
espg: 🤖 *from Claude*

Folded in c44ba02: both flat branches now call `_require_fullsphere`, raising a named `NotImplementedError` for a rect grid (no `layout` attr) or a dense-layout HEALPix grid, instead of the bare `AttributeError` / silent mis-attribution. Every manifest target (including the phase-3 full-AOI ones) is fullsphere HEALPix, so no real caller is fenced out; extending the model to rect stays out of scope until a rect target exists. Test: `test_flat_model_requires_fullsphere`.

.github/scripts/run_benchmark.py
espg: 🤖 *from Claude*

Addressed by the same fold (c44ba02): the model now fails a non-fullsphere grid with a named `NotImplementedError` ("flat object-count model assumes fullsphere HEALPix; got layout=None (RectilinearGrid)") rather than a bare `AttributeError`. It still aborts the run — deliberately: a target whose store layout the tripwire cannot model should fail loudly rather than silently skip the check; `--no-fail-on-object-mismatch` is the escape hatch if such a target is ever wanted before the model learns rect.

.github/scripts/run_benchmark.py
espg: 🤖 *from Claude*

Folded in c44ba02: `test_expected_counts_sane_for_every_manifest_config` now resolves **every** config referenced by either manifest (live matrix, provisional o10/o11, 88s, cached, full-AOI) through the model and asserts sane structure, so an unpinned provisional target cannot reach the hard-fail tripwire with a config shape the model has never seen. The o9 live-matrix configs stay exactly pinned (6+4 / 7+5); the `--no-fail-on-object-mismatch` escape hatch is documented in the flag help.

tests/test_full_aoi_series.py
```diff
@@ -204,3 +208,44 @@ def test_make_full_aoi_release_figure_renders_and_empty_is_false(tmp_path):
...
+
+def test_objects_columns_retained_and_mismatch_dropped():
+    # The two scalar object columns are retained in the release series; the
+    # mismatch description is JSON-only (dropped by the reindex), mirroring the
+    # per-merge series' JSON-only keys.
+    df = fas.records_to_frame([_record(objects_mismatch="total objects 999 != expected 27")])
+    assert list(df.columns) == fas.FULL_AOI_COLUMNS
+    assert fas.FULL_AOI_COLUMNS[-2:] == ["objects_total", "objects_expected"]
+    assert df.iloc[0]["objects_total"] == 27
+    assert df.iloc[0]["objects_expected"] == 27
```
espg: 🤖 *from Claude (review)*

[non-blocking] The `_record` fixture sets `objects_total` and `objects_expected` to the **same** value (27/27), and this test asserts each equals 27. Because the two are identical, a transposition of the columns — in `FULL_AOI_COLUMNS` ordering or in the reindex — would pass this test undetected. That matters here specifically: the release panel plots `objects_total`, so a swapped mapping would render the constant modeled `objects_expected` instead of the blown-up measured total, silently **hiding** the exact regression this metric exists to catch (a clean exact run has total == expected, so the swap is invisible except on a real regression).

Fix: give the fixture distinct values so the mapping is pinned, e.g. represent a recorded mismatch — `objects_total=999`, `objects_expected=27` in `_record`, then assert `df.iloc[0]["objects_total"] == 999` and `["objects_expected"] == 27` separately (the `objects_mismatch` override already models the mismatch case here).

espg: 🤖 *from Claude*

Folded in 29e0c87: the fixture now carries distinct values (`objects_total=27`, `objects_expected=25`) and the assertions pin each column separately, so a transposition anywhere in the flatten/reindex/render path fails the test.

espg:
> once https://github.com/englacial/zagg/issues/236 merges, should the hive targets + parity assertion land (a) as a follow-up phase on this PR, or (b) as a separate PR?

follow-up phase on this PR

> Failure semantics: per-merge hard-fails on mismatch; the full-AOI release run records only (so a release still lands its series point, and the mismatch is visible in the panel/series). Happy to make the full-AOI leg hard-fail too if you'd rather both block.

NO. We don't want a release blocked is nasa CMR is being flakey-- do not hard fail!

espg:
🤖 *from Claude*

Both answers folded into the PR state:

- **Full-AOI leg: no change needed — it already never hard-fails.** The release run records object counts through a never-raises helper (a mismatch or a flaky-CMR LIST shows up in the series/panel but can't block the release). The hard-fail exists only on the **per-merge** harness, which is opt-in via the `benchmark` label. Your "do not hard fail" is the shipped behavior; the question was only whether to *also* block releases, and that's now firmly closed as no.
- **Phase 4 (hive targets + flat↔hive parity assertion) lands on this PR** as a follow-up phase once PR #241 merges. Until then this PR is marked **`blocked`** (Blocked by #241) and returned to draft so the incomplete checklist can't merge by accident; the phase-4 commit + its review pass will flip it back to ready.

.github/scripts/run_full_aoi_benchmark.py
```diff
@@ -425,7 +540,47 @@ def run_target(
...
+    # ``parity`` detail is JSON-only, ``parity_ok`` joins the retained series.
+    parity = None
+    sibling = target.get("parity_with")
+    if sibling and store:
+        flat_store = f"{store.rsplit('/', 1)[0]}/{sibling}.zarr"
+        parity = _flat_hive_parity(
+            flat_store,
+            store,
+            grid,
+            [int(k) for k in sm.shard_keys],
```
espg: 🤖 *from Claude (review)*

[non-blocking] Parity iterates **all** `sm.shard_keys`, but object-count measurement two lines up uses `cells_with_data` (the populated/succeeded count). On a partial-failure release (a flaky Lambda shard erroring — exactly the "flaky CMR" scenario the record-only ruling exists for), the failed shard's hive leaf is absent, so `_flat_hive_parity` opens a missing leaf and records it as a mismatch → `parity_ok` flips to **False** even though nothing diverged. That conflates an infra/dispatch failure with a genuine content divergence in the one advisory signal this leg produces.

Concrete fix: drive parity off the same populated basis the object measurement uses — pass only the shards that completed with data (derive the succeeded-shard set from `summary["results"]`, or intersect with `cells_with_data`) rather than every dispatched key. A leaf missing for a shard that genuinely produced data still lands as a finding (the intended `missing_leaf_is_a_finding` case); a leaf missing because its shard errored no longer masquerades as a parity divergence.

.github/scripts/run_full_aoi_benchmark.py
```diff
@@ -425,7 +540,47 @@ def run_target(
...
+            region=region,
+        )
+    # flat<->hive output parity (issue #240 item 2): manifest-driven -- a target
+    # carrying ``parity_with`` names its flat sibling (same config modulo
+    # store_layout, already dispatched: manifest order). RECORD-ONLY; the
+    # ``parity`` detail is JSON-only, ``parity_ok`` joins the retained series.
+    parity = None
+    sibling = target.get("parity_with")
+    if sibling and store:
+        flat_store = f"{store.rsplit('/', 1)[0]}/{sibling}.zarr"
```
espg: 🤖 *from Claude (review)*

[non-blocking] The flat sibling store is located purely by name (`f"{store.rsplit('/', 1)[0]}/{sibling}.zarr"`) with no check that the sibling actually ran **this** session. `main()` supports `--target` subselection (`names = args.target or list(...)`), so re-running just the hive arm (`--target full_aoi_neon_o9_hive`, e.g. to reproduce a parity finding) leaves the sibling undispatched — yet S3 output stores persist across releases, so parity then silently compares the fresh hive store against a **stale prior-release** flat store. The verdict recorded as `parity_ok` is misleading (a spurious pass or fail) rather than `None`/unknown, and the harness has no way to tell. The docstring says it "requires the flat sibling to have dispatched FIRST" but nothing enforces or detects it.

Concrete fix: only run parity when `sibling` is in the current session's `names` set (thread the resolved `names` into `run_target`, or skip with `parity_ok=None` + a "sibling not run this session" note otherwise). On the full-manifest release path this is a no-op; it just closes the subselection hole.

.github/scripts/run_full_aoi_benchmark.py
espg: 🤖 *from Claude*

Folded in 6ec54be: parity now runs over `_ok_shard_keys(results)` — the exact `cells_with_data` predicate (status 200, no error), which also excludes granule-less shards that write no leaf (they would have false-flagged too). The dispatched-but-uncovered count is recorded as `shards_skipped` in the parity detail. A leaf missing for a shard that DID complete with data remains a genuine finding. Tests: `test_ok_shard_keys_is_the_cells_with_data_predicate`, `test_parity_recorded_runs_over_ok_shards_and_counts_skips`.

.github/scripts/run_full_aoi_benchmark.py
espg: 🤖 *from Claude*

Folded in 6ec54be: parity is now session-gated — `run_target` receives `session_targets=set(names)` from `main()`, and `_parity_recorded` skips with `parity_ok: None` plus a `skipped` reason (JSON detail) when the `parity_with` sibling was not dispatched this session, so a `--target full_aoi_neon_o9_hive`-only rerun records unknown instead of a verdict against a stale prior-release store. The full-manifest release path is unaffected (the sibling always dispatches first, manifest order — now enforced, not assumed). Test: `test_parity_recorded_gates_on_session_targets`.

Issue #62, opened 2026-06-19
null


espg:
The most impactful single investment. Both temporal (https://github.com/englacial/zagg/issues/12) and future cluster backends (https://github.com/englacial/zagg/issues/20) need the same generic fan-out → retry → measured-cost loop that currently lives as two bespoke functions in runner.py. A dispatch.py with a clean Executor protocol extracts this once. Every pipeline type (spatial, temporal, event) then inherits Lambda and local execution for free, and items 3–4 of https://github.com/englacial/zagg/issues/20 (vaex, ray/dask/slurm) plug in behind the same interface without touching pipeline logic.

This is the prerequisite that unblocks both https://github.com/englacial/zagg/issues/12 Phase 1c and https://github.com/englacial/zagg/issues/20 items 3–4.

PR #208, opened 2026-07-10
Coverage MOCs: worker-owned leaf tier + end-of-run root (issue 200)

espg:
Closes #200
Refs #198

All five phases of the coverage-MOC plan ([implementation plan](https://github.com/englacial/zagg/issues/200#issuecomment-4938764992), [answers](https://github.com/englacial/zagg/issues/200#issuecomment-4938859764), [O8 resolution](https://github.com/englacial/zagg/issues/200#issuecomment-4939477871)): the tier-0 **morton box** rides the hive commit stamp, exact cell-order occupancy ships as a **zstd-compressed bitmap sidecar** inside the leaf, the end-of-run **store-root `coverage.moc`** (shard-order ranges MOC, default-on for hive) gives readers the one-GET bootstrap, `zagg.coverage` provides the reader primitives + O7 staleness lean, and the docs (layout guide + design record) are updated. Design contract: `docs/design/sparse_coverage.md` §4 (amended via PR #206, now on `main`; O7/O8/O9 registry entries updated to resolved-with-outcome in this PR — living doc). Coverage is hive-layout-only — flat stores and rect grids are untouched.

### Phase 1 — the tier-0 box on the commit stamp

- **`zagg.grids.morton.morton_box(words)`** — the canonical ≤ 4-member MOC covering a set of occupied cells (mixed order allowed). mortie 0.9.0 public kernels only, no new dependencies: `compress_moc` canonicalizes (an occupied ancestor absorbs descendants), `common_ancestor` + `clip2order` split the DCA into its 2–4 intersecting children, each **tightened** to the common ancestor of the occupancy inside it (round-1 finding 1). Documented as not always globally minimal; the construction freezes with the mortie-side spec.
- **`process_shard(..., occupied_out=...)`** — optional out-param sink (the `ragged_out` pattern) delivering the distinct cell words holding ≥ 1 observation — exactly the `cells_with_data` cells (pooled: `cell_to_slice` keys; buffered: `StreamingAggregator.counts` keys). Needed because the streamed carrier holds all template cells. Default `None`: byte-identical.
- **`hive.build_coverage` / `stamp_commit(..., coverage=)` / `read_coverage`** — envelope `{spec: "morton-moc/1", box: [4 slots, nulls trail], cell_order, source: "worker"}` on the existing final `attrs.update` PUT (zero extra ops, debris semantics inherited); accessor strict on the spec; `generated_at` deliberately omitted at the leaf (the stamp's `written_at` is the one clock).

[...]

PR #243, opened 2026-07-16
record windowed-leaf temporal design in sparse_coverage spec (issue #237)

espg:
Refs #237 (design ratification thread; the issue stays open for the implementation PRs).

Records the temporal-partitioning design ratified on #237 ([espg ratification](https://github.com/englacial/zagg/issues/237#issuecomment-4986393261) and the follow-up) into `docs/design/sparse_coverage.md`, as three new registry entries plus the inline amendments that carry them:

- **D13 — time-windowed, write-once leaves** (option (c) from the thread): one leaf per (shard, window) under a manifest-declared schedule; D4/D5 semantics preserved verbatim (binary debris, zero shared mutable state); backfill = a new earlier-window leaf; the rejected (a)/(b) alternatives and why are recorded in the entry. Layout tree in §2 gains the windowed-leaf line, with naming deferred to the mortie spec page (mortie #62) as part of `morton-hive/2`.
- **D14 — coverage `encoding: "full"` fast path**: the stamp envelope discriminator becomes `"ranges" | "bitmap" | "full"`; popcount at stamp time; edge-of-scene shards write the real bitmap (spatial union across the window), interior dense shards skip the sidecar. §4 tier-1 carries the amendment note.
- **D15 — temporal declaration splits like coverage**: window schedule + time encoding in the static manifest (`morton-hive/2` block added to §3), actual ranges on leaf stamps (truth), time-range union on the root summary (cache). The §3 entry also records the append story explicitly (no new manifests on append; explicit-range-list schedules are the noted exception).

Single phase (docs-only). Tested: `codespell` clean on the file; no code paths touched.

## Questions for review

- ~~D13's naming placeholder (`{full_id}@{window}.zarr`) is marked as frozen-with-mortie rather than decided here — flagging that the mortie #62 spec page needs the corresponding `morton-hive/2` entry before implementation starts.~~ **Resolved** in the espg-ratified convention round after this PR opened: naming is frozen as `{full_id}_{window}.zarr` (underscore separator), recorded on the [mortie #62 spec entry](https://github.com/espg/mortie/issues/62#issuecomment-4986809092) and synced into this PR's doc by commits `30bb5d2` (vocabulary `none`/`yearly`/`monthly`/`daily`/explicit + `quarterly` reserved, `mission`→`none` rename) — decision record on the [#237 thread](https://github.com/englacial/zagg/issues/237#issuecomment-4986819344). Nothing outstanding.

### DIFF
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -124,18 +124,23 @@ touched again during a run. Contents:
   recorded explicitly for forward compatibility).
 - Pyramid declaration: which ancestor orders carry overview zarrs, and their
   aggregation methods (populated/updated by the §7 sweep).
-- **Temporal block** (D15, `morton-hive/2`, ratified on
-  [#237](https://github.com/englacial/zagg/issues/237)): time
+- **Temporal block** (D15, ratified on
+  [#237](https://github.com/englacial/zagg/issues/237)): a store carrying this
+  block declares `spec: "morton-hive/2"` (the version string of D6 covers these
+  temporal/windowed-leaf extensions). It records time
   encoding/units/epoch/calendar, the **window schedule**
   (`mission` | `yearly` | explicit range list), and the append policy. The
-  schedule is *generative*, so the manifest stays static as data accrues:
-  appending a new year to a `yearly` store adds leaves the schedule already
-  describes — no manifest touch, and **no new manifests**: the store has
-  exactly one `morton_hive.json`, and each new windowed leaf brings only its
-  own zarr metadata + D4 stamp. The one exception is the explicit-range-list
-  form: appending a window outside the declared list requires rewriting the
-  manifest (a rare, single-writer, template-time operation) — append-heavy
-  stores should prefer generative schedules. Temporal *extent* is
+  schedule is *generative*, so under `mission`/`yearly` the manifest is
+  write-once and stays static as data accrues: appending a new year to a
+  `yearly` store adds leaves the schedule already describes — no manifest
+  touch, and **no new manifests**: the store has exactly one
+  `morton_hive.json`, and each new windowed leaf brings only its own zarr
+  metadata + D4 stamp. The explicit-range-list form is the noted exception:
+  appending a window outside the declared list re-templates the manifest (a
+  rare, single-writer, template-time operation, not a worker-race write) —
+  append-heavy stores should prefer generative schedules. (This exception is
+  an implication recorded here from the ratified schedule set, not a point
+  separately ratified on the #237 thread.) Temporal *extent* is
   deliberately **not** manifest data: actual ranges live on the leaf stamps
   (truth) and in the root summary (cache), splitting static schema from
   accruing state exactly the way coverage splits under D9. The default
@@ -212,7 +217,9 @@ the data itself). Tracks [#200](https://github.com/englacial/zagg/issues/200).
   posts its completion list to a **fire-and-forget worker invoke**
   (`InvocationType="Event"`, ~10 ms of dispatcher wall clock, run-size
   independent) that writes a shard-order root MOC for the one-GET
-  bootstrap. Failure is harmless: readers degrade to the sweep MOC or the
+  bootstrap. Under D15 this root summary also carries the time-range union
+  alongside the MOC (cache, sweep-regenerable). Failure is harmless: readers
+  degrade to the sweep MOC or the
   walk, never to wrong answers. Incremental runs: the leaves carry durable
   truth, so the sweep is always a correct rebuilder, and the end-of-run
   write may union with a prior root object.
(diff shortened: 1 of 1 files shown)


espg:
🤖 *from Claude (review)*

Fresh-context adversarial review of the D13/D14/D15 docs amendment against the ratification thread on #237. Verified fidelity, internal consistency, spec-boundary discipline, and cross-references. The amendment is faithful to what was ratified — (c) windowed leaves, the `"full"` coverage fast path, and the temporal manifest/stamp/root split all match the thread, including espg's explicit "(c) it is", "fantastic, approved", and "ok, makes sense". A few consistency/precision items below; one is worth folding, the rest are minor.

**1. O8's resolved `encoding` enum isn't cross-linked to D14's extension (main finding).** O8 is marked **RESOLVED** and states the discriminator as a closed two-value set:

> the envelope gains an `encoding: "ranges" | "bitmap"` discriminator (O8, §8.2)

D14 extends it to three values (`"ranges" | "bitmap" | "full"`), and the §4 Tier-1 body carries the D14 amendment note inline. But the O8 **registry** entry — the canonical "resolved" record of this enum — has no pointer to D14, so a reader landing on O8 takes the two-value set as complete/frozen. Recommend a one-line note in O8, e.g. "(extended by D14 to add `"full"`)", mirroring how the doc already cross-notes resolutions elsewhere. This is the one spot where the "frozen" language reads as contradicted.

**2. `spec` version vs `morton-hive/2` is never tied together (precision).** D6 and the §3 `spec` bullet still exemplify `morton-hive/1`:

> `spec`: convention version string (e.g. `"morton-hive/1"`) (§3)
> **D6 — The convention is versioned** (`morton-hive/1`) in the manifest.

D13/D14/D15 introduce `morton-hive/2` features (temporal block, windowed-leaf naming), but nothing states that a store carrying the temporal block declares `spec: "morton-hive/2"`. A reader can't tell whether the version field bumps when temporal features are present. One clause tying the temporal block to a `spec: "morton-hive/2"` value would close this.

**3. Root-summary time-range union (D15) isn't reflected in §4's root-MOC lifecycle (minor).** D15 says the end-of-run root coverage object "gains the time-range union alongside the MOC." §4's root-MOC description (O9 lifecycle, lines ~210-234) still describes only the MOC and doesn't mention the time-range union it now carries. The registry entry covers it, but §4 (where the root object's lifecycle actually lives) doesn't. Consider a cross-reference so the two sections agree.

**4. Explicit-range-list manifest-rewrite exception is Claude-derived, not verbatim-ratified (minor/transparency).** D15 and the §3 temporal block assert:

> The one exception is the explicit-range-list form: appending a window outside the declared list requires rewriting the manifest

This came from the follow-up Claude comment; espg's reply pivoted to "what's actionable" without explicitly ratifying that specific point. It follows logically from the ratified schedule set (`mission | yearly | explicit range list`), so it's defensible spec content — flagging only for transparency that it's a derived consequence, not a quoted decision.

**5. "write-once at template time" vs the rewrite exception (very minor wording).** §3's header says the manifest is "never touched again during a run," and D15 calls it "write-once at template time" — then two lines later notes the explicit-range-list case "rewrites the manifest." These are reconcilable (the rewrite is an out-of-run, template-time-style op, not a worker race, and the doc says so), but "write-once" sitting adjacent to "rewrites the manifest" invites a double-take. A word like "re-templated" for the exception would remove the friction.

**Verified clean:** node invariant holds (`{full_id}@{window}.zarr` is still a `*.zarr` object, and D13 says so explicitly); spec-boundary discipline is correct (leaf-naming grammar is deferred to mortie #62 / `morton-hive/2`, shown only as an illustrative form with an explicit "frozen with the mortie spec" annotation — not frozen zagg-side); D4/D5/D9/D10 carry-over claims match the thread; rejected (a)/(b) rationale matches; `mission`-default byte-identical claim matches; all issue/anchor references (#237, #200, #208, #218, mortie #48/#62) resolve. No garbled sentences or broken markdown.

espg:
🤖 *from Claude*

Folded the adversarial-review findings into `docs/design/sparse_coverage.md` (commit `e01e2f4`):

1. **O8 ↔ D14 cross-link** — O8 (§8.2), where the envelope gains `encoding: "ranges" | "bitmap"`, now notes it is "later extended by D14 to add a third value, `"full"`", so the RESOLVED enum no longer reads as contradicted.
2. **`morton-hive/2` tied to `spec`** — the §3 temporal-block bullet now states a store carrying the block declares `spec: "morton-hive/2"`, and that D6's version string covers these temporal/windowed-leaf extensions.
3. **§4 root-MOC lifecycle** — the O9 end-of-run root-MOC bullet now notes that under D15 the root summary also carries the time-range union alongside the MOC (cache, sweep-regenerable).
4. **Explicit-range-list transparency** — marked the manifest-rewrite exception (in both §3 and D15) as an implication recorded from the ratified schedule set, not separately ratified on the thread. Handled by making the derivation explicit; stands for espg to strike or ratify.
5. **Wording double-take** — softened the §3 "write-once / stays static" juxtaposition so the explicit-range-list case reads as the noted exception ("re-templates the manifest … not a worker-race write") rather than an apparent contradiction.

`codespell` clean. PR left in draft.

Issue #235, opened 2026-07-15
multiple lambda instance sizes available at runtime?


espg:
Our lambda pipeline is heavily optimized for ATL03 at the moment; this includes pinning our lambda worker size (4096 MB, 2+ vcpus) to the optimum ATL03 shard size (o9). Looking at #213 , we surfaced that we're doing a so-so job of memory and cpu utilization for that task-- we could halve the worker size to 2048 MB at 1 vcpu, and would halve the cost of the run (or...alternatively, spend development effort to better thread and saturate the temporal worker). 

This likely will be the case for other workloads as well-- the in flight #219 + #218 + #232 S2 work might also find that having a smaller worker makes more sense.

We should update our deployment stand_up scripts to allow for (and by default create) multiple lambda configurations that we can call without having to switch to admin role to swap them out 1-for-1. Is this something we can do, describe a set of workers that are then selectable either from the agg template (my lean) and/or the invoke orchestrator call?

espg:
🤖 *from Claude*

Yes — this is straightforwardly doable, and your lean (a described set of workers, selected from the agg template) is the shape Lambda's own constraints force anyway: **memory is a function-configuration property, not an invoke property**, so per-run sizing means either mutating config (what you're rightly trying to stop doing via admin) or pre-provisioning variants. The realistic options:

**A. Pre-provisioned size variants of one function (recommended).** `template.yaml` loops a size list (e.g. `[2048, 4096, 8192]`) into N functions sharing the same code, layer, and execution role — `process-shard-2048`, `process-shard-4096`, … — created by default at standup. Selection at run time is just a function name:
- config-driven (your lean): an `output.worker: {memory: 4096}` block (or per-template default — `sentinel2_l2a.yaml` would say 2048 once #232 lands its bound) that the runner resolves to the suffixed name;
- `agg(function_name=...)` stays as the explicit override for one-off runs.
No admin at run time, no swap serialization between concurrent runs of different workloads, idle variants cost nothing (Lambda bills invocations only). Costs: N× CloudWatch/log groups and cold pools per size (warm pools don't transfer across sizes), and the account-level concurrency limit is shared across all of them (same as today).

**B. Runtime `UpdateFunctionConfiguration` per run.** What you're doing manually now, scripted. Rejected on your own grounds: needs elevated perms in the run path, serializes concurrent workloads onto one config, and a mid-fan-out change invalidates every warm sandbox (the #171 machinery treats that as a full cold restart).

[...]

PR #245, opened 2026-07-16
raster worker: store + pull-NN geometry caches (issue #244)

### LINKED ISSUE #244: Raster worker: reuse S3 stores and pull-NN geometry across asset-samples (5x per-asset overhead)
🤖 *from Claude*

## Summary

The raster worker pays **per-asset setup costs that should be per-invoke**: a fresh `S3Store` (new HTTP client, cold connection pool) and a full geometry recomputation for every `(timestep × band)` sample. On the SERC full-year o9 empirical runs (issue #218 / PR #232 measurement threads) this is most of the ~525 s `sample` phase — the same per-unit-setup disease the t-digest read path was cured of two weeks ago.

## Measured breakdown (per asset-sample, o9 shard = 1.05M cells, real c1 B04)

| stage | fresh (as shipped) | with store reuse |
|---|---|---|
| `S3Store` creation (`_store_and_path`) | **~300 ms** | 0 ms |
| `TIFF.open` (header) | 170–280 ms | 50–70 ms |
| geometry (`mort2geo` + pyproj + affine mask) | ~175 ms | ~175 ms (uncached) |
| tile fetch (5 tiles) | cold-connection latency per GET | warm pool |
| decode (async-tiff, Rust) | **15 ms** | 15 ms |
| gather (numpy) | 9 ms | 9 ms |

Fleet context (0.30.0, 4096 MB, `profile: true`, [full curve](https://github.com/englacial/zagg/pull/232#issuecomment-4986626775)): 85 timesteps × 5 bands = **425 asset-samples per invoke** → ~130 s of client construction alone, ~74 s of redundant geometry (this AOI has exactly TWO distinct source grids: the 10 m bands and 20 m `scl`), and ~2,100 tile GETs all riding cold connections. Mean invoke wall 578.5 s at `write_buffer: 1`, of which `sample` = 524 s. The decode engine is exonerated — async-tiff contributes ~6 s of the 524.

## The two fixes

1. **Connection/store reuse**: cache the obspec store per `(scheme, bucket-or-host, region, anonymous)` — one client per source bucket per worker process, shared across all samples in an invoke and across warm invocations of the sandbox.
2. **Geometry reuse**: cache `(rows, cols, valid)` per `(epsg, transform, shape)` within a shard invoke — the pull-NN mapping is invariant across timesteps and bands that share a source grid.

## Expected impact

`sample` ~524 s → ~100–200 s (geometry CPU 74 s → <1 s; fetch/open on warm pools); full-year o9 shard wall ~580 s → **~150–250 s**, with side benefits: much wider margin on the 900 s ceiling (one wb=3 invoke timed out on variance — see the curve comment), lower GB-seconds, and a stronger case for the 2048 MB worker variant (issue #235). Calibration numbers for issue #228's pre-flight model shift accordingly.

Refs #218 (pipeline), PR #232 (measurement threads), #228 (model inputs), #235 (worker sizing coupling).

espg:
Closes #244. Refs #218, PR #232 (measurement threads), #228, #235.

## What / approach

The raster worker paid per-asset costs that should be per-invoke — measured on the SERC full-year o9 runs at ~300 ms of S3 client construction + ~175 ms of redundant pull-NN geometry per asset-sample, × 425 samples per invoke, with every tile GET riding cold connections ([breakdown on the issue](https://github.com/englacial/zagg/issues/244)). Two pure-memoization fixes in `processing/raster.py`, no config surface, no new deps; handler and local runner benefit transparently.

**Phase 1 — store cache.** `_store_and_path` resolves a cache key `(kind, bucket-or-host-or-dir, region, anonymous)` and consults a lock-guarded module-level cache; construction is single-flight under the lock. Module lifetime = sandbox lifetime (espg-ratified Q1 on #244): warm invocations keep their connection pools, the issue-#171 pattern.

**Phase 2 — geometry cache.** `process_raster_shard` allocates one `geom_cache` per invoke (per-invoke scoping makes cross-shard collisions structurally impossible) and threads it through `sample_item_async` → `_sample_one`, memoizing `(rows, cols, valid)` per `(epsg, transform, shape)`. Exactly-once population is guaranteed by loop atomicity — no `await` between check and store — per the [Q2 resolution](https://github.com/englacial/zagg/issues/244#issuecomment-4986824702); the invariant is documented at the site with the upgrade path (per-key async locks) if the compute ever moves off-loop. Public `sample_asset`/`sample_asset_async` behavior unchanged (default `None` = compute per call).

## Phases

- [x] **Phase 1** — store cache (`_STORE_CACHE`/`_build_store`), key semantics, single-flight construction.
- [x] **Phase 2** — per-invoke `geom_cache` threaded through the sample path; atomicity invariant documented.
- [ ] **Phase 3** — fleet A/B after the next bundled release (espg-ratified Q3): rerun the [PR #232 protocol](https://github.com/englacial/zagg/pull/232#issuecomment-4986626775), post before/after here, refresh #228 calibration. Target: full-year o9 wall ≤ ~300 s (estimate ~150–250 s vs today's 578 s).

## How tested

`tests/test_raster.py::TestStoreAndPath` — per-test cache isolation (autouse fixture); same-href/same-vhost store identity; distinct `(bucket, anonymous)` keys distinct; constructor-count-of-one via a counting `S3Store` shim (no wall-clock assertions); LocalStore per-directory reuse. `tests/test_raster_pipeline.py::TestGeometryCache` — `grid.sample` call-count == 1 for 4 timesteps × 1 band on one source grid, == 2 for 3 timesteps × 2 bands over two grids (the SERC shape: 10 m + 20 m); cached output byte-equal to uncached. Full raster suites: 91 passed; existing golden/e2e tests unchanged (memoization has no semantic surface).

## Questions for review

- Pre-existing, untouched: `ruff` N818 on `registry.py` (long-flagged).

### DIFF
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -231,6 +231,14 @@ async def _sample_one(
         # compute-and-store is atomic on the loop and each source grid is
         # computed exactly once per invoke — no locks needed. If this compute
         # ever moves off-loop (``to_thread``), add per-key async locks.
+        # INVARIANT (issue #244 thread): the key ``(epsg, transform, shape)``
+        # is complete only because ``cells`` and ``grid`` are constants of the
+        # invoke — ``cells = grid.children(shard_key)`` is computed once (see
+        # ``process_raster_shard``) and threaded unchanged into every
+        # ``_sample_one``, and ``geom_cache`` is allocated fresh per
+        # ``process_raster_shard`` call. A future refactor that varies
+        # ``cells`` (or ``grid``) per item/group within one invoke MUST fold
+        # them into the key or drop the cache, else it returns a stale mapping.
         geom = grid.sample(cells, f"EPSG:{epsg}", transform, shape)
         if geom_cache is not None:
             geom_cache[geom_key] = geom
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -77,6 +77,29 @@ def _geo_from_ifd(ifd) -> tuple[int, tuple[float, float, float, float, float, fl
 # pattern. Lock-guarded because the running-loop fallback in ``sample_asset``
 # and hand-rolled callers can construct from other threads; construction runs
 # under the lock deliberately (single-flight per key).
+#
+# Credential lifetime (issue #244 review): every key deployed today is
+# anonymous (``sentinel2_l2a.yaml`` sets ``anonymous: true``; runner.py:676
+# and ``sample_asset*`` default it True), so the cached ``S3Store`` carries
+# ``skip_signature=True`` and signs nothing — credentials never enter the
+# picture and warm-caching a store cannot go stale. If a *signed* store
+# (``anonymous=False``) were ever cached, sandbox-lifetime caching would only
+# be safe because of how async-tiff's Rust ``object_store``-backed ``S3Store``
+# resolves credentials, and that splits by environment:
+#   - On AWS Lambda the execution-role credentials arrive as static env vars
+#     valid for the whole sandbox lifetime and do NOT rotate mid-sandbox, so a
+#     sandbox-lifetime store cannot outlive its creds — safe by construction.
+#   - Off-Lambda (EC2 instance profile / IMDS, SSO), object_store resolves
+#     through a caching credential *provider* that refreshes on expiry per
+#     request rather than freezing a token at construction, so a warm store
+#     keeps working across rotation.
+# (The Lambda case is directly grounded; the off-Lambda refresh behavior
+# reflects object_store's documented provider design — ``S3Store`` exposes a
+# ``credential_provider`` slot — and was not source-verified against the
+# installed binary wheel.) Caveat: no explicit-credential store exists here
+# (the raster path is anonymous); do NOT add one to the cache without
+# revisiting this, since a statically-supplied token would be frozen at
+# construction and eventually go stale on a warm worker.
 _STORE_CACHE: dict = {}
 _STORE_LOCK = threading.Lock()
 
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -170,9 +170,18 @@ async def _sample_one(
     region: str | None = None,
     anonymous: bool = True,
     fill=0,
+    geom_cache: dict | None = None,
 ):
     """:func:`sample_asset_async` body, also returning the raster's center
-    ``(lon, lat)`` — the ownership rule's tile-center input (#218)."""
+    ``(lon, lat)`` — the ownership rule's tile-center input (#218).
+
+    ``geom_cache`` (issue #244) memoizes the pull-NN mapping ``(rows, cols,
+    valid)`` per ``(epsg, transform, shape)``: the mapping is invariant across
+    every timestep and band that shares a source grid, so a shard invoke
+    computes it once per distinct grid (~175 ms at o9) instead of once per
+    asset-sample. ``None`` (the default, and the public ``sample_asset*``
+    path) computes per call, unchanged.
+    """
     from async_tiff import TIFF
 
     store, path = _store_and_path(href, region=region, anonymous=anonymous)
@@ -191,7 +200,18 @@ async def _sample_one(
     epsg, transform = _geo_from_ifd(ifd)
     shape = (ifd.image_height, ifd.image_width)
     center = _raster_center_lonlat(epsg, transform, shape)
-    rows, cols, valid = grid.sample(cells, f"EPSG:{epsg}", transform, shape)
+    geom_key = (epsg, transform, shape)
+    geom = geom_cache.get(geom_key) if geom_cache is not None else None
+    if geom is None:
+        # INVARIANT (issue #244 thread): no ``await`` between this check and
+        # the store below. asyncio interleaves only at await points, so the
+        # compute-and-store is atomic on the loop and each source grid is
+        # computed exactly once per invoke — no locks needed. If this compute
+        # ever moves off-loop (``to_thread``), add per-key async locks.
+        geom = grid.sample(cells, f"EPSG:{epsg}", transform, shape)
+        if geom_cache is not None:
+            geom_cache[geom_key] = geom
+    rows, cols, valid = geom
 
     th, tw = ifd.tile_height, ifd.tile_width
     vr, vc = rows[valid], cols[valid]
@@ -271,6 +291,7 @@ async def sample_item_async(
     nodata=None,
     region: str | None = None,
     anonymous: bool = True,
+    geom_cache: dict | None = None,
 ):
     """Sample every configured band of one STAC item, concurrently.
 
@@ -313,6 +334,7 @@ async def sample_item_async(
                 region=region,
                 anonymous=anonymous,
                 fill=bands[f]["fill_value"],
+                geom_cache=geom_cache,
             )
             for f in fields
         ]
@@ -508,6 +530,12 @@ def process_raster_shard(
     # orders.
     k = _shard_workers(config)
     wb = _write_buffer(config)
+    # Pull-NN geometry memo (issue #244), scoped to THIS invoke: the (rows,
+    # cols, valid) mapping embeds the shard's cells, so per-invoke scoping
+    # makes cross-shard collisions impossible by construction. A full-year
+    # Sentinel-2 shard has exactly two distinct source grids (10 m bands,
+    # 20 m scl) — this turns 425 geometry computations into 2.
+    geom_cache: dict = {}
 
     async def _run_all():
         sem = asyncio.Semaphore(k)
@@ -524,6 +552,7 @@ def process_raster_shard(
                             nodata=nodata,
                             region=region,
                             anonymous=anonymous,
+                            geom_cache=geom_cache,
                         )
                         for e in items
                     ]
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -21,7 +21,9 @@ from __future__ import annotations
 
 import asyncio
 import concurrent.futures
+import os
 import re
+import threading
 from urllib.parse import urlparse
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


src/zagg/processing/raster.py
```diff
@@ -66,33 +68,67 @@ def _geo_from_ifd(ifd) -> tuple[int, tuple[float, float, float, float, float, fl
...
+    from async_tiff.store import HTTPStore, LocalStore, S3Store
+
+    kind, loc, region, anonymous = key
+    if kind == "s3":
+        kw: dict = {}
+        if anonymous:
+            kw["skip_signature"] = True
+        if region:
+            kw["region"] = region
+        return S3Store(loc, **kw)
```
espg, line 151: 🤖 *from Claude (review)*

**Sharpest question on the store cache — credential lifetime on warm-cached `anonymous=False` stores.** For the shipped raster path this is a non-issue: `sentinel2_l2a.yaml` sets `anonymous: true` and `runner.py:676` defaults `anonymous` to `True`, so the deployed store carries `skip_signature=True` and signs nothing. But `_store_and_path` also caches signed stores (`anonymous=False`) for the *life of the sandbox*. If a hand-rolled/future caller ever caches a signed `S3Store` and the Lambda role credentials rotate mid-sandbox-life, correctness depends on whether async-tiff's `object_store`-backed `S3Store` resolves credentials **per request** (via a refreshing provider chain — fine) or **snapshots them at construction** (stale after rotation — GETs would start 403-ing on a warm worker). Worth confirming and dropping a one-line note in the `_STORE_CACHE` docstring: today all deployed keys are anonymous so this is latent, but the cache is a signed-store trap the moment `anonymous=False` is used with rotating role creds. Not a blocker.

src/zagg/processing/raster.py
```diff
@@ -66,33 +68,67 @@ def _geo_from_ifd(ifd) -> tuple[int, tuple[float, float, float, float, float, fl
...
-    from async_tiff.store import HTTPStore, LocalStore, S3Store
-
     u = urlparse(href)
     if u.scheme == "s3":
-        kw: dict = {"skip_signature": True} if anonymous else {}
-        if region:
-            kw["region"] = region
-        return S3Store(u.netloc, **kw), u.path.lstrip("/")
-    if u.scheme in ("http", "https"):
+        key = ("s3", u.netloc, region, anonymous)
```
espg, line 118: 🤖 *from Claude (review)*

**Minor / by-design:** the `s3://` branch keys on `region` (the caller-passed value, often `None` — `runner.py:675` sources it from `source_region`), while the vhost branch keys on `region or m["region"]` (the parsed region). So the *same bucket* reached via `s3://bkt/...` (region `None`) and via `https://bkt.s3.us-west-2.amazonaws.com/...` (region `us-west-2`) lands two distinct cache entries → two clients. Harmless duplication, and it cannot happen in one invoke in practice (Earth Search asset hrefs are a single style, so an invoke sees one scheme), but it is unintuitive and not covered by a test — `test_vhost_and_scheme_key_separately_but_cache` only asserts two vhost hrefs *share* a store, never that s3-vs-vhost split. If you want the dedupe, normalize `region` for the `s3://` branch too (or resolve it up front). Fine to leave as-is with a comment.

src/zagg/processing/raster.py
```diff
@@ -155,7 +200,18 @@ async def _sample_one(
     epsg, transform = _geo_from_ifd(ifd)
     shape = (ifd.image_height, ifd.image_width)
     center = _raster_center_lonlat(epsg, transform, shape)
-    rows, cols, valid = grid.sample(cells, f"EPSG:{epsg}", transform, shape)
+    geom_key = (epsg, transform, shape)
+    geom = geom_cache.get(geom_key) if geom_cache is not None else None
```
espg, line 227: 🤖 *from Claude (review)*

**Confirmed sound — the atomicity + key-completeness invariant holds.** Verified: (1) there is genuinely no `await` between the `.get` here and the `geom_cache[geom_key] = geom` store below — `grid.sample` is a blocking CPU call, so the check-compute-store is atomic on the loop and each source grid is computed exactly once per invoke even under `Semaphore(k)` concurrency (the `test_two_computes_for_two_grids` = 2 result confirms it). (2) The key `(epsg, transform, shape)` is *complete* only because `cells` and `grid` are invoke-constant — `cells = grid.children(shard_key)` is computed once at `raster.py:504` and threaded unchanged into every `_sample_one`, and `geom_cache` is freshly allocated per `process_raster_shard` call. That is the load-bearing dependency behind correctness; the in-code comment documents the `await` invariant but not the `cells`-constancy one. Worth one clause ("...key omits `cells`/`grid` because both are invoke-constant") so a future refactor that varies `cells` per item within an invoke does not silently return a stale mapping. Also verified the cached `(rows, cols, valid)` arrays are only ever read downstream (`sample_item_async` copies `valid` before `&=`; `_sample_one` indexes rows/cols into fresh buffers) — no aliasing mutation across bands. And `sample_asset_async` is byte-compatible: it never took/added `geom_cache`, calls `_sample_one` without it → `None` → compute-per-call.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of both phases. **No correctness blockers found** — the two caches are sound. `uv run pytest tests/test_raster.py tests/test_raster_pipeline.py -q` → **69 passed** locally on the branch head. Details in three inline comments; summary below.

### Verified sound
- **Geometry-cache atomicity is genuinely exactly-once.** No `await` between the `geom_cache.get` and the store (`grid.sample` is blocking CPU), so the section is atomic on the loop even under `Semaphore(k)` group concurrency. The key `(epsg, transform, shape)` is complete *because* `cells`/`grid` are invoke-constant (`cells = grid.children(shard_key)` computed once at `raster.py:504`, `geom_cache` freshly allocated per `process_raster_shard`). Per-invoke + per-grid scoping means cross-shard collisions are structurally impossible, and multi-item groups share the one `cells`, so the ownership path is fine.
- **No aliasing bug** from sharing cached `(rows, cols, valid)`: downstream reads only — `sample_item_async` copies `valid` before `&=`, `_sample_one` indexes rows/cols into fresh buffers.
- **`sample_asset_async` is byte-compatible** — signature unchanged, calls `_sample_one` without `geom_cache` → `None` → compute-per-call. Public behavior preserved.
- **Store-cache thread-safety is fine.** The `threading.Lock` critical section holds no `await`, so no loop-blocking-across-yield and no deadlock between the main loop and the `_run_sync`/`sample_asset` running-loop thread fallback. async-tiff stores are Rust (Send+Sync) with no Python-side mutation; sharing one into concurrent tasks is safe. Cache is unbounded but bounded-in-practice by distinct buckets (one for Sentinel-2).

### Non-blocking findings (ranked)
1. **[low / question] Credential lifetime on warm-cached `anonymous=False` S3Store** (`raster.py:128`). Deployed path is `anonymous=True` (`sentinel2_l2a.yaml`, `runner.py:676` default) so this is latent, but a cached *signed* store spanning a mid-sandbox credential rotation is only correct if async-tiff/`object_store` resolves creds per-request rather than at construction. Confirm, and note it in the `_STORE_CACHE` docstring.
2. **[low / by-design] Region-key duplication** (`raster.py:95`). `s3://` keys on caller `region` (often `None`); vhost keys on `region or m["region"]`. Same bucket via both schemes → two clients. Cannot occur within one invoke in practice; not covered by a test.
3. **[nit] Document the `cells`-constancy invariant** at the geom-cache site (`raster.py:204`) alongside the existing `await` invariant, so a future per-item `cells` refactor cannot silently return a stale mapping.
4. **[nit] Test isolation scope.** The autouse `_fresh_cache` fixture only guards `TestStoreAndPath`; other tests (`TestGeometryCache`, `TestSampleAsset`) populate the real module `_STORE_CACHE` with `LocalStore` entries keyed on unique pytest `tmp_path` dirs — harmless (no key reuse across tests) but the cache does accumulate for the session. The counting-shim (`monkeypatch.setattr(ats, "S3Store", ...)` seen by the call-time `from async_tiff.store import S3Store` in `_build_store`) is realistic and correct.

Phase 3 (fleet A/B) is correctly deferred. Recommend addressing 1 (confirm + docstring line) and folding the two doc nits; 2 and 4 are fine to leave.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Folded in 91eb13a — added a "Credential lifetime" paragraph to the `_STORE_CACHE` comment block.

Verified the claim before documenting:
- **Lambda (high confidence):** the execution-role credentials arrive as static env vars valid for the whole sandbox lifetime and do not rotate mid-sandbox, so a sandbox-lifetime cached signed store cannot outlive its creds — safe by construction. This is directly grounded.
- **Off-Lambda IMDS/instance-profile/SSO (moderate-high confidence):** async-tiff's `S3Store` is a wrapper around Rust `object_store`'s `AmazonS3` and exposes a `credential_provider` slot (confirmed via `help(S3Store)` on the installed 0.7.2 wheel). object_store's built-in providers resolve through a *caching* provider that refreshes on expiry per request rather than freezing a token at construction. I could not source-verify that refresh behavior against the compiled binary wheel (the `store` submodule is a native extension with no readable stubs), so the note says as much honestly.

Also documented the caveat: today every deployed key is anonymous (`skip_signature=True`), so this is latent — but an explicit-credential store must not be added to the cache without revisiting, since a statically-supplied token would freeze at construction.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Folded in 5ba2147 — extended the INVARIANT comment at the geometry-cache site. Added a clause spelling out that the key `(epsg, transform, shape)` is complete only because `cells` and `grid` are invoke-constant: `cells = grid.children(shard_key)` is computed once in `process_raster_shard` and threaded unchanged into every `_sample_one`, and `geom_cache` is allocated fresh per call. The note warns that a future refactor varying `cells` (or `grid`) per item/group within one invoke must fold them into the key or drop the cache, else it returns a stale mapping.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Leaving both by design — no code change.

**Dual-scheme key duplication (this thread):** correct that `s3://bkt/...` (region `None`) and `https://bkt.s3.us-west-2.amazonaws.com/...` (region `us-west-2`) would land two cache entries for the same bucket. But as you note it cannot happen within one invoke: Earth Search asset hrefs are a single URL style, so an invoke only ever sees one scheme, and the worst case off that path is one redundant client — a harmless duplication, never a correctness issue. Normalizing `region` up front for the `s3://` branch would add branch-specific parsing logic to guard against a case the deployed catalog can't produce, so the simpler keying stays.

**Test isolation scope:** `test_vhost_and_scheme_key_separately_but_cache` deliberately asserts only the in-scope guarantee that two vhost hrefs *share* a store; it does not assert the s3-vs-vhost split precisely because that split is the accepted harmless duplication above rather than a behavior we want to lock in. Adding a test that pins the two-entry outcome would ossify an implementation detail we'd happily dedupe later, so it's left uncovered on purpose.

Both stand for @espg if you'd prefer the dedupe folded in.

Issue #247, opened 2026-07-16
Raster hive write path: windowed (time, cells) leaves, per-leaf templates and stamps, Lambda hive branch


espg:
🤖 *from Claude*

Implementation of **PR B** from the #237 phasing ([decision record](https://github.com/englacial/zagg/issues/237#issuecomment-4986819344)) — the **raster hive write path**: Sentinel-2 stores adopt the hive layout (espg-ratified on #237) as windowed `(time, cells)` leaves per the `morton-hive/2` convention ([mortie #62 entry](https://github.com/espg/mortie/issues/62#issuecomment-4986809092), D13/D14/D15 in `docs/design/sparse_coverage.md`).

**Sequenced after #246 (PR A, temporal machinery)** — this consumes its windowing config, leaf naming/path arithmetic, stamp time-range fields, and coverage `"full"` machinery. File references are against the PR #241 branch state.

## Scope

- **Per-leaf raster templates**: replace the single flat-store template (`emit_raster_template`, called from `RasterStrategy.run`, `src/zagg/runner.py:648-660`) with per-(shard, window) leaf templates under the hive tree — each leaf's time axis is the acquisitions intersecting that shard × window, known from the catalog at dispatch. Follows the aggregation pattern (`HealpixGrid.emit_shard_template` + `hive.shard_leaf_path`).
- **Commit stamp on the slab writer's completion path**: the leaf's final write is `hive.stamp_commit` (carrying the window's actual time range + acquisition count per D15), wired where `process_raster_shard`'s slab loop completes — the raster analog of what `process_and_write_hive` does for aggregation.
- **Coverage, popcount-decided (D14)**: spatial union across the window's acquisitions; interior shards stamp `encoding: "full"` (no sidecar), edge-of-scene/swath shards write the real bitmap via `hive.encode_coverage_bitmap`/`write_coverage_sidecar`. Per-timestep validity stays data-plane nodata.
- **Lambda branch**: `mode="process_raster"` handler gains the hive branch mirroring the aggregation one (`deployment/aws/lambda_handler.py`), sharing one worker path with the local dispatcher.
- **Validation**: remove the #239 explicit rejection (raster + `store_layout: hive` becomes legal); **`sharded` stays rejected on the raster path** (espg-ratified: per-timestep slab streaming over a ShardingCodec object would read-modify-write it; raster object count is time-axis-dominated anyway).
[...]

PR #248, opened 2026-07-16
morton-hive/2 temporal machinery: windowed leaves, temporal manifest block, time-range stamps, coverage full (issue #246)

### LINKED ISSUE #246: morton-hive/2 temporal machinery: windowed leaves, manifest temporal block, stamp time ranges, coverage full
🤖 *from Claude*

Implementation of **PR A** from the #237 phasing ([decision record](https://github.com/englacial/zagg/issues/237#issuecomment-4986819344)) — the **dataset-agnostic temporal machinery** for `morton-hive/2`. The convention is frozen on the [mortie #62 spec entry](https://github.com/espg/mortie/issues/62#issuecomment-4986809092); the design registry entries are D13/D15 (+ D14's coverage `"full"`) in `docs/design/sparse_coverage.md` (PR #243).

All `file:line` references below are against the **PR #241 branch** (`claude/236-hive-dense-sharding`, the assumed-merged base for this work); its in-flight phase-5 commit (issue #238 `shard_order` removal) may shift `write.py`/`healpix.py`/`config.py` lines slightly — function names are authoritative.

## Scope

- **Config surface**: a template-time windowing declaration (schedule `none` | `yearly` | `monthly` | `daily` | explicit range list; membership timestamp field). Absent = `none` = today's behavior, byte-identical.
- **Manifest**: `build_manifest` / `ensure_manifest` (`src/zagg/hive.py:132,157`) gain the temporal block; `spec` bumps to `"morton-hive/2"` when the block is present (plain `/1` otherwise — full back-compat).
- **Leaf naming + path arithmetic**: `{full_id}_{window}.zarr` (`shard_leaf_path`, `hive.py:89`), split-on-first-`_` in the walker/node-invariant checks (`check_node_invariant`, `hive.py:110`) and reader primitives (`src/zagg/coverage.py`), label generation/parsing per the frozen grammar (lexicographic = chronological; UTC half-open boundaries).
- **Dispatch fan-out**: local + Lambda dispatchers fan per (shard, window) when a schedule is declared; workers filter observations by the window range (the temporal analog of `aoi_mask`) so boundary-straddling granules split correctly.
- **Stamp + coverage**: `stamp_commit` (`hive.py:435`) carries the leaf's actual time range + window label; `build_coverage` (`hive.py:229`) gains the D14 `encoding: "full"` fast path (popcount at stamp time, no sidecar when the subtree is fully covered); the root summary (`build_root_coverage`/`write_root_coverage`, `hive.py:520,598`) gains the time-range union (D15 cache).

## Non-goals

- Raster (that's the companion issue — PR B in the #237 phasing).
- Read-side window concat / xarray surface (PR C; the LIST-walk fallback keeps windowed stores readable meanwhile).
- Pyramids/overview window naming (reserved per D13).

## Why aggregation benefits immediately

Per-year 88S processing, seasonal/pre-post subsets, and append-as-acquired all become window leaves; t-digest mergeability makes mission-scale statistics a read-time merge over yearly leaves — the same approximation class as the worker's existing cross-buffer merge.

## Sequencing

Actionable once PR #241 (issue #236) merges — it refactors the same hive template/write path. Implementation plan follows as a comment.

espg:
Closes #246

Implements the `morton-hive/2` temporal machinery ratified on #237 (design registry D13/D14/D15, merged in #243) and planned + ratified on the #246 thread: time-windowed leaves, the manifest temporal block, window/time-range stamps, the coverage `"full"` fast path, per-(shard, window) dispatch fan-out, and the worker-side observation time filter. The window-label grammar and boundary semantics are FROZEN on the [mortie spec page](https://github.com/espg/mortie/issues/62#issuecomment-4986809092); golden-vector tests pin that table.

## Approach

- **New `src/zagg/windows.py` module** (pure functions, stdlib only): label grammar + validation, `window_label` / `window_range` / `windows_intersecting` (UTC half-open `[start, end)`), the split-on-first-`_` leaf-name parser, and the fixed-offset UTC↔dataset-time conversion (`epoch_offset` with `GPS_MINUS_UTC = 18`, TAI hook included, documented ≤1-leap-second tolerance) per the ratified Q3 answer, plus `union_time_range`/`iso_time_range`.
- **Config**: the ratified nested `output.windowing: {schedule, time_field, epoch, scale, units, windows}` block; absent (or `schedule: none`) = today's behavior, byte-identical. Hive-only, healpix-only; `time_field` must be a base-rate `data_source.variables` column (segment-rate membership deferred — review fold); raster configs rejected pointing at #247. `quarterly` is grammar-reserved and rejected pointedly; the explicit windows list ships in this round. `window_time_filters`/`windowed_cell_config` implement the observation filter as a pair of structured `ge`/`lt` predicates riding the issue #43 machinery — the temporal analog of `aoi_mask`.
- **Manifest** gains the D15 temporal block (`schedule`/`time_field`/`epoch`/`scale`/`units`/`calendar`/`append_policy`/`windows`) and bumps `spec` to `"morton-hive/2"` iff windowing is configured; `temporal` joins the frozen resume keys (a windowing change refuses resume like an orders change); `/1` manifests keep reading. **Stamps** (truth) carry the window label + the actual written `time_range` as ISO-8601 UTC strings (ratified Q2), validated fail-closed. The **root summary** (cache) carries the time-range union, accumulated by `write_root_coverage` and rebuilt by `refresh_root_coverage` from the stamps.
- **Coverage `"full"`** (D14): a popcount at stamp time short-circuits a fully-occupied subtree to `encoding: "full"` with no bitmap sidecar PUT; `bitmap_and` short-circuits reads via the shard's own MOC membership (exact, one fewer GET).
- **Dispatch** fans one work unit per (shard, window) (`runner._windowed_units`, shard-major so the #197 shuffle and biggest-first buckets survive): the shardmap's new per-granule `time_start`/`time_end` subset granules per window; legacy shardmaps degrade conservatively (every granule → every window; `bounds.temporal` enumerates generative windows, with a pointed error otherwise). Window bounds convert to dataset units once at dispatch. Async status keys become `{label}_{window}.json` so windowed units of one shard cannot collide. Both dispatchers ride the shared `process_and_write_hive` worker path; the Lambda event gains an optional `window` key (absent = byte-identical), and the worker response body returns the stamped ISO `time_range` for the root union.

## Phases

- [x] Phase 1 — window primitives (`src/zagg/windows.py`) + golden-vector tests pinned to the frozen mortie table, lexicographic=chronological property test per generative schedule
- [x] Phase 2 — config `output.windowing` block (validation + accessor) + manifest temporal block / `morton-hive/2` spec bump
- [x] Phase 3 — windowed leaf naming (`shard_leaf_path`), node-invariant + walker acceptance of windowed leaves
- [x] Phase 4 — stamp `window`/`time_range`, coverage `encoding: "full"` write + read sides, root-summary time union
- [x] Phase 5 — per-(shard, window) dispatch fan-out (local + Lambda), shardmap per-granule time metadata, worker time filter
- [x] Phase 6 — end-to-end tests + `docs/hive_layout.md` windowing section

All six adversarial self-review rounds are folded (one review + one fold pass per phase; every inline thread carries a reply).

## Testing

- `tests/test_windows.py` (55): golden vectors for every row of the frozen grammar table, split-parser round trips, half-open boundary membership, the lexicographic=chronological property over 300 random instants per generative schedule, window contiguity, epoch conversion (ATLAS SDP exactness, GPS-1980 +18 s, TAI-1958 +37 s, round trips).
- `tests/test_hive_windows.py` (90): config validation (including raster→#247, quarterly, explicit-list well-formedness, base-rate time_field), manifest `/2` + frozen-key resume semantics, windowed leaf paths + node invariant + walker (dedupe, malformed-label skip, bare+windowed siblings), windowed stamps (fail-closed `time_range`), coverage `"full"` both ways (popcount through `process_and_write_hive`, full-vs-bitmap exactness at coarse AOIs, depth-0), root time union (build/write-union/refresh/load), dispatch fan-out units (subsetting, straddler, legacy fallback, explicit), `_invoke_lambda_cell` event shape (windowed + unwindowed-byte-identical), and **`TestYearlyEndToEnd`** — the acceptance fixture through the REAL read path (`_FakeH5` h5coro stub + real `_read_group`, so the injected filters actually run): exact 1/5/6 observation splits including an obs exactly on the boundary instant (later window only), stamp ISO ranges strictly inside the window, backfill leaving committed leaves untouched, idempotent window re-run, root time union.
- `tests/test_lambda_handler.py`: windowed event → windowed leaf + injected filter values + ISO `time_range` in stamp and body; unwindowed event stays bare (spec `/1`, no new keys); setup declares the `/2` manifest.
- `none`-identity pins: unwindowed manifest/stamp/root-envelope key sets pinned exactly; unwindowed dispatch payloads/events byte-identical; all pre-existing hive/#241 parity tests green.
- Local green: `ruff format --check src tests`, `pytest -v` — 2100+ passed. `ruff check src tests` reports one **pre-existing** N818 on `src/zagg/registry.py` (`UnknownCapability`), present on main and untouched here (the CI lint bot selects E,F,W,I). `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` also fails on a clean main checkout in this environment (build-environment dependent) — both flagged rather than "fixed" per §4.

## Questions for review

- **`hive.py` is now ~1030 lines**, over the ~1000 §4 guidance (871 pre-change). The temporal machinery itself lives in `windows.py` (pure) and `config.py` (filter injection) by design, and this PR moved `union_time_range`/`iso_time_range` out — the remaining growth is the windowed-stamp/coverage-full/manifest wiring that belongs with the write path. Raising per §4 instead of splitting unilaterally: the natural follow-up split is the read-side bitmap accessors (`decode_coverage_bitmap`/`read_coverage_bitmap`, ~75 lines) into `coverage.py` ("hive owns the write side; coverage owns consumption"), if that is preferred.
- **Coverage `"full"` is gated on windowing being configured** (only `/2` stores emit it). D14 does not literally condition it on a schedule, but emitting `"full"` on schedule-none stores would break the acceptance criterion that schedule-absent output stays object-for-object identical to pre-change (a fully-occupied shard would drop its sidecar object). The mortie spec files the `"full"` discriminator under the `/2` stamp-envelope change, so gating on `/2` reads consistent — flagging the choice explicitly.
- **Per-granule temporal metadata**: ShardMap granule entries gain `time_start`/`time_end` on newly built maps (from STAC `start_datetime`/`end_datetime`). Legacy shardmaps degrade conservatively (every granule to every window; the worker filter enforces correctness) and need `bounds.temporal` for generative window enumeration — the error message says exactly that. Windowed production runs will want rebuilt shardmaps to avoid the read amplification.
- **Streaming (`aggregation.streaming`) windowed stamps omit `time_range`** (the buffered path pools no columns to min/max). The stamp field is optional by design; if streaming+windowing becomes a production combination, a running min/max in the buffered aggregator is a small follow-up.

### DIFF
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -683,8 +683,11 @@ def _validate_windowing(config: PipelineConfig) -> None:
     page but NOT implemented — rejected with a pointed message. The explicit
     windows list must be well-formed: frozen label grammar, half-open
     ``start < end``, unique labels, non-overlapping ranges. ``time_field``
-    must name a declared ``data_source`` column so the worker actually reads
-    the membership timestamps it filters on.
+    must name a declared BASE-RATE ``data_source`` variable (a flat
+    ``variables`` entry or the base level's ``variables``) — a coordinate or
+    a non-base (segment-rate) level column is rejected, since the stamp
+    ``time_range`` is pooled from read variable columns and window membership
+    is decided per observation.
     """
     from zagg import windows as _windows
 
@@ -729,21 +732,37 @@ def _validate_windowing(config: PipelineConfig) -> None:
             "output.windowing.time_field is required: the per-observation "
             "timestamp column that decides window membership"
         )
-    # The worker can only filter on columns it actually reads: the flat
-    # data_source coordinates/variables plus any readable segment-level
-    # variable a hierarchical level declares (issue #30 mapping form,
-    # ``_segment_variable_names``). The check is unconditional — a windowing
-    # store that declares no columns cannot filter on ``time_field`` at all.
-    declared = (
-        set((config.data_source or {}).get("coordinates", {}))
-        | set((config.data_source or {}).get("variables", {}))
-        | _segment_variable_names(config.data_source or {})
-    )
+    # ``time_field`` must be a BASE-RATE variable column, i.e. a
+    # ``data_source.variables`` entry (the base level reads its columns from
+    # there too — ``_validate_levels`` forbids a base-level ``variables``
+    # mapping). The stamp ``time_range`` (a windowed leaf's headline truth) is
+    # pooled from the read VARIABLE columns (``col_arrays`` in the worker),
+    # never from ``coordinates`` — so a coordinate ``time_field`` would filter
+    # yet silently drop the stamp, and a lat/lon coordinate is not a timestamp
+    # anyway. A non-base (segment-rate) level column is rejected too: window
+    # membership would be decided per whole segment, not per observation,
+    # which is not supported this round (see ``window_time_filters``).
+    ds = config.data_source or {}
+    declared = set(ds.get("variables") or {})
     if time_field not in declared:
+        if time_field in set(ds.get("coordinates") or {}):
+            raise ValueError(
+                f"output.windowing.time_field {time_field!r} is a data_source "
+                f"coordinate; the stamp time_range is pooled from read variable "
+                f"columns and a coordinate is not read as a timestamp. Declare it "
+                f"under data_source.variables (or the base level's variables)."
+            )
+        if time_field in _segment_variable_names(ds):
+            raise ValueError(
+                f"output.windowing.time_field {time_field!r} is declared on a "
+                f"non-base (segment-rate) level; segment-rate window membership is "
+                f"not supported yet (whole segments would be kept or dropped). "
+                f"Declare it on the base level or data_source.variables."
+            )
         raise ValueError(
             f"output.windowing.time_field {time_field!r} is not a declared "
-            f"data_source column (one of {sorted(declared)}); the worker can "
-            f"only filter on columns it reads"
+            f"base-rate data_source column (one of {sorted(declared)}); the "
+            f"worker filters per observation on variable columns it reads"
         )
     epoch = block.get("epoch")
     if epoch is None:
@@ -2032,39 +2051,28 @@ def window_time_filters(config: PipelineConfig, start: float, end: float) -> lis
     predicates on the declared ``time_field`` — ``ge start`` / ``lt end`` in
     DATASET units (converted once at dispatch) — so it rides the existing,
     pushdown-eligible filter machinery (issue #43) on every backend instead
-    of a bespoke row filter. The field's dataset path resolves from
-    ``variables``/``coordinates`` (base level) or a hierarchical ``levels``
-    entry (the filter then carries that level and expands to base rate, the
-    Phase-B path). Appended to :func:`filters_from_data_source`'s normalized
-    output by the hive write path, preserving any declared filters.
+    of a bespoke row filter. ``_validate_windowing`` restricts ``time_field``
+    to a base-rate ``data_source.variables`` entry (the base level reads its
+    columns from there too), so the predicate always filters at base rate
+    (``level=None``, per observation); a coordinate or non-base/segment-rate
+    ``time_field`` is rejected at validation and never reaches here. Appended
+    to :func:`filters_from_data_source`'s normalized output by the hive write
+    path, preserving any declared filters.
     """
     windowing = get_windowing(config)
     if windowing is None:
         raise ValueError("window_time_filters requires a windowed config (output.windowing)")
     field = windowing["time_field"]
-    ds = config.data_source or {}
-    path, level = None, None
-    for key in ("variables", "coordinates"):
-        p = (ds.get(key) or {}).get(field)
-        if isinstance(p, str) and p:
-            path = p
-            break
-    if path is None:
-        for lname, spec in (get_levels(config) or {}).items():
-            lvars = spec.get("variables") if isinstance(spec, dict) else None
-            if isinstance(lvars, dict) and isinstance(lvars.get(field), str):
-                # The base level's columns filter at base rate (level None).
-                path, level = lvars[field], (None if lname == get_base_level(config) else lname)
-                break
-    if path is None:
+    path = ((config.data_source or {}).get("variables") or {}).get(field)
+    if not (isinstance(path, str) and path):
         raise ValueError(
-            f"windowing time_field {field!r} has no dataset path in "
-            f"data_source.variables/coordinates/levels — validate_config should "
-            f"have rejected this configuration"
+            f"windowing time_field {field!r} has no base-rate dataset path in "
+            f"data_source.variables — validate_config should have rejected "
+            f"this configuration"
         )
     return [
-        {"level": level, "dataset": path, "op": "ge", "value": float(start)},
-        {"level": level, "dataset": path, "op": "lt", "value": float(end)},
+        {"level": None, "dataset": path, "op": "ge", "value": float(start)},
+        {"level": None, "dataset": path, "op": "lt", "value": float(end)},
     ]
 
 
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -381,6 +381,12 @@ def validate_config(config: PipelineConfig) -> None:
                 "hive root to bootstrap from)"
             )
 
+    # Temporal windowing block (issue #246, morton-hive/2): nested
+    # output.windowing declares the window schedule + time encoding. Absent =
+    # schedule none = today's behavior; the block itself is hive-only (windowed
+    # leaves are a hive-layout convention), mirroring coverage_moc's posture.
+    _validate_windowing(config)
+
     # Validate bounds structure (optional)
     if config.bounds is not None:
         allowed_keys = {"temporal", "spatial"}
@@ -654,6 +660,135 @@ def _validate_raster_config(config: PipelineConfig) -> None:
             "raster templates do not support sharded: true yet (issue #218); "
             "chunks are (1, cells_per_chunk) — one object per timestep-chunk"
         )
+    # Time-windowed hive leaves are a point-pipeline capability in this round;
+    # the raster (time, cells) product's windowing is its own design (issue
+    # #247). Reject pointedly rather than silently ignoring the block —
+    # extending the early-return validation posture from issue #239.
+    if config.output.get("windowing"):
+        raise ValueError(
+            "raster pipelines do not support output.windowing yet (issue #247); "
+            "drop the block — raster output is a (time, cells) cube, not "
+            "windowed hive leaves"
+        )
+
+
+def _validate_windowing(config: PipelineConfig) -> None:
+    """Validate the ``output.windowing`` block (issue #246, morton-hive/2).
+
+    ``{schedule, time_field, epoch, scale?, units?, windows?}`` — the nested
+    form ratified on the #246 thread. Absent/null is schedule ``none``
+    (today's behavior). A present block requires the hive store layout on a
+    healpix grid (windowed leaf names are a morton-hive convention, like
+    ``coverage_moc``). ``quarterly`` is grammar-reserved on the mortie spec
+    page but NOT implemented — rejected with a pointed message. The explicit
+    windows list must be well-formed: frozen label grammar, half-open
+    ``start < end``, unique labels, non-overlapping ranges. ``time_field``
+    must name a declared ``data_source`` column so the worker actually reads
+    the membership timestamps it filters on.
+    """
+    from zagg import windows as _windows
+
+    block = config.output.get("windowing")
+    if block is None:
+        return
+    if not isinstance(block, dict):
+        raise ValueError(
+            f"output.windowing must be a mapping "
+            f"{{schedule, time_field, epoch, ...}} (got {block!r})"
+        )
+    try:
+        schedule = _windows.check_schedule(block.get("schedule", "none"))
+    except ValueError as e:
+        raise ValueError(f"output.windowing.schedule: {e}") from e
+    grid = config.output.get("grid") or {}
+    if (grid.get("type", "healpix")) != "healpix":
+        raise ValueError(
+            "output.windowing requires a healpix grid (windowed leaves are a "
+            f"morton-hive convention; grid type is {grid.get('type')!r})"
+        )
+    if (config.output.get("store_layout") or "flat") != "hive":
+        raise ValueError(
+            "output.windowing requires output.store_layout: hive (window leaves "
+            "are hive leaf zarrs; the flat shared store has no leaves to window)"
+        )
+    if schedule == "none":
+        return
+    time_field = block.get("time_field")
+    if not isinstance(time_field, str) or not time_field:
+        raise ValueError(
+            "output.windowing.time_field is required: the per-observation "
+            "timestamp column that decides window membership"
+        )
+    declared = {
+        **(config.data_source or {}).get("coordinates", {}),
+        **(config.data_source or {}).get("variables", {}),
+    }
+    if declared and time_field not in declared:
+        raise ValueError(
+            f"output.windowing.time_field {time_field!r} is not a declared "
+            f"data_source column (one of {sorted(declared)}); the worker can "
+            f"only filter on columns it reads"
+        )
+    epoch = block.get("epoch")
+    if epoch is None:
+        raise ValueError(
+            "output.windowing.epoch is required: the dataset's zero time as an "
+            "ISO-8601 UTC instant (e.g. '2018-01-01T00:00:00Z' for ICESat-2 "
+            "delta_time)"
+        )
+    try:
+        _windows.parse_utc(epoch)
+    except ValueError as e:
+        raise ValueError(f"output.windowing.epoch: {e}") from e
+    scale = block.get("scale") or "utc"
+    if scale not in _windows.EPOCH_SCALES:
+        raise ValueError(
+            f"output.windowing.scale must be one of {_windows.EPOCH_SCALES} (got {scale!r})"
+        )
+    units = block.get("units") or "seconds"
+    if units not in _windows.UNIT_SECONDS:
+        raise ValueError(
+            f"output.windowing.units must be one of {tuple(_windows.UNIT_SECONDS)} (got {units!r})"
+        )
+    declared_windows = block.get("windows")
+    if schedule != "explicit":
+        if declared_windows is not None:
+            raise ValueError(
+                f"output.windowing.windows only applies to schedule: explicit "
+                f"(the {schedule} schedule derives its windows from labels)"
+            )
+        return
+    if not isinstance(declared_windows, list) or not declared_windows:
+        raise ValueError(
+            "output.windowing.windows is required for schedule: explicit — a "
+            "non-empty list of {label, start, end} entries"
+        )
+    seen: dict = {}
+    for entry in declared_windows:
+        if not isinstance(entry, dict) or not {"label", "start", "end"} <= set(entry):
+            raise ValueError(
+                f"each explicit window must be a {{label, start, end}} mapping (got {entry!r})"
+            )
+        try:
+            label = _windows.validate_label(entry["label"], "explicit")
+            start, end = _windows.parse_utc(entry["start"]), _windows.parse_utc(entry["end"])
+        except ValueError as e:
+            raise ValueError(f"output.windowing.windows: {e}") from e
+        if not start < end:
+            raise ValueError(
+                f"explicit window {label!r} is not half-open: start "
+                f"{entry['start']!r} must precede end {entry['end']!r}"
+            )
+        if label in seen:
+            raise ValueError(f"explicit window label {label!r} is declared twice")
+        seen[label] = (start, end)
+    ordered = sorted(seen.items(), key=lambda kv: kv[1][0])
+    for (a, (_sa, ea)), (b, (sb, _eb)) in zip(ordered, ordered[1:]):
+        if ea > sb:
+            raise ValueError(
+                f"explicit windows {a!r} and {b!r} overlap; windows must be "
+                f"disjoint half-open ranges"
+            )
 
 
 def get_raster_bands(config: PipelineConfig) -> dict:
@@ -1832,6 +1967,46 @@ def get_coverage_moc(config: PipelineConfig) -> bool:
     if flag is None:
         return get_store_layout(config) == "hive"
     return bool(flag)
+
+
+def get_windowing(config: PipelineConfig) -> dict | None:
+    """The normalized temporal windowing declaration, or ``None`` (issue #246).
+
+    ``None`` — absent block, null block, or an explicit ``schedule: none`` —
+    is today's unwindowed behavior (bare leaf names, ``morton-hive/1``).
+    Otherwise a normalized dict with defaults resolved::
+
+        {"schedule", "time_field", "epoch", "scale", "units", "windows"}
+
+    ``epoch`` and explicit-window boundaries are canonicalized to ISO-8601
+    UTC strings; ``windows`` is ``None`` except for ``schedule: explicit``.
+    The same dict feeds the manifest temporal block
+    (:func:`zagg.hive.build_manifest`) and the dispatch fan-out, so the two
+    can never disagree.
+    """
+    from zagg import windows as _windows
+
+    block = config.output.get("windowing")
+    if not block or block.get("schedule", "none") == "none":
+        return None
+    declared = None
+    if block["schedule"] == "explicit":
+        declared = [
+            {
+                "label": w["label"],
+                "start": _windows.iso_utc(_windows.parse_utc(w["start"])),
+                "end": _windows.iso_utc(_windows.parse_utc(w["end"])),
+            }
+            for w in block["windows"]
+        ]
+    return {
+        "schedule": block["schedule"],
+        "time_field": block["time_field"],
+        "epoch": _windows.iso_utc(_windows.parse_utc(block["epoch"])),
+        "scale": block.get("scale") or "utc",
+        "units": block.get("units") or "seconds",
+        "windows": declared,
+    }
 
 
 def get_aoi_mask(config: PipelineConfig) -> bool:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -2025,6 +2025,75 @@ def get_windowing(config: PipelineConfig) -> dict | None:
     }
 
 
+def window_time_filters(config: PipelineConfig, start: float, end: float) -> list[dict]:
+    """Structured filters implementing one window's ``[start, end)`` (issue #246).
+
+    The observation-level window filter is exactly a pair of structured
+    predicates on the declared ``time_field`` — ``ge start`` / ``lt end`` in
+    DATASET units (converted once at dispatch) — so it rides the existing,
+    pushdown-eligible filter machinery (issue #43) on every backend instead
+    of a bespoke row filter. The field's dataset path resolves from
+    ``variables``/``coordinates`` (base level) or a hierarchical ``levels``
+    entry (the filter then carries that level and expands to base rate, the
+    Phase-B path). Appended to :func:`filters_from_data_source`'s normalized
+    output by the hive write path, preserving any declared filters.
+    """
+    windowing = get_windowing(config)
+    if windowing is None:
+        raise ValueError("window_time_filters requires a windowed config (output.windowing)")
+    field = windowing["time_field"]
+    ds = config.data_source or {}
+    path, level = None, None
+    for key in ("variables", "coordinates"):
+        p = (ds.get(key) or {}).get(field)
+        if isinstance(p, str) and p:
+            path = p
+            break
+    if path is None:
+        for lname, spec in (get_levels(config) or {}).items():
+            lvars = spec.get("variables") if isinstance(spec, dict) else None
+            if isinstance(lvars, dict) and isinstance(lvars.get(field), str):
+                # The base level's columns filter at base rate (level None).
+                path, level = lvars[field], (None if lname == get_base_level(config) else lname)
+                break
+    if path is None:
+        raise ValueError(
+            f"windowing time_field {field!r} has no dataset path in "
+            f"data_source.variables/coordinates/levels — validate_config should "
+            f"have rejected this configuration"
+        )
+    return [
+        {"level": level, "dataset": path, "op": "ge", "value": float(start)},
+        {"level": level, "dataset": path, "op": "lt", "value": float(end)},
+    ]
+
+
+def windowed_cell_config(config: PipelineConfig, window: dict) -> tuple[PipelineConfig, dict]:
+    """One work unit's config with its window filter injected (issue #246).
+
+    Returns ``(config, windowing)``: a per-unit config copy whose normalized
+    filter list gains the :func:`window_time_filters` pair for ``window``'s
+    dataset-unit ``[start, end)`` (declared filters preserved — the explicit
+    list wins over ``quality_filter`` sugar, so normalize-then-append), plus
+    the normalized windowing declaration. Raises if the config declares no
+    ``output.windowing`` — a dispatched window without one is dispatcher/
+    config drift, never guessed around.
+    """
+    from dataclasses import replace
+
+    windowing = get_windowing(config)
+    if windowing is None:
+        raise ValueError(
+            "a window was dispatched but the config declares no output.windowing "
+            "block — dispatcher/config drift, refusing to guess the time_field"
+        )
+    ds = dict(config.data_source)
+    ds["filters"] = filters_from_data_source(ds) + window_time_filters(
+        config, window["start"], window["end"]
+    )
+    return replace(config, data_source=ds), windowing
+
+
 def get_aoi_mask(config: PipelineConfig) -> bool:
     """Whether the optional strict-AOI cell mask is enabled (issue #101).
 
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -700,6 +700,18 @@ def _validate_windowing(config: PipelineConfig) -> None:
         schedule = _windows.check_schedule(block.get("schedule", "none"))
     except ValueError as e:
         raise ValueError(f"output.windowing.schedule: {e}") from e
+    if schedule == "none":
+        # An inert ``schedule: none`` block is equivalent to an absent block
+        # (``get_windowing`` returns ``None``, no windowed output), so it must
+        # NOT require the hive/healpix layout — this early return precedes the
+        # layout guards. A stray ``windows`` key is still rejected, mirroring
+        # the generative-schedule check below (validation symmetry).
+        if block.get("windows") is not None:
+            raise ValueError(
+                "output.windowing.windows only applies to schedule: explicit "
+                "(schedule: none produces no windowed output)"
+            )
+        return
     grid = config.output.get("grid") or {}
     if (grid.get("type", "healpix")) != "healpix":
         raise ValueError(
@@ -711,19 +723,23 @@ def _validate_windowing(config: PipelineConfig) -> None:
             "output.windowing requires output.store_layout: hive (window leaves "
             "are hive leaf zarrs; the flat shared store has no leaves to window)"
         )
-    if schedule == "none":
-        return
     time_field = block.get("time_field")
     if not isinstance(time_field, str) or not time_field:
         raise ValueError(
             "output.windowing.time_field is required: the per-observation "
             "timestamp column that decides window membership"
         )
-    declared = {
-        **(config.data_source or {}).get("coordinates", {}),
-        **(config.data_source or {}).get("variables", {}),
-    }
-    if declared and time_field not in declared:
+    # The worker can only filter on columns it actually reads: the flat
+    # data_source coordinates/variables plus any readable segment-level
+    # variable a hierarchical level declares (issue #30 mapping form,
+    # ``_segment_variable_names``). The check is unconditional — a windowing
+    # store that declares no columns cannot filter on ``time_field`` at all.
+    declared = (
+        set((config.data_source or {}).get("coordinates", {}))
+        | set((config.data_source or {}).get("variables", {}))
+        | _segment_variable_names(config.data_source or {})
+    )
+    if time_field not in declared:
         raise ValueError(
             f"output.windowing.time_field {time_field!r} is not a declared "
             f"data_source column (one of {sorted(declared)}); the worker can "
--- a/src/zagg/coverage.py
+++ b/src/zagg/coverage.py
@@ -35,6 +35,7 @@ from zagg.hive import (
     read_manifest,
     read_root_coverage,
     root_coverage_words,
+    union_time_range,
 )
 from zagg.store import open_object_store
 from zagg.windows import split_leaf_name
@@ -108,20 +109,35 @@ def box_and(coverage: dict, aoi) -> np.ndarray:
 
 
 def bitmap_and(leaf_root: str, aoi, **store_kwargs) -> np.ndarray | None:
-    """Exact cell-level intersection via a leaf's bitmap sidecar.
+    """Exact cell-level intersection via a leaf's coverage (bitmap or full).
 
-    One opt-in sidecar GET (:func:`zagg.hive.read_coverage_bitmap`), then
-    ``moc_and`` against ``aoi``. ``None`` when the leaf carries no bitmap
-    (box-only phase-1 stamp, depth-0 config, debris, absent leaf) — the
-    caller falls back to the box verdict; a present-but-corrupt sidecar
-    raises (the decoder's posture). An empty array is a definitive miss:
-    the bitmap is exact, not conservative.
+    Reads the stamp envelope once. ``encoding: "full"`` (issue #246, D14 —
+    whole-subtree coverage, no sidecar object exists) short-circuits to the
+    shard's own MOC membership: ``moc_and`` against the shard id resolves
+    containment exactly, with no bitmap GET and no cell expansion. The
+    ``"bitmap"`` path pays the one opt-in sidecar GET
+    (:func:`zagg.hive.read_coverage_bitmap`, envelope reused). ``None`` when
+    the leaf carries neither (box-only phase-1 stamp, depth-0 config, debris,
+    absent leaf) — the caller falls back to the box verdict; a
+    present-but-corrupt sidecar raises (the decoder's posture). An empty
+    array is a definitive miss: both encodings are exact, not conservative.
     """
-    occupied = read_coverage_bitmap(leaf_root, **store_kwargs)
-    if occupied is None:
+    from zagg.hive import read_coverage
+    from zagg.store import open_store
+
+    coverage = read_coverage(open_store(leaf_root, **store_kwargs))
+    if not coverage:
         return None
     from mortie import moc_and
 
+    if coverage.get("encoding") == "full":
+        from zagg.grids.morton import morton_word
+
+        word = morton_word(split_leaf_name(leaf_root.rstrip("/").rsplit("/", 1)[-1])[0])
+        return moc_and(np.asarray([word], dtype=np.uint64), np.asarray(aoi, dtype=np.uint64))
+    occupied = read_coverage_bitmap(leaf_root, coverage=coverage, **store_kwargs)
+    if occupied is None:
+        return None
     return moc_and(occupied, np.asarray(aoi, dtype=np.uint64))
 
 
@@ -222,7 +238,8 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
     order = int(manifest["shard_order"])
     store = open_object_store(store_root, **store_kwargs)
     root = store_root.rstrip("/")
-    keys = []
+    keys: list = []
+    time_ranges: list = []
     stack = [""]
     while stack:
         prefix = stack.pop()
@@ -243,7 +260,8 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
                         f"listed in {ROOT_COVERAGE_NAME}"
                     )
                     continue
-                if read_commit(open_store(f"{root}/{rel}", **store_kwargs)) is None:
+                stamp = read_commit(open_store(f"{root}/{rel}", **store_kwargs))
+                if stamp is None:
                     continue  # unstamped debris (D4)
                 if _decimal_order(decimal) != order:
                     # A fixed-order ranges MOC cannot represent this leaf:
@@ -259,6 +277,10 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
                     )
                     continue
                 keys.append(morton_word(decimal))
+                # D15: windowed stamps carry the leaf's actual time range;
+                # the rebuilt root summary re-derives the union from this
+                # walk's stamps (truth), superseding any cached value.
+                time_ranges.append(stamp.get("time_range"))
                 continue
             is_digit_node = (
                 _is_base_component(name) if prefix == "" else len(name) == 1 and name in "1234"
@@ -272,7 +294,9 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
         except (FileNotFoundError, NotFoundError):
             pass
         return None
-    envelope = build_root_coverage(keys, order, source="refresh")
+    envelope = build_root_coverage(
+        keys, order, source="refresh", time_range=union_time_range(*time_ranges)
+    )
     obstore.put(store, ROOT_COVERAGE_NAME, json.dumps(envelope, indent=1).encode())
     return envelope
 
--- a/src/zagg/coverage.py
+++ b/src/zagg/coverage.py
@@ -223,8 +223,9 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
     the ranges envelope has no empty form). Windowed leaves (issue #246)
     classify as data via the frozen first-``_`` name split; several stamped
     windows of one shard are one covered shard (the MOC is spatial — the
-    builder de-duplicates words), and a malformed window label is skipped
-    with a warning, like the foreign-order carve-out.
+    builder de-duplicates words), and a malformed window label on a STAMPED
+    leaf is skipped with a warning, like the foreign-order carve-out —
+    unstamped malformed debris stays silent, as ordinary debris does.
     """
     import obstore
     from obstore.exceptions import NotFoundError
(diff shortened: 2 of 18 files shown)


src/zagg/windows.py
```diff
@@ -0,0 +1,310 @@
...
+
+#: First instant of the current constant-offset era (the 2017-01-01 leap
+#: second is the most recent one declared).
+_OFFSET_ERA = datetime(2017, 1, 1, tzinfo=timezone.utc)
+
+#: Explicit-list label charset (frozen): opaque, no ``_`` by construction —
+#: the superset grammar every window label (generative included) satisfies.
+_LABEL_RE = re.compile(r"^[0-9A-Za-z-]{1,32}$")
+_GENERATIVE_RE = {
+    "yearly": re.compile(r"^\d{4}$"),
```
espg: 🤖 *from Claude (review)*

[non-blocking] The generative label patterns use `\d`, which matches Unicode digits (e.g. `٢٠٢٥`, `２０２５`), not just ASCII `0-9`. `int()` also accepts those, so `validate_label("٢٠٢٥", "yearly")` and `window_range` would succeed on a label that is *not* in the frozen `[0-9A-Za-z-]` charset — and would then fail `leaf_name`/`split_leaf_name` (which use `_LABEL_RE`), breaking the round-trip and the "lexicographic = chronological, S3-safe" invariant the grammar exists to guarantee. `window_label` only ever emits ASCII, so generation is safe; the asymmetry is only on externally-supplied labels (config/inventory paths in later phases). Suggest `[0-9]{4}`/`{6}`/`{8}` to keep the generative REs a strict subset of the frozen charset.

src/zagg/windows.py
```diff
@@ -0,0 +1,310 @@
...
+
+    ``epoch`` is the dataset's zero as a UTC instant (ISO string or
+    ``datetime``). See the module docstring for the fixed-offset semantics
+    and tolerance; e.g. ICESat-2 ``delta_time`` is
+    ``scale="gps", epoch="2018-01-01T00:00:00Z"`` (the ATLAS SDP epoch).
+    """
+    epoch_dt = parse_utc(epoch)
+    offset = epoch_offset(scale)  # validates the scale even when it cancels
+    seconds = (parse_utc(dt) - epoch_dt).total_seconds()
+    if scale != "utc" and epoch_dt < _OFFSET_ERA:
```
espg, line 331: 🤖 *from Claude (review)*

[non-blocking] The leap-offset branch keys only on `epoch < _OFFSET_ERA` — it never looks at `dt`. The module docstring says the offset "is identical at both ends and cancels" for a post-2017 epoch, but "both ends" silently assumes `dt` is also >= 2017-01-01. For the ratified use cases this holds (ICESat-2 dt is always post-2018; GPS/TAI native epochs hit the pre-2017 branch and get the full +18/+37, which is exact by definition), so all three golden vectors are correct. But two inputs fall outside the honest tolerance without any guard:
- post-2017 epoch + a `dt` before 2017-01-01 → the leap delta between `dt` and 2017 is dropped (>1 s possible for a far-past `dt`).
- an arbitrary pre-2017 *non-native* epoch (e.g. 2000-01-01) → assumes offset-at-epoch = 0 and applies the full current offset, off by up to the whole current offset (this is noted in the docstring as "inherits up to the full current offset as error", but silently).

Realistic ICESat-2 conversion is exact, so this is not a blocker — but since these are the swap-point functions, consider either a one-line note in the `utc_to_offset`/`offset_to_utc` docstrings that the <=1 s guarantee holds only when epoch *and* dt are on the same side of 2017-01-01, or a cheap guard/warning when they are not.

src/zagg/windows.py
```diff
@@ -0,0 +1,310 @@
...
+    pattern = _GENERATIVE_RE.get(schedule, _LABEL_RE)
+    if not isinstance(label, str) or not pattern.match(label):
+        raise ValueError(
+            f"window label {label!r} does not match the {schedule} schedule's "
+            f"grammar ({pattern.pattern}; frozen on the morton-hive/2 spec)"
+        )
+    return label
+
+
+def window_range(label: str, schedule: str, windows: list[dict] | None = None):
```
espg: 🤖 *from Claude (review)*

[non-blocking] `window_range` has no return annotation, so mypy infers it and downstream callers in the coming config/manifest/dispatch phases (D15 says these consume it) get no type checking on the `(start, end)` tuple. Since it is a public primitive, an explicit `-> tuple[datetime, datetime]` would document the half-open pair and type-check the consumers. (`parse_utc(value)` similarly leaves `value` untyped — `str | datetime` would match its two branches.) Default mypy does not flag either today, so this is polish, not a CI issue.

Related: `validate_label` intentionally defers calendar validity to `window_range`, so `validate_label("202513", "monthly")` / `"20250000"` pass on width alone. That is fine for the leaf-name split guarantee, but note that any phase-2 path that validates a *generative* label without also decoding its range will accept a non-date; worth keeping in mind when config validation is wired up.

tests/test_windows.py
```diff
@@ -0,0 +1,288 @@
...
+    def test_daily(self):
+        # | daily | YYYYMMDD | -31123_20251103.zarr | calendar day |
+        assert w.window_label(_dt(2025, 11, 3, 23, 59, 59), "daily") == "20251103"
+        assert w.window_range("20251103", "daily") == (_dt(2025, 11, 3), _dt(2025, 11, 4))
+        assert w.leaf_name("-31123", "20251103") == "-31123_20251103.zarr"
+        # Leap day decodes; a non-date of the right width does not.
+        assert w.window_range("20240229", "daily")[0] == _dt(2024, 2, 29)
+        with pytest.raises(ValueError):
+            w.window_range("20230229", "daily")
+
```
espg, line 85: 🤖 *from Claude (review)*

[non-blocking] Test-coverage gaps (the suite is otherwise thorough and pins the frozen table well):
- `window_range` rejecting a *width-valid but non-calendar* generative label is tested only for `daily` (`20230229`). The `monthly` month-out-of-range path (`"202513"`, `"202500"`) and the `yearly` `"0000"` path — where the raise comes from `datetime(...)`, not the regex — are untested. A one-line `pytest.raises` per schedule would lock in that these malformed-but-well-formed labels fail.
- The pre-2017 `scale="tai"` conversion (the `+37` branch) is not pinned by any golden vector — `test_gps_native_epoch` covers the `+18` branch, and tai only appears in the round-trip where the 2018 epoch cancels the offset. A single `utc_to_offset(..., epoch="1958-01-01T00:00:00Z", scale="tai")` assertion would cover it symmetrically with the GPS case.

src/zagg/windows.py
espg: 🤖 *from Claude*

Folded in b491183. Switched the generative REs to `^[0-9]{4}$`/`{6}`/`{8}` (dropped `\d`), with a comment noting they must stay a strict subset of the frozen `_LABEL_RE` charset so an externally supplied label round-trips through `leaf_name`/`split_leaf_name`.

src/zagg/windows.py
espg: 🤖 *from Claude*

Folded in b491183. Added the explicit caveat to both `utc_to_offset` and `offset_to_utc` docstrings: the <= 1 s guarantee holds only when `epoch` **and** `dt` (resp. the resulting instant) lie on the same side of 2017-01-01, since the branch keys on `epoch` alone. No runtime guard imposed, per the ratified decision that legitimate ICESat-2/native-epoch conversion stays exact.

src/zagg/windows.py
espg: 🤖 *from Claude*

Folded in b491183. Added `-> tuple[datetime, datetime]` to `window_range` and typed `parse_utc(value: str | datetime)` to match its two branches. The related `validate_label`-defers-calendar-validity note is now also pinned by tests (see the test-coverage thread): out-of-range generative labels are exercised for yearly/monthly/daily.

tests/test_windows.py
espg: 🤖 *from Claude*

Folded in b491183. Added `pytest.raises` for the width-valid-but-non-calendar generative labels: `window_range("0000", "yearly")` and `window_range("202513"/"202500", "monthly")` (raise sourced from `datetime()`, not the regex), alongside the existing daily `20230229`. Also added `test_tai_native_epoch` pinning the pre-2017 `scale="tai"` +37 branch (epoch 1958-01-01), symmetric with `test_gps_native_epoch`.

src/zagg/config.py
```diff
@@ -654,6 +660,135 @@ def _validate_raster_config(config: PipelineConfig) -> None:
...
+        raise ValueError(
+            "output.windowing requires a healpix grid (windowed leaves are a "
+            f"morton-hive convention; grid type is {grid.get('type')!r})"
+        )
+    if (config.output.get("store_layout") or "flat") != "hive":
+        raise ValueError(
+            "output.windowing requires output.store_layout: hive (window leaves "
+            "are hive leaf zarrs; the flat shared store has no leaves to window)"
+        )
+    if schedule == "none":
```

[non-blocking] A `{schedule: none}` block is not treated as equivalent to an absent block, in two directions that both cut against the "absent = schedule none = today's behavior" framing in the block comment and the `/1`-is-`/2`-with-`none` spec claim:

1. The hive/healpix guards (lines 704, 709) run *before* the `schedule == "none"` early-return, so an inert `{schedule: none}` block is hard-rejected on a flat or rectilinear store — even though `get_windowing` returns `None` for it and it produces no windowed output. A config/template that always emits `windowing: {schedule: none}` would fail validation on a flat store while dropping the block entirely passes.
2. Because this returns before the `time_field`/`epoch`/`windows` checks, a stray `windows`, `time_field`, or `epoch` under `schedule: none` is silently accepted, whereas under a generative schedule a stray `windows` list is rejected ("windows only applies to schedule: explicit"). That is a validation-completeness asymmetry — malformed extra keys pass silently under `none` only.

Same theme on the raster path: `_validate_raster_config` rejects on `config.output.get("windowing")` truthiness, so an inert `{schedule: none}` block yields the "#247 not supported" error rather than being ignored. If the intent is that any *present* block requires hive+healpix (as the docstring states), consider saying so explicitly; if `none` is meant to be fully inert, the early-return should precede the layout guards.

src/zagg/config.py
```diff
@@ -654,6 +660,135 @@ def _validate_raster_config(config: PipelineConfig) -> None:
...
+    if not isinstance(time_field, str) or not time_field:
+        raise ValueError(
+            "output.windowing.time_field is required: the per-observation "
+            "timestamp column that decides window membership"
+        )
+    declared = {
+        **(config.data_source or {}).get("coordinates", {}),
+        **(config.data_source or {}).get("variables", {}),
+    }
+    if declared and time_field not in declared:
```

[non-blocking] The `declared` guard is `if declared and time_field not in declared` — so when a config declares neither `coordinates` nor `variables` (both absent/empty), `declared` is falsy and any `time_field` is accepted unchecked, losing the config-time reject the message promises. For a point pipeline the read set is always non-empty (atl06 has `coordinates`/`variables`), so this only bites a degenerate config, but the soft-skip is silent. Worth either a one-line comment noting the "can't check what isn't declared" intent, or dropping the `declared and` guard so an empty read set is itself rejected (a windowing store that reads no columns can't filter on `time_field` anyway). No test exercises the empty-declared branch.

src/zagg/hive.py
```diff
@@ -152,6 +165,24 @@ def build_manifest(grid, dataset: dict | None = None) -> dict:
         "pyramid": {"orders": [], "aggregation": {}},
         "generated_at": _utcnow(),
     }
+    if windowing:
+        temporal = {
+            "schedule": windowing["schedule"],
+            "time_field": windowing["time_field"],
+            "epoch": windowing["epoch"],
+            "scale": windowing["scale"],
+            "units": windowing["units"],
```
espg, line 186: 🤖 *from Claude (review)*

[non-blocking] The D15 temporal block is specified as recording "time encoding/units/**calendar**/epoch" (docs/design/sparse_coverage.md §3, and the frozen mortie comment). The manifest here writes `schedule`, `time_field`, `epoch`, `scale`, `units`, `append_policy` — no `calendar` field. For the three supported scales (utc/gps/tai) the calendar is implicitly proleptic Gregorian, so nothing is wrong today, but the spec text lists calendar as part of the block. Either add a `calendar` key (e.g. defaulting to "gregorian") to match the frozen spec, or note in the docstring/spec that calendar is intentionally deferred while only Gregorian-derived scales exist, so a later reader doesn't read the omission as a bug.

src/zagg/coverage.py
```diff
@@ -226,7 +231,18 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
...
             name = rel.split("/")[-1]
             if name.endswith(".zarr"):
-                decimal = name.removesuffix(".zarr")
+                # Windowed leaves (issue #246, D13): `{full_id}_{window}.zarr`
+                # names split on the first `_` (frozen parse rule); several
+                # windows of one shard collapse to one coverage entry below.
+                try:
+                    decimal, _window = split_leaf_name(name)
+                except ValueError:
+                    logger.warning(
```
espg, line 269: 🤖 *from Claude (review)*

[non-blocking] The malformed-window warn-and-skip sits *before* the `read_commit` stamp gate, so its posture differs from the foreign-order carve-out just below it in two ways worth a second look:

1. It fires for **unstamped** malformed debris too, whereas a valid-named unstamped leaf is silently skipped as D4 debris (line 246-247). So malformed debris is noisier than ordinary debris.
2. It cannot distinguish "torn/foreign debris" from "a real, *stamped* data leaf whose name got mangled" — both emit the same warning and drop from the MOC. If that mangled leaf is the only one, `keys` ends empty and the root `coverage.moc` is deleted (line 269-274), i.e. a store with real (misnamed) data reads as no-coverage.

The foreign-order branch (line 248) deliberately runs *after* `read_commit` so it only warns about **stamped** leaves. Consider moving this skip after the stamp check for the same "only complain about real data" symmetry, and/or wording the message so it does not read as pure debris. Escape-hatch-must-not-die is the right instinct; this is only about which leaves it stays quiet on.

src/zagg/config.py

Folded in 1597706. Moved the `schedule == "none"` early-return ahead of the hive/healpix layout guards in `_validate_windowing`, so an inert `{schedule: none}` block now validates on a flat or rectilinear store — byte-for-byte the "absent = schedule none" posture. Kept validation symmetry by rejecting a stray `windows` key under `none` with the same "only applies to schedule: explicit" message the generative path uses. Tests: `test_schedule_none_inert_on_flat_layout` (validates on flat) and `test_schedule_none_rejects_stray_windows`. Left the raster path (`_validate_raster_config`) as-is — its truthiness reject is out of this fix's scope; a present-block-requires-hive posture there is a separate call for @espg.

src/zagg/config.py

Folded in 1597706. Dropped the `declared and` guard so the check is unconditional — an empty declared set now raises the same "not a declared data_source column" error rather than silently accepting any `time_field`. Checked issue #43 Phase B: hierarchical `levels` DO declare readable columns via the non-base `variables: {name: path-template}` mapping form (issue #30), surfaced by `_segment_variable_names` and broadcast to per-photon columns the worker reads — so `declared` now folds those in alongside `data_source.coordinates`/`variables`. Tests: `test_time_field_rejected_when_no_columns_declared` (empty read set rejected) and `test_time_field_from_level_declared_variable_accepted` (a level-declared `seg_time` is accepted).

src/zagg/hive.py
espg: 🤖 *from Claude*

Folded in 1597706. Added `"calendar": "proleptic_gregorian"` to the manifest temporal block with a comment noting it is the only value this round supports (the three scales utc/gps/tai all derive from stdlib datetime, which is proleptic Gregorian), matching the D15 encoding/units/calendar/epoch listing. Updated `test_yearly_manifest_declares_v2_temporal` to pin the new key.

src/zagg/coverage.py
```diff
@@ -226,7 +231,18 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
             rel = child.rstrip("/")
             name = rel.split("/")[-1]
             if name.endswith(".zarr"):
-                decimal = name.removesuffix(".zarr")
+                # Windowed leaves (issue #246, D13): `{full_id}_{window}.zarr`
+                # names split on the first `_` (frozen parse rule); several
+                # windows of one shard collapse to one coverage entry below.
+                try:
+                    decimal, _window = split_leaf_name(name)
```
espg, line 263: 🤖 *from Claude (review)*

[non-blocking] `check_node_invariant` is the only thing that verifies a leaf`\s id equals its digit chain, but the walker in `refresh_root_coverage` never calls it — it trusts `split_leaf_name(name)[0]` and does `morton_word(decimal)` directly. So a windowed leaf parked at the *wrong* node (e.g. `-5/1/1/2/3/3/3/-9999999_2025.zarr`, id not matching the path) is silently added to the MOC at its own claimed id. This is pre-existing behavior for bare leaves (the walk always trusted the basename over the path) and is faithfully carried to windowed names, so it is not a regression — flagging only because phase 3 is where windowed leaves first enter this trust boundary, and a test asserting the walker still keys off the basename id (not the path) would pin the intended contract.

src/zagg/hive.py
```diff
@@ -103,6 +111,9 @@ def shard_leaf_path(store_root: str, shard_key) -> str:
             f"id with zagg.grids.morton.morton_word first"
         )
     rel = MortonIndexArray.from_words(np.asarray([word], dtype=np.uint64)).hive_path()[0]
+    if window is not None:
+        node, _sep, bare = rel.rpartition("/")
```
espg, line 115: 🤖 *from Claude (review)*

[non-blocking] Robustness note on the `rpartition` surgery: for a degenerate order-0 shard whose `hive_path()` is a bare `{sign+base}.zarr` with no `/`, `rel.rpartition("/")` returns `node=""`, so `rel` becomes `/-5_2025.zarr` (leading slash). It still raises correctly via `check_node_invariant` (after `strip("/")`, `len(parts) == 1 < 2`), same net result as the bare branch, so no wrong prefix escapes — but the two branches reach the raise by different routes. Harmless given order-0 leaves are not valid hive paths anyway; noting only because the task called out the leading-node-component edge. Ids with a `-` sign are handled correctly (`removesuffix` + `leaf_name` preserve the sign — covered by `test_windowed_leaf_at_the_shard_node`).

tests/test_hive_windows.py
```diff
@@ -0,0 +1,447 @@
...
+
+        _windowed(cfg)
+        grid = self._grid(cfg)
+        root = str(tmp_path / "store")
+        from zagg.config import get_windowing
+
+        hive.ensure_manifest(root, hive.build_manifest(grid, windowing=get_windowing(cfg)))
+        word = _shard_word()
+        for label in (*labels, *debris):
+            leaf = hive.shard_leaf_path(root, word, window=label)
```
espg, line 481: 🤖 *from Claude (review)*

[non-blocking] Test gap: the walker tests only ever place windowed leaves at a node (`_windowed_store` emits `window=label` for every leaf). The spec forbids mixing bare and windowed leaves in one store at write time, but the walk "must not misclassify" whatever it finds — a node holding BOTH `-5112333.zarr` (bare) and `-5112333_2025.zarr` should still resolve to a single covered shard via the `np.unique` dedupe in `build_root_coverage`. A test asserting that bare+windowed siblings of the same id collapse to one MOC entry would lock in the dedupe path from the read side (the current dedupe test uses windowed-only leaves). Confirmed `build_root_coverage` dedupes via `np.unique` (hive.py:592), so this is a coverage gap, not a bug.

src/zagg/hive.py
```diff
@@ -422,8 +482,9 @@ def read_coverage_bitmap(leaf_root: str, **store_kwargs) -> np.ndarray | None:
     coverage = read_coverage(open_store(leaf_root, **store_kwargs))
     if not coverage or coverage.get("encoding") != "bitmap" or not coverage.get("sidecar"):
         return None
-    leaf_name = leaf_root.rstrip("/").rsplit("/", 1)[-1]
-    shard = morton_word(leaf_name.removesuffix(".zarr"))
+    # Windowed leaves (issue #246) carry `{full_id}_{window}.zarr` basenames;
+    # the shard id is the part before the first `_` (the frozen parse rule).
+    shard = morton_word(split_leaf_name(leaf_root.rstrip("/").rsplit("/", 1)[-1])[0])
```
espg, line 508: 🤖 *from Claude (review)*

[non-blocking] Behavior-change note (re: cc4fb34 phase-3 change to this line): `split_leaf_name` *raises* `ValueError` on a basename that does not end in `.zarr`, whereas the old `leaf_root...rsplit(...)[-1].removesuffix(".zarr")` was a silent no-op. For every real hive leaf root this is identical (they always end in `.zarr`), so no functional change — but if any caller ever hands `read_coverage_bitmap` a leaf_root without the suffix it now throws instead of best-effort parsing. Arguably stricter-is-better; just confirming no non-`.zarr` leaf_root reaches this read path (I did not find one). No action needed unless such a caller exists.

src/zagg/hive.py
```diff
@@ -443,15 +531,32 @@ def stamp_commit(
...
+        "spec": HIVE_SPEC if window is None else HIVE_SPEC_V2,
         "complete": True,
         "cells_with_data": int(cells_with_data),
         "granule_count": int(granule_count),
         "written_at": _utcnow(),
     }
+    if window is not None:
+        stamp["window"] = str(window)
+        if time_range is not None:
+            stamp["time_range"] = [str(t) for t in time_range]
```
espg, line 579: 🤖 *from Claude (review)*

[non-blocking] The D15 docstring calls the stamp the "TRUTH half" and records "the actual `[t_min, t_max]`", but `stamp_commit` writes `time_range` verbatim (`[str(t) for t in time_range]`) with no validation. A reversed `[t_max, t_min]`, a wrong-length sequence (`["z"]`), or an unparsable instant is stored silently as truth. `union_time_range` then either poisons the cache (a reversed leaf makes `min(starts)`/`max(ends)` produce `start > end`) or drops the whole range on a `lo, hi = r` unpack failure — and any consumer reading `stamp["time_range"]` directly inherits the garbage. The asymmetry is worth reconsidering: fail-open is right for the cache union, but the truth boundary should fail *closed* — assert `len == 2`, `parse_utc` both ends, and `lo <= hi` here before writing. As-is nothing catches a bad range at the point it becomes durable truth.

src/zagg/hive.py
```diff
@@ -595,6 +711,31 @@ def root_coverage_words(envelope: dict) -> np.ndarray:
...
+            continue
+        try:
+            lo, hi = r
+            starts.append(parse_utc(lo))
+            ends.append(parse_utc(hi))
+        except (TypeError, ValueError) as e:
+            logger.warning(f"dropping malformed time_range {r!r} from the union ({e})")
+    if not starts:
+        return None
+    return [iso_utc(min(starts)), iso_utc(max(ends))]
```

[non-blocking] `iso_utc` renders at `timespec="seconds"`, so `union_time_range` floors both endpoints to whole seconds. That is safe for the min (floors earlier → wider) but *narrows* the max (floors earlier → the cached upper bound lands before the true `t_max` when it carries sub-second precision). For an extent summary, an under-reported upper bound could make a consumer that trusts the root `time_range` wrongly exclude an instant in the last fractional second. Cache-only (leaf stamps stay truth), so low severity — but if sub-second window extents are ever possible, consider ceil-rounding the max (or carrying the endpoints at full precision) so the cached union never claims *less* than the leaves.

src/zagg/coverage.py
espg: 🤖 *from Claude*

Folded in d7d4ca7. Moved the malformed-window warn-and-skip to run AFTER the `read_commit` stamp gate. `read_commit` opens the full rel path (not the decimal), so it needs no name split — I call it first, and only a STAMPED leaf whose name then fails `split_leaf_name` warns. Both concerns you raised are addressed: (1) an unstamped malformed leaf is now dropped silently as ordinary D4 debris (no louder than a valid-named unstamped leaf); (2) the warning is reserved for real (misnamed) data, matching the foreign-order branch's "only complain about real data" posture. The escape hatch stays alive either way. `test_refresh_skips_malformed_window_label` now pins both directions — unstamped-malformed stays silent, stamped-malformed warns.

src/zagg/coverage.py
```diff
@@ -107,20 +109,35 @@ def box_and(coverage: dict, aoi) -> np.ndarray:
...
+
+    coverage = read_coverage(open_store(leaf_root, **store_kwargs))
+    if not coverage:
         return None
     from mortie import moc_and
 
+    if coverage.get("encoding") == "full":
+        from zagg.grids.morton import morton_word
+
+        word = morton_word(split_leaf_name(leaf_root.rstrip("/").rsplit("/", 1)[-1])[0])
```
espg, line 135: 🤖 *from Claude (review)*

[non-blocking] Test gap on the "full" short-circuit's exactness. `bitmap_and` claims `moc_and([shard], aoi)` is exactly the bitmap path's `moc_and(all_children_at_cell_order, aoi)`. That equality holds only because `moc_and` compacts/normalizes across orders (the contract `root_coverage_and`/`box_and` rely on) — but every "full" test here (`test_full_leaf_writes_no_sidecar_and_short_circuits`, `test_partial_leaf_keeps_the_bitmap_path`) uses AOIs that are already child-order cells, where both representations trivially agree and no normalization is exercised. The interesting case — an AOI cell *coarser* than `cell_order` but inside the shard, where the bitmap path must compact fine children up to match the coarse short-circuit result — is untested. Suggest a test comparing `bitmap_and` on a `full` leaf against `bitmap_and` on an equivalent fully-populated *bitmap* leaf for a coarse-cell AOI, so a future `moc_and` change that stops compacting would be caught. (Also: no `full`-encoding test at a depth-0 config, where the subtree collapses to the shard cell itself.)

src/zagg/coverage.py
```diff
@@ -243,6 +287,10 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
                     )
                     continue
                 keys.append(morton_word(decimal))
+                # D15: windowed stamps carry the leaf's actual time range;
+                # the rebuilt root summary re-derives the union from this
+                # walk's stamps (truth), superseding any cached value.
+                time_ranges.append(stamp.get("time_range"))
```
espg, line 292: 🤖 *from Claude (review)*

[non-blocking] `refresh_root_coverage` is documented as "THE SANCTIONED ROBUSTNESS PATH … ground truth", yet its time union inherits `union_time_range`'s fail-open posture: a leaf whose stamp carries a malformed `time_range` is dropped with only a `logger.warning`, silently narrowing the rebuilt union. That is defensible for the incremental `write_root_coverage` cache path, but refresh is precisely the recovery path where a corrupt *leaf stamp* (the truth carrier) is most suspicious — a warning buried in logs is easy to miss during a crash-recovery rebuild. Consider making refresh's malformed-range handling louder (surface a count in the return/log summary) so the operator running the escape hatch notices leaf-stamp corruption rather than getting a quietly-narrowed union.

src/zagg/coverage.py
espg: 🤖 *from Claude*

Folded in d7d4ca7 — pinning test only, no walker behavior change (the trust boundary is intentional). Added `test_refresh_trusts_basename_id_over_path_node`: a stamped windowed leaf parked one digit off (`.../3/3/4/-5112333_2025.zarr`) is listed at its basename id `-5112333`, not the path node — asserting the walker keys off `split_leaf_name(name)[0]` and never re-checks `check_node_invariant`. The test comment records that this mirrors pre-#246 behavior for bare leaves (the walk always trusted the basename over the path), so it is the intended contract, not a regression.

src/zagg/hive.py
espg: 🤖 *from Claude*

Intentionally left — no code change. As you noted, the order-0 `rel.rpartition("/")` → `node=""` route still raises correctly via `check_node_invariant` (`len(parts) == 1 < 2` after `strip("/")`), same net result as the bare branch, and order-0 leaves are not valid hive paths anyway, so no wrong prefix escapes. Not worth adding a branch to unify two routes that already reach the correct raise. Flagging it here so the different route is on record.

src/zagg/hive.py
espg: 🤖 *from Claude*

Intentionally left — no code change. As you confirmed, for every real hive leaf root (always `.zarr`-suffixed) `split_leaf_name` is byte-identical to the old `removesuffix(".zarr")` no-op, and no caller hands `read_coverage_bitmap` a non-`.zarr` leaf_root (I did not find one either). Stricter-is-better here: a suffix-less leaf_root is a caller bug worth a raise rather than best-effort parsing. Left as-is; noting it so the behavior change is on record.

tests/test_hive_windows.py
espg: 🤖 *from Claude*

Folded in d7d4ca7. Added `test_refresh_dedupes_bare_and_windowed_siblings`: one node holding BOTH a stamped bare leaf (`-5112333.zarr`) and a stamped windowed leaf (`-5112333_2025.zarr`) of the same id resolves to a single MOC entry. The spec forbids mixing on the WRITE side, but the walk survives it — both stamp to the same shard word and `build_root_coverage`'s `np.unique` dedupe (hive.py:592) collapses them. Complements the existing windowed-only dedupe test by exercising the two name shapes together.

src/zagg/hive.py
espg: 🤖 *from Claude*

Folded in c1d40fe. `stamp_commit` now fails CLOSED on `time_range` before it becomes durable truth: it unpacks a 2-sequence and `parse_utc`s both ends, raising `ValueError` on a wrong length / garbage instant ("must be a 2-sequence of parseable UTC instants"), and separately rejects `t_min > t_max` ("time_range is reversed …"). The production path is unaffected — `process_and_write_hive` still builds the range via `windows.iso_time_range` from the worker min/max, so ordering already holds there; this guards direct callers. Tests in `TestWindowedStamp`: `test_reversed_time_range_rejected`, `test_wrong_length_time_range_rejected` (both the 1- and 3-element cases), `test_garbage_time_range_rejected`.

src/zagg/hive.py

Intentionally left — no code change. On reflection the union does NOT narrow relative to the leaf stamps: the stamp `time_range` is itself rendered by `iso_time_range` → `iso_utc` at `timespec="seconds"` (hive.py, the `iso_time_range(metadata["time_range"], windowing)` call), so the leaf truth is ALREADY floored to whole seconds. `union_time_range` then re-parses those already-seconds strings and re-renders at seconds, so the cached root max is byte-identical to the flooring the stamps carry — it never claims LESS than the leaves. The sub-second flooring happens exactly once, at stamp creation, consistently for both the stamp truth and the union, and is the documented `<= 1 s` window-boundary tolerance (windows.py module docstring).

Per the task guidance I deliberately did not change encoding precision. A ceil-the-max fix would make the union claim MORE than the seconds-precision stamp truth (breaking stamp == union consistency), so it is not a clean one-liner here — the right place for any sub-second precision would be `iso_time_range`/`iso_utc` for BOTH carriers together, which is a larger change left standing for espg if sub-second window extents ever become real.

src/zagg/coverage.py
espg: 🤖 *from Claude*

Folded in c1d40fe. Both gaps now covered in `tests/test_hive_windows.py`:

- `TestCoverageFull.test_full_short_circuit_matches_bitmap_for_coarse_aoi`: builds a "full" leaf AND an equivalent fully-populated *bitmap* leaf (all 4^2 order-8 children encoded in the sidecar, not "full"), then AND-s both against the shard's own order-6 word — an AOI COARSER than `cell_order` 8. The bitmap path must compact its fine children up to match the short-circuit's `moc_and([shard], aoi)`; the test asserts the two representations are array-equal, so a future `moc_and` that stopped compacting would fail here. (Confirmed the normalization is real: `moc_and(all_children_o8, [word_o6])` and `moc_and([word_o6], [word_o6])` both return `[word]`.)
- `TestProcessAndWriteHiveWindowed.test_depth0_full_popcount_short_circuits`: a depth-0 config (`child_order == parent_order`, popcount `4**0 == 1`) run through `process_and_write_hive` — asserts the single occupied cell stamps `encoding: "full"` with no sidecar object, and `bitmap_and` short-circuits. Contrasts the UNWINDOWED depth-0 leaf, which stays box-only (`full` is gated on windowing).

src/zagg/coverage.py
espg: 🤖 *from Claude*

Assessed — the existing warning suffices for refresh; no code change. `refresh_root_coverage` collects each stamp's `time_range` into `time_ranges` and passes them to `union_time_range(*time_ranges)` (coverage.py, the `build_root_coverage(..., time_range=union_time_range(*time_ranges))` call). `union_time_range` already logs a `WARNING` per malformed entry — `"dropping malformed time_range {r!r} from the union (…)"` on logger `zagg.windows` (windows.py) — so a corrupt leaf stamp IS surfaced on the refresh path, naming the offending range, and refresh is an operator-invoked escape hatch (not a silent background write), so those logs are being watched at exactly the moment the review is worried about.

The one thing the current warning does NOT do is aggregate a count into refresh's return/summary. I judged that a marginal add and, more to the point, it would touch `coverage.py`, which is outside this fold's file scope — so I'm leaving a count-in-the-summary enhancement standing for espg rather than making it unilaterally.

src/zagg/processing/worker.py
```diff
@@ -546,6 +558,14 @@ def _iter_granule_reads():
         col_arrays, cell_to_slice, n_obs_total = _concat_and_group(all_reads, grid, handoff)
         logger.info(f"  Read {n_obs_total:,} observations")
 
+    # Actual time extent (issue #246): min/max of the declared time column
+    # over the pooled (post-filter) observations — two reductions on an array
+    # already in hand. The buffered/streaming path holds no pooled columns, so
+    # windowed streaming stamps simply omit their time_range (documented).
+    if time_range_of is not None and time_range_of in col_arrays and n_obs_total:
```
espg, line 565: 🤖 *from Claude (review)*

[non-blocking] The D15 stamp `time_range` (the headline "truth" field of a windowed leaf) is reported here only when `time_range_of in col_arrays`, and `col_arrays` is populated from `data_dict` in `read._read_group` — which contains `variables` + segment broadcasts + `leaf_id`, **not** filter-only datasets, and of the `coordinates` only `latitude`/`longitude` are ever read as columns. But `_validate_windowing` accepts a `time_field` in `coordinates ∪ variables ∪ segment_variable_names`. So a config that declares `time_field` as a *coordinate* (anything other than lat/lon) passes validation, filters correctly (the ge/lt predicates read it into `arrays_by_path`), yet **silently drops** `metadata["time_range"]` here → stamp carries `window` but no `time_range`, and the root-summary union is `None`. No warning anywhere. The test fixture sidesteps this by putting `delta_time` in `variables` (comment: "must be a declared read column"), but nothing enforces the stricter *variables ∪ segment* membership that `time_range` actually needs, and no e2e test exercises the real pooling. Recommend tightening `_validate_windowing` to require `time_field ∈ variables ∪ segment_variable_names` (a non-lat/lon coordinate is never read anyway), so the acceptance criterion "stamps carry the actual UTC time range" cannot silently fail. Related coupling worth a doc line: the only way to get the stamp `time_range` is to make `time_field` an aggregated output variable, which then writes it as an output array.

src/zagg/config.py
```diff
@@ -1834,6 +1985,115 @@ def get_coverage_moc(config: PipelineConfig) -> bool:
...
+        p = (ds.get(key) or {}).get(field)
+        if isinstance(p, str) and p:
+            path = p
+            break
+    if path is None:
+        for lname, spec in (get_levels(config) or {}).items():
+            lvars = spec.get("variables") if isinstance(spec, dict) else None
+            if isinstance(lvars, dict) and isinstance(lvars.get(field), str):
+                # The base level's columns filter at base rate (level None).
+                path, level = lvars[field], (None if lname == get_base_level(config) else lname)
```

[non-blocking] When `time_field` resolves via a hierarchical `levels` entry on a non-base level, `window_time_filters` emits the ge/lt pair with `level=lname` (not `None`). In `read._read_group` that becomes a *coarse-level* (Phase-B) filter: the predicate is evaluated at segment rate and `_expand_mask_to_base` keeps/drops **whole segments**. So a segment that straddles a window boundary is admitted or excluded as a block — the per-observation split that `process_and_write_hive`'s docstring promises ("splits boundary-straddling granules correctly") does not hold for a non-base-level `time_field`; membership is only segment-granular. For ICESat-2 `delta_time` (base rate, `level=None`) this is a non-issue, but nothing restricts `time_field` to base level. Recommend either restricting `time_field` to a base-level column in `_validate_windowing`, or documenting that a level-declared `time_field` windows at segment granularity.

src/zagg/runner.py
```diff
@@ -1384,6 +1385,95 @@ def _aoi_payload_map(catalog_data: dict) -> dict:
...
+                    "windowing with a generative schedule needs per-granule time "
+                    "metadata to enumerate windows, but this shardmap predates it "
+                    "(no time_start/time_end on its granule records) — rebuild the "
+                    "shardmap with `python -m zagg.catalog`, or set bounds.temporal "
+                    "{start_date, end_date} on the run config"
+                )
+            found.update(
+                windows_intersecting(
+                    parse_utc(bounds_temporal["start_date"]),
+                    parse_utc(f"{bounds_temporal['end_date']}T23:59:59"),
```

[non-blocking] The `bounds.temporal` fallback assumes `end_date` is a bare `YYYY-MM-DD` and appends `T23:59:59`. If a config supplies a full ISO timestamp for `end_date` (e.g. `2020-12-31T00:00:00Z`), this produces the malformed `2020-12-31T00:00:00ZT23:59:59`, which `parse_utc` will reject — turning the legacy/untimed-shardmap fallback into a hard crash for a legitimate bounds value. `start_date` is passed straight to `parse_utc` (so it already tolerates either form), but `end_date` does not. Recommend parsing `end_date` first and only extending to end-of-day when it is date-only, or documenting that `bounds.temporal.end_date` must be a bare date.

tests/test_hive_windows.py
```diff
@@ -0,0 +1,1486 @@
...
+        # ...then BACKFILL the earlier window: a new leaf appears next to the
+        # committed ones, which are untouched (D13 — no resize, no rewrite).
+        grid, shard, _ = self._run_units(monkeypatch, c, root, labels={"2018"})
+        assert os.path.exists(hive.shard_leaf_path(root, shard, window="2018"))
+        assert hive.read_commit(open_store(leaf_2019))["written_at"] == stamped_at
+
+        # Idempotent window re-run (D13): same window again is wholesale
+        # replacement, same content.
+        grid, shard, _ = self._run_units(monkeypatch, c, root, labels={"2019"})
+        assert self._leaf_obs(root, shard, "2019", grid) == 5
```
espg, line 1532: 🤖 *from Claude (review)*

[non-blocking] Both `TestYearlyEndToEnd` tests request the `cfg` fixture parameter (`self, monkeypatch, cfg, tmp_path`) but never use it — each builds its own config via `c = self._cfg()` and passes `c` everywhere. Same on `test_backfill_then_root_union_and_idempotent_rerun` (L1508). Drop the unused `cfg` param from both signatures (ruff `ARG` is not in the selected set, so this is not flagged by lint, just dead surface).

tests/test_hive_windows.py
```diff
@@ -0,0 +1,1486 @@
...
+
+        leaf = hive.shard_leaf_path(root, shard, window=label)
+        grp = zarr.open_group(open_store(leaf), path=grid.group_path, mode="r", zarr_format=3)
+        return int(np.asarray(grp["count"][:]).sum())
+
+    def test_boundary_straddling_observations_split_exactly(self, monkeypatch, cfg, tmp_path):
+        from zagg.store import open_store
+
+        root = str(tmp_path / "store")
+        c = self._cfg()
```
espg, line 1482: 🤖 *from Claude (review)*

[non-blocking] The straddler `gC = [729.5, 729.75, 730.25, 730.5]` brackets the 2020 boundary (day 730.0) but never sits ON it, so the half-open tie-break — an observation exactly at the boundary instant belongs to the LATER window (worker filter is `ge start` / `lt end`) — is not exercised through the real read/filter path here. It IS pinned in `tests/test_windows.py` (window_range/windows_intersecting) and in the earlier-phase filter tests, but adding a single obs at exactly `730.0 * DAY` to gC would nail the `ge`-boundary membership through `_read_group` end-to-end (it should land in 2020, not 2019), which is the one branch this acceptance fixture is best placed to certify.

tests/test_lambda_handler.py
```diff
@@ -1100,6 +1100,117 @@ def fake(g, shard_key, urls, **kwargs):
...
+        event = _base_event(cfg_dict)
+        event["shard_key"] = TestProcessHive._WORD
+        event["child_order"] = 12
+        event["store_path"] = str(tmp_path / "hive-out")
+        event["window"] = {"label": "2019", "start": 365 * 86400.0, "end": 730 * 86400.0}
+        resp = handler_mod._handle_process(event, _context())
+        assert resp["statusCode"] == 200, resp["body"]
+
+        # The injected window filter pair reached the worker's config.
+        assert [f["op"] for f in seen["filters"][-2:]] == ["ge", "lt"]
```

[non-blocking] This assertion pins only that the injected filter pair are `ge`/`lt` ops, not their VALUES or dataset path. On the handler-forwarded path the window `{start, end}` come straight off the event and are threaded through `windowed_cell_config` → `window_time_filters` unchanged, so a regression that dropped the conversion, swapped start/end, or resolved the wrong column would still emit `[ge, lt]` and pass here. Value-correctness of the split is only pinned on the LOCAL e2e path (`TestYearlyEndToEnd`). Consider also asserting `[f["value"] for f in seen["filters"][-2:]] == [365*86400.0, 730*86400.0]` and that the filters `dataset` resolves to the delta_time path — cheap, and it closes the handler-forwarding gap.

Separately: the `fake` here never populates the `occupied_out` sink that `process_and_write_hive` passes (the real `process_shard` appends occupied cells to it), so the windowed stamp's coverage-envelope portion (D14 full/bitmap) is degenerate in this test. That is fine — the popcount branches are covered by `test_full_popcount_skips_the_sidecar` / `test_depth0_full_popcount_short_circuits` — just flagging the double contract the fake omits.

src/zagg/processing/worker.py
espg: 🤖 *from Claude*

Folded in a1be7ac. Tightened `_validate_windowing`: `time_field` must now be a base-rate `data_source.variables` entry. A `coordinates`-declared `time_field` is rejected outright (the stamp `time_range` is pooled from read VARIABLE columns / `col_arrays`, never coordinates, and a lat/lon coordinate is not a timestamp anyway), as is a non-base/segment-rate level column (see thread on config.py:2057). The `declared` set is now `set(data_source.variables)` only — the base level reads its columns from `data_source.variables` too, since `_validate_levels` forbids a base-level `variables` mapping. New tests: `test_time_field_coordinate_rejected` (coordinate → rejected) and `test_time_field_base_rate_accepted_in_hierarchical_config` (base-rate variable still accepted alongside a segment-level link).

On the "no e2e test exercises the real pooling" sub-point: that is already addressed — since phase 6, `TestYearlyEndToEnd` drives `delta_time` (a base-rate variable) through the real `_read_group` pooling and asserts the stamp `time_range` verbatim (e.g. `["2019-02-05T00:00:00+00:00", "2019-12-31T18:00:00+00:00"]`), so the "stamps carry the actual UTC range" criterion is now pinned end to end, not just via the fixture.

src/zagg/config.py

Folded in a1be7ac. Took the "restrict `time_field` to base level" option: `_validate_windowing` now rejects a non-base (segment-rate) level column with a pointed message — "segment-rate window membership is not supported yet (whole segments would be kept or dropped)". So the `level=lname` (Phase-B, coarse-level) path can no longer be reached from windowing, and the per-observation split `process_and_write_hive` promises holds for every accepted `time_field`.

`window_time_filters` is simplified to base-rate resolution only: it resolves `time_field` from `data_source.variables` and always emits `level=None` (the levels/coordinates branches are gone). This also dropped a pre-existing mypy `attr-defined` error on the old `spec.get(...)` line.

Updated the fold-added test accordingly: the old `test_time_field_from_level_declared_variable_accepted` put `seg_time` on a NON-base level, so per your call I flipped it to `test_time_field_from_non_base_level_rejected` (expects the "segment-rate window membership" rejection). Base-rate columns stay working, pinned by `test_time_field_base_rate_accepted_in_hierarchical_config` (a hierarchical config whose base-rate `time_field` lives in `data_source.variables`, accepted) and the existing `TestWindowedCellConfig` filter-injection test (still `level=None`).

src/zagg/runner.py

Folded in a1be7ac. The `bounds.temporal` fallback now only appends `T23:59:59` when `end_date` is a bare `YYYY-MM-DD` (a str of length 10 with no `T`); otherwise it is passed to `parse_utc` verbatim. A full ISO instant like `2020-12-31T00:00:00Z` no longer becomes the malformed `...ZT23:59:59` — the legacy/untimed-shardmap fallback stays robust for either bounds form (matching `start_date`, which already went straight to `parse_utc`). New test `test_bounds_full_iso_end_date_accepted` in `TestWindowedUnits` drives an untimed granule with a full-ISO `end_date` and asserts the expected `["2019", "2020"]` fan-out.

tests/test_hive_windows.py
espg: 🤖 *from Claude*

Folded in a86b9bf. Dropped the unused `cfg` fixture param from both `TestYearlyEndToEnd` signatures (`test_boundary_straddling_observations_split_exactly`, `test_backfill_then_root_union_and_idempotent_rerun`) — each already builds its own `c = self._cfg()`, so the fixture was dead.

tests/test_hive_windows.py
espg: 🤖 *from Claude*

Folded in a86b9bf. Added an obs sitting EXACTLY on the boundary instant: gC is now `[729.5, 729.75, 730.0, 730.25, 730.5]`, so day 730.0 (2020-01-01T00:00:00) drives through the real ge/lt filters. Half-open `[start, end)` (ge start) puts it in 2020 ONLY: 2019 stays 5 obs, 2020 becomes 6, and the 2020 stamp `time_range[0]` is now `2020-01-01T00:00:00+00:00` (the boundary itself). Updated the `_fakes` docstring and the boundary-membership comment; backfill test (2019 == 5) and the root union (day 300 .. 802) are unchanged.

tests/test_lambda_handler.py
espg: 🤖 *from Claude*

Folded in a86b9bf. Strengthened the handler assertion to pin the full last-two filter dicts, not just the ops: `level None`, `dataset "/{group}/land_ice_segments/delta_time"`, `op ge` value `365*86400.0`, `op lt` value `730*86400.0` — matching how `test_hive_windows.py` `TestWindowedCellConfig` pins them, so the handler-forwarded window `{start, end}` is now certified through to concrete values and dataset path.

espg:
🤖 *from Claude*

Raising per §4 (module-size guidance): `src/zagg/hive.py` ends this PR at ~1030 lines, over the ~1000-line ceiling (871 pre-change). The temporal machinery proper was kept out of it by design — grammar/conversion/union primitives live in `src/zagg/windows.py` (pure, stdlib-only) and the filter injection in `src/zagg/config.py`, and this PR also moved `union_time_range`/`iso_time_range` OUT of `hive.py` mid-implementation — but the windowed-stamp, coverage-`"full"`, and manifest-temporal wiring belongs with the write path and pushed it over.

Options, in preference order:

1. **Accept ~1030 for this PR** and split in a follow-up: move the read-side bitmap accessors (`decode_coverage_bitmap` + `read_coverage_bitmap`, ~75 lines) to `coverage.py` — consistent with the recorded split ("`hive` owns the write side and the raw accessors; `coverage` owns consumption", PR #208 round 3) and gets `hive.py` back under ~960.
2. Do that split inside this PR as one more commit (touches import sites + a handful of tests; mechanical).
3. A new `hive_root.py` for the root-summary cluster (`build_root_coverage`/`root_coverage_words`/`write_root_coverage`, ~150 lines) — bigger relief, more churn.

Proceeding with (1) unless directed otherwise; nothing in the PR blocks on this.

PR #254, opened 2026-07-16
small fixes 2026-07-16

### LINKED ISSUE #239: raster configs silently ignore store_layout / coverage_moc (validate_config returns early)
🤖 *from Claude*

Found during the 2026-07-15 pipeline-unification review.

## Problem

The raster branch of `validate_config` returns early, **before** the `store_layout` / `coverage_moc` checks (`src/zagg/config.py:347-382`) run, and `RasterStrategy` never consults `store_layout` (`src/zagg/runner.py:648-660`). So a raster config declaring `store_layout: hive` (or `coverage_moc: true`) validates cleanly and then silently writes a flat store with no coverage object. A silent config no-op is worse than a rejection.

## Fix

In `_validate_raster_config` (`src/zagg/config.py:589`), reject explicit `store_layout: hive` and explicit `coverage_moc: true` with a message pointing at #237 ("not supported on the raster path yet — see #237 for the design"). Mirror the message style of the former hive+sharded rejection. One test per rejection (red-before / green-after).

Goes away naturally when #237 lands; until then the config surface should tell the truth.

espg:
Closes #239

Bundled small-fix PR for 2026-07-16 (one open `small-fix` issue today).

## Checklist

- [x] #239 — raster configs silently ignore `store_layout` / `coverage_moc`

## What / approach

The raster branch of `validate_config` dispatches to `_validate_raster_config` and returns before the point-pipeline `store_layout` / `coverage_moc` checks run, and `RasterStrategy` never consults either key — so a raster config declaring `store_layout: hive` or `coverage_moc: true` validated cleanly and then silently wrote a flat store with no coverage object.

This adds two pointed rejections in `_validate_raster_config` (`src/zagg/config.py`), placed between the existing `sharded: true` rejection and the `output.windowing` rejection (which already cites this issue's posture):

- explicit `output.store_layout: hive` → `ValueError` ("not supported on the raster path yet — see issue #237 for the design; drop the key ...")
- explicit `output.coverage_moc: true` → `ValueError` (same shape; explicit `false` remains a harmless opt-out and is not rejected)

`store_layout: flat` (the actual raster behavior) still validates. Goes away naturally when #237 lands.

## How tested

Two new tests in `tests/test_raster_pipeline.py::TestRasterConfigValidation` (`test_hive_store_layout_rejected`, `test_coverage_moc_rejected`), confirmed red before the fix (both `DID NOT RAISE`) and green after. Full local bar: `ruff check` / `ruff format --check` clean on the changed files, `pytest -v` → 2153 passed, 30 skipped.

## Questions for review

- Pre-existing, unrelated to this change: `ruff check src tests` on current `main` fails with `N818` in `src/zagg/registry.py:64` (`UnknownCapability` should have an `Error` suffix). Not fixed here per the no-unrelated-fixes convention; flagging so it can be picked up separately.

---
_Generated by [Claude Code](https://claude.ai/code/session_014suq9BRdBuMihm2XLuwRnx)_

### DIFF
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -664,7 +664,10 @@ def _validate_raster_config(config: PipelineConfig) -> None:
     # capabilities; the raster path always writes the flat (time, cells) store
     # and never consults these keys (issue #239), so an explicit request is a
     # config mistake, not a no-op — reject it pointedly.
-    if config.output.get("store_layout") == "hive":
+    store_layout = config.output.get("store_layout")
+    if store_layout is not None and store_layout not in ("flat", "hive"):
+        raise ValueError(f"output.store_layout must be 'flat' or 'hive' (got {store_layout!r})")
+    if store_layout == "hive":
         raise ValueError(
             "output.store_layout: hive is not supported on the raster path "
             "yet — see issue #237 for the design; drop the key (raster output "
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -670,7 +670,10 @@ def _validate_raster_config(config: PipelineConfig) -> None:
             "yet — see issue #237 for the design; drop the key (raster output "
             "is a flat (time, cells) store)"
         )
-    if config.output.get("coverage_moc") is True:
+    coverage_moc = config.output.get("coverage_moc")
+    if coverage_moc is not None and not isinstance(coverage_moc, bool):
+        raise ValueError(f"output.coverage_moc must be a boolean (got {coverage_moc!r})")
+    if coverage_moc:
         raise ValueError(
             "output.coverage_moc is not supported on the raster path yet — "
             "see issue #237 for the design; drop the flag (there is no hive "
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -660,6 +660,22 @@ def _validate_raster_config(config: PipelineConfig) -> None:
             "raster templates do not support sharded: true yet (issue #218); "
             "chunks are (1, cells_per_chunk) — one object per timestep-chunk"
         )
+    # The hive shard layout and its root coverage MOC are point-pipeline
+    # capabilities; the raster path always writes the flat (time, cells) store
+    # and never consults these keys (issue #239), so an explicit request is a
+    # config mistake, not a no-op — reject it pointedly.
+    if config.output.get("store_layout") == "hive":
+        raise ValueError(
+            "output.store_layout: hive is not supported on the raster path "
+            "yet — see issue #237 for the design; drop the key (raster output "
+            "is a flat (time, cells) store)"
+        )
+    if config.output.get("coverage_moc") is True:
+        raise ValueError(
+            "output.coverage_moc is not supported on the raster path yet — "
+            "see issue #237 for the design; drop the flag (there is no hive "
+            "store root to write coverage.moc to)"
+        )
     # Time-windowed hive leaves are a point-pipeline capability in this round;
     # the raster (time, cells) product's windowing is its own design (issue
     # #247). Reject pointedly rather than silently ignoring the block —
--- a/tests/test_raster_pipeline.py
+++ b/tests/test_raster_pipeline.py
@@ -92,6 +92,18 @@ class TestRasterConfigValidation:
         with pytest.raises(ValueError, match="sharded"):
             validate_config(_raster_config(grid=grid))
 
+    def test_hive_store_layout_rejected(self):
+        cfg = _raster_config()
+        cfg.output["store_layout"] = "hive"
+        with pytest.raises(ValueError, match="issue #237"):
+            validate_config(cfg)
+
+    def test_coverage_moc_rejected(self):
+        cfg = _raster_config()
+        cfg.output["coverage_moc"] = True
+        with pytest.raises(ValueError, match="issue #237"):
+            validate_config(cfg)
+
     def test_rectilinear_grid_rejected_for_now(self):
         grid = {"type": "rectilinear", "crs": UTM18, "resolution": 10, "bounds": [0, 0, 1, 1]}
         with pytest.raises(ValueError, match="healpix"):
--- a/tests/test_raster_pipeline.py
+++ b/tests/test_raster_pipeline.py
@@ -95,13 +95,13 @@ class TestRasterConfigValidation:
     def test_hive_store_layout_rejected(self):
         cfg = _raster_config()
         cfg.output["store_layout"] = "hive"
-        with pytest.raises(ValueError, match="issue #237"):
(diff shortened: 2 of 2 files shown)


src/zagg/config.py
```diff
@@ -660,6 +660,22 @@ def _validate_raster_config(config: PipelineConfig) -> None:
...
+    # capabilities; the raster path always writes the flat (time, cells) store
+    # and never consults these keys (issue #239), so an explicit request is a
+    # config mistake, not a no-op — reject it pointedly.
+    if config.output.get("store_layout") == "hive":
+        raise ValueError(
+            "output.store_layout: hive is not supported on the raster path "
+            "yet — see issue #237 for the design; drop the key (raster output "
+            "is a flat (time, cells) store)"
+        )
+    if config.output.get("coverage_moc") is True:
```

Non-blocking: `coverage_moc is True` catches the exact `coverage_moc: true` the issue names, but a truthy non-bool (`coverage_moc: 1`, `coverage_moc: "yes"`) falls through and is silently ignored — the same silent-no-op class this PR is closing. The point path is stricter: line 369 rejects any non-bool `coverage_moc` (`isinstance(..., bool)`) before it ever tests truthiness. So the raster branch is now more permissive about malformed values than the point branch. `is True` is defensible (it correctly lets explicit `false` through and matches the issue's "explicit true" ask), but consider either mirroring the point-path bool guard or widening to a truthy test so a mistyped value doesn't slip past. Scoped as the issue asked, this is fine to leave — flagging the inconsistency.

---

src/zagg/config.py
```diff
@@ -660,6 +660,22 @@ def _validate_raster_config(config: PipelineConfig) -> None:
             "raster templates do not support sharded: true yet (issue #218); "
             "chunks are (1, cells_per_chunk) — one object per timestep-chunk"
         )
+    # The hive shard layout and its root coverage MOC are point-pipeline
+    # capabilities; the raster path always writes the flat (time, cells) store
+    # and never consults these keys (issue #239), so an explicit request is a
+    # config mistake, not a no-op — reject it pointedly.
+    if config.output.get("store_layout") == "hive":
```

Non-blocking: this rejects only `store_layout: hive`; any other non-`flat` value (a typo like `store_layout: hivve`, or `shard`) is still silently ignored on the raster path. The point path validates the enum fully — line 348 rejects anything not in `("flat", "hive")`. So an explicit-but-misspelled `store_layout` on a raster config now writes flat with no complaint, which is the same silent-config trap the issue is closing, one typo over. The issue scoped this to explicit `hive`, so leaving it is reasonable; noting the asymmetry with the point-path enum check in case you want to reject unknown values too.

---

tests/test_raster_pipeline.py
```diff
@@ -92,6 +92,18 @@ def test_sharded_rejected(self):
         with pytest.raises(ValueError, match="sharded"):
             validate_config(_raster_config(grid=grid))
 
+    def test_hive_store_layout_rejected(self):
+        cfg = _raster_config()
+        cfg.output["store_layout"] = "hive"
+        with pytest.raises(ValueError, match="issue #237"):
```
espg: 🤖 *from Claude (review)*

Nit (non-blocking): both new tests pin on the same `match="issue #237"` token. The two configs are disjoint (one sets `store_layout`, the other `coverage_moc`), so each test does exercise its own rejection path and red-before/green-after holds — but the shared anchor doesn't assert *which* message fired, and a future message edit that drops "#237" from one rejection while keeping it in the other would still pass the wrong test. Consider anchoring on `store_layout` / `coverage_moc` respectively (as `test_sharded_rejected` pins `"sharded"`) so each test is self-documenting and message-specific.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of the issue #239 fix. The core change is correct and matches the issue precisely: the two rejections live inside `_validate_raster_config`, which runs fully (the early return the issue describes is in `validate_config`'s raster branch, not inside this function), so `store_layout: hive` and `coverage_moc: true` now fail loudly instead of silently no-op'ing. Message style mirrors the neighboring sharded/windowing rejections and points at #237 as requested; the comment correctly cites #239 (implementing) vs #237 (design). Tests are genuine red-before/green-after — before the fix validation passed (return early), so `pytest.raises` would have failed. I also checked the shipped configs: no raster YAML declares either key, so nothing existing breaks.

Three non-blocking observations inline — all consistency gaps with the stricter point-path validation (truthy-non-bool `coverage_moc`, non-`hive` invalid `store_layout` strings) plus a test-anchor nit. None block; the change delivers what the issue asked for.

---

src/zagg/config.py

Folded in 4098004. The raster path now mirrors the point-path bool guard: `coverage_moc` is rejected with `output.coverage_moc must be a boolean` when it is a non-bool, before the truthiness check, so `coverage_moc: 1` / `"yes"` no longer slip past. Explicit `false` still passes through as before.

---

src/zagg/config.py

Folded in 8ee4d0e. The raster path now runs the full point-path enum check — anything not in `("flat", "hive")` raises `output.store_layout must be 'flat' or 'hive'` — before the pointed `hive`-not-supported rejection, so a typo like `hivve` no longer writes flat silently.

---

tests/test_raster_pipeline.py
espg: 🤖 *from Claude*

Folded in d4ad412. Each test now pins its own token — `match="store_layout"` and `match="coverage_moc"` respectively (matching the `test_sharded_rejected` style) — so each asserts which message fired and a future edit to one message can't silently satisfy the other test.

---

espg:
🤖 *from Claude*

Run status (2026-07-16): complete, awaiting review.

- Single-phase small-fix bundle for #239 (the only open `small-fix` today): `_validate_raster_config` now rejects explicit `output.store_layout: hive` and explicit `output.coverage_moc: true` with messages pointing at issue #237, plus one test per rejection (red-before/green-after) — commit dfd2462.
- Adversarial self-review posted 3 non-blocking findings; all 3 folded (4098004 non-bool `coverage_moc` guard, 8ee4d0e full `store_layout` enum guard, d4ad412 message-specific test anchors), with a reply on every inline thread. No findings left standing.
- CI green on d4ad412: ruff, build (x86_64 + arm64), test (3.12), test (3.13). Local: full suite 2153 passed / 30 skipped; changed files ruff- and format-clean. The ruff PR bot posted no inline comments.
- Pre-existing on `main`, not touched here: `ruff check src tests` fails N818 in `src/zagg/registry.py:64` (see PR body, Questions for review).
- Note: this routine is configured drafts-only, so the PR is left in draft rather than self-flipped to ready per CLAUDE.md section 2 — the draft-to-ready flip (and merge) is left for review. All phases are done and `waiting` is applied.

---

Issue #251, opened 2026-07-16
Deprecate HEALPix flat/fullsphere → hive; collapse store_layout to rect|hive (1.0 gate)


espg:
🤖 *from Claude*

## Summary

Deprecate and remove the **HEALPix flat / fullsphere** output path in favor of **hive+sharded**, collapsing `store_layout` so the codebase ends with exactly **two** paths: **rect** (bounded, projected) and **hive/healpix** (sparse DGGS). No `store_layout` knob — the grid determines the store.

**This is a gate for cutting the 1.x.x release.** Removing flat is a breaking change, so it must land in **0.x**; 1.0 then ships with the stable `rect | hive` surface. (Shipping 1.0 with a flat path we're about to delete would push the removal to a 2.0.)

## Motivation

- The flat/fullsphere setup Lambda (`_invoke_lambda_setup` → `HealpixGrid.emit_template`) materializes the **whole-globe DGGS coordinate template** to land a handful of shards — a measured **~107 s setup tax** on a 4-shard NEON full-AOI run (hive setup is **3.7 s**; billed but excluded from `total_wall_s` and cost). It's the vestigial eager-xdggs materialization `sparse_coverage.md` §1 pt 3 calls out ("intractable at global orders").
- Hive+sharded is the ratified default (unified-pipeline direction); flat/fullsphere is intractable at scale and we never produce global products.
- Ends the flat/hive/fullsphere layout sprawl — one DGGS path (hive), one gridded path (rect).

## Phased plan

**Phase 1 — deprecate now (no read-path dependency).**
- Mark HEALPix-flat/fullsphere deprecated; hive becomes the default (via #236 dense parity → #240 hive regression arm → default flip).
- Stop adding flat HEALPix benchmark arms; reposition the README/docs quickstart off flat+unsharded.

**Phase 2 — GATE: a morton-native sparse-DGGS READ PATH ships.**
- The domain-as-MOC + lazy dense-view reader (`sparse_coverage.md` §6) that replaces "just `xr.open_zarr` the flat store."
- **Two routes, and this must NOT be hostage to upstream:** #72 (xdggs accepts `MortonIndexDtype`) is `blocked`/`external` — the xdggs maintainers' timeline. **O3 (ship the sparse-DGGS reader as our own standalone xarray extension) is the fallback that keeps this — and therefore 1.0 — on OUR schedule.** If upstream stalls, build O3 and proceed.
[...]

Issue #250, opened 2026-07-16
Benchmark: capture + display per-phase timings (setup, read/index/aggregate) through time


espg:
🤖 *from Claude*

## Summary

Capture and **display per-phase timings through time** in the benchmark series + `docs/deployment/benchmark.md`, so we track where wall/cost actually go release-over-release. Today the full-AOI series records only the coarse orchestration phases and the worker read/index/aggregate split is thrown away — which hid a large, avoidable **setup tax** (see below).

## Motivation (what this surfaced)

On the 0.31.0 full-AOI run, wall time decomposes very unevenly and the two biggest terms were invisible on the chart:

- **`setup_s` ≈ 104–110 s (flat)** — a single Lambda (`_invoke_lambda_setup`, `runner.py:2143`, fired **once per arm before** the 4-worker fan-out, NOT per shard) that runs `HealpixGrid.emit_template` → writes the **entire fullsphere** zarr template (`spec = self._spec()`, the whole global o9 grid) just to land 4 NEON shards. This is the dense-fullsphere waste `docs/design/sparse_coverage.md` §1 calls out.
  - **Hive already collapses this to 3.7 s** (`emit_shard_template` writes one leaf lazily per shard, no global structure) — a ~29× cut. So the setup tax is a flat-fullsphere artifact hive erases (cross-ref #236 / #199).
  - `setup_s` is **billed** (a real Lambda invoke) but **excluded from `total_wall_s`** (measured before `start_time`, `runner.py:2157`) AND **excluded from the cost** (`report.cost.compute_time_s` sums only worker durations) — so flat's true wall AND cost are ~18–40% higher than the chart shows.
- **`fanout_s`** is the worker execution, but we never retain its **read / index / aggregate** split even though the workers already emit it under `profile=True` (`runner.py:2307-2344`, `summary["worker_phase_max"]`; the full-AOI harness runs `profile=True`). It's computed, logged to the runner's INFO logger, then dropped — never in the series, and it didn't surface in the workflow log.

### 1. Capture the worker phase split in the series
`run_full_aoi_benchmark.run_target` already calls `agg(..., profile=True)`, so `summary["worker_phase_max"]` (a `{phase: seconds}` dict, phases = **read / index / aggregate**) is available. Thread it into the run record and flatten it into `full_aoi_series.py`'s `FULL_AOI_COLUMNS`:
- add `phase_read_s`, `phase_index_s`, `phase_aggregate_s` (max-across-shards, matching the wall framing).
[...]

PR #255, opened 2026-07-16
Hive manifest hybrid lifecycle: async init write, finalize backstop (issue #252)

### LINKED ISSUE #252: Fold hive manifest write into finalize (take setup off the critical path)
🤖 *from Claude*

Sequenced after #236 (now **closed** — hive dense-array parity landed), so this is unblocked. Detail originally parked as a comment on #236.

## Summary

Take the hive **setup Lambda off the critical path** by folding its `morton_hive.json` write into the **finalize** step. Removes a synchronous ~3.7 s pre-worker invoke; also revives `finalize` as a meaningful tracked phase.

## Why it's safe

Today the hive dispatch runs a synchronous pre-worker setup Lambda (`_invoke_lambda_setup`, hive mode) that writes `morton_hive.json` + echoes the layout — **~3.7 s**, dominated by fixed Lambda cold-start/round-trip. The manifest is a fixed static descriptor (grid orders + dataset identity + a `parent_order`-length split schedule; **no per-shard enumeration**), so it's stable at ~3.7 s from NEON to CONUS — it does not scale.

But the manifest is **reader-facing**, and the hive workers write **self-describing leaves** (D3) — they never read it. So it isn't structurally on the critical path.

## Change

- Fold the manifest write into the finalize path (`_invoke_lambda_finalize` / the Lambda finalize handler), alongside the root `coverage.moc`. Order-correct (readers only arrive after the run), drops a whole pre-worker invoke, takes ~3.7 s off the wall.
- **Benchmark:** `finalize` stops being ~0 (it now carries the manifest + root MOC) → keep it in the #250 phase panel (don't drop it).

## The fail-fast deploy guard (gated on #251)

Setup's synchronous return also does the stale-deploy check (echoes `"layout":"hive"` so a pre-#199-phase-3 function fails fast instead of silently writing a flat store at the hive root). Finalize-folding moves that signal to the end. Its resolution depends on #251 timing:
- **Folded before flat removal** → keep a **cheap up-front version ping** (one lightweight invoke, decoupled from the manifest) to preserve fail-fast.
- **Folded after #251** (flat gone) → the "silent flat fallback" failure mode no longer exists (a stale function just errors), so drop the ping; early-worker-error abort suffices.

## Sizing / priority

**Low priority** — ~0.05 % of a multi-hour CONUS run. **Not a small-fix:** it changes the setup→finalize lifecycle, touches the Lambda handler (which mode writes the manifest), carries the ping-vs-drop-guard decision (gated on #251), and re-touches the #250 phase display. A regular `implement` task, best sequenced after the #251 flat-removal decision so the guard question resolves itself.

Refs: #236 (closed, prerequisite), #250 (finalize back in the phase panel), #251 (guard rationale disappears when flat is removed), `sparse_coverage.md` §3 (the static manifest).

espg:
Closes #252

Takes the hive setup Lambda off the critical path per the **hybrid lifecycle ratified on the issue** ([make-it-so](https://github.com/englacial/zagg/issues/252#issuecomment-5006627722), [design comment](https://github.com/englacial/zagg/issues/252#issuecomment-5005939446)): fail-fast stays in one cheap sync ping, the `morton_hive.json` write fires as an **async Event invoke of the existing `mode="setup"` hive branch at init**, and finalize's `ensure_manifest` is demoted to an **idempotent backstop**. The ratification also explicitly rejected a dispatcher-direct S3 PUT: CI/CD runs depend on OIDC with lambda-invoke-only permissions, so the orchestrator gains no S3 write access on the lambda path. #251 phase 3 has not landed (flat still exists), so the fail-fast stale-deploy guard is preserved as the up-front version ping, decoupled from the manifest write.

## What / approach

**Lambda dispatcher (`src/zagg/runner.py`)**
- Hive runs dispatch **no synchronous manifest-writing setup invoke** — the ~3.7 s pre-worker round-trip stays gone. Lifecycle: `mode="ping"` (byte-identical to phases 1–3: stale-deploy 400 fall-through + read-only `validate_manifest` precheck) → **`_invoke_lambda_setup_async`**, a fire-and-forget `InvocationType="Event"` invoke of the existing `mode="setup"` hive branch (~10 ms, the root-coverage dispatch precedent) carrying the same manifest inputs (`config`, `parent_order`, `dataset` identity, `overwrite`) → worker fan-out → finalize.
- The manifest therefore lands **seconds into the run**: a reader can start consuming completed leaves while the store builds (progressive reads — the windowed-leaf case from issue #246 makes partial output incrementally consumable by design).
- Hive **always** invokes finalize, whose `ensure_manifest` is now the **idempotent backstop** (a frozen-key-matching manifest is accepted — no second PUT). The backstop is load-bearing: worker Event invokes run with retries 0 (`template.yaml` EventInvokeConfig, issue #151 hygiene), so a lost async init write is never redelivered — finalize self-heals it. Symmetrically, finalize failure is no longer load-bearing for manifest existence, and a run that crashes mid-fan-out still left a manifest at init. The `consolidate_metadata` gate from issue #191 stays flat-only. Flat setup and finalize events stay byte-identical (pinned on the wire in `TestInvokeLambdaSetupEvent` / `TestInvokeLambdaFinalizeEvent`).
- A failed hive finalize backstop still **raises** — the manifest is required reader-facing schema (D6), unlike the fail-open coverage.moc (D9 regenerable cache).

**Fail-fast guard: `mode="ping"` (phase 2, unchanged on the wire)** — one lightweight RequestResponse invoke replaces the PR #205 layout-echo guard, catching two failure modes pre-fan-out:
- *Stale deployment*: a function that predates the issue #252 lifecycle doesn't know the mode, falls through to the process handler's logged 400 with **zero writes**, and the dispatcher raises a redeploy message.
- *Incompatible existing store*: the handler runs the **read-only** `zagg.hive.validate_manifest` against the event's manifest inputs, so a frozen-key mismatch refuses before any leaf is written (D2). Two concurrent runs racing into the same *fresh* root both pass the ping, but with the manifest landing seconds after init the collision window shrinks from run-length to seconds (see Questions (1)).

**Lambda handler (`deployment/aws/lambda_handler.py`) — zero behavioral changes in phase 4.** The hybrid reuses `_handle_setup`'s retained hive branch (already `ensure_manifest(build_manifest(...))`) as the async write target and `_handle_finalize`'s phase-1 hive branch as the backstop, exactly as ratified. Phase 4 only corrected the handler's module/`_handle_setup`/`_handle_finalize`/`_handle_ping` docstrings, which described the phase-1 "manifest rides finalize" lifecycle — flagging this here since the file is otherwise infra-scoped.

**Local dispatcher** — the manifest write is restored to **init** (pre-dispatch `ensure_manifest`; the local path writes the store directly, matching the pre-fold model), while the finalize-time `ensure_manifest` **backstop is kept**. `ensure_manifest` runs the `validate_manifest` frozen-key precheck before its PUT, so the pre-dispatch fail-fast from the review folds is preserved: a mismatched rerun still refuses before any cell runs. Net: local = write at init + idempotent backstop at finalize.

**Benchmark accounting** — `setup_s` on hive rows is the ping plus the ~10 ms Event dispatch; `finalize_s` is now the idempotent manifest backstop (comment updates in `.github/scripts/bench_metrics.py`; lifecycle text updated in `docs/hive_layout.md`, `docs/design/sparse_coverage.md`, `hive.py` docstrings). #250 (open; no phase panel in `plot_series.py` yet) should keep `finalize` in its panel per issue #252 — noted in the column comment.

## Phases

- [x] Phase 1 — fold the manifest write into finalize on both dispatch paths (runner + Lambda handler) + tests (6d25e34)
- [x] Phase 2 — lightweight `mode="ping"` version guard (dispatcher helper + handler branch) + tests (200c6bf)
- [x] Phase 3 — benchmark-accounting comments + docs lifecycle text (136b465)
- [x] Phase 4 — ratified hybrid: async init-time manifest write (Event invoke of the hive setup branch after the ping), finalize demoted to idempotent backstop, local init write restored + backstop kept, docs/bench text updated + tests (83aab6f)

Two adversarial self-review rounds folded on phases 1–3: round 1 (validate_manifest split + local pre-dispatch precheck, 27bccc3; ensure_manifest docstring, 68119c6), round 2 (scope the ping guard claim to sequential reruns, 95fcc8e; branch the ping failure message on the response body, 2d41cbb). All inline threads answered.

## Testing

- Full `uv run pytest -q`: **2170 passed, 30 skipped** (baseline on the branch: 2168 passed, 30 skipped — the delta is phase 4's new tests). `ruff check` / `ruff format --check` on `src tests` green except one **pre-existing on origin/main** item: N818 on `src/zagg/registry.py:64` (`UnknownCapability`). `deployment/aws/invoke_lambda.py` has pre-existing format drift, also on origin/main and outside CI's `src tests` scope. Both untouched — flagged, not fixed, per repo conventions.
- Phase 4 coverage: wire-level pin of the async hive setup event (`InvocationType="Event"`, exact payload, response never read) alongside the existing flat setup/finalize/ping byte-identity pins (all still passing); dispatcher-level lifecycle order (`ping → async setup → workers → finalize`, no sync setup, flat untouched); local-path init write visible before any cell runs; local finalize backstop restores a manifest lost mid-run; mismatched rerun still refuses pre-dispatch; handler-level backstop (manifest absent at finalize still ends written — `TestFinalizeHive`, unchanged).

## Questions for review

1. **Concurrent fresh-root window — narrowed by the hybrid.** Two concurrent runs racing into the same *fresh* root both pass the read-only ping, but the async init write now lands the loser's collision within **seconds of init** instead of at the losing run's finalize (run-length). The residual seconds-wide window keeps the same last-writer caveat the manifest write has always had; closing it entirely would require a synchronous stake-out write, which the ratified hybrid deliberately avoids. Treating this as resolved unless a stricter guarantee is wanted.
2. **Hive + `consolidate_metadata`** never reaches the finalize branch: `validate_config` already rejects that combination (`test_hive_rejects_consolidate_metadata`), so the hive finalize branch writing the manifest and never consolidating is consistent with the config surface.
3. The pre-existing N818 lint error (`src/zagg/registry.py:64`) and `deployment/aws/invoke_lambda.py` format drift noted above — should those get a `small-fix` issue?

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -43,7 +43,9 @@ Event payload (default / process mode):
 Setup mode (creates the zarr template once before per-cell fan-out; for a
 hive-layout config -- output.store_layout: hive, issue #199 -- it writes the
 morton_hive.json manifest instead, and each process-mode worker emits its own
-leaf template):
+leaf template. Current dispatchers no longer send hive setup -- the manifest
+rides finalize instead, issue #252 -- but the hive branch stays so older
+dispatchers keep working against this function):
 {
     "mode": "setup",
     "store_path": str,
@@ -58,10 +60,17 @@ leaf template):
     "output_credentials": dict (optional, same shape as process mode),
 }
 
-Finalize mode (consolidates zarr metadata after all cells complete):
+Finalize mode (after all cells complete: consolidates zarr metadata; for a
+hive-layout config -- issue #252 -- it writes the morton_hive.json manifest
+instead, folded off the pre-worker setup path):
 {
     "mode": "finalize",
     "store_path": str,
+    "config": dict (optional, hive only) -- same single-source config as setup;
+        its presence + store_layout selects the manifest write. With it ride
+        "parent_order", "overwrite", and the optional "dataset" identity
+        block, mirroring the hive setup event. Absent on flat runs (their
+        event is byte-identical to pre-#252).
     "output_credentials": dict (optional, same shape as process mode),
 }
 
@@ -672,11 +681,42 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
 
 
 def _handle_finalize(event: Dict[str, Any]) -> Dict[str, Any]:
-    """Consolidate zarr metadata for the store at ``event['store_path']``."""
-    from zarr import consolidate_metadata
+    """Finalize the store at ``event['store_path']``.
 
-    logger.info(f"Finalize mode: consolidating metadata at {event.get('store_path')}")
+    Flat: consolidate zarr metadata (byte-identical to before). Hive (issue
+    #252): write the root ``morton_hive.json`` manifest instead — folded off
+    the setup path so no pre-worker invoke blocks the fan-out. Order-correct
+    because the manifest is reader-facing only: workers write self-describing
+    leaves (D3) and readers arrive after the run. The manifest inputs mirror
+    the hive setup event (``config``/``parent_order``/``dataset``/
+    ``overwrite``); there is no zarr hierarchy above the leaves to
+    consolidate (D5), so the hive branch returns without touching zarr. The
+    success body echoes ``"layout": "hive"``, matching setup's echo.
+    """
+    logger.info(f"Finalize mode: finalizing store at {event.get('store_path')}")
     try:
+        if "config" in event:
+            config = load_config_from_dict(event["config"])
+            if get_store_layout(config) == "hive":
+                from zagg.config import get_windowing
+                from zagg.grids import from_config
+                from zagg.hive import build_manifest, ensure_manifest
+
+                grid = from_config(config, parent_order=event.get("parent_order"))
+                ensure_manifest(
+                    event["store_path"],
+                    build_manifest(
+                        grid, dataset=event.get("dataset"), windowing=get_windowing(config)
+                    ),
+                    overwrite=event.get("overwrite", False),
+                    **_output_store_kwargs(event),
+                )
+                return {
+                    "statusCode": 200,
+                    "body": json.dumps({"ok": True, "mode": "finalize", "layout": "hive"}),
+                }
+        from zarr import consolidate_metadata
+
         store = open_store(event["store_path"], **_output_store_kwargs(event))
         consolidate_metadata(store, zarr_format=3)
         return {"statusCode": 200, "body": json.dumps({"ok": True, "mode": "finalize"})}
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -764,7 +764,9 @@ def _handle_ping(event: Dict[str, Any]) -> Dict[str, Any]:
     into a store templated for different orders/identity refuses up front
     (the D2 mixed-order footgun) instead of after the fan-out — the writer-
     side guard the manifest fold would otherwise defer to finalize (PR #255
-    review fold). Kept while flat exists (issue #251): once flat is removed,
+    review fold). This covers sequential reruns; two concurrent runs into the
+    same fresh root both see no manifest and only collide at the losing run's
+    finalize. Kept while flat exists (issue #251): once flat is removed,
     a stale function simply errors and the ping can be dropped.
     """
     logger.info(f"Ping mode: hive preflight for {event.get('store_path')}")
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -642,9 +642,12 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
     For a hive-layout config (issue #199 phase 3) template time writes ONLY
     the ``morton_hive.json`` manifest — no global zarr template exists (zero
     metadata above the leaves, D5); each worker emits its own leaf template.
-    The optional ``dataset`` event key carries the manifest's identity block
-    (the orchestrator sources it from the ShardMap metadata, same as the local
-    path). The flat path below is byte-identical to before, bar one addition:
+    Current dispatchers no longer send hive setup (the manifest rides finalize
+    since issue #252); this branch stays so OLDER dispatchers keep working
+    against a redeployed function. The optional ``dataset`` event key carries
+    the manifest's identity block (the orchestrator sources it from the
+    ShardMap metadata, same as the local path).
+    The flat path below is byte-identical to before, bar one addition:
     the success body now ECHOES the layout it acted on (``"layout"``) — a
     stale deployment without the hive branch returns the old echo-less body,
     which the dispatcher rejects for hive runs instead of silently letting old
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -74,6 +74,25 @@ instead, folded off the pre-worker setup path):
     "output_credentials": dict (optional, same shape as process mode),
 }
 
+Ping mode (hive pre-fan-out preflight, issue #252 — writes nothing; kept while
+flat exists, issue #251):
+{
+    "mode": "ping",
+    "store_path": str,
+    "config": dict,             # same manifest inputs as hive finalize; with it
+    "parent_order": int,        #   ride "overwrite" and the optional "dataset"
+    "overwrite": bool,          #   identity block
+    "dataset": dict (optional),
+    "output_credentials": dict (optional, same shape as process mode),
+}
+Answering 200 at all is the versioning half of the guard: a function deployed
+before the issue #252 manifest-in-finalize lifecycle doesn't know the mode, so
+the event falls through to the process handler's 400 (zero writes) and the
+dispatcher fails fast with a redeploy message before any worker runs. The body
+echoes the deployed zagg version; the handler also runs the READ-ONLY
+zagg.hive.validate_manifest against any existing root so an incompatible rerun
+refuses up front (D2).
+
 Extract mode (chunk-boundary geometry extraction, issue #148 — one parquet per
 granule under an S3 prefix; a batch of granules per invocation for the fan-out):
 {
@@ -453,6 +472,7 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
 
     Default ``mode`` (or no mode) runs per-cell processing. ``mode="setup"``
     creates the zarr template; ``mode="finalize"`` consolidates metadata;
+    ``mode="ping"`` is the hive pre-fan-out preflight (issue #252);
     ``mode="coverage"`` writes the store-root ``coverage.moc`` (issue #200);
     ``mode="extract"`` extracts chunk-boundary geometry parquets (issue #148);
     ``mode="process_event"`` runs the temporal/event worker (issue #12).
@@ -468,6 +488,8 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         return _handle_setup(event)
     if mode == "finalize":
         return _handle_finalize(event)
+    if mode == "ping":
+        return _handle_ping(event)
     if mode == "coverage":
         return _handle_coverage(event)
     # Extract mode returns directly: the result_url mirror below is for the
@@ -723,6 +745,52 @@ def _handle_finalize(event: Dict[str, Any]) -> Dict[str, Any]:
     except Exception as e:
         logger.exception(e)
         return {"statusCode": 500, "body": json.dumps({"error": str(e), "mode": "finalize"})}
+
+
+def _handle_ping(event: Dict[str, Any]) -> Dict[str, Any]:
+    """Hive pre-fan-out preflight (issue #252) — writes nothing.
+
+    Answering 200 at all is the versioning half of the guard: a function that
+    predates the manifest-in-finalize lifecycle doesn't know ``mode="ping"``,
+    so the event falls through to its process handler's 400 (zero writes) and
+    the dispatcher fails fast with a redeploy message before any worker is
+    dispatched. The body echoes the deployed zagg version for observability.
+
+    The store half mirrors finalize's manifest inputs and runs the READ-ONLY
+    :func:`zagg.hive.validate_manifest` against any existing root, so a run
+    into a store templated for different orders/identity refuses up front
+    (the D2 mixed-order footgun) instead of after the fan-out — the writer-
+    side guard the manifest fold would otherwise defer to finalize (PR #255
+    review fold). Kept while flat exists (issue #251): once flat is removed,
+    a stale function simply errors and the ping can be dropped.
+    """
+    logger.info(f"Ping mode: hive preflight for {event.get('store_path')}")
+    try:
+        import zagg
+
+        if "config" in event:
+            config = load_config_from_dict(event["config"])
+            if get_store_layout(config) == "hive":
+                from zagg.config import get_windowing
+                from zagg.grids import from_config
+                from zagg.hive import build_manifest, validate_manifest
+
+                grid = from_config(config, parent_order=event.get("parent_order"))
+                validate_manifest(
+                    event["store_path"],
+                    build_manifest(
+                        grid, dataset=event.get("dataset"), windowing=get_windowing(config)
+                    ),
+                    overwrite=event.get("overwrite", False),
+                    **_output_store_kwargs(event),
+                )
+        return {
+            "statusCode": 200,
+            "body": json.dumps({"ok": True, "mode": "ping", "zagg_version": zagg.__version__}),
+        }
+    except Exception as e:
+        logger.exception(e)
+        return {"statusCode": 500, "body": json.dumps({"error": str(e), "mode": "ping"})}
 
 
 def _handle_coverage(event: Dict[str, Any]) -> Dict[str, Any]:
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -43,9 +43,9 @@ Event payload (default / process mode):
 Setup mode (creates the zarr template once before per-cell fan-out; for a
 hive-layout config -- output.store_layout: hive, issue #199 -- it writes the
 morton_hive.json manifest instead, and each process-mode worker emits its own
-leaf template. Current dispatchers no longer send hive setup -- the manifest
-rides finalize instead, issue #252 -- but the hive branch stays so older
-dispatchers keep working against this function):
+leaf template. Current dispatchers send hive setup as a fire-and-forget Event
+invoke right after the ping -- the primary manifest write, issue #252 hybrid
+-- and older synchronous dispatchers keep working against this function):
 {
     "mode": "setup",
     "store_path": str,
@@ -61,8 +61,8 @@ dispatchers keep working against this function):
 }
 
 Finalize mode (after all cells complete: consolidates zarr metadata; for a
-hive-layout config -- issue #252 -- it writes the morton_hive.json manifest
-instead, folded off the pre-worker setup path):
+hive-layout config -- issue #252 hybrid -- it re-ensures the morton_hive.json
+manifest instead, the idempotent backstop for the async init-time setup write):
 {
     "mode": "finalize",
     "store_path": str,
@@ -86,7 +86,7 @@ flat exists, issue #251):
     "output_credentials": dict (optional, same shape as process mode),
 }
 Answering 200 at all is the versioning half of the guard: a function deployed
-before the issue #252 manifest-in-finalize lifecycle doesn't know the mode, so
+before the issue #252 hive dispatch lifecycle doesn't know the mode, so
 the event falls through to the process handler's 400 (zero writes) and the
 dispatcher fails fast with a redeploy message before any worker runs. The body
 echoes the deployed zagg version; the handler also runs the READ-ONLY
@@ -642,9 +642,11 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
     For a hive-layout config (issue #199 phase 3) template time writes ONLY
     the ``morton_hive.json`` manifest — no global zarr template exists (zero
     metadata above the leaves, D5); each worker emits its own leaf template.
-    Current dispatchers no longer send hive setup (the manifest rides finalize
-    since issue #252); this branch stays so OLDER dispatchers keep working
-    against a redeployed function. The optional ``dataset`` event key carries
+    Current dispatchers send hive setup as a fire-and-forget Event invoke
+    right after the ping (issue #252 hybrid) — this branch is the PRIMARY
+    manifest write, run off the critical path, with finalize as idempotent
+    backstop; older synchronous dispatchers keep working against it
+    unchanged. The optional ``dataset`` event key carries
     the manifest's identity block (the orchestrator sources it from the
     ShardMap metadata, same as the local path).
     The flat path below is byte-identical to before, bar one addition:
@@ -709,14 +711,15 @@ def _handle_finalize(event: Dict[str, Any]) -> Dict[str, Any]:
     """Finalize the store at ``event['store_path']``.
 
     Flat: consolidate zarr metadata (byte-identical to before). Hive (issue
-    #252): write the root ``morton_hive.json`` manifest instead — folded off
-    the setup path so no pre-worker invoke blocks the fan-out. Order-correct
-    because the manifest is reader-facing only: workers write self-describing
-    leaves (D3) and readers arrive after the run. The manifest inputs mirror
-    the hive setup event (``config``/``parent_order``/``dataset``/
-    ``overwrite``); there is no zarr hierarchy above the leaves to
-    consolidate (D5), so the hive branch returns without touching zarr. The
-    success body echoes ``"layout": "hive"``, matching setup's echo.
+    #252 hybrid): re-ensure the root ``morton_hive.json`` manifest — the
+    idempotent BACKSTOP for the async init-time setup invoke (a frozen-key-
+    matching manifest is accepted, no second PUT). Worker Event invokes run
+    with retries 0 (template.yaml EventInvokeConfig), so a lost async init
+    write is never redelivered — this backstop self-heals it. The manifest
+    inputs mirror the hive setup event (``config``/``parent_order``/
+    ``dataset``/``overwrite``); there is no zarr hierarchy above the leaves
+    to consolidate (D5), so the hive branch returns without touching zarr.
+    The success body echoes ``"layout": "hive"``, matching setup's echo.
     """
     logger.info(f"Finalize mode: finalizing store at {event.get('store_path')}")
     try:
@@ -754,20 +757,20 @@ def _handle_ping(event: Dict[str, Any]) -> Dict[str, Any]:
     """Hive pre-fan-out preflight (issue #252) — writes nothing.
 
     Answering 200 at all is the versioning half of the guard: a function that
-    predates the manifest-in-finalize lifecycle doesn't know ``mode="ping"``,
+    predates the issue #252 hive lifecycle doesn't know ``mode="ping"``,
     so the event falls through to its process handler's 400 (zero writes) and
     the dispatcher fails fast with a redeploy message before any worker is
     dispatched. The body echoes the deployed zagg version for observability.
 
-    The store half mirrors finalize's manifest inputs and runs the READ-ONLY
-    :func:`zagg.hive.validate_manifest` against any existing root, so a run
-    into a store templated for different orders/identity refuses up front
-    (the D2 mixed-order footgun) instead of after the fan-out — the writer-
-    side guard the manifest fold would otherwise defer to finalize (PR #255
-    review fold). This covers sequential reruns; two concurrent runs into the
-    same fresh root both see no manifest and only collide at the losing run's
-    finalize. Kept while flat exists (issue #251): once flat is removed,
-    a stale function simply errors and the ping can be dropped.
+    The store half mirrors setup/finalize's manifest inputs and runs the
+    READ-ONLY :func:`zagg.hive.validate_manifest` against any existing root,
+    so a run into a store templated for different orders/identity refuses up
+    front (the D2 mixed-order footgun) instead of after the fan-out (PR #255
+    review fold). This covers sequential reruns; two concurrent runs into
+    the same fresh root both pass here, colliding within seconds of init
+    once the async setup write lands (issue #252 hybrid). Kept while flat
+    exists (issue #251): once flat is removed, a stale function simply
+    errors and the ping can be dropped.
     """
     logger.info(f"Ping mode: hive preflight for {event.get('store_path')}")
     try:
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -3057,12 +3057,23 @@ def _invoke_lambda_ping(
         raise RuntimeError(f"Lambda ping failed: {payload}")
     result = json.loads(payload)
     if result.get("statusCode") != 200:
+        try:
+            body = json.loads(result.get("body") or "{}")
+        except (TypeError, ValueError):
+            body = {}
+        if body.get("mode") == "ping":
+            # The deployed function KNOWS ping — this is validate_manifest
+            # refusing the store, not a stale deployment.
+            raise RuntimeError(
+                f"Lambda ping refused the store at {store_path}: "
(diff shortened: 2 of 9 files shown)


src/zagg/runner.py
```diff
@@ -1820,6 +1818,18 @@ def _accumulate(report, i, outcome):
...
+    # root coverage.moc below (a regenerable cache, D9), the manifest is
+    # REQUIRED reader-facing schema (D6), so a failed write — including a
+    # frozen-key mismatch against an existing store's manifest — raises.
+    if store_layout == "hive":
+        ensure_manifest(
+            store_path,
+            build_manifest(grid, dataset=catalog_data.get("metadata"), windowing=windowing),
+            overwrite=overwrite,
+            **store_kwargs,
+        )
```
espg, line 1846: 🤖 *from Claude (review)*

Folding `ensure_manifest` to the end of the run silently drops a **pre-fan-out safety gate**, and the issue's "why it's safe" analysis doesn't cover it. Before this PR, `ensure_manifest` ran at template time — *before* any worker wrote a leaf. `ensure_manifest` isn't just a manifest write; it's the writer-side guard (`hive.py:205-250`): on an existing store with different frozen keys it raises, and the `overwrite=True` branch does a `list_with_delimiter` and refuses when the digit tree already has shard children (the D2 "old-order leaves masquerading as legal mixed-order data" footgun).

The issue only argues *reader* ordering ("readers arrive after the run"). But this guard is about *writers*. Hive workers write self-describing leaves and **never read the manifest** (D3), so with the guard moved to finalize, a rerun/misdirected run against an incompatible existing store now:

1. **Loses fail-fast** — a whole multi-hour CONUS fan-out burns before the `ValueError` fires, where it used to fail in ~0s up front.
2. **Opens a real corruption window** — the fan-out writes *new-order* leaves into the *old-order* store during the run; `ensure_manifest` only raises afterward, i.e. after the exact D2 mixing the `overwrite` guard was designed to prevent has already happened on disk. The frozen-key check firing at the end can no longer prevent it, only report it.

This is a genuine behavioral tradeoff, not obviously a bug — but it's not one the issue signed off on, so flagging for a decision. Options: (a) keep a cheap up-front *validate-only* pass (read the existing manifest + frozen-key/overwrite check, no write) before the fan-out, preserving both fail-fast and the corruption guard while still folding the *write* into finalize; (b) accept the tradeoff but document it in the issue/PR and add a test pinning the mismatched-rerun behavior. Note the comment here at L1821-1825 already discusses the "raises on frozen-key mismatch" semantics but omits that it now raises *post*-fan-out — worth making explicit either way. The same regression applies to the lambda path (see the finalize handler comment).

---

deployment/aws/lambda_handler.py
```diff
@@ -672,11 +681,42 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
...
+                from zagg.grids import from_config
+                from zagg.hive import build_manifest, ensure_manifest
+
+                grid = from_config(config, parent_order=event.get("parent_order"))
+                ensure_manifest(
+                    event["store_path"],
+                    build_manifest(
+                        grid, dataset=event.get("dataset"), windowing=get_windowing(config)
+                    ),
+                    overwrite=event.get("overwrite", False),
```
espg, line 739: 🤖 *from Claude (review)*

Two things on this hive finalize branch:

1. **Same guard-timing regression as the local path** (see my comment on `runner.py`'s end-of-run `ensure_manifest`). On the lambda backend the fan-out has *already* invoked every per-cell worker by the time `_handle_finalize` runs, so the frozen-key / `overwrite`-with-existing-shard-data guard inside `ensure_manifest` now fires only after new-order leaves have been written into an incompatible store. `test_finalize_frozen_key_mismatch_is_500` correctly pins that the handler still *returns* 500 on mismatch — good — but nothing pins that this is now too late to prevent the D2 mixing the guard existed to stop. Worth an explicit decision (a cheap up-front validate-only ping would restore fail-fast; note the phase-2 "version ping" in the PR plan is about stale *deployments*, not frozen-key mismatch, so it won't recover this on its own).

2. **Stale "template time" wording left behind.** This move makes `hive.ensure_manifest`'s own docstring wrong — it still opens with *"Write the root manifest once at template time; verify it on reruns."* (`src/zagg/hive.py:206`), which the PR didn't touch. A reader landing there will believe the manifest is still a pre-fan-out write. Since the fold is the whole point of #252, please update that docstring (and any sibling comment that says "at template time") to say the write now rides finalize. Minor, but it's exactly the kind of misleading breadcrumb the move leaves.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase 1 (commit 6d25e34) — the manifest-write fold. The mechanical move is clean and well-tested: local drops the pre-fan-out write, lambda drops the setup invoke and threads the manifest inputs onto the finalize event, flat setup/finalize events are held byte-identical and pinned on the wire, and the handler grows a hive branch with frozen-key/resume/windowed coverage. Dispatch wiring checks out — `_finalize_fn` is defined for all three cases, `parent_order`/`overwrite`/`config_dict`/`dataset` are all in scope, and hive finalize correctly runs unconditionally (independent of the `consolidate_metadata` gate). The separate temporal `_run_lambda_events` path is untouched and never wrote a manifest, so no gap there.

One substantive finding, on both paths (inline on `runner.py` and `lambda_handler.py`): folding `ensure_manifest` to finalize doesn't just move a *write*, it moves a **pre-fan-out writer-side guard** (the frozen-key refuse + the `overwrite`-with-existing-shard-data D2 corruption check in `hive.ensure_manifest`). Since hive workers never read the manifest (D3), a rerun against an incompatible existing store now writes new-order leaves throughout the entire fan-out before the guard fires — losing fail-fast and opening the exact D2 mixed-order window the guard was built to prevent. The issue's safety argument only covers reader ordering, not this. It's a real tradeoff worth an explicit decision (a cheap up-front *validate-only* pass would preserve both while still folding the write); the phase-2 version ping won't recover it, as that targets stale deployments, not config mismatch.

Plus one doc nit folded into the second inline: `hive.ensure_manifest`'s docstring still says "Write the root manifest once at template time" — stale after this fold and left untouched by the PR.

Not flagged, per the PR plan: the fail-fast deploy ping (phase 2) and benchmark/#250 phase-panel accounting (phase 3).

---

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded on the local path via option (a): kept the WRITE at finalize, but split the read-only frozen-key guard out as `hive.validate_manifest(...)` (directly above `ensure_manifest`) and call it pre-dispatch, so a rerun/misdirected run against an incompatible existing store now refuses in ~0s before any leaf is written — restoring both fail-fast and the D2 corruption guard.

- `hive.py`: new `validate_manifest(store_root, manifest, *, overwrite=False, **store_kwargs) -> dict | None` performs exactly the existing checks (same `does not match this run` and `clear the store root first`/`list_with_delimiter` messages), writes nothing, returns the existing manifest (`None` on a fresh root); `ensure_manifest` now calls it and keeps only the write/return logic (behavior byte-identical).
- `runner.py` (local hive branch): manifest is built once up front and `validate_manifest(store_path, manifest, overwrite=overwrite, **store_kwargs)` runs pre-dispatch; finalize's `ensure_manifest` reuses that same prebuilt `manifest`.
- Tests (`tests/test_hive.py`): a LOCAL rerun into a mismatched-frozen-key root raises before any cell runs (asserts the monkeypatched `process_and_write_hive` was never called), plus direct `validate_manifest` tests (fresh root → `None`, writes nothing; match → existing; mismatch → raises).

Fix commit: 27bccc3099c9f8f8b4b0bb5f1dac4c577eccb128.

The lambda-path pre-fan-out validation lands separately in the phase-2 ping commit (see reply on the finalize-handler thread).

---

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

(2) Stale "template time" wording: fixed. `hive.ensure_manifest`'s docstring now opens with `Write the root manifest at finalize / end of run; verify it on reruns.` and explains the write rides finalize since issue #252, with the fail-fast half exposed as `validate_manifest` for pre-fan-out refusal (D2). Docstring fix commit: 68119c63d3f76743599169785e3b26547651e9a3.

(1) Same guard-timing regression on the lambda path: the pre-fan-out frozen-key validation rides the phase-2 ping commit (the one up-front invoke, issue #252, that this PR's plan ratifies) — the ping event will carry the manifest inputs and `_handle_finalize`'s counterpart will run `validate_manifest` read-only before fan-out. That commit will be linked on this thread when it lands. Note the reusable read-only guard is already in place from the local-path fold (`hive.validate_manifest`, commit 27bccc3099c9f8f8b4b0bb5f1dac4c577eccb128), so the ping handler just calls it.

---

espg: 🤖 *from Claude*

Follow-up as promised: the lambda-path pre-fan-out validation landed in phase 2, commit 200c6bf. The `mode="ping"` preflight event carries the same manifest inputs as hive finalize, and the handler runs the read-only `zagg.hive.validate_manifest` against any existing root — so a frozen-key mismatch (or an overwrite into a root with shard data) now refuses BEFORE the fan-out on the lambda path too, mirroring the local path's pre-dispatch check from 27bccc3. Pinned by `TestPingMode::test_ping_frozen_key_mismatch_is_500` (handler side) and `TestInvokeLambdaPingEvent` (wire side; a stale function's 400 fall-through and the validate refusal both abort pre-dispatch).

---

src/zagg/runner.py
```diff
@@ -2952,24 +2997,104 @@ def _invoke_lambda_setup(
...
+      handler's 400 with ZERO writes — strictly earlier and cheaper than the
+      retired PR #205 layout-echo guard, which only fired after a
+      pre-#199-phase-3 function had already written the flat GLOBAL template
+      at the hive root. This also gates out hive-capable-but-pre-fold
+      functions, whose finalize would not write the manifest.
+    - **Incompatible existing store:** the handler runs the read-only
+      ``zagg.hive.validate_manifest`` against the event's manifest inputs
+      (same keys as hive finalize), so a frozen-key mismatch — or an
+      overwrite into a root with shard data — refuses up front (D2) instead
+      of after the fan-out (PR #255 review fold).
```

Judgment call (low severity, docstring accuracy): "refuses up front (D2) instead of after the fan-out" is true for the case that matters most — a **rerun into an existing store** — but the ping's read-only `validate_manifest` narrows the D2 window for **concurrent runs into the same fresh root**, versus the old setup-writes-manifest-pre-fan-out it replaces.

Old path: `_invoke_lambda_setup` wrote `morton_hive.json` before fan-out, so a second run with different orders read the just-written manifest at *its* setup and raised **before writing any leaf**. New path: run A's ping sees no manifest (fresh root) → passes → fans out; run B's ping, launched before A's finalize, also sees no manifest → passes → fans out mixed-order leaves; the mismatch only surfaces at B's finalize `ensure_manifest`, i.e. *after* the fan-out already contaminated the tree. The guard window widened from "between the two setups" to "the whole run."

Why it matters: the sequential-rerun guard (the common footgun) is fully preserved, so this is a real but narrow regression, not a blocker. Suggested fix: soften the claim here (and in the matching `_handle_ping` / `validate_manifest` docstrings and `docs/hive_layout.md`) to note the pre-fan-out refusal covers reruns into an already-templated store, while two concurrent runs into a fresh root are only caught at the losing run's finalize. If closing that window matters, it's a separate follow-up (e.g. an early manifest stake-out), not this PR.

---

src/zagg/runner.py
```diff
@@ -2952,24 +2997,104 @@ def _invoke_lambda_setup(
...
+    )
+    payload = response["Payload"].read().decode("utf-8")
+    if response.get("FunctionError"):
+        raise RuntimeError(f"Lambda ping failed: {payload}")
+    result = json.loads(payload)
+    if result.get("statusCode") != 200:
+        raise RuntimeError(
+            f"Lambda ping failed (response body {result.get('body')!r}): either the "
+            f"deployed function predates the issue #252 manifest-in-finalize "
+            f"lifecycle — redeploy the function before dispatching hive runs — or "
```

Judgment call (low severity, error-message accuracy): this single message leads every non-200 with "the deployed function predates the issue #252 manifest-in-finalize lifecycle — redeploy the function" as the *primary* explanation, but the two failure modes are distinguishable from the response and want opposite operator actions:

- **Stale deployment**: statusCode 400, body is the process handler's key-less rejection (no `"mode": "ping"`). Redeploy is correct.
- **Incompatible store**: statusCode 500, body echoes `"mode": "ping"` and carries `validate_manifest`'s "does not match this run" refusal. The function is fine; redeploying does nothing — the operator must clear/repoint the store root.

For the 500 case, "redeploy the function" is an actively wrong first suggestion. Since `_handle_ping`'s 500 body already carries `mode: "ping"` (and the validate refusal text), you can branch the message on it, e.g. detect whether the parsed body has `mode == "ping"` and either say "the store at {store_path} was templated for a different configuration — clear the root (see error above)" or the redeploy message, rather than concatenating both. Low-value polish, but the current wording can send someone down the wrong path on the more likely of the two (a config/orders mismatch on rerun).

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phases 2–3 (commits 200c6bf, 136b465) against issue #252 and CLAUDE.md. No correctness defect survived scrutiny — the ping meets the issue's bar:

- **Stale-deployment fail-fast holds on every shape.** A pre-`mode="ping"` function routes the key-less event through `_handle_process` → 400 with zero writes; `_invoke_lambda_ping` turns any non-200 (or `FunctionError`) into a raise before the fan-out. Because ping and the finalize-writes-manifest branch ship in the same zip, a function that knows ping also writes the manifest — so the ping is a valid proxy that also gates out hive-capable-but-pre-fold functions.
- **Nothing dispatches hive setup.** `_run_lambda` sends ping for hive, setup only for flat; `_handle_setup`'s hive branch stays solely for older dispatchers.
- **Flat path byte-identical** — only the hive branch swapped setup→ping; the flat `_invoke_lambda_setup` call and event are untouched, and the wiring test asserts `"ping" not in captured` on flat.
- **Telemetry/recycle consistent** — ping flows through `lambda_handler`, so `_container_telemetry()` counts its generation like setup/finalize, and with no `result_url` it never bills `_ASYNC_INVOCATIONS_SERVED` against the recycle budget. Output-credentials and windowed-config manifest inputs mirror finalize exactly.
- **Phase-3 accounting is accurate** — bench `finalize_s`/`setup_s` notes, `hive.py`/docs lifecycle text, and the setup/finalize docstrings all match the new lifecycle. Tests (`TestPingMode`, `TestInvokeLambdaPingEvent`, wiring updates, coverage stub) cover the behavioral surface.

Two low-severity judgment calls posted inline (docstring overstates the D2 guard for the concurrent-fresh-root case; the ping's non-200 error message leads with "redeploy" even when the failure is a store-config mismatch). Neither blocks — fold or leave by design.

---

src/zagg/runner.py

Softened the claim everywhere the "refuses up front (D2)" wording appears, scoping it to sequential reruns and naming the concurrent-fresh-root window explicitly:

- `src/zagg/runner.py` `_invoke_lambda_ping` "Incompatible existing store" bullet — now: "...refuses up front (D2) instead of after the fan-out (PR #255 review fold). This covers sequential reruns; two CONCURRENT runs racing into the same fresh root both see no manifest and only collide at the losing run's finalize — a window the old setup-time write closed for the second-to-arrive dispatch. Same last-writer caveat the manifest write itself has always had."
- `deployment/aws/lambda_handler.py` `_handle_ping` docstring — one-sentence scoping (sequential reruns vs concurrent fresh-root collide-at-finalize).
- `docs/hive_layout.md` ("## The manifest") and `docs/design/sparse_coverage.md` §3 opening — a few-words scoping on each ("refuses up front on reruns; concurrent first writes still collide only at finalize").

Fix commit: 95fcc8ebef31e8d2fc4e30a2bd2dc25943b87cbd.

Closing the concurrent-fresh-root window itself (e.g. an early manifest stake-out at ping time) is left as a possible follow-up for the maintainer to decide — it predates this PR's scope, and it's the same last-writer semantics the old setup-time write had for whichever dispatch arrived second.

---

src/zagg/runner.py

Branched the non-200 message on the parsed body so the two failure modes get opposite remedies instead of one concatenated string:

- `src/zagg/runner.py` `_invoke_lambda_ping` — parse `result["body"]`; if it echoes `mode == "ping"` (the handler KNOWS ping, so this is `validate_manifest` refusing the store) raise "Lambda ping refused the store at {store_path}: {error!r} — the store was templated for a different configuration; clear the store root (or pick a new one)...". Otherwise (400 fall-through, no `mode` key) keep the "predates the issue #252 manifest-in-finalize lifecycle — redeploy the function" message. Malformed/missing body falls back to the redeploy branch.
- `tests/test_hive.py` `TestInvokeLambdaPingEvent` — `test_stale_function_fall_through_fails_fast` keeps `match="redeploy"` (its 400 body has no `mode: "ping"`); `test_validate_refusal_fails_fast` now asserts `match="clear the store root"` (its 500 body already carries `mode: "ping"`), with the comment updated to note the message points at the store remedy, not redeploy. Class docstring updated: the two modes get distinct remedies.

Fix commit: 2d41cbb665a6fa0e7be0304ff7fe2535d73ddfeb.

---

espg:
🤖 *from Claude*

Run status: **all three phases complete** and pushed (6d25e34 → 2d41cbb), per the [plan on the issue](https://github.com/englacial/zagg/issues/252#issuecomment-4990702965).

- Two adversarial self-review rounds ran; **all 4 findings folded** (27bccc3, 68119c6, 95fcc8e, 2d41cbb) and every inline thread answered with its fix sha. One item deliberately left standing as a follow-up decision: closing the concurrent-fresh-root ping window (see [this thread](https://github.com/englacial/zagg/pull/255#discussion_r3594810204) and "Questions for review" (1) in the body) — restoring it would reintroduce a pre-worker write, which is what this issue removes.
- CI at time of writing: ruff ✅ (no inline bot comments), lambda builds ✅, test matrix in progress; the full suite is green locally (2168 passed, 30 skipped vs 2151/30 on main).
- Pre-existing on main, flagged not fixed: N818 at `src/zagg/registry.py:64`, format drift in `deployment/aws/invoke_lambda.py` ("Questions for review" (3)).

**Draft state**: this routine's configuration mandates drafts-only, so the PR stays draft even though CLAUDE.md §2 would have it flipped ready (all phases done, findings folded, CI otherwise green) — leaving the draft→ready flip and merge to the maintainer. Labeling `implement` + `waiting` accordingly.

---

tests/test_hive.py
```diff
@@ -916,19 +940,76 @@ def test_local_hive_writes_manifest_not_template(self, monkeypatch, cfg, tmp_pat
...
+        from zagg.runner import agg
+
+        cfg.output["store_layout"] = "hive"
+        catalog_path, shard = self._catalog(tmp_path)
+        root = str(tmp_path / "out")
+
+        monkeypatch.setattr(runner, "get_nsidc_s3_credentials", lambda: {"accessKeyId": "a"})
+
+        def fake_hive_write(shard_key, granule_urls, grid, s3_creds, store_root, config, **kw):
+            os.remove(os.path.join(store_root, hive.MANIFEST_NAME))
```
espg: 🤖 *from Claude (review)*

**[test-gap]** This backstop test does not actually isolate the backstop behavior it advertises (init-write → lost mid-run → restored by finalize). `fake_hive_write` calls `os.remove(...)` inside the cell, but `_cell_work` in `_run_local` wraps the cell body in `try/except Exception` and swallows failures into an error envelope. So if the pre-dispatch init write ever regressed (manifest never written at init), `os.remove` would raise `FileNotFoundError`, get swallowed as a cell error, and finalize would write the manifest anyway — the test stays **green**. It therefore pins only "finalize writes the manifest," not "a manifest that *existed* was lost and *restored*."

To make it a real pin of the hybrid init→backstop path, assert the manifest exists before removing it (e.g. inside the fake, `assert os.path.exists(path)` before `os.remove`), so a missing init-write fails the test loudly instead of being masked by the swallow.

---

tests/test_hive.py
```diff
@@ -1156,21 +1270,146 @@ def test_flat_event_omits_dataset_and_matches_baseline(self, cfg):
...
+        config_dict = asdict(cfg)
+        client = _wire_client({"ok": True, "mode": "setup", "layout": "hive"})
+        _invoke_lambda_setup_async(
+            client,
+            "process-shard",
+            "s3://out/product",
+            config_dict=config_dict,
+            dataset={"short_name": "ATL06", "version": "007"},
+            parent_order=6,
+            overwrite=False,
```
espg, line 1305: 🤖 *from Claude (review)*

**[test-gap]** `_invoke_lambda_setup_async` has two conditional payload branches — `if dataset is not None` and `if output_creds_event is not None` — plus the `overwrite` field, but this wire test only exercises `dataset` (with `overwrite=False`). The `output_credentials` threading is untested on the async path, and the wiring test (`test_lambda_hive_fires_async_setup_after_ping`) monkeypatches the function out, so nothing pins that creds reach the wire.

This matters more here than for ping/finalize: the async setup is the **primary** manifest write and is fire-and-forget (`InvocationType="Event"`, no response read). If the `output_credentials` key ever drifted, the setup-mode `ensure_manifest` S3 write would fail *silently* — no dispatcher-visible error — and the miss would only surface as the finalize backstop doing the "primary" write, quietly erasing the progressive-read benefit the PR is built around. Worth one assertion that `output_credentials` (and ideally `overwrite=True`) threads into the Event payload, mirroring the finalize/ping wire pins.

---

src/zagg/runner.py
```diff
@@ -2952,24 +3011,169 @@ def _invoke_lambda_setup(
...
+    overwrite=False,
+    output_creds_event=None,
+):
+    """Fire-and-forget hive manifest write at init (issue #252 hybrid).
+
+    One ``InvocationType="Event"`` invoke of the existing ``mode="setup"``
+    hive branch (~10 ms of dispatcher wall, the root-coverage dispatch
+    precedent), posted immediately after the ping passes: the handler runs
+    ``ensure_manifest(build_manifest(...))`` in parallel with the worker
+    fan-out, so ``morton_hive.json`` lands seconds into the run and a reader
```

**[docs/nit]** "so `morton_hive.json` lands seconds into the run" (repeated in the `_run_lambda` comment at ~L2185 and across `docs/hive_layout.md` / `docs/design/sparse_coverage.md`) overstates a **best-effort** timing as a guarantee. The async setup is an `Event` invoke of the *same* function the worker fan-out saturates, so it competes for that function's concurrency: under a reserved-concurrency ceiling it can be throttled and queued behind the fan-out and land much later than "seconds" — potentially not until finalize. And on a function error (e.g. the concurrent-overwrite frozen-key raise this same commit describes), retries-0 drops it entirely, deferring the whole write to the finalize backstop.

The self-heal design is sound — this is purely a claims-vs-reality wording point. The progressive-read window is *opportunistic*, not reliable; consider softening the docs to "typically lands early in the run (best-effort; deferred to the finalize backstop under async throttling or a dropped invoke)" so a future reader doesn't treat mid-run manifest availability as a contract.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase-4 commit 83aab6f (issue #252 HYBRID lifecycle). The core design is sound and correctly implemented:

- **Event-shape symmetry verified.** `_invoke_lambda_setup_async` emits the exact retired synchronous hive-setup event shape (`mode="setup"` + `store_path`/`parent_order`/`overwrite`/`config`, optional `dataset`/`output_credentials`), which is precisely what `_handle_setup`'s hive branch reads. Zero new handler keys — the "zero handler behavior changes" claim holds.
- **Handler diff is docstring-only, confirmed.** All `deployment/aws/lambda_handler.py` hunks fall inside the module/`_handle_setup`/`_handle_finalize`/`_handle_ping` docstrings; no executable line changed.
- **Backstop is genuinely load-bearing.** `_invoke_lambda_finalize` is `RequestResponse` and raises on non-200, so a failed backstop surfaces to the dispatcher (unlike the fire-and-forget async setup). `ensure_manifest`'s frozen-key idempotency (no second PUT on a match) makes the finalize re-ensure safe. Local path mirrors this: `ensure_manifest` pre-dispatch (running `validate_manifest` first for the D2 fail-fast) + finalize re-ensure.
- **Local↔lambda symmetry holds** and the pre-dispatch fail-fast (review-fold guard) is preserved on both legs.

Three findings, none blocking correctness (all self-heal), posted inline:

1. **[test-gap]** `tests/test_hive.py:975` — `test_local_hive_finalize_backstop_restores_lost_manifest`'s `os.remove` is swallowed by `_cell_work`'s try/except, so the test passes even if the primary init write regressed; it pins "finalize writes" not "lost→restored."
2. **[test-gap]** `tests/test_hive.py:1292` — the async setup wire test never exercises `output_credentials` threading (or `overwrite=True`); since the write is fire-and-forget, a drifted creds key would fail silently and only surface via the finalize backstop.
3. **[docs/nit]** `src/zagg/runner.py:3033` (and L2185 + the two docs) — "lands seconds into the run" presents a best-effort, concurrency-contended async timing as a guarantee; suggest softening so mid-run manifest availability isn't read as a contract.

No correctness bugs found in the phase-4 diff.

---

tests/test_hive.py
espg: 🤖 *from Claude*

Folded. The fake cell now asserts the manifest exists before removing it and records a `removed.append(True)` only on the successful assert+remove; after `agg()` the test asserts `removed == [True]`. So a regressed init write (assert fails) or a swallowed `FileNotFoundError` leaves `removed` empty and the post-run check fails loudly outside `_cell_work`'s swallow — the test now pins init→lost→restored, not just "finalize writes." Fix commit: d2a908550a18d1adf4cb8f1eb049dab5ee238834.

---

tests/test_hive.py
espg: 🤖 *from Claude*

Folded. Added `TestInvokeLambdaSetupEvent::test_hive_async_event_threads_creds_and_overwrite`, a second exact-dict wire pin mirroring the baseline one: it invokes `_invoke_lambda_setup_async` with `overwrite=True` and an `output_creds_event`, then asserts the Event `Payload` equals the full expected dict including `"overwrite": True` and `"output_credentials": creds` (and re-asserts `read` is never called). So both conditional payload branches now fail the pin if `output_credentials` or `overwrite` ever drifts on the primary fire-and-forget write. Fix commit: cca5e77b5867f00f86630e60b894faeca0bb696c.

---

src/zagg/runner.py

Folded. Softened the "lands seconds into the run" wording to "typically lands within seconds of init (best-effort: the Event invoke shares worker concurrency and runs retries-0, deferring to the finalize backstop under throttling or a dropped invoke)" so mid-run availability reads as opportunistic, not a contract. Applied consistently at the `_run_lambda` dispatch comment (~L2185) and the `_invoke_lambda_setup_async` docstring in `src/zagg/runner.py`, plus `docs/hive_layout.md` and `docs/design/sparse_coverage.md`. Also softened the identical claim in `src/zagg/hive.ensure_manifest`'s docstring (`src/zagg/hive.py`), which carries the same sentence — leaving it verbatim would have been the same misleading breadcrumb. Left `deployment/aws/lambda_handler.py` untouched. Fix commit: a9b7491309ee4220d0c50252557e36ad85a49cb9.

---

espg:
🤖 *from Claude*

**Run complete — the ratified hybrid is in, all phases done, ball is with espg** (label `waiting`).

Final state at a9b7491: 4 phase commits + 6 fold commits. Phase 4 lands the hybrid exactly as ratified ([make-it-so](https://github.com/englacial/zagg/issues/252#issuecomment-5006627722)): the sync `mode="ping"` fail-fast is unchanged on the wire, the manifest write fires as an async `InvocationType="Event"` invoke of the existing `mode="setup"` hive branch immediately after the ping (`_invoke_lambda_setup_async` — zero handler changes), and finalize's `ensure_manifest` is demoted to an idempotent backstop (self-heals a lost retries-0 async write; a crashed run still leaves a manifest from init). Local dispatcher: init write restored + finalize backstop kept. Docs/bench accounting updated; the dispatcher-direct S3 PUT was not implemented per the explicit OIDC rejection.

Phase-4 review findings were folded (3 commits) with every inline thread answered; no ruff-bot comments exist on the PR.

CI green on a9b7491: ruff, Lambda builds (x86_64 + arm64), and the 3.12/3.13 test matrix. Local full suite: 2170 passed / 30 skipped (branch baseline 2168/30 — the delta is phase 4's tests); `ruff format --check` clean; `ruff check` carries only the pre-existing N818 flagged in the PR body.

---

PR #256, opened 2026-07-16
Per-stage profiling for the raster sample path (issue #249)

### LINKED ISSUE #249: Per-stage profiling for the raster sample path (open/fetch/decode/geometry/gather)
🤖 *from Claude*

## Summary

The raster worker's `profile` output splits wall into `sample` / `write` only. The 0.31.0 A/B for issue #244 showed that is **not enough to attribute variance**: per-shard walls now spread 370–700 s run-to-run (the 0.30.0 baseline was uniform at 608 ± 4 s), the sample/write split is documented-approximate at `write_buffer > 1` and was misread once already in analysis, and none of the candidate explanations (container cold start, store-cache warmth) survived controls — cold sandboxes beat warm ones in the same run. We are out of hypotheses that can be tested with the current telemetry.

## Evidence (SERC full-year o9, one profiled invoke per shard)

| shard | 0.30.0 wb1 (23:00 UTC) | 0.31.0 wb1-A (06:45) | 0.31.0 wb2 (06:55) | 0.31.0 wb1-B (07:15) |
|---|---|---|---|---|
| 533 (light) | 485 | **369** | 504 | **372** |
| 137 | 607 | 698 | 471 | 489 |
| 449 | 608 | 691 | 681 | 488 |
| 345 | 614 | 687 | 669 | 687 |

Confirmed by the run regardless of the open question: peak memory −~320 MB (the #244 store/geometry caches working); shard 533 consistently −24%. Unattributable: which stage produces the ~490 vs ~690 throughput classes (sandbox network path? source-bucket fetch behavior by time of day? head-of-line effects on the now-shared connection pool?).

## Ask

Split the `sample` bucket into per-stage accumulators inside the worker, emitted under the existing opt-in `profile` key: `open` (TIFF header), `geometry` (pull-NN mapping incl. cache hits=0s), `fetch` (tile GETs), `decode`, `gather`, plus counts (assets, tiles fetched, cache hits). Zero overhead when `profile` is absent (the #100 convention). This makes the next #244 Phase-3 A/B attributive instead of correlational.

Refs #244 (Phase 3 blocked on this being decisive), PR #232 (protocol + approximate-split caveat), #228 (the stage timings are exactly the calibration terms the pre-flight model wants).

espg:
Closes #249

## What

Splits the raster worker's `sample` profile bucket into per-stage accumulators, emitted under the existing opt-in `profile` key (the issue #100 convention) as `phase_timings["stages"]`: seconds for `open` (store lookup + TIFF header round trips), `geometry` (pull-NN mapping; a `geom_cache` hit records ~0 s), `fetch` (tile GETs), `decode`, and `gather` (tile-index derivation + numpy scatter/gather), plus counts `assets`, `tiles`, `geom_hits`. This is the telemetry issue #249 needs to make the issue #244 Phase-3 A/B attributive instead of correlational.

**Point-path `write` split (phase 3, approved scope addition):** the point pipeline's profiled phase set gains an additive `write` phase on the hive worker path. The flat Lambda handler has carried a `write` bracket since issue #100, but `process_and_write_hive` — the shared per-shard write path both backends' hive branches call, and the configuration the per-merge benchmark runs — recorded no write timing at all. It now accumulates the leaf write-out (lazy leaf template + dense chunk writes + ragged + coverage sidecar + stamp) into `phase_timings["write"]` next to `process_shard`'s `read`/`index`/`aggregate`, behind the same opt-in gate (zero timing calls when `profile` is off) and the same clean-write attach gate as the flat handler. Approved in [espg's comment](https://github.com/englacial/zagg/pull/256#issuecomment-5006361723), ratifying understanding (1) of the [example-rendering comment](https://github.com/englacial/zagg/pull/256#issuecomment-5005931022). The display-side `agg = index + aggregate` mapping is harness work and rides issue #250, **not** this PR.

## Approach

Follows the [implementation plan on the issue](https://github.com/englacial/zagg/issues/249#issuecomment-4989305160) and espg's [sign-off on both open questions](https://github.com/englacial/zagg/issues/249#issuecomment-4989732869) (local backend: yes; keep `sample`/`write` keys: yes, additive only).

- `new_stage_stats()` in `src/zagg/processing/raster.py` returns the fresh accumulator (single-sourced schema, exported). It is threaded `process_raster_shard(..., stage_stats=None)` → `sample_item_async` → `_sample_one`, exactly like the issue #244 `geom_cache` pattern.
- **Zero overhead when absent:** `stage_stats=None` (the default, and the public `sample_asset*` path) makes no timing calls at all — `_sample_one` gates every `time.time()` behind `prof = stage_stats is not None`, mirroring the `_t0 = time.time() if profile else None` convention in `worker.py` / the handler.
- **No locks:** accumulation is plain `+=` on the event loop with no `await` between read and write (atomic by the same argument as the geom_cache store; the `write_buffer` sink threads never touch the dict). Noted in the `_sample_one` docstring.
- `_handle_process_raster` allocates the accumulator when `profile` is set and emits it as `phase_timings["stages"]` next to the unchanged `sample`/`write` split.
- Local backend (`RasterStrategy`, per espg's sign-off): the accumulator is allocated only when debug logging is enabled and logged per shard at debug level; the default path passes `None`.

### Measurement semantics (important)

The stage floats are **work volume, not a wall decomposition**. Each asset-sample times its own stages, and the K×bands concurrent samples of an invoke overlap on one event loop — so a per-coroutine stage wall includes time suspended while sibling samples ran, and stage sums can exceed the invoke's wall clock. That is deliberate and documented in `new_stage_stats()`: the numbers attribute *which stage differs* between a fast and a slow invoke (the #244 question), and avoid the wall-decomposition trap the `write_buffer > 1` sample/write split hit on PR #232. Per the phase-1 self-review fold, the tile-index derivation is bucketed into `gather` (two accumulation windows) so `fetch` stays a clean network-only signal; the only untimed sliver left is the fill-array construction on the no-valid-cells early return.

## Phases

- [x] Phase 1 — per-stage accumulators + counts wired through the worker (`raster.py`), emitted by the Lambda handler under `phase_timings["stages"]`, debug-logged per shard on the local backend; minimal update to the existing handler profile test so the phase lands green
- [x] Phase 2 — tests: stage counts (`assets == timesteps × bands`, `geom_hits == assets − distinct grids`, `tiles >= assets`) and non-negative stage seconds on synthetic COGs; profiled output identical to unprofiled; zero-timing-calls trap (module `time` rebound to raise) proving the absent-`profile` gate; handler emits the full stages schema and passes `stage_stats=None` without `profile`; local-backend debug log emits stages per shard (and not at INFO)
- [x] Phase 3 — point-path `write` phase split ([approved scope addition](https://github.com/englacial/zagg/pull/256#issuecomment-5006361723)): `process_and_write_hive` accumulates the leaf write-out into an additive `phase_timings["write"]` behind the existing `profile` gate (zero timing calls when off, flat-handler attach gate); tests for the profiled write phase (streaming + sharded), the errored-shard omission, the zero-timing-calls trap, profiled/unprofiled leaf byte parity, and the runner rollup carrying `write` generically

## Testing

`uv run pytest -q`: full suite green after phase 1 (2151 passed, 30 skipped); raster-path files green after each subsequent commit and the full suite re-run on the final tree (2164 passed, 30 skipped after phase 3). `ruff format --check`: clean. `ruff check`: one **pre-existing** failure unrelated to this diff (see below).

## Questions for review

1. **Pre-existing lint failure:** `ruff check src tests` fails on main with `N818 Exception name UnknownCapability should be named with an Error suffix` (`src/zagg/registry.py:64`) — present on a clean checkout of `origin/main`, not touched here, flagged per the conventions rather than fixed.
2. **Stage boundary choices:** `open` includes the IFD geo/dtype parse and the raster-center pyproj transform (all header-derived, µs-scale); the tile-index derivation is bucketed into `gather` per the phase-1 review fold, keeping `fetch` network-only.
3. **Local-backend gate:** stages on the local path are gated on `logger.isEnabledFor(logging.DEBUG)` (there is no `profile` knob on `agg()`); that reads as the natural local flavor of "opt-in", but say the word if you'd prefer an explicit parameter.
4. **Repo scope note:** CLAUDE.md §8 lists `github.com/espg` as the source-control scope while this repo (and the routine's remote) is `github.com/englacial/zagg`; proceeded since the routine, remote, and issue author all point here — flagging so §8 can be updated if the repo moved orgs.
5. The phase-1 commit carries session-trailer lines from the harness default; noticed the repo's title-only convention after pushing, and amending would need a force-push (never done here). Subsequent commits are title-only.
6. **Hive `write` includes leaf templating:** unlike the flat handler (whose timer starts after the store open/template check, because the dispatcher owns the flat template), the hive `write` bucket counts the lazy leaf template emission — in hive the worker owns its leaf's template PUTs, so excluding them would hide real write-out cost. Documented in the `process_and_write_hive` docstring; say the word if you'd rather match the flat exclusion.

### DIFF
--- a/src/zagg/hive.py
+++ b/src/zagg/hive.py
@@ -49,6 +49,7 @@ from __future__ import annotations
 
 import json
 import logging
+import time
 import warnings
 from datetime import datetime, timezone
 
@@ -843,9 +844,15 @@ def process_and_write_hive(
     the K carriers accumulate and the whole leaf is written once
     (``write_leaf_to_zarr`` — one ShardingCodec object per array), mirroring
     the flat sharded switch in ``runner._process_and_write``.
-    ``profile`` forwards to ``process_shard`` (issue #100); the write phase is
-    interleaved with the stream (or a single post-stream pass when sharded),
-    so no separate ``write`` timing is recorded.
+    ``profile`` forwards to ``process_shard`` (issue #100), which fills
+    ``metadata["phase_timings"]`` with read/index/aggregate; the leaf write
+    work — interleaved with the stream (or a single post-stream pass when
+    sharded), plus the ragged/coverage/stamp finalize — accumulates into an
+    additive ``write`` phase alongside them (issue #249). Unlike the flat
+    path, the lazy leaf template emission counts as write: in hive the worker
+    owns its leaf's template PUTs (there is no dispatcher-side template), so
+    excluding them would hide real write-out cost. Default ``False`` makes
+    zero timing calls and leaves ``metadata`` unchanged — no probe tax.
 
     ``window`` (issue #246, D13/D15) is one dispatch unit's time window:
     ``{"label", "start", "end"}``, bounds half-open in DATASET units
@@ -877,6 +884,7 @@ def process_and_write_hive(
 
     leaf_path = shard_leaf_path(store_root, shard_key, window=window["label"] if window else None)
     box: dict = {}
+    _write_elapsed = 0.0
 
     def _leaf():
         if "store" not in box:
@@ -927,11 +935,15 @@ def process_and_write_hive(
     ragged_chunks: list = []
 
     def _write_chunk(block_index, carrier, ragged):
+        nonlocal _write_elapsed
+        _t0 = time.time() if profile else None
         store = _leaf()
         local = leaf_block_index(grid, block_index, shard_key)
         write_dataframe_to_zarr(carrier, store, grid=grid, chunk_idx=local)
         if ragged:
             ragged_chunks.append((local, ragged))
+        if profile:
+            _write_elapsed += time.time() - _t0
 
     # Occupied-cell sink (issue #200): the worker already holds the shard's
     # populated cell words; collect them here to derive the stamp's coverage.
@@ -966,7 +978,10 @@ def process_and_write_hive(
     # the ``.zarr/`` prefix — same contract as the streaming path's first
     # ``_write_chunk``.
     if sharded and chunk_results and not metadata.get("error"):
+        _t0 = time.time() if profile else None
         write_leaf_to_zarr(chunk_results, _leaf(), grid=grid, shard_key=int(shard_key))
+        if profile:
+            _write_elapsed += time.time() - _t0
     # Stamp ONLY a fully-written leaf: an errored shard (or one that streamed
     # no chunks) stays unstamped — debris by definition (D4). The stamp is the
     # last write, so its presence certifies everything before it landed — the
@@ -976,6 +991,7 @@ def process_and_write_hive(
     # The leaf write order is pinned: dense (streamed, or one object each when
     # sharded) -> ragged (one object, issue #209) -> coverage sidecar -> stamp.
     if "store" in box and not metadata.get("error"):
+        _t0 = time.time() if profile else None
         if not sharded:
             write_ragged_leaf_to_zarr(ragged_chunks, box["store"], grid=grid)
         words = np.concatenate(occupied) if occupied else None
@@ -1004,6 +1020,16 @@ def process_and_write_hive(
             window=window["label"] if window else None,
             time_range=time_range,
         )
+        if profile:
+            _write_elapsed += time.time() - _t0
+    # Write-phase split (issue #249): read/index/aggregate come from
+    # ``process_shard``; ``write`` is the leaf write-out above (template +
+    # dense chunks + ragged + coverage sidecar + stamp). Same gate as the flat
+    # Lambda handler's issue #100 write bracket: only a clean, actually-written
+    # shard carries it, so a time-to-failure never lands as a write duration
+    # and a no-data shard (no leaf) stays write-less.
+    if profile and not metadata.get("error") and "phase_timings" in metadata and "store" in box:
+        metadata["phase_timings"]["write"] = _write_elapsed
     return metadata
 
 
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -24,6 +24,7 @@ import concurrent.futures
 import os
 import re
 import threading
+import time
 from urllib.parse import urlparse
 
 import numpy as np
@@ -154,6 +155,38 @@ def _build_store(key):
     return LocalStore(loc)
 
 
+def new_stage_stats() -> dict:
+    """Fresh per-invoke stage accumulator for the sample path (issue #249).
+
+    The floats are wall-clock seconds (``time.time()`` deltas, the issue #100
+    convention) of **stage work volume**: each asset-sample times its own
+    stages independently and the K x bands concurrent samples of an invoke
+    overlap on one event loop, so a stage total is the sum of per-sample
+    elapsed walls *including* time suspended while sibling samples ran. The
+    sums attribute where the samples' time went (which stage differs between
+    a fast and a slow invoke) — they are NOT a wall decomposition and can
+    exceed the invoke's wall clock. That is deliberate: the ``write_buffer >
+    1`` sample/write remainder on PR #232 already showed wall splits go
+    approximate under overlap.
+
+    Keys — seconds: ``open`` (store lookup + TIFF header round trips + geo/
+    dtype parse), ``geometry`` (pull-NN mapping; a ``geom_cache`` hit records
+    ~0), ``fetch`` (tile GETs), ``decode``, ``gather`` (numpy scatter/gather).
+    Counts: ``assets`` (asset-samples), ``tiles`` (tiles fetched),
+    ``geom_hits`` (mappings served from ``geom_cache``).
+    """
+    return {
+        "open": 0.0,
+        "geometry": 0.0,
+        "fetch": 0.0,
+        "decode": 0.0,
+        "gather": 0.0,
+        "assets": 0,
+        "tiles": 0,
+        "geom_hits": 0,
+    }
+
+
 async def sample_asset_async(
     grid,
     cells,
@@ -194,6 +227,7 @@ async def _sample_one(
     anonymous: bool = True,
     fill=0,
     geom_cache: dict | None = None,
+    stage_stats: dict | None = None,
 ):
     """:func:`sample_asset_async` body, also returning the raster's center
     ``(lon, lat)`` — the ownership rule's tile-center input (#218).
@@ -204,9 +238,20 @@ async def _sample_one(
     computes it once per distinct grid (~175 ms at o9) instead of once per
     asset-sample. ``None`` (the default, and the public ``sample_asset*``
     path) computes per call, unchanged.
+
+    ``stage_stats`` (issue #249) accumulates per-stage seconds + counts in
+    place — see :func:`new_stage_stats` for the keys and the work-volume (not
+    wall-decomposition) semantics. ``None`` (the default, and the public
+    ``sample_asset*`` path) makes no timing calls at all — the hot path is
+    unchanged. Accumulation is plain ``+=`` on the event loop with no await
+    between read and write, so it is atomic by the same argument as the
+    ``geom_cache`` store below — no locks; the ``write_buffer`` sink threads
+    never touch this dict.
     """
     from async_tiff import TIFF
 
+    prof = stage_stats is not None
+    _t0 = time.time() if prof else None
     store, path = _store_and_path(href, region=region, anonymous=anonymous)
     tiff = await TIFF.open(path, store=store)
     ifd = tiff.ifds[0]
@@ -223,8 +268,14 @@ async def _sample_one(
     epsg, transform = _geo_from_ifd(ifd)
     shape = (ifd.image_height, ifd.image_width)
     center = _raster_center_lonlat(epsg, transform, shape)
+    if prof:
+        stage_stats["open"] += time.time() - _t0
+        stage_stats["assets"] += 1
+        _t0 = time.time()
     geom_key = (epsg, transform, shape)
     geom = geom_cache.get(geom_key) if geom_cache is not None else None
+    if prof and geom is not None:
+        stage_stats["geom_hits"] += 1
     if geom is None:
         # INVARIANT (issue #244 thread): no ``await`` between this check and
         # the store below. asyncio interleaves only at await points, so the
@@ -243,6 +294,8 @@ async def _sample_one(
         if geom_cache is not None:
             geom_cache[geom_key] = geom
     rows, cols, valid = geom
+    if prof:
+        stage_stats["geometry"] += time.time() - _t0
 
     th, tw = ifd.tile_height, ifd.tile_width
     vr, vc = rows[valid], cols[valid]
@@ -252,8 +305,17 @@ async def _sample_one(
         return np.full(rows.shape, fill, dtype=dtype), valid, center
 
     pairs = np.unique(np.stack([tr, tc], axis=1), axis=0)
+    if prof:
+        _t0 = time.time()
     tiles = await asyncio.gather(*[tiff.fetch_tile(int(c), int(r), 0) for r, c in pairs])
+    if prof:
+        stage_stats["fetch"] += time.time() - _t0
+        stage_stats["tiles"] += len(pairs)
+        _t0 = time.time()
     decoded = await asyncio.gather(*[t.decode() for t in tiles])
+    if prof:
+        stage_stats["decode"] += time.time() - _t0
+        _t0 = time.time()
 
     gathered = np.full(vr.shape, fill, dtype=dtype)
     for (trow, tcol), dec in zip(pairs, decoded):
@@ -263,6 +325,8 @@ async def _sample_one(
 
     values = np.full(rows.shape, fill, dtype=dtype)
     values[valid] = gathered
+    if prof:
+        stage_stats["gather"] += time.time() - _t0
     return values, valid, center
 
 
@@ -323,6 +387,7 @@ async def sample_item_async(
     region: str | None = None,
     anonymous: bool = True,
     geom_cache: dict | None = None,
+    stage_stats: dict | None = None,
 ):
     """Sample every configured band of one STAC item, concurrently.
 
@@ -366,6 +431,7 @@ async def sample_item_async(
                 anonymous=anonymous,
                 fill=bands[f]["fill_value"],
                 geom_cache=geom_cache,
+                stage_stats=stage_stats,
             )
             for f in fields
         ]
@@ -496,6 +562,7 @@ def process_raster_shard(
     region: str | None = None,
     anonymous: bool = True,
     on_slab=None,
+    stage_stats: dict | None = None,
 ):
     """Process one shard of a raster pipeline: every acquisition group -> slab.
 
@@ -521,6 +588,14 @@ def process_raster_shard(
         double-buffer); the returned ``slabs`` is then empty. When ``None``
         (the default) every slab accumulates into ``slabs`` and is returned,
         as before, and ``write_buffer`` is ignored.
+    stage_stats : dict, optional
+        Per-invoke stage accumulator from :func:`new_stage_stats` (issue
+        #249): the sample path adds per-stage seconds (``open`` / ``geometry``
+        / ``fetch`` / ``decode`` / ``gather``) and counts (``assets`` /
+        ``tiles`` / ``geom_hits``) in place. Stage seconds are work volume,
+        not a wall decomposition — concurrent samples overlap, so their sum
+        can exceed this call's wall (see :func:`new_stage_stats`). ``None``
+        (the default) makes no timing calls — the sample path is unchanged.
 
     Returns
     -------
@@ -584,6 +659,7 @@ def process_raster_shard(
                             region=region,
                             anonymous=anonymous,
                             geom_cache=geom_cache,
+                            stage_stats=stage_stats,
                         )
                         for e in items
                     ]
@@ -792,6 +868,7 @@ def write_raster_coords(store, grid, shard_key: int):
 
 
 __all__ = [
+    "new_stage_stats",
     "sample_asset",
     "sample_asset_async",
     "sample_item_async",
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -171,7 +171,8 @@ def new_stage_stats() -> dict:
 
     Keys — seconds: ``open`` (store lookup + TIFF header round trips + geo/
     dtype parse), ``geometry`` (pull-NN mapping; a ``geom_cache`` hit records
-    ~0), ``fetch`` (tile GETs), ``decode``, ``gather`` (numpy scatter/gather).
+    ~0), ``fetch`` (tile GETs), ``decode``, ``gather`` (tile-index derivation
+    + numpy scatter/gather).
     Counts: ``assets`` (asset-samples), ``tiles`` (tiles fetched),
     ``geom_hits`` (mappings served from ``geom_cache``).
     """
@@ -296,16 +297,20 @@ async def _sample_one(
     rows, cols, valid = geom
     if prof:
         stage_stats["geometry"] += time.time() - _t0
+        _t0 = time.time()
 
     th, tw = ifd.tile_height, ifd.tile_width
     vr, vc = rows[valid], cols[valid]
     tr, tc = vr // th, vc // tw
 
     if vr.size == 0:
+        if prof:
+            stage_stats["gather"] += time.time() - _t0
         return np.full(rows.shape, fill, dtype=dtype), valid, center
 
     pairs = np.unique(np.stack([tr, tc], axis=1), axis=0)
     if prof:
+        stage_stats["gather"] += time.time() - _t0
         _t0 = time.time()
     tiles = await asyncio.gather(*[tiff.fetch_tile(int(c), int(r), 0) for r, c in pairs])
     if prof:
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -876,6 +876,7 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
         from zagg.config import get_layout
         from zagg.grids import from_config
         from zagg.processing.raster import (
+            new_stage_stats,
             process_raster_shard,
             write_raster_coords,
             write_raster_slab,
@@ -906,10 +907,15 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
         # remainder — the split the PR #232 double-buffer decision needs
         # measured. Exact at write_buffer=1 (writes serialize in the loop);
         # at write_buffer>1 writes overlap sampling, so the remainder is
-        # approximate — A/B on ``duration_s`` instead. Default (no
-        # ``profile`` key) emits nothing: the body stays byte-identical.
+        # approximate — A/B on ``duration_s`` instead. ``stages`` (issue
+        # #249) splits the sample bucket into per-stage work volumes +
+        # counts — attribution, not a wall decomposition: concurrent samples
+        # overlap, so stage sums can exceed ``sample`` (see
+        # ``new_stage_stats``). Default (no ``profile`` key) emits nothing:
+        # the body stays byte-identical and the sample path times nothing.
         profile = bool(event.get("profile"))
         write_s = 0.0
+        stage_stats = new_stage_stats() if profile else None
 
         def _write_slab(t_idx: int, slab: dict) -> None:
             nonlocal wrote, write_s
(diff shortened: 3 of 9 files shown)


src/zagg/processing/raster.py
```diff
@@ -252,8 +305,17 @@ async def _sample_one(
         return np.full(rows.shape, fill, dtype=dtype), valid, center
 
     pairs = np.unique(np.stack([tr, tc], axis=1), axis=0)
+    if prof:
```
espg, line 312: 🤖 *from Claude (review)*

nit (attribution blind spot): the `fetch` timer starts here, but the tile-index derivation that runs between the `geometry` stop-timer (line 298) and this point is bucketed into **no** stage:

```python
    rows, cols, valid = geom
    if prof:
        stage_stats["geometry"] += time.time() - _t0   # geometry stops

    th, tw = ifd.tile_height, ifd.tile_width
    vr, vc = rows[valid], cols[valid]        # fancy index — allocates
    tr, tc = vr // th, vc // tw
    if vr.size == 0: ...
    pairs = np.unique(np.stack([tr, tc], axis=1), axis=0)  # O(n log n)
    if prof:
        _t0 = time.time()                    # fetch starts (this line)
```

`rows[valid]` and the `np.unique` over per-cell tile coords are real numpy work on potentially large `valid` arrays, and they land in neither `geometry` nor `gather`. That's a gap the concurrency caveat doesn't excuse — it's not overlap-driven overspill (which the docstrings honestly cover), it's plain unmeasured serial work, and #249's whole purpose is attributing where sample time goes for the 490 vs 690 s variance question.

Concrete fix: move the `_t0 = time.time()` fetch-start up to just after line 298 (fold the tile-index derivation into `fetch`), or record it under `gather` since it's numpy scatter/gather prep. Either bucket is defensible; leaving it uncounted is the one option that quietly understates a stage. Non-blocking.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the per-stage raster profiling (issue #249). One non-blocking nit inline; the rest of the diff holds up. What I checked and why it passes:

- **Zero-overhead gate (the #100 convention).** With `stage_stats=None`, `prof` is `False` and `_t0 = time.time() if prof else None` short-circuits — no `time.time()` call runs on the default path, every `if prof:` block is skipped, and the returned arrays are byte-identical. Matches how `worker.py` gates its phase timings. Handler emits nothing extra when `profile` is absent (the whole `if profile:` block is skipped), so the response body stays byte-identical.
- **Stage attribution.** `open` covers store lookup + `TIFF.open` + ifd/geo/dtype parse; `assets` bumps once per `_sample_one`; a `geom_cache` hit skips the compute and records `~0 s` under `geometry` while bumping `geom_hits` exactly once; `tiles += len(pairs)` matches the number of distinct tiles fetched; `fetch`/`decode`/`gather` are each cleanly delimited by `_t0` resets with no unmeasured gaps *except* the tile-index derivation noted inline. The `vr.size == 0` early return correctly records `open`/`geometry`/`assets` and skips `fetch`/`decode`/`gather`/`tiles` (zero work) — consistent with the additive schema.
- **Concurrency honesty.** `new_stage_stats`, `_sample_one`, `process_raster_shard`, and the handler comment all state that per-stage sums are *work volume including suspended time under async overlap* and can exceed wall — no docstring implies a wall decomposition. Accurate.
- **Atomicity.** Confirmed a single event loop per shard (`_run_sync(_run_all())`); each `+=` has no `await` between read and write, and the `write_buffer` `to_thread` sinks only touch `on_slab`, never `stage_stats` — so the lock-free claim is sound.
- **Handler.** `stages` is a dict of floats/ints → JSON-serializable; `sample`/`write` unchanged (additive only); `stage_stats` is non-`None` wherever it's emitted.
- **Runner debug gate.** `new_stage_stats()` allocated only under `logger.isEnabledFor(logging.DEBUG)`, one accumulator per `_one`/shard, logged per shard at debug via `shard_label` (both `logging` and `shard_label` are imported). The f-string only builds when debug is on.
- **CLAUDE.md conformance.** Terse, mirrors the #244 `geom_cache` threading pattern, stdlib `time` only (no new deps), no lint/type suppressions. The one test edit updates the profile-present key set to `{"sample","write","stages"}` and keeps the `sample+write ≈ duration` assertion — not a weakening. Dedicated stage-assertion tests are declared as phase 2, so not flagged here.

---

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Bucketed the tile-index derivation (`tr, tc`, `pairs = np.unique(...)`) into `gather` via two accumulation windows — one closing at the `vr.size == 0` early return, one closing just before the fetch `_t0` reset — so `fetch` stays a clean network-only signal for the #244 variance A/B. Also updated the `new_stage_stats` docstring's `gather` description to "tile-index derivation + numpy scatter/gather". Fixed in 1fe2acf.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of the cumulative diff, focused on the fold commit (`1fe2acf`) and phase-2 tests (`0a347ce`). **No blocking findings — this phase passes.**

**Fold commit (`gather` window split).** Verified correct. `gather` now accumulates two disjoint windows — tile-index derivation (`_t0` at the geometry close through the `np.unique` on `pairs`) and the numpy scatter/gather at the tail — with no double-count and no gap: the `fetch` `_t0` reset closes the first `gather` window before the tile GETs start. The `vr.size == 0` early return correctly folds the tile-index prep into `gather` before returning, so nothing is dropped. `new_stage_stats()` docstring updated to `gather` = *tile-index derivation + numpy scatter/gather* — matches the code.

**Zero-overhead gate tests.** Solid and non-vacuous:
- `test_default_path_makes_no_timing_calls` — `time.` is referenced in `raster.py` *only* inside `_sample_one` (the module's other time usage is `datetime.fromisoformat`), so rebinding the module-level `time` to a raising stub catches exactly the sample path and nothing else; `monkeypatch` reverts it so no leak to sibling modules. The `meta["timesteps"] == 1` assertion proves the sample path actually executed, so the trap can't pass vacuously.
- `test_no_profile_passes_no_stage_stats` — the handler's in-function `from zagg.processing.raster import process_raster_shard` resolves through the patched module attribute, and the fake's signature matches the call site; correctly asserts `stage_stats is None`.

**Count derivations checked against the fixtures.** `assets == timesteps × bands`, `geom_hits == assets − distinct source grids` (identical `(epsg, transform, shape)` across timesteps for each of red 96×96@10m and scl 48×48@20m ⇒ 6−2=4; handler single asset ⇒ 0; runner 2 shared-grid timesteps ⇒ 1), and `tiles ≥ assets` all follow from the full-coverage rect grid. Runner debug test: `zagg.runner` logger, `isEnabledFor(DEBUG)` gate flipped by `caplog.at_level(..., logger="zagg.runner")`, one shard ⇒ exactly one `stages:` line.

**Concurrency semantics.** Work-volume (not wall-decomposition) accounting is documented and correct — `+=` runs on the loop with no await between read and write of a key, atomic by the same argument as the `geom_cache` store; intervals that span `await` deliberately include suspended sibling time.

**Coverage.** Every behavioral surface is exercised (worker threading via local `agg`, handler emission, runner debug log + its off state, both zero-overhead gates, output parity). No tests weakened; none would pass with the feature broken (the sum-of-stages `> 0` and `timesteps` guards close the obvious vacuity holes).

Timing-flakiness note: the `sum(stages) > 0.0` assertions rely on `open` doing real TIFF I/O against `time.time()`'s ns-resolution clock — not a realistic 0.0 on Linux, so not flagged.

---

espg:
🤖 *from Claude*

**Status: all phases complete — awaiting review.**

- Phase 1 (4861733): `new_stage_stats()` accumulator threaded `process_raster_shard` → `sample_item_async` → `_sample_one`; handler emits `phase_timings["stages"]` next to the unchanged `sample`/`write`; local backend logs stages per shard at debug level. Adversarial self-review found [one nit](https://github.com/englacial/zagg/pull/256#discussion_r3594663409) (unattributed tile-index math), folded in 1fe2acf — `gather` absorbs it so `fetch` stays a network-only signal; thread answered.
- Phase 2 (0a347ce): tests — stage counts (`assets == timesteps × bands`, `geom_hits == assets − distinct grids`, `tiles ≥ assets`), profiled/unprofiled output parity, a raising-`time` trap proving zero timing calls when `stage_stats` is absent, handler stages schema + `stage_stats=None` without `profile`, and the local debug-log gate (on at DEBUG, silent at INFO). The phase-2 adversarial review posted [no findings](https://github.com/englacial/zagg/pull/256#pullrequestreview-4712877881). No review findings left standing.
- CI green on 0a347ce (test 3.12/3.13, ruff, lambda builds); local `pytest -q` 2158 passed / 30 skipped, `ruff format --check` clean. `ruff check` carries only the pre-existing N818 in `src/zagg/registry.py` (also fails on a clean `main`; the CI ruff job selects E,F,W,I so it doesn't trip there) — left untouched per the conventions.
- Process note: this run's phase-2 self-review + fold ran as the usual separate subagents, but the review posting was delayed and its fold pass was re-verified manually before concluding there was nothing to fold — no double-posting occurred.
- **Draft state:** CLAUDE.md §2 says a finished PR should be flipped ready-for-review, but this routine's configuration mandates drafts-only — a direct conflict, so per the conventions' conflict rule the flip is left to a human rather than guessed. Everything else is in the §2 "ready" state: phases done, findings folded, CI green.

---

espg:
this is bit too tightly scoped for the behavior that we actually want-- we don't just want to grab the profile information on the raster path, we also want to display that information in the benchmark.md output as a table and a graph.

Furthermore, we want to do that *with the other benchmarks too*.

Moving forward, this means the following:

- single shard from AOI , showing timings for each phase (read, agg, write, total wall). All those phases have there own plot, and are summarized in a table. We only do this for one configuration: o9, hive, sharded, without AOI binary mask, tdigest, with the default threading (K / shard_workers=4) on the default worker lambda (4GB). We don't report numbers normalized per 100 sq km anymore, just the cost / time per phases for the full shard. We do this per merge to main.
- full neon AOI performance and cost. As above, same configuration except that we add the AOI binary mask option. For this, we report the full cost of run (in lambda seconds / dollars), along with the breakout of the phases (read, write, etc); we also report the wall time for all shards to run. We do this per new release of zagg.
- full neon AOI on the raster pipeline. This does a year (2025) of sentinel-2 data, using the raster pipeline, for our neon AOI. Full per phase timings (with plots) as well as the total time / cost. This doesn't call sharded (obviously), but still runs in hive, and carries the AOI mask option. We do this per new release of zagg.

The per release benchmarks have a top level plot / table showing just the total cost (dual axis lambda seconds and dollars spent), and overall wall time, with a simple table that shows it in tabular format. Under that, there a separate set of plots and tables showing the diagnostic benchmarks (i.e., split per phase). As a separate section beneath that, we have the per commit to main plots and tables, which follow the same split (top level summary vs diagnostics).

Please give an example rendering of this as a comment in this PR so I can verify that things look correct. Memory color bar stays the same as previous iterations. 

Anything that's out of scope for you (i.e., touches .github workflow) should be posted as a comment to #250 , where a local agent will implement.

espg:
🤖 *from Claude*

Below is the **example rendering** of the restructured `docs/deployment/benchmark.md` requested in [espg's comment](https://github.com/englacial/zagg/pull/256#issuecomment-5005441739). Everything between the horizontal rules is the proposed page, laid out exactly per the spec: per-release section first (top-level summary, then diagnostics), then the per-commit-to-main section with the same split.

**How to read the mock:**

- **All numbers are ILLUSTRATIVE.** They are scaled from real measurements so magnitudes are plausible — the raster shard walls come from the issue #249 evidence table (372/489/488/687 s per shard on the 0.31.0 wb1-B run), the 3.7 s hive setup and the read/index/aggregate phase framing from issue #250, and every dollar figure is derived with the repo's own cost model (`LAMBDA_PRICE_PER_GB_SEC = 0.0000133334`, 4 GB — `src/zagg/dispatch.py:320-321`, so $0.0000533336 per lambda-second). None of them are from an actual run of the proposed harness.
- **The real artifacts are matplotlib PNGs** generated by the existing benchmark machinery (`.github/scripts/plot_series.py` — the shared `_render_panel_grid` / `_render_table` renderers) and embedded by raw URL from the `benchmarks` branch, exactly as today's benchmark.md does. GitHub comments can't embed unrendered PNGs, so each figure appears here twice: a boxed **[PNG placeholder]** spec (plot type, axes, series, color bar) plus a `mermaid` chart that shows the *shape only* (mermaid can't do dual axes or the memory color scale).
- **Memory color bar is unchanged:** markers colored by peak RSS as % of the 4 GB cap on the `RdYlGn_r` green→red ramp (`plot_series.py:39`), one shared horizontal colorbar with the %-of-cap scale and the absolute-MB twin axis (`plot_series.py:579-593`), and the table's `% cap` cell shaded on the same scale (`plot_series.py:214-224`).

---

# Benchmark results

Cost and runtime of the zagg compute pipeline on AWS Lambda over the NEON AOI,
tracked as two cadences: **per release** (full-AOI truth) and **per commit to
`main`** (single-shard regression tracking). All runs: arm64, 4 GB worker,
o9, hive store layout, tdigest (point pipeline), default threading
(`shard_workers=4`). Costs are reported for the full run — no per-100 km²
normalization.

## Per-release benchmarks

Two full-AOI NEON runs per release tag:

- **Point pipeline (tdigest)** — o9, hive, sharded, **with** AOI binary mask.
- **Raster pipeline (Sentinel-2, year 2025)** — o9, hive, **not** sharded, with AOI binary mask.

### Summary — total cost and wall time

```text
[PNG placeholder — full_aoi_summary.png]
matplotlib figure, 2 rows (point | raster) × 2 columns (total cost | wall time).
- Left panels: total cost vs release tag. DUAL Y-AXIS: lambda-seconds (left)
  and dollars (right). Fixed 4 GB price means USD = lambda-s × $0.0000533336,
  so the right axis is an exact relabeling of the same line. Markers colored
  by peak worker memory (% of 4 GB cap, green→red RdYlGn_r).
- Right panels: overall wall time (s) vs release tag (all shards, dispatch wall).
- x-axis: release tag (0.30.0, 0.31.0, 0.32.0, ...).
- Shared horizontal memory colorbar on top: % of cap + MB twin axis (unchanged).
```

```mermaid
xychart-beta
    title "Total cost per release (lambda-seconds; USD axis is a fixed rescale) — ILLUSTRATIVE"
    x-axis ["0.30.0", "0.31.0", "0.32.0"]
    y-axis "lambda-seconds" 0 --> 2600
    line [1010, 985, 972]
    line [2410, 2310, 2036]
```

*(top line: raster full-AOI; bottom line: point/tdigest full-AOI)*

| run | release | lambda-s | cost (USD) | wall (s) | mem (MB) | % cap |
|---|---:|---:|---:|---:|---:|---:|
| full-AOI point (tdigest, mask, hive, sharded) | 0.32.0 | 972 | $0.0518 | 273.5 | 2210 | 54% |
| full-AOI raster (S2 2025, mask, hive) | 0.32.0 | 2036 | $0.1086 | 694.1 | 2890 | 71% |

*(all values ILLUSTRATIVE; `% cap` cell shaded green→red in the rendered PNG table, as today)*

### Diagnostics — per-phase breakdown

#### Point pipeline (tdigest full-AOI)

```text
[PNG placeholder — full_aoi_point_phases.png]
matplotlib figure, one panel PER PHASE (read | agg | write | setup), each panel:
phase seconds (max across shards, matching the wall framing) vs release tag,
memory-colored markers, shared colorbar. setup is its own tracked line so the
flat→hive collapse (104–110 s → 3.7 s, issue #250) is visible; finalize is
dropped from display (~0 on this path).
```

```mermaid
xychart-beta
    title "Point pipeline per-phase seconds (max across shards) — ILLUSTRATIVE"
    x-axis ["0.30.0", "0.31.0", "0.32.0"]
    y-axis "seconds" 0 --> 200
    line [166, 158, 152]
    line [84, 76, 71]
    line [34, 30, 28]
    line [107, 106, 3.7]
```

*(lines top→bottom at left edge: read, setup, agg, write — setup collapses at the hive switch)*

| phase | seconds (max shard) | lambda-s (all shards) | cost (USD) | share of billed |
|---|---:|---:|---:|---:|
| read | 152.1 | 573 | $0.0306 | 59% |
| agg | 71.4 | 268 | $0.0143 | 28% |
| write | 28.3 | 106 | $0.0057 | 11% |
| setup (template) | 3.7 | 3.7 | $0.0002 | <1% |
| **total billed** | — | **972** | **$0.0518** | 100% |
| total wall (all shards) | **273.5** | — | — | — |

#### Raster pipeline (Sentinel-2 2025 full-AOI)

Phases are this PR's per-stage split (`phase_timings["stages"]`): `open`,
`geometry`, `fetch`, `decode`, `gather`, plus the worker `write` bucket.
**Stage seconds are work volume, not a wall decomposition** — concurrent
asset-samples overlap on one event loop, so stage sums can exceed wall (the
measurement-semantics note in this PR's description); the plots therefore show
each stage as its own series, never stacked to a wall total.

```text
[PNG placeholder — full_aoi_raster_phases.png]
matplotlib figure, one panel PER STAGE (open | geometry | fetch | decode |
gather | write), each panel: stage seconds (max shard) vs release tag,
memory-colored markers, shared colorbar (unchanged convention).
```

| stage | seconds (max shard, work volume) | notes |
|---|---:|---|
| fetch | 512 | tile GETs (network-only signal) |
| decode | 301 | |
| gather | 154 | incl. tile-index derivation (phase-1 fold) |
| write | 41 | |
| open | 38 | store lookup + TIFF headers |
| geometry | 12 | pull-NN mapping; geom-cache hits ≈ 0 s |
| **total wall (all shards)** | **694.1** | stage sum (1058) > wall by design — overlapped I/O |
| **total billed** | **2036 lambda-s / $0.1086** | |

## Per commit to `main`

Single densest shard from the NEON AOI, **one configuration only**: o9, hive,
sharded, **no** AOI binary mask, tdigest, `shard_workers=4`, 4 GB worker.
Full-shard cost/time per phase — no per-100 km² normalization. Same
top-level-summary vs diagnostics split as the per-release section.

### Summary — total cost and wall time

```text
[PNG placeholder — merge_summary.png]
matplotlib figure, 1 row × 2 panels: (left) total cost vs merge sha, DUAL
Y-AXIS lambda-seconds / dollars, memory-colored markers; (right) total wall (s)
vs merge sha. x-axis: commit shas (7 chars, rotated), as in today's per-merge
panels.
```

```mermaid
xychart-beta
    title "Per-merge total cost (lambda-seconds) — ILLUSTRATIVE"
    x-axis ["3f1c2ab", "7e90d14", "b22a9c0", "0a347ce"]
    y-axis "lambda-seconds" 0 --> 120
    line [91, 88, 84, 83]
```

| commit | lambda-s | cost (USD) | wall (s) | mem (MB) | % cap |
|---|---:|---:|---:|---:|---:|
| `0a347ce` | 83.0 | $0.00443 | 88.1 | 1650 | 40% |

### Diagnostics — per-phase breakdown

```text
[PNG placeholder — merge_phases.png]
matplotlib figure, one panel PER PHASE (read | agg | write | total wall), each
panel: seconds vs merge sha, memory-colored markers, shared %-cap + MB
colorbar. Each phase gets its own plot per the spec.
```

```mermaid
xychart-beta
    title "Per-merge phase seconds — ILLUSTRATIVE"
    x-axis ["3f1c2ab", "7e90d14", "b22a9c0", "0a347ce"]
    y-axis "seconds" 0 --> 60
    line [50.9, 49.5, 46.8, 46.2]
    line [24.0, 23.1, 22.2, 21.8]
    line [9.6, 9.4, 9.0, 8.9]
```

*(lines top→bottom: read, agg, write)*

| phase | seconds | cost (USD) |
|---|---:|---:|
| read | 46.2 | $0.00246 |
| agg | 21.8 | $0.00116 |
| write | 8.9 | $0.00047 |
| setup + untimed | 6.1 | $0.00033 |
| **total billed** | **83.0** | **$0.00443** |
| total wall | **88.1** | — |

---

**Three understandings to confirm** (answerable in one pass):

1. **Point-path phase names.** The point worker currently emits `read` / `index` / `aggregate` (`summary["worker_phase_max"]`, issue #250); the spec asks for read / **agg** / **write**. Proposed mapping: `agg = index + aggregate`, and split `write` out as its own emitted phase — the write split needs a small worker change, which fits this PR's profiling scope. Say the word if you'd rather keep `index` as a fourth displayed phase.
2. **Raster diagnostics use this PR's stage set** (`open`/`geometry`/`fetch`/`decode`/`gather` + `write`) with the work-volume caveat displayed (stage sums can exceed wall; never stacked). Alternative is max-shard coarse read/agg/write only — but that would discard exactly the attribution issue #249 asked for.
3. **Retirement:** the current live 2×2 (inline/sidecar × AOI-mask) figures and both cost-per-100 km² figures move to the Archived section (retained on the `benchmarks` branch, not regenerated), since the per-merge leg becomes the single hive configuration and per-area normalization is dropped.

Everything that touches `.github/workflows/` or `.github/scripts/` (cadence hooks, series schema, plot generation, cost computation) is out of scope for this routine and has been posted as a checklist on issue #250 for the local agent, per the directive.

Applying the `waiting` label: awaiting espg's verification that this rendering looks correct — implementation proceeds on this PR (worker-side emission) and via issue #250 (harness side) once the shape is confirmed.

---

espg:
> Proposed mapping: agg = index + aggregate, and split write out as its own emitted phase — the write split needs a small worker change, which fits this PR's profiling scope. 

approved.

> Raster diagnostics use this PR's stage set (open/geometry/fetch/decode/gather + write) with the work-volume caveat displayed (stage sums can exceed wall; never stacked)

looks great

> Retirement: the current live 2×2 (inline/sidecar × AOI-mask) figures and both cost-per-100 km² figures move to the Archived section (retained on the benchmarks branch, not regenerated), since the per-merge leg becomes the single hive configuration and per-area normalization is dropped.

correct

src/zagg/hive.py
```diff
@@ -966,7 +978,10 @@ def _write_chunk(block_index, carrier, ragged):
     # the ``.zarr/`` prefix — same contract as the streaming path's first
     # ``_write_chunk``.
     if sharded and chunk_results and not metadata.get("error"):
+        _t0 = time.time() if profile else None
```
espg, line 981: 🤖 *from Claude (review)*

The zero-overhead-when-unprofiled guarantee is only pinned for the streaming + finalize brackets, not for this sharded post-stream bracket. `test_default_path_makes_no_timing_calls` (tests/test_hive.py) installs the `hive.time` = `_Boom` trap but drives it with the non-sharded `self._grid(cfg)`, so `write_chunk` and the finalize block run but this `write_leaf_to_zarr` bracket (line 981) never executes. A regression that dropped the `if profile else None` guard here — making it an unconditional `time.time()` — would slip through: the only sharded test (`test_sharded_leaf_write_counted`) runs `profile=True`, where `time.time()` is expected to work. Consider a sharded variant of the trap test (drive `_sharded_accumulate_fake` with `profile=False` under the `_Boom` rebind) so all three brackets are covered. Low severity — the guard is present and correct; this is a test-coverage gap, not a live defect.

---

espg (commented):
🤖 *from Claude (review)*

Phase-3 diff reviewed (8a0134d): the write-phase split is correct and additive — timing brackets are placed correctly (template emission included via `_leaf()` on both streaming and sharded paths), the attach gate matches the flat handler, and the hive branch cannot double-attach or lose `write` (the handler's own gate is skipped there because `store_box` stays empty). One low-severity test-coverage gap: the zero-overhead trap doesn't exercise the sharded post-stream bracket.

---

src/zagg/hive.py
espg: 🤖 *from Claude*

Added `test_sharded_default_path_makes_no_timing_calls`: the `hive.time` = `_Boom` rebind now also drives the K>1 sharded path (`_sharded_accumulate_fake`, `profile` absent), so the post-stream `write_leaf_to_zarr` bracket — and the sharded finalize — are covered by the zero-call proof alongside the streaming trap; the stamped-leaf assertion keeps it non-vacuous. Fixed in bfb596f.

---

espg:
🤖 *from Claude*

Final state at bfb596f: phase 3 lands the [approved](https://github.com/englacial/zagg/pull/256#issuecomment-5006361723) point-path `write` split — `process_and_write_hive` (the shared hive worker path both backends call, and the per-merge benchmark configuration) now accumulates the leaf write-out into an additive `phase_timings["write"]` next to `read`/`index`/`aggregate`, behind the existing opt-in `profile` gate with zero timing calls when off. The display-side `agg = index + aggregate` mapping is harness work riding issue #250 — per the [sequencing note posted there](https://github.com/englacial/zagg/issues/250#issuecomment-5006717243), this PR merges first, then the #250 harness work lands against the merged emission schema.

The phase-3 adversarial review's finding was folded in bfb596f with the thread answered; no ruff-bot comments on the PR. One boundary choice flagged for review as Questions item (6): the hive `write` bucket includes lazy leaf templating (the worker owns its leaf's template PUTs, unlike the dispatcher-owned flat template).

CI green on bfb596f: ruff, Lambda builds (x86_64 + arm64), and the 3.12/3.13 test matrix. Local: 2164 passed / 30 skipped after phase 3 (baseline 2158/30); `ruff format --check` clean; `ruff check` carries only the pre-existing N818 flagged in the PR body.

---

PR #257, opened 2026-07-16
Flip hive default for HEALPix + deprecate flat/fullsphere; remove dense layout

### LINKED ISSUE #88: Remove deprecated HEALPix `dense` layout (fullsphere-only)
🤖 *from Claude*

The HEALPix `dense` layout is deprecated at the config boundary but not removed, leaving the code half-migrated. This tracks finishing the migration to **fullsphere-only**.

## Current state
- **Config path is already fullsphere:** `grids/__init__.py::from_config` defaults to `fullsphere` and emits a `DeprecationWarning` on `layout: dense` (lines ~35–43). Every shipped config (and both new order-19 templates) runs fullsphere.
- **But dense is still in the code:**
  - `HealpixGrid.__init__` still **defaults to `layout="dense"`** (`grids/healpix.py:63`), so a direct `HealpixGrid(...)` comes up dense.
  - ~15 dense branches in `healpix.py` (the `_position_map`, `set_populated_shards`, `populated_shards`, and the dense arms of `n_shards` / `array_shape` / `block_index` / `_spec`).
  - `schema.py` exposes a dense-pack back-compat wrapper: `xdggs_zarr_template(n_parent_cells=...)` (its own `DeprecationWarning`).
  - `runner.py` still threads `populated_shards` into `from_config` (≈ lines 381, 501) — unused for HEALPix once dense is gone.
  - Tests still assert/exercise dense: `test_default_layout_is_dense`, `test_dense_needs_populated_shards_for_n`, `test_dense_uses_position_map`, `test_dense_requires_populated_shards`, `test_dense_preserves_insertion_order`, and the dense `emit_template` shape test (`tests/test_grids.py`).

## Proposed work
1. Flip `HealpixGrid.__init__` default `layout="dense"` → `"fullsphere"`; drop the `dense` literal (or hard-error on it).
2. Remove the dense branches + the `populated_shards` / `set_populated_shards` / `_position_map` plumbing from `healpix.py`.
3. `from_config`: drop the deprecation shim and the dense/`populated_shards` handling.
4. `schema.py`: remove (or hard-error) the `n_parent_cells` dense-pack wrapper.
5. `runner.py`: drop the now-unused `populated_shards` argument to `from_config` for HEALPix.
6. Tests: retire the dense tests; update any caller passing `layout="dense"` / `populated_shards`.

## Notes
- No on-disk format change for current outputs — every shipped config already writes the fullsphere layout; this only removes the dormant alternative path.
- Behavior change for any external caller that constructs `HealpixGrid(layout="dense", ...)` directly — worth a one-line changelog note.

_Tracking only (labelled `enhancement`); add `implement` when you want it picked up._

### LINKED ISSUE #247: Raster hive write path: windowed (time, cells) leaves, per-leaf templates and stamps, Lambda hive branch

### LINKED ISSUE #253: Phase 1 — flip hive default + deprecate HEALPix flat/fullsphere (no removal)
🤖 *from Claude*

Sub-issue of #251. The **ungated slice** — unblocked now that #236 (hive dense parity) + #240/#242 (hive regression arm) are done. **No flat code is removed here** (that's Phase 3, gated on the sparse-DGGS read path); this flips the default, deprecates flat/fullsphere with a warning, and repositions docs + benchmarks.

## Current state (grounded)
- `sharded` already defaults **True** for HEALPix (`grids/__init__.py:67`, `get_sharded(config, default=True)` — #233).
- `store_layout` still defaults **flat** (`config.get_store_layout` → `... or "flat"`). **This is the flip that's pending.**
- A `dense` `DeprecationWarning` already exists in `from_config` (`grids/__init__.py:40-46`) — mirror it for flat/fullsphere.

## Steps

1. **Grid-aware default flip.** HEALPix + omitted `store_layout` → `"hive"`; rectilinear → `"flat"` (unchanged — `validate_config` already rejects hive on non-HEALPix, `config.py:353`). Net: an empty HEALPix config yields **hive + sharded**. Implement in `get_store_layout` (default on `grid.type`) or `from_config`.
2. **Deprecation warning** on an *explicit* HEALPix flat/fullsphere, mirroring the dense warning: *"store_layout: flat / fullsphere is deprecated for HEALPix; hive is the default. Flat remains for interop/debug and is removed once the sparse-DGGS read path lands (#251 Phase 3)."* Non-fatal (flat still works — ratified as the interop/debug profile until Phase 3).
3. **Shipped configs** (`src/zagg/configs/*.yaml`, HEALPix): drop any explicit `store_layout: flat` so they inherit hive; keep + annotate any intentionally pinned flat. Verify none break.
4. **Docs / README:** reposition the quickstart to hive+sharded; leave a one-line "flat = interop/debug (deprecated; removal tracked in #251)" note.
5. **Benchmarks:** make hive the primary/headline arm; **keep the flat-vs-hive A/B** for the migration measurement (the CONUS flat-vs-hive run) — do **not** add new flat arms, and freeze the flat arms after the CONUS validation lands.
6. **Tests:** flip `test_default_layout_*` (HEALPix default → hive; rect stays flat); add "empty HEALPix config → hive+sharded"; assert the flat/fullsphere `DeprecationWarning` fires; confirm rect is untouched.

## Bundle with #88
#88 (remove deprecated `dense` layout) touches the same `from_config` / deprecation surface. **Bundle into the same PR** to avoid double-touching. Note the layering: #88 removes `dense` → leaves `fullsphere`; Phase 1 *deprecates* flat/fullsphere but does **not** remove it (Phase 3 does, gated).

## Explicitly NOT in Phase 1 (→ Phase 3, gated on the read path)
Removing `HealpixGrid.emit_template`/`_spec` fullsphere, `_require_fullsphere`, `xdggs_zarr_template`, the flat setup-Lambda mode, the `bench_objects` flat/fullsphere branch; renaming flat→rect; collapsing the `store_layout` knob. Rect carve-out untouched throughout.

## Acceptance
- Empty HEALPix config → `hive + sharded`; explicit flat/fullsphere warns (non-fatal); rect unchanged (still flat).
- Benchmark hive-primary, flat-vs-hive A/B retained, no new flat arms.
- README/quickstart on hive; full suite green (`ruff` + `pytest`).

## Sequencing
Phase 1 is unblocked (#236 closed). Phases 2–3 stay in #251, gated on the O3 sparse-DGGS read path.

espg:
Phase 4 (the post-#263 rework) is in: this PR now composes with the **merged** #263 raster-hive write path — the raster carve-out described below is gone, and raster inherits the same grid-keyed hive default.

---

Closes #253. Closes #88. (Bundled per the direction in #253 — see https://github.com/englacial/zagg/issues/253#issuecomment-4989930013 — both issues touch the same `from_config` / deprecation surface.)

## What this does

**#88 (phase 1):** finishes the dense→fullsphere migration by removing the deprecated HEALPix `dense` array layout outright. `HealpixGrid` now defaults to (and only accepts) `layout="fullsphere"`; the `populated_shards` / `set_populated_shards` / `_position_map` plumbing is gone; `from_config` drops the deprecation shim and the `populated_shards` kwarg; `validate_config` rejects `layout: dense` with a pointed error; `xdggs_zarr_template(n_parent_cells=...)` hard-errors instead of emitting the dense pack; the runner's dense branches (`_maybe_warn_dense`, the raster-lambda dense guard, the two dense `from_config` call sites, the setup event's dense `n_parent_cells` computation) are removed. Phase 4 re-applied this across the dense references main gained since this branch's base: the #264 raster-setup dense check keeps its clean 400 (rationale updated — `from_config` now raises, the guard prevents a retried 500), the #263 worker-side `from_config` ValueError→400 pattern stands, and the `populated_shards` kwarg was dropped from five #263 test call sites where it had become inert.

**#253 (phase 2 + phase 4):** the `store_layout` default is now **grid-keyed and pipeline-agnostic** (`get_store_layout`): HEALPix → `hive` — for point aggregation **and** the raster pipeline (raster + hive is the real write path since #263 closed #247); rectilinear → `flat` (unchanged). Net: an **empty HEALPix config yields hive + sharded**, and an unpinned healpix **raster** config now dispatches down #263's hive path on both dispatchers. An explicit HEALPix `store_layout: flat` or `output.grid.layout: fullsphere` emits a non-fatal `DeprecationWarning` in `from_config` — now **uniformly, raster included** (see Questions). The hive gates in the shared `_validate_store_layout_keys` (#263's refactor) resolve through the grid-aware default (defaulted hive rejects `consolidate_metadata`, accepts explicit `coverage_moc: true`), as does `_validate_windowing`. The old raster→flat default arm and the raster `store_layout: hive` rejection are **gone** (main's #239 stopgap was already removed by #263; the merge did not resurrect either). The flat-arm benchmark configs (`tests/data/benchmark/configs/atl03_*_healpix_*.yaml` and now `s2_neon_o9.yaml`) pin `store_layout: flat` explicitly.

**#253 (phase 3 + phase 4):** docs/README repositioned to the hive+sharded profile, re-merged with the raster-hive docs #263 added — README overview, `docs/quickstart.md`, `docs/hive_layout.md` (default for HEALPix output, point **and** raster; raster windowing keys on STAC `datetime`), `docs/deployment/benchmark.md` (stale "raster rejects hive (#239)" text replaced: the live raster leg now **pins flat**, the pending hive benchmark target is unblocked and awaits deliberate promotion), and `tests/data/benchmark/targets_raster_neon.json` comments updated to match.

## Phases

- [x] 1. Remove the dense layout (#88) — grid, config validation, schema wrapper, runner plumbing, lambda handler caller, tests (36c4857; review folds 9af3c0c, 68ad62c)
- [x] 2. Grid-aware `store_layout` default flip to hive + flat/fullsphere `DeprecationWarning` (#253 steps 1–2, 6) + flat pins in the flat-arm benchmark configs (8dffccc; review folds 24d9ed0, 984388d)
- [x] 3. Docs/README + benchmark-docs repositioning (#253 steps 3–5 remainder) (0a73f4e)
- [x] 4. Post-#263 rework: merge main (f06a359 — #255 lifecycle, #256 profiling/#250 benchmark restructure, #217 arena, #264 raster dispatch, #263 raster hive, #267/#269 deployment); drop the raster carve-out; raster inherits the grid-keyed hive default; uniform deprecation warning; new tests (b7eb257)

## Approach notes

- **Merge, not rebase (phase 4):** main moved from 6bf9728 through five landed features that rewrote the same surfaces twice more; f06a359 resolves them with main winning on every surface it gained (the #255 ping → async-setup → finalize-backstop lifecycle replaced this PR's setup-echo guard outright — the three echo-guard tests retired with it; the #250 benchmark collapse superseded this PR's 2×2 annotations) and this branch's #88 removal + default flip + warnings + docs carried through.
- **Wire compat (phase 1):** the flat setup event keeps `"n_parent_cells": None` so it stays byte-identical for deployed handlers; the handler ignores the field (covered by `test_setup_ignores_n_parent_cells`, which survived the merge against main's since-retired dense-threading test).
- **`deployment/aws/lambda_handler.py` touched minimally** — phase 1's `from_config` caller update plus, in phase 4, only a comment-rationale update on #264's dense-400 guard. The impact analysis on the issue thread called this coupling out (https://github.com/englacial/zagg/issues/88#issuecomment-4798632890), go-ahead in https://github.com/englacial/zagg/issues/88#issuecomment-4986248423. No deploy scripts or workflows were modified beyond what the merge brought in from main, and nothing was run against AWS.
- **Stale-deployment safety:** hive runs (defaulted or explicit) fail fast at the #255 ping against a pre-#252 deployment; flat runs keep the byte-identical pre-#252 setup path.
- **Explicit flat pins added to landed (#263/#264/#256) test surfaces** — main's raster tests were authored when an unpinned config meant flat; the flip changes only the default, so each fixture that *relied* on defaulted-flat now pins it (full list under Questions).
- **External-caller note (changelog-worthy):** direct `HealpixGrid(layout="dense", ...)`, `HealpixGrid(populated_shards=...)`, `from_config(populated_shards=...)`, and `xdggs_zarr_template(n_parent_cells=...)` now raise; an omitted `store_layout` on ANY HEALPix config (point aggregation or raster) now writes a **hive** store instead of flat.

## How it was tested

- Full suite after the phase-4 merge + rework: `uv run pytest -q` — **2323 passed, 32 skipped, 0 failed** (pre-merge baseline on this branch: 2155 passed / 30 skipped; main gained ~170 tests across the five merged features).
- `ruff check src tests` / `ruff format --check src tests`: clean except the pre-existing `N818` (below).
- New/updated tests (phase 4): defaulted healpix raster resolves hive + validates (`test_default_is_hive_for_raster`), explicit raster hive accepted (`test_explicit_hive_accepted_for_raster`), defaulted raster dispatches #263's hive path on the local dispatcher (`test_defaulted_layout_dispatches_hive`) and the lambda dispatcher (hive lifecycle + hive-shaped events, no injected `store_layout` key), uniform raster flat `DeprecationWarning` (`test_raster_explicit_flat_store_warns`) with silent defaults (`test_raster_defaults_do_not_warn`).

## Questions for review

1. **Pre-existing, not fixed (flagging per convention):** `ruff check src tests` fails on `main` too with `N818` (`src/zagg/registry.py` `UnknownCapability`); the PR lint bot selects only `E,F,W,I` so CI is unaffected.
2. **`n_parent_cells` on the wire:** kept as an inert `None` in the flat setup event for deployed-handler compatibility rather than dropping the key. Fine to drop entirely with the flat setup mode in #251 Phase 3?
3. `bench/layout_materialize.py` still exercises the dense pack via `xdggs_zarr_template(n_parent_cells=...)` and would now hard-error if rerun with `--layouts dense`. Left untouched — it documents the frozen `bench/REPORT.md` measurements and isn't in CI. Confirm that's the right call.
4. ~~**Raster + `store_layout: hive` now rejects**~~ **Superseded and done**: phase 4 dropped the carve-out; raster targets hive via the merged #263 path.
5. **`merra2_storm` (temporal pipeline):** `get_store_layout` on a temporal healpix config resolves hive, but the temporal path never consumes it — no behavior change observed (suite green). Confirm no temporal-path consumer is planned before #251 Phase 3.
6. **Coverage-write default flips with the layout:** an empty HEALPix config now writes a root `coverage.moc` by default (O9 hive default; `coverage_moc: false` opts out) — this now extends to **defaulted raster hive** runs too. Confirm intended.
7. **NEW — uniform raster deprecation warning (phase 4 design call, veto welcome):** the explicit-flat `DeprecationWarning` now fires for healpix **raster** configs too (this PR originally exempted raster). Rationale: flat-on-HEALPix is the deprecated interop profile *everywhere* on issue #251's trajectory, and the raster exemption existed only because raster hive didn't — it does now. Consequence: the pinned-flat live raster benchmark leg (`s2_neon_o9.yaml`) emits the warning at dispatch (non-fatal). If raster should stay warning-exempt until the benchmark target promotes, say so and phase 4 gets a one-line follow-up.
8. **NEW — every explicit-flat pin / adjustment made to landed test surfaces in phase 4** (each because the fixture relied on defaulted-flat raster, unless noted):
   - `tests/test_raster_runner.py` — `_cfg` fixture pins `store_layout: flat` (hive tests override per test); `test_flat_default_pin_no_hive_objects` renamed → `test_flat_pin_no_hive_objects` (its premise *was* the default); 3× `from_config(cfg, populated_shards=[shard])` → `from_config(cfg)` (#88, kwarg removed).
   - `tests/test_lambda_raster_handler.py` — `_config_dict` helper pins `store_layout: flat`.
   - `tests/test_raster_pipeline.py` — flat pins in `test_coverage_moc_rejected_on_flat`, `test_coverage_moc_defaults_on_for_hive` (flat arm), `test_windowing_requires_hive_layout`; 2× `HealpixGrid(..., populated_shards=[...])` → dropped (#88); `test_flat_template_unchanged_by_refactor` expectation `(2, 4096)` → `(2, 12 * 4**16)` (#88 fullsphere — shape is metadata, writes stay sparse; the shard-sized shape came from main's dense constructor default).
   - `tests/test_lambda_handler.py` — flat pin in #264's `test_flat_setup_body_pinned_no_raster_echo`.
   - `tests/data/benchmark/configs/s2_neon_o9.yaml` — pins `store_layout: flat` (the live raster benchmark leg stays flat until the `pending_targets` hive variant is deliberately promoted — promotion is NOT done in this PR; `targets_raster_neon.json` comments updated to match).
   - `tests/test_runner.py` — 8 flat pins in the aggregation-lambda drive helpers (TestMaxRetriesPassthrough, TestInvocationPassthrough, TestSummaryKeysByteIdentical ×4, `_run_lambda_with_durations`, TestConsolidationGate): these mocks stub the FLAT lifecycle (`_invoke_lambda_setup`/`_invoke_lambda_finalize`) and pre-#255 both layouts rode it, so post-merge a defaulted-hive run hit the unstubbed ping; pinning flat preserves what they were authored to test.
   - `tests/test_runner_concurrency.py` — same, 2 pins at the inline `_run_lambda(default_config("atl06"), ...)` call sites.
9. **NEW — merge composition judgment calls:** (a) this PR's three setup-echo-guard tests (`test_defaulted_hive_requires_echo` et al.) were **retired**, not ported — the #255 ping wholly replaced the echo guard they pinned; (b) this PR's `test_hive_event_carries_dataset` retired — #255 moved dataset threading to the async setup event, already pinned by main's `test_hive_async_event_matches_baseline`; (c) `docs/deployment/benchmark.md` + `targets_raster_neon.json` carried #239-era text that was stale on main itself — rewritten here since the flip changes what an unpinned raster config means. Flag if any of these should be handled differently.


deployment/aws/lambda_handler.py
```diff
@@ -882,13 +871,6 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
...
 
         config = load_config_from_dict(event["config"])
-        if get_layout(config) == "dense":
-            # Dense layout keys blocks by populated-shard position, which the
-            # worker cannot reconstruct from one shard's event; fullsphere (the
-            # default) keys by nested id and needs no shared state.
-            error_msg = "raster lambda workers require output.grid.layout: fullsphere"
-            logger.error(error_msg)
-            return {"statusCode": 400, "body": json.dumps({"error": error_msg})}
         grid = from_config(config)
```

[MED] Dropping the explicit dense guard here changes the worker's failure semantics from a client error to a server error. The old code returned a clean `400` with an actionable message:

```python
if get_layout(config) == "dense":
    error_msg = "raster lambda workers require output.grid.layout: fullsphere"
    return {"statusCode": 400, ...}
```

Now a stale dense config falls through to `grid = from_config(config)`, which raises `ValueError("Unknown layout: 'dense'")`; that is caught by the blanket `except Exception` at the bottom of `_handle_process_raster` and returned as `statusCode 500` (your updated `test_dense_layout_500` locks this in). Two consequences:

1. `load_config_from_dict` here does not run `validate_config` (that's why the surfaced message is the generic "Unknown layout", not config.py's "must be 'fullsphere'"), so there is no upstream 400 to catch it — the only guard left is the constructor exception → 500.
2. In an async/DLQ Lambda setup a `500` triggers automatic retries and alerting, whereas the old `400` failed fast and quietly.

In practice `validate_config` now rejects dense so no live dispatcher emits it, but for a mixed-version fleet with in-flight events this is a real regression. Consider calling `validate_config(config)` (or catching the layout `ValueError`) and returning `400` to preserve fail-fast client-error semantics — this also keeps parity with the setup handler, which deliberately keeps stale events non-fatal.

---

src/zagg/config.py
```diff
@@ -1911,7 +1914,8 @@ def get_layout(config: PipelineConfig) -> str:
     Returns
     -------
     str
-        ``"fullsphere"`` (default) or ``"dense"`` (deprecated).
+        ``"fullsphere"`` (the only layout; the deprecated ``"dense"`` pack
```
espg, line 2029: 🤖 *from Claude (review)*

[LOW] After this PR `get_layout` has zero callers anywhere in the tree — it was removed from `runner.py` (imports + `_maybe_warn_dense`/`_run_local`/`_run_lambda`) and from `lambda_handler.py::_handle_process_raster`, and a repo-wide grep now finds it only at its definition and the two `__init__.py` export lines. With `dense` gone and `validate_config` rejecting every non-`fullsphere` value, this helper can only ever return `"fullsphere"`, so it's effectively vestigial. Not a blocker and removing an exported public symbol is out of scope for #88, but worth a deliberate call: either keep it as a documented public config accessor (fine) or retire it alongside the dense plumbing it existed to feed. Leaving a note so the choice is explicit rather than accidental.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase 1 (commit `36c4857`, "phase 1 of issue #253"). This phase implements the #88 dense-layout removal (the bundled slice, per #253's layering note). The core removal is clean and well-scoped; a few findings below, plus two stale docstrings I couldn't attach inline because the file isn't in the diff.

**Verified good:**
- **Setup event stays byte-identical.** `_invoke_lambda_setup` still emits the `n_parent_cells` key, hardcoded to `None`. For any valid (fullsphere) config the previous value was already `None` (`len(all_shards) if ... layout == "dense" else None`), so the wire event is unchanged for deployed handlers. Correct.
- **Deployment scope respected (§1).** Only `deployment/aws/lambda_handler.py` changed under `deployment/` — no `template.yaml`, build/standup/mirror scripts. The handler edit is minimal (docstring + drop of the dense guard/populated plumbing).
- **#253 layering respected.** Flat/fullsphere machinery is untouched — `_spec`/`emit_template` fullsphere path, `xdggs_zarr_template` (kept, now hard-errors only on `n_parent_cells`), the flat setup-Lambda mode, the `bench_objects` flat branch, and the rect carve-out all remain. Only `dense` was removed.
- **No orphaned state in `runner.py`.** The removed `layout`/`grid_type` locals and `populated_shards` branches leave `all_shards` (still used in log lines) and `grid_type` (still used for `parent_order`/`child_order`) live — no dead vars / F841 risk.
- **Tests migrated, not weakened.** The retired dense tests were dense-specific; their fullsphere siblings remain, and `test_setup_ignores_n_parent_cells` / `test_raster_pipeline` were re-pointed to assert fullsphere sizing (the pipeline tests were previously exercising the constructor's `dense` default even though production used fullsphere — a fidelity improvement).

**Findings** (2 inline: MED on the raster handler 400→500 change, LOW on the now-unused `get_layout`), plus:

- **[LOW] `src/zagg/grids/base.py:422-423`** — the `OutputGrid` Protocol class docstring still describes removed machinery: *"Implementations may be stateless (fullsphere DGGS, rectilinear) or carry a populated-shard list (dense DGGS) so block_index can resolve."* No implementation carries a populated-shard list anymore. Trim to the stateless-only description.
- **[LOW] `src/zagg/grids/base.py:476-477`** — the `block_index` Protocol docstring still says *"For stateless layouts this is pure arithmetic on shard_key. For dense layouts it consults the populated-shard list built at template time."* The dense clause is dead after this PR. (Both are outside the diff, hence flagged here rather than inline.)

Nothing blocking the phase; the MED is the one I'd weigh before ready-for-review.

---

deployment/aws/lambda_handler.py

Fixed in 68ad62c. `_handle_process_raster` now wraps `from_config(config)` in a `try/except ValueError` and returns a clean `400` with the error message, restoring fail-fast client-error semantics for stale dense events (and any other invalid grid config) in a mixed-version fleet. Chose the catch over calling `validate_config(config)` in the worker: the full validator is heavier per-invoke and its rejection text targets config authors, while the constructor error names the actual failure. `test_dense_layout_500` was flipped back to `test_dense_layout_400` asserting the 400 + "Unknown layout" body.

---

src/zagg/config.py
espg: 🤖 *from Claude*

Leaving by design. `output.grid.layout` remains a valid (accepted, validated) config key — `get_layout` stays as its public accessor even though internal callers are gone, so external consumers reading the key keep a stable surface. #251 Phase 3 collapses the layout knob entirely (grid determines the store); the accessor retires there together with the key, rather than in two steps. Its docstring already records that `"fullsphere"` is the only value since the dense removal (issue #88).

---

espg:
🤖 *from Claude*

Phase-1 review pass folded (review: https://github.com/englacial/zagg/pull/257#pullrequestreview-4712914352):

- **[MED] raster worker 500 on stale dense config** — fixed in 68ad62c (clean 400 restored; reply on the inline thread).
- **[LOW] `get_layout` zero callers** — left by design (reply on the inline thread; retires with the knob collapse in #251 Phase 3).
- **[LOW]×2 (review summary) stale `OutputGrid` protocol docstrings in `src/zagg/grids/base.py`** describing the dense/populated-shard layout — fixed in 9af3c0c.

Process note: the review was posted by a fresh-context review subagent as usual; the fold was performed by the implementing agent in a clearly separated pass (the harness spawns fold subagents asynchronously only, which conflicts with exclusive access to the working tree), without relitigating any finding. Lint + `tests/test_lambda_raster_handler.py` green on the fold commits; the full suite runs again before the phase-2 push.

---

out.zarr/morton_hive.json
```diff
@@ -0,0 +1,17 @@
+{
```
espg: 🤖 *from Claude (review)*

[HIGH] Stray generated artifact committed. `out.zarr/morton_hive.json` is a hive manifest emitted by a local run (`"generated_at": "2026-07-16T11:07:26+00:00"`), newly tracked by this commit — `git log -- out.zarr` shows `8dffccc` as its only commit, and `out.zarr/` is not in `.gitignore`. This violates CLAUDE.md §3 ("Never commit ... generated artifacts. Respect .gitignore"). It also isn't part of phase 2's grid-aware-default work — it leaked in from the working tree. Please `git rm` it (and consider adding `out.zarr/` / `*.zarr/` to `.gitignore` so the next local run doesn't re-stage it).

---

src/zagg/runner.py
```diff
@@ -2952,7 +2920,14 @@ def _invoke_lambda_setup(
...
         raise RuntimeError(f"Lambda setup error: {result.get('body')}")
-    requested = ((config_dict or {}).get("output") or {}).get("store_layout") or "flat"
+    out = (config_dict or {}).get("output") or {}
+    requested = out.get("store_layout")
+    if requested is None:
+        # Mirror config.get_store_layout's grid-aware default (issue #253):
+        # HEALPix point aggregation -> hive; rectilinear / raster -> flat.
+        ds = (config_dict or {}).get("data_source") or {}
+        healpix = (out.get("grid") or {}).get("type", "healpix") == "healpix"
+        requested = "hive" if healpix and ds.get("reader") != "raster" else "flat"
```

[MED] Grid-aware default is hand-duplicated here, creating a silent drift risk. This block re-implements `config.get_store_layout`'s default (`healpix and reader != "raster" -> hive`) on a raw `config_dict`, and the comment even says "Mirror config.get_store_layout". If the canonical default in `get_store_layout` ever changes (e.g. the raster carve-out is refined, or an S2/healpix nuance is added), this copy won't follow and the echo-guard's notion of "hive" will diverge from what the deployed worker actually resolves — exactly the mixed-fleet failure this guard exists to prevent.

Note the deployment worker already avoids this: `deployment/aws/lambda_handler.py:627` calls `get_store_layout(load_config_from_dict(config_dict))`. The dispatcher has the same inputs and `load_config_from_dict` is a cheap non-validating constructor (`config.py:159`), so:

```python
from zagg.config import load_config_from_dict
requested = get_store_layout(load_config_from_dict(config_dict))
```

keeps the two in lockstep by construction. Also worth noting: `test_defaulted_hive_requires_echo` only covers the healpix->hive branch of this copy; the raster/rect->flat branch of the duplicated logic has no test, so a drift there would be silent.

---

tests/test_coverage_root.py
```diff
@@ -60,10 +60,19 @@ def test_default_on_for_hive(self):
...
         validate_config(cfg)
 
-    def test_default_off_for_flat(self):
+    def test_default_on_for_defaulted_hive(self):
+        # Issue #253: an empty HEALPix config now defaults to hive, so the
+        # coverage default follows (O9: coverage is the hive default).
         from zagg.config import get_coverage_moc, validate_config
 
         cfg = self._cfg()
+        assert get_coverage_moc(cfg) is True
```
espg, line 69: 🤖 *from Claude (review)*

[LOW] Downstream side effect worth an explicit callout for @espg. Because `get_coverage_moc` defaults on for hive (`config.py:2032`, `get_store_layout(config) == "hive"`), flipping the HEALPix default to hive means **every empty HEALPix config now writes a root `coverage.moc` by default** — previously (flat default) it did not. This test codifies it as intended (O9), and it's a reasonable consequence of "hive is the ratified profile," but it's a behavior change beyond issue #253's stated steps 1/2/6 (which only name hive+sharded), so it should be a conscious ratification rather than an incidental one. No code change requested if intended — just flag it in the PR body's "Questions for review" so the coverage-write default flip is visible alongside the layout flip.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **phase 2** (commit `8dffccc`) of issue #253 — the grid-aware store-layout default flip. Scope: config `get_store_layout` + validate gates, `grids/__init__.py` deprecation warnings, `runner._invoke_lambda_setup` echo-guard, benchmark flat pins, and test churn. Phase-1 hunks were out of scope.

The core mechanic is sound and well-tested: `get_store_layout` resolves HEALPix→hive / rect+raster→flat with the present-but-null fallback preserved; `validate_config`, `coverage_moc`, and `_validate_windowing` gates all correctly re-resolve through the default; the raster+hive rejection and the raster exemption in the deprecation path are right; the `sentinel2_l2a` (healpix+raster→flat) case is covered; the deprecation warnings fire on explicit flat/fullsphere and stay silent on defaults and on rect; and the flat pins on the benchmark configs correctly freeze the existing flat arm without adding new ones. No weakened tests — the flat pins are behavior-preserving, and `test_defaulted_hive_requires_echo` is a genuine addition guarding the mixed-fleet case. Callers were swept: `deployment/aws/lambda_handler.py` resolves through `get_store_layout`, so no stale `"flat"` default was missed.

Three findings, none blocking correctness of the happy path:

- **[HIGH]** `out.zarr/morton_hive.json` — a generated hive manifest leaked into this commit (not gitignored, unrelated to phase 2). Remove it and gitignore `out.zarr/`.
- **[MED]** `runner._invoke_lambda_setup` hand-duplicates the grid-aware default on a raw dict instead of reusing `get_store_layout(load_config_from_dict(...))` (which the deployment worker already does) — a silent drift risk, and the raster/rect branch of the copy is untested.
- **[LOW]** The hive flip also flips `coverage_moc` default → on for empty HEALPix configs (every default run now writes a root `coverage.moc`); intended per O9, but worth an explicit callout in the PR body since it's beyond issue #253's named steps.

Details inline.

---

out.zarr/morton_hive.json
espg: 🤖 *from Claude*

Fixed in 24d9ed0. `out.zarr/` removed from the tree (`git rm -r`) and added to `.gitignore` with a note (local test-run output), so the next local run can't re-stage it. It leaked from a working-tree test run swept in by a bulk stage — good catch.

---

src/zagg/runner.py

Fixed in 984388d. The hand-mirrored default is gone — the guard now resolves `requested = get_store_layout(load_config_from_dict(config_dict or {}))`, the same pattern as the deployed worker, so dispatcher and worker stay in lockstep by construction. Also added `test_defaulted_flat_grid_needs_no_echo` (rect config, echo-less body accepted) to cover the flat branch alongside the existing `test_defaulted_hive_requires_echo`.

---

tests/test_coverage_root.py
espg: 🤖 *from Claude*

Agreed — no code change; added as an explicit item under "Questions for review" in the PR body so the coverage-write default flip (empty HEALPix config now writes a root `coverage.moc`, per the O9 hive default) is ratified consciously alongside the layout flip rather than inherited silently. The behavior is pinned by `test_default_on_for_defaulted_hive`; an explicit `coverage_moc: false` opts out.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of **phase 3** (0a73f4e) plus the two phase-2 folds (24d9ed0, 984388d). Docs/benchmark repositioning for issue #253 steps 3–5. I verified every doc claim against the committed code; almost everything checks out. Findings below are all summary-level — the one line-anchored spot sits outside a diff hunk so GitHub won't take it inline.

**Verified accurate (no action needed):**
- **`get_store_layout` framing matches the docs.** `config.py:2011-2017` returns an explicit `store_layout` verbatim, else `hive` for `grid.type == "healpix"` with `data_source.reader != "raster"`, else `flat`. The hive_layout.md carve-out ("rectilinear grids and raster pipelines keep the flat shared store") and the README/quickstart rect note both match.
- **Step 3 claim holds.** `grep store_layout src/zagg/configs/*.yaml` returns nothing — no shipped config pins a layout, so HEALPix configs inherit hive with no edit. The "shipped configs needed no edits" claim is correct.
- **The four flat benchmark targets do pin `store_layout: flat`.** The live-2×2 targets resolve to `configs/atl03_tdigest_healpix_o9.yaml` (nomask) and `configs/atl03_tdigest_healpix_o9_aoimask.yaml` (mask); both pin `store_layout: flat`. So benchmark.md's "their configs pin `store_layout: flat` explicitly" and the targets.json annotation are grounded.
- **Both manifests are still valid JSON** (parsed clean) and no target semantics changed — the `_comment` edits are annotation-only; `config`/`shardmap`/`aggregator`/`parity_with` keys are untouched.
- **benchmark.md wording tracks issue #253 step 5:** "primary arm", "retained flat-vs-hive migration A/B", "No new flat arms", "flat arms freeze once the CONUS flat-vs-hive validation lands (issue #251)" all map to "keep the flat-vs-hive A/B / do not add new flat arms / freeze the flat arms after the CONUS validation lands."
- **Relative links resolve:** `docs/hive_layout.md`, `docs/sharding.md` exist; the quickstart's `hive_layout.md` / `sharding.md` and README's `docs/hive_layout.md` all point at real files; issue links (#251) well-formed.
- **Fold 24d9ed0** (.gitignore + `out.zarr/` removal): correct and minimal — adds `out.zarr/` under a clear comment, preserves the prior `bench/drift/*.json` line, and repairs the missing trailing newline. The leaked `out.zarr/morton_hive.json` artifact is deleted.
- **Fold 984388d** (echo guard via `get_store_layout(load_config_from_dict(...))`): behavior-preserving. The old inline branch computed the identical grid-aware default; `load_config_from_dict` is imported from `zagg.config` (already the source of `get_store_layout`), so no new import edge / cycle. The new `test_defaulted_flat_grid_needs_no_echo` exercises the **flat** branch (a rectilinear config resolves flat → guard skipped → echo-less body accepted), complementing the existing hive echo test.

**Findings:**

**[LOW] `docs/hive_layout.md:66` — stale `# default: flat` comment now contradicts the same file.** The Config example still reads `store_layout: hive                # default: flat`, but phase 3 rewrote the paragraph directly above it (lines ~12-18) to state hive **is** the default. A reader copying this block sees the inline comment assert the opposite of the prose two paragraphs up — materially misleading on the one page that documents the layout. (Outside the phase-3 diff hunk, so not inline-attachable; suggest updating the comment to `# HEALPix default` or dropping it.)

**[NIT] README.md:9 / `docs/quickstart.md` — the flat carve-out omits raster pipelines.** Both say flat "stays the layout for rectilinear grids" / "Rectilinear grids keep the flat shared store", but `get_store_layout` also defaults **raster** pipelines to flat (`config.py:2015`, and `docs/hive_layout.md` itself says "rectilinear grids and raster pipelines"). Not wrong, just less complete than the authoritative page; optional to align.

No correctness or JSON-validity issues found in the phase-3 diff or the two folds. Nothing blocking from this review.

---

espg:
🤖 *from Claude*

Final state at 5043601: 3 phase commits + 6 fold commits. Phase 1 (issue #88 dense-layout removal), phase 2 (issue #253 grid-aware hive default flip + flat/fullsphere `DeprecationWarning`), phase 3 (configs/docs/benchmark repositioning) are all in, with every adversarial-review finding folded — including the two from the phase-3 review (stale `# default: flat` comment in the `hive_layout.md` example fixed in 8d8518a; raster added to the flat carve-out note in 5043601).

Local full suite green (2155 passed / 30 skipped) and `ruff` CI green on the head commit; the 3.12/3.13 test matrix and Lambda builds were still running at the time of this comment — no known failures, and the same suite passed locally on every phase.

---

espg:
> raster pipelines → flat (the raster writer only targets the shared store — an explicit hive on a raster config is now rejected by _validate_raster_config instead of being silently ignored).

This seems wrong to me. Our driving use case for the raster pipeline is explicitly for fusing Sentinel-2 and ICESat-2 data; i.e., we want to address Sentinel-2 in hive format. Seems to be a confusion here conflating flat and dense layouts... the healpix / morton / hive format and layout needs to accommodate both sparse data and dense raster data (I'm using 'dense' here literally, not in the sense of our 'dense' API). The only difference is that for denser raster data, we'll want to skip the sharding codec.

The raster module / pipeline should work with *both* healpix / hive *and* rect; the rect / hive split is one abstraction above the raster abstraction.

This needs a rework before it can merge. Ask any clarifying questions that are needed to do this correctly.

espg:
🤖 *from Claude*

Understood on the direction ([espg's comment](https://github.com/englacial/zagg/pull/257#issuecomment-5005249126)): raster targets hive exactly like sparse aggregation — Sentinel-2 addressed in hive is the driving fusion use case — with the **only** raster-specific difference being that raster leaves skip the sharding codec. The layout choice (hive vs. rect/flat) sits one abstraction *above* the raster pipeline, so this PR's raster→flat default and its rejection of explicit `store_layout: hive` on raster configs (question 4 in the PR body) were the wrong call. That carve-out will be removed; the question below is only about **sequencing**, since the raster hive *write path* does not exist yet.

**Recommendation: option (b) below** — strip this PR of all raster-specific behavior and let #247's PR B bring raster hive as one coherent move. It's the smallest diff, creates no interim half-state, and unblocks #88 and the point-aggregation default flip immediately.

## The constraint

- The raster hive write path is designed (ratified on #237) and planned as **PR B** on #247 (per-leaf raster templates, windowed `(time, cells)` leaves, stamp/coverage, Lambda hive branch). espg answered the plan's questions on #247 [here](https://github.com/englacial/zagg/issues/247#issuecomment-5005587550), with two sub-questions still in discussion — but no implementation exists yet.
- Meanwhile, **after this PR branched** (base 6bf9728), the #239 stopgap landed directly on main (9e00a43, b99814c, 1d9008b, b3fe71e; #239 is now closed). `_validate_raster_config` on current main already rejects explicit raster `store_layout` as a truth-telling interim — src/zagg/config.py:670-673 at b3fe71e:

  ```python
  if store_layout == "hive":
      raise ValueError(
          "output.store_layout: hive is not supported on the raster path "
          ...  # points at #237
  ```

  (plus rejection of unknown `store_layout` values and non-bool/truthy `coverage_moc` on the raster path, config.py:667-682).

So the rework can't just flip raster's default to hive — until PR B lands, a raster+hive config has nothing that can write it, and this PR's raster validation now also collides textually and semantically with main's stopgap.

## Sequencing question — which option?

**(a) Rework this PR now to make the grid-keyed default fully pipeline-agnostic** (healpix raster → hive too). Raster+hive configs *validate*, but dispatch raises a clear `NotImplementedError` pointing at #247 until PR B lands. Truthful, no silent no-op — but it ships a config surface that accepts what it cannot yet write.

**(b) Strip this PR of all raster-specific behavior (recommended).** Rebase onto current main (b3fe71e) and leave the raster validation surface exactly as main's #239 stopgap has it — explicit hive still rejected with the pointer at #237. This PR shrinks to: #88 dense removal + the point-aggregation healpix→hive default flip + deprecation warnings + docs. #247's PR B then removes the stopgap and brings raster hive + the same grid-keyed default for raster in one coherent move.

**(c) Hold this PR and fold the default flip into the #247 PR B branch.** One PR, no interim state — but it couples the already-finished #88 dense removal to the raster timeline.

## Rebase needed regardless

Under every option this branch needs a rebase over main's b3fe71e: the base (6bf9728) predates the four #239 stopgap commits, and both this PR and the stopgap touch `_validate_raster_config`, so the conflict has to be resolved deliberately rather than by merge luck.

Applying `waiting` pending the option choice; the rework starts next run once one is picked.

---

espg:
...I think that #247 PR B has already landed as #248 ? does that change your recommendation-- i.e., is it easier to implement here?

espg:
🤖 *from Claude*

Close — but PR #248 was PR **A**, not PR B, so the recommendation is unchanged: **(b)**.

**What landed vs. what didn't.** PR #248 (merged 2026-07-16, closes #246) is the *temporal machinery*: `src/zagg/windows.py`, the `output.windowing` config block, the morton-hive/2 manifest temporal block, windowed leaf naming/stamps, coverage `"full"`, and per-(shard, window) dispatch — for the **aggregation** write path. It explicitly kept raster out: its config validation *rejects* raster+windowing with a pointer at #247 (its own PR body: "raster configs rejected pointing at #247"). PR **B** — the raster hive write path itself (per-leaf raster templates, the slab writer's stamp wiring, the Lambda `mode="process_raster"` hive branch, per-window raster fan-out) — is the five-phase plan still sitting on #247 ([plan](https://github.com/englacial/zagg/issues/247#issuecomment-4990713286), [your answers today](https://github.com/englacial/zagg/issues/247#issuecomment-5005587550)), with one design sub-question still open there (whether round one supports `schedule: none` for raster — trade space in [the reply](https://github.com/englacial/zagg/issues/247#issuecomment-5005931110)).

**Does PR A landing make it easier to implement here? Easier, yes — but not small.** Every primitive PR B consumes now exists on main (windowed `shard_leaf_path`, `stamp_commit(window=, time_range=)`, `build_coverage(full=)`, `windows_intersecting`, the `/2` manifest writer), which is exactly why the #247 plan is mechanical. But folding it in here would graft a five-phase implementation onto a PR that is already 9 commits deep about two different things (#88 dense removal + the point-aggregation default flip), while #247's remaining design question is still unanswered — that's option (c)'s coupling problem wearing option (a)'s clothes.

**So: still (b).** Strip the raster carve-out here, rebase over the #239 stopgap, land #88 + the aggregation flip now; PR B proceeds on `claude/247-raster-hive` per its plan the moment the `schedule: none` question is settled, and brings raster hive + the grid-keyed raster default in one move. Confirm (b) (or override) and the rework lands next pass.

---

espg:
🤖 *from Claude*

**Sequencing settled (espg direction, in session):** the issue #247 implementation ("PR B", the raster hive write path) proceeds first — it is now fully ratified ([schedule-`none` confirm](https://github.com/englacial/zagg/issues/247#issuecomment-5007157978)) and its draft PR is being opened on `claude/247-raster-hive` from post-#255/#256 main. This PR waits for that implementation to close out #247 and is then reworked to fold with it (rebase over the landed raster-hive surface, drop this PR's raster carve-out, keep #88 dense removal + the aggregation default flip).

Applying `blocked` (dependency: the #247 implementation PR — number to be recorded in the body once it opens).

---

deployment/aws/lambda_handler.py
```diff
@@ -748,19 +749,9 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
...
-            else None
-        )
-        grid = from_config(
-            config,
-            parent_order=event.get("parent_order"),
-            populated_shards=populated,
-        )
+        # chunk_inner + layout from the config. The event's n_parent_cells is
+        # ignored — the dense layout it selected was removed (issue #88).
+        grid = from_config(config, parent_order=event.get("parent_order"))
```
espg, line 754: 🤖 *from Claude (review)*

[LOW] Asymmetric stale-`dense` handling in `_handle_setup`: only the raster branch clean-400s a removed-`dense` config; the point-path (hive and flat) branches let `from_config` raise into the outer `except`, which returns a **500** (retryable-shaped), not a 400.

The raster branch right above (lines 708-720) keeps its guard precisely so a stale `layout: dense` event on this invoke-only path does not surface as a retried 500:
```python
if get_layout(config) == "dense":
    ... return {"statusCode": 400, ...}   # "reject the stale deployed event with the worker's clean 400"
```
But a `dense` point-aggregation setup event hits either the hive branch (line 690) or this flat branch (line 754), both calling `from_config(config, ...)` unguarded → `ValueError("Unknown layout: 'dense'")` → outer `except` at line 760 → `{"statusCode": 500}`. So the exact scenario the raster guard was written to avoid still 500s on the point path.

This is genuinely narrow (it needs a pre-#88 dispatcher paired with a post-#88 handler, and dispatcher+handler ship in the same build, so there is essentially no real window), which is why I'm marking it LOW rather than something to block on. Flagging only for parity: if the raster defense-in-depth guard is worth keeping, a one-line equivalent on the point path (e.g. a shared early `if get_layout(config) == "dense": return 400` before the branch split) would make the whole handler consistent. Fine to leave as-is by design — the guard here is theoretical either way.

---

espg (commented):
🤖 *from Claude (review)*

Overall: the phase-4 rework is clean and faithful to the ratified directives. Verdict is COMMENT with **1 LOW** finding and no HIGH/MED — the one item is narrow defense-in-depth parity, not a correctness bug.

Resurrection sweep — clean: no `populated_shards`/`set_populated_shards`/`_position_map` plumbing survives in `src`/`deployment` (only a `# no populated_shards` comment), no raster→flat default arm, no raster `store_layout: hive` rejection, and no pre-#255 setup-echo guard. `dense` hits are all either legit "dense array" (non-ragged) usage or #88-removal comments/errors. The `n_parent_cells` wire field is correctly retained as an inert `None` (setup event byte-compat) and hard-errors in `xdggs_zarr_template`.

Gate resolution — clean: `_validate_store_layout_keys` (hive + `consolidate_metadata` + `coverage_moc`), `_validate_windowing`, and `get_coverage_moc` all resolve through the grid-aware `get_store_layout(config)`, not the raw key; the raw-key read is correctly kept only for the two `from_config` deprecation warnings (explicit `store_layout: flat` / `grid.layout` on HEALPix), with a verified rectilinear carve-out.

Merge-damage check — clean: `from_config` on main already resolved raster to `fullsphere` (populated_shards was a no-op there), so the `_healpix_setup` test shape change `(2, 4096)` → `(2, 12·4**16)` *aligns* the test with real production output rather than masking a regression; the sparse writes are verified via `_shard_cell_range`. `_invoke_lambda_setup` signature drop has no stale callers.

Pins/new tests — spot-verified: each phase-4 flat pin is genuinely load-bearing (defaulted hive would otherwise change the exercised path / hit an unstubbed ping), and the new defaulted-hive raster dispatch + uniform-warning tests pin both dispatchers. Docs (README/quickstart/hive_layout/benchmark.md) match code behavior; the raster-windowing STAC-`datetime` claims match `_validate_windowing`. (Minor, inherited & intentionally unchanged: `targets_raster_neon.json` retains `blocked_by: "#237"` to keep the inherited tripwire test green while the rewritten `_comment` describes it as unblocked-awaiting-promotion — consistent enough, noted not flagged.)

---

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Fixed in b274c9d. The removed-`dense` rejection is now ONE shared early guard in `_handle_setup`, placed right after `load_config_from_dict` and before the store-layout branch split, so it covers the hive, raster, and flat paths alike; the duplicated raster-branch dense check (and its local `get_layout` import) is removed. Added a point-path 400 test, `test_setup_dense_layout_400` in `tests/test_lambda_handler.py`. The existing raster `test_setup_dense_layout_400` needed no change — it asserts `"fullsphere"` in the error, which the shared guard's message still contains.

---

PR #258, opened 2026-07-17
Pre-provisioned Lambda worker-size variants (issue #235)

### LINKED ISSUE #235: multiple lambda instance sizes available at runtime?

espg:
Closes #235

## What / approach

Implements option (A) from the issue thread — pre-provisioned Lambda worker-size variants selectable at run time by function name, with zero AWS-side config mutation — per the [ratified plan](https://github.com/englacial/zagg/issues/235#issuecomment-4990684782) and the [answers to its open questions](https://github.com/englacial/zagg/issues/235#issuecomment-5005614732): suffix `-disk` (not `-tmp`), unsuffixed `process-shard`/`-extract` untouched at 4096/512 as the no-config default (8 functions total after standup), `worker:` as a new top-level YAML section, and `Fn::ForEach` instead of hand-enumerating ~24 resources (explanation posted [on the thread](https://github.com/englacial/zagg/issues/235#issuecomment-5005956364)).

The variant matrix (6 new functions):

| name | memory (MB) | /tmp (MB) |
|---|---|---|
| `process-shard-2048` | 2048 | 512 |
| `process-shard-4096` | 4096 | 512 |
| `process-shard-8192` | 8192 | 512 |
| `process-shard-2048-disk` | 2048 | 4096 |
| `process-shard-4096-disk` | 4096 | 6144 |
| `process-shard-8192-disk` | 8192 | 10240 |

`-disk` /tmp is memory + 2048 MB; 10240 is exactly Lambda's EphemeralStorage ceiling, so the rule holds with no clamping.

`template.yaml` gains `Transform: AWS::LanguageExtensions` and two `Fn::ForEach` loops over a `WorkerMemorySizes` CommaDelimitedList parameter (the single source of truth for the size list), with a `WorkerDiskTmp` Mappings table for the `-disk` /tmp sizes. Each loop stamps, per size: the function (sharing `DepsLayer`, the execution role, timeout, and env with `ProcessFn`), its issue-151 `EventInvokeConfig` (retries 0 / event age 60), and the issue-175 self-recycle/worker-error metric-filter pair. `EphemeralStorage` appears only in the `-disk` loop. Outputs gain the 6 variant ARNs. `stand_up.sh` now passes `CAPABILITY_AUTO_EXPAND` (review fold — acknowledges the macro; harmless if the change-set path would have processed it anyway).

Selection: a new optional top-level `worker:` block —

```yaml
worker:
  memory: 2048       # one of 2048 | 4096 | 8192
  extra_disk: false  # true -> the -disk variant
```

— resolved by one runner helper (`_resolve_function_name`): explicit `agg(function_name=...)` wins verbatim; otherwise base from `ZAGG_LAMBDA_FUNCTION_NAME` (default `process-shard`) + the config suffix; no block → byte-identical current behavior. The CLI's `--function-name` default moved from eager env-or-`process-shard` to `None` so an unset flag no longer reaches `agg` as an explicit kwarg and silently masks the config selection; an explicitly passed flag still wins verbatim.

## Phases

- [x] Phase 1 — `deployment/aws/template.yaml`: transform + ForEach variants, per-variant EventInvokeConfig + metric filters, `-disk` EphemeralStorage, extended `CreateLogMetricFilters` caveat (8 log groups), variant ARN outputs; expansion pinned by pure-Python render tests in `tests/test_lambda_build.py` (+ 3 review-fold commits incl. `CAPABILITY_AUTO_EXPAND` in `stand_up.sh`)
- [x] Phase 2 — `src/zagg/config.py`: validate the optional top-level `worker:` block (memory in {2048, 4096, 8192}, `extra_disk` bool), fail-at-load errors naming the allowed set; 13 tests in `tests/test_config.py`
- [x] Phase 3 — `src/zagg/runner.py` + `src/zagg/__main__.py`: `_resolve_function_name(config, function_name)` replacing the three inline env fallbacks (spatial/raster/temporal paths); CLI `--function-name` default becomes `None`; resolver unit tests + dispatch-seam tests + CLI tests in `tests/test_runner.py`
- [x] Phase 4 — docs: variant table + `worker:` selection section + cost caveat (halving memory halves vCPU) in `docs/deployment/lambda.md`, stack-contents table in `docs/deployment/standup.md`, `worker:` in `docs/quickstart.md`'s optional-fields example

## How tested

- `tests/test_lambda_build.py::TestWorkerSizeVariants` renders the `Fn::ForEach` expansion in pure Python (textual `${Identifier}` + `!Ref` substitution over the parameter-default size list) and asserts: the 6 concrete names/memories//tmp sizes, lockstep with `ProcessFn` (same code/layer/role/timeout/env — mirroring the existing `ExtractFn` lockstep test), per-variant async-config and metric-filter parity, metric-name uniqueness (16 filters across 8 functions), and the size-list/mapping sync.
- `tests/test_config.py::TestWorkerBlock`: accept/reject matrix for the `worker:` block incl. YAML top-level load, dict roundtrip, and temporal-pipeline validation.
- `tests/test_runner.py`: `TestResolveFunctionName` (env/config/kwarg combinations), `TestLambdaDispatchFunctionSelection` (stubbed `_run_lambda` at the dispatch seam asserting the resolved name handed to the transport, incl. the no-block byte-identical regression), `TestCliFunctionNameDefault` (argparse default `None`, explicit flag verbatim).
- Full suite + `ruff check` / `ruff format --check` green locally (phase status comments have the numbers). Pre-existing `N818` in `src/zagg/registry.py` fails plain `ruff check src tests` on `main` too — not touched here.
- No live-AWS validation by design — template is parse/render tested only; the first real standup (which settles the `CAPABILITY_AUTO_EXPAND` question live) is a human action.

## Questions for review

1. **Config key rename from the plan:** the plan proposed `extra_tmp`; since the ratified *function suffix* is `-disk`, the config key here is `extra_disk` so the YAML key and the function name it selects agree. Flagging the divergence from the plan text for sign-off.
2. **ForEach grep-ability tradeoff:** `grep process-shard-8192-disk` finds the loop, not a literal resource block. Mitigated by assembling the name strings only in each loop's `FunctionName: !Sub` line and pinning the concrete expansion in `TestWorkerSizeVariants`; if literal blocks are preferred after seeing the diff, say so and phase 1 can be redone hand-enumerated.
3. **Metric names:** variant metrics are `Worker<size>[Disk]{SelfRecycle,WorkerError}Count` in the existing `zagg/lambda` namespace — 12 new metric names. Any dashboard/alarm naming convention this should match instead?
4. **Stale doc default corrected in passing:** `docs/deployment/standup.md` said `ProcessFn` defaults to 2048 MB; the template default has been 4096 since issue #180 — fixed while adding the variants table (flagging since it predates this PR).

### DIFF
--- a/deployment/aws/template.yaml
+++ b/deployment/aws/template.yaml
@@ -1,4 +1,12 @@
 AWSTemplateFormatVersion: "2010-09-09"
+# The AWS::LanguageExtensions macro provides Fn::ForEach, which stamps the
+# issue #235 worker-size variants from one parameterized block per resource
+# family instead of ~24 hand-written near-identical resources. The macro runs
+# server-side at change-set creation (stand_up.sh deploys via change sets, so
+# no mechanical change there); the expansion is deterministic and visible in
+# the change set, or post-deploy via
+# `aws cloudformation get-template --template-stage Processed`.
+Transform: AWS::LanguageExtensions
 Description: >-
   zagg serverless backend — Lambda execution role, dependency layer, and the
   process-shard Lambda function. By default the stack creates its own execution
@@ -145,7 +153,30 @@ Parameters:
       creates /aws/lambda/<fn> log groups lazily on first invocation and
       AWS::Logs::MetricFilter requires the group to exist, so on a FRESH
       stack: create with "false", invoke each function once, then update
-      the stack with "true". Existing deployments keep the default.
+      the stack with "true". This covers all 8 log groups — the base
+      function, its -extract twin, and the 6 worker-size variants
+      (issue #235). Existing deployments keep the default.
+
+  WorkerMemorySizes:
+    Type: CommaDelimitedList
+    Default: "2048,4096,8192"
+    Description: >-
+      Memory sizes (MB) of the pre-provisioned worker-size variants
+      (issue #235). Each size S stamps two functions via Fn::ForEach:
+      <FunctionName>-S (default 512 MB /tmp) and <FunctionName>-S-disk
+      (/tmp from the WorkerDiskTmp mapping — keep that table in sync with
+      this list). The unsuffixed function and its -extract twin are
+      independent of this list and stay the no-config default.
+
+Mappings:
+  # /tmp for the -disk variants is memory + 2048 MB (issue #235 thread
+  # decision); 10240 for the 8192 variant is exactly Lambda's
+  # EphemeralStorage ceiling, so the rule holds with no clamping. Keys must
+  # cover every WorkerMemorySizes entry.
+  WorkerDiskTmp:
+    "2048": {SizeMb: 4096}
+    "4096": {SizeMb: 6144}
+    "8192": {SizeMb: 10240}
 
 Conditions:
   ShouldCreateBucket: !Equals [!Ref CreateOutputBucket, "true"]
@@ -366,6 +397,139 @@ Resources:
           MetricValue: "1"
           DefaultValue: 0
 
+  # Pre-provisioned worker-size variants (issue #235): same code zip, layer,
+  # role, timeout, and env as ProcessFn — selected at run time by function
+  # name (config `worker:` block / agg(function_name=...)), so per-run sizing
+  # needs no admin-role config mutation. Two Fn::ForEach loops over
+  # WorkerMemorySizes stamp, per size S: <FunctionName>-S (this loop, default
+  # 512 MB /tmp) and <FunctionName>-S-disk (next loop, /tmp from
+  # WorkerDiskTmp). Each variant carries the issue #151 EventInvokeConfig and
+  # the issue #175 self-recycle/worker-error metric-filter pair, mirroring
+  # the unsuffixed function's blocks above. The variant FunctionName strings
+  # are assembled ONLY in each loop's `FunctionName: !Sub` line.
+  'Fn::ForEach::WorkerSizeVariants':
+    - WorkerMemory
+    - !Ref WorkerMemorySizes
+    - 'WorkerFn${WorkerMemory}':
+        Type: AWS::Lambda::Function
+        Properties:
+          FunctionName: !Sub "${FunctionName}-${WorkerMemory}"
+          Handler: lambda_handler.lambda_handler
+          Runtime: python3.12
+          Architectures: [!Ref Architecture]
+          MemorySize: !Ref WorkerMemory
+          Timeout: !Ref Timeout
+          Role: !If [ShouldCreateRole, !GetAtt ExecutionRole.Arn, !Ref ExecutionRoleArn]
+          Layers: [!Ref DepsLayer]
+          Environment:
+            Variables:
+              MALLOC_ARENA_MAX: !Ref MallocArenaMax
+              MALLOC_TRIM_THRESHOLD_: !Ref MallocTrimThreshold
+              ZAGG_RECYCLE_RSS_MB: !Ref RecycleRssMb
+              ZAGG_RECYCLE_MAX_INVOCATIONS: !Ref RecycleMaxInvocations
+          Code:
+            S3Bucket: !Ref ArtifactBucket
+            S3Key: !Ref FunctionS3Key
+      'WorkerFn${WorkerMemory}AsyncConfig':
+        Type: AWS::Lambda::EventInvokeConfig
+        Properties:
+          FunctionName: !Ref 'WorkerFn${WorkerMemory}'
+          Qualifier: "$LATEST"
+          MaximumRetryAttempts: 0
+          MaximumEventAgeInSeconds: 60
+      'WorkerFn${WorkerMemory}SelfRecycleFilter':
+        Type: AWS::Logs::MetricFilter
+        Condition: ShouldCreateMetricFilters
+        DependsOn: 'WorkerFn${WorkerMemory}'
+        Properties:
+          LogGroupName: !Sub "/aws/lambda/${FunctionName}-${WorkerMemory}"
+          FilterPattern: '"ZAGG_SELF_RECYCLE"'
+          MetricTransformations:
+            - MetricNamespace: zagg/lambda
+              MetricName: 'Worker${WorkerMemory}SelfRecycleCount'
+              MetricValue: "1"
+              DefaultValue: 0
+      'WorkerFn${WorkerMemory}WorkerErrorFilter':
+        Type: AWS::Logs::MetricFilter
+        Condition: ShouldCreateMetricFilters
+        DependsOn: 'WorkerFn${WorkerMemory}'
+        Properties:
+          LogGroupName: !Sub "/aws/lambda/${FunctionName}-${WorkerMemory}"
+          FilterPattern: >-
+            ?"Task timed out" ?"Runtime.OutOfMemory"
+            ?"Runtime exited with error" ?"[ERROR]"
+            ?"Traceback (most recent call last)"
+          MetricTransformations:
+            - MetricNamespace: zagg/lambda
+              MetricName: 'Worker${WorkerMemory}WorkerErrorCount'
+              MetricValue: "1"
+              DefaultValue: 0
+
+  # The -disk trio (issue #235 / #217): identical to the loop above except
+  # for the -disk name suffix and an explicit EphemeralStorage from the
+  # WorkerDiskTmp mapping (memory + 2048 MB). EphemeralStorage appears ONLY
+  # here — the default-512 trio omits it (512 is free; extra /tmp bills per
+  # GB-s above 512, idle functions cost nothing either way).
+  'Fn::ForEach::WorkerDiskVariants':
+    - WorkerMemory
+    - !Ref WorkerMemorySizes
+    - 'WorkerFn${WorkerMemory}Disk':
+        Type: AWS::Lambda::Function
+        Properties:
+          FunctionName: !Sub "${FunctionName}-${WorkerMemory}-disk"
+          Handler: lambda_handler.lambda_handler
+          Runtime: python3.12
+          Architectures: [!Ref Architecture]
+          MemorySize: !Ref WorkerMemory
+          Timeout: !Ref Timeout
+          Role: !If [ShouldCreateRole, !GetAtt ExecutionRole.Arn, !Ref ExecutionRoleArn]
+          Layers: [!Ref DepsLayer]
+          EphemeralStorage:
+            Size: !FindInMap [WorkerDiskTmp, !Ref WorkerMemory, SizeMb]
+          Environment:
+            Variables:
+              MALLOC_ARENA_MAX: !Ref MallocArenaMax
+              MALLOC_TRIM_THRESHOLD_: !Ref MallocTrimThreshold
+              ZAGG_RECYCLE_RSS_MB: !Ref RecycleRssMb
+              ZAGG_RECYCLE_MAX_INVOCATIONS: !Ref RecycleMaxInvocations
+          Code:
+            S3Bucket: !Ref ArtifactBucket
+            S3Key: !Ref FunctionS3Key
+      'WorkerFn${WorkerMemory}DiskAsyncConfig':
+        Type: AWS::Lambda::EventInvokeConfig
+        Properties:
+          FunctionName: !Ref 'WorkerFn${WorkerMemory}Disk'
+          Qualifier: "$LATEST"
+          MaximumRetryAttempts: 0
+          MaximumEventAgeInSeconds: 60
+      'WorkerFn${WorkerMemory}DiskSelfRecycleFilter':
+        Type: AWS::Logs::MetricFilter
+        Condition: ShouldCreateMetricFilters
+        DependsOn: 'WorkerFn${WorkerMemory}Disk'
+        Properties:
+          LogGroupName: !Sub "/aws/lambda/${FunctionName}-${WorkerMemory}-disk"
+          FilterPattern: '"ZAGG_SELF_RECYCLE"'
+          MetricTransformations:
+            - MetricNamespace: zagg/lambda
+              MetricName: 'Worker${WorkerMemory}DiskSelfRecycleCount'
+              MetricValue: "1"
+              DefaultValue: 0
+      'WorkerFn${WorkerMemory}DiskWorkerErrorFilter':
+        Type: AWS::Logs::MetricFilter
+        Condition: ShouldCreateMetricFilters
+        DependsOn: 'WorkerFn${WorkerMemory}Disk'
+        Properties:
+          LogGroupName: !Sub "/aws/lambda/${FunctionName}-${WorkerMemory}-disk"
+          FilterPattern: >-
+            ?"Task timed out" ?"Runtime.OutOfMemory"
+            ?"Runtime exited with error" ?"[ERROR]"
+            ?"Traceback (most recent call last)"
+          MetricTransformations:
+            - MetricNamespace: zagg/lambda
+              MetricName: 'Worker${WorkerMemory}DiskWorkerErrorCount'
+              MetricValue: "1"
+              DefaultValue: 0
+
 Outputs:
   FunctionArn:
     Description: ARN of the deployed Lambda function.
@@ -385,3 +549,12 @@ Outputs:
   OutputBucket:
     Description: Bucket where results are written.
     Value: !Ref OutputBucketName
+  'Fn::ForEach::WorkerVariantArns':
+    - WorkerMemory
+    - !Ref WorkerMemorySizes
+    - 'WorkerFn${WorkerMemory}Arn':
+        Description: 'ARN of the ${WorkerMemory} MB worker-size variant (issue #235).'
+        Value: !GetAtt 'WorkerFn${WorkerMemory}.Arn'
+      'WorkerFn${WorkerMemory}DiskArn':
+        Description: 'ARN of the ${WorkerMemory} MB extra-disk worker variant (issue #235).'
+        Value: !GetAtt 'WorkerFn${WorkerMemory}Disk.Arn'
--- a/deployment/aws/template.yaml
+++ b/deployment/aws/template.yaml
@@ -165,8 +165,11 @@ Parameters:
       (issue #235). Each size S stamps two functions via Fn::ForEach:
       <FunctionName>-S (default 512 MB /tmp) and <FunctionName>-S-disk
       (/tmp from the WorkerDiskTmp mapping — keep that table in sync with
-      this list). The unsuffixed function and its -extract twin are
-      independent of this list and stay the no-config default.
+      this list). This is not a free-form knob: any added size needs a
+      matching WorkerDiskTmp entry and must keep memory+2048 within Lambda's
+      10240 MB /tmp ceiling (so nothing above 8192). The unsuffixed function
+      and its -extract twin are independent of this list and stay the
+      no-config default.
 
 Mappings:
   # /tmp for the -disk variants is memory + 2048 MB (issue #235 thread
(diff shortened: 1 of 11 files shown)


deployment/aws/template.yaml
```diff
@@ -1,4 +1,12 @@
 AWSTemplateFormatVersion: "2010-09-09"
+# The AWS::LanguageExtensions macro provides Fn::ForEach, which stamps the
+# issue #235 worker-size variants from one parameterized block per resource
+# family instead of ~24 hand-written near-identical resources. The macro runs
+# server-side at change-set creation (stand_up.sh deploys via change sets, so
+# no mechanical change there); the expansion is deterministic and visible in
+# the change set, or post-deploy via
+# `aws cloudformation get-template --template-stage Processed`.
+Transform: AWS::LanguageExtensions
```
espg, line 9: 🤖 *from Claude (review)*

**Likely-breaking: the "no mechanical change to the deploy command" claim looks wrong — `stand_up.sh` almost certainly now needs `CAPABILITY_AUTO_EXPAND`, and it is not added.**

The header comment (and the issue #235 plan) assert that because `stand_up.sh` deploys via change sets, adding `Transform: AWS::LanguageExtensions` needs no change to the deploy path. But `stand_up.sh` line 190 passes only `--capabilities CAPABILITY_NAMED_IAM`, and `aws cloudformation deploy` does not stop at change-set *creation* — it also **executes** the change set (`create-change-set` → `execute-change-set`). Templates that reference a macro/transform require the caller to acknowledge `CAPABILITY_AUTO_EXPAND`; `AWS::LanguageExtensions` is in the same macro class as `AWS::Serverless`/`AWS::Include`, and `aws cloudformation deploy` on those consistently fails with `InsufficientCapabilitiesException: Requires capabilities: [CAPABILITY_AUTO_EXPAND]` unless the flag is passed. `CAPABILITY_NAMED_IAM` and `CAPABILITY_AUTO_EXPAND` are independent — one does not imply the other.

If that holds, the *first real standup* fails before creating anything. `stand_up.sh` is explicitly in scope here (issue #235 names it, and Phase 1 lists it), so the fix is small: add `CAPABILITY_AUTO_EXPAND` to the `--capabilities` line (`--capabilities CAPABILITY_NAMED_IAM CAPABILITY_AUTO_EXPAND`). Since this routine can't deploy to confirm, please verify against the target account's behavior — but the reassurance in this comment/PR that no deploy-path change is needed should not stand unchallenged. The two inspectability claims (change-set preview and `get-template --template-stage Processed`) are accurate; only the capability claim is at issue.

---

deployment/aws/template.yaml
```diff
@@ -366,6 +397,139 @@ Resources:
...
+    - WorkerMemory
+    - !Ref WorkerMemorySizes
+    - 'WorkerFn${WorkerMemory}':
+        Type: AWS::Lambda::Function
+        Properties:
+          FunctionName: !Sub "${FunctionName}-${WorkerMemory}"
+          Handler: lambda_handler.lambda_handler
+          Runtime: python3.12
+          Architectures: [!Ref Architecture]
+          MemorySize: !Ref WorkerMemory
```
espg, line 423: 🤖 *from Claude (review)*

Minor / deploy-time only: `WorkerMemory` iterates a `CommaDelimitedList`, so its values are **strings** (`"2048"`), and `MemorySize` here (and on the `-disk` twin, L483) is `AWS::Lambda::Function`'s integer property. CloudFormation normally coerces numeric strings for numeric properties, and Fn::ForEach-over-CommaDelimitedList driving numeric props is a common AWS-documented pattern, so this is very likely fine — flagging only because the parse-only tests encode `MemorySize == "2048"` (a string) and therefore cannot catch a coercion rejection. Worth a glance at the first change-set / processed template to confirm the concrete `MemorySize` renders as expected.

---

deployment/aws/template.yaml
```diff
@@ -145,7 +153,30 @@ Parameters:
...
+      (/tmp from the WorkerDiskTmp mapping — keep that table in sync with
+      this list). The unsuffixed function and its -extract twin are
+      independent of this list and stay the no-config default.
+
+Mappings:
+  # /tmp for the -disk variants is memory + 2048 MB (issue #235 thread
+  # decision); 10240 for the 8192 variant is exactly Lambda's
+  # EphemeralStorage ceiling, so the rule holds with no clamping. Keys must
+  # cover every WorkerMemorySizes entry.
+  WorkerDiskTmp:
```
espg, line 179: 🤖 *from Claude (review)*

Consistency note (by-design, but worth making explicit): `WorkerMemorySizes` is a tunable `Parameter`, yet two hard-coded assumptions make it not safely overridable beyond the ratified `2048,4096,8192`:
1. `WorkerDiskTmp` has literal keys for exactly those three sizes — any added size fails `!FindInMap` at deploy with a missing-key error;
2. the `-disk` rule is memory+2048 against Lambda's 10240 ceiling, so any size above 8192 yields `EphemeralStorage.Size > 10240`, which is invalid.

The parameter Description does warn to keep the mapping in sync, so this is acceptable for the default set. But since it reads as a free knob, consider either a one-line note that the list is effectively fixed to what `WorkerDiskTmp` covers (and capped at 8192), or leaving the sizes as a plain inline list rather than a Parameter. The `test_size_list_and_disk_mapping_stay_in_sync` guard only covers the *default*, not an override.

---

tests/test_lambda_build.py
```diff
@@ -382,6 +382,172 @@ def test_async_event_invoke_config_pins_retries(self):
...
+            return node
+
+        out = {}
+        for key, node in section.items():
+            if not key.startswith("Fn::ForEach::"):
+                out[key] = node
+                continue
+            ident, collection, fragment = node
+            if isinstance(collection, dict) and "Ref" in collection:
+                collection = tpl["Parameters"][collection["Ref"]]["Default"].split(",")
```
espg, line 424: 🤖 *from Claude (review)*

Test-faithfulness note (not a correctness bug): `_expand_foreach` resolves the collection from `Parameters[...]["Default"]`, so the whole suite only ever exercises the **default** size list. That's the right thing to pin the ratified matrix, and the pure-Python `_subst` model (`{"Ref": ident}` → value, textual `${ident}` replacement, dict-key recursion) faithfully mirrors the macro for the constructs used here (`!Sub`, `!Ref` to the loop var and to constructed logical ids, `!GetAtt` dotted-string, `!FindInMap` key). Just be aware the tests would not catch a template that expands incorrectly under a non-default `WorkerMemorySizes` override. Fine to leave as-is given the sizes are effectively fixed (see the template-side note), but the docstring's "renders those loops the way the transform does" slightly overstates scope — it renders them for the default collection only.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase 1 (commit 2e0fa8a) — `template.yaml` Fn::ForEach variants + `TestWorkerSizeVariants`. Overall this is close and faithful to the ratified issue #235 decisions; the Fn::ForEach mechanics are correct and the tests are a genuinely good parse-time model. One finding is potentially deploy-breaking and worth resolving before this goes ready; the rest are minor/by-design notes.

**What checks out (verified against the issue thread and the surrounding template):**
- 6 variants: `2048/4096/8192` each with a default-512 twin and a `-disk` twin; `-disk` `/tmp` = memory+2048 → `4096/6144/10240`, top exactly at the 10240 ceiling. `EphemeralStorage` appears only on the `-disk` trio.
- Unsuffixed `process-shard` and `-extract` untouched; `-disk` suffix as ratified.
- EventInvokeConfig parity (retries 0 / age 60) and the issue #175 self-recycle/worker-error metric-filter pair on every variant, patterns byte-identical to the unsuffixed blocks, per-function unique metric names (16 filters total). `CreateLogMetricFilters` caveat correctly updated to 8 log groups.
- Fn::ForEach usage is correct: `!Ref` to a `CommaDelimitedList` as the collection, `${WorkerMemory}` textual substitution in logical ids / `!Sub` (leaving `${FunctionName}` for Sub), `!Ref`/`!GetAtt`/`DependsOn` with constructed logical ids, `!FindInMap` with the substituted key. No logical-id collisions. `Condition`/`!If` survive inside the loop fragment. Both files stay well under the ~1000-line limit; nothing outside `template.yaml` + tests changed.

**Findings (all inline):**
1. **(high, verify)** `deployment/aws/template.yaml` L9 / `stand_up.sh` L190 — the "no mechanical change to the deploy command" claim is likely wrong. `aws cloudformation deploy` executes the change set, and a macro/transform template almost certainly requires `CAPABILITY_AUTO_EXPAND`, which the standup does not pass (only `CAPABILITY_NAMED_IAM`). If so, the first real standup fails with `InsufficientCapabilitiesException`. `stand_up.sh` is in scope per the issue; fix is one flag. Cannot deploy from here to confirm — please verify.
2. **(low)** L420/L483 — `MemorySize` gets a *string* from the CommaDelimitedList; almost certainly coerced by CFN, but the parse-only tests can't catch a rejection.
3. **(low, by-design)** L176 — `WorkerMemorySizes` reads as tunable but is effectively fixed by the literal `WorkerDiskTmp` keys and the memory+2048 ≤ 10240 rule; the in-sync test covers only the default.
4. **(low)** `tests/test_lambda_build.py` L423 — the expansion helper only ever exercises the default size list; faithful for that, but the docstring slightly overstates scope.

No blocking correctness bug in the expansion logic itself. Finding (1) is the one to settle before ready-for-review.

---

deployment/aws/template.yaml
espg: 🤖 *from Claude*

Folded in `861e760`. `stand_up.sh` now passes `--capabilities CAPABILITY_NAMED_IAM CAPABILITY_AUTO_EXPAND`, with a one-line comment noting it acknowledges the `AWS::LanguageExtensions` macro expansion and is harmless if the change-set path would have processed it anyway. The template header comment no longer claims "no mechanical change there" — it now states that `stand_up.sh` passes `CAPABILITY_AUTO_EXPAND` to acknowledge the macro. Since the two capabilities are independent, passing the extra one is a no-op if unneeded and removes the first-standup `InsufficientCapabilitiesException` risk if it is needed. This routine can't deploy, so live verification still rests with the first human standup.

---

deployment/aws/template.yaml
espg: 🤖 *from Claude*

Leaving by design. Every CloudFormation parameter value is a string at the API level — the existing `MemorySize: !Ref MemorySize` on `ProcessFn` is equally a string under the hood, as is any numeric parameter — and Lambda's `MemorySize` property schema coerces the numeric string. `Fn::ForEach` over a `CommaDelimitedList` driving numeric properties is a documented pattern, so this matches how the template already works. The parse-only test asserting `MemorySize == "2048"` reflects the pre-coercion template text, not a rejection risk. No commit.

---

deployment/aws/template.yaml
espg: 🤖 *from Claude*

Folded in `6635eb2`. Extended the `WorkerMemorySizes` parameter Description to make the coupling explicit: it is not a free-form knob — any added size needs a matching `WorkerDiskTmp` entry and must keep `memory+2048` within Lambda's 10240 MB /tmp ceiling (so nothing above 8192). That documents the single-source-of-truth constraint you flagged without converting it to a plain inline list (keeping it a Parameter preserves the override path for the ratified set and the existing `test_size_list_and_disk_mapping_stay_in_sync` guard on the default).

---

tests/test_lambda_build.py
espg: 🤖 *from Claude*

Folded in `29b94ab`. Tightened the docstring: it now says `_expand_foreach` models the transform for the constructs this template uses, with the collection taken from the parameter's `Default`, and that the tests pin the concrete function set for the default size list. That drops the "renders those loops the way the transform does" phrasing that overstated scope and makes explicit the tests only exercise the default collection.

---

src/zagg/runner.py
```diff
@@ -81,6 +81,30 @@ def _maybe_warn_dense(layout: str) -> None:
         warnings.warn(_DENSE_DEPRECATION_MSG, DeprecationWarning, stacklevel=3)
 
 
+def _resolve_function_name(config, function_name):
```

(phase 3, f0ab1d7) Low / convention. This helper carries no type annotations, while both neighbors in the same file annotate their return: `_maybe_warn_dense(layout: str) -> None` (L79) and `_resolve_source_credentials(config) -> dict` (L1507). For consistency, `def _resolve_function_name(config, function_name) -> str:` (return is always a `str`). Not a correctness issue — flagging only to match the surrounding style (CLAUDE.md §4).

---

src/zagg/runner.py
```diff
@@ -775,8 +801,7 @@ def _run_lambda_shards(
         import boto3
         from botocore.config import Config
 
-        if function_name is None:
-            function_name = os.environ.get("ZAGG_LAMBDA_FUNCTION_NAME", "process-shard")
+        function_name = _resolve_function_name(config, function_name)
```
espg, line 804: 🤖 *from Claude (review)*

(phase 3, f0ab1d7) Low / test coverage. Of the three call sites now routed through `_resolve_function_name`, only the spatial path is exercised at the dispatch seam (`TestLambdaDispatchFunctionSelection` stubs `_run_lambda`). The raster site here (`_run_lambda_shards`) and the events site (`_run_lambda_events`, L1028) rely solely on the unit `_resolve_function_name` tests plus the fact that all three are the identical one-liner. That's defensible given the shared helper, but a one-line assertion that each of these two paths hands the resolved name to its transport would close the loop and guard against a future edit dropping the resolver call on one path. Non-blocking.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the three phase commits (3d95899 config, f0ab1d7 runner/CLI, edf0c7d docs). Verified against the ratified thread decisions in issue #235 and `template.yaml`/`stand_up.sh` as they now stand on this branch. **This is clean, well-scoped, well-tested work — no correctness or convention blockers.** Two low-severity notes inline; the rest below is confirmation of what I checked so it doesn't get re-litigated.

**Phase 2 — `config.py` `_validate_worker`.** Correct on every edge I probed: `worker={}` → rejected (`memory` missing → not in the allowed set); `memory` as `bool` guarded explicitly before the set membership (`True`/`False` won't slip through as `1`); `memory` as `"4096"` str → rejected (suffix is built from a validated int, not `str()`); unknown key `extra_tmp` → rejected (the old plan spelling can't be silently ignored); non-bool `extra_disk` → rejected, while `extra_disk` absent or explicit `None` both pass and read as false, consistent with the resolver. Validation runs before the pipeline-kind branch, so temporal/event configs are covered too (tested). Mirrors the neighboring `_validate_*` blocks in style and placement. Tests are thorough.

**Phase 3 — `runner.py` / `__main__.py`.** All three inline env fallbacks are gone and replaced by `_resolve_function_name`; grep confirms no stray `ZAGG_LAMBDA_FUNCTION_NAME` fallback remains outside the resolver, and `config` is in scope at all three call sites (spatial L386, raster `_run_lambda_shards` L804, events `_run_lambda_events` L1028). Precedence is right: explicit kwarg wins verbatim, else env base + `worker` suffix, else bare base (byte-identical prior behavior). `os` import removed from `__main__.py` with no remaining `os.` use; `os` correctly retained in `runner.py` (still used at L98). The CLI `--function-name` default → `None` is the right fix for the masking bug — an explicit `--function-name` still passes through verbatim, so no existing script breaks. Minor robustness aside (not flagged inline): an empty `worker={}` on a *hand-built* config that bypasses `load_config` resolves to the base name via `if not worker` rather than erroring — but `validate_config` rejects `{}`, and this matches how the rest of the codebase treats unvalidated hand-built configs, so leave-by-design.

**Phase 4 — docs.** Accurate against the tree: the variant tables in `lambda.md` and `standup.md` match the `WorkerDiskTmp` mapping (2048→4096, 4096→6144, 8192→10240) and the `Fn::ForEach` naming; `standup.md`'s `CAPABILITY_AUTO_EXPAND` claim matches `stand_up.sh` L192 as it now stands (fold commit 861e760) — and that capability is in fact required for `aws cloudformation deploy` against a transform template, so the doc and the script agree and are correct; the `ProcessFn` "default 4096 MB" edit matches the `MemorySize` parameter default (L83). `quickstart.md`'s `worker:` snippet is valid YAML and consistent with the schema. One stale-text nit, doc-only and outside this diff: the PR body still says "`stand_up.sh` needs no change," which the 861e760 fold superseded — worth a one-line body edit for the reader, not a code concern.

Nothing requires a change to land; the two inline notes are style/coverage polish. Not approving or requesting changes — review-only per role.

---

src/zagg/runner.py

Folded in `8d1b841`. Annotated the signature to match the neighboring helpers: `def _resolve_function_name(config: PipelineConfig, function_name: str | None) -> str:` (`PipelineConfig` was already imported in `runner.py`). `ruff check` / `ruff format` clean; `pytest tests/test_runner.py` green (187 passed).

---

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded in `248b1eb`. Added `TestLambdaResolverSeamRasterEvents` in `tests/test_runner.py` covering **both** remaining paths at the dispatch seam. It monkeypatches `zagg.runner._resolve_function_name` with a stub that records the `(config, function_name)` it receives and raises a sentinel `RuntimeError` to halt before any transport is touched, then:

- `test_events_path_routes_through_resolver` drives `_run_lambda_events` (config with `output.format: parquet` to clear the tabular guard, one URI-shaped event, an `s3://…/out.parquet` store) and asserts the stub saw that config plus the passed `function_name`.
- `test_raster_path_routes_through_resolver` drives `RasterStrategy()._run_lambda_shards` with tiny dummies — the resolver call is the first statement after the boto3 imports and before `boto3.client`, so nothing else is exercised — and makes the same assertion.

Both now guard against a future edit dropping the resolver call on either path. `ruff check` / `ruff format` clean; `pytest tests/test_runner.py` green (187 passed, +2).

---

espg:
🤖 *from Claude*

Final state at 248b1eb: 4 phase commits + 5 fold commits. Phase 1 (template `Fn::ForEach` variants + stand_up capability + expansion-pinning tests), phase 2 (top-level `worker:` config validation), phase 3 (`_resolve_function_name` + CLI shadowing fix + dispatch-seam tests), phase 4 (docs) are all in.

Adversarial-review findings, all dispositioned: (1) `CAPABILITY_AUTO_EXPAND` missing from `stand_up.sh` — folded in 861e760; (2) `WorkerMemorySizes` parameter not freely overridable — description tightened in 6635eb2; (3) ForEach-model docstring overstated scope — tightened in 29b94ab; (4) missing return annotation on the resolver — folded in 8d1b841; (5) raster/events dispatch paths untested at the resolver seam — covered in 248b1eb; (6) `MemorySize` string-vs-int coercion — left by design with rationale on the thread. Every inline thread has a reply.

CI green on 248b1eb: ruff, Lambda builds (x86_64 + arm64), and the 3.12/3.13 test matrix. Locally each phase ran green before push (`ruff check` / `ruff format --check` clean; targeted suites green, e.g. `tests/test_runner.py` 187 passed on the final folds); the final full-suite rerun was interrupted by an infrastructure restart, so the green matrix above is the authoritative full-suite signal.

---

espg:
> Config key rename from the plan: the plan proposed extra_tmp; since the ratified function suffix is -disk, the config key here is extra_disk so the YAML key and the function name it selects agree. Flagging the divergence from the plan text for sign-off.

ok. also fine with all other questions as is

PR #260, opened 2026-07-17
Arena state layout for streaming digest state (issue #217)

espg:
Refs #217 (chunk-scoped digest streaming — option (C), arena state layout).

## What

Adds an opt-in `aggregation.streaming.state_layout: arena` that holds the `StreamingAggregator` running state as contiguous CSR buffers — one sorted `cell_ids` array, a parallel counts array, and per digest field one packed `(k, 2)` float32 centroid buffer addressed by offsets — instead of the current dicts of per-cell ndarrays (`counts` + `digests`, [streaming.py:150-151 pre-change](https://github.com/englacial/zagg/blob/b3fe71e/src/zagg/processing/streaming.py#L150-L151)). Default `dict` is byte-untouched.

**Why**: the dict layout pays a measured ~290 B/cell fixed tax (dict slots + boxed keys + ndarray headers) on top of the centroid payload — 302 MB of the 861 MB digest state on the 120-granule CONUS o8 plateau shard ([measured decomposition](https://github.com/englacial/zagg/issues/217#issuecomment-5006544129)). The arena replaces that with ~24 B/cell.

**How the fold stays identical**: `_fold_arena` makes the *same* `build_tdigest`/`merge_tdigests` calls in the same per-cell order as the dict branch, then rebuilds the CSR state at its exact final size (sizes are known before allocation because the merged digests are built first). Untouched cells move via a vectorized element gather (`_ranges_to_indices`); touched cells are written per-cell. Output is therefore byte-identical to the dict layout — asserted by the A/B tests — and the transients (one buffer's fresh digests + the old arena) are freed on return.

`chunk_outputs` under the arena is a single `searchsorted` + vectorized dense fill (the dict branch loops all children per chunk).

## Phases

- [x] Arena layout: config knob + validation, `_fold_arena` rebuild, vectorized `chunk_outputs`, tests (A/B byte-equality vs dict across multi-flush merges, cell insertion, all-NaN cells; pooled equivalence at single buffer; helper unit tests)
- [x] A/B memory/runtime replay of the pinned-shard distributions through both layouts — [verdict posted on #217](https://github.com/englacial/zagg/issues/217#issuecomment-5007662673): steady state −28% (605 vs 836 MB, matching the analytic model), but the rebuild-per-flush transient peaks ~2× the dict layout (3,254 vs 1,722 MB traced) — a net loss on the OOM-relevant metric as-is. The open fork (fleet A/B for glibc fragmentation vs in-place-slot iteration vs park) is on the issue.
- [x] `arena_backing: tmp` (per espg direction on the session thread): the three `_fold_arena` centroid buffers (staging, bounded merge target, compacted state) allocate as **unlinked `/tmp`-backed memmaps** — the file is deleted the instant the mapping exists, so nothing can leak across warm invokes, and the flush transient becomes file-backed page cache the kernel can write back and evict under cgroup pressure instead of OOM-killing anonymous heap. Opt-in (`arena_backing: memory` default), arena-only (validated), byte-identical output (tested). Fleet plan per espg: merge as a **merge commit** (easy revert), then A/B on the fleet with tmp backing on/off across a regular shard and the 148/155 g difficult shards.

## Testing

`uv run pytest tests/test_streaming.py -v` — 34 passed (7 new arena tests). `ruff check` / `ruff format --check` clean on the touched files.

## Questions for review

- The knob is deliberately temporary A/B surface: if the replay confirms the arena and espg adopts it, `state_layout` should collapse (arena becomes the only layout) per the #238-style knob-collapse direction rather than shipping two containers indefinitely.
- A `/tmp`-memmap-backed arena (the (A)+(C) hybrid discussed on #217) would be a follow-up on top of this: the arena allocation site is the single place a `np.memmap` would slot in.

### DIFF
--- a/src/zagg/processing/streaming.py
+++ b/src/zagg/processing/streaming.py
@@ -241,17 +241,43 @@ class StreamingAggregator:
         buffer is an **anonymous** file mapping: the file is unlinked the
         moment the mapping exists, so the mapping itself is the only
         reference — space frees on GC and nothing can accumulate in ``/tmp``
-        across warm Lambda invokes. The payoff is failure mode: dirty pages
-        are file-backed page cache the kernel can write back and reclaim
+        across warm Lambda invokes. The reclaim payoff holds only where
+        ``/tmp`` is genuinely disk-backed (Lambda ephemeral storage): dirty
+        pages are file-backed page cache the kernel can write back and reclaim
         under cgroup pressure, so an oversized flush transient degrades into
-        paging instead of an OOM kill of anonymous heap.
+        paging instead of an OOM kill of anonymous heap. Where ``/tmp`` is
+        tmpfs (common on dev/CI boxes) those pages are RAM that counts against
+        the same cgroup and cannot be evicted — no benefit, strictly extra
+        overhead.
+
+        ``mode="w+"`` grows the file with a single sparse seek+write, so the
+        allocation succeeds even when the filesystem cannot hold the full
+        buffer; the shortfall only surfaces later as an **uncatchable SIGBUS**
+        on the page-fault writes that fill the mapping. To turn that deferred
+        crash into a clean loud error we check free space against the request
+        up front and refuse if it won't fit. ``arena_backing: tmp`` therefore
+        requires ephemeral storage sized above the arena flush transient — up
+        to three of these buffers (old + staging + bounded merge) can be live
+        at once during a fold, so size for their combined peak, not one buffer.
         """
         if self.arena_backing != "tmp" or n_rows == 0:
             return np.empty((n_rows, 2), dtype=np.float32)
         import os
         import tempfile
 
-        fd, path = tempfile.mkstemp(suffix=".arena")
+        tmp_dir = tempfile.gettempdir()
+        need_bytes = n_rows * 2 * np.dtype(np.float32).itemsize
+        st = os.statvfs(tmp_dir)
+        avail_bytes = st.f_bavail * st.f_frsize
+        if avail_bytes < need_bytes:
+            raise RuntimeError(
+                f"arena_backing: tmp needs {need_bytes} bytes for a centroid buffer but "
+                f"{tmp_dir!r} has only {avail_bytes} free; raise the function's ephemeral "
+                "storage / EphemeralStorage size (up to three such buffers — old + staging + "
+                "bounded merge — can be live concurrently during a fold, so size for their "
+                "combined peak). Overflow otherwise surfaces as an uncatchable SIGBUS on write."
+            )
+        fd, path = tempfile.mkstemp(suffix=".arena", dir=tmp_dir)
         try:
             mapped = np.memmap(path, dtype=np.float32, mode="w+", shape=(n_rows, 2))
         finally:
--- a/src/zagg/processing/streaming.py
+++ b/src/zagg/processing/streaming.py
@@ -43,6 +43,16 @@ from zagg.stats.tdigest import _DEFAULT_DELTA, build_tdigest, merge_tdigests
 _TDIGEST_FUNCTION = "zagg.stats.tdigest.build_tdigest"
 
 
+#: Safe per-cell slot bound for staged tdigest builds/merges, as a multiple of
+#: the field's delta: the k-1 compression law lands at <= ~1.13*delta centroids
+#: (measured: 289 max at delta=256, 578 at delta=512), so 2*delta never
+#: truncates; asserted at every staged write.
+_K_SLOT_FACTOR = 2
+
+#: Shared zero-size arena so releasing a field's old buffer needs no allocation.
+_EMPTY_ARENA = np.empty((0, 2), dtype=np.float32)
+
+
 def _ranges_to_indices(starts: np.ndarray, lengths: np.ndarray) -> np.ndarray:
     """Element indices for concatenated ``[start, start+length)`` ranges.
 
@@ -233,14 +243,17 @@ class StreamingAggregator:
         self._buffered_granules = 0
 
     def _fold_arena(self, col_arrays, cell_to_slice) -> None:
-        """Rebuild the CSR state with this flush's cells folded in.
+        """Rebuild the CSR state with this flush's cells folded in — staged.
 
-        Sizes are exact before allocation: fresh digests (and, for cells
-        already held, their merges) are built first, so the new arena is
-        allocated once at its final size and filled — untouched cells by a
-        vectorized element gather, touched cells by per-cell slice writes.
-        Transients are one buffer's digests plus the old arena, freed on
-        return; the steady state is pure ndarrays with no per-cell objects.
+        Every intermediate is a bulk array, never a per-cell object list:
+        fresh digests are built straight into a bounded staging arena (slot
+        bound ``min(n, 2*delta)`` — group sizes are known before any build),
+        old-only and fresh-only cells move by vectorized element gathers, and
+        only genuine merges loop. A final one-gather compaction squeezes the
+        merge-compression slack back out. Flush transient is therefore old +
+        staging + bounded-new (all payload-sized, freed in stages) — the naive
+        rebuild's list of ~1M small fresh-digest ndarrays measurably cost more
+        than the state itself (issue #217 A/B replay).
         """
         fresh_cells = np.fromiter(cell_to_slice, dtype=np.uint64, count=len(cell_to_slice))
         order = np.argsort(fresh_cells)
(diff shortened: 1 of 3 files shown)


src/zagg/processing/streaming.py
```diff
@@ -173,21 +217,84 @@ def flush(self) -> None:
...
+            keep = ~np.isin(self._cells, fresh_cells, assume_unique=True)
+            src = _ranges_to_indices(offsets[:-1][keep], k_old[keep])
+            dst = _ranges_to_indices(new_offsets[:-1][pos_old[keep]], k_old[keep])
+            new_arena[dst] = arena[src]
+            for j, i_union in enumerate(pos_fresh):
+                d = digests[j]
+                new_arena[new_offsets[i_union] : new_offsets[i_union] + len(d)] = d
+            self._offsets[name], self._arenas[name] = new_offsets, new_arena
+
+        self._cells, self._cell_counts = union, new_counts
```
espg, line 427: 🤖 *from Claude (review)*

**Arena layout silently empties the `occupied_out` sink (issue #200 coverage).** `_fold_arena` maintains state in `self._cells`/`self._cell_counts` and never writes `self.counts` (that dict is only populated by the dict branch at line 224, and stays `{}` for the whole arena run). But the worker reads `buffered.counts` to build the occupied-cell set:

```
# worker.py:579
cells = buffered.counts if buffered is not None else cell_to_slice
occupied_out.append(np.fromiter(cells.keys(), dtype=np.uint64, count=len(cells)))
```

`hive.py:950` passes `occupied_out=occupied` on every shard, and that set derives the stamp's coverage (issue #200). So a run with `state_layout: arena` and any data still appends an **empty** occupied array — the coverage MOC / occupied set is silently dropped for the entire shard, while `df_out`/ragged look correct. `empty` is overridden for arena but `counts` is not, so nothing catches it.

Failing scenario: streaming `state_layout: arena`, a shard whose cells 10/20/30 get folded → `self._cells == [10,20,30]` but `self.counts == {}` → `occupied_out` gets `array([], dtype=uint64)` instead of `{10,20,30}`. The dict layout returns the three cells.

The `_ab`/`_run` test harness never passes `occupied_out`, so this combination is entirely untested. Add a test that runs arena with `occupied_out=[]` and asserts the returned cell set equals the dict layout's.

Severity: blocking (silent data loss in a shipped sink under the new opt-in).

espg: 🤖 *from Claude*

Fixed in 320db3b. Added a layout-agnostic `StreamingAggregator.occupied_cells()` — dict layout keys `self.counts`, arena layout returns the sorted `self._cells` — and `worker.py` now calls it (`buffered.occupied_cells()`) instead of reading `buffered.counts` directly, so both layouts feed `occupied_out` (issue #200 coverage/MOC stamp). Read after the final flush, so it mirrors the merged running state.

Regression test `TestArenaLayout::test_occupied_out_fed_under_arena` drives `process_shard` with an `occupied_out` sink under both layouts, asserts the arena sink is non-empty and set-equal to the dict sink. Verified it fails on the pre-fix `buffered.counts` path (arena sink empty, `0 > 0`) and passes with the fix. `ruff check`/`format --check` clean, `pytest tests/test_streaming.py` 35 passed.

src/zagg/processing/streaming.py
```diff
@@ -43,11 +43,38 @@
 _TDIGEST_FUNCTION = "zagg.stats.tdigest.build_tdigest"
 
 
+#: Safe per-cell slot bound for staged tdigest builds/merges, as a multiple of
+#: the field's delta: the k-1 compression law lands at <= ~1.13*delta centroids
+#: (measured: 289 max at delta=256, 578 at delta=512), so 2*delta never
```
espg: 🤖 *from Claude (review)*

The `_K_SLOT_FACTOR` justification understates the true compression bound. The comment states the k-1 law "lands at <= ~1.13*delta centroids (measured: 289 max at delta=256, 578 at delta=512)". That holds for well-behaved (near-gaussian) `h_li`, but `build_tdigest`/`merge_tdigests` run on any configured `source` column, and heavy-tailed inputs push the centroid count much higher. Measured worst case over Cauchy / Pareto(0.1) inputs:

- delta=256 -> 388 centroids (1.52*delta), not 289
- delta=512 -> 776 (1.51*delta), not 578
- delta=1024 -> 1544 (1.51*delta)

So the real headroom to the `2*delta` cap is ~1.33x, not the ~1.77x the "1.13" figure implies. The cap itself is still safe (1.55*delta observed max < 2*delta), so this is not a truncation bug — but this comment is the load-bearing rationale for the magic `2`, and a future maintainer trusting "1.13*delta measured" could tighten the factor toward 1.5 and start truncating digests on tail-heavy data. Fix the stated bound to reflect the ~1.5*delta algorithmic worst case.

Severity: should-fix (safety-critical comment is empirically wrong by ~33%; code is correct).

src/zagg/processing/streaming.py
```diff
@@ -173,23 +227,135 @@ def flush(self) -> None:
...
+                u = pos_fresh[j]
+                assert len(d) <= k_bound[u], (
+                    f"merged tdigest {len(d)} centroids > slot bound {k_bound[u]} "
+                    f"(delta={delta}) — the {_K_SLOT_FACTOR}*delta compression bound broke"
+                )
+                bound_arena[bound_off[u] : bound_off[u] + len(d)] = d
+                k_exact[u] = len(d)
+            # Release the old arena and staging before compaction so the
+            # compaction peak is bound + exact, not old + stage + bound + exact.
+            self._arenas[name] = _EMPTY_ARENA
```
espg, line 411: 🤖 *from Claude (review)*

Releasing the old arena here breaks the per-field atomicity the pre-commit code had. This line sets `self._arenas[name]` to the 0-row `_EMPTY_ARENA` while `self._offsets[name]` still holds the OLD offsets (they are only reassigned together with the new arena at the final compaction line below). During that window the field is internally inconsistent: offsets index rows `0..N` into an arena that now has 0 rows.

If the compaction that follows raises before the final swap — `np.cumsum` / `_ranges_to_indices` / `bound_arena[src]` all allocate, and a `MemoryError` here is plausible precisely because this whole change targets memory-pressure workloads — the field is left with `_arenas[name] = _EMPTY_ARENA` but stale `_offsets[name]` (and `self._cells` is also still the pre-fold value, since it is only updated after the field loop). A subsequent `chunk_outputs`/`_chunk_outputs_arena` then does `arena[offsets[pos]:offsets[pos+1]]` on the empty arena and silently returns empty slices (dropped digests), not an error. The replaced code (320db3b) never mutated `_arenas[name]` until a single atomic `self._offsets[name], self._arenas[name] = new_offsets, new_arena`, so an OOM there left the field's old state intact.

Only reachable via an unexpected exception (the slot-bound asserts provably never fire — I verified the 2*delta cap holds against Cauchy/Pareto tails), so severity is nit; but if you keep the early release, consider clearing `_offsets[name]` in the same breath (or wrap so a failed fold restores the field), so no code path can observe offsets that outrun the arena.

Severity: nit.

src/zagg/processing/streaming.py
espg: 🤖 *from Claude*

Corrected the `_K_SLOT_FACTOR` doc comment to state the adversarial bound. It now reads that the k-1 compression law peaks near ~1.55*delta on heavy-tailed inputs (Cauchy/Pareto), citing your measurements (388 at delta=256, 776 at delta=512, 1544 at delta=1024), notes this is well above the ~1.13*delta seen on well-behaved data, and adds "Do not tighten below the ~1.55*delta adversarial bound." Factor stays 2. Fixed in 558f880.

src/zagg/processing/streaming.py
espg: 🤖 *from Claude*

Kept the early release (the whole point of #217 is lowering the compaction peak — bringing `arena` + `stage` back alive to swap atomically would roughly double it, since `bound_arena` alone is already comparable to the old arena) and made the failure loud instead.

Added a `self._poisoned: str | None` flag (init None). In `update()` it is set to the field name right after `self._arenas[name] = _EMPTY_ARENA; del arena, stage` and cleared to None immediately after the compaction gather assigns `_offsets[name], _arenas[name]` atomically. So it lingers set only if the gather (e.g. OOM) raises mid-rebuild. `_chunk_outputs_arena` now checks it at the top and raises `RuntimeError("... left inconsistent by a failed compaction; the running state is unusable")` rather than reading an emptied arena as a silent no-digest result. Added `test_poisoned_arena_state_raises_loudly` covering it. Fixed in 86f2ea7.

src/zagg/processing/streaming.py
```diff
@@ -149,11 +211,54 @@ def __init__(self, config: PipelineConfig, grid, handoff: str, buffer_granules:
...
+        paging instead of an OOM kill of anonymous heap.
+        """
+        if self.arena_backing != "tmp" or n_rows == 0:
+            return np.empty((n_rows, 2), dtype=np.float32)
+        import os
+        import tempfile
+
+        fd, path = tempfile.mkstemp(suffix=".arena")
+        try:
+            mapped = np.memmap(path, dtype=np.float32, mode="w+", shape=(n_rows, 2))
```
espg, line 282: 🤖 *from Claude (review)*

`arena_backing: tmp` at scale can crash the worker with an uncatchable SIGBUS, and nothing (docstring, PR body, or a config/docs note) warns that it requires raising Lambda ephemeral storage above the 512 MB default.

Mechanism: `np.memmap(path, mode="w+", shape=...)` grows the file with a single sparse write — internally `fid.seek(bytes-1); fid.write(b"\0")` (numpy `memmap.__new__`). That allocates only the last block; the rest is a hole. So this line **succeeds even when `/tmp` cannot hold the full buffer** — there is no ENOSPC at allocation. Blocks are allocated lazily on page-fault writes (here via `np.take(..., out=new_arena)` and the staged `bound_arena[dst] = ...` writes). When the filesystem fills mid-write, the kernel delivers **SIGBUS to the process**, which is not a catchable Python exception — the worker dies uncatchably rather than raising a clean error.

Scale check: this PRs own phase-2 replay traced the three concurrent `_fold_arena` centroid buffers (staging + bounded-merge + compacted, plus the prior steady arena) peaking at ~3,254 MB on the difficult shards. Under `tmp` backing those same buffers land in `/tmp`, needing ~3 GB of ephemeral storage — ~6x the 512 MB Lambda default. A fleet A/B on the 148/155 g shards with default ephemeral storage would SIGBUS-crash workers, which is easy to misread as "tmp backing is worse" when it is really an unflagged sizing/config gap.

Secondary: the "kernel writes back and reclaims under cgroup pressure" payoff in the docstring only holds where `/tmp` is genuinely disk-backed (true on Lambda). Where `/tmp` is tmpfs (common on dev/CI boxes), file-backed pages are RAM that counts against the same cgroup and cannot be evicted — no benefit, strictly extra overhead.

Suggest: (a) document that `arena_backing: tmp` requires ephemeral storage >= the arena transient peak and that overflow is an uncatchable SIGBUS, and (b) consider `fallocate`/explicit sizing so overflow surfaces as a clean ENOSPC at allocation instead of SIGBUS at write. Severity: medium — opt-in and off by default, but the failure is a silent uncatchable crash and the sizing requirement is undocumented.

espg: 🤖 *from Claude*

Folded in 2f5cb52. Both parts of the finding addressed in `_alloc_arena`:

- **Runtime guard**: before returning the tmp-backed mapping, `_alloc_arena` now does `os.statvfs(tempfile.gettempdir())` and compares `f_bavail * f_frsize` against the requested `n_rows * 2 * itemsize`. On shortfall it raises a clear `RuntimeError` naming the fix ("raise the function's ephemeral storage / EphemeralStorage size") and noting the ~3-buffer concurrent transient (old + staging + bounded merge) and the SIGBUS failure mode — a loud error at allocation instead of a deferred uncatchable SIGBUS on the page-fault write. mkstemp now pins `dir=tmp_dir` so the check and the file share a filesystem. Kept it to a single honest free-vs-request check; the nuance rides in the message rather than a predictor.
- **Docstring**: corrected the reclaim claim to hold only for disk-backed `/tmp` (Lambda) and called out that tmpfs pages are RAM against the same cgroup with no benefit; documented the ephemeral-storage sizing requirement and the sparse-`w+`/SIGBUS mechanism explicitly.

Test: `TestArenaTmpAlloc` — statvfs monkeypatched to 1 block → `RuntimeError` matching `ephemeral`; happy path allocates `(16, 2)` float32; zero-rows early-returns without touching statvfs. `tests/test_streaming.py` 47 passed; ruff check + format clean on both files.

PR #261, opened 2026-07-17
Benchmark: capture + display per-phase timings through time (issue #250)

### LINKED ISSUE #250: Benchmark: capture + display per-phase timings (setup, read/index/aggregate) through time

espg:
Closes #250

Captures per-phase/per-stage timings in the benchmark series and restructures the benchmark display around them, per the original issue items (1)–(3) **plus** the [handoff checklist](https://github.com/englacial/zagg/issues/250#issuecomment-5005935414) — in scope since PR #256 merged and espg [approved the rendering understandings](https://github.com/englacial/zagg/pull/256#issuecomment-5006361723) (agg = index + aggregate + write split; raster stage diagnostics; 2×2 + per-100 km² retirement). Coded against `main`'s merged emission (`worker_phase_max` read/index/aggregate/write; raster `phase_timings["stages"]`).

## Phases

- [x] Phase 1 — capture: `worker_phase_max` → `phase_read_s`/`phase_index_s`/`phase_aggregate_s` in `FULL_AOI_COLUMNS`, null-safe (wt_* pattern)
- [x] Phase 2 — display: per-release phase figure (superseded in phase 7 by the approved panel-per-phase diagnostics; single-figure form archived)
- [x] Phase 3 — `setup_cost_usd` as a separate, non-destructive column
- [x] Post-#255 fold: finalize restored to the display; layout-split `setup_s` semantics ([espg's question](https://github.com/englacial/zagg/pull/261#issuecomment-5007197671) answered on-thread)
- [x] Phase 4 — #256 write split: `phase_write_s` in both series; per-merge phase columns (`bench_metrics.PHASE_MAP`, raw emission names retained, agg mapping is render-time)
- [x] Phase 5 — target collapse: per-merge `targets.json` → the single `tdigest_healpix_o9_hive` config (2×2 → `provisional_targets`); full-AOI point leg → single `full_aoi_neon_o9_hive_mask` target
- [x] Phase 6 — raster release leg: `run_raster_benchmark.py` (harness-owned profiled `mode=process_raster` dispatch), `raster_series.py` (`raster_series.parquet`), pinned catalog `cat_s2_neon_2025.parquet` + `s2_neon_o9.yaml` + `targets_raster_neon.json`
- [x] Phase 7 — renderer restructure: `plot_summary.py` (dual-axis lambda-s ⇔ USD summaries, panel-per-phase point diagnostics, panel-per-stage raster diagnostics, objects tripwire figure), `finalize_cost_usd` column, summed-total display, per-100 km² dropped from tables/`latest.md`/`metrics.json`, matrix/2×2 figures retired to Archived, machine-readable companions re-pointed to the live hive target
- [x] Phase 8 — workflows: `lambda-benchmark-fullaoi.yml` gains the `raster` job (serialized after the point leg; anonymous Earth Search, no EDL). `lambda-benchmark.yml` needs **no diff** — the collapse is manifest-driven and `run_benchmark.py` already dispatches `profile=True`
- [x] Phase 9 — `docs/deployment/benchmark.md` rewritten to the approved layout (same PR as the renderer, embeds/filenames in lockstep)
- [x] Phase 10 — [espg's review](https://github.com/englacial/zagg/pull/261#issuecomment-5007928444) point (3): `profile` plumbed through the runner's raster path (`RasterStrategy` threads the opt-in event key; summary rolls `worker_stage_max`/`worker_stage_counts` + `template_s`/`lambda_time_s`/`worker_max_s`/`worker_median_s`/`max_memory_mb`); the harness now dispatches via `agg(profile=True)` and the duplicated transport is deleted
- [x] Phase 11 — point (4): raster worker RSS. `deployment/aws/lambda_handler.py`'s `process_raster` branch now reports `max_memory_mb` (sampled per-invocation peak, `_PeakRSSSampler`, issue #141 convention) + `container_hwm_mb`, point-path parity; `max_memory_mb` column in `raster_series`; raster summary/diagnostics markers colour automatically. **Note:** this edits `deployment/aws/lambda_handler.py` — espg's PR comment is the explicit ask; the diff is a sampler start/stop + two body fields
- [x] Phase 12 — point (1): hive-from-day-one gating. `targets_raster_neon.json` carries a `pending_targets` hive variant (`store_layout: hive`, blocked on issue #237); the harness applies+revalidates a target `store_layout` override and records the layout axis (`store_layout` column in `raster_series`), so the #237 flip is a one-block manifest move; a tripwire test pins today's issue #239 gate and fails the moment upstream accepts raster hive

## Testing

- Targeted suites green locally: `tests/test_benchmark.py`, `tests/test_full_aoi_series.py`, `tests/test_full_aoi_benchmark.py`, `tests/test_raster_benchmark.py` (137 passed) + ruff check/format clean on touched files.
- New coverage: phase/stage flatten null-safety (unknown future phases/stages stay JSON-only), schema-tail order pins, target-collapse pins (run-all = the hive config; retired arms resolvable via `--target`), raster stage rollup (max seconds / summed counts; unprofiled → empty, never zero-faked), offline raster dry-run e2e from the pinned catalog, summary/diagnostics render + skip contracts, archived-section behavior, per-100 km² absence in rendered outputs.
- Figures verified visually on synthetic multi-release/multi-merge data (dual-axis relabel, memory colorbar, NaN line-breaks, bottom-row-only labels).
- Workflow YAML validated syntactically (`yaml.safe_load`; job graph `full-aoi` → `raster`).

## Questions for review

1. **Cost synthesis (issue item 3/6, decision owed).** Implemented: `setup_cost_usd` + `finalize_cost_usd` as separate series columns (non-destructive — `cost_usd` keeps worker-GB-seconds semantics for a comparable history) while the **summary displays the summed total** (workers + setup + finalize) so the headline dollar figure is honest. Post-#255 this is also the only coherent option on hive: `setup_s` there is the sync residue (ping + async dispatch) and the async manifest write's billed GB-s is unobservable — folding into `cost_usd` would splice layout-dependent quantities into the primary series. Veto = drop the two columns and fold, a small diff.
2. **Raster hive** — per espg: needed, waiting on upstream. Structured in phase 12 so the flip is zero-rework (**the hive raster target is Blocked by #237**: promote `pending_targets.raster_s2_neon_2025_hive` into `targets` + delete the tripwire test when issue #237 merges). The strict AOI mask remains point-path-only (issue #101).
3. **Pinned S2 catalog**: committed `tests/data/benchmark/catalogs/cat_s2_neon_2025.parquet` (132 KB) from the operational SERC artifact (`catalog_s2_serc_2025.parquet`, 85 items, 2025, `s2:datatake_id` time key) — the cat_neon precedent. OK to canonize that artifact?
4. ~~Harness-owned raster dispatch~~ — resolved by phase 10 per espg's authorization: profile is plumbed through `agg`'s raster path and the harness transport is deleted.
5. ~~Raster memory column~~ — resolved by phase 11 per espg's ask: the raster worker reports sampled peak RSS (point-path parity) and the series/renderers carry it.
6. **`plot_series.py` is at ~1020 lines** (repo ~1000-line ceiling): the new renderers went into a separate `plot_summary.py` instead of growing it further; the retired matrix/full-AOI renderers are kept frozen-but-importable per the #193/#202 precedent. Happy to cull the frozen renderers instead if you prefer the file smaller.
7. **First live run is yours** (operator-runs-validation): the workflow/harness changes are validated offline (dry-run + YAML parse + unit tests) but cannot be fleet-tested from here — the raster leg additionally needs the benchmark role to allow the harness's direct template PUT to the store prefix (the point path's template is written by the setup Lambda; the raster template is orchestrator-written, as in `agg`'s local raster path).
8. **Sequencing**: the display maps `agg = index + aggregate` from raw retained columns, so pre-#256 workers (no `write` split) degrade to null write cells and `phase_agg_s` falls back to whichever phase exists — no backfill needed.


.github/scripts/full_aoi_series.py
```diff
@@ -90,6 +90,13 @@
     # JSON-only (dropped below).
     "store_layout",
     "parity_ok",
+    # Worker per-phase straggler split (issue #250): max seconds across shards
```
espg, line 94: 🤖 *from Claude (review)*

Minor doc drift introduced by this phase. The schema-overview comment above `FULL_AOI_COLUMNS` (lines 32-33) still frames flattening as a single-dict story:

> The harness's nested `write_throughput` dict is flattened to the `wt_*` scalar columns below\n\nAfter this phase there are **two** nested dicts flattened — `write_throughput` → `wt_*` and now `worker_phase_max` → `phase_*_s`. `flatten_record`'s own docstring was correctly updated to mention both, but this top-of-module overview wasn't, so it now under-describes the schema. Suggest tacking on "…and `worker_phase_max` to the `phase_*_s` columns" so the overview and the function docstring stay in lockstep. (Low priority; the code itself is correct.)

.github/scripts/plot_series.py
```diff
@@ -777,6 +794,102 @@ def make_full_aoi_release_figure(
...
+    # Bottom-row-only release labels, matching _render_panel_grid's convention.
+    for c in range(ncols):
+        last = max((r for r in range(nrows) if layout[r][c] is not None), default=None)
+        for r in range(nrows):
+            if layout[r][c] is not None and r != last:
+                axes[r][c].tick_params(labelbottom=False)
+
+    # Figure-level legend just below the suptitle (anchored so the two never
+    # overlap; bbox_inches="tight" keeps both in frame).
+    handles, labels = legend_ax.get_legend_handles_labels()
```
espg: 🤖 *from Claude (review)*

`make_full_aoi_phase_figure` crashes with `AttributeError: NoneType object has no attribute get_legend_handles_labels` whenever `_matrix_layout` returns an all-`None` grid while `hist` still has target rows. `legend_ax` is only assigned inside the loop (line 871), so if no target lands in any of the fixed inline/sidecar × nomask/mask slots, `legend_ax` stays `None` and line 882 dereferences it.

This is reachable: `_matrix_layout` slots on `(_aoi_axis(target), str(r.get("index_backend", "")))`. A row whose `index_backend` is null/absent (a pre-#193 legacy row, or any run that did not record the backend — the same case the `_aoi_axis`/`_matrix_mask` docstrings call out) maps to key `(..., "")`, in neither `MATRIX_COLS`, so every slot is `None`. Reproduced with the 4 full-AOI targets and `index_backend=None`:

```
layout: [[None, None], [None, None]]
make_full_aoi_phase_figure  -> CRASH: AttributeError ... get_legend_handles_labels
make_full_aoi_release_figure -> returned True   # sibling degrades gracefully
```

The sibling renderer routes through `_render_panel_grid`, which never touches a per-panel handle and just writes blank panels, so it returns True. Here the exception propagates through `main()` (the call at ~line 1037 is unguarded) and takes down the whole plot_series render / Pages build instead of omitting the section.

Fix: after the panel loop, guard `if legend_ax is None: plt.close(fig); return False` (matches the "no data -> skip the section, no broken image" contract this function already documents).

.github/scripts/full_aoi_series.py
espg: 🤖 *from Claude*

Extended the schema-overview comment above `FULL_AOI_COLUMNS` to note the `worker_phase_max` -> `phase_*_s` flatten alongside the `wt_*` one, so the overview and the `flatten_record` docstring stay in lockstep. Fixed in ae413ca.

.github/scripts/plot_series.py
espg: 🤖 *from Claude*

Added an `if legend_ax is None: plt.close(fig); return False` guard after the panel loop so an all-None layout (every row's `index_backend` null/off-matrix) honours the "return False, write nothing" contract instead of dereferencing a None `legend_ax`. Added a regression test (`test_full_aoi_phase_figure_skips_when_all_slots_off_matrix`) building the 4 full-AOI targets with `index_backend=None` and asserting False + no file written. Fixed in 1831f49.

espg:
> setup_s is the orchestrator wall around the setup invoke (runner.py:2142-2155), not the setup Lambda's billed duration (the summary doesn't carry it), so setup_cost_usd slightly overstates by the invoke round-trip overhead. Flagged in the code comment; acceptable for a ~104 s flat / ~3.7 s hive signal?

this is changing / changed by the just merged #255 ; does that impact your other question too?

espg:
🤖 *from Claude*

Yes — [PR #255](https://github.com/englacial/zagg/pull/255) bears on the fold-vs-separate question too, and it **strengthens the separate-column choice**. Post-#255, `setup_s` is layout-dependent in *meaning*: on flat it is still the orchestrator wall around the sync fullsphere-template invoke; on hive it is only the preflight ping + the ~10 ms async Event dispatch (the [hybrid lifecycle ratified on issue #252](https://github.com/englacial/zagg/issues/252#issuecomment-5006627722)). Folding it into `cost_usd` would therefore splice two different quantities into the primary cost series — exactly the discontinuity the separate column avoids. And the fold option's premise (`setup dollars ≈ setup_s × GB × $/GB-s`) no longer holds on hive at all: the manifest write's billed GB-s went fire-and-forget, so the orchestrator never observes it. `setup_cost_usd` now documents that it measures only the **sync setup-path residue** on hive rows, and never invents an async cost (0494232).

Also folded #255's explicit ask ("#250 should keep `finalize` in its panel per issue #252", noted in its `bench_metrics.py` column comment): **`finalize_s` is restored to the phase panel** (37e030e) with the post-#255 semantics in the legend/comment — ~0 on flat, but on hive the load-bearing idempotent manifest backstop (the async init write runs with retries 0, so finalize self-heals a lost manifest). Branch is merged up to `main` (44366e0) so the harness comments sit on top of #255's `bench_metrics.py` text; docs section updated to match (0494232).

The overstatement caveat you quoted survives only on the flat arm (sync wall vs billed duration); on hive the residue is seconds-scale, so the caveat is now marginal there.

tests/data/benchmark/targets.json
```diff
@@ -1,5 +1,5 @@
...
+  "description": "Benchmark targets: pinned configs + shard maps per issues #110/#133/#193/#202, collapsed to a SINGLE live per-merge configuration per issue #250 (espg-approved restructure on PR #256): o9, hive store layout, sharded output, tdigest, inline read, no AOI mask, granule_workers=4, 4 GB. The former inline/sidecar x mask/nomask 2x2 (issue #202 reset) is retired to provisional_targets (runnable by explicit --target, frozen rows retained in the series; its figures move to the Archived section). The live target doubles as the write-path regression arm: the object-count tripwire hard-fails the merge run on a sharded-write bypass (issue #215). Every-merge run dispatches ONE pinned densest shard (run_benchmark.py) with profile=True so the worker phase split (read/index/aggregate/write) lands in the series.",
```
espg, line 2: 🤖 *from Claude (review)*

The phase-5 collapse silently orphans the agent-facing "current numbers" artifacts. `plot_series.py` keys the per-merge live matrix — and the `metrics.json` / `latest.md` companions — on `_matrix_mask` (flat rows whose target ends in `_mask`/`_nomask`); the hive arm is *deliberately excluded* by both its suffix-free name and the `store_layout` gate (`plot_series.py:111-139`). Before this phase the live set was the four flat 2×2 targets, so `metrics.json` tracked the newest merge. After the collapse the sole live target is `tdigest_healpix_o9_hive`, which `_matrix_mask` drops — so in `main()` (`plot_series.py:1030-1034`) `matrix_df = df[_matrix_mask(df)]` no longer contains any post-collapse merge:

```python
matrix_df = df[_matrix_mask(df)] if not df.empty else df
has_md = not matrix_df.empty and write_latest_markdown(matrix_df, outdir / "latest.md")
has_json = not matrix_df.empty and write_latest_metrics(matrix_df, outdir / "metrics.json")
```

On the benchmarks branch (which has historical flat 2×2 rows) `matrix_df` stays non-empty, so `write_latest_metrics` picks `_latest_merge(matrix_df)` = the last *pre-collapse* commit and re-emits it every merge — `metrics.json`/`latest.md` freeze at a stale point and present old numbers as the current merge. That is the documented path an agent follows for live benchmark numbers, so this is not a merely-empty legacy panel — it is actively-stale output, distinct from the cleanly-dormant matrix *figure* (which just returns `False`).

I understand the renderer restructure (2×2 → Archived, new single-target/summary figures) is a later phase and I am not flagging that absence. The narrow question here: is `metrics.json`/`latest.md` serving stale-as-current numbers an acceptable interim state, or should the collapse re-point those companions off the retired matrix (e.g. onto the live hive target / whole retained series) in this PR so the agent-facing artifact never silently lags?

tests/data/benchmark/targets_full_aoi_neon.json
```diff
@@ -1,5 +1,5 @@
...
+  "description": "Full-AOI NEON benchmark targets (issue #202 legs 1+5), collapsed to the SINGLE point-pipeline release leg per issue #250 (espg-approved restructure on PR #256): o9, hive store layout, sharded, tdigest, inline read, WITH the strict AOI binary mask, all shards over the AOP_NEON.geojson box, recorded PER RELEASE. The former inline/sidecar x mask/nomask 2x2 and the flat parity arm are retired (their rows stay in full_aoi_series.parquet; figures move to Archived). The raster release leg (Sentinel-2 2025) rides its own manifest (targets_raster_neon.json) and harness. run_full_aoi_benchmark.py builds the shardmap over the AOI+temporal+cmr below from the PINNED catalog catalogs/cat_neon.parquet (2089 granules, offline, fixed across releases: the series tracks CODE change, not CMR drift; rebuild via run_full_aoi_benchmark._prefilter + Catalog.to_geoparquet). The whole-AOI build fans to 4 o9 shards. Dispatched with profile=True so the worker phase split (read/index/aggregate/write) lands in the series.",
```
espg, line 2: 🤖 *from Claude (review)*

The manifest collapse leaves `docs/deployment/benchmark.md` making factual claims that now contradict the shipped manifests:

- **`docs/deployment/benchmark.md:37`** — "A fifth committed target, `tdigest_healpix_o9_hive` … It deliberately stays out of the 2×2 panels" and the surrounding paragraphs (`:24-45`) still describe the inline/sidecar × mask/nomask 2×2 as the *live* per-merge matrix. After this phase the `targets` block in `targets.json` is exactly `["tdigest_healpix_o9_hive"]` (the 2×2 moved to `provisional_targets`), so it is the **only** committed target, not "a fifth," and there is no live 2×2 anymore.

- **`docs/deployment/benchmark.md:93`** — "The release matrix also carries a fifth target, `full_aoi_neon_o9_hive` … plus a **flat↔hive output-parity** read-back against its flat sibling (`parity_with` in the manifest) … recorded as `parity_ok`." This phase renamed that target to `full_aoi_neon_o9_hive_mask` (now the *sole* target, `aoi_mask: true`) and **removed** `parity_with` and the flat sibling entirely (see the diff here and `tests/test_full_aoi_benchmark.py:317-321`, which now asserts `"parity_with" not in t`). So the doc references a target name that no longer exists and a parity read-back that no longer runs.

This is the "manifest description contradicting the new shape" class, not the renderer-figure restructure (which I understand is a later phase). Suggest updating both doc paragraphs in this PR to describe the single collapsed per-merge target and the single hive+mask release leg, and dropping the flat-sibling/parity narrative. Pinned to the full-AOI manifest since `:93` is the more concrete drift (removed name + removed `parity_with`); `:37` needs the same treatment.

.github/scripts/plot_summary.py
```diff
@@ -0,0 +1,331 @@
...
+
+def make_release_objects_figure(hist: pd.DataFrame, out_png: Path) -> bool:
+    """Store object total vs release for the collapsed release leg (issue #240).
+
+    Keeps the sharded-write-bypass tripwire visible after the 2x2 retirement:
+    a bypass reads as a ~K-fold step here (record-only on the release leg).
+    """
+    if hist.empty or "objects_total" not in hist.columns or hist["objects_total"].dropna().empty:
+        return False
+    return make_diagnostics_figure(
```
espg, line 341: 🤖 *from Claude (review)*

`make_release_objects_figure` renders **object counts** (`objects_total` / `objects_expected`) but delegates to `make_diagnostics_figure`, which hardcodes `ax.set_ylabel("seconds")` on every panel (line 298). The published objects tripwire figure (and its `full_aoi_objects.png` embed in the docs) therefore labels an integer object-count axis as "seconds" — misleading to a reader who reads the store-object step off it.

Every other caller of `make_diagnostics_figure` (POINT_PHASE_PANELS, RASTER_STAGE_PANELS, merge_phases) genuinely plots seconds, so the label is only wrong on this reuse path.

Suggested fix: give `make_diagnostics_figure` a `ylabel="seconds"` keyword (default preserves current behavior) and pass `ylabel="store objects"` from `make_release_objects_figure`.

.github/scripts/plot_summary.py
```diff
@@ -0,0 +1,331 @@
...
+        # scales can never disagree.
+        usd = cost_ax.secondary_yaxis(
+            1.0,
+            functions=(lambda s: s * _USD_PER_LAMBDA_S, lambda d: d / _USD_PER_LAMBDA_S),
+        )
+        usd.set_ylabel("USD", color="C0")
+        _finish_axis(cost_ax, xs, labels)
+
+        _draw_series(
+            wall_ax, xs, hist["total_wall_s"].to_numpy(float), fracs=fracs, norm=norm, color="C1"
```
espg, line 258: 🤖 *from Claude (review)*

Defensive/consistency (not a live crash on current parquets): `make_summary_figure` accesses `hist["total_wall_s"]` (and `hist["total_billed_s"]` at line 232) with no column-presence guard, while its sibling `make_diagnostics_figure` filters panels to `c in hist.columns` and skips absent ones. If a frame reaches here without `total_wall_s` it raises `KeyError` mid-render rather than degrading.

In production the three source schemas (`FULL_AOI_COLUMNS`, `RASTER_COLUMNS`, `RECORD_COLUMNS`) all carry `total_wall_s`, and the reindex-on-write materializes it, so this is not currently reachable — but it means a future column rename, or a hand-built/legacy-parquet frame predating the column entirely, fails hard here where the diagnostics path would degrade. Reproduced by dropping `total_wall_s` from an otherwise-valid point-release frame:

```
KeyError: 'total_wall_s'  # plot_summary.py:245, hist["total_wall_s"].to_numpy(float)
```

Suggested fix: guard `total_wall_s`/`total_billed_s` presence (fall back to NaN column) so the summary degrades like the diagnostics figure instead of KeyError-ing. Leave-by-design is reasonable if you prefer the loud failure — flagging the asymmetry.

.github/scripts/run_raster_benchmark.py
```diff
@@ -0,0 +1,353 @@
...
+
+def _median(values: list[float]) -> float | None:
+    if not values:
+        return None
+    s = sorted(values)
+    mid = len(s) // 2
+    return s[mid] if len(s) % 2 else (s[mid - 1] + s[mid]) / 2.0
+
+
+def _dispatch_shards(
```
espg: 🤖 *from Claude (review)*

`_dispatch_shards` is a hand-copied mirror of `RasterStrategy._run_lambda_shards` (runner.py ~789) and must stay lockstep with the handler contract in `deployment/aws/lambda_handler.py` (~985–1076): the event envelope (`mode`/`shard_key`/`granules`/`config`/`store_path`/`time_index`/`profile`), the per-shard `time_index` slice keyed on `time_key or datetime`, and the `FunctionError` / `statusCode != 200` / `body["duration_s"]` parsing. None of that is unit-tested — `test_raster_benchmark.py` covers `stage_rollup`/`flatten`/dedup/dry-run but the transport is explicitly left to live AWS (module docstring). The whole justification for owning a transport mirror instead of reusing the runner is drift risk, yet drift is exactly what has no guard: if the handler renames a body field or the envelope shape shifts, this silently mis-parses at release time on billed invokes.

A `boto3.client` stub (monkeypatch `boto3.client` to a fake whose `invoke` returns a canned `{"Payload": BytesIO(...), "FunctionError": ...}`) would pin the emitted event fields, the `time_index` slice, `profile: true`, and the statusCode/FunctionError branches offline — the same technique the runner-transport tests could use. Recommend adding one before the first live run (PR question 7).

I verified by reading that the envelope/parse currently matches the handler; this is about keeping it matched, not a present mismatch.

tests/data/benchmark/targets.json
espg: 🤖 *from Claude*

Addressed by phase 7 (6ffb92a): `plot_series.main` now re-points the machine-readable companions (`metrics.json` / `latest.md`) at the live hive target via `plot_summary.merge_history`, and drops the retired `cost_per_100km2_usd` field from the display so the current-numbers artifacts track the collapsed target set. No further change needed here.

tests/data/benchmark/targets_full_aoi_neon.json
espg: 🤖 *from Claude*

Addressed by phase 9 (3ebb939): `docs/deployment/benchmark.md` was fully rewritten to the espg-approved layout, so its claims now match the collapsed target set. No further change needed here.

.github/scripts/plot_summary.py
espg: 🤖 *from Claude*

Fixed in 6820632. `make_diagnostics_figure` now takes a `ylabel: str = "seconds"` keyword and uses it for `set_ylabel`; `make_release_objects_figure` passes `ylabel="objects"` so the object-count axis is no longer mislabelled as seconds. Extended `test_summary_and_point_diagnostics_render` to render with the `ylabel` keyword.

.github/scripts/plot_summary.py
espg: 🤖 *from Claude*

Fixed in 3201213. Added a `_summary_ready` guard: a hist is charted only if it carries non-null `total_billed_s` and `total_wall_s`; frames lacking either (or all-null) are skipped like an empty hist, and all rows skipped -> `False` -- preserving the renderer's "return False, never raise" contract. Added a regression test (a frame with `total_wall_s` dropped -> `False`, no file, no exception).

.github/scripts/run_raster_benchmark.py
espg: 🤖 *from Claude*

Fixed in b803d73. Added an offline `_dispatch_shards` test set to `tests/test_raster_benchmark.py` (mocked boto via `monkeypatch.setattr("boto3.client", ...)` with a stub client that records events and replays a scripted Lambda envelope). Covers: (1) happy path -- asserts the event carries `mode="process_raster"`, `profile=True`, an int `shard_key`, the `store_path`, and only this shard's own `time_index` slice, with the body parsed from the nested proxy envelope; (2) `FunctionError` -> `Lambda error:` result; (3) `statusCode != 200` -> `body["error"]` surfaced; (4) invoke exception -> error string.

espg:
> Raster leg deviations from the approved mock (pending upstream): no hive

we need hive; we can wait if needed for the upstream issue to close and then enable it

> Pinned S2 catalog: committed tests/data/benchmark/catalogs/cat_s2_neon_2025.parquet (132 KB) from the operational SERC artifact (catalog_s2_serc_2025.parquet, 85 items, 2025, s2:datatake_id time key) — the cat_neon precedent. OK to canonize that artifact?

do we need this-- and if so, why? We *can* if it's needed... but isn't the element 84 stac endpoint public and reachable for the stac query from the CI/CD?

> Harness-owned raster dispatch: zagg.runner's raster lambda transport neither threads profile nor surfaces per-shard phase_timings, so the harness mirrors its transport with "profile": true (no src/zagg change, per the issue's scope note). If you'd rather plumb profile through agg and delete the harness transport, that's a runner follow-up.

not sure I understand this-- what's going on here?

> Raster memory column: the raster handler reports no RSS, so raster_series has no memory column and raster markers render uncoloured (the mock's shaded raster % cap cell isn't currently measurable). Worker-side RSS reporting would be a src/zagg follow-up.

I genuinely thought we had this-- either through #249 , or just already from when we implemented the raster module. We should go ahead and add it-- I don't mind it being another phase on this PR if needed, sound like a short pass

espg:
🤖 *from Claude*

All four points addressed — three as new phases (10–12), one answered below.

> we need hive; we can wait if needed for the upstream issue to close and then enable it

Done as **phase 12** (1e68c04), structured for a zero-rework flip: `targets_raster_neon.json` now carries a `pending_targets.raster_s2_neon_2025_hive` variant (`store_layout: hive`, blocked on issue #237), the harness applies + re-validates a target-level `store_layout` override, and `raster_series` gained the `store_layout` axis column. **When #237 merges, the only change is promoting that block into `targets` and deleting the tripwire test** (`test_pending_hive_target_blocked_on_237`, which pins today's issue #239 rejection and will fail the moment upstream accepts raster hive — a built-in reminder). Until then the live raster leg runs flat/fullsphere.

> do we need this [the pinned catalog] -- ... isn't the element 84 stac endpoint public and reachable from the CI/CD?

The endpoint is reachable from CI — the pin isn't about reachability, it's about **series comparability**: a live STAC query's item set evolves under your feet even with a fixed datetime range + collection, because Earth Search reprocesses (new `s2:processing_baseline` items appear, ids/assets change), so release N and N+1 would silently measure different workloads and the per-stage series would confound code change with catalog drift. That's exactly why the point leg pins `cat_neon.parquet` instead of querying CMR per release. A middle path exists if you'd rather not commit the parquet: **pin the 85 item IDs (a few KB of JSON) and fetch those exact items live at run time** — deterministic workload, no binary in-tree, at the cost of a release run failing if e84 ever retires an item. Happy to switch to that; the 132 KB parquet just follows the existing precedent.

> not sure I understand this [harness-owned dispatch] -- what's going on here?

What was going on: the worker's raster branch only emits its stage timings when the *event* carries `"profile": true`, but `zagg.runner`'s raster lambda transport never set that key and threw away each worker's response body beyond `timesteps` — so `agg(profile=True)` silently profiled nothing on the raster path (the kwarg fell into `**_ignored`). To capture stages without touching `src/zagg` (the issue's original scope note), the harness had duplicated the invoke loop with the profile key. Since your point (4) authorizes src phases here, **phase 10** (b60a9a0) does the clean fix: `RasterStrategy` now threads `profile` into the events and rolls the workers' `phase_timings` into `summary["worker_stage_max"]`/`["worker_stage_counts"]` (plus always-on `template_s` / `lambda_time_s` / `worker_max_s` / `worker_median_s` / `max_memory_mb` rollups), and the harness just calls `agg(..., profile=True, max_retries=1)` — the duplicated transport is deleted.

> I genuinely thought we had this [raster memory] ... We should go ahead and add it

Verified: the point path (`process`), `extract`, and `process_event` branches all report `max_memory_mb`, but the `process_raster` branch never did — #249 added stage timings only. **Phase 11** (6766c3f) adds it with point-path parity: `_PeakRSSSampler` (the issue #141 sampled per-invocation peak, so warm containers don't inflate it) + `container_hwm_mb` fallback in the raster body, rolled up through the runner into `raster_series.max_memory_mb`, and the raster summary/diagnostics markers now take the same green→red %-cap colouring as the point legs. This touches `deployment/aws/lambda_handler.py` (sampler start/stop + two body fields) — flagged in the PR body with your comment as the explicit ask. Note the fleet needs a redeploy before the first raster release run reports memory; rows from older workers read back null (uncoloured), never zero.

Docs updated to match (56c4a2c). CI-side: nothing else changed in the workflows.

docs/deployment/benchmark.md
```diff
@@ -1,116 +1,166 @@
...
+  strict AOI mask** (issue #101 is point-path-only); the AOI scoping is the
+  catalog's STAC-query clip.
+
+### Summary — total billed cost and wall time
+
+One row per pipeline: **total billed cost** on a dual axis — billed
+lambda-seconds (left) and USD (right, an *exact* relabeling at the fixed 4 GB
+price) — and overall wall, against the release tag. Markers on both rows carry
+the peak-RSS colour scale (green→red % of the 4 GB cap, MB twin axis): the
+raster worker reports the same sampled per-invocation peak as the point path
```
espg, line 50: 🤖 *from Claude (review)*

The phase-12 docs sync updated this claim but missed the sibling docstring in the renderer it describes. `.github/scripts/plot_summary.py` `make_summary_figure` (lines ~222-224) still reads:

> Markers carry the unchanged memory colour scale where the series reports memory (the raster handler doesn't -- its markers stay uncoloured).

After phase 11 the raster handler DOES report `max_memory_mb`, and `_mem_norm_and_fracs` already reads that column (`plot_summary.py` ~line 144), so raster markers will now be coloured — the exact opposite of what the docstring says. (Same file, `raster_release_history` line ~98: the parenthetical "(harness-owned template)" is also now stale — phase 10 moved template emission into the runner/`agg`, so it is runner-owned.) Neither is in this PR's diff, so update them in a follow-up commit so the docs-sync is actually complete.

src/zagg/runner.py
```diff
@@ -613,6 +638,16 @@ class RasterStrategy:
     observation count); ``timesteps`` separately carries the global datatake
     count. A dashboard summing ``total_obs`` across kinds mixes units — read it
     per kind.
+
+    ``profile=True`` (issue #250) threads the opt-in ``profile`` event key to
+    the workers (byte-identical payload when off) and rolls their
+    ``phase_timings`` up into ``summary["worker_stage_max"]`` (straggler-maxed
+    seconds per issue #249 stage + the ``write`` bucket) and
+    ``summary["worker_stage_counts"]`` (summed work counts); the local backend
+    profiles the same stages in-process. The lambda summary also carries the
```
espg, line 652: 🤖 *from Claude (review)*

This RasterStrategy docstring is accurate, but the public `agg` `profile` param doc (runner.py lines ~188-193, not in this diff) is now point-path-only and misleading for the raster path this PR wires:

- it scopes profiling to "When ``True`` (lambda backend) ... forwards ``profile`` into each cell event" — but on the raster path `profile=True` also profiles the LOCAL backend in-process and rolls into the summary (`worker_stage_max`/`worker_stage_counts`);
- it names the stages "read/index/aggregate/write" — the raster worker emits `open`/`geometry`/`fetch`/`decode`/`gather` + the `write` bucket;
- it says "the run prints a per-phase worker breakdown" — the raster path rolls into the summary, no print.

A caller doing `agg(reader=raster, profile=True)` (the harness) reads the wrong contract. Worth a one-line raster caveat on the `agg` profile param doc.

docs/deployment/benchmark.md
espg: 🤖 *from Claude*

Synced both stale sibling docstrings in `.github/scripts/plot_summary.py`. (i) `make_summary_figure` no longer claims the raster handler leaves markers uncoloured — phase 11 added raster `max_memory_mb` and `_mem_norm_and_fracs` reads it, so both point and raster legs colour; wording now says both handlers report `max_memory_mb`. (ii) `raster_release_history` no longer says "harness-owned template" — the runner emits the template (reported as `template_s`); kept the accurate part that the raster leg has no sync setup/finalize invokes so its total stays `cost_usd`. Fix: 6b703ab.

src/zagg/runner.py
espg: 🤖 *from Claude*

Updated the `agg` `profile` numpydoc entry in `src/zagg/runner.py` to cover the raster path. It now states raster profiles on BOTH backends, emits the issue #249 stage set (open/geometry/fetch/decode/gather + write), and rolls into `summary["worker_stage_max"]`/`["worker_stage_counts"]` rather than printing — while keeping the existing point-path wording (lambda-only, read/index/aggregate/write, printed breakdown). Fix: c588c2d.

PR #263, opened 2026-07-17
Raster hive write path (issue #247)

### LINKED ISSUE #247: Raster hive write path: windowed (time, cells) leaves, per-leaf templates and stamps, Lambda hive branch

espg:
Closes #247
Refs #239

## What

The **raster hive write path** — PR B of the #237 phasing ([decision record](https://github.com/englacial/zagg/issues/237#issuecomment-4986819344)): Sentinel-2 raster stores adopt the morton-hive layout as windowed `(time, cells)` leaves per `morton-hive/2` (D13/D14/D15 in `docs/design/sparse_coverage.md`).

Implements the fully ratified plan on #247:
- [Plan](https://github.com/englacial/zagg/issues/247#issuecomment-4990713286) (five phases)
- [Module placement + Q2 confirmation](https://github.com/englacial/zagg/issues/247#issuecomment-5005587550) (worker path in `processing/raster.py`)
- [Follow-up answers](https://github.com/englacial/zagg/issues/247#issuecomment-5005931110) (`time_field` optional for raster — membership is the acquisition's STAC `datetime`, recorded in the manifest; #239 is closed, so `Refs`, not `Closes`)
- [Final ratification](https://github.com/englacial/zagg/issues/247#issuecomment-5007157978) (schedule `none` **is** supported: one bare `{full_id}.zarr` leaf per shard carrying the full time axis, re-run = whole-leaf replacement, D14 `"full"` gated off exactly as aggregation gates it)

## Phases

- [x] **Phase 1 — config: raster + hive becomes legal.** The #239 stopgap rejections and PR A's `output.windowing` raster rejection are removed; raster configs route through the **shared** store_layout/coverage_moc/windowing checks (`_validate_store_layout_keys`, `_validate_windowing_windows`). `windowing.time_field` optional for raster (present only as `datetime`); `epoch`/`scale`/`units` rejected on raster; `get_windowing` records the resolved fixed encoding. Raster + `sharded: true` kept as the **permanent** ratified exclusion.
- [x] **Phase 2 — per-leaf raster templates.** `raster_leaf_spec` + `emit_raster_leaf_template` beside the flat template (shared member helpers, flat spec byte-identical): leaf arrays `(T_leaf, cells_per_shard)`, same dtypes/fills/`(1, cells_per_chunk)` chunking, root group as the D4 stamp target; leaf `time` + `cell_ids` coords written at template time; `write_raster_leaf_slab` at leaf-local indices.
- [x] **Phase 3 — worker write path + commit stamp.** `process_and_write_raster_hive` in `processing/raster.py` — the raster analog of `process_and_write_hive`, shared by both dispatchers: lazy leaf template on first slab, slabs at leaf-local indices, coverage → `stamp_commit` FINAL (D15: window label + actual ISO-UTC acquisition extent + acquisition count). D14 popcount via a new `occupied_out` seam on `process_raster_shard`; `"full"` gated on windowed units; per-timestep validity stays data-plane nodata. Local dispatcher hive branch with pre-dispatch `ensure_manifest`, `_raster_windowed_units` fan-out (group membership = earliest STAC datetime; groups never split across leaves), finalize backstop + root `coverage.moc` with D15 time union.
- [x] **Phase 4 — dispatch fan-out + Lambda hive branch.** `_handle_process_raster` gains the hive branch running the same phase-3 worker (`window` from the event, ISO `time_range` mirrored in the response); hive events carry **no `time_index`** (leaf-local axis) and flat raster process events stay **byte-identical** (pinned by test). The hive manifest rides the #252 lifecycle exactly as aggregation: shared ping (read-only precheck) → async `mode:"setup"` manifest write → per-unit synchronous invokes → `mode:"finalize"` idempotent backstop (ordered **before** the all-failed raise, mirroring local) → fire-and-forget `mode:"coverage"` root union (fail-open, D9). Dispatcher parity is tested end-to-end by driving the **real handler** through a scripted client and byte-comparing leaf sets + stamps against a local-backend run.
- [x] **Phase 5 — tests + docs.** Edge-of-scene popcount fixture at parent 14/child 16 (interior shard: `encoding:"full"`, **no** `coverage.moc` object in the leaf, asserted via listing; swath-edge shard: real bitmap sidecar decoding to the exact occupied set). Window re-run idempotency (sibling leaves byte-untouched, manifest not rewritten) + backfill (earlier window adds a leaf; committed leaf + manifest byte-untouched; root summary unions the ranges). Flat-raster default pin (no hive objects anywhere in a default-layout store). Raster-hive section in `docs/hive_layout.md`. No benchmark object-count arm (rides after this PR per the ratified phasing).

## Merges from main during the run

Main advanced twice while this PR was in flight; both merged in (no rebase), full suite green after each:

1. **`0ed364e`** — main `73296ba → 18cdc74` (#250 raster profiling/benchmark series + #217 arena streaming). Composition: the hive worker's `write` bucket feeds the #250 `worker_stage_max`/`worker_stage_counts` rollups through the same `_fold_raster_stages` seam as the flat sink timing; the #217 arena work touches only the aggregation `process_shard` seam (no raster overlap). Fallout commit `7663f97` flips #250's two **deliberate tripwire tests** in `tests/test_raster_benchmark.py` to the new reality: raster+hive now validates and the harness `store_layout` override rides end-to-end for both layouts — but the pending `raster_s2_neon_2025_hive` benchmark target is **not** promoted here (the hive benchmark arm rides after this PR per the ratified phasing; its promotion is the follow-up's first step).
2. **`a2fa210`** — main `18cdc74 → bcc5014` (#264 raster lambda dispatch via ping + **sync** setup invoke emitting the flat template worker-side, plus #267/#269 infra). Composition: **layout splits before pipeline** — `_handle_setup`'s hive branch (manifest-only, D5) now precedes the #264 raster branch, so a hive-raster setup event writes `morton_hive.json` only, while flat raster setup stays byte-for-byte #264 (load-bearing template → synchronous invoke); hive setup stays the #252 **async** Event invoke (manifest is reader-facing, not load-bearing) with the finalize backstop. Both layouts share #264's pipeline-neutral ping; the flat lifecycle (`ping → sync setup → fan-out`) is pinned by main's own tests, and the stale pre-#264 flat-events pin here was rewritten onto the `_lifecycle` helper.

## How tested

- Per-phase targeted suites plus full-suite runs after phases 1, 3, 5, each review fold, and **each merge** (final: 2317 passed, 32 skipped, 0 failed).
- Highlights: torn-worker debris + wholesale re-run replacement; no-data unit creates no prefix; schedule-`none` bare leaf (`/1` stamps, no extent claim); disjoint-timestep occupied-union pin; both dispatchers producing identical leaf sets/stamps via the real Lambda handler; hive Lambda lifecycle event shapes; finalize backstop runs even when every shard fails; D15 time-range unions at the root on both backends.
- `ruff check` / `ruff format --check` clean (the pre-existing N818 in `src/zagg/registry.py` excepted).

## Questions for review

- **Module size**: `processing/raster.py` is now ~1170 lines (was 886) — past the ~1000-line guidance. Per the ratified plan ([module placement answer](https://github.com/englacial/zagg/issues/247#issuecomment-5005587550)) the worker path lands there deliberately (raster owns its write path; `hive.py` stays the primitive library). The review thread records a concrete split seam (`processing/raster_write.py`) if preferred — left standing for a call.
- **`_raster_windowed_units` landed with phase 3, not 4**: the local dispatcher's hive branch isn't functional for windowed configs without the fan-out, so the helper shipped with the dispatcher switch; phase 4 reuses it for Lambda. Same code surface as planned, one phase earlier.
- **Hive stamp `granule_count`** counts the unit's asset-carrying entries (acquisitions), not raw records — the D15 "acquisition count" reading; the response body still carries `granule_count`/`skipped` with the flat path's semantics.
- **Benchmark tripwire disposition** (merge 1): flipping the two #250 tripwires without promoting the pending hive target keeps the ratified "benchmark arm after PR B" phasing — if promotion should instead ride this PR, it is a two-line manifest change plus deleting one test.


tests/test_raster_pipeline.py
```diff
@@ -88,21 +88,109 @@ def test_band_requires_dtype(self):
...
+        cfg = _raster_config()
+        cfg.output["store_layout"] = "hive"
+        cfg.output["windowing"] = {
+            "schedule": "explicit",
+            "windows": [
+                {"label": "a", "start": "2019-02-01", "end": "2019-01-01"},
+            ],
+        }
+        with pytest.raises(ValueError, match="half-open"):
             validate_config(cfg)
```
espg, line 217: 🤖 *from Claude (review)*

**should-fix (test-coverage):** the explicit-schedule raster path is only exercised on its *rejection* (half-open). The happy path — a valid `schedule: explicit` raster windowing normalizing through `get_windowing` — is untested, and that is the one raster branch in `get_windowing` (config.py:2137-2158) where the `windows` list is actually populated (`declared` built + `iso_utc`-canonicalized, then returned as `windows: declared`). A regression there (e.g. dropping canonicalization, or the raster branch shadowing `declared`) would sail through CI: `test_windowing_validates_and_normalizes` only pins the generative case (`windows: None`).

Suggest adding a sibling to `test_windowing_validates_and_normalizes` with a valid two-window explicit list and asserting the full normalized dict, including the canonicalized `windows` entries — mirroring the point-path assertion so the two normalizations stay locked together.

---

src/zagg/config.py
```diff
@@ -2109,6 +2143,20 @@ def get_windowing(config: PipelineConfig) -> dict | None:
...
+        # fixed encoding any ISO-8601 UTC instant normalizes to (UTC seconds
+        # since the Unix epoch). _validate_windowing rejects the conversion
+        # knobs on raster configs, so nothing here can disagree with it.
+        return {
+            "schedule": block["schedule"],
+            "time_field": "datetime",
+            "epoch": "1970-01-01T00:00:00+00:00",
+            "scale": "utc",
+            "units": "seconds",
+            "windows": declared,
```
espg, line 2162: 🤖 *from Claude (review)*

**nit (docs):** the `get_windowing` docstring above (lines 2117-2129) still describes only the point-path normalization — it says `epoch` "is canonicalized to ISO-8601 UTC strings," which for raster is misleading: on the raster branch `epoch`/`scale`/`units` are *fixed* (Unix-epoch UTC seconds), not derived from the block. A reader diffing the two return dicts would miss that raster hardcodes the encoding. One sentence in the docstring noting the raster branch returns the fixed `datetime`/Unix-epoch encoding would keep the accessor's contract self-describing. The inline comment here is good; the top-level docstring is the gap.

---

src/zagg/config.py
```diff
@@ -709,42 +669,73 @@ def _validate_raster_config(config: PipelineConfig) -> None:
...
+        if (grid or {}).get("type", "healpix") != "healpix":
+            raise ValueError(
+                "output.store_layout: hive requires a healpix grid (hive node names "
+                f"are morton decimal digits; grid type is {(grid or {}).get('type')!r})"
+            )
+        if config.output.get("consolidate_metadata"):
+            raise ValueError(
+                "output.store_layout: hive has no store-root zarr hierarchy to "
+                "consolidate (D5/D12) — drop output.consolidate_metadata"
+            )
```
espg, line 723: 🤖 *from Claude (review)*

**nit (test-coverage):** extracting `_validate_store_layout_keys` newly routes raster configs through two shared cross-checks that have no raster test:

1. **`hive` + `consolidate_metadata: true` rejection** (this line) — now reachable on the raster path, but the matrix only covers `sharded`, `coverage_moc`, and `windowing`. A one-liner asserting `raster + store_layout: hive + consolidate_metadata: true` raises would pin it.
2. **`get_coverage_moc` default-ON for raster hive** — with raster+hive legal, `get_coverage_moc` (config.py:2100) now returns `True` by default for any raster hive config (O9). That is the ratified intent (phase-4 root coverage), but nothing locks it; a test asserting `get_coverage_moc(raster+hive) is True` would guard the default against a later regression.

Both are latent until the phase-2/4 write path lands, so this is coverage hygiene, not a correctness bug — the extraction itself preserves the point-path behavior exactly (same `grid`, same `consolidate_metadata` truthiness check).

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of phase 1 (commit 70cd708, config: raster + hive becomes legal). Verdict: **the phase-1 acceptance criteria are all met and the logic is correct** — I found no blocking issues. Three non-blocking findings below (1 should-fix, 2 nits), all in test/doc coverage rather than behavior.

What I verified against the ratified plan and checked line-by-line:

- **Extraction preserves point-path behavior.** `_validate_store_layout_keys` reproduces the original `validate_config` block exactly — same `grid = config.output.get("grid")` binding (matches line 275), same `store_layout`/`consolidate_metadata`/`coverage_moc` order and messages. The only change (local `consolidate_metadata` var → re-fetch via `config.output.get(...)`) is behavior-identical, and the point path still bool-type-checks `consolidate_metadata` upstream at 355-359. Raster is an early-return branch (265-266), so there is no double-validation.
- **schedule `none` / explicit / stray keys are handled identically.** `schedule: none` returns from the shared early-return (775-786) *before* the raster branch, so it correctly skips the hive/healpix guards and still rejects a stray `windows`. Generative + stray `windows` and the explicit-list grammar both route through the shared `_validate_windowing_windows`, reached from the raster branch (824) the same way the point path reaches it (884).
- **raster + flat + windowing** lands on the shared hive-only guard (793-797) before the raster branch — matches the criterion and `test_windowing_requires_hive_layout`.
- **`get_windowing` raster normalization is consistent with `_validate_windowing`.** Validation rejects `epoch`/`scale`/`units` on raster (817-823) and get_windowing hard-codes the fixed Unix-epoch UTC encoding (2152-2158) — they cannot disagree. `time_field` is optional and fixed to `datetime` on both sides.
- **The `sharded` rejection** is correctly kept and reworded as the permanent ratified exclusion; it fires before the store-layout checks, so `test_hive_plus_sharded_rejected` asserts the sharded message wins (fine — just note the "hive_plus" framing is incidental).

Findings (none blocking):
1. **should-fix** — `tests/test_raster_pipeline.py:184` — no happy-path test for `schedule: explicit` raster windowing normalizing through `get_windowing`; the only raster branch that populates the `windows` list is untested.
2. **nit** — `src/zagg/config.py:2116` — `get_windowing` docstring describes only the point-path normalization; the "epoch is canonicalized" line is misleading for the raster fixed-encoding branch.
3. **nit** — `src/zagg/config.py:723` / `:2100` — the newly-shared `hive + consolidate_metadata` rejection and the raster-hive `get_coverage_moc` default-ON are both reachable but untested.

Everything else (message accuracy, terseness, the `_validate_windowing_windows` split, docstrings on the extracted helpers) reads clean and matches the surrounding code conventions.

---

tests/test_raster_pipeline.py
espg: 🤖 *from Claude*

Folded in 405046e. Added `test_windowing_explicit_list_normalizes` as a sibling to the generative case: a valid two-window `schedule: explicit` raster config, asserting the full normalized dict from `get_windowing` including the canonicalized `windows` list (labels + ISO-8601 UTC start/end), so the explicit-path normalization is now pinned alongside the generative one.

---

src/zagg/config.py
espg: 🤖 *from Claude*

Folded in 9d08a79. Extended the `get_windowing` docstring to note that the raster branch uses the fixed STAC `datetime` field with `epoch`/`scale`/`units` hardcoded to the Unix-epoch UTC-seconds encoding, rather than canonicalizing a declared `epoch` off the block, so the accessor's contract now describes both return shapes.

---

src/zagg/config.py
espg: 🤖 *from Claude*

Folded in ba78f8b. Added two tests to `TestRasterConfigValidation`: `test_hive_consolidate_metadata_rejected` (raster + hive + `consolidate_metadata: true` raises "no store-root zarr hierarchy") and `test_coverage_moc_defaults_on_for_hive` (`get_coverage_moc` returns `True` for raster+hive, `False` for raster flat), pinning both newly-routed shared checks.

---

src/zagg/processing/raster.py
```diff
@@ -872,15 +997,178 @@ def write_raster_coords(store, grid, shard_key: int):
...
+    # order is pinned: dense slabs -> coverage sidecar -> stamp.
+    meta["cells_with_data"] = 0
+    if "store" in box:
+        _t0 = time.time() if profile else None
+        words = occupied[0] if occupied and occupied[0].size else None
+        # D14 popcount: a fully-occupied subtree stamps encoding "full" — no
+        # sidecar PUT. Gated on a windowed unit (/2 stores) so schedule-none
+        # output mirrors aggregation's gate (hive.process_and_write_hive).
+        depth = int(grid.child_order) - int(grid.parent_order)
+        full = window is not None and words is not None and np.unique(words).size == 4**depth
```
espg, line 1137: 🤖 *from Claude (review)*

**should-fix (test coverage).** This is the only place the raster-specific D14 popcount decides `encoding: "full"` vs `"bitmap"` (`np.unique(words).size == 4**depth`, then the no-sidecar branch below). But every phase-2/3 hive test samples a 96×96 tiff that covers only a fraction of the shard, so `valid.sum() < cells_per_shard` and `full` is **always False** — `test_windowed_leaf_stamp_and_coverage`/`test_schedule_none_bare_leaf` both assert `encoding == "bitmap"`. The `full` path (correct `4**depth`, `bitmap=None`, no `write_coverage_sidecar` PUT, `build_coverage(..., full=True)`) ships unverified, yet the ratified acceptance list names "coverage popcount deciding full vs bitmap" explicitly.

Phase 5's checklist does list an "edge-of-scene popcount fixture (full vs bitmap)" — so this is tracked — but flagging that the fixture must exercise the **raster** full branch (a windowed unit whose shard is fully covered by the source raster → assert `stamp["coverage"]["encoding"] == "full"` and that no `coverage.moc` sidecar object was written), not just inherit the aggregation-side coverage tests.

---

tests/test_raster_pipeline.py
```diff
@@ -730,3 +974,204 @@ def time():
...
+        data = np.full((96, 96), value, dtype=np.uint16)
+        _write_tiff(tmp_path / "h0.tif", data)
+        _write_tiff(tmp_path / "h1.tif", np.full((96, 96), 777, dtype=np.uint16))
+        granules = [
+            _entry("g0", {"red": str(tmp_path / "h0.tif")}, T0, time_key="dt-1"),
+            _entry("g1", {"red": str(tmp_path / "h1.tif")}, T1, time_key="dt-2"),
+        ]
+        return cfg, grid, shard, granules, str(tmp_path / "hivestore")
+
+    def test_windowed_leaf_stamp_and_coverage(self, tmp_path):
```
espg, line 1015: 🤖 *from Claude (review)*

**nit (test coverage).** The occupied-cell union is a spatial OR across timesteps (`occupied_acc[valid] = True` per group in `process_raster_shard`), but every hive test here writes the same 96×96 footprint for both timesteps, so `union == single-timestep valid set`. A regression that dropped the OR — e.g. rebinding/overwriting `occupied_acc` instead of OR-ing, or only sampling the last group — would pass all current assertions. Consider a two-timestep case where t0 and t1 occupy **disjoint** cell subsets and assert the stamped coverage decodes to the union of both, which is what pins the D14 "spatial union across acquisitions" contract (and confirms per-timestep validity stays data-plane nodata rather than leaking into coverage).

---

src/zagg/runner.py
```diff
@@ -1515,6 +1684,49 @@ def _windowed_units(cells: list[tuple], windowing: dict, bounds_temporal: dict |
...
+                continue
+            if not e.get("datetime"):
+                raise ValueError(f"raster granule entry {e.get('id')!r} carries no datetime")
+            groups.setdefault(e.get("time_key") or e["datetime"], []).append(e)
+        by_window: dict = {}
+        for _key, items in groups.items():
+            earliest = min(parse_utc(e["datetime"]) for e in items)
+            labels = windows_intersecting(earliest, earliest, schedule, declared)
+            if labels:
+                by_window.setdefault(labels[0], []).extend(items)
```
espg, line 1873: 🤖 *from Claude (review)*

**nit (behavior / divergence from `_windowed_units`).** `windows_intersecting(earliest, earliest, ...)` can return **more than one** label for an `explicit` schedule with overlapping declared windows, and taking `labels[0]` assigns the group to only the **first-declared** window — the group is silently absent from the other overlapping leaves. The aggregation analog `_windowed_units` does the opposite: a record whose span intersects several windows is emitted into *every* one (`span[0] < hi and span[1] >= lo`, no first-match). For generative schedules this never bites (windows are disjoint, `labels` is length-1), so only overlapping explicit lists differ.

This is likely intended for raster's group-atomic model (a datatake shouldn't be duplicated across leaves), but the divergence is undocumented and the docstring's "belongs to *the* window containing its earliest datetime" assumes non-overlap. Either add a one-line note that overlapping explicit windows resolve to the first-declared label (and why), or confirm the drop-from-others behavior is the ratified intent for @espg.

---

src/zagg/processing/raster.py
```diff
@@ -872,15 +997,178 @@ def write_raster_coords(store, grid, shard_key: int):
     return store
 
 
+def process_and_write_raster_hive(
```
espg, line 1017: 🤖 *from Claude (review)*

**nit (module size — offering the decomposition the PR body asked about).** The PR flags `processing/raster.py` at ~1172 lines under §4's ~1000 guideline and asks whether to split. A clean, cohesive seam exists: the entire zarr template/slab/hive-write section — `_raster_array_spec`/`_check_raster_grid`/`_raster_members`, `raster_group_spec`/`raster_leaf_spec`, `emit_raster_template`/`emit_raster_leaf_template`, `write_raster_slab`/`write_raster_leaf_slab`/`write_raster_coords`, `_shard_cell_range`, and `process_and_write_raster_hive` (roughly the "── lean (time, cell) template + slab writer ──" banner onward) — depends on the item-sampling half only through `process_raster_shard`'s public signature. Lifting it to a sibling `processing/raster_write.py` (item sampling + `raster_time_index` stay in `raster.py`) brings both files back under the guideline without an artificial cut. Offered only — the ratified plan deliberately homes the worker path here, so leave-by-design is a fine answer; noting the concrete split so it's on record.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of phases 2–3 (478827a, 78c7022, 6fe3ec5) — the phase-2 leaf-template refactor and the phase-3 `process_and_write_raster_hive` worker + local dispatcher hive branch. Reviewed against the ratified #247 plan and `hive.process_and_write_hive` as the convention baseline.

**Correctness held up.** I specifically checked and found no defect in:
- **Leaf-local time indexing** — `raster_time_index([granules])` builds the unit's own axis, `process_raster_shard` returns `time_index[key]` as `t_idx`, and `write_raster_leaf_slab` writes `arr[t_idx, :]`; T_leaf and `cells_per_shard` line up, and the foreign-manifest guard can't trip since the index is built from the same subset.
- **Occupied-cell union under concurrency** — `occupied_acc[valid] = True` runs in the serial `as_completed` drain loop (not the threaded sink), no await between read and write; the single `occupied_out.append(cells[occupied_acc])` after the loop is race-free.
- **D15 stamp truth** — `time_range` = `[min,max]` of item instants (not group coordinates), rendered at seconds precision; only on windowed units; `meta["time_range"]` feeds the root union; `stamp_commit`'s reversed/parse guards are satisfied.
- **Debris/stamp ordering** — lazy `_leaf()` (no prefix on a no-data unit), a torn slab write propagates out before the `if "store" in box` stamp block (unstamped debris, D4), re-run replaces wholesale; the `warnings` suppression matches the aggregation posture.
- **`full` vs `bitmap` gating and the schedule-none bare leaf** — `full` gated on `window is not None` exactly as aggregation gates it; profile `sample`/`write`/`stages` split matches the flat raster handler's #256 convention; manifest pre-dispatch + idempotent backstop and the fail-open root `coverage.moc` mirror `_run_local`. Flat-path spec is pinned byte-identical and the sharded-reject message change is unasserted.

**Findings (all non-blocking):**
1. **should-fix (test)** — `processing/raster.py` `full`/`bitmap` popcount branch is never exercised: all hive tests are edge shards, so `encoding` is always `"bitmap"` and the raster `4**depth` + no-sidecar path ships unverified. Tracked for phase 5; flagging it must cover the raster full branch specifically.
2. **nit (test)** — the occupied union across timesteps isn't tested with disjoint per-timestep validity; a dropped-OR regression would pass today.
3. **nit (divergence)** — `_raster_windowed_units` takes `labels[0]`, so overlapping `explicit` windows assign a group to only the first-declared leaf, unlike `_windowed_units`' multi-window membership. Likely intended; undocumented.
4. **nit (module size)** — offered a concrete `raster_write.py` split for the ~1172-line file the PR body flagged; leave-by-design is fine.

Nothing here blocks advancing the phases. Inline comments have the specifics.

---

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Already covered — phase 5 (`5c891ea`, pushed after this review snapshotted) added `TestRasterHivePopcount::test_interior_full_no_sidecar_edge_bitmap` in `tests/test_raster_pipeline.py`. It exercises the raster D14 full branch exactly as flagged: a parent-14/child-16 fixture where an interior fully-covered shard (`n_valid == n_cells`) stamps `coverage.encoding == "full"` with no `coverage.moc` object (asserted via `os.listdir(leaf)`) and the reader short-circuits (`read_coverage_bitmap(...) is None`), while a swath-edge shard (`0 < nv < nc`) stamps `bitmap` and writes the real sidecar. No code change needed; the branch is now pinned. Confirmed passing locally (`pytest -k Popcount`).

---

tests/test_raster_pipeline.py
espg: 🤖 *from Claude*

Folded in `950ee3c`. Added `TestRasterHiveWorker::test_disjoint_timesteps_coverage_is_union`: two timesteps (distinct `time_key`s) whose 96×96 footprints occupy disjoint cell subsets of one schedule-none bare leaf — the second raster shifted ~480 m (via `_write_tiff(..., origin=...)` + a shifted transform) so that, within this shard, `(valid0 & ~valid1).any()` and `(valid1 & ~valid0).any()` both hold (neither footprint is a superset). The test asserts `read_coverage_bitmap(leaf, coverage=stamp["coverage"])` equals the sorted OR of both valid cell sets and `stamp["cells_with_data"]` equals the union size, so dropping the `occupied_acc` OR (or sampling only the last group) now fails. One implementation note: I offset west rather than east — the shard's covered patch sits at the raster's east edge, so an east shift pushes t1's exclusive cells off this shard (`valid1 ⊆ valid0`); a west shift keeps both timesteps' exclusive cells inside the shard, which is what the union assertion needs.

---

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded as a doc fix in `d9eb130`. The overlapping-explicit-windows premise can't actually occur: `config._validate_windowing_windows` rejects overlapping declared ranges ("windows must be disjoint half-open ranges"), and generative schedules partition time, so `windows_intersecting(earliest, earliest, ...)` returns at most one label for any schedule this code can see. `labels[0]` is therefore THE containing window, not a first-match tiebreak. I extended `_raster_windowed_units`'s docstring with a sentence stating that invariant (citing `_validate_windowing_windows`) so the `labels[0]` access is self-explaining rather than looking like a silent drop. No behavior change — if you'd rather harden it further (e.g. assert `len(labels) <= 1`), that's a scope call I've left standing for you.

---

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Leaving by design — no restructure in this fold. The ratified plan (issue #247, comment 5005587550) deliberately homes the worker path in `processing/raster.py`, and splitting the zarr template/slab/hive-write section out to a sibling `processing/raster_write.py` is a side-effecting scope change (a new module + import surface) that's espg's call, not a self-review's. The PR body's "Questions for review" already stands the `processing/raster.py` module-size question (~1172 lines vs the §4 ~1000 guideline) for espg. Recording the concrete seam you identified so it's on file for that decision: the `── lean (time, cell) template + slab writer ──` section (`_raster_array_spec`/`_check_raster_grid`/`_raster_members`, the `*_spec`/`emit_*`/`write_*` helpers, `_shard_cell_range`, `process_and_write_raster_hive`) depends on the item-sampling half only through `process_raster_shard`'s public signature, so it lifts cleanly if espg wants the cut. Left as-is pending that.

---

src/zagg/runner.py
```diff
@@ -912,13 +1039,55 @@ def _one(pair):
...
                     timesteps_written += body["timesteps"]
 
         wall_time = time.time() - t0
         if cells and errors == len(cells):
             raise RuntimeError(f"all {errors} raster shard(s) failed; last error: {last_error}")
+        if store_layout == "hive":
+            # Finalize backstop (issue #252 hybrid): idempotent ensure_manifest
+            # worker-side — required reader-facing schema (D6), so a failure
+            # raises, unlike the fail-open coverage dispatch below.
+            _invoke_lambda_finalize(
```
espg, line 1172: 🤖 *from Claude (review)*

**should-fix** — backend divergence on the all-shards-failed path. Here the `if cells and errors == len(cells): raise` on the line just above (1049) fires *before* this hive block, so on a fully-failed Lambda run `_invoke_lambda_finalize` (the manifest backstop) **and** the root `coverage.moc` dispatch are skipped. The raster **local** backend does the opposite — it runs `ensure_manifest(...)` (the manifest backstop) and the coverage union *before* the identical `raise ... all raster shard(s) failed` (see the `if store_layout == "hive":` block near line 817, raise at 851). So the two backends disagree on exactly the corner this PR's parity thesis is about.

It also matters *more* on Lambda than on local: the local pre-dispatch `ensure_manifest` (line 700) is a durable synchronous write, so the post-fan-out backstop is belt-and-suspenders; on Lambda the pre-dispatch manifest write is the fire-and-forget async `mode:"setup"` invoke (retries-0, can be dropped — per `_invoke_lambda_setup_async`'s own docstring), so `finalize` is the *only* reliable manifest backstop. Skipping it on all-error can leave a store with neither leaves nor `morton_hive.json` until a re-run.

Suggest reordering to mirror local: run the `hive` finalize/coverage block, then the all-error `raise` — or, if raising first is deliberate (to surface the shard errors rather than a finalize error), say so, since it's a visible split from the local flavor the docstring calls "exactly as aggregation." One caveat worth weighing: aggregation's Lambda path has no all-error raise and always finalizes, so "exactly as aggregation" argues for finalize-then-decide too.

---

deployment/aws/lambda_handler.py
```diff
@@ -982,7 +982,18 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
...
     try:
-        required = ["shard_key", "granules", "config", "store_path", "time_index"]
+        # The hive path (issue #247) needs no ``time_index``: the worker
+        # builds its own leaf-local index from the dispatched subset. Peek at
+        # the raw config dict (no load yet) so a flat event — including one
+        # with no config at all — reports the flat requirements byte-identical
+        # to before.
+        cfg_dict = event.get("config")
+        hive = isinstance(cfg_dict, dict) and (
+            (cfg_dict.get("output") or {}).get("store_layout") == "hive"
```

**nit** — the hive peek `(cfg_dict.get("output") or {}).get("store_layout")` assumes `config["output"]` is a dict. A malformed event whose `config` is a dict but whose `output` is a truthy non-dict (e.g. `{"config": {"output": "fullsphere"}}`) raises `AttributeError: 'str' object has no attribute 'get'` right here — inside the `try`, so it surfaces as a 500 with a confusing `str(e)` ("'str' object has no attribute 'get'") *before* `load_config_from_dict` can emit its clean validation error. Well-formed dispatcher events always carry a dict `output`, so this is defense-only, but the peek's whole point is to run *before* the load — so a `isinstance(cfg_dict.get("output"), dict)` guard (or `cfg_dict.get("output") if isinstance(...) else {}`) keeps a bad config on the clean load-error path rather than a raw-AttributeError 500. The flat requirements-echo it's protecting stays byte-identical either way.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of **phases 4 & 5** (commits `f7f3496`, `5c891ea`). Two findings inline; the rest of the surface holds up well.

**Verified clean (no finding):**
- **Dispatcher parity.** Both lifecycle events and the process events match aggregation: ping → async `setup` → per-unit synchronous invokes → `finalize` backstop → fire-and-forget `coverage`. Hive events carry no `time_index`; flat events stay byte-identical (the `set(...)` pin is real). Lifecycle events carry `config` + `parent_order` + `dataset={short_name,version}` exactly as the aggregation path (line 2388), and `build_manifest` reads only those two dataset keys, so the two-key projection can't drift the manifest.
- **Manifest windowing on Lambda.** `_handle_setup/_handle_finalize/_handle_ping` all thread `get_windowing(config)` into `build_manifest`, so the Lambda-written manifest is `morton-hive/2` and matches the local one — the parity test's `spec`/`temporal` equality would catch a regression.
- **Required-params peek vs `get_store_layout`.** Null/absent `store_layout` → flat; the peek treats only the literal `"hive"` as hive, agreeing with the default pin. Flat/config-less 400 error string keeps `time_index` (pinned).
- **Validation ordering.** The dense-layout rejection sits before the hive/flat fork, so the hive branch does not skip it.
- **Coverage dispatch conditions.** Fired only when `done` is non-empty (≥1 unit wrote); `build_root_coverage` de-dups shard keys via `np.unique`, so multiple windows on one shard collapse correctly; `union_time_range` drops the `None`s from no-data/unwindowed units. `cells_with_data` counts a boolean-masked subset of `cells`, so `words.size` is already unique — no over-count.
- **Popcount fixture (phase 5).** Interior `encoding:"full"` + no `coverage.moc` object (asserted via listing) and edge bitmap decoding to the exact occupied set; both premises are asserted with fixture-drift guards (`n_valid == n_cells`, `edge is not None`).
- **Idempotency/backfill (phase 5).** Re-run replaces leaf B wholesale, leaves sibling A and `morton_hive.json` byte-untouched; backfill unions the root summary across runs via the GET-union-PUT (`[T0,T1]`). Flat default pin asserts no hive objects anywhere.
- **Docs.** The new `hive_layout.md` raster section matches the code: `epoch/scale/units` and non-`datetime` `time_field` are genuinely *rejected* (`config.py` ~799–820), `sharded:true` permanently excluded (~671), schedule-`none` gate + `"full"` gating, and the D15 stamp-truth wording all check out.
- **Scope.** Only `deployment/aws/lambda_handler.py` was touched under `deployment/aws/` (issue #247 names the handler), and only `_handle_process_raster` within it.

Not re-raising the standing module-size question or the phase-2/3 threads.

---

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded in `3033bfc`. Reordered so the hive `finalize`/coverage block runs BEFORE the all-shards-failed raise, mirroring the local backend (where `ensure_manifest` + root coverage run ahead of the identical raise). The raise message/behavior is byte-identical, just later. As you noted, this matters more on Lambda: the pre-dispatch manifest write is the droppable retries-0 async `setup` invoke, so `finalize` is the only reliable manifest write and must run even on all-error. With every shard failed `done` is empty, so no coverage invoke fires (kept that natural gating); the finalize backstop still runs. Added a comment on the ordering, and pinned it with `TestRasterHiveLambdaBackend::test_all_failed_finalizes_before_raise`: all `process_raster` events 500 (two daily windows), asserts `RuntimeError` (match `boom`) still propagates AND the recorded modes carry `finalize` after the process events with no `coverage` event.

---

deployment/aws/lambda_handler.py

Folded in `0d80df3`. Made the pre-load hive peek shape-safe: `output = cfg_dict.get("output") if isinstance(cfg_dict, dict) else None` then `hive = isinstance(output, dict) and output.get("store_layout") == "hive"`. A truthy non-dict `output` (e.g. a string) now reads as non-hive and falls through to the flat requirements / the normal `load_config_from_dict` error path instead of raising `AttributeError` → opaque 500. The flat requirements-echo stays byte-identical. Pinned with `TestProcessRasterHiveMode::test_malformed_output_falls_through_to_flat`: an event with `config: {"output": "hive"}` (string `output`) and no other keys returns a 400 naming the flat requirements (`time_index` included), not a 500.

---

PR #266, opened 2026-07-17
Raster template behind the sync setup invoke (issue #264)

### LINKED ISSUE #264: Raster lambda path writes the store template from the orchestrator — move it behind the setup invoke
🤖 *from Claude*

## Summary

The raster lambda path writes the store template **from the orchestrator** (`RasterStrategy.run` → `emit_raster_template`, `runner.py:729`), which espg confirms is wrong: on the lambda backend the orchestrator should only ever invoke. The 0.32.0 per-release benchmark's new raster leg (#250 / PR #261) exposed it: the CI OIDC role (`zagg-benchmark`) is **deliberately lambda-invoke-only** (the #252 ratification: "CI/CD runs depend on OIDC with lambda-invoke-only permissions, so the orchestrator gains no S3 write access on the lambda path"), so the template PUT dies with `AccessDenied` before any worker runs — [failing job](https://github.com/englacial/zagg/actions/runs/29620830771/job/88015539402). Every prior raster fleet run was dispatched from espg's machine, whose credentials can write S3, so the path was never exercised under the invoke-only role.

## Fix (mirrors the flat point-path setup — #252 lifecycle)

The raster template is **load-bearing before fan-out** (workers `set_basic_selection` into its arrays), which is the flat point path's exact situation — so the fix is the *sync* setup invoke, not the #255 async hybrid (that applies only when the artifact isn't load-bearing, like the hive manifest):

1. **Handler** (`deployment/aws/lambda_handler.py`, explicitly in scope per this issue): extend `_handle_setup` with a raster branch — the event's config identifies the raster pipeline; the branch rebuilds the grid and calls `emit_raster_template(store, grid, config, times_us, overwrite)`. `times_us` is catalog-derived and unavailable in the handler, so it rides in the event payload (int64 list, JSON-safe).
2. **Runner** (`src/zagg/runner.py`): on the lambda backend, `RasterStrategy.run` replaces the direct `emit_raster_template` call with the sync setup invoke (reuse the `_invoke_lambda_setup` machinery/event shape), preceded by the `mode="ping"` stale-deploy guard the point path already runs. Local backend unchanged (direct write is correct there).
3. **Audit the rest of the raster lambda path for orchestrator-side store writes** — status markers (`*.zarr.status/`), finalize/consolidation, anything else that PUTs from the dispatcher — and move each behind an invoke or confirm it is already worker/finalize-side. The template was the first 403; it must be the last.
4. **Tests**: wire-level pin of the raster setup event (the `TestInvokeLambdaSetupEvent` precedent), handler-branch test, and a strategy-level ordering test (ping → setup → fan-out; no orchestrator store writes on the lambda path).

## Acceptance

- The #261 raster benchmark leg completes from the `zagg-benchmark` OIDC role with no `s3:PutObject` from the orchestrator.
- Flat/point setup and hive lifecycle events stay byte-identical (pinned).
- Local raster runs byte-identical to today.

Refs: #250 / PR #261 (raster leg), #252 / PR #255 (lifecycle + invoke-only ratification), #249 (raster stage profiling), [failing run](https://github.com/englacial/zagg/actions/runs/29620830771).

espg:
Closes #264

Moves the raster (time, cells) template write off the orchestrator and behind the synchronous `mode="setup"` Lambda invoke, mirroring the flat point-path setup lifecycle from [issue #252](https://github.com/englacial/zagg/issues/252). The 0.32.0 raster benchmark leg (#250 / PR #261) exposed the gap: the CI OIDC `zagg-benchmark` role is deliberately lambda-invoke-only, so `RasterStrategy.run`'s direct `emit_raster_template` PUT died with `AccessDenied` before any worker ran ([failing job](https://github.com/englacial/zagg/actions/runs/29620830771/job/88015539402)). The raster template is load-bearing before fan-out (workers write slabs into its arrays), so the fix is the **sync** setup invoke — not the #255 async hybrid, which applies only to non-load-bearing artifacts like the hive manifest (see the [PR #255 lifecycle rationale](https://github.com/englacial/zagg/pull/255)).

## What / approach

**Handler (`deployment/aws/lambda_handler.py` — explicitly in scope per issue #264):** `_handle_setup` gains a raster branch keyed on `data_source.reader == "raster"` (checked before the hive branch): it rebuilds the grid via `from_config(config)` — exactly as the `mode="process_raster"` worker does, so template chunking cannot drift from worker writes (the issue #99 lesson) — and calls `emit_raster_template(store, grid, config, times_us, overwrite)`. `times_us` is catalog-derived and unavailable in the handler, so it rides in the event as an int64 list. The success body echoes `"pipeline": "raster"`, giving the dispatcher a stale-deployment guard (the PR #205 layout-echo precedent): a deployed function without this branch would fall through to the point-path `grid.emit_template` and write the wrong template — the echo check turns that into a loud pre-fan-out redeploy error. Flat and hive setup handling is byte-identical (bodies pinned in new tests, wire events pinned in the existing `TestInvokeLambdaSetupEvent`).

**Runner (`src/zagg/runner.py`):** on the lambda backend, `RasterStrategy.run` no longer opens the store or emits the template orchestrator-side. `_run_lambda_shards` now runs `_invoke_lambda_ping` (the issue #252 stale-deploy guard — a pre-#252 function 400s the ping with zero writes; the hive manifest precheck is a no-op for a raster config) followed by the new synchronous `_invoke_lambda_raster_setup` before fan-out; `template_s` brackets the ping + setup invokes, keeping the summary schema and the `raster_series.parquet` column semantics ("setup-ish wall before fan-out") intact. The local backend is byte-identical to today (direct `emit_raster_template` at init, correct where the operator's credentials can write S3).

**Stale-deployment posture** (the "degrade loudly" acceptance): two nested guards. A pre-#252 function falls through the ping to a 400 with zero writes → redeploy error before any invoke that writes. A #252-capable but pre-#264 function passes the ping but its flat setup branch answers without the `pipeline: "raster"` echo → redeploy error before any worker is dispatched (the error text also says to rerun with overwrite, since the stale function's flat branch may have written point-path template objects at the root before responding).

## Phases

- [x] Phase 1 — handler raster-setup branch + handler tests + flat/hive body pins (d14226f)
- [x] Phase 2 — runner: ping → sync setup invoke on the lambda backend, wire-level event pin + lifecycle-order test, local path byte-identical (413c505)
- [x] Phase 3 — audit the entire raster lambda path for orchestrator-side store writes (report below; no code change needed beyond phases 1–2)
- [x] Phase 4 — docs: raster benchmark-leg lifecycle text in `docs/deployment/benchmark.md` (93717b7)

Review round 1 (phase 1) folded: reject empty `times_us` (5066b20), clean 400 on dense layout in the setup branch (b8aa7da), refuse `overwrite=False` time-coordinate clobber at unchanged length (813edb8). Review round 2 (phase 2) folded: de-hive the shared ping wording (3ce2b41), redeploy hint on setup non-200 (6aa3d82), `times_us` sync-payload-cap note (7f56513); one finding left standing by design — the setup echo guard is post-write on a stale pre-#264 function (see "Stale-deployment posture" above and the inline thread; the window closes permanently once the fleet redeploys).

## Orchestrator-write audit (phase 3)

Swept `agg` → `RasterStrategy.run` → `_run_lambda_shards` and every helper on the raster lambda path for orchestrator-side store writes. **Result: none remain.** The template was the only one; specifically:

- **Template** — was the direct `emit_raster_template` at `runner.py:729`; now behind the sync setup invoke (phases 1–2). The lambda branch returns before `open_store` is ever called; the lifecycle test guards this (`test_lifecycle_order_and_setup_event_pinned` monkeypatches `open_store` to assert no `s3://` open).
- **Status markers (`*.zarr.status/`)** — not used on the raster path: shard fan-out is synchronous `RequestResponse` with no `result_url` key in the event (`_event` in `_run_lambda_shards`), so the issue #151 result-object channel (worker-written anyway, not orchestrator-written) never engages. The `.status/` prefixes in `runner.py` (`_run_lambda`, `_run_lambda_events`) belong to the spatial/temporal async paths.
- **Finalize / consolidation** — the raster path has no finalize leg at all: `_run_lambda_shards` returns the summary straight after fan-out; `consolidate_metadata` sites in `runner.py` are on the Spatial strategy only, and `validate_config` rejects `consolidate_metadata` on raster configs (issue #239).
- **Coverage artifacts (`coverage.moc`)** — hive-only dispatch (`_invoke_lambda_coverage` under the hive finalize block of the spatial path), and raster rejects `store_layout` entirely (issue #239); also already worker-side by design (issue #200).
- **Everything else on the path is read-only or in-memory:** `_load_catalog` (local file), `_select_cells`, `_check_signature`, `raster_time_index`, `from_config`, `_resolve_function_name`, `_build_output_creds_event`, `_dry_run_summary`, `raise_for_fd_exhaustion`.

The `docs/deployment/benchmark-cicd.md` claim "**No S3** — the Zarr template write happens inside the Lambda, so the orchestrator never touches the bucket" was previously true only for the point leg; it now holds for the raster leg too (noted in the phase-4 doc edit).

## Testing

- Full `uv run pytest -q`: **2275 passed, 32 skipped** (post-phase-2; fold commits re-ran the touched suites green: 224 passed across the raster/handler/hive files). One deselected failure, **pre-existing and environmental**: `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` shells out to `build_function.sh`, whose pip resolved against a Python 3.10 interpreter on this machine (cp310 wheels → `zarr>=3.1.5` unresolvable). Untouched by this PR — flagged, not fixed.
- `uv run ruff check src tests` / `ruff format --check` green on all touched files (known pre-existing N818 on `src/zagg/registry.py:64` left alone, as flagged on PR #255).
- Coverage highlights: exact wire pin of the raster setup event + `RequestResponse` invocation type; lifecycle order `ping → setup → process_raster` at the strategy level; stale-ping and stale-setup-echo fail-fast with zero subsequent invokes; creds/endpoint threading on all three lifecycle events; flat/hive setup events and bodies pinned byte-identical; handler setup→worker end-to-end proving workers slab-write into the setup-emitted template; local-path behavior unchanged (existing local tests untouched and green).

## Questions for review

- `template_s` on lambda rows now measures ping + sync setup invoke wall (previously the orchestrator-local template write). Same "setup-ish wall before fan-out" semantics the column documents, so `raster_series.parquet` history stays comparable — flagging in case a stricter split (ping vs setup) is wanted in the series.

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -679,8 +679,19 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
         if (config.data_source or {}).get("reader") == "raster":
             import numpy as np
 
+            from zagg.config import get_layout
             from zagg.processing.raster import emit_raster_template
 
+            if get_layout(config) == "dense":
+                # Symmetry with the worker (_handle_process_raster): dense keys
+                # blocks by populated-shard position, unreconstructable from an
+                # invoke-only setup event, so from_config below would raise an
+                # opaque RuntimeError -> 500. Reject it with the worker's clean
+                # 400 instead. RasterStrategy.run rejects dense pre-dispatch too
+                # (runner.py), so this is defense-in-depth (issue #264).
+                error_msg = "raster lambda workers require output.grid.layout: fullsphere"
+                logger.error(error_msg)
+                return {"statusCode": 400, "body": json.dumps({"error": error_msg})}
             store = open_store(event["store_path"], **_output_store_kwargs(event))
             grid = from_config(config)
             times_us = np.asarray(event["times_us"], dtype=np.int64)
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -45,7 +45,9 @@ hive-layout config -- output.store_layout: hive, issue #199 -- it writes the
 morton_hive.json manifest instead, and each process-mode worker emits its own
 leaf template. Current dispatchers send hive setup as a fire-and-forget Event
 invoke right after the ping -- the primary manifest write, issue #252 hybrid
--- and older synchronous dispatchers keep working against this function):
+-- and older synchronous dispatchers keep working against this function. For
+a raster-pipeline config -- data_source.reader: raster, issue #264 -- it
+writes the raster (time, cells) template instead, from a synchronous invoke):
 {
     "mode": "setup",
     "store_path": str,
@@ -57,6 +59,10 @@ invoke right after the ping -- the primary manifest write, issue #252 hybrid
     "dataset": dict (optional, hive only) -- {"short_name", "version"} identity
         block for the manifest, sourced from the ShardMap metadata by the
         orchestrator (matching the local dispatcher). Absent on flat runs.
+    "times_us": [int, ...] (raster only, issue #264) -- the catalog-derived
+        time coordinate, int64 microseconds since the epoch; the orchestrator
+        owns the global timestep index and threads it here so the template
+        write needs no S3 access from the dispatcher.
     "output_credentials": dict (optional, same shape as process mode),
 }
 
@@ -639,6 +645,16 @@ def _handle_extract(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
 def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
     """Create the zarr template at ``event['store_path']``.
 
+    For a raster-pipeline config (``data_source.reader: raster``, issue #264)
+    this writes the ``(time, cells)`` raster template + its ``time``
+    coordinate via ``emit_raster_template`` — the orchestrator dispatches it
+    as a synchronous setup invoke before fan-out (the template is
+    load-bearing: workers write slabs into its arrays), so an invoke-only
+    dispatcher (the CI OIDC role) never needs S3 write access. ``times_us``
+    (the catalog-derived int64 time coordinate) rides in the event; the
+    success body echoes ``"pipeline": "raster"`` so the dispatcher can refuse
+    a stale deployment that fell through to the point-path template below.
+
     For a hive-layout config (issue #199 phase 3) template time writes ONLY
     the ``morton_hive.json`` manifest — no global zarr template exists (zero
     metadata above the leaves, D5); each worker emits its own leaf template.
@@ -660,6 +676,28 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
     logger.info(f"Setup mode: creating template at {event.get('store_path')}")
     try:
         config = load_config_from_dict(event["config"])
+        if (config.data_source or {}).get("reader") == "raster":
+            import numpy as np
+
+            from zagg.processing.raster import emit_raster_template
+
+            store = open_store(event["store_path"], **_output_store_kwargs(event))
+            grid = from_config(config)
+            times_us = np.asarray(event["times_us"], dtype=np.int64)
+            emit_raster_template(
+                store, grid, config, times_us, overwrite=event.get("overwrite", False)
+            )
+            return {
+                "statusCode": 200,
+                "body": json.dumps(
+                    {
+                        "ok": True,
+                        "mode": "setup",
+                        "pipeline": "raster",
+                        "timesteps": int(times_us.size),
+                    }
+                ),
+            }
         if get_store_layout(config) == "hive":
             from zagg.config import get_windowing
             from zagg.hive import build_manifest, ensure_manifest
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -684,6 +684,14 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
             store = open_store(event["store_path"], **_output_store_kwargs(event))
             grid = from_config(config)
             times_us = np.asarray(event["times_us"], dtype=np.int64)
+            if times_us.size == 0:
+                # A zero-timestep template is degenerate: the arrays get a
+                # 0-length time axis no worker can slab-write into.
+                # RasterStrategy.run refuses an empty catalog before it would
+                # dispatch (runner.py); guard the load-bearing invoke-only
+                # writer too, so a hand-rolled or drifted event can't write an
+                # unusable success-shaped store (issue #264).
+                raise ValueError("raster setup received empty times_us (no timesteps to template)")
             emit_raster_template(
                 store, grid, config, times_us, overwrite=event.get("overwrite", False)
             )
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -3278,7 +3278,17 @@ def _invoke_lambda_raster_setup(
         raise RuntimeError(f"Lambda raster setup failed: {payload}")
     result = json.loads(payload)
     if result.get("statusCode") != 200:
-        raise RuntimeError(f"Lambda raster setup error: {result.get('body')}")
+        # A pre-#264 function reaches the flat branch and calls
+        # grid.emit_template on the raster config, which can raise (a raster
+        # template has no point-path shape) → 500. A 500 is indistinguishable
+        # from a genuine failure here, so surface the redeploy hint alongside
+        # the body rather than swallowing the stale-deployment case.
+        raise RuntimeError(
+            f"Lambda raster setup error: {result.get('body')} — if this is a "
+            f"pre-#264 deployment, its flat point-path branch may have raised "
+            f"on the raster config; redeploy the function, then rerun with "
+            f"overwrite to replace anything it wrote"
+        )
     try:
         body = json.loads(result.get("body") or "{}")
     except (TypeError, ValueError):
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -3264,6 +3264,11 @@ def _invoke_lambda_raster_setup(
         "store_path": store_path,
         "overwrite": overwrite,
         "config": config_dict,
+        # times_us rides inline in this sync RequestResponse payload (6 MB
+        # cap, not the 256 KB async/Event cap); int64 μs stamps serialize at
+        # ~17 B each → ~350K-timestep headroom, orders above any real catalog
+        # (the pinned NEON benchmark catalog is 85 items). No chunking
+        # fallback — a pathological series would surface as a boto ClientError.
         "times_us": [int(t) for t in times_us],
     }
     if output_creds_event is not None:
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -633,8 +633,11 @@ class RasterStrategy:
     simple transport: synchronous invokes with transient-only retries; the
     issue-151 async result-object channel and the preflight concurrency probe
     are follow-ups (raster shards are seconds of COG windows, well inside NAT
-    idle limits). Either way the runner owns the template + global time index
-    before fan-out (the single-writer append design).
+    idle limits). Either way the runner owns the global time index before
+    fan-out (the single-writer append design); the template write is
+    orchestrator-side on the local backend but rides the sync setup invoke on
+    lambda (issue #264 — the dispatcher may hold an invoke-only role with no
+    S3 write access, e.g. the CI OIDC benchmark role).
 
     Summary schema note: ``total_obs`` is shared with the spatial/temporal
     strategies so a caller sees one summary shape across pipeline kinds. On the
@@ -716,26 +719,19 @@ class RasterStrategy:
             raise ValueError("catalog carries no raster granule entries (no assets/datetime)")
 
         resolved_endpoint = output_endpoint_url or get_output_endpoint_url(config)
-        zarr_store = open_store(
-            store_path,
-            region=region,
-            credentials=output_credentials,
-            endpoint_url=resolved_endpoint,
-        )
-        # Template emission is orchestrator-owned; time it (always-on, the
-        # issue #180 bracket convention) so the benchmark harness can split
-        # setup-ish wall from the fan-out (issue #250).
-        template_t0 = time.time()
-        emit_raster_template(zarr_store, grid, config, times_us, overwrite=overwrite)
-        template_s = time.time() - template_t0
-
         if backend == "lambda":
+            # No orchestrator-side store write on the lambda path (issue
+            # #264): the template rides the sync setup invoke inside
+            # _run_lambda_shards, so the dispatcher needs only
+            # lambda:InvokeFunction (the CI OIDC invoke-only role).
             return self._run_lambda_shards(
                 config,
                 cells,
                 time_index,
                 grid,
                 store_path,
+                times_us=times_us,
+                overwrite=overwrite,
                 max_workers=max_workers,
                 region=region,
                 function_name=_ignored.get("function_name"),
@@ -743,8 +739,21 @@ class RasterStrategy:
                 output_credentials=output_credentials,
                 output_endpoint_url=resolved_endpoint,
                 profile=profile,
-                template_s=template_s,
             )
+
+        zarr_store = open_store(
+            store_path,
+            region=region,
+            credentials=output_credentials,
+            endpoint_url=resolved_endpoint,
+        )
+        # Local backend: template emission stays orchestrator-owned (the
+        # process writes the store directly); time it (always-on, the issue
+        # #180 bracket convention) so the benchmark harness can split
+        # setup-ish wall from the fan-out (issue #250).
+        template_t0 = time.time()
+        emit_raster_template(zarr_store, grid, config, times_us, overwrite=overwrite)
+        template_s = time.time() - template_t0
 
         source = config.data_source or {}
         src_kwargs = {
@@ -858,6 +867,8 @@ class RasterStrategy:
         grid,
         store_path,
         *,
+        times_us,
+        overwrite,
         max_workers,
         region,
         function_name,
@@ -865,15 +876,19 @@ class RasterStrategy:
         output_credentials,
         output_endpoint_url,
         profile=False,
-        template_s=None,
     ):
         """Fan shards out, one synchronous ``mode="process_raster"`` invoke each.
 
-        The worker gets the shard's ShardMap entries plus only its own slice of
-        the global time index (the template — runner-emitted before this — is
-        the shared truth for the full axis). Transient invoke faults retry with
-        backoff; a Lambda ``FunctionError`` or non-200 envelope is a shard
-        error (per-shard isolation, all-error raise as on the local backend).
+        Before fan-out the lifecycle is ping → sync setup (issue #264): the
+        lightweight ``mode="ping"`` fails fast on a pre-#252 deployment with
+        zero writes, then ``_invoke_lambda_raster_setup`` writes the template
+        WORKER-SIDE — the orchestrator never PUTs to the store, so an
+        invoke-only role (the CI OIDC benchmark role) dispatches cleanly. The
+        worker gets the shard's ShardMap entries plus only its own slice of
+        the global time index (the setup-emitted template is the shared truth
+        for the full axis). Transient invoke faults retry with backoff; a
+        Lambda ``FunctionError`` or non-200 envelope is a shard error
+        (per-shard isolation, all-error raise as on the local backend).
         """
         import boto3
         from botocore.config import Config
@@ -901,6 +916,30 @@ class RasterStrategy:
         output_creds_event = _build_output_creds_event(
             output_credentials, output_endpoint_url, region
         )
+        # Ping → sync setup, bracketed as template_s (the setup-ish wall the
+        # benchmark splits from fan-out, issue #250). The template is
+        # load-bearing before fan-out — workers write slabs into its arrays —
+        # so the setup invoke is synchronous (the flat point-path lifecycle,
+        # NOT the #252 async hybrid, whose artifact is not load-bearing).
+        template_t0 = time.time()
+        _invoke_lambda_ping(
+            client,
+            function_name,
+            store_path,
+            config_dict=config_dict,
+            overwrite=overwrite,
+            output_creds_event=output_creds_event,
+        )
+        _invoke_lambda_raster_setup(
+            client,
+            function_name,
+            store_path,
+            config_dict=config_dict,
+            times_us=times_us,
+            overwrite=overwrite,
+            output_creds_event=output_creds_event,
+        )
+        template_s = time.time() - template_t0
 
         def _event(shard_key, granules):
             keys = {e.get("time_key") or e.get("datetime") for e in granules if e.get("assets")}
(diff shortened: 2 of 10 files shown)


deployment/aws/lambda_handler.py
```diff
@@ -660,6 +676,28 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
...
     try:
         config = load_config_from_dict(event["config"])
+        if (config.data_source or {}).get("reader") == "raster":
+            import numpy as np
+
+            from zagg.processing.raster import emit_raster_template
+
+            store = open_store(event["store_path"], **_output_store_kwargs(event))
+            grid = from_config(config)
+            times_us = np.asarray(event["times_us"], dtype=np.int64)
```
espg, line 697: 🤖 *from Claude (review)*

[non-blocking] The handler accepts an **empty `times_us`** and silently writes a degenerate zero-timestep template, returning 200 with `"timesteps": 0`. Verified locally: `emit_raster_template(store, grid, cfg, np.asarray([], np.int64))` succeeds, producing `time.shape == (0,)` and `red.shape == (0, 51539607552)`.

The runner refuses this loudly — `RasterStrategy.run` raises `ValueError("catalog carries no raster granule entries (no assets/datetime)")` (runner.py:717) before it would ever dispatch. But on the invoke-only lambda path this handler branch **is** the load-bearing writer, and it has no equivalent guard. A hand-rolled or edge-case payload with `times_us: []` gets a success-shaped body and an unusable store. Consider mirroring the runner's empty-check here (raise → 500), and pin it — `TestRasterSetupMode` covers missing `times_us` (KeyError→500) but not empty.

deployment/aws/lambda_handler.py
```diff
@@ -660,6 +676,28 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
...
+            emit_raster_template(
+                store, grid, config, times_us, overwrite=event.get("overwrite", False)
```
espg, line 707: 🤖 *from Claude (review)*

[non-blocking] **Silent time-coord overwrite on `overwrite=False` when the timestep COUNT is unchanged but the values differ.** `emit_raster_template` runs `spec.to_zarr(store, ..., overwrite=False)` and then unconditionally `arr[:] = times_us`. `to_zarr(overwrite=False)` only raises `ContainsGroupError` when the spec differs (e.g. a different `n_time` → different shape); when the store already holds a same-length time axis it is a no-op, so the subsequent `arr[:] = times_us` silently rewrites the `time` coordinate with the new (shifted) timestamps — no overwrite flag required.

`TestRasterSetupMode.test_setup_rerun_overwrite_semantics` (and the local `test_overwrite_and_rerun` it claims parity with) only exercise a CHANGED count (2→1), which trips the shape guard; neither covers same-count/different-values. On this invoke-only path a rerun over a different catalog with the same datatake count would leave the persisted `time` coord mismatched against the `time_index` the workers slab-write by. This is inherited from `emit_raster_template` (shared with the local path, so out of strict phase-1 scope) — flagging because the PR advertises overwrite-refuse safety and this is the one gap the tests miss.

deployment/aws/lambda_handler.py
```diff
@@ -660,6 +676,28 @@ def _handle_setup(event: Dict[str, Any]) -> Dict[str, Any]:
     logger.info(f"Setup mode: creating template at {event.get('store_path')}")
     try:
         config = load_config_from_dict(event["config"])
+        if (config.data_source or {}).get("reader") == "raster":
+            import numpy as np
+
+            from zagg.processing.raster import emit_raster_template
+
+            store = open_store(event["store_path"], **_output_store_kwargs(event))
+            grid = from_config(config)
```
espg, line 696: 🤖 *from Claude (review)*

[nit] **Dense-layout rejection asymmetry vs the worker.** This branch calls `from_config(config)` with no `populated_shards`, so a raster config carrying `output.grid.layout: dense` raises an opaque `RuntimeError("HealpixGrid(layout='dense') requires populated_shards before n_shards")` → caught → 500 (verified locally). The sibling worker `_handle_process_raster` guards the same case explicitly and returns a clean 400: `"raster lambda workers require output.grid.layout: fullsphere"` (lambda_handler.py:1050). `RasterStrategy.run` also rejects dense before invoking (runner.py:698), so this is defense-in-depth only and cannot fire on a runner-dispatched event — but the two handler branches diverge in message quality. Consider the same explicit dense guard here for symmetry.

espg:
🤖 *from Claude (review)*

Adversarial review of phase 1 (d14226f): **3 findings, all non-blocking** (2 non-blocking, 1 nit), posted inline on `deployment/aws/lambda_handler.py`. Summary: empty `times_us` silently accepted (handler lacks the runner's guard); same-count/different-values time-coord silently overwritten even at `overwrite=False` (inherited from `emit_raster_template`, test gap); dense-config rejection asymmetry vs the worker branch.

Verified clean: grid construction matches the worker (`from_config(config)`, no `populated_shards` — identical for fullsphere, drift-free); `times_us` int64 serialization is sound; raster branch ordering before hive is safe (raster+hive is config-rejected per #239); flat/hive setup bodies stay byte-identical (new pins hold); `TestRasterSetupMode` passes locally (4/4) and the end-to-end setup→worker test genuinely exercises slab writes. No handler-side blocker for the phase-2 runner wiring.

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Fixed in 5066b20. Added an explicit `times_us.size == 0` guard in the raster setup branch that raises `ValueError("raster setup received empty times_us (no timesteps to template)")` before `emit_raster_template` runs, so the invoke-only writer refuses the degenerate zero-timestep template the same way `RasterStrategy.run` refuses an empty catalog pre-dispatch. It rides `_handle_setup`'s existing try/except to a 500 — matching this function's error style (no explicit 400s; the sibling `test_setup_missing_times_us_500` is also a 500). Pinned with `TestRasterSetupMode.test_setup_empty_times_us_500` (500 + `"times_us"` in the error body), the gap next to the existing missing-key test.

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Fixed in b8aa7da. Mirrored the worker's dense-layout guard in the raster setup branch: an explicit `get_layout(config) == "dense"` check now returns the same clean 400 (`"raster lambda workers require output.grid.layout: fullsphere"`) before `from_config` can raise its opaque RuntimeError -> 500. Same message as `_handle_process_raster` so the two handler branches stay symmetric. Pinned with `TestRasterSetupMode.test_setup_dense_layout_400`, parallel to the worker's `test_dense_layout_400`.

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Fixed in 813edb8 (in `emit_raster_template`, so both the local and invoke-only paths inherit it). When `overwrite=False`, before writing the coordinate it now opens the existing `time` array (if any) and, when a same-length axis holds values that differ from `times_us`, raises `ContainsGroupError` — the same exception type `to_zarr(overwrite=False)` already raises on a COUNT/shape mismatch, so overwrite=False now *uniformly* refuses a differing template (the contract `test_overwrite_and_rerun` states). A fresh store (`ArrayNotFoundError` on the open) and an identical-values rerun both stay no-ops; empty-times templates still emit.

Pinned both paths:
- local: `test_raster_pipeline.py::test_overwrite_false_refuses_same_count_different_values` (idempotent same-values, refuse same-count/shifted with the store left intact, then overwrite=True rewrites)
- handler: `TestRasterSetupMode::test_setup_rerun_same_count_different_values_500` (500 on the shifted rerun, persisted coord unchanged, overwrite=True -> 200)

You're right it was inherited shared behavior rather than phase-1-introduced; folded it here since it's the exact gap the PR's advertised overwrite-refuse safety leaves open and the fix is contained to `emit_raster_template`.

src/zagg/runner.py
```diff
@@ -901,6 +916,30 @@ def _run_lambda_shards(
         output_creds_event = _build_output_creds_event(
             output_credentials, output_endpoint_url, region
         )
+        # Ping → sync setup, bracketed as template_s (the setup-ish wall the
+        # benchmark splits from fan-out, issue #250). The template is
+        # load-bearing before fan-out — workers write slabs into its arrays —
+        # so the setup invoke is synchronous (the flat point-path lifecycle,
+        # NOT the #252 async hybrid, whose artifact is not load-bearing).
+        template_t0 = time.time()
+        _invoke_lambda_ping(
```
espg, line 925: 🤖 *from Claude (review)*

[non-blocking] The raster path reuses `_invoke_lambda_ping`, but that helper's stale-deployment failure is worded for hive: it raises "predates the issue #252 hive dispatch lifecycle — redeploy the function before dispatching **hive** runs" (line 3377) and logs "Hive preflight OK" on success (line 3383). A raster operator hitting a pre-#252 function now gets hive-specific guidance for a raster run — message/observability drift introduced by the reuse. `test_stale_ping_fails_fast_no_setup` only asserts `match="redeploy"`, so it won't catch the wording. Consider a pipeline-neutral message (or a raster-aware variant), or at least note the drift.

src/zagg/runner.py
```diff
@@ -3191,6 +3230,68 @@ def _invoke_lambda_setup_async(
...
+        FunctionName=function_name,
+        InvocationType="RequestResponse",
+        Payload=json.dumps(event),
+    )
+    payload = response["Payload"].read().decode("utf-8")
+    if response.get("FunctionError"):
+        raise RuntimeError(f"Lambda raster setup failed: {payload}")
+    result = json.loads(payload)
+    if result.get("statusCode") != 200:
+        raise RuntimeError(f"Lambda raster setup error: {result.get('body')}")
```

[non-blocking] The nice "redeploy the function" remedy (lines 3286-3292) only fires on a **200** whose body lacks the `pipeline: raster` echo. But a post-#252/pre-#264 function reaches the flat branch and calls `grid.emit_template` on a raster/fullsphere config, which can just as easily **raise** (a raster template has no point-path shape) → the handler returns 500 → this line raises the generic `Lambda raster setup error: <body>` and the operator never sees the redeploy guidance for exactly the stale window the echo guard was built for. `test_stale_setup_echo_fails_fast` only exercises the 200-with-wrong-echo shape. Consider recognizing the stale 500 here too (or routing both to the same redeploy hint).

src/zagg/runner.py
```diff
@@ -3191,6 +3230,68 @@ def _invoke_lambda_setup_async(
...
+    try:
+        body = json.loads(result.get("body") or "{}")
+    except (TypeError, ValueError):
+        body = {}
+    if body.get("pipeline") != "raster":
```
espg, line 3310: 🤖 *from Claude (review)*

[non-blocking] Confirming an accepted tradeoff, not a defect: this echo guard is **post-write**. A pre-#264 (but post-#252) function's flat branch has already PUT a wrong point-path template into the store before it returns 200 and trips this check, so the store is left holding a bad template that a rerun-with-`overwrite` must clobber — the ping can't pre-empt it because a raster config isn't hive (validate_manifest is a no-op). The docstring + error text already state the overwrite remedy and it mirrors the PR #205 precedent; flagging so it's an explicit sign-off that the leftover-artifact window is acceptable.

src/zagg/runner.py
```diff
@@ -3191,6 +3230,68 @@ def _invoke_lambda_setup_async(
...
+    (the PR #205 layout-echo precedent); anything else raises a redeploy
+    error before any worker is dispatched. The ping preceding this call
+    already gated out pre-#252 functions with zero writes.
+    """
+    event = {
+        "mode": "setup",
+        "store_path": store_path,
+        "overwrite": overwrite,
+        "config": config_dict,
+        "times_us": [int(t) for t in times_us],
```
espg, line 3272: 🤖 *from Claude (review)*

[nit] `times_us` rides inline in the RequestResponse setup payload. This is a **synchronous** invoke, so the cap is 6 MB (not the 256 KB async/Event cap), and int64 μs stamps serialize at ~17 bytes each → ~350K timesteps before the ceiling — comfortably above any realistic raster series. Just noting there is no chunking/fallback, so a pathologically dense catalog would surface as an opaque boto `ClientError` on this invoke rather than a clear message. No action needed now.

espg:
🤖 *from Claude (review)*

Phase 2 adversarial review (commit `413c505`, runner + tests) — **4 findings, none blocking**:

1. [non-blocking] `_invoke_lambda_ping` reuse carries hive-worded stale/redeploy messaging + "Hive preflight OK" log onto the raster path (runner.py:925 / 3377 / 3383).
2. [non-blocking] The `pipeline: raster` echo guard's redeploy remedy only fires on a 200 with wrong echo; a stale function that 500s at the flat `emit_template` bypasses it into a generic error (runner.py:3281).
3. [non-blocking] Confirming the echo guard is post-write — a stale function leaves a bad template needing overwrite; documented + matches PR #205, flagging for explicit sign-off (runner.py:3286).
4. [nit] `times_us` inline payload ceiling — sync cap is 6 MB (~350K timesteps), no chunking fallback (runner.py:3267).

Verified clean: local-path byte-identity (template write correctly moved below the lambda branch), `template_s` bracketing + summary threading, numpy int64 JSON-safety (`[int(t) ...]` + wire pin), the `open_store` orchestrator-write guard test, and the echo guard's airtightness against echo-less/hive/layout bodies.

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed in 3ce2b41. Generalized the shared ping to be pipeline-neutral: the stale-deploy error now reads "predates the issue #252 dispatch lifecycle — redeploy the function before dispatching this run" (drops "hive dispatch"/"hive runs"), and the success log is now "Preflight OK" (drops "Hive"). The pinned `tests/test_hive.py::TestInvokeLambdaPingEvent` still passes ("redeploy" and "clear the store root" strings preserved). Also strengthened `test_stale_ping_fails_fast_no_setup` to assert `"hive" not in str(excinfo.value).lower()`, so the drift is now caught on the raster path.

src/zagg/runner.py

Fixed in 6aa3d82. Extended the non-200 branch: a 500 cannot be told apart from a genuine failure, so the message now carries the redeploy hint alongside the body — "... — if this is a pre-#264 deployment, its flat point-path branch may have raised on the raster config; redeploy the function, then rerun with overwrite to replace anything it wrote". Extended `test_non_200_raises` to assert `"redeploy" in str(excinfo.value)`.

src/zagg/runner.py
espg: 🤖 *from Claude*

Leaving by design — this is the acknowledged sign-off, not a defect. The echo guard is intentionally post-write: a raster config is not hive, so `validate_manifest` is a no-op and the ping cannot pre-empt the flat branch, exactly as the PR body stale-deployment posture section states and matching the PR #205 precedent. The error text already directs a rerun-with-overwrite to clobber the leftover template. The residual leftover-artifact window closes permanently once the fleet is redeployed. Kept standing for espg sign-off; no code change.

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed in 7f56513. Added a comment at the `times_us` line noting the sync RequestResponse 6 MB cap (not the 256 KB async/Event cap), the ~17 B/int64-stamp → ~350K-timestep headroom (orders above the pinned 85-item NEON benchmark catalog), and that there is no chunking fallback (a pathological series surfaces as a boto ClientError). No behavioral change.

src/zagg/runner.py
```diff
@@ -3191,6 +3230,83 @@ def _invoke_lambda_setup_async(
...
+    if response.get("FunctionError"):
+        raise RuntimeError(f"Lambda raster setup failed: {payload}")
+    result = json.loads(payload)
+    if result.get("statusCode") != 200:
+        # A pre-#264 function reaches the flat branch and calls
+        # grid.emit_template on the raster config, which can raise (a raster
+        # template has no point-path shape) → 500. A 500 is indistinguishable
+        # from a genuine failure here, so surface the redeploy hint alongside
+        # the body rather than swallowing the stale-deployment case.
+        raise RuntimeError(
```
espg, line 3300: 🤖 *from Claude (review)*

[non-blocking] After the two fold rounds, the redeploy hint added in 6aa3d82 is now appended to *every* non-200 from the setup invoke, but the only non-200 a correctly-deployed (post-#264) function returns on the normal runner path is the legitimate `ContainsGroupError` clobber-refusal (500) added in 813edb8. `RasterStrategy.run` fails fast on dense (line ~701) and empty catalog (line ~718) *before* dispatch, so the handler's dense-400 / empty-500 are unreachable via `agg`; the reachable current-function 500 is "store already holds a different-valued time coord, pass overwrite=True".

For that case the operator (e.g. rerunning the raster benchmark over a changed catalog without `overwrite=True`) sees a "if this is a pre-#264 deployment... redeploy the function" hint that does not apply. It's softened by the `if this is a pre-#264 deployment` hedge, the real body (`A group exists in store ... at path ...`) is preserved, and "rerun with overwrite" happens to be the correct fix — so this is messaging noise, not a functional bug. If you want to disambiguate, gate the redeploy hint on the body *not* looking like a ContainsGroupError (or key it off the `pipeline` echo being absent), leaving a clean clobber message for the legitimate overwrite case.

espg:
🤖 *from Claude (review)*

Adversarial review of the remaining unreviewed commits (phase 4 doc, fold rounds 1 and 2, and the phase-3 orchestrator-write audit). **1 finding: 1 non-blocking, 0 blocking.**

- [non-blocking] `src/zagg/runner.py:3291` — the 6aa3d82 redeploy hint is appended to every non-200 setup response, conflating the legitimate 813edb8 `ContainsGroupError` clobber-refusal (500) with a stale deployment. Messaging noise only (hedged, real body preserved, "rerun with overwrite" is the correct fix). Inline thread has the detail.

**Phase-3 audit verdict: verified accurate.** `RasterStrategy.run`'s lambda branch returns `_run_lambda_shards` before any `open_store` (runner.py:727 vs the local `open_store` at 744); `_run_lambda_shards` does ping → sync setup → fan-out → return summary, with no orchestrator-side `open_store`, `emit_raster_template`, `.status/`, finalize, `consolidate_metadata`, or `coverage.moc` write on the path. The `_event` payload carries no `result_url`, so the status channel never engages. Template write is worker-side via the sync setup invoke.

Also verified: 813edb8 raises `zarr.errors.ContainsGroupError` (a `ValueError` subclass) — matched to the local-path `test_overwrite_and_rerun` contract and mapped to 500 by the handler's catch-all; `open_array` on a missing time array (empty store *and* group-without-array) raises `ArrayNotFoundError` (caught), so the fresh-store and local hot paths are safe. Doc claims in benchmark.md are accurate: `template_s` is the raster analogue of the point/hive path's `setup_s` (runner.py:2569), and the benchmark-cicd.md "template write happens inside the Lambda / No S3" cross-reference (lines 98-99) now genuinely holds for both legs. `uv run pytest -q tests/test_raster_runner.py tests/test_lambda_raster_handler.py tests/test_raster_pipeline.py` → 87 passed.

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed in b434068. Gated the stale-deployment hedge off the overwrite-refusal case: the non-200 branch now always surfaces the handler body, but appends the "if this is a pre-#264 deployment... redeploy" hint only when the body does NOT look like the ContainsGroupError clobber-refusal. The gate is a conservative, best-effort substring match on zarr's `A group exists in store` text (commented as messaging, not control flow), so the legitimate "rerun with overwrite" 500 gets a clean message while a raising pre-#264 flat branch (indistinguishable from a genuine 500) still carries the redeploy remedy.

Extended `TestInvokeLambdaRasterSetupEvent.test_non_200_raises` to pin both shapes: a generic 500 (`{"error": "boom"}`) still asserts `redeploy` in the message, and a ContainsGroupError-shaped 500 asserts `redeploy` is absent while the `A group exists in store` store message is preserved.

PR #268, opened 2026-07-18
Unblock 0.32 standup: signed dist reads + ForEach reference idioms (issues #267, #269)

### LINKED ISSUE #267: stand_up.sh hardcodes --no-sign-request; fails now that the dist bucket requires auth
🤖 *from Claude*

`sliderule-public-cors` no longer allows anonymous reads, so every `--no-sign-request` S3 call in `deployment/aws/stand_up.sh` (versions.json resolve at :96/:126, the staged-minor HEAD checks at :123, and the artifact downloads at :176-177) now 403s even when the caller has valid credentials — the flag actively discards them. Verified: anonymous HEAD on `0.32/lambda_layer_arm64.zip` → 403 Forbidden; authenticated HEAD (nasa profile) → 200, 50.2 MB, staged 2026-07-17T23:48. The artifacts are staged; only the auth mode is wrong. Bit espg's 0.32 standup for the #258 worker-size variants tonight.

Fix: default to **signed** requests (use ambient credentials), with `STAND_UP_ANON=1` restoring the old anonymous mode for genuinely public mirrors. Out of scope but related: `publish_mirror.sh` and any workflow that anon-reads the same bucket should be audited the same way (follow-up).

Refs: #258 / #235 (the standup this blocked), #174 (the staged-minor check).

### LINKED ISSUE #269: Fn::ForEach worker-variant references use bare-string placeholders — changeset creation fails
🤖 *from Claude*

The first live standup of the #258 worker-size variants failed at changeset creation:

```
Template format error: Unresolved resource dependencies [WorkerFn${WorkerMemory}] in the Resources block
```

Root cause: the `AWS::LanguageExtensions` `Fn::ForEach` transform substitutes `${Identifier}` in logical-ID **keys** and inside `Fn::Sub` strings, but **not** in bare strings passed to `Ref`/`GetAtt`/`DependsOn`. The [official examples](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-foreach-example-resource.html) cross-reference looped resources exclusively via a nested `Fn::Sub` inside `Ref`/`GetAtt` (e.g. `SubnetId: !Ref {Fn::Sub: 'PublicSubnet${Identifier}'}`) and never use `DependsOn` in a loop (it forbids intrinsics — implicit `Ref` dependencies instead). PR #258's template used the bare-string forms at 8 sites (EventInvokeConfig `FunctionName`, MetricFilter `DependsOn` ×4, Output `GetAtt` ×2), and the plain-string `MetricName`s would additionally collide across variants un-substituted. The pure-Python render tests substituted textually per-loop, so they over-approximated the transform and could not catch this — exactly the risk #258 flagged with "the first real standup is a human action".

Fix (in PR #268, bundled with the standup auth fix since both block the same standup):
- `Ref`/`GetAtt` to looped logical IDs → nested `'Fn::Sub'` form.
- `DependsOn` → removed; the MetricFilter `LogGroupName` now derives from `Ref(function)` (== the function name), giving the same ordering as an implicit dependency.
- `MetricName`s → `!Sub`.
- Output `Description`s de-parameterized (no intrinsics allowed there).

Refs: #235 / PR #258, PR #268, the [failing standup run](https://github.com/englacial/zagg/issues/267).

espg:
Closes #267
Closes #269

The distribution bucket (`sliderule-public-cors`) no longer allows anonymous reads, so `stand_up.sh`'s hardcoded `--no-sign-request` now 403s every dist-bucket call even when the caller's credentials could read the objects — the flag actively discards them. Verified: anonymous HEAD on `0.32/lambda_layer_arm64.zip` → 403; authenticated (nasa profile) → 200, 50.2 MB. This blocked espg's 0.32 standup for the #258 worker-size variants.

## What

All five dist-bucket reads (versions.json resolve ×2, staged-minor HEAD checks, artifact downloads) switch from hardcoded `--no-sign-request` to `$DIST_SIGN`, which defaults **signed** (ambient credentials) and restores anonymous mode via `STAND_UP_ANON=1` for genuinely public self-hosted mirrors. String (not array) expansion deliberately: macOS `/bin/bash` 3.2 + `set -u` rejects the empty-array idiom, and the flag is a single spaceless token.

## Testing

`bash -n` clean. Live validation is the standup itself (operator-run): espg reruns `env AWS_PROFILE=nasa-admin OUTPUT_BUCKET=sliderule-public ./deployment/aws/stand_up.sh` from this branch — the previously-failing HEAD is confirmed to pass signed with these credentials.

## Questions for review

- The script header still describes the bucket as public — left as-is pending the wider audit (issue #267 notes `publish_mirror.sh` and any workflows that anon-read the same bucket need the same treatment; out of scope here).

## Second fix bundled (issue #269) — Fn::ForEach reference idioms

The auth fix exposed the next blocker: changeset creation failed with `Unresolved resource dependencies [WorkerFn${WorkerMemory}]`. The LanguageExtensions transform does not substitute `${Identifier}` in bare strings passed to Ref/GetAtt/DependsOn (only in logical-ID keys and Fn::Sub strings — per the official examples). Fixed per the documented idioms: nested `'Fn::Sub'` inside Ref/GetAtt for looped logical ids; DependsOn removed in favor of an implicit dependency (MetricFilter LogGroupName now derives from Ref(function) == function name); MetricNames wrapped in !Sub (they must stay unique per variant); Output Descriptions de-parameterized (no intrinsics allowed). ForEach stays — no static enumeration needed.

Verified: YAML parses; all 31 TestWorkerSizeVariants render tests pass unchanged (their textual substitution is agnostic to the nested forms); the one test_lambda_build failure is the known pre-existing environmental py3.10 item. Live validation is espg's rerun of the same standup command from this branch.

### DIFF
--- a/deployment/aws/stand_up.sh
+++ b/deployment/aws/stand_up.sh
@@ -78,6 +78,15 @@ DIST_BUCKET="${DIST_BUCKET:-sliderule-public-cors}"
 DIST_PREFIX="${DIST_PREFIX:-}"                       # keys: [<prefix>/]<minor>/<zip>
 DIST_REGION="${DIST_REGION:-us-west-2}"
 
+# Distribution reads are signed with the caller's credentials by default — the
+# dist bucket no longer allows anonymous reads (issue #267), so an unsigned
+# request 403s even for callers whose credentials could read it. Set
+# STAND_UP_ANON=1 to restore unsigned reads against a genuinely public mirror
+# (e.g. a self-hosted DIST_BUCKET copy). Expanded unquoted (single flag, no
+# spaces) so the empty default vanishes — macOS bash 3.2 + `set -u` rejects
+# the empty-array idiom.
+DIST_SIGN="${STAND_UP_ANON:+--no-sign-request}"
+
 dist_key()   { local p="$DIST_PREFIX"; [ -n "$p" ] && p="$p/"; echo "${p}${1}/${2}"; }  # minor, zip
 dist_root()  { local p="$DIST_PREFIX"; [ -n "$p" ] && p="$p/"; echo "${p}${1}"; }       # versions.json path
 
@@ -93,7 +102,7 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 # CORS bucket's versions.json) -> explicit LAMBDA_VERSION -> the repo's git tag
 # (works from a fresh clone) -> the installed zagg.
 if [ "${LAMBDA_VERSION:-}" = "latest" ]; then
-    MINOR="$(aws s3 cp "s3://$DIST_BUCKET/$(dist_root versions.json)" - --region "$DIST_REGION" --no-sign-request 2>/dev/null \
+    MINOR="$(aws s3 cp "s3://$DIST_BUCKET/$(dist_root versions.json)" - --region "$DIST_REGION" $DIST_SIGN 2>/dev/null \
         | python3 -c 'import json,sys; print(json.load(sys.stdin)["latest"])' 2>/dev/null || true)"
     [ -z "$MINOR" ] && { echo "ERROR: could not read 'latest' from s3://$DIST_BUCKET/$(dist_root versions.json)"; exit 1; }
     echo "Resolved LAMBDA_VERSION=latest -> $MINOR"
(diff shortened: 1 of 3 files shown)


deployment/aws/stand_up.sh
```diff
@@ -78,6 +78,15 @@ DIST_BUCKET="${DIST_BUCKET:-sliderule-public-cors}"
...
 DIST_REGION="${DIST_REGION:-us-west-2}"
 
+# Distribution reads are signed with the caller's credentials by default — the
+# dist bucket no longer allows anonymous reads (issue #267), so an unsigned
+# request 403s even for callers whose credentials could read it. Set
+# STAND_UP_ANON=1 to restore unsigned reads against a genuinely public mirror
+# (e.g. a self-hosted DIST_BUCKET copy). Expanded unquoted (single flag, no
+# spaces) so the empty default vanishes — macOS bash 3.2 + `set -u` rejects
+# the empty-array idiom.
+DIST_SIGN="${STAND_UP_ANON:+--no-sign-request}"
```
espg, line 88: 🤖 *from Claude (review)*

`${STAND_UP_ANON:+--no-sign-request}` enables anonymous reads for **any non-empty value**, not just `1` — the `:+` expansion only tests null-vs-non-null, ignoring the value.

Scenario: the config/usage docs say "Set STAND_UP_ANON=1". A user who instead sets `STAND_UP_ANON=0` (or `false`/`no`) intending to *force* signed reads gets `--no-sign-request`, which now 403s against the private dist bucket — precisely the #267 failure, resurfaced and hard to diagnose (the value looks falsy). The only "off" is unset/empty. If you want to reject falsy strings, gate on an explicit match (e.g. `[ "${STAND_UP_ANON:-}" = 1 ] && DIST_SIGN=--no-sign-request`); otherwise a doc note that any non-empty value enables it would suffice.

Severity: low.

espg:
🤖 *from Claude (review)*

Documentation-drift finding (cannot attach inline — the lines are outside the diff hunk).

The header still describes the dist bucket as public: line 4-5 "Lambda code ... is read from the **public** distribution bucket", and line 74 "The **public** CORS bucket the release pipeline stages to: listable". These now contradict the new `DIST_SIGN` comment block (lines 81-87) which states the bucket "no longer allows anonymous reads (issue #267)".

Scenario: a reader who trusts the header expects anonymous access and is surprised the script signs by default; two comment blocks in the same file disagree. Purely documentation — no runtime effect.

The PR marks the header out of scope; raising it so that call is explicit rather than an oversight. A minimal touch ("public distribution bucket" -> "release distribution bucket" at line 4, and dropping "public" at line 74) removes the contradiction without expanding scope. Standing for @espg to accept or decline.

Severity: low (doc-only).

espg:
🤖 *from Claude*

CI red was mine: the phase-2 template fix updated the reference idioms but not the render-test pins — and my pre-push verification was a false green (`uv run --directory` collected the main tree's unmodified tests against its unmodified template; espg caught the CI failure). Fixed in b0e59ad's follow-up: the pins now assert the documented forms — `{"Ref": {"Fn::Sub": logical}}` for looped async configs, `DependsOn` absent with the implicit-ref `LogGroupName` (Sub list form), `!Sub`-wrapped MetricNames (uniqueness check unwraps them), and the two-element `GetAtt` list in Outputs. Verified from inside the branch worktree this time: 30 passed, only the pre-existing environmental py3.10 build test fails (identical on main). The template itself was already live-validated by espg's successful CloudFormation deploy.

deployment/aws/stand_up.sh
espg: 🤖 *from Claude*

Folded in da5221e: `DIST_SIGN` now sets `--no-sign-request` only for truthy `STAND_UP_ANON` values (1/true/yes) via a case statement — `STAND_UP_ANON=0`/`false` stays signed as a user would intend, closing the resurfaced-403 path you described.

espg: 🤖 *from Claude*

Correction to the reply above: this PR merged ~30 s before my fold push, so the cited sha landed on a briefly-recreated (now deleted) branch. The fold rides a one-line follow-up instead: PR #271 (`ef4bbf8`), cherry-picked onto post-merge main.

PR #270, opened 2026-07-18
Spill-partition aggregation: streaming.mode spill (issue #217)

### LINKED ISSUE #259: sharded: true silently no-ops without chunk_inner — single-cell reads cost a whole-shard object fetch
🤖 *from Claude*

## Summary

With `chunk_inner` unset — which is every shipped config today — `HealpixGrid` **silently force-disables sharding** (`healpix.py:129-131`: `if self.chunks_per_shard <= 1: sharded = False`). Since #215/#233 made `sharded: true` the default, this means the production default *believes* it is sharded but the ShardingCodec never engages anywhere. Espg flagged this as a trap to fix ([#217 discussion](https://github.com/englacial/zagg/issues/217)): "we don't want sharding disabled if chunk_inner is omitted."

## Verified behavior at K=1 (chunk_inner unset)

- **Dense arrays**: `_shard()` returns the spec unwrapped (`healpix.py:598-610`). Object count is unchanged (1 object/shard either way, since the regular chunk already spans the shard) — the "byte-identical to a pre-#215 unsharded write" claim in the comment is true.
- **Ragged arrays**: `rag_shard = None` (`healpix.py:648-651`) → a regular vlen array whose chunk is the whole shard. Also 1 object/shard.
- So nothing is *lost* in object counts today. What never materializes is **sub-object read granularity** — and that's the bite:

**Measured on a real store** (`s3://sliderule-public/zagg-bench/issue217-ccheck.zarr`, the #217 (C)-check run: CONUS o8 shard `321241241`, 120 granules, 82.6M obs, 1.04M occupied cells):

| array | objects for this shard | size |
| --- | ---: | ---: |
| `19/h_tdigest/c/149276` | 1 | **223 MB** |
| `19/morton`, `19/cell_ids` | 1 each | 33.6 MB |
| `19/count` | 1 | 16.6 MB |

Reading **one cell's digest** costs a 223 MB whole-object fetch. There is no sharding index to consult because there is no ShardingCodec. A denser shard (the 148/155-granule OOM shards of #217) will be proportionally worse.

Note this is *not* fixable by `sharded: true` alone: even wrapped in a ShardingCodec, K=1 means inner chunk == shard, so the index would resolve every cell to the same whole-shard span. Sub-object granularity exists **only** via `chunk_inner` — it is the one knob that makes the ShardingCodec do anything (relevant to the keep/deprecate question raised in [#217](https://github.com/englacial/zagg/issues/217#issuecomment-5005973590); #238 discussed and dropped the separate `shard_order` sub-shard object-granularity knob, whose "2-GET single-cell reads" rationale implicitly assumed an engaged sharding index).

## Options

1. **Default `chunk_inner` when sharded** (recommended direction): when `sharded: true` (now the default) and `chunk_inner` is unset, derive one — e.g. `chunk_order = parent_order + 2` (K=16) or a target inner-chunk cell count (`cells_per_chunk ≈ 64²–256²`). Gives the production default real sub-object reads. Needs: a benchmark point for the read-side win vs shard-index overhead, and a decision on the value. Interacts with #217 — the same `chunk_inner` becomes the digest-state scoping unit if chunk-scoped streaming lands, and with the (A)/(C) write-independence constraint (per-chunk independent writes need the unsharded path, `runner.py` RMW warning) — so the escape-valve config (`sharded: false`) must keep the ability to *unset* it.
2. **Warn loudly** when `sharded` is requested (explicitly or by default) but no-ops at K=1. Zero behavior change, but the default config then always warns — noisy unless (1) also lands.
3. **Document only.** Cheapest, leaves the 223 MB single-object reads in place.

Grid-layout constraint check: `chunk_inner` finer than `parent_order` requires grid `layout: fullsphere` (`healpix.py:98-105`) — which **is** the default, and is orthogonal to `output.store_layout: hive` (#236/#237), so option (1) is compatible with the unified hive+sharded target.

Refs: #215 / #233 (sharded default), #209 (vlen ragged layout), #236 / #237 (hive parity), #217 (chunk-scoped streaming; `chunk_inner` keep/deprecate discussion), #238 (dropped `shard_order` knob).

espg:
Refs #217 — implements the ratified spill-partition plan (option A): [plan + amendments](https://github.com/englacial/zagg/issues/217#issuecomment-5009132382), [design integration (canonical phase list)](https://github.com/englacial/zagg/issues/217#issuecomment-5009293773), motivated by the [fleet A/B](https://github.com/englacial/zagg/issues/217#issuecomment-5009144609) showing the heavy o8 trio is **time-bound on streaming merge CPU** (~500–600 s of per-flush `build_tdigest`+`merge_tdigests`), not memory-bound. Spill deletes that term: flushes append grouped observations to unlinked per-partition files in `/tmp`, and each cell is aggregated **once, from complete data**, after the reads — byte-identical to the pooled path in the single-block regime, at every reducer the pooled path supports.

**Closes #259** — phase 7 folds in the omitted-`chunk_inner` sharding no-op fix (auto-derive the 64² inner chunk), so merging this PR resolves it. Issue #217 stays open for espg's fleet validation of the spill path.

*(This PR is authored by a Claude routine run on the `claude/217-spill` branch; espg reviews and merges.)*

## Phases

- [x] **1. Spill writer/reader** (`728a329`, fold `504669b`) — `src/zagg/processing/spill.py`: per-block, per-partition packed append files (cells + schema columns, raw bytes). Files are `mkstemp`'d and **unlinked at birth** — the open fd is the only reference, so no `/tmp` leak across warm invokes and no teardown pathway to miss. Partition key = inner-chunk id (`clip2order`; K=1 → single partition); production geometry (`chunk_inner: 13` in every shipped HEALPix config = 1,024 inner chunks per o8 shard) later forced a **contiguous partition-group cap** — `56d15c6` clips at `min(chunk_order, parent_order + 3)`, so at most 4³ = 64 files per block, far under Lambda's ~1,024 nofile default, while each group still reads back as one contiguous cell span for the pooled per-chunk build (a consumed-group tripwire guards the iter_chunks-contiguity invariant, fold `a90db58`). Exact byte accounting on write. Standalone `check_tmp_headroom` statvfs guard naming the `-disk` variant fix (deliberately not the #260 arena guard, which phase 6 deletes).
- [x] **2. Worker integration, full reducer surface** (`f6d4ccd`) — `aggregation.streaming.mode: merge (default) | spill` in the existing block (`get_streaming`); merge mode byte-identical to before. Spill flush = `_concat_and_group` → append grouped columns to the current block's partitions. **The central reuse decision:** a partition read back + `_group_columns` is exactly a pooled carrier, so `chunk_outputs` drives the pooled machinery (`_pool_chunk_columns` → `_eval_chunk_precompute` → `_aggregate_chunk_cells`) — every pooled reducer (scalar functions, expressions, vectors, tdigest, **located ragged**, `chunk_precompute`) works with zero reimplementation and single-block output is byte-identical to pooled **by construction** (partition rows are the chunk's rows in global read order; the stable sort reproduces the pooled per-cell slices). Threshold-triggered block close (formula below); multi-block folds mergeable reducers per block (counts by summation, tdigests via `merge_tdigests` — one merge round per ~memory-sized block instead of per 25-granule buffer); a non-mergeable config that actually overflows raises `SpillOverflowError` naming the remedies (bigger tier / `-disk` variant / finer `parent_order`), re-raised loudly through the worker's tolerated per-granule except. `spill_write_s`/`spill_read_s`/`spill_bytes` land in `phase_timings` (the approved `/tmp` throughput instrumentation, retiring the #235 leftover). `occupied_out` fed from per-flush uniques.
- [x] **3. Local A/B replay** — run by the **main session** (working nasa auth) with this branch's `replay_spill_ab.py` driver against the real count slabs (`issue217-ccheck.zarr` shard 3779012668573089800 = 148 g, `issue217-ccheck-counts.zarr` shard 3780367266898509832 = 120 g). Table and verdicts below; **kill criterion passed decisively**, plus one finding fixed in `b7281d3` (+ folds `23f0b87`, `0ffd4a9`).
- [x] **4. Fleet validation prep** — see "Fleet validation (espg or main session runs)" below.
- [x] **5. Async read/reduce overlap** (`a0340a1`, fold `88c19e4`) — block N reduces (disk→memory) on one worker thread while reads stream into block N+1 (network→disk). At most one closed block in flight (the next close joins the previous reduce), so `/tmp` holds closing + filling — exactly the threshold formula's reservation — and blocks fold in close order, keeping the merge sequence (and bytes) identical to sequential; reducer failures park and re-raise at the next join/finalize as `SpillReduceError`, which the worker's tolerated per-granule except re-raises (review fold `88c19e4`) so a failed block can never be silently dropped from the output. `overlap=False` keeps the sequential path for A/B.
- [x] **6. #260 surface removal** — `state_layout`/`arena_backing` knobs, the arena code paths (`_alloc_arena`/`_fold_arena`/`_chunk_outputs_arena`/CSR state/poisoning), and their tests deleted (espg: dormant-then-collapse; the rebase merge foreclosed a clean revert). `get_streaming` now rejects **unknown keys** loudly (the #239 discipline), so a stale arena config fails with a pointed message instead of silently running the dict path. Net knob count is flat: two knobs removed, `mode` added.
- [x] **7. Auto-derive `chunk_inner` when omitted — closes #259** (`77f7cdb`) — `grids/__init__.py` `from_config`: with sharding on (the default since #215/#233) and `chunk_inner` omitted, a fullsphere HEALPix grid now derives the 64×64 (`4^6 = 4096`-cell) inner chunk — `chunk_order = child_order − 6`, floored at `parent_order` — instead of silently collapsing to `chunk_order == parent_order` (K==1) and disabling sharding, which cost a whole-shard object fetch per single-cell read (the #259 bug). This reproduces the value the flagship configs (`atl03_tdigest_healpix` / `atl03_gain_bias_healpix` / `atl03_tdigest_located_healpix`) already hand-set as `chunk_inner: 13`, making the win automatic for any config that omits the knob. Derivation is skipped when `child_order − 6 ≤ parent_order` (shard already spans ≤ one inner chunk → the value would just reproduce the K==1 no-op; `atl06`/`atl06_nullable` at parent 6 / child 12 stay unchanged), when `sharded: false` (no bundling → deriving would only add K regular objects), and for dense layout (deprecated — its constructor rejects `chunk_inner` finer than `parent_order`). Settles the "K=4 vs K=16?" question from the [read benchmark](https://github.com/englacial/zagg/issues/259#issuecomment-5009973508): a single-cell `count` read at the 64² geometry is 2 requests / 32 KB / ~130 ms cold, vs 1.05–4.19 MB at the coarser shardings and a 16.8 MB whole-object fetch unsharded — the hand-set 64² wins, so the fix is making it automatic, not sweeping K. Independent of the spill machinery; the flat-config production stores already carry `chunk_inner: 13`, so no shipped config's behavior changes. Tests: 5 in `tests/test_grids.py::TestFromConfig` (derive-when-omitted, floor-at-parent, explicit-not-overridden, not-derived-when-unsharded, not-derived-for-dense); full non-slow suite green (2292 passed).

## Phase-3 A/B replay results (main session runs, real count slabs)

| slab | arm | in-read-loop agg | post-read build/emit | peak RSS | spill bytes (write / read wall) |
|---|---|---:|---:|---:|---|
| 120 g — 82.6M obs, 1.04M cells, 5 flushes | dict-merge | 503.4 s | 2.1 s | 2,528 MB | — |
| | spill (K=1) | **6.1 s** | 148.7 s | 5,787 MB | 1,652 MB (0.5 s / 0.8 s, local SSD) |
| 148 g — 152.7M obs, 1.01M cells, 6 flushes | dict-merge | 972.4 s | 1.8 s | 3,567 MB | — |
| | spill (K=1) | **9.2 s** | 232.5 s | 8,338 MB | 3,054 MB (1.0 s / 1.8 s) |

**Verdict (a) — kill criterion passed decisively.** The in-read-loop aggregation term collapses **82× / 106×** (503.4→6.1 s, 972.4→9.2 s), and total aggregation CPU drops **3.3× / 4.0×**. The dict arm's 148 g in-loop term alone (972 s) exceeds the 900 s Lambda ceiling — fully explaining the fleet trio deaths; spill's replacement cost (one-shot builds from complete data) is 149–233 s of post-read build, exactly the plan's "order 100 s, not 500" prediction.

**Verdict (b) — build-peak finding, fixed on this branch.** The K=1 partition build peaked at ~2.7–3.5× the partition's spilled bytes (5,787 MB over 1,652; 8,338 over 3,054 — read-back columns, stable-sort gather copies, and per-cell outputs coexist). The original threshold derivation under-reserved for this: at the 4096`-disk` tier a derived ~2.4 GB block would have built at ~7 GB → OOM. Fixed in `b7281d3`: the derivation now divides by a measured `_BUILD_MULT = 3` (constant documented against these numbers, pinned by test), giving `0.8 × 0.75 × memory × K / 3 = 0.2 × memory × K`, still capped at 45% of `/tmp` free space. The build unit is one partition **group**, so the group count (64 at production geometry, `56d15c6`) divides the build working set — ~3 GB of spill reads back in ~50 MB units — which covers this finding with headroom while keeping fds bounded.

## Block threshold (documented formula)

A closing block's reduce working set is its largest partition **group** (~block/K, K = the partition-group count, at most 64) at the measured ~3× build peak, live beside the read buffer — so block bytes ≲ `0.8 × (memory − read buffer) × K / 3` ([design point 5](https://github.com/englacial/zagg/issues/217#issuecomment-5009293773) + the phase-3 build-peak measurements above). The read buffer + slack are budgeted at 25% of memory, giving `0.2 × memory × K`, capped at 45% of `/tmp` free space (closing + filling block must coexist). Memory comes from `AWS_LAMBDA_FUNCTION_MEMORY_SIZE`, else the cgroup v2 limit, else physical RAM. Injectable for tests/ops via `SpillAggregator(block_bytes=...)`.

## Testing

- `tests/test_spill.py` (46 tests): round-trip byte-exactness (multi-dtype, multi-append, read-then-append, row-alignment + schema-drift guards), partition routing incl. the routing law vs `iter_chunks`, fd hygiene (no visible files even mid-write; every fd closed; worker leaves `/tmp` empty), exact byte accounting, statvfs guard; **byte-identity vs pooled** across the full reducer matrix (scalar `nanmean`, expressions, vector histogram, tdigest, located tdigest, `chunk_precompute` anchor), NaN cells, multi-flush (where merge mode is only approximate), K=4 partitioned chunks via `chunk_results`; forced multi-block via injected tiny threshold — mergeable spill is **byte-identical to merge mode** when blocks align with flushes (same per-cell build+merge sequence), counts exact vs pooled, non-mergeable overflow raises through `process_shard`; overlap-vs-sequential identity + reducer-error propagation (at next close and at finalize); occupied/profile/empty-shard parity. Byte-equality tests pin `shard_workers: 1` (the granule-pool + shared fake-read-iterator nondeterminism trap).
- `tests/test_streaming.py`: merge-mode regression untouched (dict-path tests unchanged); arena tests deleted with the surface; unknown-key rejection added; a dict-streaming occupied_out-vs-pooled test preserves the coverage the arena test carried.
- Full suite `uv run pytest -m "not slow"` green locally on every phase; `ruff check` + `ruff format --check` green (mypy has pre-existing failures on `main` unrelated to this change — untouched, per §4).

## Fleet validation (espg or main session runs)

Requires a fleet build containing this PR (deploy is espg's; agents never run standup). Arm driver: the session scratchpad `run_spill_arm.py` (the `run_ab_arm.py` pattern with a `mode` argument; `pooled` arm drops the streaming block for the byte-identity reference). Submap: `data/conus/conus_regression_submap_o8_oom5.json` on the `conus_benchmark` branch (the 148/155/176 g trio). The arm config leaves `chunk_inner` as the config ships it (13, production geometry): the 64-group partition cap (`56d15c6`) already divides the build working set per the phase-3 build-peak finding, so no geometry override is needed.

```bash
# spill arm — the configuration where every current mode fails
AWS_PROFILE=nasa uv run python run_spill_arm.py \
  data/conus/conus_regression_submap_o8_oom5.json spill trio-spill process-shard-4096-disk 25

# pooled reference (byte-identity check target), same shards
AWS_PROFILE=nasa uv run python run_spill_arm.py \
  data/conus/conus_regression_submap_o8_oom5.json pooled trio-pooled process-shard-4096-disk

# merge baseline for the runtime delta (expected: 900 s wall, as on 0.32.0)
AWS_PROFILE=nasa uv run python run_spill_arm.py \
  data/conus/conus_regression_submap_o8_oom5.json merge trio-merge process-shard-4096-disk 25
```

Acceptance criteria (from the [plan](https://github.com/englacial/zagg/issues/217#issuecomment-5009132382)):
1. 148/155/176 g complete **under the 900 s ceiling** at `process-shard-4096-disk` — where every current mode fails.
2. Output **byte-identical to the pooled arm** (single-block regime): obs counts and digest bytes match per shard.
3. NEON-scale parity: runtime and peak memory within noise of merge-mode streaming on light shards.
4. No `/tmp` leaks across warm invokes; the summary reports `spill_bytes` and derived write/read MB/s per memory tier (the #235 throughput numbers, measured with the design's exact access pattern).

## Questions for review

- Phase-1 commit `728a329` carries a `Co-Authored-By` trailer against the repo's title-only convention (harness default applied before the repo style was re-checked); later commits are title-only. Can't be amended without a force-push, so it stands.
- `get_streaming` now rejects unknown keys in `aggregation.streaming` (it previously ignored them). This is deliberate — a stale `state_layout: arena` config must not silently change behavior — but it is a validation tightening; flag if any shipped config carries extra keys.


src/zagg/processing/spill.py
```diff
@@ -0,0 +1,246 @@
...
+
+    def read_partition(
+        self, part_key: int, *, close: bool = False
+    ) -> tuple[np.ndarray, dict[str, np.ndarray]]:
+        """Read one partition back as ``(cells, {name: values})``.
+
+        Rows come back in exact append order (flush order, within-flush order
+        preserved), so a stable sort by cell reproduces the pooled path's
+        per-cell row order. ``close=True`` closes the partition's file after
+        the read — its (already unlinked) bytes free immediately.
```
espg, line 301: 🤖 *from Claude (review)*

`append` enforces the schema contract loudly (the drift `raise` just above) but leaves the equally load-bearing **row-alignment** contract — `part_ids`, `cells`, and every column being equal-length — completely unchecked, and a violation corrupts silently rather than failing.

The run split derives boundaries from `part_ids` (`bounds`/`ends` use `len(part_ids)`), while the empty-guard on the line above keys off `len(cells)`, and `write_segment` records `n_rows = len(cells[s:e])` yet writes each column slice in full. So if a column (or `part_ids`) is a different length than `cells`, the packed byte stream and the in-memory segment map disagree with no error. Verified:

```
part_ids = [0,0,0]; cells = [10,11,12]; cols = {'v': [1,2,3,4,5]}   # 5 != 3
block.append(part_ids, cells, cols)
block.read_partition(0)  ->  cells=[10,11,12], v=[1,2,3]
```

The two extra column values (`4,5`) are written to the file but fall outside the recorded segment — orphan bytes sitting *before* the append position. With a second column or a follow-on segment those orphans are misread as the next field/segment, i.e. exactly the silent data loss the module's own test docstring warns is unacceptable ("they replace the in-memory pool, so any loss is silent data corruption"). Phase 2 always feeds equal-length arrays by construction, so this only bites a wiring bug — but that's precisely the case where a loud failure matters most, and the guard is O(1):

```python
n = len(cells)
if len(part_ids) != n or any(len(v) != n for v in col_dict.values()):
    raise ValueError("spill append: part_ids/cells/columns must be row-aligned")
```

Suggest placing it right beside the schema-drift check so both halves of the append contract fail the same way.

espg: 🤖 *from Claude*

Added the O(1) row-alignment guard right after the schema-drift check in `append`: it raises `ValueError("spill append: part_ids/cells/columns must be row-aligned")` when `part_ids`, `cells`, or any column length disagree — before any bytes are written, so no orphan bytes can land in the stream. Covered by a new `TestRoundTrip.test_row_misalignment_raises_before_writing` (mismatched column and part_ids lengths both raise; the prior append still round-trips intact). Fixed in 504669b.

src/zagg/processing/spill.py
```diff
@@ -0,0 +1,666 @@
...
+            block.close()
+
+    def _join_reducer(self) -> None:
+        """Wait for the in-flight block reduce; re-raise its failure loudly."""
+        if self._reducer is not None:
+            self._reducer.join()
+            self._reducer = None
+        if self._reduce_err is not None:
+            err, self._reduce_err = self._reduce_err, None
+            raise RuntimeError(
```

The re-raise here is not loud in the real worker — it is swallowed, and the parked error is cleared before it escapes, so a failed overlap reduce silently drops a block's data from the emitted output.

`_join_reducer` re-raises a plain `RuntimeError`, and it fires inside the next `_close_block`, which runs on the `granule_done() -> flush() -> _close_block()` path. In `process_shard` that call sits inside the tolerated per-granule handler:

```python
    if buffered is not None:
        buffered.granule_done()
except SpillOverflowError:
    raise
except Exception as e:
    logger.warning(f"  Error processing file {s3_url}: {e}")
    continue
```

`SpillOverflowError` is deliberately re-raised, but this `RuntimeError` is not — it hits `except Exception` and is downgraded to warn-and-continue. And because the line just above does `err, self._reduce_err = self._reduce_err, None` before raising, `_reduce_err` is already cleared, so the failure is gone for good: the reducer for the current `block` is never started (the `raise` short-circuits before `self._reducer = threading.Thread(...)`), the shard keeps reading, and finalize then emits merged state that silently omits both the failed block and the dropped one. That contradicts this method's own "re-raise its failure loudly" docstring and the phase-5 "results identical to sequential" contract.

Fix: give the overlap failure a distinct type (e.g. `SpillReduceError(RuntimeError)`) and re-raise it in `process_shard` alongside `SpillOverflowError` (`except (SpillOverflowError, SpillReduceError): raise`), so a mid-read reduce failure aborts the shard instead of being logged and dropped. The finalize-path join (outside the read loop) already propagates correctly; only the next-close join inside the loop is swallowed. `TestSpillOverlap` does not catch this because it drives `SpillAggregator` directly, bypassing the worker's tolerated `except`.

Fixed in 88c19e4. Added `SpillReduceError(RuntimeError)` in spill.py, raised it (in place of the plain `RuntimeError`, keeping the `from err` chain and message) from `_join_reducer`, and extended the worker read loop’s tolerated except to `except (SpillOverflowError, SpillReduceError): raise` so a mid-read overlap-reduce failure now aborts the shard instead of being downgraded to warn-and-continue. Added a through-`process_shard` regression test (`test_overlap_reduce_error_raises_through_worker`, monkeypatching `_fold_block` to raise under tiny blocks) asserting `SpillReduceError` propagates, and tightened the two direct-drive overlap error tests to the new type.

src/zagg/processing/spill.py
```diff
@@ -0,0 +1,693 @@
...
+_BUILD_MULT = 3
+
+
+def _default_block_bytes(n_partitions: int, tmp_dir: str | None = None) -> int:
+    """Default spill-block threshold (issue #217 design comment).
+
+    The formula: a closing block's reduce working set is its **largest
+    partition** (~block/K) at the measured :data:`_BUILD_MULT` build peak
+    (columns + sort copies + outputs), live alongside the read buffer — so
+    block bytes ≲ ``0.8 x (memory - read buffer - outputs) x K /
```

The rewritten derivation double-counts **outputs**. `_BUILD_MULT` is now defined as the measured *process-wide* peak over partition bytes, and its comment states that peak already includes the per-cell outputs:

> the read-back columns, the stable-sort gather copies, and **the per-cell outputs** coexist at ~2.6-3.5x the partition bytes

But this docstring still folds outputs into `_BUILD_MULT` *and* reserves them a second time in the 25% carve-out:

> ...at the measured `_BUILD_MULT` build peak (columns + sort copies + **outputs**)... so block bytes ≲ `0.8 x (memory - read buffer - **outputs**) x K / _BUILD_MULT`

The pre-edit version was coherent — the working set was "partition **plus** the block's outputs" (partition alone, outputs called out separately), so subtracting outputs in the reservation matched. Once outputs got absorbed into the measured `_BUILD_MULT` multiple, the separate `- outputs` term became a second charge for the same bytes.

Effect is safe (it shrinks the derived block, erring conservative), so this is not an OOM risk — but the written model is now self-contradictory. That matters because the `_BUILD_MULT` comment explicitly invites future re-tuning ("do not lower without re-measuring the replay"); the next person re-deriving from this docstring can't tell whether outputs are inside the 3x or inside the 25%. Either drop "outputs" from the `_BUILD_MULT` parenthetical (leaving it as the read-buffer/output carve-out) or drop it from the `- outputs` reservation term, so it's charged exactly once.

tests/test_spill.py
```diff
@@ -0,0 +1,790 @@
...
+            f_bavail = 10**7
+            f_frsize = 4096
+
+        monkeypatch.setattr(spill.os, "statvfs", lambda _p: _Big())
+        one = spill._default_block_bytes(1)
+        # 0.8 x 0.75 x mem / _BUILD_MULT: the phase-3 replays measured the
+        # partition build (columns + sort copies + outputs) peaking at
+        # ~2.6-3.5x the partition bytes, so the reservation divides by 3 —
+        # without it a derived K=1 block builds at ~3x its bytes and OOMs.
+        assert one == int(0.8 * 0.75 * 4096 * 2**20 / spill._BUILD_MULT)
```
espg: 🤖 *from Claude (review)*

This assertion doesn't actually pin the measured `3x` the class docstring promises ("must reserve the measured ~3x build peak"). It recomputes the expected value using `spill._BUILD_MULT` itself:

```python
assert one == int(0.8 * 0.75 * 4096 * 2**20 / spill._BUILD_MULT)
```

Since both the code and the test divide by the same `spill._BUILD_MULT` symbol, this is tautological in the constant — if someone changed `_BUILD_MULT = 3` to `2` (exactly the "lower without re-measuring" the constant's comment warns against), the code and this expected value move together and the test stays green. The only thing pinned is the formula *shape*, not the reservation strength.

The neighboring assert (`_default_block_bytes(4) == 4 * one`) does meaningfully pin K-linearity, and `test_tmp_cap_binds_when_small` hardcodes its expected value (`int(0.45 * 1000 * 4096)`). For symmetry — and to make this test the guard the comment relies on — pin the multiplier with a literal: `assert one == int(0.2 * 4096 * 2**20)` (or assert `spill._BUILD_MULT == 3` explicitly). Then an unreviewed change to the constant fails a test.

src/zagg/processing/spill.py

Reworked the `_default_block_bytes` docstring so outputs are charged exactly once: the parenthetical now reads "read-back columns + sort copies + per-cell outputs — outputs are inside this multiple, charged here and nowhere else", and the reservation term dropped the second charge to `0.8 x (memory - read buffer) x K / _BUILD_MULT` (the 25% carve-out is now read buffer + slack only). Formula and `_BUILD_MULT` unchanged; behavior identical. Fixed in 23f0b87.

tests/test_spill.py
espg: 🤖 *from Claude*

De-tautologized the assertion: it now pins the constant directly with `assert spill._BUILD_MULT == 3` (comment cites the phase-3 ~2.73-3.50x peaks and "do not lower without re-measuring"), and the expectation compares against the literal `int(0.8 * 0.75 * 4096 * 2**20 / 3)` rather than `spill._BUILD_MULT`, so lowering the constant to 2 now fails the test. Fixed in 0ffd4a9.

src/zagg/processing/spill.py
```diff
@@ -0,0 +1,725 @@
...
+        self._reducer: threading.Thread | None = None
+        self._reduce_err: BaseException | None = None
+        # Cross-block mergeable running state (only ever fed on block close).
+        self._counts: dict[int, int] = {}
+        self._digests: dict[str, dict[int, np.ndarray]] = {n: {} for n in self._digest_fields}
+        # Per-flush unique cell words; unioned lazily by occupied_cells().
+        self._occupied: list[np.ndarray] = []
+        # Single-block reduce cache: (part_key, col_arrays, cell_to_slice) for
+        # the most recently loaded partition. Chunks sharing a partition group
+        # reuse it, and iter_chunks visits each group as one contiguous run,
```
espg, line 472: 🤖 *from Claude (review)*

The new group-sharing reuse in `_chunk_outputs_exact` makes correctness depend on an **unguarded** invariant, and its failure mode is silent data loss rather than an error.

Pre-commit each chunk had its own partition, consumed exactly once at its single `iter_chunks` visit. Now several chunks share a group file, so the single-slot `_loaded` cache + `read_partition(key, close=True)` (which *deletes* the partition from the block) only stay correct because `iter_chunks` visits a group's chunks **consecutively**. If that ordering were ever violated — a different grid, a future `generate_morton_children`/`iter_chunks` change, a non-nested clip — the sequence `load(G)` → consume → `load(G')` → revisit chunk of `G` would hit `_load_partition(G)`, find `G` no longer in `partition_keys()`, and fall into the **empty-columns else-branch** (spill.py:662-667). That branch is legitimately reached for a genuinely unpopulated group, so it cannot tell "never populated" from "already consumed" — the revisited chunk silently emits zeros instead of raising.

I confirmed the invariant *does* hold today (o8/`chunk_inner:7`: iter order = 64 runs over 64 unique groups, contiguous; dedup sequence matches the sorted-children order the unit test pins), so this is not a live bug. But the invariant is subtle (morton nesting × enumeration order × clip2order monotonicity) and now load-bearing with no tripwire.

Concrete fix: track consumed group keys and distinguish the two empty cases in `_load_partition`, e.g.

```python
if key in self._loaded_keys:   # a set, added on each close=True read
    raise SpillReduceError(
        f"spill group {key} re-requested after close; iter_chunks must "
        f"visit a group's chunks contiguously"
    )
```

so a future ordering regression aborts the shard loudly instead of shipping a shard with silently-empty cells.

espg: 🤖 *from Claude*

Folded in a90db58. Added a `self._consumed: set[int]` tripwire: `_load_partition` records a group key when it read-and-closes a real partition, and now raises `RuntimeError` (naming the iter_chunks-contiguity invariant) if an already-consumed key is re-requested — instead of silently falling into the empty-columns branch. A genuinely-unpopulated group (never read, never consumed) still loads empty as before. Test `test_consumed_group_re_request_trips` drives load(A) → load(B) → re-request(A) → RuntimeError, and asserts an untouched absent key still yields empty outputs.

espg:
🤖 *from Claude*

Added **phase 7 — auto-derive `chunk_inner` when omitted (`77f7cdb`), which closes #259.** `from_config` now derives the 64² (`4^6 = 4096`-cell) inner chunk (`chunk_order = child_order − 6`) when a fullsphere HEALPix config has sharding on (the default) but omits `chunk_inner`, instead of collapsing to K==1 and silently disabling sharding. It reproduces the `chunk_inner: 13` the flagship configs already hand-set, so no shipped config's behavior changes — it just makes the win automatic for any config that forgets the knob. Guards: skipped when `child_order − 6 ≤ parent_order` (already ≤ one inner chunk → K==1 no-op), when `sharded: false`, and for dense layout (deprecated, rejects finer `chunk_inner`).

This lands the fix the [#259 read benchmark](https://github.com/englacial/zagg/issues/259#issuecomment-5009973508) pointed to: a single-cell `count` read at 64² is 2 requests / 32 KB / ~130 ms cold vs 1.05–4.19 MB at coarser shardings and a 16.8 MB whole-object fetch unsharded — the hand-set geometry already wins, so the answer to "K=4 vs K=16?" is neither; automate the 64² default.

Tests: 5 new in `tests/test_grids.py::TestFromConfig` (derive-when-omitted, floor-at-parent, explicit-not-overridden, unsharded opt-out, dense opt-out); full non-slow suite green (2292 passed, 26 skipped). Fresh-context adversarial self-review on the commit found no correctness issues; its one wording nit on the derivation comment is folded in `3a5dd9b`. Independent of the spill machinery — phase 7 touches only `grids/__init__.py` + its test.

Issue #272, opened 2026-07-18
Benchmark the spill path + fix per-release series/table rendering


espg:
🤖 *from Claude*

## Benchmark the spill path + fix the per-release series/table rendering

Two coupled workstreams in the Lambda benchmark harness, now that streaming `mode: spill` (#217, PR #270) is merged and the disk-provisioned worker variants (#235/#258) are live.

---

### A. Add a `spill` column to the tdigest benchmark matrix

Spill only helps where there's a streaming-merge CPU term to delete — i.e. **tdigest** targets (not scalar/gain_bias). Add a spill arm alongside the existing (merge) arm so each tdigest target reports both.

**Mechanics (no new provisioning — #258 made this yaml-selectable):**
- New target field `streaming_mode: merge | spill`. `run_benchmark.py` injects it into `config.aggregation.streaming.mode` at dispatch (mirroring how `sharded` / `index_backend` / `handoff` are already injected).
- Spill targets also carry `worker: {memory: 4096, extra_disk: true}`, injected into the config's top-level `worker:` block. `runner._resolve_function_name` then routes the invoke to **`process-shard-4096-disk`** (4096 MB mem / **6144 MB /tmp**, pre-provisioned by #258). One catch: `run_benchmark.py` currently passes `--function-name process-shard` **explicitly**, which per #258 "wins verbatim" and masks the worker block — so worker-block targets must dispatch with `function_name=None` (let the config's worker suffix resolve), matching the #258 CLI-default change.
- **Scope (confirmed by @espg):** live **every-merge** arm, alongside merge. The `-disk` variants already exist in the benchmark account (742127912612), so the every-merge run stays green.

**Higher billing — ephemeral-storage cost (the key cost change):**
- The merge arm dispatches to `process-shard` (512 MB /tmp = exactly the free tier → **$0 ephemeral**). The spill arm's `-disk` worker has 6144 MB /tmp → **(6144 − 512)/1024 = 5.5 GB billable** ephemeral, charged per GB-second of runtime.
- Add `LAMBDA_EPHEMERAL_PRICE_PER_GB_SEC` (+ `FREE_EPHEMERAL_GB = 0.5`) and fold the ephemeral term into `bench_metrics` cost so `cost_per_shard_usd` / `cost_per_100km2_usd` reflect the true `-disk` bill. Record `tmp_mb` + the ephemeral cost component as new (appended) columns so the merge-vs-spill cost delta is visible in the series/table.
[...]

PR #273, opened 2026-07-18
Benchmark spill arm: ephemeral cost + matrix column + plot fixes (issue #272)

### LINKED ISSUE #237: Design: hive store layout for the raster pipeline (S2 adopts hive)

espg:
Refs #272 · **Closes #237** — benchmark the shipping **spill** path (streaming `mode: spill` on the `-disk` worker), promote the raster leg to hive (#237), and fix the per-release series/table rendering. Design ratified by @espg across #272 + this thread.

*(Authored by a Claude routine on `claude/benchmark-spill-column`; @espg reviews and merges.)*

## Design (settled)

- **Live per-merge leg → spill on `process-shard-4096-disk`, as a read-backend A/B**: two arms, `tdigest_healpix_o9_hive` (inline, **green**) and `tdigest_healpix_o9_hive_sidecar` (sidecar, **purple**), both spill+disk. This re-expands the #193/#202 inline-vs-sidecar matrix (that #250 collapsed) onto the shipping spill path. Plot = one panel/figure, two lines (green=inline, purple=sidecar); plots keep wall + total lambda-seconds/cost.
- We measure **spill because it's what ships** (@espg): on light shards it's within Lambda noise of merge (+~3% wall / <5% cost — accepted for robustness); on heavy shards it's a rescue.
- **Ephemeral `/tmp` billing** (the `-disk` worker's 5.5 GB over the 512 MB free tier) is folded into `cost_per_shard_usd`.
- **Full-AOI release TABLE** gains a **time-to-first-consumer** column (min over shards of the `.status` post time); the run **wall** stays result-ready = max over shards. Plots unchanged.
- **Series floor at 0.33.0** — pre-0.33 rows are wrong (cost predates the ~100 s async-setup term; the last pre-floor point was a 0 s no-run that collapsed the axis). Render-only; parquet untouched.

## Phases

- [x] **Ephemeral cost model** (`1b9a88a`) — `bench_metrics`: `LAMBDA_EPHEMERAL_PRICE_PER_GB_SEC = 3.09e-8`, folded into `cost_per_shard_usd`; `streaming_mode`/`tmp_mb`/`ephemeral_cost_usd` columns. Tests.
- [x] **Dispatch plumbing** (`c49a5d4`) — `run_benchmark` injects `streaming_mode` + `worker`, resolves the `-disk` variant off the base name (`process-shard` → `process-shard-4096-disk`; per-merge `process-shard-test` → `process-shard-test-4096-disk`), threads `tmp_mb`. Tests.
- [x] **Version floor 0.33.0** (`1e4703f`) — `plot_summary` filters every history; sub-0.33 fixtures bumped; floor test.
- [x] **Two-arm inline/sidecar target** (`287eb3b`) — sidecar companion target; the `run all` matrix is now the two spill+disk arms; matrix test updated.
- [x] **Two-line green/purple panel** (`27ac8dc`) — `merge_history` returns both arms; `make_merge_ab_figure` overlays inline (green) + sidecar (purple) on the cost + wall panels; the per-phase diagnostics gain an `arm_col` overlay. Rendering verified with matplotlib (2 new tests).
- [x] **Per-release tables + first-consumer column** (`028ac56`) — a dedicated `_render_release_table` (release schema, distinct from the per-merge one) renders the missing full-AOI + raster tables, embedded on the page; the full-AOI harness computes **time-to-first-consumer** = `min` over shards of the `.status` result-ready post time (#274/#275's `LastModified`), added as `first_consumer_s` to the series schema and shown in the table (plots stay wall + cost). Merged `main` for #275's post-time plumbing.
- [x] **Raster promotion → hive** (`8832993`, closes #237) — `s2_neon_o9` → `store_layout: hive`, the `pending_targets` hive variant moved into `targets`, gate test deleted. Merged `main` for the #259 raster-unsharded fix (`7aed66f`) so the raster grid resolves hive + **unsharded** (raster never shards, #247). @espg ratified the #237 promotion on #272.
- [x] **Setup/finalize panels removed** (`e2a8b86`) — dropped from the diagnostics figure (#274 minimizes both; `setup_s`/`finalize_s` stay as columns so a future creep shows as `total_wall` diverging from read+agg+write).

## Fleet evidence (real Lambda, spill vs merge, hive, inline)

| shard | merge wall | spill wall | Δ | outcome |
|---|---:|---:|---:|---|
| NEON o9 (light, 3.3 M) | 118.7 s | 121.9 s | +2.7% | both complete (noise) |
| CONUS o8 120 g (82.6 M) | 703.8 s | 570.9 s | **−18.9% (& −18.7% cost)** | spill wins |
| CONUS o8 148 g (152.7 M) | **timeout, $0 output (~$0.048 billed)** | 779.9 s | rescue | **spill completes, merge fails** |

Sidecar beats inline everywhere (NEON + both heavy, ~8–15% faster/cheaper, equal memory). Peak RSS ≤ 3.07 GB on the heaviest — no OOM (the local-replay 8.3 GB figure was bogus). **NEON AOI o9 (4 shards):** first-consumer 87.0 s (inline) / 74.5 s (sidecar); all-consumable wall 138.0 / 119.5 s.

## Testing

`uv run pytest tests/test_benchmark.py tests/test_full_aoi_series.py tests/test_raster_benchmark.py tests/test_benchmark_shardmap.py` green; `ruff` clean. Remaining plot phases add their own tests.

## Questions for review
1. The live o9 pin is a light shard where spill is marginally worse than merge — kept spill per "measure what ships"; flagging that the per-merge series will show that small tax (the win is on the heavy/​release shards).
2. Raster promotion (#237) was ratified by @espg on #272 and is done here — flagging that it retires the flat raster arm from the release series (the flat rows on the benchmarks branch predate it; the corrected hive series starts fresh at the next release).

### DIFF
--- a/.github/scripts/plot_series.py
+++ b/.github/scripts/plot_series.py
@@ -942,17 +942,17 @@ def main(argv: list[str] | None = None) -> int:
     # PNGs persist on the benchmarks branch under the Archived section.
     merge_hist = plot_summary.merge_history(df)
     if not merge_hist.empty:
-        if plot_summary.make_summary_figure(
-            [("per-merge point (hive, densest shard)", merge_hist, "commit")],
-            outdir / "merge_summary.png",
-        ):
+        # Issue #272: the per-merge leg is the inline-vs-sidecar A/B on spill+disk
+        # -- two lines per figure (inline green, sidecar purple), cost + wall.
+        if plot_summary.make_merge_ab_figure(merge_hist, outdir / "merge_summary.png"):
             rendered.append("merge_summary")
         if plot_summary.make_diagnostics_figure(
             merge_hist,
             plot_summary.POINT_PHASE_PANELS,
             "commit",
             outdir / "merge_phases.png",
-            "zagg per-merge point — per-phase seconds vs merge",
+            "zagg per-merge point — per-phase seconds vs merge (inline vs sidecar)",
+            arm_col="target",
         ):
             rendered.append("merge_phases")
     latest = _latest_of(merge_hist).to_dict(orient="records")
--- a/.github/scripts/plot_series.py
+++ b/.github/scripts/plot_series.py
@@ -279,32 +279,36 @@ def _render_release_table(recs: list[dict], title: str, out_png: Path) -> bool:
     if not recs:
         return False
     fmt = bench_metrics._fmt
-    headers = [
-        "target",
-        "release",
-        "shards",
-        "obs",
-        "first consumer (s)",
-        "wall (s)",
-        "total $",
-        "mem (MB)",
-    ]
+
+    def _has(key):
+        return any(r.get(key) is not None and r.get(key) == r.get(key) for r in recs)
+
+    # Schema-adaptive so the SAME renderer serves the point and raster legs: the
+    # point series records ``total_obs`` + ``first_consumer_s``; the raster
+    # series records ``slabs_written`` and computes no first-consumer. Show a
+    # column only where the leg actually has the data (else the raster table
+    # carried two permanently-n/a columns and an over-claiming caption).
+    has_fc = _has("first_consumer_s")
+    count_key = "total_obs" if _has("total_obs") else "slabs_written"
+    count_hdr = "obs" if count_key == "total_obs" else "slabs"
+    headers = ["target", "release", "shards", count_hdr]
+    if has_fc:
+        headers.append("first consumer (s)")
+    headers += ["wall (s)", "total $", "mem (MB)"]
     cell_text = []
     for r in recs:
         cost_cell = fmt(r.get("total_cost_usd"), ".5f")
         cost_cell = f"${cost_cell}" if cost_cell != "n/a" else "n/a"
-        cell_text.append(
-            [
-                str(r.get("target") or "")[-26:],
-                str(r.get("ref") or r.get("commit") or "")[:14],
-                fmt(r.get("n_shards_ok"), ",.0f"),
-                fmt(r.get("total_obs"), ",.0f"),
-                fmt(r.get("first_consumer_s"), ".1f"),
-                fmt(r.get("total_wall_s"), ".1f"),
-                cost_cell,
-                fmt(r.get("max_memory_mb"), ".0f"),
-            ]
-        )
+        row = [
+            str(r.get("target") or "")[-26:],
+            str(r.get("ref") or r.get("commit") or "")[:14],
+            fmt(r.get("n_shards_ok"), ",.0f"),
+            fmt(r.get(count_key), ",.0f"),
+        ]
+        if has_fc:
+            row.append(fmt(r.get("first_consumer_s"), ".1f"))
+        row += [fmt(r.get("total_wall_s"), ".1f"), cost_cell, fmt(r.get("max_memory_mb"), ".0f")]
+        cell_text.append(row)
     fig, ax = plt.subplots(figsize=(11, 0.5 * len(recs) + 1.2))
     ax.axis("off")
     table = ax.table(cellText=cell_text, colLabels=headers, loc="center", cellLoc="right")
@@ -861,8 +865,11 @@ def write_index(
         f'<h3>{title}</h3>\n<img src="{name}.png" alt="{name}">'
         for name, title in (
             ("full_aoi_summary", "Summary — total billed cost (lambda-s ⇔ USD) and wall"),
-            ("full_aoi_point_table", "Point pipeline — latest release (with time-to-first-consumer)"),
-            ("full_aoi_raster_table", "Raster pipeline — latest release (with time-to-first-consumer)"),
+            (
+                "full_aoi_point_table",
+                "Point pipeline — latest release (with time-to-first-consumer)",
+            ),
+            ("full_aoi_raster_table", "Raster pipeline — latest release"),
             ("full_aoi_point_phases", "Point pipeline — per-phase seconds (max shard)"),
             (
                 "full_aoi_raster_phases",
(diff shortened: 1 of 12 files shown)


.github/scripts/plot_series.py
```diff
@@ -996,6 +1055,23 @@ def main(argv: list[str] | None = None) -> int:
...
+    point_latest = _latest_of(point_hist).to_dict(orient="records")
+    if _render_release_table(
+        point_latest,
+        f"zagg full-AOI point (release {str((point_latest[0] if point_latest else {}).get('ref') or '')})",
+        outdir / "full_aoi_point_table.png",
+    ):
+        rendered.append("full_aoi_point_table")
+    raster_latest = _latest_of(raster_hist).to_dict(orient="records")
+    if _render_release_table(
+        raster_latest,
```
espg, line 1076: 🤖 *from Claude (review)*

The raster release table reuses the point-schema `_render_release_table`, but two of its columns are permanently `n/a` for raster because the raster series doesn't carry those keys:

- **`obs`** → `_render_release_table` reads `r.get("total_obs")` (plot_series.py:301), but `RASTER_COLUMNS` has no `total_obs` — the raster observation tally is recorded as `slabs_written` (run_raster_benchmark records `slabs_written=summary.get("total_obs")`, and `records_to_frame` reindexes to `RASTER_COLUMNS`, dropping anything else). So the value *exists* but under a different name and the raster "obs" column renders blank.
- **`first consumer (s)`** → `r.get("first_consumer_s")` (plot_series.py:302) is not in `RASTER_COLUMNS`, and only `run_full_aoi_benchmark.py` (the point harness) computes it — the raster harness never does. So this column is always `n/a` for raster, yet the Pages caption at plot_series.py:865 promises "Raster pipeline — latest release (with time-to-first-consumer)".

Net: the raster table renders (no crash — every cell is None-safe via `_fmt`), but ships with two dead/misleading columns. `test_release_table_renders_with_first_consumer` only feeds a point-shaped record, so nothing catches the raster mismatch.

Options: map `total_obs` → `slabs_written` (and drop or relabel the first-consumer column) for the raster call, or split into a raster-specific renderer, or at minimum drop the "with time-to-first-consumer" promise from the raster caption.

espg: 🤖 *from Claude*
Fixed in b46e73c. `_render_release_table` is now schema-adaptive: the first-consumer column appears only when a record carries `first_consumer_s` (point leg), and the count column reads `total_obs` (point, header "obs") or falls back to `slabs_written` (raster, header "slabs") — so the raster table no longer shows two n/a columns. The raster Pages caption dropped the "(with time-to-first-consumer)" claim. Added `test_release_table_adapts_to_raster_schema` (raster-shaped record → no first-consumer column + slabs; point-shaped → first-consumer + obs).

PR #275, opened 2026-07-18
Shrink async-dispatch wall overhead: result-ready LastModified wall + overlapped manifest check

### LINKED ISSUE #274: Shrink async-dispatch wall overhead: result-ready LastModified wall + overlapped manifest check
🤖 *from Claude*

## Shrink the async-dispatch wall overhead (light-shard latency)

On a light shard the end-to-end **wall** is ~20 s longer than the **worker** compute, and almost all of that gap is dispatch overhead, not work. Measured on a real NEON o9 spill run (`process-shard-4096-disk`, us-west-2):

- wall **121.9 s** vs worker_max_s **102.6 s** → **19.3 s gap**.
- Phase timings: `setup_s` **3.67 s** (cold) — but the *second* arm of the same A/B on a warm container was **0.18 s**; `finalize_s` **3.16 s** (both arms, warm or cold); `fanout_s − worker_max_s` ≈ **8.8 s** (async-queue + cold-start + poll-detection + result write); ~3.7 s client/dispatch overhead.

Three separable contributors, none touching the worker. Two are worth fixing now (espg-ratified design below); the third is out of scope.

### Which wall we report (decision)

**The benchmark's headline wall is "result-ready"** — as soon as the results are finished and usable by a consumer (espg). For a hive store the `morton_hive.json` manifest is written **async at init/setup**, *before* the workers finish, so the store is consumer-readable the instant the worker posts its `.status` result. Therefore result-ready = the worker's `.status` post time, and the finalize backstop (which runs *after* that) is excluded from the reported wall — Fix 1 measures to exactly this point, and Fix 2 removes the backstop from the healthy run's real latency too.

Run-level: a consumer *could* start on partial results as each shard lands, but for our purposes **result-ready = when ALL shards of the run are consumable = `max` over shards of the `.status` post time**. For the single-shard benchmark that's just the one shard; for a multi-shard release run it's the slowest shard's post time.

### Fix 1 — measure wall to the true result-post time (`LastModified`)

The poll-detection lag (avg ~2.75 s of the gap) is a **measurement artifact**: the worker posts its `.status/<run>/<label>.json` result at time T, but the dispatcher records the wall at T + (however long until its next 5 s poll caught it). We don't care *when we noticed* — we want *when the result landed*.

**Change:** the result object's S3 `LastModified` **is** the post time. `_fetch_result` (`runner.py:3161`) already GETs the object via `obstore.get`; return its `.meta.last_modified` alongside the body, and in `_poll_lambda_result` (`runner.py:3099`) set `wall_time = post_time.timestamp() - wall_start` instead of `time.time() - wall_start`.

- Removes the poll-detection lag from the reported wall with **zero** change to the poll interval, worker, or S3 GET rate — the wall is the same true value whether we poll every 5 s or 30 s.
- **Decision (espg): `LastModified` second-granularity (~±1 s) is accepted** — still ~5× tighter than the poll-lag, and works against the already-deployed fleet (no worker change). A finer sub-second option (worker stamps a `posted_at` ms UTC in the body) is explicitly **not** needed now.
- The poll interval (`_ASYNC_POLL_INTERVAL_S = 5.0`, `runner.py:2898`) stays as-is — this supersedes any idea of tightening it.

### Fix 2 — replace the blocking finalize backstop with an overlapped manifest check

`_invoke_lambda_finalize` (`runner.py:3604`) is a **synchronous `RequestResponse`** invoke the dispatcher blocks on (~3.2 s: a cold finalize container + the round-trip). On hive it runs `ensure_manifest` (`lambda_handler.py:765`), the **idempotent backstop** for the `morton_hive.json` already written async at init (#252 hybrid). On the happy path (the async init write succeeded — virtually always) it GETs the manifest, sees matching frozen keys, and returns **without a PUT** — pure overhead. It's only load-bearing in the rare case the async init write was lost (worker Event invokes run `retries: 0`, #151).

**Change (espg's design):** the manifest is required reader-facing schema (D6), so we still guarantee it exists — but confirm it **client-side, overlapped with the run**, instead of a blocking invoke:

- During the fan-out (already hundreds of seconds), a lightweight background checker GETs `morton_hive.json` **every ~10 s**; the moment it's present, **set a flag and stop checking** (a `threading.Event`, since it runs alongside the per-shard result pollers). The dispatcher does **not** block on this — it overlaps the fan-out, so it costs ~0 wall (the async init write fires at setup and the fan-out ≫ 10 s, so the flag is set on the first check in practice).
- At finalize: flag **set** → **skip the backstop invoke entirely** (happy path, 0 wall). Flag **unset** (rare lost async write) → invoke the finalize backstop to self-heal (unchanged behavior).

Removes the ~3.2 s finalize invoke from every healthy hive run. Flat-layout finalize (metadata consolidation, `mode="finalize"` / issue #191) is a different path and out of scope — this is hive-only (the `ensure_manifest` backstop).

### Non-goals
- **Not** tightening the poll interval — Fix 1 supersedes it (bounding the 1700-worker fan-out S3 GET rate is still the interval's only job).
- **Cold start** (~2.8 s init on the first invoke) — the benchmark *forces* cold via `_force_cold_containers` (`runner.py:3173`) for honest numbers; production-warm skips it. Not "waste," out of scope.

### Tests
- Fix 1: `_poll_lambda_result` computes `wall_time` from a mocked result whose `LastModified` is earlier than "now" → asserts the wall is the post-delta, not the detection-delta; the poll interval doesn't affect it.
- Fix 2: hive fan-out where the manifest GET returns present → finalize backstop invoke is **not** called (flag set); manifest absent through the run → backstop **is** invoked (flag unset, self-heal); the checker thread stops on first hit and joins cleanly.

Scope: `src/zagg/runner.py` (+ possibly `deployment/aws/lambda_handler.py` only if the finalize event shape changes — it should not). Independent of the spill/benchmark work (#272 / PR #273). Ship as its own PR (`claude/dispatch-wall`).

espg:
Closes #274

## What

Shrinks the async-dispatch wall overhead measured on light shards (~19 s gap between end-to-end wall and worker compute), per the espg-ratified design on #274. Two separable, worker-independent fixes; both scoped to `src/zagg/runner.py`. The finalize event shape is unchanged.

### Fix 1 — result-ready wall via S3 `LastModified`

The poll-detection lag was a measurement artifact: the worker POSTs its `.status` result at time T, but the dispatcher recorded the wall at T + (however long until its next 5 s poll caught it).

- `_fetch_result` now returns `(envelope, last_modified)` — the result object's obstore `GetResult.meta["last_modified"]` (an `ObjectMeta` TypedDict) IS the worker's POST time.
- `_poll_lambda_result` sets `wall_time = post_time.timestamp() - wall_start` (falling back to the real clock only if the store reports no timestamp).
- The poll interval (`_ASYNC_POLL_INTERVAL_S = 5.0`) is **unchanged**; deadline/timeout logic stays on `time.time()`. Only the *reported* per-shard wall uses the post time. `±1 s` `LastModified` granularity is accepted (issue decision).
- Run-level result-ready = max over shards of the post time; that aggregation already lives in the benchmark consumer reading `results[*]["wall_time"]` — no new summary key invented.

### Fix 2 — overlapped manifest check replaces the blocking finalize backstop (HIVE only)

`_invoke_lambda_finalize` was a synchronous `RequestResponse` invoke (~3.2 s) the dispatcher blocked on as the idempotent `morton_hive.json` backstop, pure overhead on the happy path (the async init write of #252 virtually always succeeds).

- During the hive fan-out a lightweight background checker (`_start_manifest_checker`) GETs `morton_hive.json` every ~10 s, sets a `threading.Event` the moment it's present, then stops. It overlaps the (hundreds-of-seconds) fan-out and is joined at finalize — the dispatcher never blocks on it.
- At finalize: flag **set** → skip the backstop invoke entirely (happy path, ~0 wall); flag **unset** (rare lost retries-0 Event write) → invoke the finalize backstop to self-heal (unchanged path).
- Flat-layout finalize (metadata consolidation, `mode="finalize"`, #191) is a different path and untouched. Store credentials/obstore kwargs are shared with the result poller via a new `_output_store_kwargs` helper.

## Phases

- [x] Phase 1 — Fix 1 (result-ready `LastModified` wall) + tests
- [x] Phase 2 — Fix 2 (overlapped manifest check) + tests

## How tested

- `uv run pytest tests/test_runner.py tests/test_dispatch.py tests/test_runner_concurrency.py -m "not slow"` green.
- Fix 1: new `test_wall_is_result_post_time_not_detection` — mocked fetch whose `LastModified` is 3 s past `wall_start` with no-op sleeps → asserts `wall_time == 3 s` (not the ~0 s detection time), identical at poll intervals 5 s and 30 s. Existing async/temporal/`_result_fetcher` tests updated to the `(envelope, LastModified)` fetch contract.
- Fix 2: `TestManifestChecker` — `_start_manifest_checker` unit tests (hit sets the flag + thread joins; absent re-checks across ticks then stops; a transient predicate fault is swallowed), plus hive `_run_lambda` integration: manifest present -> `_invoke_lambda_finalize` NOT called + checker joined (no hang); absent -> it IS called (self-heal); a wiring test asserts the predicate reads `morton_hive.json` at the store root.
- `ruff check` / `ruff format --check` clean.

## Questions for review

- (none yet)

### DIFF
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -16,6 +16,7 @@ import logging
 import os
 import random
 import statistics
+import threading
 import time
 import uuid
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -2493,11 +2494,28 @@ def _run_lambda(
     # manifest inputs (config + dataset identity from the ShardMap metadata,
     # the same source as the local path) ride the finalize event; flat
     # finalize events stay byte-identical.
+    # Hive manifest checker handles (issue #274 Fix 2): the background thread is
+    # started after the async setup invoke (below) and joined at fan-out end;
+    # _finalize_fn reads ``manifest_found`` to decide whether the backstop invoke
+    # is still needed. None on the flat path (no checker runs there).
+    manifest_found = None
+    manifest_stop = None
+    manifest_thread = None
+
     if get_store_layout(config) == "hive":
         md = catalog_data.get("metadata") or {}
         dataset = {"short_name": md.get("short_name"), "version": md.get("version")}
 
         def _finalize_fn():
+            # issue #274 Fix 2: the checker overlapped the fan-out. Manifest
+            # present -> the async init write (#252) landed; SKIP the backstop
+            # invoke (happy path, ~0 wall). Absent -> rare lost retries-0 Event
+            # write; invoke the finalize backstop to self-heal (unchanged path).
+            if manifest_found is not None and manifest_found.is_set():
+                logger.info(
+                    "morton_hive.json present (overlapped check) -- skipping finalize backstop"
+                )
+                return None
             return _invoke_lambda_finalize(
                 state["lambda_client"],
                 function_name,
@@ -2574,6 +2592,16 @@ def _run_lambda(
             overwrite=overwrite,
             output_creds_event=output_creds_event,
         )
+        # Overlap the fan-out with the client-side morton_hive.json check
+        # (issue #274 Fix 2): the async setup write above typically lands
+        # within seconds, so by finalize the flag is set and the blocking
+        # backstop invoke is skipped. Reads via the poller's output-store path.
+        from zagg.hive import read_manifest
+
+        manifest_kwargs = _output_store_kwargs(output_creds_event, region)
+        manifest_found, manifest_stop, manifest_thread = _start_manifest_checker(
+            lambda: read_manifest(store_path, **manifest_kwargs) is not None
+        )
     else:
         _invoke_lambda_setup(
             state["lambda_client"],
@@ -2622,6 +2650,13 @@ def _run_lambda(
         )
     finally:
         executor.shutdown()
+        # Stop the overlapped manifest checker and read its verdict (#274 Fix 2).
+        # It ran the whole fan-out; stopping now wakes it from its interval wait
+        # and joins cleanly (a final check already ran), leaving manifest_found
+        # settled before _finalize_fn consults it.
+        if manifest_stop is not None:
+            manifest_stop.set()
+            manifest_thread.join(timeout=_MANIFEST_CHECK_JOIN_TIMEOUT_S)
     fanout_s = time.time() - start_time
 
     # Consolidate metadata via Lambda (same rationale as setup -- avoids
@@ -2897,6 +2932,14 @@ _DEFAULT_FUNCTION_TIMEOUT_S = 900
 _ASYNC_POLL_MARGIN_S = 90.0
 _ASYNC_POLL_INTERVAL_S = 5.0
 
+# Hive manifest checker (issue #274 Fix 2). The morton_hive.json backstop is
+# confirmed CLIENT-side, overlapped with the fan-out, instead of a blocking
+# finalize invoke: a background thread GETs the manifest every INTERVAL until
+# it's present (async init write, #252), then the finalize backstop is skipped.
+# JOIN_TIMEOUT bounds the wait when stopping the checker at fan-out end.
+_MANIFEST_CHECK_INTERVAL_S = 10.0
+_MANIFEST_CHECK_JOIN_TIMEOUT_S = 30.0
+
 # Async (Event) invoke requests cap at 256 KB (vs 6 MB synchronous). Budget a
 # little under it so the dispatch pre-flight fails with a remedy before
 # Lambda's raw RequestEntityTooLargeException does; the realistic trigger is a
@@ -2975,6 +3018,39 @@ def _fetch_result(result_store, key):
     except (FileNotFoundError, NotFoundError):
         return None
     return json.loads(bytes(data)), last_modified
+
+
+def _start_manifest_checker(check_present, *, interval_s=_MANIFEST_CHECK_INTERVAL_S):
+    """Overlap the hive fan-out with a background morton_hive.json check (#274 Fix 2).
+
+    ``check_present`` is a zero-arg predicate returning True once the manifest is
+    readable. The moment it is, ``found`` is set and the thread stops; otherwise
+    it re-checks every ``interval_s`` until ``stop`` is set (the caller sets it and
+    joins at fan-out end). The dispatcher NEVER blocks on this -- it overlaps the
+    (hundreds-of-seconds) fan-out, so the async init-time manifest write (#252) has
+    landed by the first check in practice and the finalize backstop is skipped. A
+    transient GET fault is swallowed and retried next tick, never fatal. At least
+    one check always runs (even if ``stop`` is already set), so joining right after
+    an instant fan-out still yields a definitive present/absent verdict. Returns
+    ``(found, stop, thread)``; the thread is a daemon.
+    """
+    found = threading.Event()
+    stop = threading.Event()
+
+    def _run():
+        while True:
+            try:
+                if check_present():
+                    found.set()
+            except Exception:
+                pass  # transient GET fault -- re-check next tick, never fatal
+            if found.is_set() or stop.is_set():
+                return
+            stop.wait(interval_s)
+
+    thread = threading.Thread(target=_run, name="hive-manifest-check", daemon=True)
+    thread.start()
+    return found, stop, thread
 
 
 def _invoke_lambda_event(
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -2919,27 +2919,37 @@ _POLL_RETRY_CONFIG = {
 }
 
 
+def _output_store_kwargs(output_creds_event, region):
+    """obstore kwargs for the output store, shared by the async result poller
+    and the hive manifest checker (issues #151, #274).
+
+    Mirrors the worker's ``_output_store_kwargs``: the explicit
+    output-credentials block when supplied, else the ambient chain. Carries the
+    short per-request :data:`_POLL_RETRY_CONFIG` (issue #186) so a single 5xx
+    can't block for minutes and silently overrun a poll/check deadline.
+    """
+    kwargs = {"region": region, "retry_config": _POLL_RETRY_CONFIG}
+    if output_creds_event:
+        kwargs["region"] = output_creds_event.get("region", region)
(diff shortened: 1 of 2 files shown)


espg:
🤖 *from Claude (review)*

**Phase 1 review (commit `ab8f718`, "result-ready wall via S3 LastModified") — looks correct.** Read the full diff against `origin/main`, verified the obstore API against the installed version, and ran the affected tests. No material findings; the change matches the issue #274 Fix 1 design. Detail on what I checked:

- **Tuple-unpack contract is contained.** `_fetch_result` is only ever consumed via the `_result_fetcher` closure, and every `result_fetch()` call goes through `_poll_lambda_result` (grepped `src/`, `tests/`, `deployment/` — the temporal `_invoke_lambda_event` and the spatial dispatcher are the only two callers, and both delegate the unpack to `_poll_lambda_result`). No bare-dict consumer was left behind. Tests were updated to the `(envelope, LastModified)` shape at all three fixture sites.
- **obstore API is right.** On the pinned `obstore==0.11.0`, `obstore.get(...).meta` is a plain `dict` with a tz-aware UTC `last_modified` (`datetime`, `tzinfo=UTC`), so `.meta["last_modified"]` (subscript, not the `.meta.last_modified` attribute the issue prose sketched) is correct, and `post_time.timestamp()` yields POSIX seconds on the same epoch as `wall_start = time.time()`. Reading `.meta` before `.bytes()` is safe (meta comes from response headers) — verified by round-trip.
- **`_output_store_kwargs` refactor is byte-for-byte equivalent** to the old inline block (region override, `credentials`, `endpoint_url`, `retry_config`).
- **Deadline logic correctly still uses `time.time()`** (real-clock timeout); only the *reported* success wall switches to post-time. The `post_time is None` fallback is unreachable in practice (S3 always stamps LastModified) but harmless as defensive code.
- **`test_wall_is_result_post_time_not_detection` is not a trivial pass:** with `sleep` no-op'd, detection would give ~0s and the `approx(3.0, abs=0.5)` assertion would fail — so it genuinely distinguishes post-time (3.0s) from detection-time, and the two misses run well under the 60s deadline without tripping the error path.

Two non-blocking observations for your awareness (neither is a phase-1 bug):
- **Clock-skew edge:** `post_time` (S3 service clock) minus `wall_start` (client clock) introduces a small new skew term the old `time.time()-time.time()` form didn't have; on an NTP-synced fleet with ~100s shards it's sub-second and well inside the accepted ±1s LastModified granularity, so it only matters pathologically (could nudge a very fast result slightly negative).
- **Multi-shard origin:** the per-shard `wall_time` is `post_time - per_shard_wall_start`, so a future run-level rollup of `max(per_shard wall_time)` is not identical to `max(post_time) - run_start` when shards are staggered. Out of scope here (issue notes the single-shard bench is one shard), just flagging for when Fix 2 / the bench aggregation lands.

FYI (not part of this commit): the working tree currently carries an unused `import threading` from the in-progress, uncommitted phase-2 — it will fail the ruff bot if committed as-is, but it's absent from `ab8f718`, which is ruff-clean.

espg (commented):
🤖 *from Claude (review)*

**Phase 2 (`3c2b540`, overlapped hive manifest check) reviewed — correct, no changes requested.** Adversarial pass on the probes that mattered:

- **Closure/scoping.** `manifest_found` is a proper cell variable: it is initialized to `None` before `_finalize_fn` is defined and reassigned later in the same `_run_lambda` scope, so the hive `_finalize_fn` reads it at *call* time via the closure cell (not the `None` snapshot from definition time). No `UnboundLocal` — it is always bound at the top before any branch. The flat/consolidate `_finalize_fn` never references it. Verified against `runner.py:2503-2528` and `2601-2606`.
- **Ordering / no race.** The `finally` runs `executor.shutdown()` then `manifest_stop.set(); manifest_thread.join(...)` (`2652-2662`); `executor.finalize()` (which calls `_finalize_fn`) runs *after* the `finally` (`2670`). So the flag is settled before it is read — airtight, no mid-check read.
- **Thread lifecycle / hangs.** Checker is a daemon; join is bounded by `_MANIFEST_CHECK_JOIN_TIMEOUT_S`; a blocking GET can outlast the join but the daemon can never hang process exit, and an un-set flag then just falls through to the (idempotent) backstop invoke — safe fallback. On a healthy run the thread has already exited on first hit, so `stop.set()`+`join` is ~instant → no wall added, preserving the overlap intent. If `dispatch()` raises, the checker is still stopped (it is in the `finally`); `finalize` is skipped exactly as before this PR.
- **`_run` loop.** `while True` runs `check_present()` first, so ≥1 check even if `stop` is pre-set; faults are swallowed and retried next tick; prompt exit on `found`/`stop` before the `stop.wait(interval_s)`. Matches the docstring.
- **Backstop unchanged / flat untouched.** Manifest-absent path invokes `_invoke_lambda_finalize` with identical args; flat path keeps `manifest_found=None` and its own `_finalize_fn`.
- **Predicate late-binding.** `from zagg.hive import read_manifest` executes at runtime, after the tests' `monkeypatch` of `zagg.hive.read_manifest`, so the patched fn is bound; `_output_store_kwargs(output_creds_event, region)` is the same kwargs builder the result poller uses (`2601` / `2995`), carrying region/credentials/endpoint_url + retry_config. `read_manifest` reads `morton_hive.json` at the store root, confirmed by `test_checker_predicate_reads_manifest_at_store_root`.
- **Tests are real, not trivially passing.** Skip-vs-invoke is proven through the actual closure (`n==0` vs `n==1`); the present-case asserts the mock thread was joined (ordering); the `_start_manifest_checker` unit tests exercise a real thread — first-hit stop+join, re-check across ticks (`next(calls) > 1`), and fault-swallowing. Ran the class 5× (`uv run --extra test pytest tests/test_runner.py::TestManifestChecker`): 6 passed each time, no flakiness.

The local `from zagg.hive import ...` follows the established in-function import pattern elsewhere in `runner.py` (753/871/2076/…). Comments/docstrings match behavior; no dead code. `runner.py` is ~3978 lines but that is pre-existing, not introduced here.

Issue #280, opened 2026-07-18
Parallelize the spill reducer (intra-block partitions + inter-block k-way) — blocked by #279


espg:
🤖 *from Claude*

## Summary

Follow-on to #279. Once the t-digest reduce loop is vectorized numpy (GIL-releasing) and the fold is a single order-independent k-way merge, the `SpillAggregator` reducer can be parallelized for another ~2–4× on heavy shards. This is an **engine-level** change to `src/zagg/processing/spill.py` (the reducer thread + the memory budget), *not* a user-side agg function — at most gated by a worker knob (e.g. `streaming.reduce_workers: N`). Like #279 it ships in the packaged zagg, so fleet use needs a layer redeploy; it is fully testable locally without one.

**Blocked by #279** — specifically:
- The reduce hot loop is a Python-scalar k1 loop today (holds the GIL), so threading it buys nothing on CPU until #279 makes it vectorized numpy. Only then does parallelism scale the reduce.
- Cross-block parallelism (Flavor B below) is only *correct* once #279's k-way merge decouples per-block reduction from the final order-independent combine.

## Current model

`_close_block` → `_reduce_one` on one `threading.Thread`; `_join_reducer` before the next close. So reads overlap **one** block's reduce, but blocks reduce strictly serially in close order, and `_fold_block` walks its partitions sequentially (spill.py:598-621).

## Two flavors of parallelism

**Flavor A — intra-block, across partitions (clean, no correctness change).**
`_fold_block` loops `for key in block.partition_keys()`. Partitions are disjoint morton cell-groups — within a block a cell lives in exactly one partition — so folding partitions concurrently touches **disjoint cells**. No associativity/order issue. Each partition worker returns its own `{cell: digest}` and the disjoint results are `dict.update`-merged single-threaded (avoid mutating shared `self._digests`/`self._counts` from N threads). Parallelizes both the partition read-back I/O and the (post-#279) numpy CPU. Realistic degree = Lambda vCPUs (~2 near 1.8 GB, up to 6 at 10 GB).

**Flavor B — inter-block, parallel block reduces (rides on #279 Phase 3).**
The same cell recurs in every block, so folding two blocks concurrently into shared `self._digests[cell]` is racy *and* order-dependent. With #279's k-way merge, each block reduces independently to its own per-block-per-cell digest (no shared state), then a single k-way merge per cell combines across blocks at the end — parallel-safe and block-order-invariant.

## Central design question — the memory budget

[...]

PR #282, opened 2026-07-18
Vectorize t-digest build/merge + order-independent k-way fold (issue #279)

### LINKED ISSUE #279: t-digest agg hot path is unvectorized: ~15× build + ~6× fold available in pure numpy
🤖 *from Claude*

## Summary

Since the spill/read-path work in #270 (and the benchmark changes in #273) shifted wall-clock off the read path, the t-digest **aggregation** compute is now the tall pole on very heavy shards — hundreds of seconds spent building initial digests and folding in follow-on blocks. The cause is not that this needs a compiled extension; it's that the hot path in `src/zagg/stats/tdigest.py` is essentially **unvectorized**. There is a ~15× build speedup and a ~6× fold speedup available in **pure numpy** (Lambda-clean, no new deps).

## Diagnosis

Both `build_tdigest` and `merge_tdigests` run the k1 compression as a **scalar Python `for` loop over every element** (`tdigest.py:213` and `tdigest.py:336`), and each iteration calls `_k1_scale` (`tdigest.py:88`), which evaluates `float(np.arcsin(<scalar>))` — a full numpy dispatch **per observation**. The only vectorized step in the whole routine is the initial `np.sort`. So on a heavy shard the compression loop is where the time goes, and it barely touches numpy.

The fold path compounds it: `streaming.flush` (`streaming.py:207-211`) and `spill._fold_block` (`spill.py:616-620`) accumulate a cell with a **left-fold of pairwise `merge_tdigests`** — one re-sort + re-compression per block. That's O(blocks) compressions per cell.

A side effect worth calling out (it changes how we should frame Phase 3): t-digest merge is **not associative/commutative** — `(A⊕B)⊕C ≠ A⊕(B⊕C)`. Folding the same 50 blocks forward vs. reversed produced **different digests** (600 vs 601 centroids, not byte-equal). The output is reproducible run-to-run today only because block arrival order is currently pinned (flush order / serialized block-close order); it is *not* order-independent, so anything that reorders block completion (e.g. parallelizing the spill overlap reducer) would silently change the bytes. Pairwise folding also drifts *above* the δ budget (600 vs 512 centroids).

## Measurements

Local microbench, `normal` data, δ=512 (harness attached below):

**Build** — `build_tdigest(values)`:

| n obs | current | vectorized-k1 (identical output) | fully vectorized |
|---:|---:|---:|---:|
| 100k | 74 ms | 31 ms | 4.2 ms |
| 500k | 375 ms | 158 ms | 23 ms |
| 1M | 765 ms | 315 ms | **50 ms (15×)** |

**Fold** — pairwise `merge_tdigests` left-fold vs a single k-way merge:

| blocks × obs | pairwise fold (current) | single k-way merge |
|---:|---:|---:|
| 20 × 5k | 23 ms | 5 ms (4.7×) |
| 50 × 5k | 62 ms | 10 ms (6.0×) |
| 100 × 10k | 121 ms | 19 ms (6.4×) |

The k-way column also produces digests at the δ budget (~515 vs 600 centroids). Accuracy against the true pooled quantiles is a **wash-to-slight-trade**, *not* a k-way win: k-way is marginally better mid-distribution, while pairwise's over-budget drift happens to give marginally finer resolution at the extreme (p99.9) tail. All three (pairwise / k-way / single-build-over-raw) sit within a hair of the true quantiles. So the case for k-way is **speed + order-independence + budget-adherence**, not accuracy — see the fold phase below.

## Proposed plan (phased)

**Phase 1 — vectorize the k1 evaluations (byte-identical output).**
Precompute the whole `k1(rank)` vector with one `np.arcsin` call before the loop, replacing the per-element scalar `np.arcsin`. The loop stays; each step becomes plain float arithmetic. ~2.4× for free, **zero** accuracy/parity change. Lands safely against the golden parity vectors with no expected-output churn. This is the low-risk beachhead.

**Phase 2 — replace the per-element build loop with a per-centroid loop (~15×).**
`k1_inv` is closed-form (`q = 0.5*(sin(π(m/δ − 0.5)) + 1)`), so from each centroid's left edge we can jump directly to the next boundary rank instead of testing every observation — **≤ δ (~512) Python iterations regardless of n** — then compute means/weights with `np.add.reduceat` over a `cumsum`. Keeping the exact greedy-reset boundary rule makes this reproducible byte-for-byte against Phase 1's output (verify against the golden vectors). The "fully vectorized" column above uses the slightly-different *ideal-threshold* variant; the reduceat-on-greedy-boundaries version gets the same speed with exact parity.

**Phase 3 — k-way merge in the fold instead of pairwise left-fold (~6×).**
In `streaming.flush` and `spill._fold_block`, collect all of a cell's per-block digests and compress once (`concatenate → argsort → single k1 scan`) instead of `merge_tdigests(held, fresh)` per block. Fewer sorts, fewer re-compressions. This is the change that most directly attacks the "follow-on appends of additional tdigest blocks" cost.

Motivation is **speed + order-independence + budget adherence**, explicitly *not* an accuracy claim (see Measurements). Because merge is non-associative, the pairwise fold is order-dependent; k-way (sort-then-compress-once) is invariant to block arrival order, which removes a latent reproducibility hazard as the spill path parallelizes. It also holds the centroid count at δ, keeping object size predictable (#209). The far-tail resolution the pairwise drift accidentally buys is better recovered, if wanted, by raising δ (Phase 2 makes that cheap) than by relying on over-budget drift.

Caveat: this changes the merge grouping, so it moves the golden vectors for multi-block cells — expected; re-bless with the documented rationale above, and add a test asserting order-invariance (forward-fold vs reverse-fold bytes now match).

Phases 2 and 3 also carry the located-digest channel (`_centroid_ancestors`, issue #87) unchanged — the boundary/`starts` computation is shared, so the location reduction rides along without extra work.

## Acceptance criteria

- `ruff check`/`ruff format --check`/`pytest -v` green; `pre-commit run --all-files` clean.
- Phase 1: existing t-digest tests + golden parity vectors pass **unchanged**.
- Phase 2: output byte-identical to Phase 1 on the golden vectors (or, if we accept the ideal-threshold variant, a documented re-bless with p50/p99 error bounds shown to be equal-or-better).
- Phase 3: multi-block golden vectors re-blessed with rationale; single-block path unchanged.
- A benchmark test (marked `slow`) asserting the build/fold speedups hold, so we don't regress back into the scalar loop.

## Open questions

1. **Parity strictness for Phase 2** — do we want strict byte-identity (reduceat over the exact greedy boundaries), or is the ideal-threshold variant (simpler, marginally *more* accurate, but different bytes) acceptable with a re-bless? Recommendation: strict byte-identity, to keep Phase 2 a pure speedup.
2. **Whether to also raise δ** now that heavy shards are cheap — out of scope here, but the 15× headroom makes a larger default δ (more tail accuracy) affordable if desired. Flagging, not proposing.
3. Rust/compiled path: explicitly **not** proposed. It would add maybe 2–4× on top of vectorized numpy while adding the cross-arch Lambda-layer build burden (§7). The numpy wins should take heavy shards from hundreds of seconds to tens; revisit only if that's still not enough.

Add `implement` (and I'll open a draft PR working the phases) or `plan`/`discuss` if you'd rather iterate on the approach first. Benchmark harness (`bench_td.py`, `bench_merge.py`) ready to fold into the PR.

espg:
Closes #279.

## What & why

After #270/#273 moved wall-clock off the read path, the t-digest **aggregation** compute became the tall pole on heavy shards — the hot path in `src/zagg/stats/tdigest.py` was essentially unvectorized (a scalar Python loop with a `float(np.arcsin(...))` per observation). This PR vectorizes it in **pure numpy** (no new deps, Lambda-clean) for ~15× on build and ~6× on the spill fold, and adds an order-independent k-way fold.

Approach:
- **One shared compression core** `_compress(means, weights, delta)` replaces the per-element scalar loop with a precomputed-k1 array + one `searchsorted` per centroid, so the loop runs ~δ times regardless of n. It reproduces the old greedy rule's **centroid boundaries and weights bit-for-bit** (verified across 200 randomized build trials and 150 merge trials); only the per-centroid means shift at the float-ULP level (single `sum/weight` vs. incremental Welford), which vanishes under the float32 cast. Both `build_tdigest` and `merge_tdigests` route through it.
- **`merge_tdigests_kway(digests, ...)`** — concatenate-sort-compress once, so the result is **order-independent** (t-digest merge is not associative) and holds the centroid count at the δ budget instead of drifting above it.
- **`build_tdigest_pairwise`** — a new reducer with identical build output but the pre-#279 **pairwise** fold law, for anyone who wants the old order-dependent semantics.
- **Fold wiring**: on the spill path, the standard `build_tdigest` folds k-way (parts accumulated per block, collapsed once at finalize); `build_tdigest_pairwise` keeps the running pairwise fold. Merge-mode keeps the pairwise fold for both variants (see caveat below).

## Where each fold law runs

| Path | `build_tdigest` (standard) | `build_tdigest_pairwise` |
|---|---|---|
| Pooled (no `streaming`) | one-shot build, no fold | one-shot build, no fold |
| Merge-mode streaming | pairwise | pairwise |
| Spill, single-block | one-shot build, no fold | one-shot build, no fold |
| Spill, multi-block | **k-way** | pairwise |

k-way runs in exactly one place: the standard reducer on the spill multi-block path — the path #280 will parallelize.

## Why merge-mode isn't k-way

k-way holds each block's per-cell digest as a "part" and compresses once at the end. On spill each part is a saturated ~δ-centroid digest over a large slab and blocks are few, so parts-per-cell is tiny. On the **merge-mode** path flushes are fine-grained: each per-flush per-cell digest is often loss-free (< δ obs → one centroid per obs), so accumulating parts ≈ retaining all raw observations, which would break streaming's memory bound. So merge-mode must stay pairwise — it's forced by memory characteristics, not a preference.

## Phases

- [x] Phase 1 — vectorize `build_tdigest` core (~15×; 765→51 ms at 1M obs). Byte-identical output.
- [x] Phase 2 — vectorize `merge_tdigests` (pairwise) via the shared core (~2× per merge). Byte-identical.
- [x] Phase 3 — `merge_tdigests_kway` + `build_tdigest_pairwise` + fold-law wiring in spill/streaming.

## Benchmarks (local, δ=512, normal data)

| n obs | build before | build after |
|---:|---:|---:|
| 100k | 74 ms | 4.7 ms |
| 1M | 765 ms | 51 ms |

Pairwise fold (50 blocks × 5k): 62 → 31 ms; the k-way single-pass over the same is ~10 ms.

## How tested

- Full suite green: **2390 passed, 34 skipped** (`uv run pytest`). ruff check + format clean; mypy clean on every file this PR touches.
- New unit tests (`tests/test_tdigest.py`): k-way order-independence (reversed + 3 permutations → byte-identical), the contrasting pairwise order-*dependence*, k-way ≠ pairwise past two contributors, weights/bounds/sorted, empties/single/skip-empty, one-shot quantile tolerance, located channel, `build_tdigest_pairwise` identical-to-standard.
- New integration tests (`tests/test_spill.py`): pairwise multi-block byte-identical to merge-mode under real compression; standard k-way diverges from pairwise on multi-block cells while counts stay exact; standard k-way tracks pooled quantiles within tolerance.
- Parity harnesses confirming boundaries/weights bit-identical to the pre-#279 scalar loop (build + merge).

## Questions for review

1. **Mode-dependent output for standard `build_tdigest`.** Because k-way ≠ pairwise, the *same config* under `mode: merge` vs `mode: spill` (multi-block, ≥3 compressing blocks) now yields different digest bytes — both valid within t-digest tolerance, diverging only when compression actually happens. `build_tdigest_pairwise` restores byte-identity with merge-mode (the old #270 contract). Is the mode-dependence acceptable as the default, or should the default reducer stay pairwise with k-way opt-in?
2. **δ unchanged** (still 512) — raising it for far-tail accuracy is out of scope here (deferred per #279); flagging only.
3. **Pre-existing, not addressed here:** `tests/test_lambda_build.py::test_function_build_succeeds` fails locally because `deployment/aws/build_function.sh` resolves against a Python <3.11 interpreter (can't satisfy `zarr>=3.1.5`). Unrelated to this change and a deploy script (§1) — flagging, not touching. The mypy baseline outside these files (`worker.py`, `catalog/*`) also remains dirty and is left alone.

Follow-up #280 (parallelize the spill reducer) is blocked on this landing.

### DIFF
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -55,7 +55,12 @@ from zagg.config import (
     get_data_vars,
     get_output_signature,
 )
-from zagg.stats.tdigest import _DEFAULT_DELTA, build_tdigest, merge_tdigests
+from zagg.stats.tdigest import (
+    _DEFAULT_DELTA,
+    build_tdigest,
+    merge_tdigests,
+    merge_tdigests_kway,
+)
 
 #: Floor for the spill-enable /tmp check: below this, even a degraded
 #: many-block run is pointless — fail at config time instead of thrashing.
@@ -433,12 +438,18 @@ class SpillAggregator:
         except ValueError:
             self._mergeable = False
         self._count_fields: list[str] = []
-        self._digest_fields: dict[str, tuple[str, int]] = {}  # name -> (source, delta)
+        # name -> (source, delta, pairwise). ``pairwise`` selects the cross-block
+        # fold law: False -> order-independent k-way (build_tdigest, the default),
+        # True -> pairwise left-fold (build_tdigest_pairwise). See issue #279.
+        self._digest_fields: dict[str, tuple[str, int, bool]] = {}
         if self._mergeable:
+            from zagg.processing.streaming import _TDIGEST_PAIRWISE_FUNCTION
+
             for name, meta in agg_fields.items():
                 if get_output_signature(meta)["kind"] == "ragged":
                     delta = int((meta.get("params") or {}).get("delta", _DEFAULT_DELTA))
-                    self._digest_fields[name] = (meta.get("source") or "h_li", delta)
+                    pairwise = meta.get("function") == _TDIGEST_PAIRWISE_FUNCTION
+                    self._digest_fields[name] = (meta.get("source") or "h_li", delta, pairwise)
                 else:
                     self._count_fields.append(name)
         if hasattr(grid, "chunk_order") and int(getattr(grid, "chunks_per_shard", 1)) > 1:
@@ -465,6 +476,13 @@ class SpillAggregator:
         # Cross-block mergeable running state (only ever fed on block close).
         self._counts: dict[int, int] = {}
         self._digests: dict[str, dict[int, np.ndarray]] = {n: {} for n in self._digest_fields}
+        # k-way fold accumulator: for order-independent fields, each block's
+        # per-cell digest is stashed here and collapsed once at finalize
+        # (merge_tdigests_kway), rather than pairwise-folded per block. Each part
+        # is a saturated per-block digest and blocks are few, so this stays small.
+        self._digest_parts: dict[str, dict[int, list[np.ndarray]]] = {
+            n: {} for n, (_, _, pairwise) in self._digest_fields.items() if not pairwise
+        }
         # Per-flush unique cell words; unioned lazily by occupied_cells().
         self._occupied: list[np.ndarray] = []
         # Single-block reduce cache: (part_key, col_arrays, cell_to_slice) for
@@ -598,10 +616,12 @@ class SpillAggregator:
     def _fold_block(self, block: SpillBlock) -> None:
         """Fold one block into the running mergeable state, per partition.
 
-        This is the StreamingAggregator merge sequence at block granularity:
-        counts by summation (exact), tdigests built fresh per cell and merged
-        under the field's delta — one merge round per block instead of per
-        buffer, which is the ~6x merge-CPU collapse the design targets.
+        Counts fold by summation (exact). tdigests are built fresh per cell and
+        then, per the field's fold law: **k-way** fields stash the block digest
+        for one order-independent collapse at finalize (:meth:`_finalize_kway`);
+        **pairwise** fields merge it into the running digest here. Either way it
+        is one build round per block instead of per buffer — the ~6x merge-CPU
+        collapse the design targets (issue #279).
         """
         from zagg.processing.aggregate import _group_columns
 
@@ -613,12 +633,30 @@ class SpillAggregator:
             del cells, cols
             for cell, (start, end) in cell_to_slice.items():
                 self._counts[cell] = self._counts.get(cell, 0) + (end - start)
-                for name, (source, delta) in self._digest_fields.items():
+                for name, (source, delta, pairwise) in self._digest_fields.items():
                     fresh = build_tdigest(col_arrays[source][start:end], delta=delta)
-                    held = self._digests[name].get(cell)
-                    self._digests[name][cell] = (
-                        fresh if held is None else merge_tdigests(held, fresh, delta=delta)
-                    )
+                    if pairwise:
+                        held = self._digests[name].get(cell)
+                        self._digests[name][cell] = (
+                            fresh if held is None else merge_tdigests(held, fresh, delta=delta)
+                        )
+                    else:
+                        self._digest_parts[name].setdefault(cell, []).append(fresh)
+
+    def _finalize_kway(self) -> None:
+        """Collapse each k-way field's per-block parts into one digest per cell.
+
+        Runs once, after the final block is folded and before the first emission.
+        The single-pass k-way merge is order-independent (t-digest merge is not
+        associative), so the result does not depend on block reduce order — the
+        property #280 relies on to parallelize the reducer.
+        """
+        for name, cell_parts in self._digest_parts.items():
+            delta = self._digest_fields[name][1]
+            dest = self._digests[name]
+            for cell, parts in cell_parts.items():
+                dest[cell] = merge_tdigests_kway(parts, delta=delta)
+            cell_parts.clear()
 
     # -- post-read emission ----------------------------------------------------
 
@@ -640,6 +678,7 @@ class SpillAggregator:
         key = int(partition_ids(self.grid, children[:1])[0]) if len(children) else 0
         if self._loaded is None or self._loaded[0] != key:
             self._load_partition(key)
+        assert self._loaded is not None  # _load_partition always sets it before returning
         _, col_arrays, cell_to_slice = self._loaded
         chunk_pooled = _pool_chunk_columns(col_arrays, cell_to_slice, children)
         chunk_scalars = _eval_chunk_precompute(self.config, chunk_pooled)
@@ -688,12 +727,14 @@ class SpillAggregator:
         if not self._finalized:
             # Join the in-flight overlap reduce (its failure re-raises here),
             # then fold the final (still-open, never threshold-closed) block
-            # into the running state once, before the first emission.
+            # into the running state once, before the first emission. The k-way
+            # collapse runs after every block is folded so it sees all parts.
             self._join_reducer()
             try:
                 self._fold_block(self._block)
             finally:
                 self._block.close()
+            self._finalize_kway()
             self._finalized = True
         children = np.asarray(children)
         n_cells = len(children)
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -478,8 +478,16 @@ class SpillAggregator:
         self._digests: dict[str, dict[int, np.ndarray]] = {n: {} for n in self._digest_fields}
         # k-way fold accumulator: for order-independent fields, each block's
         # per-cell digest is stashed here and collapsed once at finalize
-        # (merge_tdigests_kway), rather than pairwise-folded per block. Each part
-        # is a saturated per-block digest and blocks are few, so this stays small.
+        # (merge_tdigests_kway), rather than pairwise-folded per block. Holding
+        # all parts to finalize is by design: order-independence forbids
+        # incremental collapse (folding subsets then combining is a tree
+        # reduction of a non-associative op, which reintroduces the very
+        # order-dependence k-way exists to remove). Cost is bounded: each part is
+        # a saturated ≤δ-centroid digest (~4 KB at δ=512, δ·8 bytes), so peak is
+        # ~B×C×(δ·8) over B closed blocks and C occupied cells — small relative
+        # to the multi-GB per-partition build peak the block threshold already
+        # reserves. Tighter memory-budget accounting against that reservation is
+        # owned by the #280 parallel-reduce work.
         self._digest_parts: dict[str, dict[int, list[np.ndarray]]] = {
             n: {} for n, (_, _, pairwise) in self._digest_fields.items() if not pairwise
         }
--- a/src/zagg/stats/tdigest.py
+++ b/src/zagg/stats/tdigest.py
@@ -85,22 +85,14 @@ __all__ = [
 _DEFAULT_DELTA = 512
 
 
-def _k1_scale(q: float, delta: float) -> float:
-    """Dunning's k1 scale function: k(q) = delta * (arcsin(2q - 1)/pi + 1/2).
+def _k1_scale_array(q: np.ndarray, delta: float) -> np.ndarray:
+    """Dunning's k1 scale ``k(q) = delta * (arcsin(2q - 1)/pi + 1/2)``, vectorized.
 
-    Maps the cumulative rank fraction q ∈ [0, 1] onto [0, delta].  Its
+    Maps each cumulative rank fraction ``q ∈ [0, 1]`` onto ``[0, delta]``.  The
     derivative is largest at the tails (q → 0 or 1) and smallest at the median,
     so bounding each centroid to span ≤ 1 unit of k yields narrow, high-
     resolution centroids in the tails and wide centroids in the middle — the
-    defining t-digest property.  A digest holds ~delta centroids regardless of
-    the observation count, and is loss-free while the count stays ≤ delta.
-    """
-    qc = min(1.0, max(0.0, q))
-    return delta * (float(np.arcsin(2.0 * qc - 1.0)) / np.pi + 0.5)
-
-
-def _k1_scale_array(q: np.ndarray, delta: float) -> np.ndarray:
-    """Vectorized :func:`_k1_scale` over an array of rank fractions.
+    defining t-digest property.
 
     Evaluating the ``arcsin`` for every rank in **one** ufunc call is the whole
     point: the scalar per-observation ``float(np.arcsin(...))`` in the old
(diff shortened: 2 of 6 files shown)


src/zagg/processing/spill.py
```diff
@@ -465,6 +476,13 @@ def __init__(
         # Cross-block mergeable running state (only ever fed on block close).
         self._counts: dict[int, int] = {}
         self._digests: dict[str, dict[int, np.ndarray]] = {n: {} for n in self._digest_fields}
+        # k-way fold accumulator: for order-independent fields, each block's
+        # per-cell digest is stashed here and collapsed once at finalize
+        # (merge_tdigests_kway), rather than pairwise-folded per block. Each part
+        # is a saturated per-block digest and blocks are few, so this stays small.
```

**[memory, medium-high] The "blocks are few → stays small" assumption is weakest exactly where spill runs.** Unlike the pairwise fold — which keeps one running digest per cell (bounded at ~δ) — the k-way path holds *every* block's per-cell digest in `_digest_parts` until `_finalize_kway()` at first emit. Peak memory for a k-way field is therefore `O(B × C)` digests held simultaneously (B = closed blocks, C = occupied cells), vs `O(C)` for pairwise. But B is not small precisely in the case spill exists for: spill engages under memory pressure / large shards, and B = spill_bytes / block_bytes grows with exactly that data volume. So a pathological many-block shard accumulates the most parts right when the memory budget is tightest — the parts buffer can eat into the headroom the block-threshold formula reserved. Nothing bounds or caps B×C here. Is this quantified anywhere against the block-threshold reservation, or should `_finalize_kway` collapse incrementally (e.g. periodically fold accumulated parts per cell) so peak stays O(C)? At minimum the comment overstates the guarantee.

src/zagg/stats/tdigest.py
```diff
@@ -317,53 +363,121 @@ def merge_tdigests(
...
+        empty = np.empty((0, 2), dtype=np.float32)
+        return (empty, np.empty(0, dtype=np.uint64)) if located else empty
+    if len(arrs) == 1:
+        # A single contributor is already a valid digest; no re-compression, and
+        # copy the location channel so the return never aliases the caller.
+        out = arrs[0].astype(np.float32)
+        return (out, locs[0].copy()) if located else out
+
+    combined = np.concatenate(arrs, axis=0)
+    order = np.argsort(combined[:, 0], kind="stable")
```
espg: 🤖 *from Claude (review)*

**[correctness, medium] Order-independence is not strict when means tie across digests.** The claim (and #280's parallel-reduce invariant) is that the fold result is permutation-invariant. But `kind="stable"` breaks ties by *concatenation* position, which is the input digest order. If two centroids from different digests have the exact same mean, `[d1, d2]` and `[d2, d1]` place them in opposite relative order in `combined`. Consequences: (a) if a k1 boundary falls between the tied pair, one goes to the left output centroid and one to the right — with unequal weights the cumulative-weight boundaries then shift and the output arrays genuinely differ, not just at ULP; (b) even with no boundary split, the `reduceat` weighted-mean sum runs in a different order → ULP-level mean drift. `np.array_equal` (as `test_order_independent` uses) would catch either. `standard_normal` inputs never tie so the test passes, but discrete/quantized sources (integer counts, quality flags, repeated elevations) do tie. To make it truly order-independent, tie-break deterministically on something intrinsic (e.g. lexsort on `(mean, weight)`) rather than input position, or state the caveat. This matters because #280 wants to combine parts in nondeterministic parallel order.

src/zagg/processing/spill.py
```diff
@@ -613,12 +633,30 @@ def _fold_block(self, block: SpillBlock) -> None:
...
-                    )
+                    if pairwise:
+                        held = self._digests[name].get(cell)
+                        self._digests[name][cell] = (
+                            fresh if held is None else merge_tdigests(held, fresh, delta=delta)
+                        )
+                    else:
+                        self._digest_parts[name].setdefault(cell, []).append(fresh)
+
+    def _finalize_kway(self) -> None:
```
espg, line 654: 🤖 *from Claude (review)*

**[behavior-change, medium — confirm intended] The standard `build_tdigest` reducer is no longer byte-deterministic across the merge/spill decision.** Pre-#279, a `build_tdigest` field produced byte-identical digests whether it ran through merge mode or multi-block spill — the old `test_mergeable_multi_block_byte_identical_to_merge_mode` asserted exactly that, and it has been repurposed to the *pairwise* variant. Now the standard reducer folds k-way on the multi-block spill path but pairwise on the merge path, so identical input + identical config can emit different digest bytes depending on whether spill engaged. And that decision is environment-driven (memory tier / `block_bytes`), not user intent — so reproducibility now depends on the runtime, for the standard reducer. (Single-block spill still short-circuits through the len==1 k-way path and matches, so the divergence is multi-block-only.) This looks deliberate — k-way buys order-independence for #280, and `build_tdigest_pairwise` exists as the escape hatch — but please confirm nothing downstream pins standard-reducer merge==spell byte-identity (golden vectors, moczarr O3 parity fixtures, cached shard outputs). The removed guarantee is worth a one-line note in the PR body so a future reader knows it was dropped on purpose.

src/zagg/stats/tdigest.py
```diff
@@ -73,30 +73,89 @@
...
+
+    Shared core for both :func:`build_tdigest` (unit weights over sorted values)
+    and the merge paths (weighted sub-centroids). ``means`` must be ascending
+    and ``weights`` strictly positive, both 1-D float64.
+
+    Returns ``(out_means, out_weights, starts)``: the compressed centroid means
+    (weighted averages), their weights, and ``starts`` — each output centroid's
+    first sub-centroid index, which the located channel reduces over.
+
+    The centroid **boundaries and weights are bit-identical** to the old scalar
```
espg: 🤖 *from Claude (review)*

**[docs, low] "bit-identical to the old scalar loop" is very slightly overstated for boundaries too, not just means.** Two float re-associations can flip a boundary decision at the ULP:
1. The old loop tested `k_right - k_left <= 1.0` (a subtraction); `_compress` tests `searchsorted(k_right, k_left + 1.0, side="right")` (an addition). `a - b <= 1.0` and `a <= b + 1.0` are not identical in IEEE-754, so a sub-centroid whose k-span sits within an ULP of exactly 1.0 can land on the opposite side.
2. `total = float(cumw[-1])` (sequential `cumsum`) replaces the merge path's old `n_total = combined[:, 1].sum()` (pairwise summation) — different rounding of the denominator can shift a rank fraction by an ULP.

Both are astronomically rare and produce an equally-valid digest, so this is not a correctness bug — but the docstring's "boundaries and weights are bit-identical" claim would be more accurate as "identical except for float-ULP effects at k-span/denominator boundaries." Weights follow the boundaries, so they inherit the same caveat. (There is no test asserting equality against the removed scalar loop, so nothing breaks — the loss-free `n=500, δ=512` test still holds because it is nowhere near a 1.0-span boundary.)

src/zagg/stats/tdigest.py
espg: 🤖 *from Claude*

Fixed in 2c3cf78. Replaced the `np.argsort(combined[:, 0], kind="stable")` tie-break — which broke mean-ties by concatenation (input) order — with `np.lexsort((combined[:, 1], combined[:, 0]))` (primary key means, secondary weights). The ordering is now intrinsic to `(mean, weight)` and independent of the input digest order; sub-centroids with identical `(mean, weight)` are interchangeable so their residual order is immaterial. `combined_locs` already indexes by `order`, so it inherits the new ordering with no further change. The pairwise `merge_tdigests` sort is left as-is (order-dependent by design). Added `test_order_independent_with_tied_means` under `TestMergeTDigestsKway` that builds digests from `rng.integers(0, 5, ...)` (guaranteed cross-digest ties) and asserts byte-identical output across several permutations; the existing `test_order_independent` still passes.

src/zagg/stats/tdigest.py
espg: 🤖 *from Claude*

Softened in b8c9404. The docstring no longer claims boundaries/weights are "bit-identical" by construction. It now states that `_compress` reproduces the old greedy rule and was verified bit-identical across randomized trials, but notes the join comparison is re-associated (`searchsorted` on `k_left + 1.0` vs the loop`s `k_right - k_left <= 1.0`) and the rank denominator uses `cumw[-1]` vs a pairwise `.sum()`, so a boundary within an ULP of a 1.0 k-span could in principle flip — exact equality is not guaranteed by construction. Means keep their existing float-ULP caveat.

src/zagg/processing/spill.py

Leaving the design as-is; documented the cost in f2134ac. Holding every block`s per-cell part until finalize is intrinsic to the k-way guarantee: order-independence forbids incremental collapse, because folding subsets and then combining is a tree reduction of a non-associative op — that reintroduces exactly the order-dependence k-way exists to remove. So `_finalize_kway` collapsing periodically per cell would defeat the point.

The peak is bounded and quantified in the new comment: each part is a saturated ≤δ-centroid digest (~4 KB at δ=512, δ·8 bytes), so peak is ~B×C×(δ·8) over B closed blocks and C occupied cells — small relative to the multi-GB per-partition build peak the block-threshold formula already reserves. You are right that the comment previously overstated the guarantee ("blocks are few"); it now states the bound rather than asserting smallness. Tighter memory-budget accounting against the block-threshold reservation is owned by the #280 parallel-reduce work, where the parts buffer is combined in nondeterministic parallel order.

src/zagg/processing/spill.py
espg: 🤖 *from Claude*

Intended behavior — no code change. k-way was ratified as the standard fold precisely to buy order-independence for #280; `build_tdigest_pairwise` is the byte-identity escape hatch that preserves the pre-#279 merge==spill guarantee for anyone who needs it. This divergence is covered by two tests: `test_pairwise_multi_block_byte_identical_to_merge_mode` (pairwise reducer stays byte-identical merge vs spill) and `test_standard_kway_diverges_from_pairwise_multi_block` (asserts the standard k-way reducer diverges from pairwise on multi-block cells, counts unchanged).

Verified nothing downstream pins standard-reducer merge==spill byte-identity. Grepped `golden|parity|byte.identical|moczarr` across `tests/` and `src/`: the spill byte-identity tests (`test_full_reducer_matrix_byte_identical_to_pooled`, etc.) compare spill vs pooled within one mode, not merge-mode vs spill-mode; the only merge-vs-spill byte-identity assertion is the pairwise one above. No golden vectors or moczarr O3 parity fixtures live in this repo — moczarr is a separate reader that consumes the stored `(k,2)` digests and does not care how they were folded. Already flagged in the PR body`s "Questions for review".

PR #288, opened 2026-07-18
add process-global ambient S3 store cache (issue #287)

### LINKED ISSUE #287: Sidecar read path rebuilds an S3 client + credential chain per granule (missing the #244 store-cache)
🤖 *from Claude*

## Problem

On the sidecar read path, the worker **rebuilds a fresh S3 client + walks the full botocore credential chain once per granule**. On a 784-granule o9 shard this fired **675 `Found credentials in environment variables.` log lines in ~885 s** (one per granule that reached a read), each carrying ~300 ms of client/TLS + credential-resolution setup on the **critical read path**.

## Root cause (traced)

- `SidecarIndex._fetch_manifest` (`h5coro_hidefix/zagg_backend.py:326`) calls `open_object_store(self.store)` **on every manifest fetch** — no caching.
- `zagg/store.py::_s3_object_store` (`:202-210`) constructs a **brand-new `Boto3CredentialProvider()` per call**.
- `Boto3CredentialProvider.__init__` (`obstore/auth/boto3.py:90-101`) **eagerly** builds `boto3.Session()` and calls `session.get_credentials()`, which walks the default chain → `EnvProvider` logs the line (`botocore/credentials.py:1252`).

It's cached per-granule-id in `SidecarIndex._cache`, so it fires **~once per granule** (the chunk byte-range path is clean — h5coro's `S3Driver` is built once per granule with *explicit* creds, bypassing the chain, so chunks don't churn).

This is **the same symptom issue #244 already fixed on the raster path** with a module-global `_STORE_CACHE` (`raster.py:73-156`, which measured "~300 ms/client, 425 clients per invoke"). The sidecar manifest path was simply left out of that fix.

## Impact

At ~300 ms/client, 675 rebuilds ≈ **~200 s** of pure setup on the 85 °S pole shard's ~885 s wall — larger than the ~14 % (~130 s) that keeps that shard from fitting the 900 s Lambda ceiling (issue #148). Fixing it plausibly clears the 85–86 °S ring shards, and speeds every sidecar run fleet-wide.

## Fix

Memoize the obstore store per process in `zagg/store.py`, mirroring raster's `_STORE_CACHE` (issue #244). Scope it to the **ambient-credential, no-extra-kwargs** case — exactly the sidecar's `open_object_store(self.store)` call — so the explicit-credential (output writes, `runner.py:2995`), custom-endpoint, and read-only-retry (`temporal.py:636`) callers fall through to a fresh build unchanged. This keeps the fix in zagg (no `h5coro-hidefix` change needed): the sidecar calls `open_object_store` per fetch, but it now returns the cached store.

**Credential safety over a 900 s run** (per the #244 review): `Boto3CredentialProvider` auto-refreshes on each call (30-min ttl); on Lambda the execution-role creds are static per sandbox and don't rotate mid-invoke, so a sandbox-lifetime store cannot outlive its creds. Statically-supplied *explicit* credentials are deliberately **not** cached (they would freeze) — the excluded branch preserves today's behavior.

Refs #244 (the raster precedent), #148 (the pole tractability this unblocks), #279/#282 (the t-digest work whose gains this churn was masking).

espg:
Closes #287

## What this does

Memoizes the obstore `S3Store` per process on the **sidecar manifest-fetch hot path**, so the worker builds the S3 client + walks the botocore credential chain **once per sandbox** instead of **once per granule**.

`SidecarIndex._fetch_manifest` calls `open_object_store(self.store)` on every manifest fetch (once per granule). Without caching, each call constructed a fresh `Boto3CredentialProvider`, whose `__init__` eagerly builds `boto3.Session()` and walks the default credential chain (~300 ms of client/TLS + credential resolution, plus the `Found credentials in environment variables.` log line) — on the read critical path. On one 784-granule o9 shard that fired 675 rebuilds.

## Approach

Adds a process-global `_OBJECT_STORE_CACHE` (dict) guarded by `_OBJECT_STORE_LOCK` to `zagg/store.py`, mirroring the raster path's `_STORE_CACHE` (issue #244). In `open_object_store`, the cache is entered **only** for the ambient hot-path case:

```python
if credentials is None and endpoint_url is None and not kwargs:
    with _OBJECT_STORE_LOCK:
        store = _OBJECT_STORE_CACHE.get(path)
        if store is None:
            store = _s3_object_store(path)
            _OBJECT_STORE_CACHE[path] = store
    return store
```

Keyed on the path string. Single-flight build under the lock. This is exactly the sidecar's bare `open_object_store(self.store)` call. Every other caller — explicit `credentials` (output writes, `runner.py`), custom `endpoint_url`, or any `kwargs` (read-only retry config on the temporal path, coverage, hive) — falls through to a fresh `_s3_object_store(...)` build, byte-identical to today. The fix lives entirely in zagg; no `h5coro-hidefix` change is needed, since the backend keeps calling `open_object_store` per fetch but now gets the cached store back.

## Credential-lifetime safety

- The cached store carries a live `Boto3CredentialProvider` that **refreshes per call** (30-min ttl). On Lambda the execution-role credentials are static per sandbox and don't rotate mid-invoke, so a sandbox-lifetime store cannot outlive its credentials — the same reasoning ratified for #244.
- **Statically-supplied explicit tokens are deliberately not cached** (they would freeze on a warm worker): the excluded branch preserves today's fresh-build behavior for every explicit-credential caller.

## How tested

- New `TestObjectStoreCache` class in `tests/test_store.py`: ambient store built once and reused across N calls (asserts `_s3_object_store`/provider constructed once, all callers get the same instance); distinct paths → distinct stores; explicit-credentials / endpoint / kwargs each bypass the cache (fresh build per call); local paths not cached. An autouse `_clear_object_store_cache` fixture resets the process-global cache around every test so state can't leak between tests (the pre-existing ambient tests rely on this).
- `uv run pytest tests/test_store.py -v` → **32 passed** (6 new + all pre-existing, none regressed).
- `uv run pre-commit run --files src/zagg/store.py tests/test_store.py`: ruff check, ruff-format, codespell all pass. (mypy reports only pre-existing errors in unrelated modules — 85 on the baseline branch; zero in the two touched files. CI's lint job runs ruff only.)

## Questions for review

- **Impact is unverified until deploy.** The motivating hypothesis is that the ~675 client rebuilds on the 85–86°S pole shards collapse to 1/sandbox, dropping ~200 s of setup off the ~885 s wall and plausibly clearing the 900 s Lambda ceiling (issue #148). That needs a deploy + a re-run of the 85–86°S ring shards to confirm — the change here is the enabler, not the measurement.

### DIFF
--- a/src/zagg/store.py
+++ b/src/zagg/store.py
@@ -1,6 +1,7 @@
 """Store factory for opening Zarr stores from path strings."""
 
 import copy
+import threading
 from datetime import timedelta
 from pathlib import Path
 
@@ -98,6 +99,26 @@ def open_store(
     return LocalStore(Path(path).resolve(), read_only=read_only)
 
 
+# Ambient-credential object-store cache (issue #287): one obstore ``S3Store``
+# per ``s3://`` path per PROCESS, for the ambient (execution-role) hot path only.
+# The sidecar index backend (``h5coro_hidefix.zagg_backend.SidecarIndex``) calls
+# ``open_object_store(self.store)`` once per granule to fetch that granule's
+# manifest parquet; without this cache each call built a fresh
+# ``Boto3CredentialProvider`` whose ``__init__`` eagerly walks the botocore
+# credential chain (~300 ms of client/TLS + "Found credentials..." per granule),
+# on the read critical path — 675 rebuilds on one 784-granule o9 shard. Mirrors
+# the raster ``_STORE_CACHE`` (issue #244). Module lifetime == sandbox lifetime:
+# ``Boto3CredentialProvider`` refreshes per call (30-min ttl) and Lambda role
+# creds are static per sandbox, so a cached store cannot outlive its creds.
+# Scoped deliberately to the ``credentials is None and endpoint_url is None and
+# not kwargs`` case (the sidecar's exact call): explicit-credential output
+# writes, custom endpoints, and retry-config/anonymous callers fall through to a
+# fresh build, byte-identical to before — a statically-supplied token must NOT
+# be cached (it would freeze on a warm worker).
+_OBJECT_STORE_CACHE: dict = {}
+_OBJECT_STORE_LOCK = threading.Lock()
+
+
 def open_object_store(
     path: str,
     credentials: dict | None = None,
@@ -111,8 +132,20 @@ def open_object_store(
     objects -- e.g. the per-shard async result JSON a Lambda worker writes next
     to the output store for the orchestrator to poll. Path forms and credential
     handling match ``open_store``; a local directory is created if absent.
+
+    Ambient ``s3://`` stores (no explicit ``credentials``/``endpoint_url`` and no
+    extra ``kwargs``) are cached per process and reused across calls (issue #287)
+    -- this is the sidecar manifest-fetch hot path. Every other call builds a
+    fresh store, unchanged.
     """
     if path.startswith("s3://"):
+        if credentials is None and endpoint_url is None and not kwargs:
+            with _OBJECT_STORE_LOCK:
+                store = _OBJECT_STORE_CACHE.get(path)
+                if store is None:
+                    store = _s3_object_store(path)
+                    _OBJECT_STORE_CACHE[path] = store
+            return store
         return _s3_object_store(
             path,
             credentials=credentials,
--- a/tests/test_store.py
+++ b/tests/test_store.py
@@ -25,6 +25,17 @@ def mock_s3(monkeypatch):
     return s3_cls, prov_cls
 
 
+@pytest.fixture(autouse=True)
+def _clear_object_store_cache():
+    """Reset the process-global ambient store cache (issue #287) around every
+    test so one test's cached store can't leak into another's assertions."""
+    from zagg import store
+
+    store._OBJECT_STORE_CACHE.clear()
+    yield
+    store._OBJECT_STORE_CACHE.clear()
+
+
 class TestOpenStore:
     def test_local_absolute_path(self, tmp_path):
         store = open_store(str(tmp_path / "test.zarr"))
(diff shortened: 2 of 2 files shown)


espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial pass over the diff — **clean, nothing to fold.** What I checked:

- **Cache-key collisions:** key is the bare `path`. Safe because the cache is entered only under `credentials is None and endpoint_url is None and not kwargs`, which pins region to the `_s3_object_store` default (`us-west-2`) and rules out any `skip_signature`/`retry_config`/endpoint variation. So `path` fully determines the store — no two distinct store configs can collide on one key.
- **Thread-safety:** the whole get-or-build runs under `_OBJECT_STORE_LOCK` (single-flight), not a lossy double-checked lock — identical to raster's `_STORE_CACHE`/`_STORE_LOCK` in `processing/raster.py:134` (issue #244). No torn read, no duplicate build.
- **Autouse fixture × existing tests:** `_clear_object_store_cache` clears the process-global dict before and after each test, so the now-active cache can't leak across tests. The pre-existing `TestS3RetryConfig` cases call `open_store` (the Zarr factory — uncached) not `open_object_store`, and `test_default_retry_config_explicit_creds` passes explicit creds (bypass), so none touch the cache; the one ambient `open_object_store` case (`test_s3_returns_bare_store_with_ambient_creds`) is isolated by the fixture. 32/32 green confirms no regression.
- **Any ambient caller needing a *fresh* store:** none. Output writes pass explicit creds (bypass), the temporal read-only retry passes `retry_config` kwargs (bypass), coverage/hive pass kwargs — only the sidecar's bare `open_object_store(self.store)` hits the cached branch, and obstore `S3Store` is already shared across the `granule_workers` pool on the raster path, so sharing is established-safe.
- **mypy:** zero errors in `store.py`/`test_store.py`; the `_OBJECT_STORE_CACHE: dict` annotation matches the raster precedent.

Two non-blocking notes, both inherited from the ratified #244 pattern and non-issues at sidecar scale (a worker touches one store path): the cache never evicts (unbounded, but bounded in practice), and the global lock briefly serializes concurrent *first* builds of distinct paths. Neither warrants a change; flagging for the record.

Issue #284, opened 2026-07-18
Optionally auto-deploy the Lambda fleet on release (publish.yml deploy job)


espg:
🤖 *from Claude*

## Motivation

Today, cutting a release publishes artifacts but does **not** update the running Lambda fleet — a human still runs `stand_up.sh` (CloudFormation). Lambda serves an internal copy of the code snapshotted at the last deploy; it never re-reads S3 or auto-points at the newest zip. This issue is to decide whether (and how safely) to **auto-deploy the fleet as part of the release**, so a tagged release optionally lands on Lambda without a manual step.

## Current wiring (for reference)

`publish.yml` on a `*.*.*` tag: run tests → build wheel → publish PyPI → build dual-arch Lambda zips (`lambda-build-reusable.yml`) → create GitHub release + attach the four zips → *(conditional, only if the repo vars `LAMBDA_DIST_BUCKET` / `LAMBDA_RELEASE_ROLE_ARN` / `LAMBDA_AWS_REGION` are set)* upload the zips to the S3 dist bucket. **No CloudFormation, no `update-function-code` anywhere in it.**

- The change to deploy lives in the **function zip** (`build_function.sh` packages `src/zagg`); the **layer** (`build_layer.sh`, third-party deps) only changes when deps do.
- `template.yaml` functions reference **versioned** S3 keys (`FunctionS3Key`, `LayerS3Key`).
- `stand_up.sh` resolves the current version from `versions.json`, stages to a same-region bucket if needed, and does a CloudFormation stack **update** — CFN sees the changed key and rolls the function `Code`. This is the step that actually moves the fleet.

## Proposed approach

Add an **opt-in** deploy job to `publish.yml` (`needs: lambda-build`/`github-release`), gated so it can never fire by accident:

- **Gate**: existing `LAMBDA_RELEASE_ROLE_ARN` / `LAMBDA_DIST_BUCKET` / `LAMBDA_AWS_REGION` **plus** a new explicit toggle (e.g. `LAMBDA_AUTO_DEPLOY == 'true'`), so setups without it keep today's build-only behavior.
- **Auth**: OIDC `assume-role` into `LAMBDA_RELEASE_ROLE_ARN` (fleet account `742127912612`), least-privilege — CloudFormation + Lambda + S3 scoped to the stack(s). Not an admin role.
[...]

PR #289, opened 2026-07-18
Async result-object channel for the raster Lambda transport (issue #286)

### LINKED ISSUE #286: Raster benchmark leg fails on GitHub runners: synchronous invoke severed at ~4 min (needs #151-style async channel)
🤖 *from Claude*

The per-release **raster benchmark leg** (`0.35.0 Lambda Benchmark (full-AOI, per release)` → `raster` job) fails on GitHub-hosted runners, and it's a CI-transport limitation, not a code or Lambda bug.

## Symptom

The `Run raster benchmark (S2 2025, every shard over the AOI)` step exits 1 after ~4.5 min with, for every shard:

```
raster shard 4331422421 failed: Connection was closed before we received a valid
response from endpoint URL:
"https://lambda.us-west-2.amazonaws.com/2015-03-31/functions/process-shard/invocations".
```

(Example failing run: `actions/runs/29662915145`, job `raster`.)

## Root cause

`.github/scripts/run_raster_benchmark.py` invokes the base `process-shard` Lambda **synchronously** — one `RequestResponse` invoke per raster shard (the raster transport is "one synchronous `mode=process_raster` invoke per shard"; it deliberately does **not** use the issue #151 async result-object channel that the spatial path uses).

On 0.35.0 each of the 4 S2-2025 shards (85 granules each) takes **~486–529 s (~8–9 min)**. CloudWatch `/aws/lambda/process-shard` REPORTs for those invocations show them **completing** — Duration 485,869–528,860 ms, Max Memory ~0.9–1.1 GB, **no `Task timed out`, no `Runtime.OutOfMemory`**. The Lambda succeeds and writes its output.

But **GitHub-hosted runners sever a synchronous Lambda invoke's HTTP connection at ~4 min**, so `boto3` on the runner raises `Connection was closed…` and the job fails while the Lambda is still running (and then finishes fine). So the failure is purely the client giving up early — the same failure mode issue #151 fixed for the *spatial* benchmark path, which the raster path never received.

This is not new to 0.35.0 (raster doesn't use the t-digest touched by #279/#282); it fails on any release where a raster shard's runtime crosses ~4 min.

## Confirmation

Running the identical command locally (no 4-min runner sever) completes cleanly against the same `process-shard` 0.35.0 function — so the invoke works; only the CI runner's short synchronous-HTTP tolerance breaks.

## Fix options

1. **Give the raster path the #151 async channel** — async `Event` invoke + poll an S3 result object (poll deadline = function `Timeout` + margin), mirroring `_run_lambda`/`dispatch` on the spatial path. This is the durable fix and matches the ~900 s shard budget.
2. **Shrink raster shards** so each stays under ~4 min (finer parent_order for the S2 raster target, or fewer granules/shard).
3. **Run the raster leg on a runner without the ~4-min sever** (self-hosted).

Recommend (1) for parity with the spatial transport. Refs #151, #237/#272 (raster leg promotion), the failing run above.

espg:
Closes #286

## What

Gives the raster Lambda transport the issue-#151 async result-object channel, so a raster shard running longer than a GitHub runner's ~4 min synchronous-HTTP (NAT idle) tolerance no longer severs the dispatcher. The shard fan-out now **defaults to async** (`invocation="async"`): each shard fires `InvocationType="Event"` (fire-and-forget) and the dispatcher polls a per-shard result object the worker mirrors, exactly as the spatial path has since #151. `invocation="sync"` keeps the legacy synchronous `RequestResponse` transport (byte-identical events, no result object).

## Approach (mirrors the #151 spatial async path)

- **Worker mirror is already generic.** The handler-level `result_url` mirror (`lambda_handler` at `deployment/aws/lambda_handler.py:507-538`) covers **every** per-unit handler that falls through it — spatial `process`, temporal `process_event`, **and** raster `process_raster`. So a `process_raster` response is already mirrored to `result_url` whenever the event carries one; **no worker/handler code change was needed** for Phase 1 — only a test pinning that behavior. (This is why the diff touches no `deployment/aws/` code, per the golden rules.)
- **Dispatch** (`RasterStrategy._run_lambda_shards`): extracted the per-shard invoke into a module-level `_invoke_lambda_raster` — the raster twin of `_invoke_lambda_cell` — that supports both transports (sync `RequestResponse`, or async `Event` + poll via the shared `_poll_lambda_result` / `_result_fetcher`), reusing the 256 KB async-payload preflight and the transient-only retry rules. `_run_lambda_shards` builds the async result channel (`<store>.status/<run>/<shard_label>.json`, per-window suffix on hive) and reads the function timeout once for the poll deadline (`function_timeout + _ASYNC_POLL_MARGIN_S`).
- **Template write stays on the sync setup invoke (issue #264).** Only the **shard** fan-out goes async; the ping/setup (flat) and ping/async-setup/finalize (hive) lifecycle is unchanged, so the load-bearing template still lands before fan-out and an invoke-only dispatcher role (CI OIDC) is unaffected.
- **Plumbing:** threaded `invocation` through `RasterStrategy.run` → `_run_lambda_shards`. `agg` already forwarded `invocation`; `run_raster_benchmark.py` calls `agg` with defaults, so it goes async with no change.

## Phases

- [x] **Phase 1** — raster worker writes a result object (pinned by test; already generic at the handler level, no `deployment/aws/` change)
- [x] **Phase 2** — async Event dispatch + poll in the shard fan-out; template write stays on the sync setup invoke; 256 KB preflight + `invocation="sync"` fallback preserved
- [x] **Phase 3** — plumbing + tests (worker mirror, `_invoke_lambda_raster` sync/async unit, full-run poll + delayed-result, hive per-window naming, sync byte-identical)

## How tested

- `uv run pytest tests/test_raster_runner.py tests/test_lambda_raster_handler.py -q` → green.
- Full suite: **2405 passed, 35 skipped, 1 failed** — the single failure is `test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, an **environmental** failure (the deploy build script can't resolve `zarr>=3.1.5` platform wheels in this sandbox); it is pre-existing and unrelated to this change.
- New tests: `TestInvokeLambdaRaster` (sync reads response / no `result_url`; FunctionError + non-200 as shard errors, no retry; async Event carries `result_url` + polls; tolerates a delayed result; missing-at-deadline is an error with **no** re-invoke; oversized async payload raises before dispatch; transient invoke fault retries then polls) and `TestRasterLambdaAsyncBackend` (default async threads the result channel with `[ping RequestResponse, setup RequestResponse, process_raster Event]`; result object named by shard label; delayed result tolerated end-to-end; `invocation="sync"` omits `result_url`; hive windowed units get distinct `_<window>.json` objects). Existing `TestRasterLambdaBackend` / `TestRasterHiveLambdaBackend` were pinned `invocation="sync"` and retained as the byte-identical synchronous-transport regression suite.
- `ruff check` + `ruff format` clean on the changed files.

## Questions for review

1. **Template sequencing (issue #264).** I kept the setup/template emission on its own **synchronous** invoke and made only the shard fan-out async — the simplest correct version, since the flat template is load-bearing before fan-out. I did not adopt the hive-style async-setup + finalize-backstop for the flat setup. Flagging in case you'd prefer the shard channel and setup to share one lifecycle.
2. **Deploy + CI re-run is the real confirmation.** These tests mock the Lambda client + result box (no AWS); the actual >4 min-shard fix is confirmed only by a deploy and a benchmark re-run on a GitHub runner.
3. **Pre-existing lint/type debt (not touched, per §4):** `mypy` reports 3 pre-existing errors in `runner.py:1775/1780/1861` (`_raster_windowed_units` / `windows_intersecting`) and `codespell` flags `runner.py:1309` `statics` — both pre-date this PR and are outside the diff. Flagging rather than fixing.

### DIFF
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -615,16 +615,20 @@ class RasterStrategy:
     the template emission (and, later, the single-writer resize on append).
 
     Backends: ``"local"`` (thread pool, in-process workers) and ``"lambda"``
-    (one synchronous ``mode="process_raster"`` invoke per shard —
-    espg-confirmed scope on issue #218). The lambda cut is deliberately the
-    simple transport: synchronous invokes with transient-only retries; the
-    issue-151 async result-object channel and the preflight concurrency probe
-    are follow-ups (raster shards are seconds of COG windows, well inside NAT
-    idle limits). Either way the runner owns the global time index before
-    fan-out (the single-writer append design); the template write is
-    orchestrator-side on the local backend but rides the sync setup invoke on
-    lambda (issue #264 — the dispatcher may hold an invoke-only role with no
-    S3 write access, e.g. the CI OIDC benchmark role).
+    (one ``mode="process_raster"`` invoke per shard — espg-confirmed scope on
+    issue #218). The lambda shard fan-out defaults to the issue-151 async
+    result-object channel (``invocation="async"``, issue #286): each shard is
+    a fire-and-forget ``InvocationType="Event"`` invoke and the dispatcher
+    polls a per-shard result object the worker mirrors, so a shard running
+    longer than a GitHub runner's ~4 min synchronous-HTTP (NAT idle) tolerance
+    no longer severs the dispatcher. ``invocation="sync"`` keeps the legacy
+    synchronous ``RequestResponse`` transport (byte-identical events, no
+    result object). The preflight concurrency probe stays a follow-up. Either
+    way the runner owns the global time index before fan-out (the
+    single-writer append design); the template write is orchestrator-side on
+    the local backend but rides the SYNC setup invoke on lambda even under
+    async shard dispatch (issue #264 — the dispatcher may hold an invoke-only
+    role with no S3 write access, e.g. the CI OIDC benchmark role).
 
     Summary schema note: ``total_obs`` is shared with the spatial/temporal
     strategies so a caller sees one summary shape across pipeline kinds. On the
@@ -661,6 +665,7 @@ class RasterStrategy:
         output_credentials,
         output_endpoint_url,
         profile=False,
+        invocation="async",
         **_ignored,
     ):
         from zagg.processing.raster import (
@@ -738,6 +743,7 @@ class RasterStrategy:
                 store_layout=store_layout,
                 dataset={"short_name": md.get("short_name"), "version": md.get("version")},
                 profile=profile,
+                invocation=invocation,
             )
 
         # Local backend: template emission stays orchestrator-owned (the
@@ -940,8 +946,23 @@ class RasterStrategy:
         store_layout="flat",
         dataset=None,
         profile=False,
+        invocation="async",
     ):
-        """Fan shards out, one synchronous ``mode="process_raster"`` invoke each.
+        """Fan shards out, one ``mode="process_raster"`` invoke each.
+
+        The shard fan-out defaults to the issue-151 async result-object
+        channel (``invocation="async"``, issue #286): each shard is a
+        fire-and-forget ``InvocationType="Event"`` invoke and the dispatcher
+        polls ``<store>.status/<run>/<shard_label>.json`` (per-window suffix on
+        hive) for the worker's mirrored envelope, with the poll deadline keyed
+        to the function timeout + margin. No synchronous connection sits open
+        while a shard runs, so a shard longer than a GitHub runner's ~4 min NAT
+        idle tolerance survives (the #286 failure mode). ``invocation="sync"``
+        keeps the legacy synchronous ``RequestResponse`` invoke, byte-identical
+        to the pre-#286 transport (no ``result_url`` on the event). The
+        ping/setup lifecycle stays SYNCHRONOUS under either mode (only the
+        shard fan-out goes async) so the load-bearing template write still
+        lands before fan-out (issue #264, below).
 
         Before fan-out the lifecycle is ping → setup. Flat (issue #264): the
         lightweight ``mode="ping"`` fails fast on a pre-#252 deployment with
@@ -965,6 +986,8 @@ class RasterStrategy:
         import boto3
         from botocore.config import Config
 
+        if invocation not in ("async", "sync"):
+            raise ValueError(f"Unknown invocation: {invocation!r} (expected 'async' or 'sync')")
         function_name = _resolve_function_name(config, function_name)
         max_workers = min(max_workers or 64, len(cells)) or 1
         client = boto3.client(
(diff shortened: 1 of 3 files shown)


espg:
🤖 *from Claude*

**Adversarial self-review (fresh context, Opus): clean.** No inline threads — every called-out risk area checked out:

- **Sync/local byte-identity** — the `_invoke_lambda_raster` sync branch is byte-for-byte the old inline `_one` loop (same `RequestResponse` invoke, same `FunctionError`/non-200/success mapping); the async `event = {**event, "result_url": ...}` rebuild is a copy so the caller's event never gains the key; setup reads + `result_prefix` are gated behind `invocation == "async"`, so sync does zero extra AWS calls; the local backend never touches `invocation`.
- **Poll timeout** — `function_timeout + _ASYNC_POLL_MARGIN_S` (990s), deadline from first-attempt wall_start, function timeout read on the right client with a 900s fallback — matches the spatial default.
- **Result-object naming** — `<label>.json` / `<label>_<window>.json` under a per-run `uuid4().hex` prefix, identical to the spatial `_cell_work`; windowed triples only exist on hive, so no flat collision.
- **Template write** — stays a single synchronous setup invoke before fan-out under both flat and hive (issue #264), unaffected by `invocation`.
- **Lost/duplicate delivery** — a successful Event invoke polls and returns (no re-invoke of a running shard); fetch faults are swallowed as misses; a deadline miss / FunctionError is a shard error, not a re-dispatch; only client-side transient invoke faults retry — the established `_invoke_lambda_cell` precedent, and the write is idempotent.

**One folded finding (parity nit, 3056e3e):** the async non-200 mapping guarded `status is not None`, so a landed envelope *lacking* a `statusCode` key would read as success in async but `"status None"` in sync. Unreachable with the real worker (it always emits `statusCode`), but I dropped the clause so async matches the sync branch exactly, and added `test_async_non_200_result_surfaces_error_like_sync`.

Left as a draft; the template-sequencing choice under "Questions for review" is the one open design point for you.

espg:
🤖 *from Claude*

Added a small related commit (`656023b`) at @espg's direction: the raster release benchmark now dispatches to the **`-disk` production variant** (`process-shard-4096-disk`) instead of the bare `process-shard`, so the raster leg runs on the same function tdigest/prod use.

- `run_raster_benchmark.run_target` now mirrors `run_benchmark.run_target`'s worker-suffix resolution: a target's `worker: {memory, extra_disk}` block resolves `base + -<memory>[-disk]`. Absent → the base name, unchanged.
- `targets_raster_neon.json`'s target gains `worker: {memory: 4096, extra_disk: true}`.
- Dry-run confirms: `-> process-shard-4096-disk @ us-west-2`.

This is deploy-alignment (issue #284: `LAMBDA_PROD_FUNCTION_NAME` was just pointed at `process-shard-4096-disk`), folded here since it's the same `run_raster_benchmark.py`. Separable on request.

PR #295, opened 2026-07-19
ShardMap.reproject: derive a shard map at another HEALPix order (issue #294)

### LINKED ISSUE #294: ShardMap.reproject: derive a shard map at another HEALPix order (coarsen = regroup, refine = scoped re-intersect) instead of rebuilding from catalog
🤖 *from Claude*

## Motivation

The order sweeps (CONUS #202, 88S #148) currently rebuild the **full** shard map from the 555k-granule ATL03 catalog **once per HEALPix order** — each is a minutes-long `ShardMap.build` (bbox/temporal prefilter + per-granule `morton_coverage` intersection). But HEALPix nesting means a shard map at one order is derivable from another without touching the catalog again: an o6 shard is exactly the morton parent of four o7 shards. We're using this by hand right now (deriving the CONUS o6 25-shard set by coarsening the o7 map). This issue is to make it a first-class, both-directions utility.

## Proposed API

`ShardMap.reproject(target_grid)` (or `reproject_shardmap(sm, target_grid)`) → a new `ShardMap` at `target_grid`'s `parent_order`, carrying the same granule assignments, with a refreshed `grid_signature` and `metadata` (record the source order + method).

## Two directions — asymmetric by nature

**Coarsen (finer → coarser, e.g. o7 → o6): pure regroup, exact, no geometry.**
- `parent = clip2order(target_order, shard_key)` for each shard (the same `mortie.clip2order` primitive `HealpixGrid.shards_of` uses).
- Group source shards by `parent`; each coarse shard's granule list = **union of its children's granule lists, deduplicated by granule id** (a granule in multiple children counts once in the parent).
- Exact because footprint→cell assignment nests: a granule intersects a coarse cell **iff** it intersects one of its finer children. O(#shards); no catalog, no `morton_coverage`.

**Refine (coarser → finer, e.g. o7 → o8): scoped re-intersection, still cheap.**
- Cannot be a pure regroup — the coarse map doesn't record which child cell each granule fell in. But it *can* avoid the full-catalog rebuild: for each coarse shard, re-run the footprint→cell intersection (`morton_coverage`/`ShardMap`'s `_intersect_*`) at the target order **over only that shard's granules** (a few hundred, vs 555k). Union across shards.
- Reproduces the direct `ShardMap.build` at the finer order exactly (same intersection, just scoped to the pre-filtered granule set), at a fraction of the cost.

## Correctness / tests
- **Coarsen is exact:** `reproject(o8→o7)` must equal `ShardMap.build(...o7)` — same shard_keys, same per-shard granule sets. Test on the committed CONUS o8/o9 maps.
- **Refine reproduces the build:** `reproject(o7→o8)` equals the direct o8 build (shard_keys + granule sets).
- **Round-trip:** `coarsen(o8→o7)` then `refine(o7→o8)` returns the original o8 map (granule sets identical; keys identical).
- Guard: `target_order` within `[0, child_order]`; reject a target coarser/finer than the grid supports; carry `grid.signature()` correctly so the runner's spatial-signature check still passes.

## Reference implementation (coarsen direction, live now)
The CONUS o7→o6 derivation in the current order-sweep work is the coarsen-direction reference — `clip2order` + union-dedup. The refine direction is the new piece.

Refs #202 (CONUS order sweep — the immediate consumer), #148 (88S order sweep), the `ShardMap`/`grids.healpix` (`clip2order`/`shards_of`) machinery.

espg:
Closes #294.

Derives a shard map at another HEALPix order from an existing one, instead of rebuilding from the 555k-granule catalog per order (the CONUS #202 / 88S #148 order sweeps).

## What
`ShardMap.reproject(target_grid, catalog=None) -> ShardMap`, both directions relative to the source `parent_order`:
- **Coarsen** (target < source): pure regroup — `mortie.clip2order(target, key)` per shard, group by parent, dedup-union children's granule lists by id. Exact by HEALPix nesting; O(#shards), no geometry.
- **Refine** (target > source): scoped re-intersection via `_intersect_mortie` restricted to each source shard's own descendants — needs `catalog=` for footprints (raises a clear `ValueError` naming the requirement if omitted). Reproduces the direct `ShardMap.build` at the finer order without touching the full catalog.
- Same order: no-op copy. Guards reject mismatched `child_order` / non-HEALPix signatures; refreshes `grid_signature`.

## Tested (`tests/test_shardmap.py::TestReproject`, 8 cases)
coarsen-matches-direct-build, dedup, refine-reproduces-build, refine-without-catalog raises, round-trip, no-op, mismatched-child_order, non-HEALPix rejection. `pytest tests/test_shardmap.py` → 70 passed, 1 skipped (pre-existing spherely skip); ruff clean.

## Real-world validation
Built the CONUS o7 map (3,276 shards) and derived o6 by `reproject` coarsen (883 shards; per-shard `max(children) <= union <= sum(children)` held for all 883). Reprojected o7→o8 (refine) matched the committed direct o8 build on **95/100** shards; the 5 diffs are legitimate CONUS-polygon-boundary overhang (some o7 parents straddle the edge; a few o8 children fall just outside the region-restricted direct build's candidate set — `reproject` has no region arg, by design). Same overhang class as issue #101.

## Questions for review
- Refine's region-independence: `reproject` intentionally has no polygon/region argument, so refining a boundary-straddling shard can keep a child the region-restricted direct build dropped. Acceptable (it never drops a real intersection), but flagging the semantics.

### DIFF
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -608,9 +608,12 @@ class ShardMap:
         re-intersected at ``target_order`` (the same ``morton_coverage_moc``
         machinery :meth:`build` uses), restricted to that shard's own
         descendant cells (``generate_morton_children(shard_key,
-        target_order)``) so the result reproduces the direct
-        :meth:`build` at the finer order exactly, at a fraction of the cost
-        (only this shard's granules, not the whole catalog). Refine always
+        target_order)``). In the interior this reproduces the direct
+        :meth:`build` at the finer order; at a region/AOI boundary it may
+        **over-include** child shards a region-restricted direct build would
+        drop (the #101 whole-shard overhang class -- reproject applies no
+        region clip), but it never drops a real intersection. Costs only this
+        shard's granules, not the whole catalog. Refine always
         re-intersects via the mortie MOC path regardless of the source's
         ``backend`` (a spherely-built source is not reproduced by it), so the
         derived map records ``backend="mortie"``.
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -722,6 +722,9 @@ class ShardMap:
         }
         meta["total_shards"] = len(new_keys)
         meta["total_pairs"] = sum(len(g) for g in new_granules)
+        # The dropped per-shard AOI mask (aoi_mask=None below) must not still be
+        # advertised in the derived map's metadata.
+        meta.pop("aoi_mask", None)
         return ShardMap(target_grid.spatial_signature(), new_keys, new_granules, meta, None)
 
     def to_json(self, path: str) -> None:
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -725,6 +725,16 @@ class ShardMap:
         # The dropped per-shard AOI mask (aoi_mask=None below) must not still be
         # advertised in the derived map's metadata.
         meta.pop("aoi_mask", None)
+        # This reproject ran no timed build, so the source build's timing must
+        # not describe the derived map.
+        meta.pop("build_wall_s", None)
+        if method == "refine":
+            meta["mortie_order"] = mortie_order
+        else:
+            # Coarsen is a pure regroup: no geometry backend and no order-based
+            # build ran, so the source's ``backend``/``mortie_order`` don't apply.
+            meta.pop("backend", None)
+            meta.pop("mortie_order", None)
         return ShardMap(target_grid.spatial_signature(), new_keys, new_granules, meta, None)
 
     def to_json(self, path: str) -> None:
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -610,7 +610,10 @@ class ShardMap:
         descendant cells (``generate_morton_children(shard_key,
         target_order)``) so the result reproduces the direct
         :meth:`build` at the finer order exactly, at a fraction of the cost
-        (only this shard's granules, not the whole catalog).
+        (only this shard's granules, not the whole catalog). Refine always
+        re-intersects via the mortie MOC path regardless of the source's
+        ``backend`` (a spherely-built source is not reproduced by it), so the
+        derived map records ``backend="mortie"``.
         """
         from mortie import clip2order, generate_morton_children
 
@@ -729,6 +732,10 @@ class ShardMap:
         # not describe the derived map.
         meta.pop("build_wall_s", None)
         if method == "refine":
+            # Refine re-intersects via the mortie MOC path (``_intersect_mortie``)
+            # regardless of the source's backend -- a spherely-built (exact-S2)
+            # source is not reproduced by it -- so record what actually ran.
+            meta["backend"] = "mortie"
             meta["mortie_order"] = mortie_order
         else:
             # Coarsen is a pure regroup: no geometry backend and no order-based
--- a/src/zagg/catalog/shardmap.py
+++ b/src/zagg/catalog/shardmap.py
@@ -562,6 +562,168 @@ class ShardMap:
             meta["aoi_mask"] = True
         return cls(grid.spatial_signature(), shard_keys, granules, meta, aoi_mask)
 
+    def reproject(self, target_grid, catalog=None) -> "ShardMap":
+        """Derive a ShardMap at ``target_grid``'s ``parent_order`` (issue #294).
+
+        HEALPix nesting means a shard map at one order is derivable from
+        another **without touching the source catalog again** in the coarsen
+        direction, and with only a scoped (per-shard) re-intersection in the
+        refine direction -- either is far cheaper than a full
+        :meth:`build` over the whole catalog.
+
+        Parameters
+        ----------
+        target_grid : HealpixGrid
+            Grid to reproject onto. Must share ``child_order`` (the leaf
+            resolution) with the source grid -- reprojecting across different
+            DGGS resolutions isn't meaningful, only the shard (dispatch) order
+            changes.
+        catalog : Catalog, optional
+            Required only when refining (``target_grid.parent_order >`` this
+            map's ``parent_order``): the shard map itself stores only
+            ``{"id", "s3", "https"}`` per granule, not footprint geometry, so
+            recovering which finer cell each granule falls in needs the
+            granule records (``catalog.granule_records()``) back.
+
+        Returns
+        -------
+        ShardMap
+            New map at ``target_grid.parent_order``. ``metadata`` records the
+            source order and ``reproject: {method: "coarsen"|"refine"|"noop"}``.
+
+        Notes
+        -----
+        **Coarsen** (``target_order < source_order``): pure regroup, exact,
+        no geometry. Each source shard's key coarsens via
+        ``mortie.clip2order(target_order, shard_key)``; shards sharing a
+        coarse parent are grouped and their granule lists unioned, deduplicated
+        by granule ``id`` (a granule spanning multiple children counts once in
+        the parent). Exact because footprint-to-cell assignment nests: a
+        granule intersects a coarse cell iff it intersects one of its finer
+        children.
+
+        **Refine** (``target_order > source_order``): cannot be a pure
+        regroup -- the coarse map never recorded which child cell a granule
+        fell in. Instead, for each source shard, its own granules are
+        re-intersected at ``target_order`` (the same ``morton_coverage_moc``
+        machinery :meth:`build` uses), restricted to that shard's own
+        descendant cells (``generate_morton_children(shard_key,
+        target_order)``) so the result reproduces the direct
+        :meth:`build` at the finer order exactly, at a fraction of the cost
+        (only this shard's granules, not the whole catalog).
+        """
+        from mortie import clip2order, generate_morton_children
+
+        target_order = getattr(target_grid, "parent_order", None)
+        target_child_order = getattr(target_grid, "child_order", None)
+        if target_order is None or target_child_order is None:
+            raise ValueError(
+                "reproject: target_grid must be a HEALPix grid (parent_order/child_order)"
+            )
+        source_order = self.grid_signature.get("parent_order")
+        source_child_order = self.grid_signature.get("child_order")
+        if source_order is None or self.grid_signature.get("type") != "healpix":
+            raise ValueError("reproject: source ShardMap must be a HEALPix grid_signature")
+        if source_child_order != target_child_order:
+            raise ValueError(
+                f"reproject: child_order must match (source={source_child_order}, "
+                f"target={target_child_order}) -- reproject only changes the shard "
+                "(parent_order), not the leaf DGGS resolution"
+            )
+        if not (0 <= target_order <= target_child_order):
+            raise ValueError(
+                f"reproject: target_order {target_order} outside [0, child_order="
+                f"{target_child_order}]"
+            )
+
+        if target_order == source_order:
+            meta = dict(self.metadata or {})
+            meta["reproject"] = {
+                "source_parent_order": int(source_order),
+                "target_parent_order": int(target_order),
+                "method": "noop",
+            }
+            return ShardMap(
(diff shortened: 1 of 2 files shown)


src/zagg/catalog/shardmap.py
```diff
@@ -562,6 +562,168 @@ def build(
...
+
+        meta = dict(self.metadata or {})
+        meta["reproject"] = {
+            "source_parent_order": int(source_order),
+            "target_parent_order": int(target_order),
+            "method": method,
+        }
+        meta["total_shards"] = len(new_keys)
+        meta["total_pairs"] = sum(len(g) for g in new_granules)
+        return ShardMap(target_grid.spatial_signature(), new_keys, new_granules, meta, None)
```
espg, line 748: 🤖 *from Claude (review)* — **MEDIUM: reproject silently drops `aoi_mask` while metadata still advertises it → a lying manifest.**

Coarsen/refine construct the new map with `aoi_mask=None` (correct — the source per-shard mask is parallel to the old `shard_keys` and would be wrong-length/misaligned against the new shard set). But `meta = dict(self.metadata or {})` on line 717 carries the source's `"aoi_mask": True` forward unchanged. Verified end-to-end: after reprojecting an AOI-masked source, `metadata["aoi_mask"] == True` while the `aoi_mask` field is `None`, so `to_json`/`to_parquet` write **no** `aoi_mask` payload yet the metadata claims one is present. A downstream reader trusting the metadata flag is misled.

Separately, the semantic drop itself is silent: reprojecting an AOI-masked map yields an **unmasked** map (issue #101 whole-shard overhang returns) with no warning. The mask depends only on (grid, AOI), so it is recomputable — but at minimum, when dropping the field, pop `meta["aoi_mask"]` so the metadata doesn't lie.

src/zagg/catalog/shardmap.py
```diff
@@ -562,6 +562,168 @@ def build(
...
+                    bucket = new_granules_map.setdefault(int(k), {})
+                    for i in idxs:
+                        entry = _granule_entry(sub_records[i])
+                        bucket[entry["id"]] = entry
+
+            new_keys = sorted(new_granules_map)
+            new_granules = [list(new_granules_map[k].values()) for k in new_keys]
+            method = "refine"
+
+        meta = dict(self.metadata or {})
```
espg, line 723: 🤖 *from Claude (review)* — **MEDIUM: stale `build_wall_s` / `mortie_order` / `backend` carried into the derived map's metadata.**

`meta = dict(self.metadata or {})` copies the *source build's* provenance, and only `total_shards`/`total_pairs` are refreshed (lines 723–724). Verified on an o12→o11 coarsen: the result carries `build_wall_s: 0.0` (the source build's timing — this reproject ran no timed build), `mortie_order: 12` (source order; the map is now o11), and `backend: mortie` (for coarsen **no geometry backend ran at all**). These now misdescribe the derived map. `build` recomputes all three; reproject should refresh or drop them (e.g. drop `build_wall_s`/`mortie_order`, and for refine set `backend`/`mortie_order` to what refine actually used). Contrast the task's own note that `total_pairs` etc. must not be left to lie — `total_pairs` is fixed, these three aren't.

src/zagg/catalog/shardmap.py
```diff
@@ -562,6 +562,168 @@ def build(
...
+                    raise ValueError(
+                        f"reproject: refine needs footprints for all granules in the source "
+                        f"map, but the catalog is missing {len(missing)} of them (e.g. "
+                        f"{missing[:5]}{extra})"
+                    )
+                sub_records = [records_by_id[gid] for gid in ids]
+                descendants = set(
+                    int(s) for s in generate_morton_children(int(shard_key), target_order)
+                )
+                shard_to_idx = _intersect_mortie(
```
espg, line 705: 🤖 *from Claude (review)* — **LOW–MEDIUM: refine always uses the mortie MOC backend, regardless of how the source was built.**

The source guard only requires a HEALPix `grid_signature`; a HEALPix map can legitimately be built with `backend="spherely"` (exact S2). Refine unconditionally re-intersects with `_intersect_mortie` — the ~0.01% polar-omission MOC path (module docstring, lines 22–23) — so for a spherely-built source it does **not** reproduce that source's build at the finer order, and (compounding the metadata finding above) `meta["backend"]` still reads `"spherely"`. The "reproduces the direct build exactly" claim only holds when the source was itself mortie-built. Either restrict refine to mortie-built sources (guard on `meta["backend"]`) or pick the backend from the source's recorded backend.

src/zagg/catalog/shardmap.py
```diff
@@ -562,6 +562,168 @@ def build(
...
+        granule intersects a coarse cell iff it intersects one of its finer
+        children.
+
+        **Refine** (``target_order > source_order``): cannot be a pure
+        regroup -- the coarse map never recorded which child cell a granule
+        fell in. Instead, for each source shard, its own granules are
+        re-intersected at ``target_order`` (the same ``morton_coverage_moc``
+        machinery :meth:`build` uses), restricted to that shard's own
+        descendant cells (``generate_morton_children(shard_key,
+        target_order)``) so the result reproduces the direct
```
espg: 🤖 *from Claude (review)* — **LOW: docstring overstates "reproduces the direct `build` at the finer order exactly."**

Refine scopes to each source shard's *full* descendant set (`generate_morton_children(shard_key, target_order)`) with no region/polygon clip, so at region boundaries it keeps child shards the region-restricted direct `build` drops — exactly the 95/100 case in the PR body and the #101 overhang class. The claim is exact only in the interior (or when no `region`/AOI restricted the direct build). The behavior is arguably fine (never drops a real intersection), but the docstring should match the PR's own "Questions for review" caveat rather than say "exactly" unqualified.

tests/test_shardmap.py
```diff
@@ -928,6 +928,98 @@ def test_beams_on_missing_collection_metadata_raises(self, grid, fake_spherely):
...
+            "method": "coarsen",
+        }
+
+    def test_coarsen_dedups_granule_spanning_multiple_children(self, fine_grid, coarse_grid):
+        # A granule wide enough to land in >=2 fine shards under the same
+        # coarse parent must count once in the coarsened granule list.
+        cat = _catalog([_item("Gwide", -76.60, -76.52)])
+        sm_fine = ShardMap.build(cat, fine_grid, backend="mortie")
+        gs_fine = _granule_shards(sm_fine)
+        assert len(gs_fine["Gwide"]) >= 1  # sanity: built onto >=1 fine shard
```
espg: 🤖 *from Claude (review)* — **LOW: `test_coarsen_dedups_...` doesn't actually exercise the dedup path.**

The test comment says `Gwide` is "wide enough to land in >=2 fine shards under the same coarse parent," but the assertions only check `len(gs_fine["Gwide"]) >= 1` (line 965) and "no duplicate ids within a shard list" (lines 971–973). The dedup branch (`seen[g["id"]] = g` unioning across children) only runs when the granule appears in **≥2 fine children sharing one coarse parent**; if `Gwide` lands in a single fine shard, dedup never triggers and the test still passes — so it doesn't prove dedup works. Strengthen by asserting the scenario is real, e.g. that `Gwide` appears in ≥2 fine shards that clip to the same coarse parent (or that its total fine appearances exceed its coarse appearances).

src/zagg/catalog/shardmap.py
```diff
@@ -562,6 +562,168 @@ def build(
...
+            for i, p in enumerate(parents.tolist()):
+                groups.setdefault(int(p), []).append(i)
+
+            new_keys = sorted(groups)
+            new_granules = []
+            for k in new_keys:
+                seen: dict = {}
+                for i in groups[k]:
+                    for g in self.granules[i]:
+                        seen[g["id"]] = g
```
espg: 🤖 *from Claude (review)* — **LOW: coarsen (and the no-op branch) alias the source's granule dicts by reference.**

`seen[g["id"]] = g` (here) and `[list(g) for g in self.granules]` in the no-op branch (line 649) reuse the *same* granule dict objects in the derived map; a mutation to a granule dict in either map would leak into the other. Refine builds fresh entries via `_granule_entry`, and `build` always produces fresh dicts, so this is an inconsistency. Harmless if maps are treated as immutable (they generally are), but worth a fresh-copy for consistency, or a note that the derived map shares granule payloads with its source.

espg (commented):
🤖 *from Claude (review)*

Adversarial review of `ShardMap.reproject` (issue #294). **No blocking correctness bug** — I checked the core claim adversarially and it holds: coarsen is exact by HEALPix nesting (a footprint intersects a coarse cell iff it intersects a child), and refine's per-shard scoping is sound because a fine child cell is a subset of its coarse parent, so any granule reaching a child of coarse cell Y was necessarily assigned to Y in the source map — no fine assignment can be missed. The finer-MOC ⊆ coarser-MOC region property means refine also never under-assigns; the 95/100 boundary diffs are over-inclusion (reproject has no region clip), matching the #101 overhang class, consistent with "never drops a real intersection."

The findings below are all **metadata hygiene / test-strength / docstring accuracy**, none blocking. Ranked by severity in the inline comments.

src/zagg/catalog/shardmap.py
espg: 🤖 *from Claude*

Fixed in 8c717e9. Coarsen/refine now `meta.pop("aoi_mask", None)` before constructing the derived map (which carries `aoi_mask=None`), so the metadata no longer advertises a mask the map does not carry. The no-op branch is unchanged — it preserves the source `aoi_mask` and the flag correctly stays. Left the semantic recompute-on-reproject out of scope for this fold (mask dropping is intended behavior; only the lying flag was the bug).

src/zagg/catalog/shardmap.py
espg: 🤖 *from Claude*

Fixed in dbbe882 (order/timing) and 94b6250 (refine backend). The derived metadata now: drops `build_wall_s` (no timed build ran); on refine sets `mortie_order` to the target-resolved order and `backend="mortie"`; on coarsen drops both `mortie_order` and `backend` (pure regroup — no order-based build and no geometry backend ran). The honest fields (`total_shards`/`total_pairs`, footprint, collection, reproject) stay.

src/zagg/catalog/shardmap.py
espg: 🤖 *from Claude*

Fixed in 94b6250 — took the smaller correct option: refine still uses the mortie MOC path unconditionally, but now stamps `backend="mortie"` on the refined result (in dbbe882/94b6250) so the metadata reflects what actually ran rather than the source `backend` (e.g. a spherely source no longer leaves `backend="spherely"` on a mortie-produced map). Documented it in the docstring too (1019895): refine always re-intersects via the mortie MOC path regardless of source backend, and a spherely-built source is not reproduced by it.

src/zagg/catalog/shardmap.py

Fixed in 1019895. Docstring no longer says "exactly" unqualified — it now states refine reproduces the direct build **in the interior**, and at a region/AOI boundary may **over-include** child shards a region-restricted direct build would drop (the #101 whole-shard overhang class — reproject applies no region clip), but never drops a real intersection. Matches the PR body caveat.

tests/test_shardmap.py
espg: 🤖 *from Claude*

Fixed in bd45e62. The test now proves the dedup scenario is real: it asserts `Gwide` spans ≥2 fine shards, then uses `clip2order(11, ...)` to assert ≥2 of those fine shards coarsen to a common parent (so the union-across-children branch actually runs). After reproject it asserts `Gwide`'s coarse shard count is strictly fewer than its fine count and that each coarsened shard list has no duplicate ids — i.e. the duplicates collapsed to exactly one per parent.

src/zagg/catalog/shardmap.py

Fixed in dd9fcb5. Coarsen now does `seen[g["id"]] = _granule_entry(g)` and the no-op branch does `[[_granule_entry(g) for g in shard] for shard in self.granules]`, so both build fresh dicts via `_granule_entry` — matching refine and `build`. A reprojected map no longer shares granule dict objects with its source, so a mutation in one can't leak into the other.

espg:
reproject is the wrong name here... should probably be reshard

Issue #299, opened 2026-07-19
Template-hash leaf naming: multi-product hive store (breaking, pre-1.0)


espg:
🤖 *from Claude*

Sub-issue of #296 — slice (3): rename leaf stores from the morton id to a hash of the agg template, turning the hive tree into a multi-product store. **Breaking store change; pre-1.0 timing per the #262 rationale.**

## Context

The morton id is currently encoded twice: as the digit-tree path components *and* as the leaf basename — `check_node_invariant` in `src/zagg/hive.py` literally asserts `full_id == head + "".join(digits)`. The basename is redundant, so it's free for product identity: leaves become `{template_hash}.zarr`, several aggregations colocate as siblings under one morton prefix, and discovery ("what aggs exist here?") is a prefix listing. This is the nix-store provenance pattern: the template says *what*, the shardmap says *where*, and the two are independent (no config references a shardmap — verified).

## Decisions settled on #296

- **Name = canonicalized template hash.** Parse the YAML, serialize to sorted-key JSON, sha256, truncate. Never hash raw bytes — comment/whitespace/key-order edits must not change identity. (Note: there is no content hashing anywhere in the codebase today; existing "signatures" are dict fingerprints, e.g. `grid.signature()`.)
- **Catalog identity lives in the sidecar, not the name.** The output is `f(template, shard, catalog snapshot, code version)` — the template hash alone proves "already ran" only for a frozen catalog, and ATL03 is a living collection. The stats sidecar records granule count + a hash of sorted granule ids + zagg version; the dedup/estimate check consults **name and sidecar**, never the name alone.
- **Add a logical content hash of the written data to the sidecar.** Outputs have been byte-identical between runs in practice (#282 was the rare, explicitly flagged exception, and still `np.isclose`-equivalent), so a data hash turns "probably identical" into checkable — nix-like verification a posteriori, without folding the catalog into the name and churning names on every catalog growth. Hash the *decoded* arrays, not compressed bytes (codec/library version bumps can change bytes while data is identical).
- **Hash → template registry at the store root**: write the template itself as `{hash}.yaml` next to the data, or the names are opaque (nix keeps derivations in the store for exactly this reason).

## Blast radius (coordinate before landing)

- `check_node_invariant` loses its self-validation; placement validation moves to the sidecar's recorded `shard_key`.
- Windowed leaves `{full_id}_{window}.zarr` (issue #246) become `{hash}_{window}.zarr`.
[...]

Issue #298, opened 2026-07-19
Invoke cost reporting (max/estimated/actual) and notebook dispatch wrapper


espg:
🤖 *from Claude*

Sub-issue of #296 — slice (2): report max / estimated / actual cost in the invoke return, and add a notebook-facing dispatch wrapper with a progress bar.

## Context

The cost math already exists but only surfaces as a printed summary: `runner.py` computes `gb_seconds = total_lambda_time × memory_gb` and `estimated_cost_usd = gb_seconds × LAMBDA_PRICE_PER_GB_SEC`, printed at end of run. The return payload has no cost block, no precomputed ceiling, and there is no progress UI anywhere in the package.

Goal, per #296:

- **Max cost** up front: `n_shards (shardmap) × per-second rate for the configured memory (template) × 900 s`. Cheap and exact as a ceiling.
- **Actual cost** in the structured return: rolled up from per-worker `duration_s × memory` (an estimate of the bill — billed duration differs slightly from measured; AWS Cost Explorer reconciles the truth, see the dashboards sub-issue).
- **Estimated cost** (optional, deferred): from prior-run sidecars at the shard prefix — depends on the sidecar and template-hash-naming sub-issues.
- **Notebook wrapper**: call invoke from IPython, get a properly formatted response object and a tqdm-style progress bar driven by the existing async result-polling loop.

## Dependency note (settled on #296)

tqdm is already in the dependency closure (earthaccess → pqdm → tqdm), and the orchestrator/invoke side runs outside Lambda, so the `lambda` extra is unaffected. Decision: **optional import** — progress bar if tqdm is importable, plain logging otherwise. No new hard dependency.

espg:
🤖 *from Claude*

## Implementation plan

**Phase 1 — cost block in the run return.** Extend the runner/dispatch summary (`src/zagg/runner.py` cost math, `src/zagg/dispatch.py` `RunReport`) with a structured block: `max_cost_usd` (= `n_shards × rate(memory_gb) × 900`, computable *before* invoke from shardmap + template), `actual_cost_usd` (existing `gb_seconds × LAMBDA_PRICE_PER_GB_SEC` rollup), `estimated_cost_usd: None` placeholder. Rate table keyed by arch (arm64/x86_64) — the price constant already exists. Return it from the dispatch entrypoints, print max before fan-out.

[...]

PR #302, opened 2026-07-19
Per-shard stats sidecar: schema, worker sidecars, run parquet (issue #297)

### LINKED ISSUE #297: Per-shard stats sidecar: standardized schema, persisted worker results
🤖 *from Claude*

Sub-issue of #296 — slice (1): standardize the per-shard stats sidecar and persist what workers already report.

## Context

The telemetry #296 wants already flows back through the worker return envelope — it just isn't kept or schema'd:

- Spatial workers return `metadata` with `duration_s`, `total_obs`, `cells_with_data`, `files_processed`, `time_range`, `max_memory_mb`, `container_hwm_mb`, and (when `profile` is on) `phase_timings` with the read/index/aggregate split plus `spill_write_s`/`spill_read_s`/`spill_bytes` (`src/zagg/processing/worker.py`).
- Raster workers return the analogous `phase_timings` (`sample`/`write` + per-stage `open`/`geometry`/`gather`/`fetch`/`decode` stats) (`src/zagg/processing/raster.py`).
- The async invoke path already mirrors a per-shard result JSON next to the store (`_write_result` in `deployment/aws/lambda_handler.py`) — but it's transient plumbing, not a kept record.

Today this information is only recoverable by grepping CloudWatch. Goal: a versioned, standardized per-shard sidecar written at the leaf on success, plus a run-level parquet (one row per shard, including failures) at the store root.

## Design points (from the #296 discussion)

- **Mergeable by construction**: the schema stores only associative stats — counts, sums, min/max, t-digests for distributions. No means/medians stored; derive at read. This makes the up-tree rollup (the unified-sweep sub-issue) a trivial fold.
- **Two formats, one source**: JSON per leaf (single small object, human-readable, fetchable by gridlook); parquet per run at the root (what duckdb/Athena/Grafana want). Both derive from the same worker dict.
- **Sidecars only exist on success.** Timed-out/OOM shards write nothing. The dispatcher's `RunReport` (`src/zagg/dispatch.py`) writes the run-level parquet *including* failure rows; the structured CloudWatch `processing_complete` log stays as the failure-forensics channel. This does not replace CloudWatch — it replaces grepping it for success-path stats.
- **Fields that future slices depend on**: `schema_version`, template hash, catalog identity (granule count + hash of sorted granule ids for the shard), zagg version, and a logical content hash of the written data — these are what the template-hash-naming sub-issue's dedup check consults.

espg:
Closes #297

Per-shard stats sidecar: a versioned, mergeable-by-construction record built from what workers already report, persisted two ways from one source — a `stats.json` sidecar SIBLING to each hive leaf `.zarr` (success only), and a run-level parquet at the store root (one row per shard, failure rows included). Decisions ratified on the issue thread (sibling placement, `stats.json` naming until #299's `{hash}.stats.json`, envelope ride, `invoked_by` resolved dispatcher-side via `sts get-caller-identity`) are implemented as agreed.

## Approach

- **`src/zagg/telemetry.py`** (new): `SCHEMA_VERSION = 1` record schema + `build_record` (pure builder from the existing worker `metadata` dict) + `merge` — an associative/commutative fold. Only associative stats are stored (counts, sums, min/max); identity-like fields (`shard_key`, `granules_sha256`, `invoked_by`, `lambda`, ...) merge as equal-or-`None` with `None` absorbing, which is what keeps the fold associative. `template_hash` is a nullable placeholder until #299 lands the hasher. Catalog identity is granule count + sha256 over the sorted granule ids.
- **Phase-timing collection is now always-on** in workers (the issue #297 decision: drop the `profile` gate for *collection*; `profile` still gates the verbose raster per-stage `stages` block and the summary rollups). `process_shard`, `process_and_write_hive`, `process_and_write_raster_hive`, and both Lambda per-unit handlers now always emit `phase_timings`.
- **Sidecar + envelope**: on success, the worker writes the record as a sibling of the leaf `.zarr` (`stats.json`; windowed leaves get `stats_{window}.json` so two windows of one shard cannot collide — mirrors the `{full_id}_{window}.zarr` leaf naming). The record also rides the response envelope (`body["stats"]`) so the async dispatcher builds the run parquet with no second S3 listing. `invoked_by` is resolved once per run by the dispatcher (`sts get-caller-identity` → ARN + userid), stamped into the invoke payload, and copied verbatim by the worker. Request ids / function ARNs of the run itself are omitted per the issue decision.
- **Run parquet**: one row per shard at the store root, `stats_{run_id}_{timestamp}.parquet`, written fail-open by the dispatcher from the `RunReport` results (spatial local + lambda) and the raster fan-out loops — including failure rows (error class, duration until failure, retry count) that have no sidecar. Written with the core `fastparquet` engine (pyarrow stays off the worker path per the #130 pivot); readable by pyarrow/duckdb.
- CloudWatch structured logging is unchanged.

## Phases

- [x] Phase 1 — schema + merge fold, pure/no-I/O (`src/zagg/telemetry.py`; unit tests for fold associativity/commutativity, merge-of-children == direct)
- [x] Phase 2 — workers write the leaf sidecar + envelope ride + always-on timing collection (`processing/worker.py`, `processing/raster.py`, `hive.py`, `deployment/aws/lambda_handler.py` application code, local-runner call sites; tests against the local-store path)
- [x] Phase 3 — run-level parquet from the dispatcher rollup (`runner.py` reading the `RunReport`), failure rows included; parquet round-trip test (fastparquet write, pyarrow + fastparquet read-back)
- [x] Follow-up (thread): raster read-volume counters (`raster_bytes_read` / `raster_px_decoded` / `raster_px_sampled`) in the record; raster summaries carry `run_stats_path`
- [x] Scope addition (directed on the PR #303 thread): `main` merged in post-#303 (merge, not rebase — pushed history is never rewritten; espg's "rebase" wording achieved via merge) and the #298 cost block extended to the **raster** Lambda path (`cost: {max, estimated, actual}` + `gb_seconds`/`price_per_gb_sec`), with the worker record pricing through the same arch-keyed table. The **temporal** path already received its cost block in PR #303 itself (`_run_lambda_events`) — verified, nothing further to add; local backends stay unmetered by design (parity with the spatial local path).

## Testing

- `tests/test_telemetry.py`: schema fields, cost pricing from the lambda config block, `invoked_by` verbatim copy, non-numeric phase entries dropped; merge identity, sums/maxes, equal-or-`None` collapse, success AND-fold, associativity, commutativity, merge-of-children == direct, empty/schema-version error paths.
- Phase 2: sidecar round-trip + windowed naming units; spatial hive Lambda handler writes the sibling record equal to `body["stats"]` with `invoked_by` verbatim; no-data hive shard writes no sidecar (record still rides the envelope); flat spatial carries envelope record only; raster hive Lambda + local backends checked end-to-end; hive/raster/worker phase-timing tests updated for always-on collection; object-count model tests updated (+1 per-shard object).
- Phase 3: `flatten_record` column mapping (phase/lambda/invoked_by flattening), `failure_record` + error-class derivation, parquet write/read round-trip (fastparquet + pyarrow), `_lambda_result_rows` success/stale-worker/failure branches, root listings across hive + flat pins updated; an autouse conftest guard keeps unit tests from PUT-ing the run parquet to real S3 (local-path writes stay live so the wiring is integration-tested).
- Full suite green locally (`ruff check`, `ruff format --check`, `pytest`) except two pre-existing items, both unrelated and untouched: `test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` (this machine's `pip` resolves for CPython 3.10, so `zarr>=3.1.5` cannot resolve) and a `ruff` `N818` on `src/zagg/registry.py:64` (`UnknownCapability`) that trips the full `select=[..., N]` config but not the CI lint bot's `E,F,W,I` selection.

## Questions for review

0. **`.github/scripts/bench_objects.py` was touched** (a CI benchmark *model* script, not a workflow): the sidecar is a new known per-leaf object, so the object-count model had to learn it (classify the node-sibling `stats*.json` as per-shard; expected per-shard +1) or every hive benchmark run and `test_benchmark_objects` would flag it as a stray object. Change is minimal and test-covered; flagging because CLAUDE.md is conservative about `.github/` — please eyeball that hunk.
1. ~~Flat layouts get no sidecar object~~ — **resolved on the thread**: hive is becoming the default; flat keeps the envelope record + parquet row only.
2. **Windowed sidecar naming**: the ratified name is `stats.json`, but a windowed store puts every window's leaf of a shard in one node directory, so a bare `stats.json` would self-clobber. Implemented `stats_{window}.json` for windowed leaves (bare leaves keep `stats.json` exactly). Both converge to `{hash}.stats.json` under #299.
3. **Local runs carry `invoked_by: null`** — the STS resolve is done only on the lambda dispatch paths (a local run may have no AWS credentials at all). Rows/records from local runs have a null identity column.
4. ~~Raster summaries don't carry `run_stats_path`~~ — **resolved on the thread**: all four summaries now carry it unconditionally (`None` when the fail-open write skipped, path on success); the raster parquet is still written before the all-failed raise. Follow-up on the thread also added the raster read-volume counters `raster_bytes_read` / `raster_px_decoded` / `raster_px_sampled` to the record (over-provision = decoded/sampled, derived at read).
5. ~~Overlap with the #298 PR~~ — **resolved**: PR #303 merged to main and `main` is merged into this branch (b3fec5f, clean); `telemetry.build_record` now prices via #303's `LAMBDA_PRICE_PER_GB_SEC_BY_ARCH` table so there is one rate source.

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -1123,6 +1123,10 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
                 "cells_with_data": meta.get("cells_with_data", 0),
                 "total_obs": meta["timesteps"],
                 "duration_s": time.time() - start_time,
+                # Read-volume counters (issue #297) for the stats record.
+                "raster_bytes_read": meta.get("raster_bytes_read"),
+                "raster_px_decoded": meta.get("raster_px_decoded"),
+                "raster_px_sampled": meta.get("raster_px_sampled"),
             }
             if meta.get("time_range") is not None:
                 body["time_range"] = meta["time_range"]
@@ -1222,6 +1226,10 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             "cells_with_data": grid.cells_per_shard if wrote else 0,
             "total_obs": meta["timesteps"],
             "duration_s": time.time() - start_time,
+            # Read-volume counters (issue #297) for the stats record.
+            "raster_bytes_read": meta.get("raster_bytes_read"),
+            "raster_px_decoded": meta.get("raster_px_decoded"),
+            "raster_px_sampled": meta.get("raster_px_sampled"),
         }
         # Worker memory telemetry (issue #250): sampled per-invocation peak,
         # container high-water fallback (see the process-mode stamp).
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -1139,9 +1139,11 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             if "phase_timings" in meta:
                 body["phase_timings"] = meta["phase_timings"]
             # Per-shard stats record (issue #297): envelope ride + the leaf
-            # sidecar (only when the unit wrote a leaf — timesteps > 0 means
-            # slabs streamed, so the leaf exists and is stamped). Fail-open on
-            # the sidecar PUT; the record still rides the envelope.
+            # sidecar (only when the unit wrote a leaf — ``leaf_written`` is the
+            # accurate signal, set iff a slab streamed and the leaf was stamped;
+            # a unit with acquisitions but no occupied cell writes no leaf, so
+            # ``timesteps`` alone would orphan a sidecar). Fail-open on the
+            # sidecar PUT; the record still rides the envelope.
             from zagg.telemetry import build_record, lambda_env, raster_granule_ids, write_sidecar
 
             record = build_record(
@@ -1152,7 +1154,7 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
                 lambda_config=lambda_env(),
             )
             body["stats"] = record
-            if meta["timesteps"]:
+            if meta.get("leaf_written"):
                 from zagg.hive import shard_leaf_path
 
                 try:
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -29,6 +29,12 @@ Event payload (default / process mode):
         Forwarded to ``process_shard`` so the worker fills the ``aoi_mask``
         column; absent when ``output.aoi_mask`` is off (the column is not
         allocated), keeping the flag-off event and outputs byte-identical.
+    "invoked_by": dict (optional, issue #297) -- {"arn", "userid"} caller
+        identity the dispatcher resolved once per run via sts
+        get-caller-identity; the worker copies it VERBATIM into the per-shard
+        stats record (sidecar + envelope). Workers cannot see the invoker
+        themselves (Event invokes carry no caller identity). Absent -> the
+        stats record carries null.
     "result_url": str (optional, issue #151) -- where to ALSO write this
         invocation's response envelope as JSON (e.g.
         "s3://bucket/out.zarr.status/<run_id>/<shard_label>.json", where the
@@ -1128,8 +1134,37 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
             body["max_memory_mb"] = (
                 sampled_peak if sampled_peak is not None else body["container_hwm_mb"]
             )
-            if profile and "phase_timings" in meta:
+            # Always-on collection (issue #297); the per-stage ``stages`` block
+            # inside stays gated on ``profile`` (raster.py).
+            if "phase_timings" in meta:
                 body["phase_timings"] = meta["phase_timings"]
+            # Per-shard stats record (issue #297): envelope ride + the leaf
+            # sidecar (only when the unit wrote a leaf — timesteps > 0 means
+            # slabs streamed, so the leaf exists and is stamped). Fail-open on
+            # the sidecar PUT; the record still rides the envelope.
+            from zagg.telemetry import build_record, lambda_env, raster_granule_ids, write_sidecar
+
+            record = build_record(
+                shard_key=shard_key,
+                metadata=body,
+                granule_ids=raster_granule_ids(event["granules"]),
+                invoked_by=event.get("invoked_by"),
+                lambda_config=lambda_env(),
+            )
+            body["stats"] = record
+            if meta["timesteps"]:
+                from zagg.hive import shard_leaf_path
+
+                try:
+                    window = event.get("window")
+                    leaf = shard_leaf_path(
+                        event["store_path"],
+                        shard_key,
+                        window=window["label"] if window else None,
+                    )
+                    write_sidecar(leaf, record, **_output_store_kwargs(event))
+                except Exception as e:
+                    logger.warning(f"stats sidecar write failed (fail-open, issue #297): {e}")
             return {"statusCode": 200, "body": json.dumps(body)}
 
         time_index = {k: int(v) for k, v in event["time_index"].items()}
@@ -1156,10 +1191,9 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
 
         def _write_slab(t_idx: int, slab: dict) -> None:
             nonlocal wrote, write_s
-            w0 = time.time() if profile else 0.0
+            w0 = time.time()
             write_raster_slab(store, grid, shard_key, t_idx, slab)
-            if profile:
-                write_s += time.time() - w0
+            write_s += time.time() - w0
             wrote = True
 
         _slabs, meta = process_raster_shard(
@@ -1195,13 +1229,25 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
         body["max_memory_mb"] = (
             sampled_peak if sampled_peak is not None else body["container_hwm_mb"]
         )
+        # Always-on sample/write split (issue #297); ``stages`` (issue #249)
+        # stays profile-gated verbosity.
+        body["phase_timings"] = {
+            "sample": (time.time() - start_time) - write_s,
+            "write": write_s,
+        }
         if profile:
-            total = time.time() - start_time
-            body["phase_timings"] = {
-                "sample": total - write_s,
-                "write": write_s,
-                "stages": stage_stats,
-            }
+            body["phase_timings"]["stages"] = stage_stats
+        # Stats record (issue #297): envelope ride only — a flat store has no
+        # per-shard leaf for a sidecar sibling (see the PR #302 discussion).
+        from zagg.telemetry import build_record, lambda_env, raster_granule_ids
+
+        body["stats"] = build_record(
+            shard_key=shard_key,
+            metadata=body,
+            granule_ids=raster_granule_ids(event["granules"]),
+            invoked_by=event.get("invoked_by"),
+            lambda_config=lambda_env(),
+        )
         return {"statusCode": 200, "body": json.dumps(body)}
     except Exception as e:
         logger.error(f"raster worker failed: {e}", exc_info=True)
@@ -1416,7 +1462,7 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
                 store = _get_store()
                 if store is None:
                     return  # template missing — recorded in write_error, skip the rest
-                _t0 = time.time() if profile else None
+                _t0 = time.time()
                 try:
                     # write_dataframe_to_zarr no-ops on an empty carrier, so no per-chunk
                     # emptiness check is needed. Use each chunk's own block_index.
@@ -1430,8 +1476,7 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
                     logger.error(f"Failed to write zarr to {store_path}: {e}")
                     write_error["msg"] = f"Failed to write zarr: {e}"
                     return
-                if profile:
-                    _write_elapsed += time.time() - _t0
+                _write_elapsed += time.time() - _t0
 
             chunk_results = [] if sharded else None
             _df_out, metadata = process_shard(
@@ -1454,13 +1499,12 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
             if sharded and chunk_results:
                 store = _get_store()
                 if store is not None:
-                    _write_t0 = time.time() if profile else None
+                    _write_t0 = time.time()
                     try:
                         write_shard_to_zarr(
                             chunk_results, store, grid=grid, shard_key=int(shard_key)
                         )
-                        if profile:
-                            _write_elapsed += time.time() - _write_t0
+                        _write_elapsed += time.time() - _write_t0
                     except Exception as e:
                         logger.error(f"Failed to write zarr to {store_path}: {e}")
                         write_error["msg"] = f"Failed to write zarr: {e}"
@@ -1471,14 +1515,13 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         if write_error:
             metadata["error"] = write_error["msg"]
 
-        # Record the write-phase timing (issue #100): read/index/aggregate come from
-        # ``process_shard``; ``write`` is the time spent in the streaming callback /
-        # sharded write. Only attach it on a clean write (no ``error``) so a time-to-
-        # failure is never folded in as a real write duration; the no-data path wrote
-        # nothing (``_write_elapsed`` stays 0) but also has no chunks, so writing 0 is
-        # harmless — gate on a populated ``phase_timings`` and no error to match the
-        # old "write absent on failure / no-data" contract.
-        if profile and not metadata.get("error") and "phase_timings" in metadata and store_box:
+        # Record the write-phase timing (issue #100; always-on collection since
+        # issue #297): read/index/aggregate come from ``process_shard``; ``write``
+        # is the time spent in the streaming callback / sharded write. Only attach
+        # it on a clean write (no ``error``) so a time-to-failure is never folded
+        # in as a real write duration; the no-data path wrote nothing and has no
+        # chunks — "write absent on failure / no-data" is unchanged.
+        if not metadata.get("error") and "phase_timings" in metadata and store_box:
             metadata["phase_timings"]["write"] = _write_elapsed
 
         # Peak worker RSS (issues #120, #141): captured here, after the write phase,
@@ -1504,6 +1547,36 @@ def _handle_process(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         metadata["cpu_seconds"] = round(
             (cpu_t1.user - cpu_t0.user) + (cpu_t1.system - cpu_t0.system), 3
         )
+
+        # Per-shard stats record (issue #297): rides the response envelope
+        # (``body["stats"]``) so the dispatcher builds the run parquet with no
+        # second S3 listing; hive leaves additionally get the ``stats.json``
+        # sidecar SIBLING to the leaf .zarr — success only (an error-free hive
+        # shard is by construction written + stamped). ``invoked_by`` is copied
+        # verbatim from the invoke payload (the dispatcher resolves it; the
+        # worker cannot). Fail-open: a sidecar PUT failure never fails a shard
+        # whose data landed — the record still rides the envelope.
+        from zagg.telemetry import build_record, lambda_env, write_sidecar
+
+        record = build_record(
+            shard_key=int(shard_key),
+            metadata=metadata,
+            granule_ids=event.get("granule_urls"),
+            invoked_by=event.get("invoked_by"),
+            lambda_config=lambda_env(),
+        )
+        metadata["stats"] = record
+        if get_store_layout(config) == "hive" and not metadata.get("error"):
+            from zagg.hive import shard_leaf_path
+
+            try:
+                window = event.get("window")
+                leaf = shard_leaf_path(
+                    store_path, int(shard_key), window=window["label"] if window else None
+                )
+                write_sidecar(leaf, record, **_output_store_kwargs(event))
+            except Exception as e:
+                logger.warning(f"stats sidecar write failed (fail-open, issue #297): {e}")
 
         # Log structured result
         logger.info(
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -1125,6 +1125,12 @@ def process_and_write_raster_hive(
     # prefix, and a worker error raised out above, leaving debris (D4). Write
     # order is pinned: dense slabs -> coverage sidecar -> stamp.
     meta["cells_with_data"] = 0
+    # Accurate leaf-written signal for the stats-sidecar gate (issue #297): set
+    # iff a slab streamed (``"store" in box``), so both dispatchers gate the
+    # sidecar PUT on leaf existence rather than the ``timesteps`` proxy (a unit
+    # with acquisitions but no occupied cell writes no leaf). ``phase_timings``
+    # cannot serve as the gate — it rides only under ``profile``.
+    meta["leaf_written"] = "store" in box
     if "store" in box:
         _t0 = time.time()
         words = occupied[0] if occupied and occupied[0].size else None
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -230,6 +230,7 @@ async def _sample_one(
     fill=0,
     geom_cache: dict | None = None,
     stage_stats: dict | None = None,
+    io_stats: dict | None = None,
 ):
     """:func:`sample_asset_async` body, also returning the raster's center
     ``(lon, lat)`` — the ownership rule's tile-center input (#218).
@@ -249,6 +250,15 @@ async def _sample_one(
     between read and write, so it is atomic by the same argument as the
     ``geom_cache`` store below — no locks; the ``write_buffer`` sink threads
     never touch this dict.
+
+    ``io_stats`` (issue #297) accumulates the read-volume counters for the
+    per-shard stats record, in place and by the same on-loop ``+=`` argument:
+    ``bytes_read`` (compressed tile bytes fetched), ``px_decoded`` (pixels in
+    the decoded tiles — whole tiles are read to sample a few cells), and
+    ``px_sampled`` (cell samples actually gathered). Unlike ``stage_stats``
+    this is ALWAYS-ON in the shard workers (the counters are a ``len()`` and
+    two multiplies per asset — no timing calls); ``None`` (the public
+    ``sample_asset*`` path) counts nothing.
     """
     from async_tiff import TIFF
 
@@ -304,6 +314,9 @@ async def _sample_one(
     vr, vc = rows[valid], cols[valid]
     tr, tc = vr // th, vc // tw
 
+    if io_stats is not None:
+        io_stats["px_sampled"] += int(vr.size)
+
     if vr.size == 0:
         if prof:
             stage_stats["gather"] += time.time() - _t0
@@ -318,6 +331,9 @@ async def _sample_one(
         stage_stats["fetch"] += time.time() - _t0
         stage_stats["tiles"] += len(pairs)
         _t0 = time.time()
+    if io_stats is not None:
+        io_stats["bytes_read"] += sum(len(t.compressed_bytes) for t in tiles)
+        io_stats["px_decoded"] += len(pairs) * th * tw
     decoded = await asyncio.gather(*[t.decode() for t in tiles])
     if prof:
         stage_stats["decode"] += time.time() - _t0
@@ -394,6 +410,7 @@ async def sample_item_async(
     anonymous: bool = True,
     geom_cache: dict | None = None,
     stage_stats: dict | None = None,
+    io_stats: dict | None = None,
 ):
     """Sample every configured band of one STAC item, concurrently.
 
@@ -438,6 +455,7 @@ async def sample_item_async(
                 fill=bands[f]["fill_value"],
                 geom_cache=geom_cache,
                 stage_stats=stage_stats,
+                io_stats=io_stats,
             )
             for f in fields
         ]
@@ -675,6 +693,12 @@ def process_raster_shard(
     # same argument as the geom_cache store; allocated only when a sink was
     # passed, so the default path is unchanged.
     occupied_acc = np.zeros(len(cells), dtype=bool) if occupied_out is not None else None
+    # Read-volume counters (issue #297): always-on inputs for the stats
+    # record — compressed bytes fetched, pixels decoded (whole tiles), and
+    # cell samples gathered. Their sampled/decoded ratio is the extract's
+    # over-provision at read time; stored raw (associative sums), never as a
+    # ratio, per the mergeable-by-construction schema rule.
+    io_stats = {"bytes_read": 0, "px_decoded": 0, "px_sampled": 0}
 
     async def _run_all():
         sem = asyncio.Semaphore(k)
@@ -693,6 +717,7 @@ def process_raster_shard(
                             anonymous=anonymous,
                             geom_cache=geom_cache,
                             stage_stats=stage_stats,
+                            io_stats=io_stats,
                         )
                         for e in items
                     ]
@@ -760,6 +785,10 @@ def process_raster_shard(
         "granule_count": len(granules),
         "skipped": skipped,
         "timesteps": len(groups),
+        # Read-volume counters (issue #297) for the stats record.
+        "raster_bytes_read": io_stats["bytes_read"],
+        "raster_px_decoded": io_stats["px_decoded"],
+        "raster_px_sampled": io_stats["px_sampled"],
     }
     return slabs, metadata
 
--- a/src/zagg/processing/raster.py
+++ b/src/zagg/processing/raster.py
@@ -332,7 +332,15 @@ async def _sample_one(
         stage_stats["tiles"] += len(pairs)
         _t0 = time.time()
     if io_stats is not None:
-        io_stats["bytes_read"] += sum(len(t.compressed_bytes) for t in tiles)
+        # compressed_bytes is one Buffer (chunky) or a list of Buffers (planar);
+        # normalize so bytes_read counts bytes, not buffers, if the single-band
+        # guard above (bits_per_sample != 1) is ever relaxed for multi-band COGs.
+        # A lone Buffer answers len() with its byte count; only the planar
+        # list/tuple needs per-buffer summing.
+        io_stats["bytes_read"] += sum(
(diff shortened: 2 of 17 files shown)


src/zagg/telemetry.py
```diff
@@ -0,0 +1,193 @@
...
+    error = metadata.get("error")
+    duration_s = float(metadata.get("duration_s") or 0.0)
+    gb_seconds = est_cost = None
+    if lambda_config and lambda_config.get("memory_mb"):
+        gb_seconds = duration_s * lambda_config["memory_mb"] / 1024.0
+        est_cost = gb_seconds * LAMBDA_PRICE_PER_GB_SEC
+    phase_timings = {
+        k: float(v)
+        for k, v in (metadata.get("phase_timings") or {}).items()
+        if isinstance(v, (int, float)) and not isinstance(v, bool)
```
espg, line 142: 🤖 *from Claude (review)*

`build_record` keeps every numeric `phase_timings` entry as a "timing", but the worker already puts a non-timing into that dict: `worker.py:717` sets `phase_timings["spill_bytes"] = buffered.spill_bytes` (a byte count, next to the `*_s` second values on 715/716). This filter passes it through, so `spill_bytes` rides in `phase_timings` and `merge` sums it per-key like a duration. The sum is numerically fine (bytes are additive), but in the phase-3 run parquet the `phase_timings` block gets flattened to columns that are otherwise seconds — a bytes value in a seconds-typed column will mislead any cost/latency query.

Suggest segregating it: either drop non-`*_s` keys here (keep only known timing names), or split byte/volume metrics into a separate `spill_bytes` field on the record so the timing block stays homogeneous. Cheapest is to filter on the key suffix rather than value type.

src/zagg/telemetry.py
```diff
@@ -0,0 +1,193 @@
...
+    records = list(records)
+    if not records:
+        raise ValueError("merge requires at least one record")
+    versions = {r.get("schema_version") for r in records}
+    if versions != {SCHEMA_VERSION}:
+        raise ValueError(f"cannot merge stats records with schema_version(s) {sorted(versions)}")
+    out: dict[str, Any] = {"schema_version": SCHEMA_VERSION}
+    for key in _EQ_OR_NONE_KEYS:
+        first = records[0].get(key)
+        out[key] = first if all(r.get(key) == first for r in records) else None
```

Aliasing asymmetry vs `build_record`. Here the equal-or-`None` merge does `out[key] = first` where `first = records[0].get(key)`. For the two dict-valued identity fields (`lambda`, `invoked_by`) that stores the *same object* the first input record holds — the merged record shares a nested dict with `records[0]`. Meanwhile `build_record` is careful to defensively copy both (`dict(lambda_config)`, `dict(invoked_by)`). So a fold result is not independent of its inputs the way a freshly built record is.

Today the records are treated as immutable so nothing breaks, but phases 2–3 mutate rows (the dispatcher stamps failure columns / assembles parquet rows), and a mutation of a rolled-up row would silently corrupt the leaf record it aliases. Cheap fix: copy dict values when carrying them through, e.g. `out[key] = dict(first) if isinstance(first, dict) else first`, mirroring `build_record`. (`phase_timings` is already rebuilt fresh, so only the two eq-or-None dicts are exposed.)

tests/test_telemetry.py
```diff
@@ -0,0 +1,162 @@
...
+        assert granules_sha256([]) is None
+        assert granules_sha256(None) is None
+
+
+class TestMerge:
+    def test_single_record_is_identity(self):
+        rec = _record()
+        _assert_records_close(merge([rec]), rec)
+
+    def test_sums_and_maxes(self):
```
espg, line 172: 🤖 *from Claude (review)*

Coverage gap: the merge fold of the cost fields is never exercised. Every `TestMerge` case builds records with `lambda_config=None`, so across the whole merge suite `gb_seconds` / `est_cost_usd` are `None` and `lambda` is `None` — the `_SUM_OR_NONE_KEYS` summation and the `lambda` equal-or-`None` collapse are asserted only against the empty branch. Given "mergeable-by-construction cost" is the headline property of this schema, the associative fold of populated cost should be pinned: e.g. merge two records with distinct `lambda_config` memory sizes and assert `gb_seconds`/`est_cost_usd` equal the sum of the parts, that a shared `lambda` block survives while differing ones collapse to `None`, and that a mixed populated/`None` set sums only the populated ones. `test_lambda_config_prices_cost` covers `build_record` pricing but not the fold.

src/zagg/telemetry.py
espg: 🤖 *from Claude*

Fixed in f5d9eed. `build_record` now splits any `*_bytes` phase entries out of the seconds-only `phase_timings` block and surfaces the volume as its own top-level `spill_bytes` field. It joins `_SUM_OR_NONE_KEYS` so the fold stays associative (byte counts add; `None` when unmetered). The worker's own `metadata["phase_timings"]` dict is unchanged — only the record schema splits it. Added `test_spill_bytes_split_out_of_timings` and a `spill_bytes is None` assertion to `test_schema_fields`.

src/zagg/telemetry.py

Fixed in 80f10a0. The equal-or-`None` branch now does `out[key] = dict(first) if isinstance(first, dict) else first`, so a rolled-up record no longer aliases the `lambda`/`invoked_by` dict of `records[0]` — mirroring `build_record`'s defensive copies. `test_cost_fields_shared_lambda_survives` pins it with `m["lambda"] is not a["lambda"]` / `is not b["lambda"]`.

tests/test_telemetry.py
espg: 🤖 *from Claude*

Fixed in 3ba0ecd. Added three `TestMerge` cases exercising the populated cost fold: `test_cost_fields_fold` (distinct `lambda_config` memory sizes → `gb_seconds`/`est_cost_usd` sum, differing `lambda` collapses to `None`), `test_cost_fields_shared_lambda_survives` (shared `lambda` block survives the collapse), and `test_cost_fields_mixed_populated_and_none` (one metered + one `None` → sum covers only the populated part, `lambda` collapses to `None`).

deployment/aws/lambda_handler.py
```diff
@@ -1128,8 +1134,37 @@ def _handle_process_raster(event: Dict[str, Any]) -> Dict[str, Any]:
...
+
+            record = build_record(
+                shard_key=shard_key,
+                metadata=body,
+                granule_ids=raster_granule_ids(event["granules"]),
+                invoked_by=event.get("invoked_by"),
+                lambda_config=lambda_env(),
+            )
+            body["stats"] = record
+            if meta["timesteps"]:
```

The raster-hive sidecar is gated on `meta["timesteps"]`, but a nonzero timestep count does not prove a leaf was written. In `process_and_write_raster_hive` the leaf is created + stamped only inside `if "store" in box:` (raster.py ~L1128), which is reached only when at least one slab was actually streamed via `_write_slab`. A unit with acquisitions (`timesteps > 0`) but no sampled/occupied cells returns with `"store" not in box` — no `.zarr`, no stamp, and `meta["phase_timings"]` absent — yet `meta["timesteps"]` is still truthy.

Why it matters: `write_sidecar` then PUTs `stats.json` as a SIBLING of a leaf that does not exist — an orphan sidecar, violating the ratified "sidecar on success only, where the leaf actually exists" invariant. It also breaks `.github/scripts/bench_objects.py`, whose hive model attributes that node’s `stats.json` to a shard it expects to have a full leaf underneath (per_shard/total mismatch → the object-count benchmark fails). The same weak gate is in the local path (`runner.py`, `store_layout == "hive" and meta.get("timesteps")`).

Fix: gate on the same signal the stamp uses — e.g. `if "phase_timings" in meta:` (populated only under `"store" in box`), or add an explicit `wrote`/`stamped` flag to the raster meta and gate both the Lambda and local sidecar writes on it. The spatial paths already use the accurate `not metadata.get("error")` leaf gate.

src/zagg/runner.py
```diff
@@ -844,6 +846,27 @@ def _write_slab(t_idx, slab):
...
                     f"raster shard {shard_label(grid, int(shard_key))} stages: {stage_stats}"
                 )
+            # Per-shard stats record (issue #297), the raster-local flavor:
+            # hive units that wrote a leaf (timesteps > 0 -> stamped) get the
+            # sidecar sibling; the record rides ``meta`` for the run parquet
+            # either way. Fail-open on the sidecar PUT.
+            meta.setdefault("duration_s", time.time() - unit_t0)
+            record = build_record(
+                shard_key=int(shard_key),
+                metadata=meta,
```
espg, line 931: 🤖 *from Claude (review)*

The local raster path builds the stats record straight from `meta`, but `process_raster_shard` / `process_and_write_raster_hive` never set `total_obs` (the raster meta carries `timesteps`, `granule_count`, `skipped`, `cells_with_data`). `build_record` reads `metadata.get("total_obs") or 0`, so this record gets `n_obs = 0`. The Lambda raster handler, by contrast, injects `body["total_obs"] = meta["timesteps"]` *before* `build_record` (lambda_handler.py ~L1124).

Why it matters: the same raster shard yields `n_obs = timesteps` on Lambda but `n_obs = 0` locally — a silent cross-backend divergence in a schema whose entire purpose is a backend-independent, mergeable record (and it flows into the phase-3 run parquet). It is untested: `test_raster_runner.py` asserts `schema_version`/`success`/`shard_key`/`phase_timings` but not `n_obs`.

Fix: mirror the handler before `build_record` here — e.g. `meta["total_obs"] = meta.get("timesteps", 0)` (raster’s obs-count convention) — or pass the obs count explicitly, and add an `n_obs` assertion to the raster-runner test.

src/zagg/runner.py
```diff
@@ -2389,6 +2463,9 @@ def _run_lambda(
     # over a not-yet-built name.
     session = boto3.Session()
     state: dict = {}
+    # Caller identity for the per-shard stats records (issue #297): resolved
+    # once per run, stamped into every cell event; workers copy it verbatim.
+    invoked_by = _resolve_invoked_by(session, region)
```
espg, line 2793: 🤖 *from Claude (review)*

The two `_resolve_invoked_by` call sites use different client seams. RasterStrategy (~L1058) calls `_resolve_invoked_by(boto3, region)` — the module — and its comment explicitly relies on tests patching `boto3.client` to intercept the `sts get-caller-identity`. Here `_run_lambda` passes a `boto3.Session()` instance, whose `.client(...)` is NOT intercepted by a `boto3.client` patch.

Why it matters: spatial lambda-path tests that patch `boto3.client` will not intercept this STS call, so it either makes a real network call (fail-open → `None`, leaving spatial `invoked_by` stamping effectively untested) or, in a credentialed CI environment, resolves and stamps the runner’s real ARN/UserId into test-produced records. Either way the spatial `invoked_by` plumbing is unverified and behaves differently from the raster path.

Fix: route both sites through the same seam (pass the module, or the shared `session`, consistently), and add a spatial `_run_lambda` test asserting `invoked_by` reaches the event payload via the same mock the other AWS calls use.

deployment/aws/lambda_handler.py

Folded in 32e690c. Added an explicit `meta["leaf_written"] = "store" in box` flag in `process_and_write_raster_hive` (raster.py, set alongside `cells_with_data`), and switched BOTH sidecar gates to it: the Lambda hive branch (`if meta.get("leaf_written")`, lambda_handler.py) and the local `RasterStrategy._one` (`store_layout == "hive" and meta.get("leaf_written")`, runner.py).

One correction to the suggested `"phase_timings" in meta` gate: that key rides only under `profile` (`if profile and "store" in box:` in raster.py), so it would drop the sidecar on every non-profiled leaf write — hence the explicit `leaf_written` flag (your second suggestion) instead. Added `test_hive_no_leaf_writes_no_orphan_sidecar` (test_lambda_raster_handler.py): patches `process_raster_shard` to report `timesteps=1` without streaming a slab, and asserts `read_sidecar(leaf) is None` while the record still rides the response envelope.

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded in 612f41a. Mirrored the Lambda handler in `RasterStrategy._one` (runner.py): `meta.setdefault("total_obs", meta.get("timesteps", 0))` before `build_record`, so `n_obs` is the timestep count on both backends instead of `0` locally. Added an assertion to `test_schedule_none_end_to_end` (test_raster_runner.py): `record["n_obs"] == red.shape[0] == 2`, tying the record obs count to the leaf time axis.

src/zagg/runner.py
espg: 🤖 *from Claude*

Partial-fold in 5b44c06 — documented the seam + added the missing spatial tests, but kept the per-path seam by design.

On unification: I verified the two harnesses patch DIFFERENT seams. The raster `_run_lambda_shards` tests patch `boto3.client` (test_raster_runner.py, e.g. `monkeypatch.setattr(boto3, "client", ...)`), so the module route (`_resolve_invoked_by(boto3, ...)`) hits the fake. The spatial `_run_lambda` tests patch `boto3.Session` (test_runner.py:2780, `monkeypatch.setattr(boto3, "Session", lambda *a, **k: MagicMock())`), so the session route (`_resolve_invoked_by(session, ...)`) degrades to None via the shape-check. Routing both through one seam would leave the OTHER family unpatched, sending a real STS `get-caller-identity` from unit tests — exactly the failure mode you flagged, just relocated. So full unification is not safe without also rewriting several spatial harness sites; I judged that out of scope for a fold.

What I changed: (a) explanatory comments at both call sites (runner.py ~1073 raster, ~2579 spatial) and in the `_resolve_invoked_by` docstring, naming the harness-seam constraint and that both a module and a `Session` expose `.client`. (b) The spatial verification you asked for, split into two focused pieces so no real network is possible:
- `TestResolveInvokedBy` (test_runner.py): a stubbed session returning a real `{Arn, UserId}` dict yields `{"arn","userid"}` and calls `.client("sts", region_name=...)` once; a bare `MagicMock()` session degrades to `None` (shape-check); a raising session degrades to `None` (fail-open).
- `TestInvokeLambdaCellEvent.test_invoked_by_adds_event_key` / `test_default_event_has_no_invoked_by_key`: assert `invoked_by` threads into the captured event payload when present, and the key is absent (byte-identical to pre-#297) when `None`.

Happy to take the full unification as a follow-up if you want the harnesses reworked to a single seam.

src/zagg/runner.py
```diff
@@ -1880,6 +1967,88 @@ def _raster_windowed_units(cells: list[tuple], windowing: dict) -> list:
...
+    data landed — the leaf sidecars remain the durable per-shard truth.
+    """
+    if not rows:
+        return
+    try:
+        from zagg.telemetry import write_run_parquet
+
+        path = write_run_parquet(store_path, rows, run_id=run_id, store_kwargs=store_kwargs)
+        if summary is not None:
+            summary["run_stats_path"] = path
```
espg, line 2175: 🤖 *from Claude (review)*

`summary["run_stats_path"]` is set **only inside the try, after a successful PUT**. On the fail-open branch (`except` → warn), when `rows` is empty (the early `return` above), or whenever the write is skipped, the key is never added — so `run_stats_path` is a *conditionally present* summary key.

Two consequences:
- **Non-deterministic key set in production.** `_run_lambda` passes `summary=summary`, so a lambda run that PUTs the parquet emits `run_stats_path`, but the exact fail-open scenario this helper is built for (the invoke-only OIDC role that can't PUT) emits a summary *without* it. A consumer indexing `summary["run_stats_path"]` `KeyError`s on precisely the fail-open runs. This breaks the byte-identical-key-set invariant `TestSummaryKeysByteIdentical` exists to protect.
- **The gap is masked, not caught.** `_LAMBDA_KEYS` (tests/test_runner.py:1488) was left *without* `run_stats_path`, and `test_lambda_summary_keys_and_cost` passes only because the autouse conftest guard short-circuits the `s3://` store before the key is set. Production `_run_lambda` on a real s3 store *does* add the key on success, so the pinned lambda key set silently diverges from prod.

Fix: set `summary["run_stats_path"] = None` unconditionally (before the empty-rows `return` / outside the try), overwrite with `path` on success, and add `run_stats_path` to `_LAMBDA_KEYS`. Then the summary schema is stable regardless of write outcome.

tests/conftest.py
```diff
@@ -51,3 +51,26 @@ def point_words(n, seed, lat0=45.0, lon0=45.0, spread=1e-4):
...
+    credentials could reach a real bucket). Local-path writes stay live so
+    the parquet wiring is still integration-tested; a test that wants the
+    s3 branch re-patches ``zagg.runner._write_run_stats`` itself.
+    """
+    from zagg import runner
+
+    real = runner._write_run_stats
+
+    def guard(store_path, rows, **kwargs):
+        if str(store_path).startswith("s3://"):
```
espg: 🤖 *from Claude (review)*

The guard only short-circuits `s3://` store paths — every **local relative** store path (e.g. the `"./out.zarr"` used by `TestSummaryKeysByteIdentical` and the other `_run_local` tests) is passed straight through to the real `_write_run_stats`, which does a live `obstore.put` into the repo working directory. Because `run_id` is a fresh `uuid4` per call and nothing cleans up, these files **accumulate**: this worktree already carries a stray `./out.zarr/` holding six `stats_*.parquet` objects (plus `morton_hive.json`) left by prior test runs. That pollutes the working tree and risks being committed by an `git add`-all, and it means the parquet-write wiring is "integration tested" by scribbling into the repo root instead of a `tmp_path`.

Fix: point those `_run_local` tests at a `tmp_path` store instead of `"./out.zarr"`, or broaden the guard to also short-circuit / redirect non-tmp local writes. At minimum the accumulating `out.zarr/` must not be produced in the tree.

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded in 8062c02. `_write_run_stats` now sets `summary["run_stats_path"] = None` unconditionally up front (before the empty-rows `return` and outside the `try`), and only overwrites it with the real path once the PUT lands — so the fail-open branch, the empty-rows early return, and a skipped write all leave the key *present* (None) instead of absent. The summary schema is now stable regardless of write outcome.

Also added `run_stats_path` to `_LAMBDA_KEYS` (tests/test_runner.py), matching `_LOCAL_KEYS`. To stop the autouse conftest guard from masking the key on the `s3://` short-circuit, its skip branch now mirrors the same contract (`summary["run_stats_path"] = None`), so `test_lambda_summary_keys_and_cost` exercises the real key set rather than passing by omission. Both summary-key pin tests (TestSummaryKeysByteIdentical, 7 cases) stay green.

tests/conftest.py
espg: 🤖 *from Claude*

Folded in 8c91982. Broadened the autouse guard (tests/conftest.py) so it short-circuits **any non-absolute** store path in addition to `s3://` — `if sp.startswith("s3://") or not os.path.isabs(sp)`. The relative `"./out.zarr"` stores used by TestSummaryKeysByteIdentical and the other `_run_local` tests no longer reach the live `obstore.put`, so no more `stats_*.parquet` accumulates in the repo cwd. The skip branch still stamps `run_stats_path = None` so the summary schema stays stable (per the sibling fold).

The tests that assert the parquet actually exists all drive **absolute** `tmp_path` stores, so they keep exercising the real write: test_hive.py TestRunnerWiring (`str(tmp_path / "store")`), test_coverage_root.py (`tmp_path / "out"`), test_raster_runner.py flat pin (`tmp_path / "out.zarr"`), and test_benchmark_objects.py hive tests (`tmp_path / "out"`) — all green (199 passed across the affected files; 421 across the full gate).

Verified the stray `./out.zarr/` no longer regrows its six `stats_*.parquet` after a full test run — only a single fixed-name `morton_hive.json` is left, and that comes from a separate hive-manifest write path (not the run-stats sidecar), so it overwrites rather than accumulates; out of scope for the run-stats guard. `rm -rf ./out.zarr` locally; not committed.

espg:
> Flat layouts get no sidecar object — the ratified placement ("sibling next to the leaf .zarr") only exists on hive; flat stores have no per-shard leaf. Flat shards still get the record in the envelope and a row in the run parquet. If a flat per-shard object is wanted, the .status/ prefix convention could host it — say the word.

hive is becoming the default, this is fine

> Spatial summaries always carry run_stats_path (None when the fail-open write skipped, the path on success — review fold made the key unconditional); raster summaries don't carry it — the raster paths write the parquet before their all-failed guard (so failure evidence persists even when the run raises), which is before the summary dict exists; the spatial local/lambda summaries do carry the key. Easy to add if wanted.

feel like raster should carry it-- whats the difference in stats here? still carry obs, cost... do we also carry MBs of raster read, percentage miss / over provision on the extract?

espg:
🤖 *from Claude*

**Flat sidecar** — acknowledged, leaving as-is (envelope record + run-parquet row for flat shards, no per-shard object).

**Raster `run_stats_path`** — done in 6fb5397: both raster summaries (local + lambda) now carry the key with the spatial semantics (path on success, `None` when the fail-open write skipped). The parquet is still written *before* the all-failed raise so failure evidence persists; `_write_run_stats` now returns the path and the summaries fold it in afterwards.

**What the raster record carries vs spatial** — same schema (it's one schema, off-raster fields are null and vice versa):

- **Carried all along**: `n_obs` (= timesteps written, the raster obs analog), `n_granules` (dispatched acquisition entries, incl. skipped asset-less ones), `granules_sha256` (over STAC item ids / datetimes), `cells_with_data` (occupied-cell union), `duration_s`, `gb_seconds` + `est_cost_usd` (function memory × duration), `max_memory_mb` / `container_hwm_mb`, the lambda config block, `invoked_by`, success/error, timestamp. `phase_timings` in the record is the always-on `{sample, write}` split; the per-stage seconds (`open`/`geometry`/`fetch`/`decode`/`gather`) and counts (`assets`/`tiles`/`geom_hits`) are the issue #249 verbosity block — they ride the *envelope* under `--profile` only and are deliberately excluded from the record (non-numeric block; the parquet flattens phase columns as seconds).
- **Was NOT carried — added now in b4cd109**, since both turned out trivially countable at the fetch stage:
  - `raster_bytes_read` — compressed COG tile bytes actually fetched (`len(Tile.compressed_bytes)` summed at the async-tiff fetch), i.e. the source-egress volume. Decoded volume is derivable as `raster_px_decoded × dtype size`.
  - `raster_px_decoded` + `raster_px_sampled` — pixels in the fetched tiles vs cell samples actually gathered. **Over-provision on the extract = `px_decoded / px_sampled`**, derived at read (stored raw, never as a ratio, keeping the schema mergeable-by-construction; all three fold as None-aware sums like `spill_bytes`).

One precision on "percentage miss": at *tile* granularity there is no over-fetch by construction — the sampler `np.unique`s the tiles touched by ≥ 1 sampled cell and fetches only those, so every fetched tile contributes. The waste is *pixel*-level (whole tiles decoded to sample a few cell centers), which is exactly what `px_decoded` vs `px_sampled` measures — at NEON-o9-like geometries expect ratios in the hundreds, dropping as shard order coarsens. (`geom_hits/assets` is a different thing — geometry-*cache* hit rate, issue #244 — and stays in the profile stages.)

**Cost fold-in from #303** (per the direction on that thread): `main` is merged in (merge commit b3fec5f — merge, not rebase: pushed history is never rewritten per the repo rules; same integration), and 4141ee9 adds the #298 cost block to the raster Lambda summary — `cost: {max_cost_usd, estimated_cost_usd, actual_cost_usd}` plus `gb_seconds`/`price_per_gb_sec` and the legacy flat `estimated_cost_usd` actual-rollup, with failed units' landed durations included in the billed rollup. The worker-side record now prices through the same #298 arch-keyed table (`platform.machine()` → table key, flat-rate fallback), so sidecar and summary agree on the rate. The temporal path already got its cost block in #303 itself (`_run_lambda_events`) — nothing to add there; local backends stay unmetered by design, matching the spatial local path.

Full suite green locally; CI running on 4141ee9.

src/zagg/runner.py
```diff
@@ -1294,13 +1370,53 @@ def _one(pair):
...
         # null-safe on bodies from a worker that predates either field.
         ok_bodies = [b for _k, b in ok_units]
         durations = [float(b["duration_s"]) for b in ok_bodies if b.get("duration_s") is not None]
         mems = [float(b["max_memory_mb"]) for b in ok_bodies if b.get("max_memory_mb") is not None]
+        # Actual-cost rollup (issues #298/#297): billed durations across
+        # successes AND landed failures (a failed unit's body still carries
+        # its duration until failure), priced like the spatial path. The
+        # structured cost block mirrors _run_lambda's; estimate_cost_usd is
+        # the Phase-3 stub (None until the #297/#299 history exists).
+        billed_s = sum(durations) + sum(
```
espg, line 1408: 🤖 *from Claude (review)*

The distinctive new behavior of this commit — billing **landed failures** into the rollup — is untested. `billed_s` adds `duration_s` from every `stats_failures` body, but the only cost-block test (`test_raster_lambda_backend`) drives a single *successful* unit (`duration_s: 30.0`, `cells_error == 0`); the partial-failure test (`test_partial_shard_failure`, `cells_error == 1`) runs the **local** backend, which has no cost block. So deleting the `+ sum(... for _k, _e, b in stats_failures ...)` term entirely would keep every test green.

Add a lambda-path test with a mixed fan-out: one 200 body (e.g. 30 s) and one non-200 body that still carries a `duration_s` (e.g. `{"statusCode": 500, "body": {"error": "boom", "duration_s": 12.0}}`), then assert `summary["gb_seconds"] == (30+12) * memory_gb` and `cost["actual_cost_usd"]` reflects both. That pins the failed-unit-billing claim and guards against the ok/failed lists double-counting a unit.

src/zagg/processing/raster.py
```diff
@@ -318,6 +331,9 @@ async def _sample_one(
         stage_stats["fetch"] += time.time() - _t0
         stage_stats["tiles"] += len(pairs)
         _t0 = time.time()
+    if io_stats is not None:
+        io_stats["bytes_read"] += sum(len(t.compressed_bytes) for t in tiles)
```

`len(t.compressed_bytes)` means byte-length **only** for the chunky/pixel-interleaved case. Per the async-tiff stub, `Tile.compressed_bytes -> Buffer | list[Buffer]` — "a single buffer for pixel-interleaved (chunky) data, or a **list of buffers** for band-interleaved (planar) data." On the planar branch `len()` returns the *buffer count* (e.g. 3), not bytes, silently collapsing `bytes_read` to a handful.

Today this is safe purely because `_sample_one` hard-rejects multi-sample rasters three lines up (`len(ifd.bits_per_sample) != 1` -> ValueError), and single-band COGs are always chunky. But nothing here ties the `len()` to that guarantee, so if the single-band restriction is ever relaxed (multi-band COG support) this counter becomes garbage with no test catching it. Either normalize — `b = t.compressed_bytes; nbytes = len(b) if isinstance(b, (bytes, bytearray, memoryview)) else sum(len(x) for x in b)` — or add a one-line comment pinning the byte semantics to the single-band guard so the assumption is explicit.

tests/test_lambda_raster_handler.py
```diff
@@ -522,6 +527,65 @@ def test_schedule_none_bare_leaf(self, handler_mod, tmp_path):
...
+        assert record["invoked_by"] == event["invoked_by"]
+        assert record["n_granules"] == 1
+        # Always-on sample/write collection flows into the record.
+        assert {"sample", "write"} <= set(record["phase_timings"])
+        assert record["max_memory_mb"] is not None
+        # Read-volume counters (issue #297): whole tiles are fetched+decoded
+        # to sample the shard's cells, so decoded pixels bound samples and the
+        # over-provision ratio is derivable at read.
+        assert record["raster_bytes_read"] > 0
+        assert record["raster_px_decoded"] >= record["raster_px_sampled"] > 0
```
espg: 🤖 *from Claude (review)*

This invariant (`px_decoded >= px_sampled`) is fixture-specific, not universal — and the "over-provision = px_decoded / px_sampled" framing in the counter comments (raster.py ~698, telemetry.py ~155) inverts with it. `px_sampled` accumulates `vr.size` = **valid output cells**, while `px_decoded` = `len(pairs) * th * tw` = **source pixels in the fetched tiles**. When the output grid is finer than the source raster, many cells NN-pull the same pixel, so cell count can exceed decoded pixel count. This is reachable in practice: S2 is 10 m, but HEALPix order 16 (~6 m) / 17 (~3 m) are finer, so a high-order run oversamples and `px_sampled > px_decoded`, making the ratio < 1 and this assertion false.

The counters are each individually correct — only the *interpretation* and this assert are regime-bound. Suggest softening to note the assumption (source finer than output) or dropping the `>=` and asserting each `> 0`, so a legitimate oversampling fixture does not trip it.

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded in 437ae81. Added `TestRasterLambdaBackend::test_mixed_fanout_bills_landed_failure` — a 2-shard sync fan-out where shard0 lands a 200 body (`duration_s: 30.0`) and shard1 lands a 500 whose body still carries `duration_s: 12.0`. It asserts `gb_seconds == (30 + 12) * 4` and `cost["actual_cost_usd"]`/`estimated_cost_usd` reflect both durations, while `lambda_time_s == 30.0` stays ok-only. Deleting the `stats_failures` billing term drops it to `30 * 4` and fails here; the assertion `cells_error == 1 and cells_with_data == 1` confirms a partial failure does not trip the all-failed raise.

src/zagg/processing/raster.py
espg: 🤖 *from Claude*

Folded in 2f994de. Normalized the tally: `sum(len(x) for x in cb) if isinstance(cb, (list, tuple)) else len(cb)` per tile, so bytes_read counts bytes on both the chunky (single Buffer) and planar (list of Buffers) branches. One correction to the suggested form: async-tiff `Buffer` is not a `bytes`/`bytearray`/`memoryview`, so the `isinstance(cb, (bytes, bytearray, memoryview))` variant would have fallen through to `sum(len(x) for x in cb)` on the real single-Buffer path and iterated it byte-by-byte (`int has no len()`) — it broke 20 raster-pipeline tests locally. Keying on `(list, tuple)` (the planar case) instead routes correctly: a lone Buffer answers `len()` with its byte count, only the planar list needs per-buffer summing. Full raster suite green.

tests/test_lambda_raster_handler.py
espg: 🤖 *from Claude*

Folded in 3aa5d3d. Relaxed both regime-bound asserts — `tests/test_lambda_raster_handler.py:556` and the same one in `tests/test_raster_runner.py` (`test_schedule_none_end_to_end`) — from `raster_px_decoded >= raster_px_sampled > 0` to asserting each counter `> 0`, so a legitimate oversampling fixture (grid finer than source) no longer trips them. Also softened the over-provision framing where it appears: the `io_stats` comment in `raster.py` and the `build_record` comment in `telemetry.py` now note the decoded/sampled ratio reads as read-time over-provision only when the output grid is coarser than the source, and that a finer grid pushes it below 1. Counters left individually correct and stored raw.

PR #303, opened 2026-07-19
Invoke cost reporting (max/estimated/actual) and notebook dispatch wrapper

### LINKED ISSUE #298: Invoke cost reporting (max/estimated/actual) and notebook dispatch wrapper

espg:
Closes #298

Invoke cost reporting (max / estimated / actual) in the run return, plus a notebook-facing dispatch wrapper with a progress bar. Design per the plan comment and follow-ups on #298: new module for the wrapper, CLI gates on max cost with a skippable prompt while the notebook call never blocks, tqdm via optional import, and the Phase-3 estimator stubbed only (deferred behind #297/#299).

## Approach

**Cost block (phase 1).** `dispatch.py` gains an arch-keyed price table (`LAMBDA_PRICE_PER_GB_SEC_BY_ARCH`) and `max_cost_usd(n_shards, memory_gb, timeout_s, arch)` — the pre-invoke ceiling `n_shards × rate(arch) × memory_gb × 900 s`, computable from shardmap + config alone. `RunReport` carries `max_cost_usd` / `estimated_cost_usd` / `actual_cost_usd` (stamped by the runner). Both Lambda paths (`_run_lambda`, `_run_lambda_events`) log the ceiling before fan-out and add a structured `summary["cost"] = {max_cost_usd, estimated_cost_usd: None, actual_cost_usd}` block. The legacy flat `estimated_cost_usd` summary key keeps its pre-existing meaning (the billed-duration rollup) for existing consumers.

**One-arch price table.** Only arm64 is deployed (`deployment/aws/template.yaml` `Architecture` default `arm64`; no x86_64 variant is stood up, and `runner.py`'s pricing comment already says "arm64 pricing"), so the table carries a single arm64 row with the arch recorded — an x86_64 deployment adds its rate as a second entry. An unknown arch raises `KeyError` instead of pricing silently wrong.

**Worker-variant memory.** `_worker_memory_gb(config)` resolves the billed memory from the config's `worker:` block (issue #235 variants 2048/4096/8192 MB), defaulting to the flat 4 GB constant. It now also feeds `LambdaExecutor.measure_cost` and the rollup, so a `worker: {memory: 8192}` run bills 8 GB instead of the old hardcoded 4 — a small behavioral fix; the default path is byte-identical.

**Notebook wrapper + progress (phase 2).** New `src/zagg/notebook.py` (per the ratified decision — `dispatch.py` stays free of UI concerns): `run()` wraps `agg` and returns a `RunView` with a `_repr_html_` cost/counters/failures report; `max_cost_preview()` resolves the ceiling from shardmap + config with the same unit accounting as `_run_lambda` (cell selection + windowed-unit expansion); `confirm_max_cost()` is the blocking yes/no gate, wired CLI-only in `__main__` behind `--yes`/`-y` (the notebook path only displays — it never prompts, and a test asserts `input` is unreachable there). Progress rides a new `on_progress(done, total, cost_usd)` callback threaded through `agg` into every fan-out accumulator (spatial local/lambda, temporal local/lambda via a shared `_with_progress` wrapper; raster local/lambda loops tick with `cost_usd=None` since that path has no per-unit pricing). tqdm is an optional import (`tqdm.auto`): bar when importable, logging fallback otherwise — no dependency changes, `lambda` extra untouched.

**Estimator stub + example notebook (phase 3).** `dispatch.estimate_cost_usd(catalog_data, template_hash=None, history=None)` is the interface stub for the ratified pilot-first estimator: tier 1 scales same-template-hash prior actuals by per-shard granule/obs counts; tier 2 falls back to cross-template regression on `n_obs`. It always returns `None` until issues #297/#299 land the history it reads; both Lambda cost blocks now populate `estimated_cost_usd` through it so the wiring point exists. The example notebook is `notebooks/cost_reporting.ipynb` (repo convention places example notebooks under `notebooks/` with a README table row + Binder badge, not `docs/` — see Questions): Binder-runnable, reads only the in-tree `sm_healpix_o9.json` fixture, and shows max → progress → actual with the one non-Binder-able step (a live Lambda invoke) replayed by a clearly-labeled stand-in `agg` that drives the identical wrapper code path. Executed end-to-end locally (all cells green, `max >= actual` asserted in-notebook).

## Phases

- [x] Phase 1 — cost block in the run return: price table + `max_cost_usd` in `dispatch.py`, `RunReport` cost fields, ceiling logged pre-fan-out, `summary["cost"]` on both Lambda paths, tests.
- [x] Phase 2 — notebook wrapper (`src/zagg/notebook.py`): tqdm-or-logging progress from the dispatch result stream, `_repr_html_` report object, CLI confirm gate with `--yes`/`-y`.
- [x] Phase 3 — estimator stub (pilot-first / regression tiers, returns None; deferred behind #297/#299) + example notebook.

## Testing

- `tests/test_dispatch.py::TestMaxCostUsd`: ceiling math vs the price table, memory-variant scaling, arm64 alias pinning, unknown-arch `KeyError`, `RunReport` defaults.
- `tests/test_runner_concurrency.py::TestCostBlock`: mocked `_run_lambda` — cost-block shape, `max >= actual` at the worst billed duration (900 s), worker-variant pricing, `_worker_memory_gb` resolution.
- `tests/test_runner.py::TestSummaryKeysByteIdentical`: pinned Lambda key set extended with the new `cost` key (the one intended summary-shape change).
- `tests/test_notebook.py`: ceiling preview against the in-tree `sm_healpix_o9.json` shardmap (unit count, memory-variant scaling, `max_cells` clamp), tqdm-present and tqdm-absent (`sys.modules` shim) progress paths, `RunView` HTML (cost block, escaping, benign-error exclusion, local no-metered-cost), the non-blocking wrapper (progress driven from a stubbed `agg`, bar closed on raise, `input` unreachable), and the CLI gate (decline aborts pre-`agg` with exit 1, confirm proceeds, `--yes` skips, local backend never gated).
- `on_progress` plumbing: per-completed-shard callback on the mocked spatial Lambda harness (running cost monotone, final == actual rollup) and on the temporal local path (`cost_usd=None`).
- Local: `ruff check` / `ruff format --check` / full `pytest -q` — green except the pre-existing failures noted below.

## Questions for review

1. **Overlap with the PR for #297** (branch `claude/297-stats-sidecar`): both touch `runner.py` and `dispatch.py` (`RunReport`). The cost-block changes here are kept minimal and self-contained; whichever PR merges second rebases.
2. **Timeout constant.** The ceiling prices with `_DEFAULT_FUNCTION_TIMEOUT_S` (900 s, the template default and Lambda hard cap) rather than the live `function_timeout_s`, which is only readable after preflight builds the client — the issue specifies a pre-invoke ceiling, so the constant is used. A function deployed with a shorter Timeout would make the ceiling conservative, never low.
3. **Raster path.** `RasterStrategy._run_lambda_shards` computes no cost today (no `gb_seconds`/`estimated_cost_usd` in its summary), so it gets no cost block in this PR either — adding one there is mechanical if wanted.
4. **Notebook location.** The task sketch said "under docs/", but the repo convention (README "Example Notebooks" table, Binder badges pointing at `lab/tree/notebooks`, `.binder/` config) keeps runnable notebooks in `notebooks/` — followed the convention; happy to move it if docs/ is preferred.
5. Pre-existing failures on clean `main`, not touched per §4: `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` (local docker-dependent build), ruff `N818` on `src/zagg/registry.py:64` (`UnknownCapability`), and format drift in `tests/test_lambda_raster_handler.py`.

https://claude.ai/code/session_019zYV2pvzhVp8cv7e7H4JY3

### DIFF
--- a/src/zagg/dispatch.py
+++ b/src/zagg/dispatch.py
@@ -368,6 +368,12 @@ def estimate_cost_usd(catalog_data=None, *, template_hash=None, history=None) ->
     returns ``None`` and the run's ``cost.estimated_cost_usd`` stays a
     placeholder.
 
+    The temporal event fan-out has no shardmap (its fan-out unit is the
+    event, resolved at call time, not a catalog shard), so
+    ``_run_lambda_events`` wires this stub with no ``catalog_data`` and its
+    estimate stays ``None`` until an event-history tier exists -- out of
+    scope for issue #298; both tiers below key on per-shard counts.
+
     Two tiers, tried in order once wired:
 
     * **Tier 1 -- pilot-first.** Prior actuals for the *same* template hash at
--- a/src/zagg/dispatch.py
+++ b/src/zagg/dispatch.py
@@ -337,7 +337,7 @@ LAMBDA_PRICE_PER_GB_SEC = LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[LAMBDA_ARCH]
 
 
 def max_cost_usd(
-    n_shards: int,
+    n_units: int,
     memory_gb: float,
     *,
     timeout_s: float,
@@ -346,15 +346,16 @@ def max_cost_usd(
     """Hard ceiling on a fan-out's Lambda bill, computable before any invoke.
 
     Every work unit is one Lambda invocation billed at ``memory_gb`` for at
-    most the function timeout, so ``n_shards * rate(arch) * memory_gb *
-    timeout_s`` bounds one clean pass over the shardmap (issue #298).
+    most the function timeout, so ``n_units * rate(arch) * memory_gb *
+    timeout_s`` bounds one invoke per work unit (issue #298). A work unit is
+    a shard on the spatial path and an event on the tabular/temporal path.
     Transient-fault retries (:data:`LAMBDA_RETRY`, rare) re-bill a unit and
     are deliberately not folded in -- the ceiling prices the run as planned,
     not the worst-case retry storm. ``arch`` keys the price table
     (:data:`LAMBDA_PRICE_PER_GB_SEC_BY_ARCH`); an unknown arch raises
     ``KeyError`` rather than pricing silently wrong.
     """
-    return n_shards * LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[arch] * memory_gb * timeout_s
+    return n_units * LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[arch] * memory_gb * timeout_s
 
 
 class LambdaExecutor:
--- a/src/zagg/dispatch.py
+++ b/src/zagg/dispatch.py
@@ -358,7 +358,12 @@ def max_cost_usd(
     return n_units * LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[arch] * memory_gb * timeout_s
 
 
-def estimate_cost_usd(catalog_data=None, *, template_hash=None, history=None) -> float | None:
+def estimate_cost_usd(
+    catalog_data: dict | None = None,
+    *,
+    template_hash: str | None = None,
+    history: list | None = None,
+) -> float | None:
     """Estimated run cost from prior-run history (issue #298 Phase 3 -- stub).
 
     Interface placeholder for the pilot-first estimator ratified on issue
--- a/src/zagg/dispatch.py
+++ b/src/zagg/dispatch.py
@@ -77,6 +77,13 @@ class RunReport:
     counters; cost is accumulated per-result via :meth:`Executor.measure_cost`.
     ``runner.py`` reads this to build the public summary dict and to print cost
     -- this module never formats or prints.
+
+    The run-level cost block (issue #298) is stamped by the runner, not the
+    loop: ``max_cost_usd`` is the pre-invoke ceiling (:func:`max_cost_usd`),
+    ``actual_cost_usd`` the post-run rollup of billed durations, and
+    ``estimated_cost_usd`` the prior-history estimate (``None`` until the
+    issue #297/#299 sidecar history exists). All three stay ``None`` on runs
+    that carry no metered cost (local backend).
     """
 
     results: list[dict] = field(default_factory=list)
@@ -84,6 +91,9 @@ class RunReport:
     cells_error: int = 0
     total_obs: int = 0
     cost: CellCost = field(default_factory=CellCost)
+    max_cost_usd: float | None = None
+    estimated_cost_usd: float | None = None
+    actual_cost_usd: float | None = None
 
 
 # ---------------------------------------------------------------------------
@@ -313,12 +323,38 @@ class LocalExecutor:
         self._pool.shutdown()
 
 
-# arm64 Lambda pricing, $/GB-second, and the function's memory in GB. Matches
-# the constants inlined into ``_run_lambda`` before this extraction; surfaced
-# here so :meth:`LambdaExecutor.measure_cost` and the runner's presentation
-# read one source.
+# Lambda pricing, $/GB-second, keyed by CPU architecture, and the default
+# function memory in GB. The deployed fleet is arm64-only (template.yaml
+# ``Architecture`` default; no x86_64 variant is stood up), so the table
+# carries one row -- an x86_64 deployment adds its rate here (issue #298).
+# ``LAMBDA_PRICE_PER_GB_SEC`` stays the flat arm64 constant so
+# :meth:`LambdaExecutor.measure_cost` and the runner's presentation keep
+# reading one source, byte-identical to the pre-table path.
+LAMBDA_ARCH = "arm64"
+LAMBDA_PRICE_PER_GB_SEC_BY_ARCH = {"arm64": 0.0000133334}
 LAMBDA_MEMORY_GB = 4.0  # issue #193: production worker sized to 4 GB (2.3 vCPU)
-LAMBDA_PRICE_PER_GB_SEC = 0.0000133334
+LAMBDA_PRICE_PER_GB_SEC = LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[LAMBDA_ARCH]
+
+
+def max_cost_usd(
+    n_shards: int,
+    memory_gb: float,
+    *,
+    timeout_s: float,
+    arch: str = LAMBDA_ARCH,
+) -> float:
+    """Hard ceiling on a fan-out's Lambda bill, computable before any invoke.
+
+    Every work unit is one Lambda invocation billed at ``memory_gb`` for at
+    most the function timeout, so ``n_shards * rate(arch) * memory_gb *
+    timeout_s`` bounds one clean pass over the shardmap (issue #298).
+    Transient-fault retries (:data:`LAMBDA_RETRY`, rare) re-bill a unit and
+    are deliberately not folded in -- the ceiling prices the run as planned,
+    not the worst-case retry storm. ``arch`` keys the price table
+    (:data:`LAMBDA_PRICE_PER_GB_SEC_BY_ARCH`); an unknown arch raises
+    ``KeyError`` rather than pricing silently wrong.
+    """
+    return n_shards * LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[arch] * memory_gb * timeout_s
 
 
 class LambdaExecutor:
--- a/src/zagg/dispatch.py
+++ b/src/zagg/dispatch.py
@@ -358,6 +358,45 @@ def max_cost_usd(
     return n_units * LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[arch] * memory_gb * timeout_s
 
 
+def estimate_cost_usd(catalog_data=None, *, template_hash=None, history=None) -> float | None:
+    """Estimated run cost from prior-run history (issue #298 Phase 3 -- stub).
+
+    Interface placeholder for the pilot-first estimator ratified on issue
+    #298; the implementation is deferred behind issues #297 (per-shard stats
+    sidecars) and #299 (template-hash identity), which create the history it
+    reads. Until both land there is nothing to estimate from, so this always
+    returns ``None`` and the run's ``cost.estimated_cost_usd`` stays a
+    placeholder.
+
+    Two tiers, tried in order once wired:
+
+    * **Tier 1 -- pilot-first.** Prior actuals for the *same* template hash at
+      any shard (the dev -> single-shard test -> fleet workflow means a pilot
+      run usually exists), scaled to the remaining shards by per-shard
+      granule/observation counts from the shardmap. One measured shard scaled
+      by relative shard size is a much tighter prior than a global fit.
+    * **Tier 2 -- cross-template regression (fallback).** No same-hash pilot:
+      regress cost on per-shard ``n_obs`` across the sidecar history of other
+      templates.
+
+    Parameters
+    ----------
+    catalog_data : dict, optional
+        The loaded shardmap: per-shard granule/obs counts for the tier-1
+        scaling and the tier-2 regressors.
+    template_hash : str, optional
+        The run's template identity (issue #299) used to select tier-1 priors.
+    history : optional
+        Prior-run sidecar records (issue #297) to estimate from.
+
+    Returns
+    -------
+    float or None
+        ``None`` until the sidecar history exists.
+    """
+    return None
+
+
 class LambdaExecutor:
     """Fan out one synchronous boto3 ``invoke`` per unit (the rich backend).
 
--- a/src/zagg/notebook.py
+++ b/src/zagg/notebook.py
@@ -20,7 +20,7 @@ from __future__ import annotations
 import html
 import logging
 
-from zagg.config import PipelineConfig, get_windowing
+from zagg.config import PipelineConfig, get_pipeline_type, get_store_layout, get_windowing
 from zagg.dispatch import LAMBDA_ARCH, max_cost_usd
 
 logger = logging.getLogger(__name__)
@@ -40,13 +40,30 @@ def max_cost_preview(
 ) -> dict:
     """Resolve the pre-invoke cost ceiling from shardmap + config alone.
 
-    Mirrors ``_run_lambda``'s unit accounting -- cell selection
-    (``max_cells``/``morton_cell``) and the issue #246 windowed-unit expansion
-    -- so the previewed ceiling equals the one the run logs before fan-out.
+    Mirrors the dispatching path's unit accounting -- cell selection
+    (``max_cells``/``morton_cell``) and the windowed-unit expansion -- so the
+    previewed ceiling equals the one the run logs before fan-out. The unit
+    accounting branches on pipeline kind exactly as :func:`~zagg.runner.agg`:
+    the spatial path expands with :func:`~zagg.runner._windowed_units` (issue
+    #246); the ``reader: raster`` re-route mirrors ``RasterStrategy.run``, whose
+    windowed fan-out (:func:`~zagg.runner._raster_windowed_units`, issue #247)
+    fires only on the hive layout and groups by acquisition, so its unit count
+    differs from the spatial expansion for a windowed-raster config. A temporal
+    run has no shardmap ceiling -- its fan-out unit is the ``events=`` item
+    resolved at call time, not a catalog shard -- so this raises ``ValueError``.
     Returns ``{n_units, memory_gb, arch, timeout_s, max_cost_usd}``. No AWS
     access and no grid-signature check: this is display math, not dispatch.
     """
     from zagg import runner
+
+    kind = get_pipeline_type(config)
+    if kind == "temporal":
+        raise ValueError(
+            "temporal runs take events=; no shardmap ceiling (the fan-out unit "
+            "is the event, resolved at call time, not a catalog shard)"
+        )
+    if kind == "spatial" and (config.data_source or {}).get("reader") == "raster":
+        kind = "raster"
 
     catalog_path = catalog or config.catalog
     if not catalog_path:
@@ -54,7 +71,12 @@ def max_cost_preview(
     catalog_data = runner._load_catalog(catalog_path)
     cells = runner._select_cells(catalog_data, morton_cell=morton_cell, max_cells=max_cells)
     windowing = get_windowing(config)
-    if windowing is not None:
+    if kind == "raster":
+        # Mirror RasterStrategy.run: windowed fan-out only on the hive layout;
+        # otherwise one unit per selected cell (cells == _select_cells).
+        if get_store_layout(config) == "hive" and windowing is not None:
+            cells = runner._raster_windowed_units(cells, windowing)
+    elif windowing is not None:
         cells = runner._windowed_units(cells, windowing, (config.bounds or {}).get("temporal"))
     memory_gb = runner._worker_memory_gb(config)
     timeout_s = runner._DEFAULT_FUNCTION_TIMEOUT_S
--- a/src/zagg/notebook.py
+++ b/src/zagg/notebook.py
@@ -269,7 +269,13 @@ def run(config: PipelineConfig, **kwargs) -> RunView:
 
     total = None
     if kwargs.get("events") is not None:
-        total = len(list(kwargs["events"]))
+        # Materialize once and rebind: a one-shot iterable (generator) would be
+        # exhausted by the count, then forwarded empty to agg -- TemporalStrategy's
+        # ``list(events)`` would yield [] and the run would silently process zero
+        # events. List/tuple inputs are unchanged by the round-trip.
+        events = list(kwargs["events"])
+        kwargs["events"] = events
+        total = len(events)
     elif kwargs.get("catalog") or config.catalog:
         preview = max_cost_preview(
             config,
--- a/src/zagg/notebook.py
+++ b/src/zagg/notebook.py
@@ -1,0 +1,289 @@
+"""Notebook-facing dispatch wrapper: progress bar + rich run report (issue #298).
+
+:func:`run` wraps :func:`zagg.runner.agg` with a per-shard progress bar driven
+by the runner's ``on_progress`` callback (one tick per completed work unit,
+with the running Lambda cost in the postfix) and returns a :class:`RunView`
+whose ``_repr_html_`` renders the cost block, run counters, and failures in
+Jupyter. The notebook path **never blocks on cost** (ratified on issue #298):
+the pre-invoke ceiling is displayed, not confirmed. The CLI's yes/no gate
+(:func:`confirm_max_cost`) lives here too and is wired in ``zagg.__main__``
+behind ``--yes``/``-y``.
+
+tqdm is an *optional* import -- it already rides the dependency closure on
+orchestrator installs (earthaccess -> pqdm -> tqdm) but is deliberately not a
+zagg dependency: a bar when importable, a logging fallback otherwise. Workers
+never import this module, so the ``lambda`` extra is untouched.
+"""
+
+from __future__ import annotations
+
+import html
+import logging
+
+from zagg.config import PipelineConfig, get_windowing
+from zagg.dispatch import LAMBDA_ARCH, max_cost_usd
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Pre-invoke ceiling
+# ---------------------------------------------------------------------------
+
+
+def max_cost_preview(
+    config: PipelineConfig,
+    catalog: str | None = None,
+    *,
+    max_cells: int | None = None,
+    morton_cell: str | None = None,
+) -> dict:
+    """Resolve the pre-invoke cost ceiling from shardmap + config alone.
+
+    Mirrors ``_run_lambda``'s unit accounting -- cell selection
+    (``max_cells``/``morton_cell``) and the issue #246 windowed-unit expansion
+    -- so the previewed ceiling equals the one the run logs before fan-out.
+    Returns ``{n_units, memory_gb, arch, timeout_s, max_cost_usd}``. No AWS
+    access and no grid-signature check: this is display math, not dispatch.
+    """
+    from zagg import runner
+
+    catalog_path = catalog or config.catalog
+    if not catalog_path:
+        raise ValueError("No catalog specified (pass catalog= or set catalog: in config)")
+    catalog_data = runner._load_catalog(catalog_path)
+    cells = runner._select_cells(catalog_data, morton_cell=morton_cell, max_cells=max_cells)
+    windowing = get_windowing(config)
+    if windowing is not None:
+        cells = runner._windowed_units(cells, windowing, (config.bounds or {}).get("temporal"))
+    memory_gb = runner._worker_memory_gb(config)
+    timeout_s = runner._DEFAULT_FUNCTION_TIMEOUT_S
+    return {
+        "n_units": len(cells),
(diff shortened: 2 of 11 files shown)


src/zagg/runner.py
```diff
@@ -1511,11 +1542,22 @@ def _accumulate(report, i, result):
...
     total_lambda_time = report.cost.compute_time_s
-    gb_seconds = total_lambda_time * LAMBDA_MEMORY_GB
+    gb_seconds = total_lambda_time * memory_gb
     estimated_cost = gb_seconds * LAMBDA_PRICE_PER_GB_SEC
 
+    # Structured cost block (issue #298), mirroring the spatial summary; the
+    # legacy flat ``estimated_cost_usd`` key keeps the actual-cost rollup.
+    report.max_cost_usd = run_max_cost
+    report.actual_cost_usd = estimated_cost
+    cost_block = {
```
espg, line 1625: 🤖 *from Claude (review)*

[test-gap] The events path (`_run_lambda_events`) got its own copy of the phase-1 logic — the pre-invoke ceiling log, `memory_gb = _worker_memory_gb(config)` feeding the `LambdaExecutor` (line ~1483), the `gb_seconds = total_lambda_time * memory_gb` rollup, and this `cost_block` — but nothing tests it. `TestCostBlock` in `tests/test_runner_concurrency.py` only drives `_run_lambda` (spatial), and `TestSummaryKeysByteIdentical._LAMBDA_KEYS` pins only the spatial summary's key set, not the tabular events summary. So a regression on this side — dropping the `cost` key, or reverting `memory_gb` back to the flat `LAMBDA_MEMORY_GB` in the events rollup — would ship green. The PR body's own testing section confirms only `_run_lambda` is mocked.

Suggested fix: add an events-path case (mock `_run_lambda_events` the way `TestCostBlock._run` mocks `_run_lambda`) asserting the `cost` block shape, that `estimated_cost_usd` is `None`, and that a `worker: {memory: 8192}` config bills 8 GB in both `gb_seconds` and `max_cost_usd`.

src/zagg/runner.py
```diff
@@ -2350,6 +2393,18 @@ def _run_lambda(
     if windowing is not None:
         cells = _windowed_units(cells, windowing, (config.bounds or {}).get("temporal"))
 
+    # Pre-invoke cost ceiling (issue #298): every unit is one invoke billed at
```
espg, line 2476: 🤖 *from Claude (review)*

[question] The ceiling is computed here, but this block sits *after* the `if dry_run: return _dry_run_summary(...)` early return a few lines up (line ~2386). `_dry_run_summary` (line 1944) returns only `total_cells`/granule stats — no cost. So `agg(..., dry_run=True)`, the one path a user runs specifically to preview a fan-out before spending, never logs or returns the max-cost ceiling — the single number that's fully computable without any invoke and is the stated point of #298 ("max computed pre-invoke... print max before fan-out"). It's surfaced only on the real-spend path.

Is that intended for phase 1 (ceiling deferred to the phase-2 CLI gate), or should the dry-run summary carry `max_cost_usd` too? If deferred, worth a one-line note; if not, hoisting the `_worker_memory_gb`/`max_cost_usd` call above the `dry_run` return and adding it to `_dry_run_summary` is mechanical. (Note the windowed unit expansion also happens after the dry_run return, so a dry-run ceiling would need that accounted for.)

src/zagg/dispatch.py
```diff
@@ -313,12 +323,38 @@ def shutdown(self) -> None:
...
+# reading one source, byte-identical to the pre-table path.
+LAMBDA_ARCH = "arm64"
+LAMBDA_PRICE_PER_GB_SEC_BY_ARCH = {"arm64": 0.0000133334}
 LAMBDA_MEMORY_GB = 4.0  # issue #193: production worker sized to 4 GB (2.3 vCPU)
-LAMBDA_PRICE_PER_GB_SEC = 0.0000133334
+LAMBDA_PRICE_PER_GB_SEC = LAMBDA_PRICE_PER_GB_SEC_BY_ARCH[LAMBDA_ARCH]
+
+
+def max_cost_usd(
+    n_shards: int,
```
espg: 🤖 *from Claude (review)*

[docs] The parameter is named `n_shards`, but both call sites pass something else: `_run_lambda_events` passes the event count (`n = len(event_list)`), and `_run_lambda` passes the post-windowing unit count (`len(cells)` after `_windowed_units`, i.e. shard×window pairs, not shards). The runner comments consistently call these "units"/"events". Since this is a public helper imported by tests, the misnomer will mislead callers into thinking it only counts shards.

Suggested fix: rename the param to `n_units` (or `n_invokes`) and adjust the docstring's "one clean pass over the shardmap" wording to "one invoke per work unit" — purely cosmetic, no behavior change.

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed in 07942f6. Added `TestCostBlockEvents` in tests/test_runner_concurrency.py, mirroring `TestCostBlock` but driving `_run_lambda_events` (mocked `compute_available_workers`, `_invoke_lambda_event`, `_write_tabular_output`; 3 events x 1 s billed, sync invocation, no live AWS). It asserts the `cost` block shape (`max_cost_usd` == `max_cost_usd(3, 4.0, timeout_s=900)`, `estimated_cost_usd is None`, `actual_cost_usd` == the legacy `estimated_cost_usd` rollup), that the ceiling bounds the actual even at the 900 s worst case, and that a `worker: {memory: 8192}` config bills 8 GB in both `gb_seconds` (3 x 1 s x 8 GB) and doubles `max_cost_usd` -- so a revert to flat `LAMBDA_MEMORY_GB` on the events rollup now fails.

src/zagg/runner.py
espg: 🤖 *from Claude*

Leave-by-design (with a one-line note added in bdbd7ce). Deferred to the phase-2 CLI gate, for the windowing reason you flagged: `_windowed_units` expands the shard list into (shard, window) units *after* the dry-run return, so a ceiling computed on the pre-windowing `len(cells)` here would undercount temporal/windowed runs -- the wrong direction for something meant to be an upper bound. A correct dry-run ceiling has to expand units first, which means either reordering `_windowed_units` above the dry-run return (changing `total_cells` semantics for windowed dry runs) or duplicating the windowing expansion just for the ceiling. Phase 2 adds a precompute helper that expands units and owns the dry-run ceiling, so the honest fix lands there rather than half-done here. Added a comment at the dry-run return in src/zagg/runner.py recording exactly this so the omission is intentional, not an oversight.

src/zagg/dispatch.py
espg: 🤖 *from Claude*

Fixed in 15a086e. Renamed the `max_cost_usd` param `n_shards` -> `n_units` in src/zagg/dispatch.py and reworded the docstring from "one clean pass over the shardmap" to "one invoke per work unit", with an added line noting a work unit is a shard on the spatial path and an event on the tabular/temporal path. No behavior change: both call sites (`_run_lambda`, `_run_lambda_events`) and the tests pass the count positionally, so nothing else needed touching.

src/zagg/notebook.py
```diff
@@ -0,0 +1,289 @@
...
+    path never prompts, per issue #298; the blocking gate is CLI-only). The
+    bar total is the previewed unit count for catalog runs and ``len(events)``
+    for temporal runs; when neither is resolvable the bar degrades to an
+    unsized tqdm (or the logging fallback).
+    """
+    from zagg import runner
+
+    total = None
+    if kwargs.get("events") is not None:
+        total = len(list(kwargs["events"]))
```
espg: 🤖 *from Claude (review)*

[bug] `run()` exhausts a one-shot `events` iterable before dispatch. `total = len(list(kwargs["events"]))` materializes the iterable purely to count it, but then the *same* object is forwarded to `agg(...)` via `**kwargs`. `TemporalStrategy` does `event_list = list(events)` on it — for a generator/iterator (which `agg`'s docstring and TemporalStrategy both accept: "an iterable of ... tuples") that second `list()` yields `[]`, so the run silently processes **zero** events and reports success. Only `list`/`tuple` inputs survive.

Fix: materialize once and rebind, e.g.
```python
if kwargs.get("events") is not None:
    events = list(kwargs["events"])
    kwargs["events"] = events
    total = len(events)
```
A test passing a generator to `notebook.run(..., events=(e for e in ...))` and asserting the callback fires per event would lock it in (the current `test_events_run_sizes_bar_from_event_count` passes a list, so it misses this).

src/zagg/notebook.py
```diff
@@ -0,0 +1,289 @@
...
+    from zagg import runner
+
+    catalog_path = catalog or config.catalog
+    if not catalog_path:
+        raise ValueError("No catalog specified (pass catalog= or set catalog: in config)")
+    catalog_data = runner._load_catalog(catalog_path)
+    cells = runner._select_cells(catalog_data, morton_cell=morton_cell, max_cells=max_cells)
+    windowing = get_windowing(config)
+    if windowing is not None:
+        cells = runner._windowed_units(cells, windowing, (config.bounds or {}).get("temporal"))
```
espg, line 80: 🤖 *from Claude (review)*

[bug] The pre-invoke ceiling assumes **spatial** unit accounting, but silently applies it to raster and temporal runs, so the previewed/gated cost can mismatch the actual fan-out.

`max_cost_preview` always calls the *spatial* `runner._windowed_units` (line 58). The raster path fans out over `runner._raster_windowed_units` instead (`RasterStrategy.run`, runner.py:~730), whose membership rules are entirely different — it groups by `time_key`, places each acquisition group in exactly one window, and drops asset-less / out-of-window groups (see the `_raster_windowed_units` docstring). For a **windowed raster** config the two produce different `(shard, window)` counts, so:
- the CLI gate (`__main__.py:114`) prints/prompts a ceiling with the wrong unit count, and
- `notebook.run()`'s bar total (`preview["n_units"]`) is wrong for a windowed-raster catalog run (non-windowed raster is fine — cells == `_select_cells`).

For a **temporal** config the mismatch is total: units are events built from specs, not `_select_cells` shards. Via the CLI this is also a dead-end — the gate computes a shard-based ceiling and prompts the user, then `agg` immediately raises `temporal pipeline requires events=` (the CLI has no events source).

The runner already routes on `get_pipeline_type(config)` (+ the `reader == "raster"` override) in `agg`. Suggest `max_cost_preview` branch the same way: use `_raster_windowed_units` for raster, and either skip the gate for temporal or size it from the event count. At minimum, guard the CLI gate so it only fires for spatial/raster kinds and document the raster-window caveat.

src/zagg/runner.py
```diff
@@ -93,6 +95,42 @@ def _resolve_function_name(config: PipelineConfig, function_name: str | None) ->
...
+    ``cost_usd`` is the report's running metered rollup on Lambda paths
+    (``metered=True``) and ``None`` where nothing is billed. ``None`` hook ->
+    the accumulator is returned untouched, keeping the default path identical.
+    """
+    if on_progress is None:
+        return accumulate
+
+    def wrapped(report, i, result):
+        accumulate(report, i, result)
+        on_progress(i, total, report.cost.cost_usd if metered else None)
```

[question] An `on_progress` that raises aborts a *paid* run mid-fan-out. `_with_progress.wrapped` calls `on_progress(...)` with no guard, and `dispatch`'s `as_completed` loop calls `accumulate` (the wrapped fn) directly — so an exception from the callback propagates out of `dispatch`, unwinds `_run_lambda`, and drops the summary. By then the async Lambda fan-out has already been invoked (money spent) and the store may be partially written, yet the caller gets an exception instead of the `RunView`. The `agg` docstring even states the hook "must not raise" — but nothing enforces it, and the notebook's own sink (`_TqdmProgress.update` → tqdm `refresh`/`set_postfix`) can raise in odd/headless terminals.

A progress display is cosmetic; its failure should not kill a run that has already cost money. Consider guarding the callback here, e.g.
```python
def wrapped(report, i, result):
    accumulate(report, i, result)
    if on_progress is not None:
        try:
            on_progress(i, total, report.cost.cost_usd if metered else None)
        except Exception:
            logger.debug("on_progress hook raised; ignoring", exc_info=True)
```
(counters/results are already folded before the hook runs, so swallowing here loses only the tick). If the "must not raise" contract is intended to be strict instead, a test that a raising hook is *not* swallowed would document that choice.

src/zagg/runner.py
```diff
@@ -859,6 +917,10 @@ def _write_slab(t_idx, slab):
                     last_error = e
                     logger.warning(f"raster shard {label} failed: {e}")
                     continue
+                finally:
+                    # Progress hook (issue #298): local raster has no metered cost.
+                    if on_progress is not None:
+                        on_progress(i, len(cells), None)
```

[test-gap] The raster `on_progress` plumbing is untested. This commit adds `enumerate(as_completed(...), 1)` + an `on_progress(i, len(cells), None)` tick to **both** raster loops — `_run_local_shards` (this line, inside a `finally`) and `_run_lambda_shards` (runner.py:1217) — but no test exercises either. `grep -rn on_progress tests/` shows coverage only for the spatial-lambda path (`test_runner_concurrency.py:198`) and temporal-local path (`test_runner.py:2535`); nothing asserts the raster ticks fire, are 1-based/monotone, carry `cost_usd=None`, or (for the local loop's `finally`) still tick on a failed shard.

The raster local tick sits in a `finally` after a bare-`except … continue`, so its correctness on the failure path is exactly the kind of thing a test should pin. Add a raster progress test mirroring the spatial/temporal ones (a mocked shard harness with one failing shard, asserting the recorded `(done, total, cost)` sequence).

src/zagg/notebook.py
espg: 🤖 *from Claude*

Fixed in 07a1d67. `run()` now materializes the iterable once and rebinds it: `events = list(kwargs["events"]); kwargs["events"] = events; total = len(events)` — so the count and the object forwarded to `agg` are the same materialized list, and a generator no longer arrives empty at `TemporalStrategy`'s `list(events)`. Added `TestRunWrapper.test_events_generator_not_exhausted_by_count` in tests/test_notebook.py: passes a one-shot generator of 3 events through a faked `agg` that does `list(events)`, and asserts both the bar total is 3 AND `agg` saw all 3 events (the pre-fix code would show total 3 but an empty `agg` events list). List/tuple inputs are unchanged by the round-trip.

src/zagg/notebook.py
espg: 🤖 *from Claude*

Fixed in cacaa20. `max_cost_preview` now branches on pipeline kind exactly as `agg` does (`get_pipeline_type` + the `reader == "raster"` re-route):

- **spatial** → `_windowed_units` (unchanged).
- **raster** → mirrors `RasterStrategy.run`: windowed fan-out via `_raster_windowed_units` **only** on the hive layout (`get_store_layout(config) == "hive" and windowing is not None`); otherwise one unit per selected cell (`cells == _select_cells`). Matches the runner exactly — the flat/non-windowed raster path there never expands.
- **temporal/event** → raises `ValueError("temporal runs take events=; no shardmap ceiling ...")`.

And the `__main__` gate now only engages for the catalog-driven kinds: added `and get_pipeline_type(config) != "temporal"` to the gate condition, so a temporal config prints no ceiling and goes straight to `agg` (which raises its own `requires events=` for a CLI run) rather than computing a meaningless shard-based prompt.

Tests (tests/test_notebook.py): `test_temporal_config_has_no_shardmap_ceiling` (preview raises), `test_raster_nonwindowed_counts_selected_cells` (raster flat counts cells; monkeypatched `_windowed_units` asserts the spatial expansion never runs), `test_windowed_raster_uses_raster_units` (hive+windowing routes to `_raster_windowed_units`, not `_windowed_units`), and `TestCliGate.test_temporal_backend_skips_gate` (no prompt, no ceiling printed, agg still called).

src/zagg/runner.py

Fixed in 24c82d1. Guarded the callback everywhere it fires: `_with_progress.wrapped` now wraps `on_progress(...)` in `try/except Exception` that logs a warning (`logger.warning("on_progress hook raised; ignoring", exc_info=True)`) and continues; the two raster inline ticks (`_run_local_shards` finally, `_run_lambda_shards`) got the same guard. Counters/results are folded before the hook runs, so swallowing loses only the tick. Softened the `agg` docstring: dropped "must not raise", kept "must be cheap", and noted a raising hook is caught and logged rather than propagated — a cosmetic display must not unwind a paid fan-out mid-flight.

Test: `TestTemporalStrategy.test_raising_on_progress_hook_does_not_unwind_run` in tests/test_runner.py drives a local run with a hook that raises `RuntimeError` and asserts the run completes and still returns its summary (`total_events == 2`, `events_with_data == 2`). This exercises the `_with_progress` guard directly; the raster inline guards are exercised by the raster on_progress tests added for the test-gap finding.

src/zagg/runner.py

Fixed in 59bc303. Added raster on_progress coverage in tests/test_raster_runner.py:

- **Local loop** — `TestRasterAgg.test_local_on_progress_ticks_including_failed_shard`: a two-shard flat-layout run where shard1's lone granule carries an asset map lacking the configured band (so `sample_item_async` raises for that shard only, while shard0 writes). Asserts the recorded `(done, total, cost)` sequence is `[1, 2]` (1-based, monotone), every `total == 2` (the runner's authoritative count), every `cost is None`, and — the key point for the `finally` — that the FAILED shard still ticked (`cells_error == 1`, `cells_with_data == 1`, both ticks present).
- **Lambda loop** — `TestRasterLambdaBackend.test_on_progress_ticks_on_lambda_fanout`: reuses the existing `_FakeLambdaClient` + `_lifecycle` mocked harness (`invocation="sync"`), asserting the fan-out ticks `[(1, 1, None)]` — the single shard fixture ticks once, and the ping/setup lifecycle invokes never tick (only the `process_raster` fan-out does).

The lambda harness exists, so both loops are covered. The `_run_lambda_shards` tick uses the single-shard `manifest` fixture (one tick); the local test carries the multi-shard + failure case.

src/zagg/runner.py
```diff
@@ -1511,11 +1610,24 @@ def _accumulate(report, i, result):
...
     total_lambda_time = report.cost.compute_time_s
-    gb_seconds = total_lambda_time * LAMBDA_MEMORY_GB
+    gb_seconds = total_lambda_time * memory_gb
     estimated_cost = gb_seconds * LAMBDA_PRICE_PER_GB_SEC
 
+    # Structured cost block (issue #298), mirroring the spatial summary; the
+    # legacy flat ``estimated_cost_usd`` key keeps the actual-cost rollup.
+    # estimate_cost_usd is the Phase-3 stub (None until #297/#299 land).
+    report.max_cost_usd = run_max_cost
+    report.estimated_cost_usd = estimate_cost_usd()
```
espg, line 1623: 🤖 *from Claude (review)*

[question] The temporal `_run_lambda_events` wires the estimator as `estimate_cost_usd()` with no `catalog_data`, but a temporal run has no shardmap at all — `max_cost_preview` raises `ValueError` for `kind == "temporal"` ("the fan-out unit is the event, resolved at call time, not a catalog shard"), and both documented tiers key on per-shard granule/obs counts. So the stub's documented interface (`catalog_data` = "the loaded shardmap") can never be satisfied at this call site; when #297/#299 land, this path is a structural `None` regardless of history.

Why it matters: it leaves the deferred implementer guessing whether the pilot-first estimator is meant to cover the temporal event fan-out at all. Benign now (returns `None`), but worth resolving the intent before the impl.

Concrete fix: either drop the temporal wiring (leave `estimated_cost_usd` as the `RunReport` default here) or add one line to the `estimate_cost_usd` docstring noting the temporal event fan-out is out of scope (no shardmap → no per-shard scaling), so the #297/#299 author do’t try to fit it.

espg, line 1623: 🤖 *from Claude (review)*

[question] The temporal `_run_lambda_events` wires the estimator as `estimate_cost_usd()` with no `catalog_data`, but a temporal run has no shardmap at all: `max_cost_preview` raises `ValueError` for `kind == "temporal"` (the fan-out unit is the event, resolved at call time, not a catalog shard), and both documented tiers key on per-shard granule/obs counts. So the documented interface (`catalog_data` = the loaded shardmap) can never be satisfied at this call site; when #297/#299 land, this path stays a structural `None` regardless of history.

Why it matters: it leaves the deferred implementer guessing whether the pilot-first estimator is meant to cover the temporal event fan-out at all. Benign now (returns `None`), but worth resolving the intent before the impl.

Concrete fix: either drop the temporal wiring (leave `estimated_cost_usd` at the `RunReport` default here) or add one line to the `estimate_cost_usd` docstring noting the temporal event fan-out is out of scope (no shardmap, no per-shard scaling), so the #297/#299 author does not try to fit it.

src/zagg/dispatch.py
```diff
@@ -313,12 +323,78 @@ def shutdown(self) -> None:
...
+      granule/observation counts from the shardmap. One measured shard scaled
+      by relative shard size is a much tighter prior than a global fit.
+    * **Tier 2 -- cross-template regression (fallback).** No same-hash pilot:
+      regress cost on per-shard ``n_obs`` across the sidecar history of other
+      templates.
+
+    Parameters
+    ----------
+    catalog_data : dict, optional
+        The loaded shardmap: per-shard granule/obs counts for the tier-1
```
espg, line 396: 🤖 *from Claude (review)*

[question] Signature-fit for the future tiers — two gaps the deferred impl will hit, worth capturing while the interface is being set:

1. **Cell selection is not threaded.** The spatial call passes the full `catalog_data` but not the selected cell set (`max_cells`/`morton_cell`). Tier 1 says "scaled to the remaining shards" — but `_run_lambda` may run a strict subset of `catalog_data["shard_keys"]`, and a single-shard pilot run (the exact dev → single-shard test → fleet workflow this estimator targets) is itself a `max_cells`/`morton_cell` run. Estimating from all shards when only some run would mis-scale.
2. **No target run config.** Turning a prior actual into a dollar estimate at a *different* worker-memory variant (issue #235: 2/4/8 GB) needs the target run’s `memory_gb`/`arch`/`timeout` — none are in the signature (no config passed). A pilot billed at 4 GB scaled by obs count alone would misprice an 8 GB fleet run.

Minor: the docstring types the params (`catalog_data : dict`, `template_hash : str`) but the signature leaves them untyped, unlike the fully-typed `max_cost_usd` right above.

Why it matters: none of this blocks the stub (it returns `None`), but the wired call sites bake in an interface that #297/#299 will likely have to widen. Concrete fix: either pass `cells`/`config` (or a small resolved context) now, or note in the docstring that the selected-unit set and target-run pricing are still to be added.

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed with the small doc option in 31e069e. Kept the `estimate_cost_usd()` call at the temporal site (so the `estimated_cost_usd` field stays a uniform stub-wired placeholder at both call sites) and added a docstring paragraph to `estimate_cost_usd` in src/zagg/dispatch.py recording exactly this: the temporal event fan-out has no shardmap (its fan-out unit is the event, resolved at call time, not a catalog shard), so `_run_lambda_events` wires the stub with no `catalog_data` and its estimate stays `None` until an event-history tier exists -- out of scope for issue #298, since both tiers key on per-shard counts. So the deferred implementer sees the intent stated in the stub itself rather than having to infer it from the call site.

src/zagg/dispatch.py
espg: 🤖 *from Claude*

Partly fixed, partly leave-by-design.

**(c) types — fixed in 5ac97a8.** Annotated the stub signature to match the docstring: `catalog_data: dict | None = None, *, template_hash: str | None = None, history: list | None = None`, consistent with the fully-typed `max_cost_usd` right above. Pure annotation, no behavior change (still returns `None`); both call sites (`estimate_cost_usd(catalog_data)` spatial, `estimate_cost_usd()` temporal) typecheck.

**(a) cell selection + (b) target-run pricing — leave-by-design, recorded for the deferred impl.** Both are real and correct, but the fix is to *widen the signature with inputs no code reads yet* (selected shard keys, `memory_gb`, `arch`). Adding defaulted-`None` keyword params to a stub that returns `None` is a speculative interface with no consumer — exactly the dead abstraction §4 asks us not to bake in, and as you note the signature is expected to evolve when #297/#299 land and create the history it reads. The honest place to finalize the shape is alongside the estimator logic that actually uses `cells`/`config`, where the resolved-context vs loose-kwargs call can be decided against real use. So the finding stands recorded here for the #297/#299 implementer: thread the selected-unit set (max_cells/morton_cell) so a single-shard pilot does not mis-scale from all shards, and pass the target run pricing identity (memory_gb/arch/timeout, issue #235) so a pilot billed at one variant reprices correctly at another.

src/zagg/runner.py
espg: 🤖 *from Claude*

Duplicate of the thread below (same finding, posted twice by the review pass). Addressed in 31e069e: the temporal call is kept so both Lambda cost blocks stay uniformly stub-wired, and the `estimate_cost_usd` docstring now states the temporal event fan-out has no shardmap (its fan-out unit is the event), passes no `catalog_data`, and stays `None` until an event-history tier exists — out of scope for issue #298.

espg:
🤖 *from Claude*

Rolled into PR #302 as directed:

- `main` (with this PR's cost machinery) is integrated into `claude/297-stats-sidecar` via merge commit b3fec5f — a merge rather than a rebase, since pushed history is never rewritten per the repo rules; same integration effect.
- 4141ee9 adds the #298 cost block to the **raster** Lambda summary (`cost: {max_cost_usd, estimated_cost_usd, actual_cost_usd}` + `gb_seconds`/`price_per_gb_sec`, legacy flat `estimated_cost_usd` = actual rollup; failed units' landed durations are billed into the rollup), and makes the per-shard stats record price through the same arch-keyed table so sidecar and summary agree on the rate.
- The **temporal** path already carries the cost block from this PR itself (`_run_lambda_events` got the ceiling + structured block in phase 1 here) — verified on the merged code, nothing further to add. Its `estimated_cost_usd` stays the documented `None` stub (no shardmap to estimate from) and its actual rollup prices from the report's billed durations, so nothing is faked.
- Local backends (spatial and raster alike) remain unmetered by design — no cost block there.

Details and shas also on the PR #302 thread; CI running there.

Issue #305, opened 2026-07-19
Morton convention spec: normative encoding, distinct dggs name, self-declared UUID (implements #262 / D16)


espg:
🤖 *from Claude*

Implements the spec half of #262 (D16): morton as a real, declared convention — "not just 'our proxy knows'" — which the ratified public-bucket goal (browser-direct source.coop example) requires.

## Scope

1. **mortie spec page** ([mortie#62](https://github.com/espg/mortie/issues/62), still open): normative bit layout, resolution/Δ table, frozen-convention statement; the order-29 point-id/area-word parse non-injectivity note; the D16 order-29→24 clip rule for point-encoded ids (NESTED ids are float64-exact only through order 24). The morton-hive leaf-naming grammar is already recorded there (D13); this completes the page for v1.0.
2. **zagg conventions block**: the dggs attrs entry for morton-declared stores — distinct grid **`name: "morton"`** (rationale recorded in D16: scheme-blind readers must hard-reject, not silently misread; matches moczarr's xdggs `grid_name`), `coordinate: "morton"`, and the **O10 discriminator** `resolution: "exact" | "point"` (proposal — needs espg sign-off).
3. **Convention identity**: an interim **self-declared convention UUID** via the zarr-conventions mechanism (the schema supports self-declared entries), so morton stores are identifiable now; the upstream dggs-registry ask (#72 ask 3) is thereby *required-not-nice-to-have* for the public-store path, but not a blocker for shipping the self-declared form.
4. Registry-spec consumers: moczarr and gridlook (espg/gridlook#1 Phase 6 detector branch + TS BigInt decode, decimal-string render form per D1) reference this entry rather than private knowledge.

## Decisions for espg

1. **O10 field**: name (`resolution`?), values (`"exact" | "point"`), placement (dggs attrs block vs manifest), default (`"point"` for raw conversions). Proposal as stated in the doc amendment.
2. **Interim self-declared UUID** now vs waiting for an upstream zarr-conventions/dggs registry entry.
3. mortie#62 page content sign-off once drafted (it lives in the mortie repo; drafting can start immediately).

espg:
🤖 *from Claude*

[...]

Issue #304, opened 2026-07-19
Morton-only writer flip: stop writing cell_ids (implements #262 / D16)


espg:
🤖 *from Claude*

Implements the writer half of #262 (D16 — morton-only storage; doc amendment PR carries the decision record). The moczarr fabrication layer — the named gate — is merged (espg/moczarr PR #7), so this is unblocked.

## Context: the flip is smaller than #262's sketch implies

The writer **already always emits the `morton` array** — `coords_of` returns both `morton` and `cell_ids` (`src/zagg/grids/healpix.py:423-426`). `cell_ids` is the redundant coordinate, and the #135 knob (`cell_ids_encoding: morton`) currently stores morton words *inside* `cell_ids` with `name: "healpix"` + `indexing_scheme: "morton"` attrs (`healpix.py:669-685`) — exactly the scheme-blind silent-misread hazard the #262 verification identified. So the work is: stop writing `cell_ids`, re-point the convention attrs at `morton` under a **distinct grid `name: "morton"`** (matching moczarr's xdggs registration — unknown-name readers hard-reject with a diagnostic instead of rendering garbage), and regenerate fixtures.

## Phases (0.x sequencing per D16)

1. **Attrs + discriminator**: dggs block becomes `name: "morton"`, `coordinate: "morton"`, plus the O10 `resolution: "exact" | "point"` field (default `"point"` for raw lat/lon conversions) — emitted whenever the store is morton-declared. Signatures (`grid.signature()`, `cell_ids_encoding` field) updated coherently.
2. **Default flip**: `cell_ids` no longer written by default; an `emit_cell_ids: true` escape hatch remains for transition stores (gridlook demos until the hub-proxy fabrication / TS decode lands — espg/gridlook#1 Phases 4/6).
3. **Removal**: drop NESTED emission and retire the `cell_ids_encoding` knob (#135 inverts: morton stored, NESTED derived). Fixtures/goldens regenerate; bench object counts update (one fewer u64 array per leaf — the #236/#240 currency moves favorably).

## Sequencing

Lands in the **same break window as #299** (leaf rename): moczarr#11 already tracks the #299 adoption, and extending it to cover both changes means moczarr's golden parity vectors regenerate **once**. gridlook Phase 6 then targets the final layout once.

## Decisions for espg

1. Confirm the phase 2 escape hatch (`emit_cell_ids`) vs going straight to removal — the hatch exists only to keep browser-direct demo stores viewable until the gridlook morton path lands.
[...]

Issue #301, opened 2026-07-19
Cost/provenance visualization: gridlook choropleth, duckdb, Grafana


espg:
🤖 *from Claude*

Sub-issue of #296 — slice (5): surface the cost/provenance record spatially (gridlook) and tabularly (duckdb now, Grafana later). Depends on the sidecar sub-issue for data; nothing here blocks the other slices.

## Context

**Spatial (gridlook) — the MVP needs zero new viewer code.** The gridlook Phase 3 branch already has property-driven vector choropleth ("e.g. zagg shard outlines colored by granule_count" is the code comment in `src/lib/layers/vectorChoropleth.ts`), hover readout of feature properties, and deep-linkable layer params (espg/gridlook#1). Emitting shard outlines as GeoJSON with `est_cost_usd` / `total_obs` / `duration_s` properties gives cost-on-the-globe immediately. Native sidecar reading belongs in gridlook Phase 6 (morton-hive awareness), where the viewer at zoom N reads the level-N stats rollup — cost pyramids fall out of the same sweep as data pyramids (unified-sweep sub-issue).

**Tabular — duckdb over the run parquet first.** Same pattern as xopr (github.com/englacial/xopr) and the STAC catalog: no standing infrastructure, notebook-native, covers ~90% of the reporting need. Grafana is the upgrade path, not the starting point.

**Grafana prior art (2i2c).** Their stack: dedicated IAM role → small Python server querying the AWS Cost Explorer API (tag-based attribution) → JSON via Grafana's Infinity plugin (`2i2c-org/jupyterhub-cost-monitoring` + their infrastructure repo). Two takeaways for zagg:

- **Two complementary cost layers**: self-reported per-shard GB-s estimates are immediate and spatial but are estimates; Cost Explorer with cost-allocation tags on the Lambda functions/variants is authoritative but ~24 h delayed and spatially blind. Use sidecars for the map, Cost Explorer for reconciliation. Requires activating cost-allocation tags on the fleet account.
- **Hosting**: zagg is serverless and has no cluster to run Grafana on; 2i2c's runs on hub k8s. Since gridlook-jupyter already targets cryocloud (2i2c-operated), the natural home is cryocloud's existing Grafana — a conversation to have with 2i2c before building anything.

## Related design points from #296

- The target bucket is cryocloud-viewable, not public — but sidecars should still omit account-identifying fields (request ids, function ARNs); they add nothing to cost/provenance and already live in `RunReport`/CloudWatch.
[...]

Issue #300, opened 2026-07-19
Unified second-pass rollup sweep: stats, shardmap sub-maps, MOC, overviews


espg:
🤖 *from Claude*

Sub-issue of #296 — slice (4): one second-pass sweep that owns every derived, up-tree artifact. Extends #201's scope; #201's open design questions (per-field overview semantics, level schedule, orchestration, trigger) remain the discussion of record there.

## Context

Four artifact families all share the same shape — derived from leaf truth, mergeable up the morton prefix tree, regenerable, never load-bearing (the D9 cache principle from `docs/design/sparse_coverage.md`):

1. **Overview/pyramid zarrs** — already settled as a second-pass sweep on #201 (never write-time; aggregates across worker-shard boundaries).
2. **MOC regeneration** — already sweep-owned per #201.
3. **Shardmap sub-map rollups** — `ShardMap.reproject` (issue #294) makes coarsen an *exact* pure regroup via `mortie.clip2order` (granule union, deduped). Place a sub-map at each leaf prefix, fold it up each level of the tree; #296's composability observation made concrete.
4. **Stats/cost rollups** — the per-shard sidecars (sidecar sub-issue) are mergeable by construction (sums/counts/min/max/t-digests), so the rollup is the same fold as the pyramid.

Unifying these in one sweep is the cohesion #296 asks for: one idempotent pass, one orchestration answer, one staleness story.

## Design points

- **Idempotent and non-load-bearing**: the sweep can fail or lag without corrupting anything; readers must never *require* a rollup.
- **Staleness is detectable, not prevented**: after a leaf re-run, interior rollups are stale until the sweep re-runs. Stamp every rollup with generation info (merged-leaf count + max leaf timestamp) so consumers can detect it; regenerate opportunistically.
- **Rollup == direct**: for every artifact family, the folded result at level N must equal direct computation at level N (the #294 coarsen-exactness property, generalized) — this is the core test obligation.

espg:
🤖 *from Claude*

## Implementation plan

Ordered cheapest-first so the sweep skeleton is proven on pure-metadata folds before it touches data.

[...]

PR #306, opened 2026-07-19
sparse_coverage doc amendment: D16 morton-only storage + O10 discriminator

### LINKED ISSUE #262: Design: morton-only storage — drop stored NESTED cell_ids ("NESTED is fabricated, never stored"; pre-1.0)
🤖 *from Claude*

# Design: morton-only storage — stop writing NESTED `cell_ids`; "NESTED is fabricated, never stored"

**espg-ratified in-session (2026-07-17); filed for the record and for the `sparse_coverage.md` amendment.** Proposed as a new decision (next free D-number) amending §6's "cell_ids (HEALPix NESTED) remains the interop encoding" — the interop encoding remains NESTED, but it becomes a **derived view fabricated on demand, never stored**. Timing ratified as **now, not soon** — this is a pre-1.0 (0.x) breaking store change, same logic as #251: shipping 1.0 with a coordinate we're about to delete costs a 2.0.

## The decision

- zagg stops writing the `cell_ids` (NESTED uint64) array to leaves. `morton` (packed mortie words) becomes the only stored cell coordinate and the declared convention coordinate.
- Any consumer that wants NESTED gets it **fabricated exactly, on demand** (`mort2healpix` is vectorized arithmetic): moczarr fabricates it Python-side; the gridlook-jupyter hub proxy fabricates it at serve time for browsers; a TS conversion serves browser-direct public stores (below).
- This is the third instance of an existing design principle: dense views fabricated lazily from the MOC domain (§6), caches derived from leaf truth (D9), and now NESTED derived from morton.

## Why (ratified rationale)

1. **Mixed-order capability in one coordinate** — the decisive one. NESTED u64 can't represent mixed orders in one array (order is side-metadata; healpix-land needed zuniq/nuniq for this). Morton words carry order intrinsically. D2 already anticipates mixed orders (coarse shards in sparse regions), and #201's pyramid/overview products are inherently multi-order — morton-only makes both first-class instead of scheme-breaking.
2. **Kills the dual-encoding cost**: one less u64 array per leaf to write, store, test, and keep consistent (8 B/cell — ~8 MB raw per shard at parent 9 / child 19 — plus one array's objects per leaf in the #236/#240/#242 object-count currency), and deletes the NESTED↔morton double-encode maintenance #72 complains about.
3. **The reader stack is ready**: moczarr's `open_hive()` + `MortonIndex` via xdggs's registry (espg/moczarr#6) mean we no longer depend on stored NESTED for a working xarray story — the 2026 situation is materially different from when #72 was filed.

## Refinements ratified with the decision

- **Public-bucket spec path stays a goal.** The hub-proxy fabrication covers cryocloud, but we still want to **host a source.coop example that renders browser-direct** — so the morton scheme must be a real, declared spec (not just "our proxy knows"), and the gridlook fork (later upstream) needs native TS morton support for that path.
- **Order-29 dual-encoding rule.** Two order-29 encodings exist: (a) genuinely order-29 resolution, and (b) **unknown resolution encoded at order 29** — the default for any raw lat/lon conversion, expected to be common. For browser/Number-safety (NESTED ids are float64-exact only through order 24): case (b) is **clipped on the fly to order 24** (equally valid — the resolution is unknown, so 24 is as true as 29); genuinely-finer-than-24 data takes other measures (hub-side fabrication, aggregation). **Open sub-question (O-a below): the discriminator** — readers need declared metadata to tell (a) from (b) (manifest/attrs field, e.g. `resolution: exact | point`); this also intersects mortie's documented point-id/area-word parse non-injectivity at order 29.
- **Breaking-change scope check**: the morton datatype itself has been stable; the parts of the store spec still in flux are elsewhere (windowing, coverage envelopes) — so the legacy-format surface this creates is expected to be small and mostly fixture-regeneration.

## What it requires (phasing sketch — implementation issues carved out after doc amendment)

1. **Spec**: a morton `indexing_scheme`/`name` entry — normative encoding on the mortie spec page (espg/mortie#62), plus either the zarr-conventions/dggs registry ask (#72 ask 3, now required-not-nice-to-have) or an interim self-declared convention UUID (the conventions mechanism supports this). Includes the O-a discriminator.
2. **moczarr fabrication layer — THE GATE** (mirrors #251's reader-gate pattern): a lazily-derived `cell_ids` view on `open_hive` results, exact via `mort2healpix`, so Python consumers keep NESTED for free. zagg's writer flip does not land until this ships.
3. **zagg writer**: `emit cell_ids` knob (default on) → default flip → removal, sequenced as 0.x phases; dggs convention attrs updated to name morton as the coordinate; fixtures/goldens regenerate; bench object counts update.
4. **Viewers**: gridlook-jupyter proxy fabricates NESTED at serve time (cryocloud path — browser unchanged); gridlook fork adds a detector branch for the morton convention + TS morton decode (BigInt words, decimal-string render form per D1, the order-29→24 clip rule) for the browser-direct/source.coop path.

## Verification (dispatched, results to be posted on this thread)

- **artools/#213**: confirm the integration never consumes the stored `cell_ids` coordinate (believed pure declarative-temporal).
- **gridlook hover/selection**: confirm it requires only "ids are global NESTED pixel numbers at the declared nside" — i.e., fabricated NESTED satisfies it bit-for-bit.

moczarr's opener is already tolerant of `cell_ids` absence (`set_coords` only lifts coordinates that exist), so reader-side carry-both is cheap.

## Cross-references

#198 (design doc home; this amends §6), #72 (reframed again: the upstream convention ask is now on the critical path for public stores), #135 (`cell_ids_encoding` inverts: morton stored, cell_ids derived), #251 (same pre-1.0 gate logic; sequencing sibling), #201 (pyramids = the mixed-order beneficiary), espg/moczarr#1 (fabrication layer lands there), espg/gridlook#1 (viewer-side items).

Doc amendment PR (sparse_coverage.md §6 + decisions registry) to follow once espg confirms the D-number/O-number assignments and the verification results land.

### LINKED ISSUE #296: grafana monitoring board
I'm been growing increasingly frustrated with updating benchmarks-- it's a hassle, error prone, and time consuming. It occurs to me that the problem is broader; we need a way to estimate, track, and report costs more generally. My view on this is that we should standardize on some bookkeeping infrastructure and design.

We already deliver data to a spatially sharded location-- why not just add sidecar files with the zarr arrays that have performance and data summary information? We should take the klugde of information that we currently grep for out of cloudwatch, and summerize it at the shard level. I.e., a json or parquet file that contains things like read/write/agg split in seconds (or similar info we log for the raster module), observation count, etc. One part of my thinking is that this should integrate with both gridlook ( #74 and #262 ) and with pyramid / overview generation ( #201 ). This actually touches on several architecture components-- and hopefully unifies them into a single cohesive system.

On the invoke side-- we should start reporting both the max, the estimated (optional), and the actual run cost in the return... it would also be nice if we put together a proper python wrapper so that we can call invoke from within an ipython notebook and get proper formatted responses and a tdqm style tracking bar. The max cost for the invoke is easy-- number of shards (know from shard map) times cost per second of instance (known from agg template yml) times 900. The actual we log next to the data; an estimate could be something if we see evidence of a prior run in at the shard prefix level (the last piece can optionally be differed.

This also means a substantial rethinking of the file layout that we currently have. We're keeping the zarr stores named by the morton shard index-- but we should stop doing this, since we already have the morton index from the s3 path prefix concatenation. Instead, we should save a short hash of the agg template yaml, and use that as base name for what we write. This pattern matches the provenance design from nixos / nix store; for us it works conveniently since we have invariant guarantees we can depend on from, shardmap , stac catalogs, and our agg templates. Our agg templates are invariant with regard to the shardmap ; agg template says what, and shardmap says where. That means that we can use the hash of the template and know if that agg has been run prior... and we have no collision from shard to shard. The shardmap itself is composable (see #294 and associated PR) and can get degraded up or down a morton order. This means we can place a shardmap sub-map with the data at the leave prefix, and then aggregate that shardmap up each hierarchy of the zagg morton prefix tree. STAC catalogs have the same subset / coarsen / merge property too.

This begins to suggest that we have something like s3://sliderule-public-cors/zagg/* with morton prefix beginning immediately, and then can track all the aggregations we do... including (potentially) checking if we even have that aggregation already, or if we have an estimate of how much that aggregation might cost given prior runs. We can visualize cost as well-- either directly (and spatially) in gridview, or in tabular form with graphana... or potentially a merge of both backends?

regardless, one of the things that we want to have is a record of what the data cost; we can see in it the provenance data, and merge it as metadata in the viewer so that information is carried along.

Big items to do: standardize what metadata, and what file types, plus the merge and split logic (most / all of this data can be parsed either direction-- leaves building up to the root, or tracing down the root towards the leaves). Also, define the pyrimid semantics and module. Some of this lives in buckets that might have data expiration policies... can we touch/refresh files that are cached or accessed to stop deletion?

Obviously a lot to talk about here-- what are your thoughts, worries, concerns? What do you like about the design, and what risks do you see? And suggestions or changes / enhancements to make this stronger? Let me know pros and cons you see.

espg:
Closes #262
Closes #296

The consolidated spec amendment: records the espg-ratified morton-only-storage decision (#262) **plus every design decision settled over the last ten days** that the spec didn't yet reflect, swept from the #296-family, #251/#253, #236/#233/#241, #238, #209/#210, #294, and #300 threads. Every new entry cites its provenance inline — thread-ratified comments are linked directly; in-session ratifications are marked as recorded, per the same discipline the phase-1 review enforced on D16.

## Phases

- [x] Phase 1 — D16 (morton-only storage) + O10 (order-29 discriminator) + §6 rewrite; adversarial review run, both findings folded
- [x] Phase 2 — consolidation: D17–D22, O3 resolved, O11 opened; §1/§2/§3/§6/§7 amended to match
- [x] Phase 3 — **D19 revised to hash-first product roots** (espg-ratified 2026-07-20 after pushing on terminal-leaf semantics; [trade study on this thread](https://github.com/englacial/zagg/pull/306#issuecomment-5019696069)): `{store_root}/{hash}/` product roots, each subtree a complete unmodified morton-hive store — #299 becomes **additive, not breaking**; D3/D13 survive untouched; the singular-`coverage.moc` inconsistency in the phase-2 layout block is fixed; D16's break-window note updated (the writer flip is now the only breaking change in this family)
- [x] Phase 7 — **final rulings + reader model** (espg-ratified 2026-07-20; [record on this thread](https://github.com/englacial/zagg/pull/306#issuecomment-5025396393)): O8 constants blessed as-shipped (bit convention + range ordering contract, zstd level non-normative); **O11 → RESOLVED** (per-named-zarr-array sha256 over decoded values, at write, all arrays incl. coordinates); **D23 tokens ratified** (`all` + `{window}.stats.json` alignment — PR #307 shipped exactly these); **O14** opened (reader DataTree model: product + resolution axes yes, spatial axis never, interop hierarchy as D12 derived cache)
- [x] Phase 6 — **espg-directed spec-gap additions** ([record](https://github.com/englacial/zagg/pull/306#issuecomment-5025140978)): O12 retention proposal (prefix rules per product, tags per window), O13 publication profile (prefix copy minus operator telemetry, roll-up summary preserved), D22 debris-collection family, §2 concurrency contract (same-leaf concurrency out-of-contract after review fold), §8.3 standing-test-obligations index, D19 refinements (hash-is-truth, full-length + short-fingerprint display, URL-safe names), D24 `none`-handling leans (deferred)
- [x] Phase 5 — **D19 rev 2 + D24** (espg-ratified 2026-07-20 in-session; [record + analysis on this thread](https://github.com/englacial/zagg/pull/306#issuecomment-5024857031)): product roots become **named** (`{name}/`, hash demoted to metadata); the hash shrinks to the **semantic core** (aggregation block + data-source semantics + grid type/indexing — cell order, packaging, worker knobs, streaming mode all excluded) as a frozen manifest key with the canonical core at `{name}/aggregation.yaml`; **D24 resolution polymorphism** — cell order is a query/packaging axis (composes per aggregator merge laws; composability class `exact|approximate|none`; one resolution per (shard, window) with a **cross-order overwrite guard**; allowed-set frozen keys; gated on mortie#116/moczarr#8 for readers); D20 gains timestamp-first run-record names + `run_id`/`semantic_hash` in sidecars
- [x] Phase 4 — **D23: window-only leaf naming** (`{window}.zarr`, `morton-hive/3`; espg-ratified 2026-07-20 in-session; [trade study on this thread](https://github.com/englacial/zagg/pull/306#issuecomment-5019934798)): completes the product/space/time axis separation (product = root, space = path, time = basename); `all.zarr` token and `{window}.stats.json` sidecar alignment recorded as **leans, not ratified**; `/1`–`/2` grammars stay frozen for existing stores; D3/D13 carry supersession pointers; adversarial review's two findings folded (layout block keeps ratified sidecar names with the lean marked; D3/D19 version-scoped)

## What phase 2 adds

**New decisions (registry §8.1):**
- **D17** — hive+sharded is the HEALPix default; flat/fullsphere deprecated (dense path removed, PR #257); ShardingCodec one-object-per-array leaves; raster leaves unsharded by design; `grid.shard_order` removed (#238).
- **D18** — ragged output = sharded vlen-bytes with golden-bytes framing pin (#209); typed ZDType deferred behind the #210 gates.
- **D19** — **named product roots + semantic-core hash** (phase-5 revision; the name is the address, the hash is the integrity check): a multi-product store is a directory of complete morton-hive stores under human-readable `{name}/` roots; `semantic_hash` (output-defining subset only) is a frozen manifest key with the canonical core at `{name}/aggregation.yaml`; literal per-run templates archive with run records; catalog identity in the sidecar; immutable-provenance opt-in `{name}+{catalog-hash}/` (#296/#299; both trade studies + the phase-5 record on this thread).
- **D20** — per-shard stats sidecar (sibling object, associative-only schema, `invoked_by`, `run_id`, `semantic_hash`) + run-level parquet at the product root including failure rows, timestamp-first names (#297).
- **D24** — resolution polymorphism (phase 5): cell order excluded from product identity; per-aggregator composability class; one resolution per (shard, window) with a cross-order overwrite guard; manifest order fields become allowed sets.
- **D21** — `path_grouping` manifest parameter, default 1; grouping is ergonomics, never a schema fork (#300).
- **D22** — one unified sweep owns overview zarrs, MOC regen, sub-shardmap rollups (full ShardMap JSON), stats rollups; generation stamps; rollup == direct obligation; MOC-first reader contract strengthening D10 (#300).

**Registry updates:** O3 → RESOLVED (standalone moczarr, per the #251 ratification); O11 opened (logical content hash — the one item still awaiting espg sign-off); amendment pointers added at D2, D3, D10, D11, D13.

**Section edits:** §1 fullsphere reframed as legacy; §2 layout block, node invariant, and D2/D3/D13 bullets updated for hash naming, sidecars, and `path_grouping`; §3 gains `path_grouping` + per-artifact-family rollup schedules and the per-product manifest qualification; §6 reflects O3's resolution; §7 lists all four sweep artifact families with generation-stamp semantics.

All amendments are within-section — §ordering is unchanged, so `docs/hive_layout.md` and `docs/deployment/benchmark.md` cross-references stay valid.

## Closure map

Design-record issues close on this PR; implementation issues close on their own PRs — nothing floats:

- **Closes here**: #262 (morton-only design; implementation carved to #304/#305, attached as its sub-issues) and #296 (cost/provenance discussion; decisions recorded here, work carved to sub-issues #297–#301).
- **Close on their own PRs**: #297 → PR #302 (wired), #298 → PR #303 (wired); #299, #300, #301, #304, #305 → future implementation PRs.
- **Deliberately not closed here**: #251 (phases 2–3 pending), #201 (open pyramid questions), #210 (gated follow-up).

## Testing

Docs-only; codespell clean via the pre-commit hook. No code paths touched.

## Questions for review

- **Provenance caveat, stated plainly**: entries marked "in-session, recorded on <link>" trace to Claude-posted records of session ratifications under the espg account; the directly espg-typed thread approvals are the linked comments on #297 (two), #298, #299, #300, #301, #236, #238, #251, and PR #257. Merging this PR is the act that ratifies the *doc text*; if any recorded item doesn't match your intent, flag the entry.
- **O11 is a proposal, not a decision** — per-array sha256 over decoded values, exact-match, no float tolerance. It's written as open; sign-off belongs on #299/#305.
- **Flag from the sweep** (not incorporated, surfacing for a call): #276 item 5 notes PR #208's coverage-MOC contract constants (bitmap order vs literal order-19, zstd level 3, root-MOC range ordering) merged without explicit sign-off, and they sit beneath O8's "espg-ratified" text. If you bless them, O8 gets a one-line addendum; if not, they need a thread.
- D16/O10 numbering from phase 1 stands (registry ran D15/O9); phase 2 continues D17–D22/O11.
- `Closes #296`/`Closes #262` leave their open sub-issues attached under closed parents — detach the lines if you'd rather keep the umbrellas open until all children land.

### DIFF
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -820,16 +820,25 @@ rollup leaves all leaf reads intact.
   provenance attrs; the pyramid is the store's resolution axis, partially
   materialized. Reader support is gated on mortie#116 mixed-order morton
   (tracked as moczarr#8); the schema needs no bump — this is what the
-  coordinate system was built for. Open sub-point (deferred pending
-  discussion, espg leans recorded 2026-07-20): how a `none`
-  (non-composable) field behaves — the recorded lean is **per-field
-  exclusion from coarser materializations** (composable fields roll up;
-  non-composable fields exist only at native resolution — matching
-  #201's "which fields degrade" question) rather than min-semantics
-  pinning the whole product; and whether `none` fields are permitted at
-  all — lean: allowed, with a **loud warning** at template validation.
-  Per-field exclusion makes "resolution of a product" per-field, which
-  readers must handle; not ratified.
+  coordinate system was built for. Sub-point **ruled** (espg-ratified
+  in-session, 2026-07-20, on the #201 thread): `none` (non-composable)
+  fields are **allowed, with a loud warning** at template validation,
+  and at overview orders the default is **per-field exclusion** —
+  composable fields roll up, `none` fields exist only at native
+  resolution (min-semantics product pinning rejected; "resolution of a
+  product" is per-field, which readers handle via the manifest's
+  per-family declarations). An **explicitly declared derived summary**
+  is available as the opt-in: e.g. an auto-digest of a roster field's
+  raw values **under a different field name** at overview orders, so
+  overview schema never silently differs from source; the declaration
+  lives in the **pyramid block, never the semantic core** (leaf truth is
+  unchanged — two products differing only in overview-summary
+  declarations are the same product), and derived fields carry their own
+  composability class and O11 hashes. Deterministic subsampling
+  (seeded reservoir per cell, seed = f(cell, window, field) for O11
+  reproducibility) is **deferred** until a concrete consumer (the #265
+  HHDC class) asks; concatenation ("roster pyramids") is **rejected** —
+  volume never shrinks, every level duplicates the raw data beneath it.
 
 ### 8.2 Open for review (input needed)
 
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -625,7 +625,10 @@ rollup leaves all leaf reads intact.
   is **mergeable by construction** — only associative stats (counts, sums,
   min/max, t-digests; never stored means) — so up-tree rollups are a pure
   fold (D22). The record also rides the async result envelope; the
-  dispatcher writes a **run-level parquet at the store root** (one row per
+  dispatcher writes a **run-level parquet at the product root** — a run maps
+  to one template ⇒ one product tree (D7/D19), so it lands under `{hash}/`,
+  never the multi-product store root, whose node invariant admits only
+  `{hash}.yaml` + `{hash}/` — with one row per
   shard, *including failure rows* sourced from the run report — sidecars
   exist only on success; CloudWatch structured logs remain the failure
   forensics channel). Sidecars omit account-identifying fields (request
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -154,8 +154,14 @@ products by listing `{store_root}/` and reading each
   no objects (D4/D5), their run records are distinct by name
   (timestamp + run id, D20), and the manifest precheck admits both
   (frozen keys match by construction under D19). Same-(shard, window)
-  concurrency degrades to D4 semantics — last writer wins; the loser's
-  partial output is stamped-over or remains ignorable debris. Interior
+  concurrency is **out of contract**: *sequential* re-runs give clean D4
+  last-writer-wins replacement, but a leaf is several objects (per-array
+  ShardingCodec objects D17, the ragged object D18, root attrs, coverage
+  sidecar) and S3 PUTs are not transactional across them — two
+  *simultaneous* different-content co-writers can interleave those objects
+  and leave a stamped-but-torn leaf, which is neither replacement nor
+  ignorable debris. Detection is content verification (the O11 hash) or a
+  re-run; the supported contract stays shard-disjoint. Interior
   rollups may go stale during any of this, which D22's generation stamps
   make detectable, never silently wrong. Cross-product concurrency shares
   nothing at all.
@@ -936,7 +942,9 @@ registry — CI coverage should be auditable against this list:
   no-op) and nothing-load-bearing (deleting every rollup leaves leaf
   reads green).
 - **Merge-fold algebra** (D20): associativity/commutativity of the stats
-  record fold; merge-of-children == direct.
+  record fold; merge-of-children == direct — exactly for count/sum/min/max,
+  up to the approximate class (`np.isclose`) for order-dependent t-digest
+  members (cf. D24).
 - **Semantic-hash canonicalization** (D19): syntactic edits (whitespace,
   key order, comments) never change the hash; packaging-knob edits
   (orders, chunking, worker size, streaming mode) never change the hash;
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -551,10 +551,12 @@ rollup leaves all leaf reads intact.
   the manifest's `shard_order` (tree/dispatch order) is unrelated and stays.
   Full flat-machinery removal remains gated on the O3 reader (the #251
   phase-2 gate).
-- **D18 — Ragged output = sharded vlen-bytes** (espg-ratified in-session,
-  recorded on
-  [#209](https://github.com/englacial/zagg/issues/209#issuecomment-4940116927);
-  tracking [#210](https://github.com/englacial/zagg/issues/210)). Ragged
+- **D18 — Ragged output = sharded vlen-bytes** (measurement + recommendation
+  on
+  [#209](https://github.com/englacial/zagg/issues/209#issuecomment-4940116927),
+  with espg's in-session ratification of the vlen-bytes adoption recorded in
+  the [#210](https://github.com/englacial/zagg/issues/210) body; tracking
+  [#210](https://github.com/englacial/zagg/issues/210)). Ragged
   t-digest arrays store as `bytes` dtype + vlen-bytes codec under the
   ShardingCodec — one object per shard, 2-GET single-cell reads — with
   element interpretation as an attrs convention and a **golden-bytes framing
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -353,6 +353,11 @@ new wiring on the read side.
 4. Within the leaf zarr, the `morton` coordinate locates the observation(s)
    in the containing cell.
 
+(Under D24 a target's orders may be regionally heterogeneous — the
+declared sets bound them; the *leaf's* actual orders come from its
+morton words / MOC, and the truncation predicate is unchanged.
+Mixed-order reader support is gated on mortie#116 / moczarr#8.)
+
 Per-observation LISTs are forbidden in the join loop: at 2,000 workers ×
 millions of photons the walk is the robustness path, not the join path (D10).
 
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -5,8 +5,9 @@ temporal-partitioning amendments (D13–D15) ratified on
 [#237](https://github.com/englacial/zagg/issues/237); morton-only-storage
 amendment ratified on
 [#262](https://github.com/englacial/zagg/issues/262) (D16; open item O10
-carved from it); 2026-07 consolidation (D17–D24, O3 resolved, O11–O13
-opened, §8.3 test-obligations index added)
+carved from it); 2026-07 consolidation (D17–D24; O3/O11 resolved, O8
+constants blessed, O12–O13 opened, D23 tokens ratified, §8.3
+test-obligations index added)
 recording decisions settled on the
 [#251](https://github.com/englacial/zagg/issues/251)/[#236](https://github.com/englacial/zagg/issues/236)/[#209](https://github.com/englacial/zagg/issues/209)
 and [#296](https://github.com/englacial/zagg/issues/296)-family threads —
@@ -69,13 +70,14 @@ what zagg consumes:
                                     the manifest path_grouping param, D21)
       {window}.zarr/             <- leaf, basename = time window (D23,
                                     morton-hive/3); `all.zarr` for
-                                    schedule: none (reserved token — lean).
+                                    schedule: none (reserved token, ratified;
+                                    excluded from the window grammar).
                                     /1–/2 stores keep {full_id}.zarr and
                                     {full_id}_{window}.zarr (D3/D13)
-      stats_{window}.json        <- per-shard stats sidecar, sibling; ratified
-                                    D20 naming (stats.json for schedule: none).
-                                    D23 lean, not ratified: {window}.stats.json
-                                    / all.stats.json (follow-up to #302)
+      {window}.stats.json        <- per-shard stats sidecar, sibling (D20,
+                                    D23-aligned for /3; all.stats.json for
+                                    schedule: none). /1–/2 stores keep
+                                    stats_{window}.json / stats.json
       <sub-shardmap JSON>        <- leaf sub-map for sweep rollups (D22)
 ```
 
@@ -769,12 +771,15 @@ rollup leaves all leaf reads intact.
   D20 sidecar `shard_key` (an fsck pays one GET per leaf); D3's
   "unambiguous if moved" softens to "recoverable from attrs" — the moved
   case reduces to a downloader that materializes the prefix tree into
-  folder names (espg-noted in-session). **Leans recorded, not ratified**:
-  the `schedule: none` reserved token — proposal `all.zarr` (reads as
-  all-time; cannot collide with the digit-shaped window grammar;
-  `none.zarr` matches the schedule literal but reads worse) — and the
-  sidecar alignment `{window}.stats.json` / `all.stats.json` (D20
-  naming; follow-up to the merged PR #302). Rejected alternative: time
+  folder names (espg-noted in-session). **Both leans ratified**
+  (espg, in-session 2026-07-20): the `schedule: none` reserved token is
+  **`all`** (`all.zarr`; reads as all-time; cannot collide with the
+  digit-shaped window grammar; excluded from the window grammar forever
+  on the mortie spec page), and the sidecar aligns to
+  **`{window}.stats.json` / `all.stats.json`** for `/3` stores (the
+  spec-keyed seam shipped in PR #307 makes both a constant flip; `/1`–`/2`
+  stores keep the frozen `stats_{window}.json` / `stats.json` names).
+  Rejected alternative: time
   as a *path* level (`{name}/{window}/{morton…}`) would make
   window-scoped ops prefix-cheap but duplicates the morton tree per
   window and shatters the dominant read — a time series at a location —
@@ -877,7 +882,12 @@ rollup leaves all leaf reads intact.
   a third value, `"full"`). Bit convention (frozen
   with the mortie spec, golden-vector-pinned on PR #208): bit i = the i-th
   shard-subtree cell in ascending packed-word order (base-4 D1 digit tail,
-  digits 1..4 → 0..3), MSB-first per byte.
+  digits 1..4 → 0..3), MSB-first per byte. *Addendum (espg-blessed
+  in-session, 2026-07-20 — closing the #276 item-5 flag)*: the PR #208
+  contract constants stand as-shipped — the bitmap bit convention and the
+  root-MOC range ordering are **contract** (both golden-vector-pinned);
+  zstd level 3 is **non-normative** (any zstd level decodes identically;
+  only "zstd stream" is contract).
 - **O9 — end-of-run root MOC default**: **RESOLVED — on** for hive stores
   ([espg](https://github.com/englacial/zagg/issues/200#issuecomment-4938859764):
   coverage MOCs are the default for healpix templates; PR #208 implements
@@ -895,21 +905,31 @@ rollup leaves all leaf reads intact.
   documented point-id/area-word parse non-injectivity at order 29; field
   name, values, and placement to be frozen with the mortie spec page
   ([mortie#62](https://github.com/espg/mortie/issues/62)).
-- **O11 — logical content hash of outputs** (carved from D19; **proposal
-  awaiting espg sign-off** — espg asked for expanded context on
-  [#299](https://github.com/englacial/zagg/issues/299#issuecomment-5017263033),
-  expansion posted
-  [in reply](https://github.com/englacial/zagg/issues/299#issuecomment-5017334704)).
-  Proposal: per-array sha256 over *decoded* values (raw C-order bytes at the
-  declared dtype, after decompression — never stored object bytes, which
-  churn on codec/library upgrades), recorded in the D20 sidecar as
-  `{array_name: hash}` plus one combined hash. Exact bytes, no float
-  tolerance: any value change — including flagged code changes that pass
-  `np.isclose` (the PR #282 class) — flips the hash by design;
+- **O11 — logical content hash of outputs: RESOLVED — adopted as
+  proposed** (espg-ratified in-session, 2026-07-20; carved from D19,
+  discussion trail on
+  [#299](https://github.com/englacial/zagg/issues/299#issuecomment-5017263033)
+  and
+  [the expansion](https://github.com/englacial/zagg/issues/299#issuecomment-5017334704)).
+  Per-array sha256 over *decoded* values — **"per-array" means per named
+  zarr array (variable) in the leaf** (each data field, the ragged bytes
+  array, `morton`, every coordinate), hashed over that array's full
+  decoded contents (raw C-order bytes at the declared dtype, after
+  decompression); the ShardingCodec's inner chunks and the
+  one-object-per-array packaging are invisible to it by construction —
+  never stored object bytes, which churn on codec/library upgrades.
+  Recorded in the D20 sidecar as `{array_name: hash}` plus one combined
+  hash (hash of the sorted per-array hashes). **Computed at write**,
+  in-worker (the data is already in memory — near-free); **scope: all
+  arrays including coordinates** (cheap, deterministic). Exact bytes, no
+  float tolerance: any value change — including flagged code changes that
+  pass `np.isclose` (the PR #282 class) — flips the hash by design;
   interpretation pairs the hash with the sidecar's recorded zagg version.
-  Motivation: outputs have been byte-identical between runs in practice, so
-  this turns D19's "probably already ran" dedup into a verifiable claim
-  without folding catalog identity into leaf names.
+  Three jobs: the verification half of D19's `semantic_hash` ("intended
+  identical" vs "actually byte-identical"); the mismatch localizer
+  ("only `h_li_tdigest` differs in this leaf"); and the detection
+  mechanism for stamped-but-torn leaves under the §2 concurrency
+  contract's out-of-contract case.
 - **O12 — retention/expiration (proposal, espg-directed to record
   2026-07-20)**: **product-level** retention is a *prefix* lifecycle rule
   (one rule per `{name}/` — delete or transition a whole product); 
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -5,7 +5,7 @@ temporal-partitioning amendments (D13–D15) ratified on
 [#237](https://github.com/englacial/zagg/issues/237); morton-only-storage
 amendment ratified on
 [#262](https://github.com/englacial/zagg/issues/262) (D16; open item O10
-carved from it); 2026-07 consolidation (D17–D22, O3 resolved, O11 opened)
+carved from it); 2026-07 consolidation (D17–D23, O3 resolved, O11 opened)
 recording decisions settled on the
 [#251](https://github.com/englacial/zagg/issues/251)/[#236](https://github.com/englacial/zagg/issues/236)/[#209](https://github.com/englacial/zagg/issues/209)
 and [#296](https://github.com/englacial/zagg/issues/296)-family threads —
@@ -64,12 +64,14 @@ what zagg consumes:
     <run records (parquet)>      <- run-level telemetry, one row/shard (D20)
     {sign+base}/{d1}/{d2}/.../   <- one digit per level (D2; digit-chunking is
                                     the manifest path_grouping param, D21)
-      {full_id}.zarr/            <- self-describing leaf (D3), zarr v3; dense
-                                    arrays via ShardingCodec (D17)
-      {full_id}_{window}.zarr/   <- time-windowed leaf (D13); naming frozen on
-                                    the mortie spec page (morton-hive/2)
-      stats.json                 <- per-shard stats sidecar, sibling (D20;
-                                    stats_{window}.json for windowed leaves)
+      {window}.zarr/             <- leaf, basename = time window (D23,
+                                    morton-hive/3); `all.zarr` for
+                                    schedule: none (reserved token — lean).
+                                    /1–/2 stores keep {full_id}.zarr and
+                                    {full_id}_{window}.zarr (D3/D13)
+      {window}.stats.json        <- per-shard stats sidecar, sibling (D20,
+                                    naming aligned by D23; all.stats.json
+                                    for the degenerate case)
       <sub-shardmap JSON>        <- leaf sub-map for sweep rollups (D22)
 ```
 
@@ -90,8 +92,10 @@ bare store; `{hash}.yaml` entries ⇒ product directory).
 - **Full morton id at the leaf** (D3): `.../1/2/3/-31123.zarr/` is
   self-describing without parsing its path, greppable in inventories,
   unambiguous if moved. (Unchanged by D19 — product identity lives at the
-  product *root*, above the tree, so leaf naming and the within-product
-  walk semantics survive untouched.)
+  product *root*, above the tree. *Superseded by D23 for `morton-hive/3`
+  stores*: the basename becomes the time window; the full id stays
+  recoverable from the path arithmetically and from the stamp/sidecar
+  `shard_key`.)
 - **Time-windowed leaves** (D13, ratified on
   [#237](https://github.com/englacial/zagg/issues/237)): a store whose
   manifest declares a temporal window schedule (§3) partitions each shard's
@@ -112,7 +116,8 @@ bare store; `{hash}.yaml` entries ⇒ product directory).
   schedule lives in the manifest (D10 preserved). The no-partitioning
   degenerate case (`schedule: none`) keeps the bare `{full_id}.zarr` name and
   is byte-identical to the pre-D13 layout — a `morton-hive/1` store *is* a
-  `/2` store with `schedule: none`.
+  `/2` store with `schedule: none`. (Leaf *naming* is revised by D23 for
+  `morton-hive/3` stores; the windowing semantics here are unchanged.)
 - **Node invariant**: below a product root, a node contains *only* digit
   children (`[1-4]/`), `*.zarr` objects, and the declared leaf-adjacent
   sidecars — the per-shard stats record (D20) and the sub-shardmap JSON
@@ -177,7 +182,8 @@ again during a run. Contents:
 - **Temporal block** (D15, ratified on
   [#237](https://github.com/englacial/zagg/issues/237)): a store carrying this
   block declares `spec: "morton-hive/2"` (the version string of D6 covers these
-  temporal/windowed-leaf extensions). It records time
+  temporal/windowed-leaf extensions; `/3` = the same semantics under D23
+  window-only leaf naming). It records time
   encoding/units/epoch/calendar, the membership timestamp field, the **window
   schedule** (`none` | `yearly` | `monthly` | `daily` | explicit range list;
   `quarterly` grammar-reserved), and the append policy. Label grammar and
@@ -474,7 +480,9 @@ rollup leaves all leaf reads intact.
   — grammar and boundary semantics recorded on the
   [mortie spec page](https://github.com/espg/mortie/issues/62#issuecomment-4986809092)
   as part of `morton-hive/2`. (Unchanged by D19 — leaf naming survives the
-  product-root design intact.) Reserved (lean, not decided): §7
+  product-root design intact. *Naming revised by D23 for `morton-hive/3`*:
+  the basename becomes the window alone; the `/2` grammar stays frozen for
+  `/2` stores.) Reserved (lean, not decided): §7
   overview/pyramid zarrs inherit window naming (per-window overviews, with
   an optional all-time overview as a derived artifact).
 - **D14 — Coverage `encoding: "full"` fast path**
@@ -669,6 +677,36 @@ rollup leaves all leaf reads intact.
   direction of `ShardMap.reproject` and its no-region semantics are still
   under review on PR #295 — the sweep depends only on the exact coarsen
   direction.)
+- **D23 — Leaf basename = time window (`{window}.zarr`), `morton-hive/3`**
+  (espg-proposed and ratified in-session, 2026-07-20; recorded here and on
+  the PR #306 thread). Completes the axis separation D19 began: product =
+  root prefix (D19), space = digit path (D1/D2), **time = basename** —
+  each identity axis in exactly one place, none encoded twice. This
+  removes the last path/basename redundancy (the #296 observation that
+  started the naming work, applied to its final instance): the morton id
+  currently appears in both the path and the leaf name. Listing a shard
+  node returns the temporal inventory directly (`2019.zarr`,
+  `2020.zarr`, …) — what append planning and time-series discovery
+  actually ask. Spec bump to `morton-hive/3` under D6 versioning: `/1`
+  and `/2` stores remain valid forever under their frozen grammars;
+  readers discriminate by the manifest `spec` string; the `/3` grammar is
+  to be re-frozen on the mortie spec page
+  ([mortie#62](https://github.com/espg/mortie/issues/62)). Costs accepted
+  with the decision: the name==path self-check moves to the stamp attrs /
+  D20 sidecar `shard_key` (an fsck pays one GET per leaf); D3's
+  "unambiguous if moved" softens to "recoverable from attrs" — the moved
+  case reduces to a downloader that materializes the prefix tree into
+  folder names (espg-noted in-session). **Leans recorded, not ratified**:
+  the `schedule: none` reserved token — proposal `all.zarr` (reads as
+  all-time; cannot collide with the digit-shaped window grammar;
+  `none.zarr` matches the schedule literal but reads worse) — and the
+  sidecar alignment `{window}.stats.json` / `all.stats.json` (D20
+  naming; follow-up to the merged PR #302). Rejected alternative: time
+  as a *path* level (`{hash}/{window}/{morton…}`) would make
+  window-scoped ops prefix-cheap but duplicates the morton tree per
+  window and shatters the dominant read — a time series at a location —
+  across W prefixes; reads dominate window expiry, so time stays at the
+  leaf.
 
 ### 8.2 Open for review (input needed)
 
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -5,7 +5,7 @@ temporal-partitioning amendments (D13–D15) ratified on
 [#237](https://github.com/englacial/zagg/issues/237); morton-only-storage
 amendment ratified on
 [#262](https://github.com/englacial/zagg/issues/262) (D16; open item O10
-carved from it); 2026-07 consolidation (D17–D23, O3 resolved, O11 opened)
+carved from it); 2026-07 consolidation (D17–D24, O3 resolved, O11 opened)
 recording decisions settled on the
 [#251](https://github.com/englacial/zagg/issues/251)/[#236](https://github.com/englacial/zagg/issues/236)/[#209](https://github.com/englacial/zagg/issues/209)
 and [#296](https://github.com/englacial/zagg/issues/296)-family threads —
@@ -57,11 +57,13 @@ what zagg consumes:
 
 ```
 {store_root}/                    <- multi-product form (D19): a directory of
-  {hash}.yaml                       stores + a content-addressed registry
-  {hash}/                        <- product root: each product subtree is a
-    morton_hive.json                COMPLETE morton-hive store — bare-named
+  {name}/                        <- NAMED product root — each product subtree
+    morton_hive.json                is a COMPLETE morton-hive store: bare-named
     coverage.moc                    manifest (§3) + optional root MOC (§4, O9)
-    <run records (parquet)>      <- run-level telemetry, one row/shard (D20)
+    aggregation.yaml             <- canonical semantic core (D19); its hash is
+                                    a frozen manifest key, not a path name
+    <run records (parquet)>      <- run-level telemetry, one row/shard;
+                                    timestamp-first names (D20)
     {sign+base}/{d1}/{d2}/.../   <- one digit per level (D2; digit-chunking is
                                     the manifest path_grouping param, D21)
       {window}.zarr/             <- leaf, basename = time window (D23,
@@ -77,10 +79,14 @@ what zagg consumes:
 ```
 
 A bare single-product store (today's layout — `morton_hive.json` at the
-store root, no `{hash}/` level) remains fully valid: the D19 product root is
+store root, no `{name}/` level) remains fully valid: the D19 product root is
 *additive*, and a product subtree is byte-identical to a bare store. A
 reader distinguishes the two forms by what sits at the root (a manifest ⇒
-bare store; `{hash}.yaml` entries ⇒ product directory).
+bare store; only name-shaped prefixes ⇒ product directory). Product names
+must not match the base-component grammar (`-?[1-6]`) so the walker's child
+classification stays unambiguous; gridlook and other viewers enumerate
+products by listing `{store_root}/` and reading each
+`{name}/morton_hive.json` directly — no name↔hash translation layer.
 
 - **Ids are morton decimal strings** (D1): sign + base digit (constant width,
   12 values `1..6`/`-1..-6`), then one digit per order, digits `1-4`, never
@@ -124,9 +130,9 @@ bare store; `{hash}.yaml` entries ⇒ product directory).
   sidecars — the per-shard stats record (D20) and the sub-shardmap JSON
   (D22) — with nothing else, ever: the walker's child classification
   depends on the name set being closed. The product root alone also carries
-  the manifest, MOC objects, and run-level telemetry records (D20); the
-  *store* root of a multi-product directory carries only `{hash}.yaml`
-  registry entries and `{hash}/` product roots (D19).
+  the manifest, MOC objects, the semantic core (`aggregation.yaml`, D19),
+  and run-level telemetry records (D20); the *store* root of a
+  multi-product directory carries only `{name}/` product roots (D19).
 - **Termination condition**: S3 has no empty directories (a prefix exists iff
   ≥1 object lies beneath it) and LIST is strongly consistent, so a
   delimiter-LIST returning no digit-shaped children is a definitive "nothing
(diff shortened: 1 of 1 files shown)


docs/design/sparse_coverage.md
```diff
@@ -419,6 +426,38 @@ write path (§2) is load-bearing; this phase is optimization.
...
+  refinements: the dggs attrs use a **distinct grid `name: "morton"`** —
+  never `name: "healpix"` + `indexing_scheme: "morton"`, which scheme-blind
+  readers silently misread as NESTED (garbage renders); a distinct name
+  makes them hard-reject with a diagnostic, and matches moczarr's xdggs
+  registration. Unknown-resolution point encodings at order 29 are clipped
+  on the fly to order 24 for Number-safe browser paths (NESTED ids are
+  float64-exact only through order 24; genuinely-finer-than-24 data takes
+  other measures — hub-side fabrication, aggregation). The writer flip
+  sequences as 0.x phases (emit knob default-on → default flip → removal),
+  bundled into the same break window as the
```
espg: 🤖 *from Claude (review)*

Fidelity: the #299 break-window bundling reads as part of the ratified record, but it isn't on the #262 thread. The ratified issue body (2026-07-17, "espg-ratified in-session") never mentions #299 — its phasing sketch item (3) only says "fixtures/goldens regenerate." #299 ("Template-hash leaf naming", a sub-issue of #296) is referenced only in a later Claude-authored execution comment ([#issuecomment-5017510905](https://github.com/englacial/zagg/issues/262#issuecomment-5017510905): "bundled into the #299 break window") — a session sequencing decision, not an espg on-thread ratification.

Because this sentence sits inside a D16 entry headed "espg-ratified in-session 2026-07-17," it presents the bundling as ratified. Suggest a D15-style qualifier — the same doc already distinguishes derived implications with "(That exception is an implication recorded from the ratified schedule set, not separately ratified on the thread.)" A parallel parenthetical here (e.g. "the #299 bundling is a session sequencing decision, not separately ratified on #262") keeps the fidelity clean. The #299 cross-reference itself is accurate (it is the leaf-rename issue).

docs/design/sparse_coverage.md
```diff
@@ -2,7 +2,9 @@
 
 **Status**: draft. Tracks [#198](https://github.com/englacial/zagg/issues/198);
 temporal-partitioning amendments (D13–D15) ratified on
-[#237](https://github.com/englacial/zagg/issues/237).
+[#237](https://github.com/englacial/zagg/issues/237); morton-only-storage
+amendment (D16, O10) ratified on
```
espg: 🤖 *from Claude (review)*

Minor: "(D16, O10) ratified on #262" groups O10 with the ratified decision, but O10 is an *open* item — its `resolution: "exact" | "point"` field is explicitly a proposal (§8.2, sign-off deferred to #305). Only the amendment/number assignment and D16 are ratified; the discriminator's content is not. The O10 entry reads correctly as a proposal, so this is only about the top-of-doc provenance line possibly being misread as ratifying O10's content. Consider "D16 ratified, O10 opened" or "the D16/O10 amendment recorded on #262" to preserve the decided-vs-proposal line the rest of the doc is careful about.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Folded. Appended a D15-style qualifier immediately after the #299 bundling sentence in the D16 entry: "(The #299 break-window bundling is a sequencing implication recorded from in-session planning, not separately ratified on the thread.)" — so the bundling no longer reads as part of the espg-ratified record. Fix commit 72ce5449e3ecdcdcf43e0e54493eb7231943abd1.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Folded. Reworded the top-of-doc provenance line so ratification attaches to D16 only: "morton-only-storage amendment ratified on #262 (D16; open item O10 carved from it)." O10 no longer reads as ratified. Fix commit 08aa823a9cc76f82496c8efd382055027dab5e0a.

docs/design/sparse_coverage.md
```diff
@@ -419,6 +494,146 @@ write path (§2) is load-bearing; this phase is optimization.
...
+- **D17 — Hive+sharded is the HEALPix default; flat/fullsphere deprecated;
+  dense leaf arrays write through the ShardingCodec.**
+  (Phase-1 ratification recorded on
+  [#251](https://github.com/englacial/zagg/issues/251#issuecomment-4989697559);
+  landed via PR #257 (default flip + dense-layout removal), #233 (`sharded:
+  true` default), #241 (hive dense parity); leaf object model ratified
+  in-session, recorded on
+  [#236](https://github.com/englacial/zagg/issues/236#issuecomment-4986241727).)
+  Every HEALPix config — point aggregation and raster — targets hive; the
+  grid determines the layout (no `store_layout` knob survives, #238); rect
```
espg: 🤖 *from Claude (review)*

D17 says "the grid determines the layout (**no `store_layout` knob survives, #238**)", but that's inaccurate on two counts and contradicts this same entry two sentences down:

- `store_layout` was **not** removed. PR #257 (via #253) made the *default* grid-keyed (`get_store_layout` → hive for HEALPix); it explicitly kept `store_layout: flat` as a knob that emits a `DeprecationWarning`, and the flat-arm benchmark configs in this repo still `pin store_layout: flat`. D17's own next-but-one sentence confirms flat is still present: "Full flat-machinery removal remains gated on the O3 reader." A knob whose value still selects still-present machinery hasn't been removed — the *default* flipped, the knob didn't disappear.
- The `#238` citation is wrong here. Issue #238 is titled "Demote or drop **grid.shard_order**", and D17 already cites it correctly for exactly that in the very next sentence (`grid.shard_order ... is removed (#238)`). The store_layout default flip is #253 / PR #257, not #238.

It also leaves a live contradiction with the surviving O9 text in §8.2, which still reads "default true under `store_layout: hive`" — accurate only *because* the knob survives.

Suggest: "the store_layout **default** is grid-keyed (hive for HEALPix; PR #257 / #253); an explicit `store_layout: flat` survives with a `DeprecationWarning` until the O3-gated flat removal." Drop the #238 cite from this clause (keep it on the `grid.shard_order` sentence).

docs/design/sparse_coverage.md
```diff
@@ -419,6 +494,146 @@ write path (§2) is load-bearing; this phase is optimization.
...
+  array per leaf via the ShardingCodec (always-accumulate); `sharded: false`
+  remains a legitimate opt-out, and **raster leaves skip the sharding codec
+  by design** — the only raster/aggregation difference at the leaf
+  ([espg on PR #257](https://github.com/englacial/zagg/pull/257#issuecomment-5005249126)).
+  `grid.shard_order` (sub-shard object granularity) is removed
+  ([#238](https://github.com/englacial/zagg/issues/238#issuecomment-4986885796));
+  the manifest's `shard_order` (tree/dispatch order) is unrelated and stays.
+  Full flat-machinery removal remains gated on the O3 reader (the #251
+  phase-2 gate).
+- **D18 — Ragged output = sharded vlen-bytes** (espg-ratified in-session,
```
espg: 🤖 *from Claude (review)*

Provenance mismatch. D18 is labeled "espg-ratified in-session, **recorded on** [#209 comment 4940116927]", but that linked comment is a Claude-posted recommendation whose closing line is literally "**espg to steer the adopt decision**" — it *predates* any ratification and records a pending recommendation, not an approval. #209 has exactly two comments and both are `🤖 *from Claude*`; there is no espg-typed adopt anywhere on the thread.

This breaks the discipline the rest of the set keeps: D16 / D21 / D22's "recorded on <link>" each point at a comment that explicitly *states* the in-session ratification ("Ratified decisions (espg, in-session)"). Here the cite points at the ask, not the record — so as written it reads as if the thread ratified something it didn't.

Either link the comment that actually records espg's adopt, or relabel to match reality using the same honest-caveat form already used for the D15 append-exception and the D16 #299-bundling note — e.g. "recommendation posted on [#209 comment 4940116927]; adopt decided in-session, not separately recorded on the thread."

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Fixed in ca0d41d. Corrected the D17 clause: it now reads "the grid keys the default layout (hive for HEALPix; PR #257 / #253), and an explicit `store_layout: flat` survives only as a deprecated escape hatch (emits a `DeprecationWarning`) until the O3-gated flat removal." The erroneous `#238` cite is dropped from this clause; the correct `#238` cite stays on the `grid.shard_order` removal sentence below. This removes the contradiction with the surviving §8.2 O9 text and the "Full flat-machinery removal remains gated on the O3 reader" sentence — the default flipped, the knob did not disappear.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Fixed in f78537a. Relabeled D18 provenance to match the record: "measurement + recommendation on [#209 comment 4940116927], with espg's in-session ratification of the vlen-bytes adoption recorded in the #210 body; tracking #210." The #209 comment link stays (it is the measurement + the "espg to steer the adopt decision" recommendation), and the ratification is now cited to where it is actually recorded — the #210 body, which reads "adopted via the head-to-head measurement and espg's in-session ratification" (issue authored by espg). No longer claims the #209 thread ratified anything.

espg:
🤖 *from Claude*

## D19 revision (phase 3, edaccd7): hash-first product roots — trade study

espg pushed on a specific consequence of the phase-2 D19 (template-hash *leaf basenames*): earlier decisions leaned on telling a terminal leaf by the absence of further prefix encoding beneath it, and colocating products at different shard orders in one tree breaks that. Working the question honestly reversed the design. Ratified in-session 2026-07-20; recorded here for the thread.

### Does losing position-terminal detection break anything?

Sharpening the premise first: the "no prefixes below ⇒ terminal" assumption was already half-abandoned pre-hash — D11 puts overview zarrs at *ancestor* nodes ("never inferred from position"; hence the `role` attr), and D13 allows several `.zarr` per node. What remained true was narrower: *within one product*, source shards partition space. Colocation breaks that only for walkers that don't filter by product. So:

- **Hot-path readers: immune by construction.** MOC-first readers (D10/D21) take occupancy — cells *and* orders, morton words carry order — from the MOC and compute paths. Terminal-ness never comes from tree position.
- **Sweeps: no regression.** D22 sweeps work from run records, never by listing.
- **What actually degrades**: audit/fallback walks must descend the *union* tree (digit children holding only another product's finer data yield nothing for the walked product — bounded waste, rare path), and — the decisive part — *product-scoped operations*.

### The real cost of shared-tree colocation

S3's only cheap scoping primitive is the prefix, and leaf-basename colocation spends the prefix on space instead of product. Everything per-product then becomes the expensive form:

- **Delete a product** = enumerate-by-basename over the whole tree — exactly the full recursive enumeration D10 (as strengthened) declares out-of-contract.
- **Lifecycle/expiration per product** = per-object tagging at write, because prefix-filtered rules can't see products.
- **Access policy per product**: S3 policies are prefix-based — one public demo product + one private product in a bucket is *impossible* under colocation, trivial under hash-first. (Matters for the source.coop public-example goal.)
- **Per-product inventory/analytics**: prefix-scoped configs, same story.
- **Naming surface**: every root and leaf-adjacent artifact needs hash-prefixing — and the phase-2 draft already leaked once (a singular `coverage.moc` survived in the §2 block). That's evidence about the design, not just the draft.

### Why hash-first (`{store_root}/{hash}/{morton tree}`) wins

It keeps everything the multi-product store was for — one root, `{hash}.yaml` content-addressed registry, product discovery = one root listing, dedup = HEAD a computed path — while restoring per-product tree purity. The near-decisive property: **a product subtree is a complete, unmodified morton-hive store.** Bare-named manifest and MOC; `{full_id}.zarr` leaves; D3 and D13's frozen grammar untouched; `moczarr.open_hive` works on a product root today, unchanged. The multi-product store becomes *a directory of stores plus a registry* — a much smaller idea than a new tree grammar. Consequences:

- **#299 stops being a breaking change.** Existing single-product stores are already valid; the product-root form is additive. Readers distinguish root forms by content (manifest ⇒ bare store; `{hash}.yaml` entries ⇒ product directory).
- The D20 sidecar naming shipping in PR #302 (`stats.json` / `stats_{window}.json`) is final — no future `{hash}.` rename.
- The D16 writer flip is now the *only* remaining breaking store change in this family (its entry updated accordingly).
- D7 generalizes cleanly: "one store per dataset" → "one product tree per template" under a shared root; cross-product joins stay §5 manifest + truncation arithmetic, which never depended on physical colocation.

**What hash-first gives up**: a single spatial prefix spanning all products — i.e., region-scoped *storage* operations across every product at once (mirror-all-products-for-Antarctica as one prefix copy; one-pass all-product spatial audits). We could not name a planned workload that needs this: cross-product *reads* are arithmetic and per-product GETs in both layouts, and region ops within a product still get a spatial prefix (`{hash}/{base}/{digits}`).

Verdict recorded in D19: closer to obvious than close (~80/20), with the loss being a property the arithmetic-read design deliberately never depended on.

### What phase 3 changed in the doc

§2 layout block (product-root form + bare-store validity note + the `coverage.moc` fix), node invariant (store root vs product root), §3 manifest text (bare-named per product tree), D3/D13 amendment pointers reverted to "unchanged by D19", D16's break-window note updated, D19 rewritten with this rationale. D20/D21/D22/O11 unaffected. Scope revision posted on #299.

docs/design/sparse_coverage.md
```diff
@@ -86,10 +113,14 @@ what zagg consumes:
...
-  classification depends on it. The root alone also carries the manifest and
-  MOC objects.
+- **Node invariant**: below a product root, a node contains *only* digit
+  children (`[1-4]/`), `*.zarr` objects, and the declared leaf-adjacent
+  sidecars — the per-shard stats record (D20) and the sub-shardmap JSON
+  (D22) — with nothing else, ever: the walker's child classification
+  depends on the name set being closed. The product root alone also carries
+  the manifest, MOC objects, and run-level telemetry records (D20); the
+  *store* root of a multi-product directory carries only `{hash}.yaml`
+  registry entries and `{hash}/` product roots (D19).
```
espg: 🤖 *from Claude (review)*

Post-pivot survivor: run-record placement now contradicts D20. This line asserts the *store* root of a multi-product directory carries **only** `{hash}.yaml` registry entries and `{hash}/` product roots, and the layout block (line 64) + the preceding sentence both place the run-level telemetry parquet under the **product** root (`{hash}/`). But D20 (line 628, untouched by phase 3) still reads: "the dispatcher writes a **run-level parquet at the store root**."

Before phase 3 those were the same prefix (single tree, hashes at leaves), so "store root" was unambiguous. Hash-first splits them, and a run maps to exactly one template ⇒ one product tree (D7/D19), so the parquet belongs at the product root — the layout block and this node-invariant line are right, D20 is the stale reference. Worth fixing because it is load-bearing: a store-root parquet would itself violate this very node invariant (a non-registry, non-`{hash}/` object at the store root) and muddy the bare-store-vs-product-directory discrimination rule (line 79: manifest ⇒ bare store; `{hash}.yaml` ⇒ product dir). Suggest updating D20 to say "product root."

espg: 🤖 *from Claude*

Fixed in 680155f. Corrected D20: the dispatcher now "writes a **run-level parquet at the product root** — a run maps to one template ⇒ one product tree (D7/D19), so it lands under `{hash}/`, never the multi-product store root, whose node invariant admits only `{hash}.yaml` + `{hash}/`." This removes the store-root contradiction — the parquet was the last post-pivot survivor placing a non-registry object at the store root, which (as you noted) would itself violate the very node invariant on line 122. Layout block (line 64) and the node-invariant line now agree with D20.

docs/design/sparse_coverage.md
```diff
@@ -47,24 +56,46 @@ contract: [mortie #48](https://github.com/espg/mortie/issues/48) discussion →
...
+    coverage.moc                    manifest (§3) + optional root MOC (§4, O9)
+    <run records (parquet)>      <- run-level telemetry, one row/shard (D20)
+    {sign+base}/{d1}/{d2}/.../   <- one digit per level (D2; digit-chunking is
+                                    the manifest path_grouping param, D21)
+      {window}.zarr/             <- leaf, basename = time window (D23,
+                                    morton-hive/3); `all.zarr` for
+                                    schedule: none (reserved token — lean).
+                                    /1–/2 stores keep {full_id}.zarr and
+                                    {full_id}_{window}.zarr (D3/D13)
+      {window}.stats.json        <- per-shard stats sidecar, sibling (D20,
```
espg, line 77: 🤖 *from Claude (review)*

Lean-vs-ratified: this sidecar line presents the D23 alignment as settled. The leaf line above correctly flags `all.zarr` as `(reserved token — lean)`, but here `{window}.stats.json` / `all.stats.json` is given as the layout with only "naming aligned by D23" and no lean marker. D23's own entry (§8) lists the sidecar alignment under **"Leans recorded, not ratified"** (`{window}.stats.json` / `all.stats.json` — D20 naming; follow-up to #302). So the normative §2 layout block over-states a lean as applied.

Note this also *replaced* the ratified D20 sidecar names — the pre-phase-4 block showed `stats.json` / `stats_{window}.json` — with the unratified `{window}.stats.json` / `all.stats.json` form. As written, §2 now documents the lean naming as the store layout while dropping the currently-ratified D20 names. Consider marking the sidecar line as lean (as `all.zarr` is), or keeping the ratified D20 names with the D23 alignment noted as a lean.

docs/design/sparse_coverage.md
```diff
@@ -395,7 +479,10 @@ write path (§2) is load-bearing; this phase is optimization.
   `{full_id}_{window}.zarr`, underscore separator, split on the first `_`
   — grammar and boundary semantics recorded on the
   [mortie spec page](https://github.com/espg/mortie/issues/62#issuecomment-4986809092)
-  as part of `morton-hive/2`. Reserved (lean, not decided): §7
+  as part of `morton-hive/2`. (Unchanged by D19 — leaf naming survives the
+  product-root design intact. *Naming revised by D23 for `morton-hive/3`*:
```
espg, line 525: 🤖 *from Claude (review)*

Version-scoping: D13's registry entry got the D23 supersession note here, but two sibling references were left unqualified and now read as contradicting D23 for `/3` stores:

- **D3 registry entry** (~line 430): `**D3 — Full morton id at the leaf** (\`{full_id}.zarr\`), self-describing. (Unchanged by D19: …)` — still presents `{full_id}.zarr` as the leaf naming with no D23 note, even though the D3 *bullet* in §2 did get a "*Superseded by D23 for morton-hive/3*" note. The registry entry and the bullet are now inconsistent with each other.
- **D19 registry entry body** (~lines 592–593): describes a product subtree as "bare-named manifest and MOC, `{full_id}.zarr` leaves (D3 and D13's frozen grammar untouched)". A `/3` product subtree's leaves are `{window}.zarr`, so this characterization is now version-incomplete — worth scoping to `/1`–`/2` (the `(D3 and D13's frozen grammar untouched)` parenthetical is about D19 not touching naming, but the sentence still asserts `{full_id}.zarr` as *the* leaf form).

Both would benefit from the same `/3` qualifier D13/D3-bullet received, so the registry stays internally consistent.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Folded in 684229524cc9cf9181813ce3469db2f7e1dcfdc4. The §2 layout block now keeps the ratified D20 sidecar names as the current form and marks the D23 alignment explicitly as a lean: the sidecar line reads `stats_{window}.json  <- per-shard stats sidecar, sibling; ratified D20 naming (stats.json for schedule: none). D23 lean, not ratified: {window}.stats.json / all.stats.json (follow-up to #302)`. This restores the pre-phase-4 D20 names the phase-4 edit had replaced, and the lean now carries the same not-ratified caveat as the `all.zarr` reserved-token marker on the leaf line above — matching D23's own "Leans recorded, not ratified" §8 entry. Alignment/column style preserved.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Folded in 60ac0c727184ea27105b6b9f5883bb19c8d2dca9. Both siblings now carry the same `/3` qualifier D13/the D3-bullet received:

- **D3 registry entry** (~line 430): appended the D3-bullet's supersession note — "*Superseded by D23 for `morton-hive/3` stores*: the basename becomes the time window; the full id stays recoverable from the path and from the stamp/sidecar `shard_key`." The registry entry and the §2 bullet are now consistent.
- **D19 registry body** (~line 593): version-scoped to "`{full_id}.zarr` leaves under `/1`–`/2` grammars (`{window}.zarr` under `/3` — D23; D3 and D13's grammar untouched by D19)". The `/3` leaf form is no longer contradicted, and the D19-does-not-touch-naming parenthetical is preserved.

codespell passes.

espg:
🤖 *from Claude*

## D23 trade study (phase 4, a8ed644): window-only leaf naming — `{window}.zarr`, `morton-hive/3`

espg proposed dropping the morton id from leaf basenames — `{full_id}.zarr` / `{full_id}_{window}.zarr` → `{window}.zarr` — keeping the prefix encoding and formalizing temporal chunking as the naming convention. Ratified in-session 2026-07-20; the session analysis recorded here for the thread.

### What it buys

- **Completes the axis separation D19 started**: product = root prefix, space = digit path, **time = basename** — each identity axis in exactly one place, none encoded twice. The morton id currently appears in both the path and the leaf name; this is the same redundancy observation that opened #296, applied to its final instance.
- **The naming convention *becomes* the temporal chunking contract** instead of a suffix bolted onto an id: listing a shard node returns the temporal inventory directly (`2019.zarr`, `2020.zarr`, …) — which is exactly what append planning and gridlook time-series discovery ask.
- Shorter keys; window semantics stop hiding behind a "split on first `_`" parsing rule.

### What it costs (four items, all modest)

1. **D13's grammar is frozen on the mortie spec page as `morton-hive/2`** — this un-freezes it: spec bump to `/3`, a second naming dialect for readers. Mitigations: D6 versioning exists precisely for this; the parse is trivial; it's additive and spec-discriminated (existing `/1`/`/2` stores stay valid forever), not a migration. *espg: "don't care, that's easy."*
2. **The name==path self-check dies.** With `{full_id}.zarr` an auditor verifies placement from the key alone; with `{window}.zarr` identity verification needs the stamp attrs / D20 sidecar `shard_key` (one GET per leaf in an fsck), and D3's "unambiguous if moved" softens to "recoverable from attrs." *espg: no real moved-case exists — the closest is downloading a shard, which reduces to a downloader that materializes the prefix tree into folder names.*
3. **`schedule: none` needs a reserved token** — the bare-leaf case has no window to name. Proposal `all.zarr` (reads as "all time," cannot collide with the digit-shaped window grammar; `none.zarr` matches the schedule literal but reads worse). *espg: the hairiest item — recorded as a **lean, not ratified**.*
4. **Sidecar naming should align**: `{window}.stats.json` / `all.stats.json` sorts adjacent to its leaf. PR #302 merged with the ratified D20 names (`stats.json` / `stats_{window}.json`), so this is a follow-up (in flight, `claude/297-window-sidecar-naming`) — implemented behind a single naming seam so the token stays a one-constant change pending ratification. *Also recorded as a lean.*

### Rejected alternative — time as a path level

If window-scoped *operations* (expire old windows) were the goal, time would want to be a path component (`{hash}/{window}/{morton…}`), because S3 lifecycle can't filter on basenames. But that axis order duplicates the morton tree per window and shatters the dominant read — a time series at a location — across W prefixes. Reads dominate window expiry by a wide margin, so space stays contiguous and time stays at the leaf; window expiry is an enumerate-or-tag operation under either naming, so `{window}.zarr` loses nothing there.

### Verdict

Favor — ~70/30, versus hash-first's ~80/20: the win is formalization and elegance rather than operational capability, but it's the right long-term shape, everything is pre-1.0, and landing it in the same `morton-hive/3` bump as the D19 product-root introduction means readers absorb one new spec version, not two. The open details are exactly the two leans (`all.zarr` token, sidecar alignment) plus re-freezing the `/3` grammar on mortie#62 (heads-up posted there).

espg:
🤖 *from Claude*

## D19 revision 2 + D24 (phase 5, a422270): named roots, semantics-only hash, resolution polymorphism

Session ratifications recorded (espg, 2026-07-20):

**D19 rev 2 — the name is the address, the hash is the integrity check.** Product roots become human-readable `{name}/` (gridlook enumerates `store_root/{name}/morton_hive.json` directly — no translation layer). The hash survives demoted to metadata, and it hashes **only the output-defining subset**: the `aggregation` block, `data_source` semantics, grid *type + indexing scheme*. Excluded as packaging: cell order (→ D24), parent/shard order, chunking/sharding, worker size, streaming mode (merge-vs-spill is `np.isclose` and shares one store, actual mode recorded per-run), read knobs. Rationale: hashing the whole template made o8-vs-o9 different products, which is wrong (aggregation composes) and blocks mixed-order processing. The `semantic_hash` is a frozen manifest key (name reuse with different semantics refuses up front, like any frozen-key mismatch); the literal template is deliberately *not* the product-level record (it carries run-varying packaging) — the product root holds the canonical semantic core (`aggregation.yaml`), each run archives its literal template with its run record, sidecars carry `run_id` to join. This formalizes the `spatial_signature`/`output_field_signature` seam (#89) the code already has.

**D24 — cell order is a resolution axis, not identity.** Aggregation composes across orders (exactly for count/sum/min/max; `np.isclose` for t-digest — same epistemic class as merge-vs-spill). One product tree may carry regionally heterogeneous resolution (o19 polar shards, o17 mid-latitude). The design had already committed to the pillars: D22's rollup==direct *is* the composition claim; D16 chose morton words because they carry order; D11's `role` attr anticipated coarse source; #217's mergeable-reducer machinery provides the merge laws. Frozen with it: per-product **composability class** (`exact|approximate|none`; `none` pins cell order); **one resolution per (shard, window)** — same-order rerun = D4 replacement, but **cross-order writes into an occupied leaf refuse with a useful error** (espg-directed guard); manifest `cell_order`/`shard_order` become declared allowed sets. Coarse source and sweep overviews unify — the pyramid is the store's resolution axis, partially materialized. Reader support gated on mortie#116 → moczarr#8; no schema bump needed.

**D20 additions**: run-record names go timestamp-first (`stats_{timestamp}_{run_id}.parquet` — lexicographic = chronological, key-range time queries); user scoping stays the `invoked_by` column; sidecars gain `run_id` + `semantic_hash`. Implementation follow-up dispatched.

docs/design/sparse_coverage.md
```diff
@@ -419,6 +523,259 @@ write path (§2) is load-bearing; this phase is optimization.
...
+  replacement; **writing a different cell order into an occupied leaf
+  refuses with a useful error** (espg-directed) — intentional
+  re-resolution means rerunning at a parent order that isn't occupied, or
+  explicitly clearing the leaf. (3) Manifest `cell_order` and
+  `shard_order` become declared allowed sets/ranges (§3); per-leaf truth
+  is the morton words + MOC. (4) Coarse source and sweep-built overviews
+  unify — the same multi-order tree, distinguished only by `role`/
+  provenance attrs; the pyramid is the store's resolution axis, partially
+  materialized. Reader support is gated on mortie#116 mixed-order morton
+  (tracked as moczarr#8); the schema needs no bump — this is what the
```
espg, line 822: 🤖 *from Claude (review)*

**§5 cross-resolution join still assumes a scalar `cell_order`/`shard_order`, not reconciled with D24's regionally-heterogeneous resolution.**

D24 (this entry) establishes that "one product tree may carry **regionally heterogeneous resolution** (e.g. o19 cells in polar shards, o17 mid-latitude)" and, in point (3), turns manifest `cell_order`/`shard_order` into "declared allowed sets/ranges … per-leaf truth is the morton words + MOC." But §5's cross-resolution join (lines 347–351, out of this hunk) still reads them as single values:

- L347: "Read the *target* dataset's manifest → its `cell_order`/`shard_order`."
- L348–351: "Truncate the fine observation's morton decimal string **to the target's cell order** … Truncate further **to the target's shard order** → its shard's hive path."

For a polymorphic store the manifest now returns a *set*, so "the target's cell order" / "the target's shard order" is underspecified — the region's actual order is per-leaf truth, which the up-front scalar-truncation arithmetic in steps 2–3 can't select without consulting the MOC/leaf.

This is a doc-coherence gap, not a broken decision: D24 gates reader support on mortie#116 mixed-order morton / moczarr#8 (L776–777), so the mixed-order read path is legitimately deferred. But as written §5 reads as *contradicting* D24 rather than pointing at it. A one-line forward pointer — in §5 or in D24 point (3) — that the join generalizes to the allowed-set case behind the mortie#116 gate would close it. Homogeneous-resolution stores (the singleton-set case) are unaffected.

espg: 🤖 *from Claude*

Folded in d2519df. Added a forward pointer right after the §5 cross-resolution join steps rather than editing the numbered arithmetic (which stays correct for the homogeneous / singleton-set case):

> (Under D24 a target's orders may be regionally heterogeneous — the declared sets bound them; the *leaf's* actual orders come from its morton words / MOC, and the truncation predicate is unchanged. Mixed-order reader support is gated on mortie#116 / moczarr#8.)

This makes §5 point *at* D24 (declared allowed sets + per-leaf MOC truth, reader path deferred behind mortie#116 / moczarr#8) instead of reading as a contradiction of it; homogeneous stores are unaffected.

espg:
🤖 *from Claude*

## Phase 6 (bb2780e): espg-directed spec-gap additions

From the session's "what's still missing" review, with espg dispositions:

- **O12 — retention/expiration (recorded as proposal)**: product-level retention = one *prefix* lifecycle rule per `{name}/`; window-level retention = *tag*-scoped rules (`ephemeral=true` filters expiration; keep-alive = retag; never access-refresh or self-copy "touch"). Open: tag vocabulary, who tags at write, manifest-vs-tags for pin records.
- **O13 — publication profile (shape espg-ratified)**: publishing to a public bucket = **prefix copy minus operator telemetry, telemetry replaced by a simple roll-up summary** preserving cost-estimation utility (totals; no `invoked_by` ARNs/request ids). Open: summary schema, placement, sub-shardmap treatment.
- **Debris collection** added to D22 as an optional fifth, audit-class sweep family (unstamped prefixes past a horizon; never load-bearing).
- **Concurrency contract** recorded once in §2 as a derived property: shard-disjoint concurrent runs share nothing; same-(shard,window) degrades to D4 last-writer-wins; rollup staleness stays D22-detectable; cross-product shares nothing.
- **§8.3 standing test obligations** — a consolidated index of every normative test claim in the registry (rollup==direct, merge-fold algebra, semantic-hash canonicalization, manifest guard, D24 guards, naming dialects, golden byte pins, stamp semantics) so CI coverage is auditable against the spec.
- **D19 refinements** (espg-confirmed): the manifest `semantic_hash` is truth and `aggregation.yaml` is a D9-class derived cache; the hash is stored full-length with git-style short-fingerprint display; product names are URL-safe by requirement (charset frozen with the mortie#62 grammar).
- **D24 open sub-point recorded, deferred**: `none`-field handling — lean is per-field exclusion from coarser materializations (not min-semantics product pinning), and `none` likely allowed with a loud warning; not ratified, pending discussion.

Normative/rationale split: proposed as an explicit #305 acceptance criterion (comment there).

docs/design/sparse_coverage.md
```diff
@@ -104,6 +148,17 @@ what zagg consumes:
...
   objects at digit nodes, no shared mutable state, no create-group races
   across 2,000 workers.
+- **Concurrency contract** (a derived property, recorded 2026-07-20 —
+  assembled here so it exists in one place): concurrent runs against one
+  product are supported whenever they are **shard-disjoint** — they share
+  no objects (D4/D5), their run records are distinct by name
+  (timestamp + run id, D20), and the manifest precheck admits both
+  (frozen keys match by construction under D19). Same-(shard, window)
+  concurrency degrades to D4 semantics — last writer wins; the loser's
+  partial output is stamped-over or remains ignorable debris. Interior
```
espg: 🤖 *from Claude (review)*

The same-`(shard, window)` clause borrows the **sequential-retry** debris model to describe **simultaneous** co-writers, and slightly overpromises for the differing-content case.

"last writer wins; the loser's partial output is stamped-over or remains ignorable debris" reads cleanly when a torn worker leaves an *unstamped* prefix that a later retry overwrites. But two runs targeting the identical `(shard, window)` write to the **same** `.zarr/` prefix, which is **multiple objects** (one per array under the ShardingCodec — D17; the ragged object — D18; root attrs; coverage sidecar), and S3 PUTs are not transactional across those objects. So:

- If both writers produce byte-identical content (the normal rerun / dedup-collision case, D19's "outputs byte-identical in practice"), interleaving is harmless and "last writer wins" holds trivially.
- If the two concurrent runs have **different** content (e.g. same product/`semantic_hash`, same shard+window, but different catalogs mid-append — both pass the D19 pre-check because neither sees the other's not-yet-written sidecar), object-level last-writer-wins per key can interleave arrays A/B across writers and still end **stamped** — a stamped-but-internally-inconsistent leaf, which is neither "the loser's partial output stamped-over" nor "ignorable debris."

Framing suggestion: state that same-leaf concurrency is clean only when the content is identical (idempotent replacement); differing simultaneous co-writers to one multi-object leaf are outside the write-once guarantee. The clean either/or (`stamped-over | ignorable debris`) is the *sequential* retry model, not simultaneous co-writers. Not a blocker — the supported contract is explicitly shard-disjoint — but the degraded-case sentence claims a bit more than the debris model gives.

espg:
🤖 *from Claude (review)*

Fresh-context adversarial review of **phase 6 only** (bb2780e), against the full file at that commit. Phases 1–5 not re-reviewed.

**Verified clean:**
- **§8.3 test-obligations index** (line-by-line against cited entries): rollup==direct + sweep-idempotence + nothing-load-bearing (D22 / §7); merge-fold (D20); semantic-hash canonicalization + name-grammar (D19, incl. the base-component exclusion at §2 and the URL-safe charset refinement); manifest guard (§3, incl. `path_grouping` absent⇒1 normalization and `cell_order`/`shard_order` allowed-set membership); D24 guards (occupied-leaf cross-order refusal, same-order D4 replace, class derivation from merge-law flags); D23 naming dialects (`/3` round-trip, `/1`–`/2` byte-unchanged, spec-string discrimination); golden pins (D18 VLenArray framing, O8 bit convention / PR #208 vectors); stamp/debris (D4). No invented obligations, no misattributed D-numbers.
- **Provenance discipline**: O12 (proposal), O13 (shape ratified / details open), D24 sub-point (leans, "not ratified"), D19 refinements (espg-confirmed), concurrency contract ("derived property recorded", not a new decision), GC fifth family (optional, audit-class) — all marked correctly, no overclaim.
- **O12/O13 consistency**: O12 product-level = prefix rule (matches D19's prefix-op rationale), window-level = tag mechanism (a new sub-product granularity, not a contradiction). O13 telemetry-stripping is complementary to D20's "sidecars omit account-identifying fields", going one step further (per-product totals only).
- **Numbering/format/links**: O12/O13 next-free; §8.3 placed after §8.2, before §9; header status line updated to "O11–O13 opened, §8.3 … added"; entry style matches O10/O11.

**One finding** posted inline (concurrency contract, same-`(shard,window)` degraded case).

**One soft observation (pre-existing, optional, out of phase-6 scope to fix here):** the new §8.3 "Merge-fold algebra (D20)" line lists "associativity/commutativity of the stats record fold" as a flat obligation. That faithfully mirrors D20's own "only associative stats (… t-digests …)" framing, so it doesn't contradict its cited entry — but D24 elsewhere notes t-digest merge is order-dependent (`np.isclose`, not exact). A literal associativity/commutativity test would need to exclude t-digest or use tolerance. Flagging only so whoever writes the CI test doesn't take the index line as exact-for-all-stats; the underlying D20↔D24 nuance predates this phase.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Folded in de8c645. The degraded-case sentence no longer borrows the sequential-retry debris model for simultaneous co-writers. It now states same-`(shard, window)` concurrency is **out of contract**: sequential re-runs give clean D4 last-writer-wins replacement, but a leaf is several objects (per-array ShardingCodec objects D17, ragged object D18, root attrs, coverage sidecar) and S3 PUTs aren't transactional across them, so two simultaneous different-content co-writers can interleave those objects into a stamped-but-torn leaf — neither replacement nor ignorable debris. Detection is content verification (the O11 hash) or a re-run; the supported contract stays shard-disjoint.

Same commit also applied your §8.3 observation: the "Merge-fold algebra (D20)" line now qualifies `merge-of-children == direct` as exact for count/sum/min/max and up to the approximate class (`np.isclose`) for order-dependent t-digest members, cf. D24.

espg:
🤖 *from Claude*

## Phase 7 (3ccc3aa): final rulings folded — the amendment is complete

espg rulings from the session, now in the doc:

- **O8 addendum**: the PR #208 constants stand as-shipped — bitmap bit convention and root-MOC range ordering are **contract** (golden-pinned); zstd level 3 is **non-normative**. Closes the #276 item-5 flag.
- **O11 → RESOLVED, adopted as proposed**: per-array sha256 over decoded values — "per-array" pinned as **per named zarr array (variable) in the leaf**, over its full decoded contents, packaging-invisible; computed **at write**, in-worker; **all arrays including coordinates**; exact bytes, no float tolerance. Three recorded jobs: D19 verification half, mismatch localizer, torn-leaf detector for the §2 concurrency contract.
- **D23 tokens ratified**: `schedule: none` ⇒ **`all`** (`all.zarr`, excluded from the window grammar forever on mortie#62), and the sidecar aligns to **`{window}.stats.json` / `all.stats.json`** for `/3` stores — both one-constant flips through the PR #307 seam, which shipped with exactly these values.

Remaining open by design: O12/O13 (proposals as recorded), the D24 `none`-field sub-point (deferred pending discussion), and gridlook/mortie#116-gated reader work. Per espg's direction the amendment is now **frozen for merge**; the #201 plan was posted against it.

docs/design/sparse_coverage.md
```diff
@@ -419,6 +548,288 @@ write path (§2) is load-bearing; this phase is optimization.
...
+  records. What a shared spatial tree would have offered — one prefix
+  spanning all products — serves no planned workload (§5 reads are
+  arithmetic). **Catalog identity lives in the sidecar, never the name**:
+  granule count + sha256 of sorted granule ids + zagg version;
+  dedup/`has_run` consults the computed path, the `semantic_hash`, *and*
+  the sidecar catalog identity (a catalog-grown shard is "stale", not
+  "hit"). Immutable-provenance naming (product root
+  `{name}+{catalog-hash}/`) stays an opt-in for frozen-catalog archival
+  runs. The output content hash that makes dedup *verifiable* is O11
+  (proposal; it complements the semantic hash — "intended identical" vs
```
espg: 🤖 *from Claude (review)*

Stale ratification state: the D19 body still calls the content hash **O11 (proposal)**, but phase 7 resolved it. The header now reads "O3/O11 resolved" and the §8.2 entry is "**O11 — logical content hash of outputs: RESOLVED — adopted as proposed**". This is the one surviving "proposal" reference to the content hash in D19's prose — recommend dropping "(proposal)". The "intended identical vs actually byte-identical" gloss stays correct (it's job (1) in the resolved O11's "three jobs").

docs/design/sparse_coverage.md
```diff
@@ -466,14 +882,138 @@ write path (§2) is load-bearing; this phase is optimization.
...
+  time dimension inside each node (windows-as-nodes would shatter the
+  dominant time-series read); the **spatial digit axis is never tree
+  nodes** — a node per morton digit recreates the D5/D12 metadata storm
+  client-side. The Rust-backed fast `open_datatree` path applies to
+  zarr-native hierarchies, which the hive tree deliberately is not
+  (D12) — it plugs in via D12's escape hatch instead: the sweep's
+  optional interop materialization can generate a consolidated,
+  `open_datatree`-able hierarchy *as a derived cache* for non-moczarr
+  consumers, while the MOC-first opener remains the truth path (the
+  derived-views principle, third instance). Phasing: (1) `open_hive` →
```
espg: 🤖 *from Claude (review)*

"Third instance" collides with D16. D16 already claims that slot — "**Third instance** of the derived-views principle (§6 dense views, D9 caches)" — enumerating (1) §6 dense views, (2) D9 caches, (3) D16 morton-only fabrication. O14's MOC-first-truth-path-with-derived-cache is therefore a *fourth* instance, not a third. Recommend "fourth instance" (or drop the ordinal to avoid having to renumber again).

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Fixed in a254bba — dropped "(proposal)"; the D19 body now reads "The output content hash that makes dedup *verifiable* is O11 (resolved — adopted; …)", and the "intended identical vs actually byte-identical" gloss stays as you noted.

While in this file I also folded the reviewer's for-the-record §4 flag (no thread of its own): the tier-0 pad-sentinel sentence (~line 271) still framed the choice as open ("lean is null … repetition-padding the viable alternative … frozen with the mortie-side spec (O8)"). Since O8 is RESOLVED with JSON null decided, it now reads "the pad sentinel is null (base-0 words / JSON `null`), decided with the mortie-side spec (O8); repetition-padding would also have worked … but null is the frozen choice." — 5a0cfd4.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Fixed in f4fe0de — O14 now reads "the derived-views principle, fourth instance". D16 keeps the third slot (1: §6 dense views, 2: D9 caches, 3: D16 morton-only fabrication); O14's MOC-first-truth-path-with-derived-cache is the fourth.

PR #307, opened 2026-07-20
D23 sidecar naming seam + temporal cost-block status (refs #297/#298)

espg:
Refs #297, Refs #298. Design record: D23 in `docs/design/sparse_coverage.md` on PR #306's branch (espg-proposed and ratified in-session, 2026-07-20).

Two small follow-ups to the merged PR #302, directed on the PR threads:

## 1. D23 sidecar-naming seam (`{window}.stats.json`)

D23 makes `morton-hive/3` leaf basenames the time window alone (`{window}.zarr`; `all.zarr` reserved for `schedule: none`), and aligns the stats sidecar to `{window}.stats.json` / `all.stats.json`.

**Choice: (a) — a spec-keyed seam, no change to what current writers emit.** No `/3` writer exists yet, so renaming today's `/2` output would fork the frozen `/1`–`/2` grammar for nothing. Instead:

- `zagg.windows.leaf_name_v3(window)` — the `/3` leaf-basename seam (`{window}.zarr`, `all.zarr` for `None`) the issue #299 writer flips to; `leaf_name` stays the frozen `/1`–`/2` grammar.
- `zagg.windows.SCHEDULE_NONE_TOKEN = "all"` — **the `all` token is a recorded LEAN, not ratified** (see D23 §8.1 "leans recorded"); it is a single constant, so ratifying a different token is a one-line change ("token pending ratification, one-constant change").
- `zagg.telemetry.sidecar_key(leaf_name, spec)` (+ `sidecar_path`/`write_sidecar`/`read_sidecar` pass-through): `spec == "morton-hive/3"` derives the sidecar from the leaf stem (`{window}.stats.json`, `all.stats.json`) — the token flows from the leaf name itself, one source; any other/absent spec keeps the frozen legacy names (`stats.json` / `stats_{window}.json`) byte-identically. The #299 flip is a spec switch at the writer call sites, not a rename.

## 2. Temporal cost-block status (verification, no code)

Verified against `main` (post PR #303 merge): the temporal dispatch path **already carries the full cost block** — `_run_lambda_events` computes the pre-invoke `max_cost_usd` ceiling and stamps `summary["cost"] = {max_cost_usd, estimated_cost_usd, actual_cost_usd}` with the billed-duration actual rollup, exactly like the spatial (and now raster) paths. The fold commit "note temporal path out of scope for estimator (issue #298)" concerns only the Phase-3 **estimator stub** (`estimated_cost_usd` stays `None` on temporal because there is no shardmap to estimate from — documented in `estimate_cost_usd`'s docstring); `max` and `actual` are real values there. Nothing to add; recording the status here per the directive.

## Phases

- [x] Phase 1 — the D23 naming seam (`windows.leaf_name_v3` + `SCHEDULE_NONE_TOKEN`, spec-keyed `telemetry.sidecar_key` and sidecar I/O pass-through) with tests pinning the `/3` mapping, the frozen legacy mapping, and the token's single-source derivation
- [x] Temporal cost-block verification (document-only, above)

## Testing

`tests/test_telemetry.py`: `/3` mapping (`2019.zarr` → `2019.stats.json`; `leaf_name_v3(None)` → `all.zarr` → `all.stats.json`), legacy specs and absent-spec unchanged, malformed leaf names raise, `/3` write/read round-trip on a local store (and the legacy reader not seeing the `/3` name — spec selects). Full suite green locally; the two pre-existing environment-only items noted on PR #302 are unchanged.

## Questions for review

1. **`all` token**: implemented as the single `SCHEDULE_NONE_TOKEN` constant per the recorded lean — flip the constant if ratification picks another spelling (`none`, `full`, …).
2. **Seam shape**: `sidecar_key` takes the manifest `spec` string rather than a bool, so future spec bumps route through the same switch. If you'd rather the `/3` naming ride `zagg.windows` entirely (sidecar name computed next to `leaf_name_v3`), easy to move before #299 builds on it.

### DIFF
--- a/src/zagg/telemetry.py
+++ b/src/zagg/telemetry.py
@@ -364,14 +364,34 @@ def write_run_parquet(
 # ---------------------------------------------------------------------------
 
 
-def sidecar_key(leaf_name: str) -> str:
-    """Sidecar object name for a leaf zarr basename.
+#: Manifest ``spec`` string selecting the D23 window-only naming grammar.
+#: ``/1``/``/2`` (and an absent spec) keep the frozen legacy sidecar names.
+SPEC_V3 = "morton-hive/3"
 
-    Bare leaves get :data:`SIDECAR_NAME`; windowed leaves (issue #246) get
+
+def sidecar_key(leaf_name: str, spec: str | None = None) -> str:
+    """Sidecar object name for a leaf zarr basename, keyed by store spec.
+
+    Legacy (``spec`` absent / ``morton-hive/1`` / ``/2``): bare leaves get
+    :data:`SIDECAR_NAME`; windowed leaves (issue #246) get
     ``stats_{window}.json`` — a hive node directory holds every window's leaf
     of its one shard, so a bare ``stats.json`` would self-clobber across
-    windows. Mirrors the ``{full_id}_{window}.zarr`` leaf naming.
+    windows. Mirrors the ``{full_id}_{window}.zarr`` leaf naming. Frozen: what
+    every current writer emits, unchanged.
+
+    ``morton-hive/3`` (:data:`SPEC_V3`, D23 — window-only leaf naming, no
+    writer yet): the sidecar is the leaf stem + ``.stats.json`` —
+    ``{window}.stats.json``, and ``all.stats.json`` for the ``schedule:
+    none`` :data:`~zagg.windows.SCHEDULE_NONE_TOKEN` leaf. Derived from the
+    leaf basename itself, so the token has ONE source
+    (:func:`zagg.windows.leaf_name_v3`) and the issue #299 writer flip is a
+    spec switch here, not a rename.
     """
+    if spec == SPEC_V3:
+        stem = leaf_name.removesuffix(".zarr")
+        if not stem or stem == leaf_name:
+            raise ValueError(f"{leaf_name!r} is not a leaf zarr name")
+        return f"{stem}.stats.json"
     from zagg.windows import split_leaf_name
 
     _full_id, window = split_leaf_name(leaf_name)
@@ -381,13 +401,13 @@ def sidecar_key(leaf_name: str) -> str:
     return f"{stem}_{window}.{ext}"
 
 
-def sidecar_path(leaf_path: str) -> str:
+def sidecar_path(leaf_path: str, spec: str | None = None) -> str:
     """Absolute path of a leaf's stats sidecar (sibling of the ``.zarr``)."""
     prefix, _, name = leaf_path.rstrip("/").rpartition("/")
-    return f"{prefix}/{sidecar_key(name)}"
+    return f"{prefix}/{sidecar_key(name, spec)}"
 
 
-def write_sidecar(leaf_path: str, record: dict, **store_kwargs) -> None:
+def write_sidecar(leaf_path: str, record: dict, spec: str | None = None, **store_kwargs) -> None:
     """PUT ``record`` as the leaf's stats sidecar (success path only, #297)."""
     import obstore
 
@@ -395,11 +415,13 @@ def write_sidecar(leaf_path: str, record: dict, **store_kwargs) -> None:
 
     prefix, _, name = leaf_path.rstrip("/").rpartition("/")
     obstore.put(
-        open_object_store(prefix, **store_kwargs), sidecar_key(name), json.dumps(record).encode()
+        open_object_store(prefix, **store_kwargs),
+        sidecar_key(name, spec),
+        json.dumps(record).encode(),
     )
 
 
-def read_sidecar(leaf_path: str, **store_kwargs) -> dict | None:
+def read_sidecar(leaf_path: str, spec: str | None = None, **store_kwargs) -> dict | None:
     """The leaf's stats sidecar record, or ``None`` when absent."""
     import obstore
     from obstore.exceptions import NotFoundError
@@ -408,7 +430,9 @@ def read_sidecar(leaf_path: str, **store_kwargs) -> dict | None:
 
     prefix, _, name = leaf_path.rstrip("/").rpartition("/")
     try:
-        data = obstore.get(open_object_store(prefix, **store_kwargs), sidecar_key(name)).bytes()
+        data = obstore.get(
+            open_object_store(prefix, **store_kwargs), sidecar_key(name, spec)
+        ).bytes()
     except (FileNotFoundError, NotFoundError):
         return None
     return json.loads(bytes(data))
--- a/src/zagg/telemetry.py
+++ b/src/zagg/telemetry.py
@@ -368,6 +368,10 @@ def write_run_parquet(
 #: ``/1``/``/2`` (and an absent spec) keep the frozen legacy sidecar names.
 SPEC_V3 = "morton-hive/3"
 
+#: Specs that key the frozen legacy sidecar names (bare / ``stats_{window}.json``).
+#: An absent spec (``None``) is a ``morton-hive/1`` store by definition.
+_LEGACY_SPECS = (None, "morton-hive/1", "morton-hive/2")
+
 
 def sidecar_key(leaf_name: str, spec: str | None = None) -> str:
     """Sidecar object name for a leaf zarr basename, keyed by store spec.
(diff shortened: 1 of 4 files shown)


src/zagg/windows.py
```diff
@@ -221,6 +221,29 @@ def windows_intersecting(
...
+
+    Space lives in the digit path, time in the basename — the morton id no
+    longer repeats in the leaf name. ``None`` (``schedule: none``) is the
+    reserved :data:`SCHEDULE_NONE_TOKEN` leaf. No ``/3`` writer exists yet
+    (issue #299 flips the writers to this seam); ``/1``/``/2`` stores keep
+    :func:`leaf_name`'s frozen grammar forever (D6 versioning).
+    """
+    if window is None:
+        return f"{SCHEDULE_NONE_TOKEN}.zarr"
+    validate_label(window)
```
espg, line 249: 🤖 *from Claude (review)*

**Reserved `all` token collides with an explicit window label; `leaf_name_v3` does not enforce the invariant D23 relies on.**

D23's recorded rationale for `all.zarr` is that it "cannot collide with the digit-shaped window grammar." That holds only for the *generative* schedules (`yearly`/`monthly`/`daily` are all digits). But `leaf_name_v3(window)` validates via `validate_label(window)` with the **default `schedule="explicit"`**, whose grammar is `_LABEL_RE = [0-9A-Za-z-]{1,32}` — which admits the literal `"all"`. So an explicit windows list may legally declare `{label: "all", ...}`, and then:

- `leaf_name_v3("all")` -> `"all.zarr"` == `leaf_name_v3(None)` -> `"all.zarr"` (the reserved `schedule: none` leaf), and
- their sidecars `all.stats.json` collide too.

Under `/1`/`/2` the full_id prefix made the namespace collision-free; under `/3` the window *is* the whole basename, so the reserved token shares the namespace with author-supplied explicit labels. The `/3` writer (#299) would inherit this hole silently.

Fix: enforce the reservation at this seam — e.g. `leaf_name_v3` (or `validate_label`) should reject `window == SCHEDULE_NONE_TOKEN`, so a declared explicit label equal to the token raises rather than aliasing the no-schedule leaf. (Distinct from the PR's Q1, which is about *which spelling* to ratify, not enforcing the token against explicit labels.)

src/zagg/telemetry.py
```diff
@@ -364,14 +364,34 @@ def write_run_parquet(
...
+
+    ``morton-hive/3`` (:data:`SPEC_V3`, D23 — window-only leaf naming, no
+    writer yet): the sidecar is the leaf stem + ``.stats.json`` —
+    ``{window}.stats.json``, and ``all.stats.json`` for the ``schedule:
+    none`` :data:`~zagg.windows.SCHEDULE_NONE_TOKEN` leaf. Derived from the
+    leaf basename itself, so the token has ONE source
+    (:func:`zagg.windows.leaf_name_v3`) and the issue #299 writer flip is a
+    spec switch here, not a rename.
     """
+    if spec == SPEC_V3:
```
espg, line 399: 🤖 *from Claude (review)*

**Unknown/future spec silently falls back to LEGACY sidecar names — a silent wrong-name, not a loud failure.**

`sidecar_key` (and the `sidecar_path`/`write_sidecar`/`read_sidecar` pass-throughs) select V3 by exact equality `spec == SPEC_V3`; every other non-None spec drops to the legacy branch. That makes the else-branch a catch-all, so:

- a future `morton-hive/4` that *also* intends window-only naming would silently get `stats.json` / `stats_{window}.json` sidecars, and
- a typo (`"Morton-Hive/3"`, a trailing space) reads/writes the *wrong* sidecar name with no error.

The real hazard is a writer/reader spec mismatch: a `/3`+ store whose writer emits `{window}.stats.json` but is read back with a spec this function doesn't recognize -> `read_sidecar` keys legacy `stats.json`, gets `NotFound`, returns `None` — the sidecar looks absent rather than raising. For a versioned grammar switch, silently defaulting an *unrecognized* spec to the frozen legacy grammar is the risky failure mode.

Consider validating `spec` against a known set (`{None, "morton-hive/1", "morton-hive/2", SPEC_V3}`) and raising on anything else, so a future spec bump is a loud, deliberate change here. (The PR's Q2 is about seam *shape* — string vs bool, or moving naming into `windows` — not the failure mode on an unrecognized spec.)

src/zagg/telemetry.py
```diff
@@ -364,14 +364,34 @@ def write_run_parquet(
...
+        stem = leaf_name.removesuffix(".zarr")
+        if not stem or stem == leaf_name:
+            raise ValueError(f"{leaf_name!r} is not a leaf zarr name")
+        return f"{stem}.stats.json"
```
espg, line 410: 🤖 *from Claude (review)*

**V3 branch is looser than legacy: it never validates the stem against the label grammar, so a malformed leaf yields a malformed sidecar key instead of raising.**

The legacy branch routes through `split_leaf_name` -> `validate_label`, so a bad window label raises. The V3 branch only guards `not stem or stem == leaf_name` (empty stem / missing `.zarr`) and otherwise returns `f"{stem}.stats.json"` verbatim. Nothing checks that `stem` is a legal label (`[0-9A-Za-z-]{1,32}`), so:

- `sidecar_key("a/b.zarr", SPEC_V3)` -> `"a/b.stats.json"` — an embedded `/` escapes the leaf directory into a different prefix;
- `sidecar_key("foo_bar.zarr", SPEC_V3)` -> `"foo_bar.stats.json"` — `_` is forbidden by the frozen charset yet passes here.

The PR frames the token as flowing "from the leaf name itself," but that trust is unchecked: a malformed leaf silently produces a malformed sidecar key. The public entry points feed `name` via `rpartition("/")` so the slash case is masked in practice, but `sidecar_key` is public and tested directly, and the strictness asymmetry with legacy is a latent trap for the #299 writer.

Fix: run `validate_label(stem)` in the V3 branch (permitting `SCHEDULE_NONE_TOKEN`), so the two branches reject the same malformed inputs. A test asserting a `/`- or `_`-bearing name raises under `SPEC_V3` would pin it.

src/zagg/windows.py
espg: 🤖 *from Claude*

Folded. `leaf_name_v3` now raises `ValueError` when `window == SCHEDULE_NONE_TOKEN` (naming D23 and the reservation) — an explicit label equal to `all` can no longer alias the no-schedule leaf; only `leaf_name_v3(None)` owns `all.zarr`. Added `TestLeafNameV3.test_explicit_label_equal_to_reserved_token_raises`. Fix in c8a5e0a.

src/zagg/telemetry.py
espg: 🤖 *from Claude*

Folded. `sidecar_key` now enumerates: `spec == SPEC_V3` -> v3 names, `spec in _LEGACY_SPECS` (`(None, "morton-hive/1", "morton-hive/2")` module tuple) -> legacy, anything else raises `ValueError`. A future `morton-hive/4` or a typo (`"Morton-Hive/3"`) is now a loud failure, not a silent legacy fallback that reads the wrong sidecar name. Added `test_unknown_spec_raises` (unknown raises; /1 and /2 still legacy). Fix in e313f83.

src/zagg/telemetry.py
espg: 🤖 *from Claude*

Folded. The v3 branch now runs `validate_label(stem)` (imported from `zagg.windows`) after the `.zarr`/empty guard, matching the legacy branch strictness: `sidecar_key("a/b.zarr", SPEC_V3)` and `sidecar_key("foo_bar.zarr", SPEC_V3)` now raise on the grammar rather than returning a malformed key. The `all` schedule-none token satisfies the explicit grammar, so it keeps passing. Added `test_v3_malformed_stem_raises`. Fix in 765d95f.

espg:
🤖 *from Claude*

Both Questions-for-review are now answered by espg rulings (recorded in PR #306 phase 7): the `schedule: none` token is **ratified as `all`** — `SCHEDULE_NONE_TOKEN` stays exactly as shipped — and the `/3` sidecar alignment (`{window}.stats.json` / `all.stats.json`) is **ratified as implemented** by this PR's seam. No changes needed; the seam shape (spec-string-keyed) also stands. This PR is ready to merge as-is.

PR #309, opened 2026-07-20
stats records: timestamp-first names, run_id, semantic_hash (refs #297/#299)

espg:
Refs #297, Refs #299

Three small, mechanical follow-ups to the telemetry machinery PR #302 merged, per the espg-ratified D19 rev 2 + D20 (+D24) design record in `docs/design/sparse_coverage.md` on PR #306's branch — phase-5 record comment: https://github.com/englacial/zagg/pull/306#issuecomment-5024857031.

## What changed

1. **Run-record names go timestamp-first** — `run_parquet_key` now emits `stats_{timestamp}_{run_id}.parquet` (was `stats_{run_id}_{timestamp}.parquet`), so a lexicographic listing of `stats_*.parquet` at the store root is chronological and time-range queries can prune on the key alone (D20). Schema/content unchanged; `SCHEMA_VERSION` stays 1 (unreleased).
2. **Sidecars carry `run_id`** — the per-shard record gains a `run_id` field, threaded exactly like `invoked_by`: the dispatcher generates it once per run, stamps it through the invoke payload, and the worker copies it verbatim (`src/zagg/telemetry.py::build_record`, the three `build_record` sites in `deployment/aws/lambda_handler.py`, and all four dispatcher paths in `src/zagg/runner.py` — local/lambda × spatial/raster). The run parquet reuses the same id in its object name, so leaf sidecar → run record → (once #299 lands) literal-template joins close. Merge-fold treats it as an identity field (equal-or-`None` collapse, absorbing). Dispatcher-built failure rows and the stale-worker fallback rows are stamped too (`failure_record`/`_lambda_result_rows` gained pass-throughs). On the lambda paths the run id also names the async `.status/` prefix (the spatial path already worked this way; the raster path now shares one id instead of two unrelated uuids).
3. **`template_hash` → `semantic_hash`** — per D19 rev 2 the record's hash is the *semantic-core* hash of the config, not a whole-template hash. Rename only; it stays a nullable placeholder until issue #299 lands the hasher. Docstring points at D19 rev 2.

## Phases

- [x] Phase 1: all three changes (one mechanical commit) + tests

## Testing

- `uv run pytest` — 2516 passed, 35 skipped; the one failure (`test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`) is a pre-existing local-Docker environment failure (pip inside the build image resolves a pre-3.12 Python and can't satisfy `zarr>=3.1.5`), reproduced on a clean stash of this diff — flagged below.
- `ruff check` / `ruff format --check` green on every touched file. Repo-wide `ruff check src tests` trips a pre-existing `N818` in `src/zagg/registry.py` (`UnknownCapability`), also present on `main` — not touched here.
- New/updated tests: `run_parquet_key` shape + round-trip suffix order; `run_id` verbatim copy, equal-or-`None` merge collapse, flatten column, failure-record stamp (`tests/test_telemetry.py`); `run_id` event-key threading on the spatial cell invoke (`tests/test_runner.py`); worker copies `run_id` verbatim into envelope + sidecar records (`tests/test_lambda_handler.py`); raster flat event key set now includes the always-on `run_id` stamp (`tests/test_raster_runner.py`).

## Questions for review

- **Nothing found depending on the old parquet name pattern.** All in-repo consumers match `stats_*` + `.parquet` prefix/suffix only (`tests/test_hive.py`, `tests/test_raster_runner.py`, `tests/test_coverage_root.py`, `tests/conftest.py::_no_s3_run_stats`); no benchmark/duckdb helper globs on the `stats_{run_id}_{ts}` order. External notebooks/scripts outside the tree weren't audited.
- `estimate_cost_usd` in `src/zagg/dispatch.py` still takes a `template_hash=` keyword (the #298 Phase-3 stub interface). Left as-is since it's not the record field; happy to rename it to `semantic_hash=` here if preferred, or leave it for #299's hasher PR.
- `deployment/aws/lambda_handler.py` is touched (three `build_record` call sites + event docstring) — treated as worker application code, same as PR #302 which introduced the `invoked_by` threading there; no infra/deploy scripts touched.
- The raster lambda `_event` stamps `run_id` unconditionally (the id always exists), while the spatial `_invoke_lambda_cell` keeps the `None`-guarded optional-key pattern for signature symmetry with `invoked_by`. Both are exercised by tests; flagging the asymmetry in case a uniform posture is preferred.

https://claude.ai/code/session_019zYV2pvzhVp8cv7e7H4JY3

### DIFF
--- a/src/zagg/telemetry.py
+++ b/src/zagg/telemetry.py
@@ -3,7 +3,7 @@
 One versioned record per processed shard, built from the worker's existing
 ``metadata`` dict. Two consumers, one source: the JSON **sidecar** written as a
 SIBLING object next to a hive leaf ``.zarr`` (``stats.json``; the
-``{hash}.stats.json`` naming arrives with issue #299 — ``template_hash`` is a
+``{hash}.stats.json`` naming arrives with issue #299 — ``semantic_hash`` is a
 nullable placeholder until then), and the **run-level parquet** the dispatcher
 writes at the store root (one row per shard, failure rows included).
 
@@ -57,7 +57,8 @@ _SUM_OR_NONE_KEYS = (
 _MAX_OR_NONE_KEYS = ("max_memory_mb", "container_hwm_mb")
 _EQ_OR_NONE_KEYS = (
     "shard_key",
-    "template_hash",
+    "run_id",
+    "semantic_hash",
     "granules_sha256",
     "zagg_version",
     "lambda",
@@ -112,6 +113,7 @@ def build_record(
     metadata: dict,
     granule_ids: Iterable[str] | None = None,
     invoked_by: dict | None = None,
+    run_id: str | None = None,
     lambda_config: dict | None = None,
 ) -> dict:
     """Build one shard's stats record from the worker's ``metadata`` dict.
@@ -122,6 +124,9 @@ def build_record(
     telemetry keys when the caller stamped them. ``invoked_by`` is copied
     VERBATIM from the invoke payload — the dispatcher resolves it via
     ``sts get-caller-identity`` once per run; workers cannot see the invoker.
+    ``run_id`` is threaded the same way (dispatcher-generated, stamped through
+    the invoke payload, copied verbatim) so a leaf sidecar joins back to its
+    run-level parquet.
     ``lambda_config`` is :func:`lambda_env` on Lambda, ``None`` locally;
     when present it prices ``gb_seconds`` / ``est_cost_usd`` from
     ``duration_s`` (the billed-duration approximation the dispatcher's cost
@@ -155,7 +160,11 @@ def build_record(
     return {
         "schema_version": SCHEMA_VERSION,
         "shard_key": int(shard_key),
-        "template_hash": None,  # nullable until issue #299 lands the hasher
+        "run_id": run_id,
+        # The semantic-core hash (D19 rev 2, docs/design/sparse_coverage.md):
+        # hashes the config's semantic core, NOT the whole template. Nullable
+        # until issue #299 lands the hasher.
+        "semantic_hash": None,
         "zagg_version": _zagg_version(),
         "n_shards": 1,
         "n_granules": int(n_granules),
@@ -245,7 +254,7 @@ def _zagg_version() -> str:
     return zagg.__version__
 
 
-def failure_record(*, shard_key=None, error, duration_s=None) -> dict:
+def failure_record(*, shard_key=None, error, duration_s=None, run_id=None) -> dict:
     """Skeleton record for a shard with no worker record (issue #297 phase 3).
 
     Timed-out / OOM / dropped shards write no sidecar and return no envelope
@@ -257,6 +266,7 @@ def failure_record(*, shard_key=None, error, duration_s=None) -> dict:
     record = build_record(
         shard_key=shard_key if shard_key is not None else -1,
         metadata={"error": str(error) or "unknown failure", "duration_s": duration_s},
+        run_id=run_id,
     )
     if shard_key is None:
         record["shard_key"] = None
@@ -267,7 +277,8 @@ def failure_record(*, shard_key=None, error, duration_s=None) -> dict:
 _ROW_SCALARS = (
     "schema_version",
     "shard_key",
-    "template_hash",
+    "run_id",
+    "semantic_hash",
     "zagg_version",
     "n_shards",
     "n_granules",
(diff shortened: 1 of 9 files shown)


src/zagg/telemetry.py
```diff
@@ -155,7 +160,11 @@ def build_record(
     return {
         "schema_version": SCHEMA_VERSION,
         "shard_key": int(shard_key),
-        "template_hash": None,  # nullable until issue #299 lands the hasher
+        "run_id": run_id,
+        # The semantic-core hash (D19 rev 2, docs/design/sparse_coverage.md):
+        # hashes the config's semantic core, NOT the whole template. Nullable
+        # until issue #299 lands the hasher.
+        "semantic_hash": None,
```
espg, line 169: 🤖 *from Claude (review)*

The record's key set changes here — `template_hash` → `semantic_hash` and the new `run_id` field — but `SCHEMA_VERSION` stays `1`. That directly contradicts the invariant documented at `src/zagg/telemetry.py:32-33`:

```
#: Version stamped into every record; bump on any key change (issue #297).
SCHEMA_VERSION = 1
```

PR #302 already shipped v1 records (with `template_hash`, no `run_id`), so there are now two distinct on-disk record shapes both stamped `schema_version=1`: the parquet column set and sidecar key set differ, yet nothing that keys on `schema_version` to know the columns can tell them apart. Any downstream table (Athena/duckdb) built on #302 output has a `template_hash` column this PR silently drops.

Note the countervailing pull: `merge()` raises `ValueError` on a `schema_version` mismatch, so bumping to 2 would make a hive rollup that mixes stale-container (v1) and new (v2) sidecars crash during a rolling deploy — though `merge()` is not yet wired into any production finalize path, so that is latent, not active. Either way one of the two should move: bump `SCHEMA_VERSION`, or update the "bump on any key change" comment to state the deliberate no-bump policy (and why) so the invariant and the code agree.

src/zagg/telemetry.py
```diff
@@ -3,7 +3,7 @@
 One versioned record per processed shard, built from the worker's existing
 ``metadata`` dict. Two consumers, one source: the JSON **sidecar** written as a
 SIBLING object next to a hive leaf ``.zarr`` (``stats.json``; the
-``{hash}.stats.json`` naming arrives with issue #299 — ``template_hash`` is a
+``{hash}.stats.json`` naming arrives with issue #299 — ``semantic_hash`` is a
```
espg, line 6: 🤖 *from Claude (review)*

Remaining `template_hash` reference not covered by this rename: `src/zagg/dispatch.py:364` (`estimate_cost_usd(*, template_hash: str | None = None, ...)`) plus its docstring at 398-399 ("The run's template identity (issue #299) used to select tier-1 priors") and the tier-1 text ("same template hash"). Its live test still passes `template_hash="abc"` (`tests/test_dispatch.py:177`).

That estimator is the intended consumer of exactly this identity field — tier-1 priors are selected by the run's config-identity hash, which after D19 rev 2 is `semantic_hash` (the semantic core), not a whole-template hash. So the tree now carries the old name on the reader and the new name on the record. The stub returns `None` and is deferred to #299/#298, so this is low severity — flagging it here so #299 unifies the two names (and the "same template hash" priors language) rather than leaving a silent naming mismatch across the join.

src/zagg/telemetry.py
espg: 🤖 *from Claude*

Kept `SCHEMA_VERSION = 1` and fixed the real defect the finding exposes — the stale comment — in 18f4343. The version gates *released* schema changes; this schema is unreleased (#302 merged days ago, nothing durable keys on v1 columns yet), and the key set changes here (`template_hash`→`semantic_hash`, new `run_id`) were ratified to rev in place with no bump in the D19 rev 2 / D20 record (https://github.com/englacial/zagg/pull/306#issuecomment-5024857031, which states "Schema/content unchanged; SCHEMA_VERSION stays 1 — unreleased"). The comment now says the bump-on-key-change rule applies once the schema is released and pre-release key changes rev in place, so the invariant and the code agree.

src/zagg/telemetry.py
espg: 🤖 *from Claude*

Fixed in 13b54e9: renamed the `estimate_cost_usd` keyword `template_hash` → `semantic_hash` in `src/zagg/dispatch.py` (param, the Parameters entry, the "template identity (issue #299)" and "same template hash"/"template-hash identity" docstring text) and updated the test at `tests/test_dispatch.py:177`. `grep template_hash src tests` now returns nothing, so the reader and the record carry the same name.

PR #311, opened 2026-07-20
Unified second-pass rollup sweep: stats + MOC + sub-shardmap families, hook + CLI (issue #300)

### LINKED ISSUE #300: Unified second-pass rollup sweep: stats, shardmap sub-maps, MOC, overviews

espg:
Closes #300

## What

Implements the D22 **unified second-pass rollup sweep** (`src/zagg/sweep.py`): one idempotent bottom-up pass over a hive store's digit tree that folds leaf artifacts into interior-node rollups, per registered artifact family, per the design record (`docs/design/sparse_coverage.md` §7 / D22 / D20 / D9) and the decisions on the issue thread ([trigger + full-ShardMap-JSON sub-maps](https://github.com/englacial/zagg/issues/300#issuecomment-5017291722), [schedule decoupling](https://github.com/englacial/zagg/issues/300#issuecomment-5017464707)).

Core mechanics (all families share the engine):

- **Discovery from run records, never a recursive LIST** (D22): callers pass the leaves a run touched — `(shard_key, window)` pairs from the dispatcher's run report (end-of-run hook) or from the run-level stats parquets at the product root (manual CLI). The walk visits only those leaves' ancestor paths; untouched siblings contribute through their existing stored rollups (≤ 4 candidate GETs per node, no LIST), so incremental runs accumulate exactly.
- **Generation stamps** (D22): every rollup records `{n_leaves, max_leaf_timestamp}`. A leaf re-run carries a later timestamp than every earlier record, so ancestor stamps stop matching — staleness is *detectable, not prevented* — and the next sweep rewrites exactly the stale chain. Skip-if-current is a two-part test: the generation stamp, backstopped by payload equality (covers same-second rewrites — both timestamp sources are whole-second). A second pass over an unchanged tree PUTs nothing.
- **Nothing load-bearing** (D9): rollups are regenerable caches. No existing read path consults them; deleting every rollup leaves all leaf reads (and the whole pre-existing test suite) green.
- Rollup objects are JSON sidecars at digit nodes named `{family}.rollup.json` — deliberately distinct from the leaf sidecar names (`stats.json` / `coverage.moc`): under D24 mixed-order stores a node can be leaf and interior at once, so sharing leaf names would self-clobber; the names list as objects, never digit-shaped prefixes, so the §5 discovery walk is unaffected.
- Shard-node rollups record the `windows` they merged, so an append run touching one window never drops its committed siblings (the union is re-read from leaf sidecars — truth — never re-derived from a prior rollup).
- The manifest's `spec` string is threaded into leaf-sidecar naming, so the PR #307 D23 seam resolves per store (legacy names today; the `/3` flip is issue #299's writer change). The sub-map sidecar name derives from the same seam (`shardmap.json` / `shardmap_{window}.json` / `{window}.shardmap.json`).

## Phases

- [x] **Phase 1 — sweep skeleton + stats rollup family.** Engine (bottom-up walk, generation stamps, skip-if-current), family registry, stats rollups via the existing `zagg.telemetry.merge` fold (D20 — associative by construction, so fold-of-folds == direct).
- [x] **Phase 2 — MOC regen family.** Committed-leaf coverage (D4 stamps — debris invisible) folded into interior MOC rollups in the same walk; the sweep is now a regenerator of the store-root `coverage.moc` via the same GET-union-PUT writer the runner uses (the runner's end-of-run fail-open write stays the fast path, per O7/D9), with a skip-when-covered gate. The O8 in-leaf bitmap contract is untouched (only the stamp envelope is read).
- [x] **Phase 3 — sub-shardmap rollup family, end-to-end.** Leaf sub-maps are **full ShardMap JSON** (the ratified format), written **by the worker on success** next to the stats sidecar — the local backends in-process, the Lambda workers from a new size-gated `submap` event block (`grid_signature` + catalog `metadata` + the `{id, s3, https}` granule entries the event's bare `granule_urls` can't reconstruct; raster events already carry entries, so their block is identity-only). The block is dropped (never fatal) when it would push an async event over the 256 KB cap — the sub-map is a D9 regenerable artifact and the sweep CLI is the backstop. Old deployments ignore the unknown key; old dispatchers send no block — both fail open. The fold: window union deduped by granule id at the shard node, then `ShardMap.reproject`'s exact coarsen regroup (merged in PR #295) level by level — rollup at order N == direct reproject at N (§8.3), verified against the production `reproject` as oracle. (Phase 1 had registered this family as a stub while #295 was open; this phase replaces it.)
- [x] **Phase 4 — window column + run-record discovery + manual CLI.** `window` joins the per-shard record and run parquet (threaded like `run_id`, per the ruling on the thread); run-parquet `shard_key` becomes a nullable UInt64 so packed morton words round-trip exactly (today the failure-row NaN coerces the column to float64, corrupting keys above 2^53 — found while building discovery); `python -m zagg.sweep` discovers leaves column-first from the run records, with a bounded per-shard-node LIST fallback for pre-column records only.
- [x] **Phase 5 — trigger wiring (D8-compliant).** Local backends: end-of-run in-process hook (fail-open, off the critical path) — the local "dispatcher" is also the worker and writes with the user's own credentials, so the D8 IAM rationale doesn't apply (stated here for veto). Lambda backends: **the dispatcher never PUTs** (D8 standing rule) — the hook is a fire-and-forget **`mode: "sweep"` worker Event invoke** (async, retries-0, fail-open, manual CLI as the D9 regeneration backstop), following the root-MOC `mode: "coverage"` pattern and coordinated with PR #315's `mode: "stats"` transport (size-gated inline leaves, discovery fallback).

**Registered but stubbed** (visible slots; `get_family` refuses with a pointer):
- `overview` — overview zarrs are issue #201's separate follow-on PR; only the D22 registration slot is reserved here.
- `debris` — the optional audit-class fifth family (D22), not implemented.

**Benchmark object model** (`.github/scripts/bench_objects.py`, the issue #240 tripwire): the model now counts the leaf sub-map as the second node-dir sibling, and sweep rollups get their own `objects_rollups` bucket — second-pass D9 caches, excluded from the write-path total the model audits (the #215 per-shard bypass guard is untouched: rollups never live inside a leaf prefix). This is the Python model module the unit tests import, not workflow/deploy config; it was updated the same way for the #297 sidecar.

**Sweep-invoke reliability vs #313's fire-verify-refire** (stated per the ruling): the `mode:"stats"` transport verifies-and-refires because a dropped invoke silently loses the run record. The `mode:"sweep"` invoke stays **plain fire-and-forget, retries-0, no refire** — a dropped invoke costs rollup *staleness only* (D9: every rollup is a regenerable cache; leaf truth is untouched), the next run's hook re-folds the same ancestor chains, and `python -m zagg.sweep` is the explicit backstop. Post-#313 ordering caveat documented at the dispatch site: the stats and sweep Event invokes race, so the oversized-work-set `discover: true` fallback may miss the racing run's records — same fail-open class.

**Post-#314 (morton-only writer flip)**: no interaction by construction — the sweep families read commit stamps, stats sidecars, and sub-map JSON, none of which carry `cell_ids`; the merged benchmark model tests (3-array leaves) and the full suite pass on the merged tip.

**Merged with post-branch main** (no rebase — the branch history is append-only per repo rules): the #299 semantic-hash stack and #305 landed mid-review, so `origin/main` is merged in with a merge commit. Reconciliations: `build_record` threads BOTH `window` (this PR) and `semantic_hash` (#299); the local sub-map emission now passes the manifest `spec` through the same #307 naming seam main's sidecar writes adopted; the benchmark hive model carries main's metadata ceiling (manifest + optional MOC + run parquet + `aggregation.yaml`) alongside this PR's per-shard sub-map sibling and rollup bucket.

## How tested

`tests/test_sweep.py`, keyed to the §8.3 standing obligations:

- **Rollup == direct**, per family: stats rollups equal `telemetry.merge` over the exact leaf record set; MOC rollup words at every level equal the direct union of committed shard words (root `coverage.moc` refreshed, `source: "sweep"`); sub-map rollups equal the run-level ShardMap reprojected directly to the rollup order by the production coarsen (shared granules dedup once).
- **Idempotence**: second pass over an unchanged tree has `written == 0`, byte-identical rollups, no root-MOC PUT.
- **Staleness**: a leaf re-run makes exactly its ancestor chain rewrite — both the timestamp fast path and the same-second payload-equality backstop are covered.
- **Incremental accumulation**: sweeping only a new window / sibling shard keeps earlier contributions; the root-MOC refresh unions with bases the sweep never visited.
- **Nothing load-bearing**: deleting every `*.rollup.json` leaves leaf sidecar reads intact and a re-sweep regenerates identical payloads.
- **Worker emission**: the local backend writes the leaf sub-map on success and none on failure (fake-hive-write integration, the `test_coverage_root` pattern); handler emission is additive (`submap` event block absent in the existing harness events, which stay green).
- Edge cases: missing sidecar/sub-map skipped, malformed sub-map counts `failed` (loud, not absent), unstamped debris invisible to the MOC family, foreign-order refs skipped with a warning, corrupt rollup rebuilt, windowed stamps union their D15 time ranges, empty work set is a no-op, no-manifest refusal, stub-family refusals, sub-map naming across legacy//3 grammars.

Local per push: `ruff check` / `ruff format --check` clean; `pytest` full suite green, with two local-environment-only exceptions flagged rather than fixed (§4): `test_lambda_build::test_function_build_succeeds` fails on this machine because the build env's pip cannot resolve `zarr>=3.1.5` (the PR's own CI `build / build-arm64` + `build-x86_64` checks pass), and `registry.py:64`'s pre-existing `N818` (outside the PR lint bot's `--select=E,F,W,I`).

## Questions for review

1. **Local-hook interpretation of D8** (phase 5): stated above — on the local backends the orchestrator process *is* the worker (it already PUTs every leaf), so the in-process hook doesn't cross the D8 boundary. Veto if the sweep should be invoke-shaped there too.
2. **Sub-map payload cap fallback**: units whose `submap` block would overflow the async event cap ship without it (logged at debug); their leaf sub-maps appear only once something re-emits them — a `--catalog` backfill flag on the sweep CLI is the natural home but is not in this PR; say the word if it should be. The dense-polar-shard case (#148-scale granule lists) is the realistic trigger.
3. **PR #315 interplay** (phase 5): the `mode: "sweep"` invoke will mirror its size-gated transport; whichever PR lands second carries a trivial rebase in `runner.py`'s post-run block.
4. **Rollup schedule**: dense every-order this round; sparse per-family manifest-declared schedules are issue #201 Phase C (confirmed on the thread), and the engine's level walk is compatible.

### DIFF
--- a/.github/scripts/bench_objects.py
+++ b/.github/scripts/bench_objects.py
@@ -276,18 +276,26 @@ def store_object_counts(
             if key in (hive.MANIFEST_NAME, hive.ROOT_COVERAGE_NAME) or _is_run_parquet(key):
                 metadata += 1
                 continue
-            # Sweep rollups (issue #300): `{family}.rollup.json` at any digit
-            # node — second-pass D9 caches with their own bucket, never
-            # write-path objects (and never inside a leaf prefix, so the #215
-            # per-shard guard is untouched).
-            if key.endswith(".rollup.json"):
-                rollups += 1
-                continue
+            # Leaf-prefix membership FIRST (issue #300 review): an object that
+            # lands INSIDE a leaf `.zarr/` prefix is that shard's data object —
+            # rollup-named or not — so it counts into ``per_shard`` and trips
+            # the exact #215 per-shard guard. Only keys OUTSIDE every leaf
+            # prefix are eligible for the rollup / sibling buckets below.
             for prefix, label in leaf_of.items():
                 if key.startswith(prefix):
                     per_shard[label] = per_shard.get(label, 0) + 1
                     break
             else:
+                # Sweep rollups (issue #300): `{family}.rollup.json` at a digit
+                # node — second-pass D9 caches with their own bucket, never
+                # write-path objects. Legit rollups sit at `{node}/` (siblings
+                # of the leaf `.zarr/`), so classifying them here — after the
+                # leaf-membership check has failed — leaves the per-shard guard
+                # untouched while a MISPLACED in-leaf rollup falls to the
+                # attribution above and trips #215 instead of vanishing.
+                if key.endswith(".rollup.json"):
+                    rollups += 1
+                    continue
                 node, _, name = key.rpartition("/")
                 is_sibling = any(
                     name == f"{stem}.json"
--- a/.github/scripts/bench_objects.py
+++ b/.github/scripts/bench_objects.py
@@ -148,10 +148,14 @@ def expected_object_counts(
         # Leaf fixed objects: leaf root zarr.json + group zarr.json + one
         # zarr.json per array, plus the in-leaf coverage.moc sidecar (written
         # for any populated leaf when the leaf has depth, i.e. child_order >
-        # parent_order), plus the stats.json sidecar SIBLING to the leaf
-        # (issue #297 — one per successful shard, in the leaf's node dir).
+        # parent_order), plus TWO node-dir siblings per successful shard: the
+        # stats.json sidecar (issue #297) and the shardmap.json leaf sub-map
+        # (issue #300 — same success gate; on the Lambda path it may be
+        # legitimately absent for a unit whose submap event block was dropped
+        # over the async payload cap, which then surfaces here as a mismatch
+        # worth seeing rather than a modeled window).
         sidecar = 1 if grid.child_order > grid.parent_order else 0
-        lo = hi = 3 + len(members) + sidecar
+        lo = hi = 4 + len(members) + sidecar
         for m in members:
             blocks = m["blocks_per_shard"]
             hi += blocks
@@ -225,6 +229,7 @@ def store_object_counts(
     per_shard: dict[str, int] = {}
     other: list[str] = []
     metadata = 0
+    rollups = 0
 
     if store_layout == "flat":
         _require_fullsphere(grid)
@@ -259,9 +264,10 @@ def store_object_counts(
             hive.shard_leaf_path("", int(k)).lstrip("/") + "/": grid.shard_label(int(k))
             for k in shard_keys
         }
-        # The per-shard stats sidecar (issue #297) is a SIBLING of the leaf
-        # .zarr — ``{node}/stats.json`` (``stats_{window}.json`` windowed) —
-        # so attribute it to the node's shard rather than pooling it in
+        # The per-shard stats sidecar (issue #297) and the leaf sub-map
+        # (issue #300) are SIBLINGS of the leaf .zarr — ``{node}/stats.json``
+        # and ``{node}/shardmap.json`` (``_{window}`` suffixed when windowed)
+        # — so attribute them to the node's shard rather than pooling them in
         # ``other``.
         stats_of = {
             prefix.rstrip("/").rsplit("/", 1)[0] + "/": label for prefix, label in leaf_of.items()
@@ -270,16 +276,25 @@ def store_object_counts(
             if key in (hive.MANIFEST_NAME, hive.ROOT_COVERAGE_NAME) or _is_run_parquet(key):
                 metadata += 1
                 continue
+            # Sweep rollups (issue #300): `{family}.rollup.json` at any digit
+            # node — second-pass D9 caches with their own bucket, never
+            # write-path objects (and never inside a leaf prefix, so the #215
+            # per-shard guard is untouched).
+            if key.endswith(".rollup.json"):
+                rollups += 1
+                continue
             for prefix, label in leaf_of.items():
                 if key.startswith(prefix):
                     per_shard[label] = per_shard.get(label, 0) + 1
                     break
             else:
                 node, _, name = key.rpartition("/")
-                is_stats = name == "stats.json" or (
-                    name.startswith("stats_") and name.endswith(".json")
+                is_sibling = any(
+                    name == f"{stem}.json"
+                    or (name.startswith(f"{stem}_") and name.endswith(".json"))
+                    for stem in ("stats", "shardmap")
                 )
-                if is_stats and node + "/" in stats_of:
+                if is_sibling and node + "/" in stats_of:
                     per_shard[stats_of[node + "/"]] = per_shard.get(stats_of[node + "/"], 0) + 1
                 else:
                     other.append(key)
@@ -290,6 +305,7 @@ def store_object_counts(
         "objects_total": len(keys),
         "objects_metadata": metadata,
         "objects_per_shard": per_shard,
+        "objects_rollups": rollups,
         "objects_other": len(other),
         "other_keys": other[:20],
     }
@@ -346,7 +362,13 @@ def object_count_mismatch(measured: dict, expected: dict) -> str | None:
     know every object the run writes.
     """
     problems = []
-    total = measured["objects_total"]
+    # Sweep rollups (issue #300) are second-pass D9 caches, not write-path
+    # objects: the end-of-run sweep may or may not have landed them by
+    # measurement time (fire-and-forget on Lambda), so they are tallied in
+    # their own bucket and excluded from the write-path total this model
+    # audits (the #215 bypass guard below is untouched — rollups never live
+    # inside a leaf prefix).
+    total = measured["objects_total"] - measured.get("objects_rollups", 0)
     meta = measured["objects_metadata"]
     meta_lo, meta_hi = expected["metadata_min"], expected["metadata"]
     if not (meta_lo <= meta <= meta_hi):
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -939,6 +939,7 @@ class RasterStrategy:
                 metadata=meta,
                 granule_ids=raster_granule_ids(granules),
                 run_id=run_id,
+                window=window["label"] if window else None,
             )
             meta["stats"] = record
             if store_layout == "hive" and meta.get("leaf_written"):
@@ -2571,6 +2572,7 @@ def _run_local(
                 metadata=meta,
                 granule_ids=_resolve_urls(records, driver),
                 run_id=run_id,
+                window=window["label"] if window else None,
             )
             meta["stats"] = record
             if store_layout == "hive" and not meta.get("error"):
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -957,17 +957,24 @@ class RasterStrategy:
                 # entries as full ShardMap JSON, sibling to the stats sidecar
                 # (same in-process rationale as the aggregation local path).
                 try:
-                    from zagg.sweep import write_leaf_submap
+                    from zagg.sweep import submap_emittable, write_leaf_submap
 
-                    write_leaf_submap(
-                        store_path,
-                        int(shard_key),
-                        granules,
-                        grid_signature=catalog_data["grid_signature"],
-                        metadata=catalog_data.get("metadata"),
-                        window=window["label"] if window else None,
-                        store_kwargs=store_kwargs,
-                    )
+                    sig = catalog_data["grid_signature"]
+                    if submap_emittable(sig, granules):
+                        write_leaf_submap(
+                            store_path,
+                            int(shard_key),
+                            granules,
+                            grid_signature=sig,
+                            metadata=catalog_data.get("metadata"),
+                            window=window["label"] if window else None,
+                            store_kwargs=store_kwargs,
+                        )
+                    else:
+                        logger.debug(
+                            f"leaf sub-map skipped for shard {shard_key}: non-HEALPix grid "
+                            f"or id-less entries (unmergeable, issue #300)"
+                        )
                 except Exception as e:
                     logger.warning(f"leaf sub-map write failed (fail-open, issue #300): {e}")
             return meta, stage_stats, write_s
@@ -2623,17 +2630,24 @@ def _run_local(
                 # (the Lambda path threads them via the event's submap block).
                 # Fail-open, like the sidecar.
                 try:
-                    from zagg.sweep import write_leaf_submap
+                    from zagg.sweep import submap_emittable, write_leaf_submap
 
-                    write_leaf_submap(
-                        store_path,
-                        int(shard_key),
-                        records,
-                        grid_signature=catalog_data["grid_signature"],
-                        metadata=catalog_data.get("metadata"),
-                        window=window["label"] if window else None,
-                        store_kwargs=store_kwargs,
-                    )
+                    sig = catalog_data["grid_signature"]
+                    if submap_emittable(sig, records):
+                        write_leaf_submap(
+                            store_path,
+                            int(shard_key),
+                            records,
+                            grid_signature=sig,
+                            metadata=catalog_data.get("metadata"),
+                            window=window["label"] if window else None,
+                            store_kwargs=store_kwargs,
+                        )
+                    else:
+                        logger.debug(
+                            f"leaf sub-map skipped for shard {shard_key}: non-HEALPix grid "
+                            f"or id-less entries (unmergeable, issue #300)"
+                        )
                 except Exception as e:
                     logger.warning(f"leaf sub-map write failed (fail-open, issue #300): {e}")
             return {"shard_key": shard_key, "ok": True, "meta": meta}
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -808,6 +808,10 @@ class RasterStrategy:
                 output_endpoint_url=resolved_endpoint,
                 store_layout=store_layout,
                 dataset={"short_name": md.get("short_name"), "version": md.get("version")},
+                catalog_signature={
+                    "grid_signature": catalog_data.get("grid_signature"),
+                    "metadata": catalog_data.get("metadata"),
+                },
                 profile=profile,
                 invocation=invocation,
                 on_progress=on_progress,
@@ -947,6 +951,23 @@ class RasterStrategy:
                     write_sidecar(leaf, record, **store_kwargs)
                 except Exception as e:
                     logger.warning(f"stats sidecar write failed (fail-open, issue #297): {e}")
+                # Leaf sub-map (issue #300, D22): the unit's raster ShardMap
+                # entries as full ShardMap JSON, sibling to the stats sidecar
+                # (same in-process rationale as the aggregation local path).
+                try:
+                    from zagg.sweep import write_leaf_submap
+
+                    write_leaf_submap(
+                        store_path,
+                        int(shard_key),
+                        granules,
+                        grid_signature=catalog_data["grid_signature"],
+                        metadata=catalog_data.get("metadata"),
+                        window=window["label"] if window else None,
+                        store_kwargs=store_kwargs,
+                    )
+                except Exception as e:
+                    logger.warning(f"leaf sub-map write failed (fail-open, issue #300): {e}")
             return meta, stage_stats, write_s
 
         stage_max: dict = {}
@@ -1075,6 +1096,7 @@ class RasterStrategy:
         output_endpoint_url,
         store_layout="flat",
         dataset=None,
+        catalog_signature=None,
         profile=False,
         invocation="async",
         on_progress=None,
@@ -1252,6 +1274,12 @@ class RasterStrategy:
                 # byte-identical to pre-#247 runs.
                 if window is not None:
                     ev["window"] = window
+                # Leaf sub-map fields (issue #300, D22): raster events already
+                # carry the unit's full ShardMap entries in ``granules``, so
+                # only the catalog identity rides here; the worker writes the
+                # full ShardMap JSON sub-map next to the leaf on success.
+                if catalog_signature is not None:
+                    ev["submap"] = catalog_signature
             else:
                 keys = {e.get("time_key") or e.get("datetime") for e in granules if e.get("assets")}
                 ev["time_index"] = {k: time_index[k] for k in keys}
@@ -2555,6 +2583,25 @@ def _run_local(
                     write_sidecar(leaf, record, **store_kwargs)
                 except Exception as e:
                     logger.warning(f"stats sidecar write failed (fail-open, issue #297): {e}")
+                # Leaf sub-map (issue #300, D22): the unit's ShardMap entries as
+                # full ShardMap JSON, sibling to the stats sidecar. The local
+                # "worker" is in-process, so the catalog fields are in scope
+                # (the Lambda path threads them via the event's submap block).
+                # Fail-open, like the sidecar.
+                try:
+                    from zagg.sweep import write_leaf_submap
+
+                    write_leaf_submap(
+                        store_path,
+                        int(shard_key),
+                        records,
+                        grid_signature=catalog_data["grid_signature"],
+                        metadata=catalog_data.get("metadata"),
+                        window=window["label"] if window else None,
+                        store_kwargs=store_kwargs,
+                    )
+                except Exception as e:
+                    logger.warning(f"leaf sub-map write failed (fail-open, issue #300): {e}")
             return {"shard_key": shard_key, "ok": True, "meta": meta}
         except Exception as e:
             return {"shard_key": shard_key, "ok": False, "error": e}
@@ -2911,6 +2958,16 @@ def _run_lambda(
         granule_urls = _resolve_urls(records, "s3")
         ds = _clamped_data_source(config.data_source, len(granule_urls))
         cell_config_dict = {**config_dict, "data_source": ds} if ds is not None else config_dict
+        # Leaf sub-map fields (issue #300, D22): the worker writes the full
+        # ShardMap JSON sub-map at the leaf prefix on success, but the event's
+        # bare granule_urls can't reconstruct the {id, s3, https} entries — so
+        # the unit's entries + the catalog identity ride a dedicated block
+        # (size-gated in _invoke_lambda_cell; dropped, never fatal).
+        submap = {
+            "grid_signature": catalog_data["grid_signature"],
+            "metadata": catalog_data.get("metadata"),
+            "granules": records,
+        }
         return _invoke_lambda_cell(
             state["lambda_client"],
             grid.block_index(int(shard_key)),
@@ -2930,6 +2987,7 @@ def _run_lambda(
             label=label,
             invoked_by=invoked_by,
             run_id=run_id,
+            submap=submap,
             **extra,
         )
 
@@ -4400,6 +4458,7 @@ def _invoke_lambda_cell(
     label=None,
     invoked_by=None,
     run_id=None,
+    submap=None,
 ):
     """Invoke Lambda for a single cell with retry logic.
 
@@ -4425,6 +4484,12 @@ def _invoke_lambda_cell(
     connection sits idle while the shard runs, so long shards survive NAT idle
     timeouts (GitHub-hosted runners sever synchronous invokes at ~4 min). When
     ``None`` (legacy sync path) the invoke is byte-identical to before.
+    ``submap`` (issue #300) forwards the unit's leaf sub-map fields
+    (``grid_signature``/``metadata``/``granules`` — the ShardMap entries the
+    worker cannot derive from bare urls) so the worker writes the D22 full
+    ShardMap JSON sub-map next to the leaf; size-gated below (dropped, never
+    fatal, when it would push an async event over the 256 KB cap) and omitted
+    (``None``) the event stays byte-identical.
     """
     wall_start = time.time()
 
@@ -4484,6 +4549,20 @@ def _invoke_lambda_cell(
     # rather than letting every attempt fail on Lambda's raw
     # RequestEntityTooLargeException (issue #151).
     payload = json.dumps(event)
+    # Leaf sub-map block (issue #300): attached only when it FITS — a unit
+    # whose granule entries would push an async event over the cap keeps its
+    # pre-#300 payload (the sub-map is a regenerable D9 artifact; the manual
+    # sweep CLI is the backstop), so the cap gate below can never start
+    # rejecting a run that dispatched fine before.
+    if submap is not None:
+        with_submap = json.dumps({**event, "submap": submap})
+        if invocation_type != "Event" or len(with_submap) <= _ASYNC_PAYLOAD_CAP_BYTES:
+            payload = with_submap
+        else:
+            logger.debug(
+                f"cell {label or shard_key}: dropping submap block ({len(with_submap):,} "
+                f"bytes over the async budget); leaf sub-map deferred to the sweep CLI"
+            )
     if invocation_type == "Event" and len(payload) > _ASYNC_PAYLOAD_CAP_BYTES:
         raise ValueError(
             f"cell {label or shard_key} event payload is {len(payload):,} bytes, over the "
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -44,6 +44,7 @@ from zagg.config import (
     get_pipeline_type,
     get_store_layout,
     get_store_path,
+    get_sweep,
     get_windowing,
 )
 from zagg.dispatch import (
@@ -1046,6 +1047,18 @@ class RasterStrategy:
         run_stats_path = _write_run_stats(
             store_path, rows, run_id=run_id, store_kwargs=store_kwargs
         )
+        # End-of-run rollup sweep (issue #300): in-process on the local
+        # backend (this process is the worker — see _run_local's note; D8
+        # constrains only the Lambda paths). Only units that wrote a leaf
+        # contribute work; fail-open inside sweep_after_run (D9).
+        if store_layout == "hive" and get_sweep(config):
+            from zagg.sweep import leaves_from_stats_records, sweep_after_run
+
+            leaves = leaves_from_stats_records(
+                [m.get("stats") for m in ok_metas if m.get("leaf_written")]
+            )
+            if leaves:
+                sweep_after_run(store_path, leaves, store_kwargs=store_kwargs)
         # Per-shard isolation lets one bad shard be counted and skipped, but a
         # run where EVERY shard raised (e.g. a config band whose ``asset`` is
         # absent from every granule) would otherwise return a success-shaped,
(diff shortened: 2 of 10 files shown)


src/zagg/sweep.py
```diff
@@ -0,0 +1,539 @@
...
+    except Exception as e:
+        logger.warning(f"sweep[{fam.name}]: merge failed at node {node}; skipping ({e})")
+        counts["failed"] += 1
+        return None
+
+
+def _generation(n_leaves: int, timestamps) -> dict:
+    """The D22 generation stamp: merged-leaf count + max leaf timestamp."""
+    stamps = [t for t in timestamps if t is not None]
+    return {"n_leaves": int(n_leaves), "max_leaf_timestamp": max(stamps) if stamps else None}
```
espg, line 736: 🤖 *from Claude (review)*

[should-fix] Generation-stamp blind spot at sub-second re-runs. The D22 staleness guarantee here rests entirely on `max_leaf_timestamp` moving forward on a re-run (docstring lines 19-24: "a leaf re-run bumps its artifact timestamp past every earlier stamp"). But both timestamp sources have **1-second resolution**: stats records use `datetime.now(...).isoformat(timespec="seconds")` (`telemetry.build_record`), and the MOC family reads `stamp.get("written_at")` which `hive._utcnow()` also stamps at `timespec="seconds"`. Two runs of the same leaf within the same wall-clock second produce an **identical** `(n_leaves, max_leaf_timestamp)` generation, so `skip-if-current` treats the stale rollup as current and never rewrites the chain — silently serving stale rollups. This is realistic for fast local shards re-run back-to-back (the exact "local execution" path phase 1 targets). Not relitigating the D22 stamp *definition* (count + max ts) — flagging that the underlying clock resolution defeats the "detectable" half of "detectable, not prevented" below one second. Please either fold a content discriminator into the stamp compare (e.g. a payload digest, kept out of the D22 stamp itself) or document the ≥1s-between-reruns assumption loudly here and in §7 of the design record.

tests/test_sweep.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def test_leaf_rerun_makes_ancestors_detectably_stale(self, tmp_path):
+        _write_manifest(tmp_path)
+        _put_leaf(tmp_path, "-311", timestamp="2026-01-01T00:00:00+00:00")
+        _put_leaf(tmp_path, "-321", timestamp="2026-01-02T00:00:00+00:00")
+        run_sweep(str(tmp_path), _leaf_refs("-311", "-321"))
+        stale_gen = _rollup(tmp_path, "-3")["generation"]
+        # Re-run of one leaf: a rewritten sidecar always carries a LATER
+        # timestamp than every earlier record, so ancestor stamps mismatch.
+        _put_leaf(tmp_path, "-311", n_obs=99, timestamp="2026-01-03T00:00:00+00:00")
```
espg, line 130: 🤖 *from Claude (review)*

[should-fix] Staleness test masks the real timestamp path. This test manufactures distinct day-apart timestamps by passing `timestamp=` into `_put_leaf`, which overrides `build_record`'s auto-stamp. So it never exercises the production timestamp — `build_record` stamps `timespec="seconds"`, and a real re-run of `-311` moments after the first would carry the SAME timestamp, leaving `(n_leaves, max_leaf_timestamp)` unchanged and the ancestor chain silently NOT rewritten (see the sub-second blind-spot finding on `sweep.py:478`). An assert that only ever sees hand-injected, guaranteed-monotonic timestamps will pass on an implementation that is broken for real back-to-back re-runs. Suggest adding a case that re-runs a leaf via the real `build_record` path within one process tick and asserts either detected-staleness (if a content discriminator is added) or an explicit xfail documenting the ≥1s assumption.

src/zagg/sweep.py
```diff
@@ -0,0 +1,539 @@
...
+    contributing. Generation is the children's sum/max — fold-of-folds equals
+    the direct leaf fold because every family's merge is associative (§8.3).
+    """
+    from zagg.hive import _decimal_order
+
+    children = []
+    for digit in "1234":
+        art = computed.get(node + digit)
+        if art is None:
+            art = _read_rollup(store, fam, node + digit)
```
espg, line 689: 🤖 *from Claude (review)*

[nit] Emptied shard keeps contributing through its stale on-store rollup. When a shard is in this run's work set but all its window leaves now read empty, `_rollup_shard_node` returns `None` **without deleting** the shard's existing `stats.rollup.json`. Here `computed.get(node+digit)` is then `None`, so the interior falls through to `_read_rollup(node+digit)` and picks up the STALE shard rollup — the emptied shard still contributes to its parent, and the sweep has no path to represent removal. Under append-only + D9 this may be intended (leaves are not normally deleted), but the interaction is silent: a shard that lost its data is invisible to the sweep until something else rewrites the parent. Please confirm this is intended and, if so, add a one-line note here; otherwise the emptied-shard rollup needs an explicit delete/tombstone.

tests/test_sweep.py
```diff
@@ -0,0 +1,335 @@
...
+    def test_foreign_order_leaf_ref_is_skipped_with_warning(self, tmp_path, caplog):
+        _write_manifest(tmp_path)
+        summary = run_sweep(str(tmp_path), [(morton_word("-3111"), None)])
+        assert summary["skipped_leaves"] == ["-3111"]
+        assert summary["n_leaves"] == 0
+        assert "mixed-order" in caplog.text
+
+    def test_no_manifest_raises(self, tmp_path):
+        with pytest.raises(ValueError, match="not a hive store root"):
+            run_sweep(str(tmp_path), [])
```
espg, line 233: 🤖 *from Claude (review)*

[nit] No coverage for the empty work set. `run_sweep(root, [])` on a store *with* a manifest is a live path (a run that touched nothing, or a caller filtering all leaves out): it exercises the immediate `break` in the frontier walk (`sweep.py:363`) and `MocFamily.finish(tops=[])` returning `root_moc_written: False` (`sweep.py:178`). Only the no-manifest raise is tested here. A one-liner asserting `run_sweep(root, [])` returns zeroed per-family counts and does not PUT a root MOC would lock in the empty-store behavior.

src/zagg/sweep.py
espg: 🤖 *from Claude*

Fixed in d02fb7a. The skip-if-current test is now two-part: every node recomputes its merged payload BEFORE the skip decision, and the PUT is skipped only when the stored generation stamp AND the stored payload both match. The generation stamp (count + max leaf timestamp, unchanged per D22) stays the fast path; payload equality is the backstop for the sub-second case you flagged. It closes end-to-end without touching the stamp schema or telemetry timestamp resolution: shard nodes re-read their leaf sidecars every pass, so a same-second content change flips the payload compare; interior nodes fold the freshly computed child payloads (the rewritten child is in `computed`), so the rewrite cascades up the chain. The failed-merge path is unchanged (counts["failed"], no write). Module docstring rewritten to state the two-part test; regression test in 249cda5 exercises it with a frozen same-second clock.

tests/test_sweep.py
espg: 🤖 *from Claude*

Addressed in 249cda5. Added `test_leaf_rerun_same_second_rewrites_via_payload_compare`, which uses the REAL `build_record` auto-stamp (no hand-injected `timestamp=`) and freezes the wall clock so a back-to-back re-run of `-311` lands in the same second — exactly the production path. It re-puts the leaf with different content and asserts the full ancestor chain (`-311`, `-31`, `-3`) is rewritten and the base payload reflects the new obs; it also asserts the generation stamp is IDENTICAL across the re-run, proving the payload-equality backstop (d02fb7a) is what caught the change, not the stamp. The old day-apart test is kept and renamed `test_leaf_rerun_day_apart_bumps_generation_stamp` with a comment noting it covers the generation-stamp fast path (hand-injected timestamps there are intentional, to force the day gap).

src/zagg/sweep.py
espg: 🤖 *from Claude*

Confirmed intended; leaving the behavior by design, with a note added in be18678. The store is append-only at the leaf level — leaf deletion/GC is the registered debris family, deliberately stubbed — and rollups are D9 regenerable caches, so an emptied/vanished leaf whose shard rollup returns `None` this pass keeps contributing through its prior on-store rollup until a deletion-aware family exists. That is regeneration-safe, not stale-serving: nothing here is truth. I added a paragraph to the `_rollup_interior` docstring at this fallback (the `_read_rollup(node+digit)` branch) spelling this out, so the interaction is no longer silent. An explicit delete/tombstone is out of scope for this PR — it is the debris-family work.

tests/test_sweep.py
espg: 🤖 *from Claude*

Added in add05d4. `test_empty_work_set_writes_nothing` runs `run_sweep(root, [])` on a manifest-bearing store (a stamped + sidecar-ed leaf present) and asserts `n_leaves == 0`, all four counts (`written`/`current`/`empty`/`failed`) zero for both `stats` and `moc`, `root_moc_written is False`, no `*.rollup.json` object written anywhere under the root, and `read_root_coverage(root) is None` (no root `coverage.moc` created). This locks in the immediate frontier-walk break and `MocFamily.finish(tops=[])` behavior you pointed at.

src/zagg/runner.py
```diff
@@ -947,6 +951,23 @@ def _write_slab(t_idx, slab):
...
+                # entries as full ShardMap JSON, sibling to the stats sidecar
+                # (same in-process rationale as the aggregation local path).
+                try:
+                    from zagg.sweep import write_leaf_submap
+
+                    write_leaf_submap(
+                        store_path,
+                        int(shard_key),
+                        granules,
+                        grid_signature=catalog_data["grid_signature"],
```

[should-fix] This raster local path (and the identical raster-lambda `submap` block wired via `catalog_signature`, runner.py ~1279) emits a leaf sub-map whose `grid_signature` is the **rectilinear** `spatial_signature()` — `{type, crs, affine, shape, chunk_shape}`, with **no `parent_order`/`child_order`** (grids/rectilinear.py:299). And raster ShardMap entries can carry **no `id`** — `raster_granule_ids` falls back to `datetime` (telemetry.py:91). `submap` is now a DEFAULT family, so any sweep of a raster hive store (#237) hands these to `SubmapFamily.merge`, which unconditionally reads `signature["parent_order"]` and `entry["id"]` (sweep.py:376/368) → `KeyError` → the node is counted `failed`, never folded. So phase 3 writes raster leaf sub-maps the fold can never consume, guaranteeing `failed` spam and zero rollups on raster stores. Sub-shardmap rollups are inherently HEALPix-morton (`reproject` is HEALPix-only). Fix: gate the leaf-submap write on a HEALPix `grid_signature` (skip rectilinear) at both raster call sites, mirroring the reproject contract.

src/zagg/sweep.py
```diff
@@ -0,0 +1,727 @@
...
+                for entry in entries:
+                    bucket[entry["id"]] = dict(entry)
+        keys = sorted(buckets)
+        granules = [list(buckets[k].values()) for k in keys]
+        signature = dict(payloads[0]["grid_signature"])
+        meta = dict(payloads[0].get("metadata") or {})
+        meta.pop("reproject", None)  # never inherit a child fold's stamp
+        meta["total_granules"] = len({e["id"] for g in granules for e in g})
+        folded = ShardMap(signature, keys, granules, meta)
+        if int(signature["parent_order"]) == int(order):
```
espg, line 438: 🤖 *from Claude (review)*

[should-fix] `merge` assumes every leaf/child payload is a HEALPix ShardMap with per-entry `id`: line 368 does `bucket[entry["id"]]`, line 374 `{e["id"] ...}`, and this line `int(signature["parent_order"])`. When those keys are missing the `_merged` wrapper catches the bare `KeyError` and counts the node **`failed`** (sweep.py:508) — the same channel as a genuinely malformed sub-map (the "loud, not absent" broken-writer signal, per `read_leaf`). So an *unsupported grid* (rectilinear raster leaf sub-map, see the runner.py:964 finding) is indistinguishable from a *corrupt* one, and a raster sweep reports `failed` counts that look like data corruption. Harden here even if the raster writer is gated: detect a non-HEALPix `grid_signature` (no `parent_order`, or `type != "healpix"`) and skip it cleanly (return a no-op / raise a typed `SweepUnsupported` the engine counts as `empty`/skipped, not `failed`), and treat a missing `entry["id"]` the same way rather than as a broken writer.

src/zagg/sweep.py
```diff
@@ -0,0 +1,727 @@
...
+        buckets: dict[int, dict] = {}
+        for p in payloads:
+            for key, entries in zip(p["shard_keys"], p["granules"]):
+                bucket = buckets.setdefault(int(key), {})
+                for entry in entries:
+                    bucket[entry["id"]] = dict(entry)
+        keys = sorted(buckets)
+        granules = [list(buckets[k].values()) for k in keys]
+        signature = dict(payloads[0]["grid_signature"])
+        meta = dict(payloads[0].get("metadata") or {})
```
espg, line 434: 🤖 *from Claude (review)*

[nit] The fold inherits provenance from `payloads[0]` only (`meta = dict(payloads[0].get("metadata"))`, and `signature = dict(payloads[0]["grid_signature"])` on line 371). `total_granules` is recomputed, but `collection`/`short_name`/`version`/`footprint` are silently taken from the first child. For a single-product store this is fine; for a store that accreted leaves from more than one catalog (mixed collections under one root), the rollup advertises one arbitrary child's identity. `grid_signature` is safe (all children of a node share `child_order`), so only the descriptive metadata is at risk. Worth a one-line note in the docstring that the submap fold assumes a single-catalog store, or reconcile the identity fields (e.g. drop/merge conflicting keys) — otherwise the rollup metadata is quietly misleading on a multi-catalog store.

tests/test_sweep.py
```diff
@@ -0,0 +1,579 @@
...
+        assert r31["grid_signature"]["parent_order"] == 1
+        assert r31["metadata"]["reproject"]["method"] == "coarsen"
+        # Rollup == direct (§8.3): the base rollup equals the run-level
+        # ShardMap reprojected straight to order 0 by the production coarsen.
+        direct = ShardMap(
+            dict(SUBMAP_SIG),
+            [morton_word(d) for d in sorted(per_shard)],
+            [[_entry(g) for g in per_shard[d]] for d in sorted(per_shard)],
+            {"collection": "TEST_001"},
+        ).reproject(_ReprojectTarget(SUBMAP_SIG, 0))
```
espg, line 473: 🤖 *from Claude (review)*

[nit] The "rollup == direct (§8.3)" oracle reprojects with `_ReprojectTarget(SUBMAP_SIG, 0)` — the same shim the production fold uses, not a real `HealpixGrid`. So this proves the fold agrees with itself, not with the grid the design contract is written against. If the shim ever diverges from a real grid's `spatial_signature()` at the coarsened order (an extra/renamed key, a differently-typed `parent_order`), both sides move together and the test stays green while stored rollups carry a `grid_signature` a genuine reproject would not. Strengthen by building the oracle target from an actual `from_config`/`HealpixGrid` at order 0 (or add a focused assert that `_ReprojectTarget(SUBMAP_SIG, N).spatial_signature()` equals a real grid's at order N), so the shim is pinned to the thing it stands in for.

src/zagg/sweep.py
```diff
@@ -0,0 +1,727 @@
...
+        meta.pop("reproject", None)  # never inherit a child fold's stamp
+        meta["total_granules"] = len({e["id"] for g in granules for e in g})
+        folded = ShardMap(signature, keys, granules, meta)
+        if int(signature["parent_order"]) == int(order):
+            # Shard-node fold: already at the node's order — a window union,
+            # not a reprojection; keep counts honest without a noop stamp.
+            meta.update(total_shards=len(keys), total_pairs=sum(len(g) for g in granules))
+        else:
+            folded = folded.reproject(_ReprojectTarget(signature, order))
```
espg, line 451: 🤖 *from Claude (review)*

[nit] Provenance-stamp accuracy across multi-level folds. At an interior node the fold coarsens the *children* (`signature` comes from `payloads[0]`, whose `parent_order` is the child order), so the resulting `metadata.reproject.source_parent_order` records the immediate child order — not the leaf order a *direct* reproject-to-N would record. The data (keys/granules) still equals direct (coarsen is transitive), which is what §8.3 requires and what `test_rollup_equals_direct_reproject` checks, so this is not a data bug. But the stamp reads as "reprojected from order N-1" at every level, which is misleading provenance for anyone who reads it as "the source order this map derives from." Either document that the stamp records the last fold hop (not the origin order), or set `source_parent_order` to the true leaf/shard order carried through the fold.

src/zagg/runner.py

Fixed in d8cc6ae. Added a shared guard `submap_emittable(grid_signature, granules)` in sweep.py and gated all four emission sites on it (runner local agg + raster local, and the handler's two `submap` branches). A unit whose `grid_signature` is not HEALPix (no `parent_order`/`child_order`, or `type != "healpix"`) or whose entries lack an `id` now skips the write with a `logger.debug` instead of persisting a payload the fold counts `failed` — no per-shard warn-spam on raster stores. `write_leaf_submap` stays strict for direct callers. Covered by `test_submap_emittable_gates_unmergeable_units` (rectilinear signature and id-less raster-shaped entries both return False).

src/zagg/sweep.py
espg: 🤖 *from Claude*

Fixed in 142be2b. In `read_leaf`, after the required-keys check (still a raised `ValueError` → `failed`, the corrupt path), a payload whose `grid_signature` lacks `parent_order`/`child_order` or has `type != "healpix"` is now treated as UNSUPPORTED: warn once-ish (`_warn_unsupported_submap`, deduped per grid type so a raster sweep does not spam) and `return None`, so the engine counts it `empty` (the skip path), not `failed`. In `merge`, the id-less case is now a clean named `ValueError` (`"...missing required 'id' field"`) rather than a bare `KeyError` — no longer reachable from stored artifacts since `read_leaf` filters unsupported leaves first. New `test_unsupported_signature_counts_empty_not_failed` asserts `empty == 1`, `failed == 0`.

src/zagg/sweep.py
espg: 🤖 *from Claude*

Documented in 1349daf. Added a docstring to `SubmapFamily.merge` noting that identity metadata (`collection`/`short_name`/`version`/`footprint`) is inherited from the first child payload, uniform by construction under the one-store-per-product invariant (D7/D19: one product tree per semantic core, one catalog family); `grid_signature` is safe regardless since a node's children share `child_order`, and a multi-catalog store would need reconciliation this fold does not attempt. Doc-only, no behavior change.

tests/test_sweep.py
espg: 🤖 *from Claude*

Fixed in 96a4c74. Added `test_reproject_shim_matches_real_grid_signature`, which pins `_ReprojectTarget(SUBMAP_SIG, N).spatial_signature()` against a real `zagg.grids.healpix.HealpixGrid(N, child_order).spatial_signature()` at orders 0, 1, and SHARD_ORDER. They match on every key except `layout`: SUBMAP_SIG carries the store-layout token (`"flat"`) while a live grid reports its DGGS layout (`"fullsphere"`) — reconciled honestly by comparing modulo `layout` (reproject only swaps the dispatch order, never the DGGS resolution). The `rollup == direct` oracle is now anchored to a genuine grid, not just the shim.

src/zagg/sweep.py
espg: 🤖 *from Claude*

Documented in 190334b. Added a comment at the interior-fold `reproject` branch in `SubmapFamily.merge` noting that `metadata["reproject"]["source_parent_order"]` records the immediately-lower fold level (payloads[0]'s parent_order, one order below this node) — the last hop, not the leaf/shard origin order. Coarsen is transitive, so the folded data equals a direct reproject straight to this order (§8.3, `test_rollup_equals_direct_reproject`); the stamp naming the last hop is by design, not a data discrepancy. Doc-only.

.github/scripts/bench_objects.py
```diff
@@ -270,16 +276,25 @@ def store_object_counts(
             if key in (hive.MANIFEST_NAME, hive.ROOT_COVERAGE_NAME) or _is_run_parquet(key):
                 metadata += 1
                 continue
+            # Sweep rollups (issue #300): `{family}.rollup.json` at any digit
+            # node — second-pass D9 caches with their own bucket, never
+            # write-path objects (and never inside a leaf prefix, so the #215
+            # per-shard guard is untouched).
+            if key.endswith(".rollup.json"):
```

[should-fix] The `.rollup.json` bucket check runs *before* the leaf-prefix membership loop, so it classifies on name alone. A stray object named `*.rollup.json` that lands *inside* a leaf `.zarr/` prefix is silently counted in `objects_rollups` and `continue`d — it never reaches the `per_shard` attribution nor the `other` list, so it's excluded from the audited total (line 371), evades the exact per-shard #215 tripwire, and doesn't bump `objects_other`. That is exactly the class of bypass this model exists to catch. The comment ("rollups never live inside a leaf prefix") is a *design invariant* — but the benchmark's job is to fail when the invariant is violated, not to trust it.

Because legit rollups sit at `{node}/` (siblings of the leaf `.zarr/`, never `startswith` a `leaf_of` prefix), gating the bucket on non-leaf-membership is safe and strictly stricter. Fix: only treat as a rollup when the key is outside every leaf prefix, e.g. move the check into the `else:` of the `for prefix, label in leaf_of.items()` loop (alongside the sibling test), or guard it with `if key.endswith(".rollup.json") and not any(key.startswith(p) for p in leaf_of):`. A misplaced rollup-named object then falls through to `other` and trips the model.

tests/test_sweep.py
```diff
@@ -0,0 +1,706 @@
...
+        df.to_parquet(
+            tmp_path / "stats_20260101T000000Z_old2.parquet",
+            engine="fastparquet",
+            index=False,
+            object_encoding="utf8",
+        )
+        refs = discover_leaves(str(tmp_path))
+        assert refs == []
+        assert "inexact" in caplog.text
+
```
espg, line 776: 🤖 *from Claude (review)*

[nit] Test-coverage gap: no test threads a *windowed* run record all the way through `discover_leaves` into a real `run_sweep`. This test asserts discovery *returns* windowed `(key, window)` pairs, and `TestSweepCli.test_end_to_end_discovers_and_folds` runs a real sweep but only on an *unwindowed* store; `leaves_from_stats_records` is window-tested in isolation. So the seam that actually matters at scale — a windowed leaf name computed from the discovered `window` label feeding the fold — is never exercised end-to-end, and a mismatch between the discovery-side sidecar-name grammar and the fold-side reader would pass CI. Consider extending the CLI end-to-end test (or adding a sibling) with a windowed manifest + windowed leaves so a real `run_sweep` folds them via discovered windows.

espg, line 776: 🤖 *from Claude (review)*

[nit] Test-coverage gap: no test threads a *windowed* run record all the way through `discover_leaves` into a real `run_sweep`. This test asserts discovery *returns* windowed `(key, window)` pairs, and `TestSweepCli.test_end_to_end_discovers_and_folds` runs a real sweep but only on an *unwindowed* store; `leaves_from_stats_records` is window-tested in isolation. So the seam that actually matters at scale — a windowed leaf name computed from the discovered `window` label feeding the fold — is never exercised end-to-end, and a mismatch between the discovery-side sidecar-name grammar and the fold-side reader would pass CI. Consider extending the CLI end-to-end test (or adding a sibling) with a windowed manifest + windowed leaves so a real `run_sweep` folds them via discovered windows.

espg, line 776: 🤖 *from Claude (review)*

[nit] Test-coverage gap: no test threads a *windowed* run record all the way through `discover_leaves` into a real `run_sweep`. This test asserts discovery *returns* windowed `(key, window)` pairs, and `TestSweepCli.test_end_to_end_discovers_and_folds` runs a real sweep but only on an *unwindowed* store; `leaves_from_stats_records` is window-tested in isolation. So the seam that actually matters at scale — a windowed leaf name computed from the discovered `window` label feeding the fold — is never exercised end-to-end, and a mismatch between the discovery-side sidecar-name grammar and the fold-side reader would pass CI. Consider extending the CLI end-to-end test (or adding a sibling) with a windowed manifest + windowed leaves so a real `run_sweep` folds them via discovered windows.

.github/scripts/bench_objects.py

Folded in 0a28fa6. Moved the `.rollup.json` bucket classification into the `else:` of the `for prefix, label in leaf_of.items()` loop, so leaf-prefix membership is now tested FIRST: an object landing inside a leaf `.zarr/` prefix (rollup-named or not) counts into that shard's `objects_per_shard` and trips the exact #215 guard, and only keys OUTSIDE every leaf prefix reach the rollup bucket. Kept the invariant comment, reframed as "classify here, after leaf-membership has failed — a misplaced in-leaf rollup falls to the attribution above and trips #215 instead of vanishing." Added `test_hive_misplaced_rollup_counts_into_shard`: a hand-planted `{leaf}.zarr/stats.rollup.json` now attributes to the shard (`per_shard == {label: 2}`) while the legit sibling `{node}/stats.rollup.json` stays in the bucket (`objects_rollups == 1`).

tests/test_sweep.py
espg: 🤖 *from Claude*

Folded in 7869e50. Added `TestSweepCli.test_end_to_end_windowed_discovery_folds_both_windows`: writes two real windowed leaf sidecars for one shard, writes a real run record carrying the `window` column via `write_run_parquet`, then threads `discover_leaves` -> `run_sweep(families=("stats",))` and asserts the shard-node rollup folded BOTH windows (generation `n_leaves == 2`, `windows == ["2019", "2020"]`, `payload == merge` of the two sidecar records) — exercising the discovery-side sidecar-name grammar against the fold-side reader seam. (This round-3 nit posted three identical times; replying on each so no thread is left open.)

espg: 🤖 *from Claude*

Folded in 7869e50. Added `TestSweepCli.test_end_to_end_windowed_discovery_folds_both_windows`: writes two real windowed leaf sidecars for one shard, writes a real run record carrying the `window` column via `write_run_parquet`, then threads `discover_leaves` -> `run_sweep(families=("stats",))` and asserts the shard-node rollup folded BOTH windows (generation `n_leaves == 2`, `windows == ["2019", "2020"]`, `payload == merge` of the two sidecar records) — exercising the discovery-side sidecar-name grammar against the fold-side reader seam. (This round-3 nit posted three identical times; replying on each so no thread is left open.)

espg: 🤖 *from Claude*

Folded in 7869e50. Added `TestSweepCli.test_end_to_end_windowed_discovery_folds_both_windows`: writes two real windowed leaf sidecars for one shard, writes a real run record carrying the `window` column via `write_run_parquet`, then threads `discover_leaves` -> `run_sweep(families=("stats",))` and asserts the shard-node rollup folded BOTH windows (generation `n_leaves == 2`, `windows == ["2019", "2020"]`, `payload == merge` of the two sidecar records) — exercising the discovery-side sidecar-name grammar against the fold-side reader seam. (This round-3 nit posted three identical times; replying on each so no thread is left open.)

tests/test_benchmark.py
```diff
@@ -2048,12 +2048,13 @@ def test_run_target_threads_store_layout():
...
 
 def test_hive_config_expected_counts_root_moc_optional():
-    # The committed hive arm's model: per-shard DATA is exact (per-leaf count,
+    # The committed hive arm's model: per-shard DATA is exact (11 objects/leaf,
     # the sharded-write-bypass tripwire), but the store-root coverage.moc is a
     # fail-open, regenerable D9 cache (runner.write_root_coverage) that may be
-    # present OR absent -- so store-root metadata is a [1, 2] window (the
-    # morton_hive.json manifest always; +coverage.moc when it lands) and the
-    # total is [12, 13]. A real bypass still fails on the exact per-shard count.
+    # present OR absent -- so store-root metadata is a [1, 3] window (the
```
espg: 🤖 *from Claude (review)*

[should-fix] Stale merge-carried comment: contradicts the assert it introduces. This function-top block says store-root metadata is a `[1, 3]` window and `the total is [14, 16]`, but the assert below (`metadata_min: 1`, `metadata: 4`, `total_min: 12`, `total_max: 15`) and the sibling inline comment (`... aggregation.yaml (optional, issue #299) -> [1, 4]`) both say the window is `[1, 4]` and the total is `[12, 15]`. The `[1, 3]`/`[14, 16]` text predates the #299 metadata bump (3->4, merge 55beb20) and the #304 array-drop total update ([14,16]->[12,15], merge 3b1b012) — it was never reconciled in either merge. Fix: `store-root metadata is a [1, 4] window (the morton_hive.json manifest always; +coverage.moc, +the run stats parquet, and +aggregation.yaml when they land) and the total is [12, 15].`

espg: 🤖 *from Claude*

Reconciled the stale function-top comment to match the assert and sibling inline comment: store-root metadata is now a `[1, 4]` window (morton_hive.json always; +coverage.moc, the run stats parquet, and aggregation.yaml when they land) and the total is `[12, 15]`. Fixed in 6af16e3.

PR #312, opened 2026-07-20
Morton convention: O10 resolution discriminator + self-declared UUID (issue #305)

espg:
Closes #305. Implements the spec half of #262 (D16): morton as a real, declared convention, per the espg-ratified rulings recorded on the issue thread ([O10 ruling record](https://github.com/englacial/zagg/issues/305#issuecomment-5025784723), [acceptance criterion](https://github.com/englacial/zagg/issues/305#issuecomment-5025141183)).

## What this does

1. **mortie spec page (cross-repo)** — [espg/mortie PR #118](https://github.com/espg/mortie/pull/118) (mortie#62) drafts `docs/specification.md`, the normative v1.0 conventions page: bit layout, resolution/Δ table (drift-pinned against `order2res`), decimal grammar + order-29 point/area parse non-injectivity, the `resolution: "exact" | "point"` discriminator and the 29→24 clip rule (`point`-only), the dggs convention block (`name: "morton"`, `coordinate: "morton"`, self-declared UUID), the morton-hive `/1`–`/3` grammars (windowed leaves, the ratified `all` token, `/3` sidecar names, product-name grammar with base-component exclusion), and the coverage-bitmap/root-MOC contracts (zstd level non-normative). espg sign-off on that page is decision 3 of this issue.
2. **zagg conventions surface** (`src/zagg/grids/morton.py`, `src/zagg/grids/healpix.py`):
   - `MORTON_CONVENTION` — the self-declared `zarr_conventions` entry with the **permanent** UUID `3e22156d-ea9e-4e01-95fe-e3809a4b41e7` (minted once; pinned by test so accidental regeneration fails loudly), name `morton-dggs`, spec/schema URLs pointing at the mortie page. The upstream dggs-registry ask (#72 ask 3) proceeds in parallel; `zarr_conventions` is a list so both entries later coexist.
   - `RESOLUTION_EXACT` / `RESOLUTION_POINT` — the O10 discriminator constants, importable where the temporal/point paths will need `point` later (the seam; no speculative point-writer implementation).
   - `HealpixGrid._dggs_attrs` now emits `resolution: "exact"` in the `dggs` block — grid-derived cell coordinates are exact **by construction**, under every `cell_ids_encoding`.
   - The attrs **flip** to `name: "morton"` / `coordinate: "morton"` + the convention entry is deliberately NOT here — that is #304's phase 1 (the writer-flip PR, based on this branch).
[...]

PR #314, opened 2026-07-20
Morton-only writer flip: stop writing cell_ids (issue #304)

### LINKED ISSUE #304: Morton-only writer flip: stop writing cell_ids (implements #262 / D16)

espg:
Closes #304. **Blocked by #312** (issue #305 — this branch is based on `claude/305-morton-convention` and consumes its `MORTON_CONVENTION` / `RESOLUTION_EXACT` constants; merge #312 first, then this).

Implements the writer half of #262 (D16 — morton-only storage): stop writing the redundant `cell_ids` array, re-point the convention attrs at the `morton` coordinate under the **distinct grid `name: "morton"`** (never `name: "healpix"` + `indexing_scheme: "morton"` — scheme-blind readers must hard-reject, not silently misread; matches moczarr's xdggs `grid_name`). This is the **one remaining breaking store change** in this family (D19 made #299 additive); moczarr golden-vector coordination for this change is tracked on espg/moczarr#11.

## Phases

- [x] Phase 1 — **attrs flip behind the `cell_ids_encoding` seam**: a morton-declared store (`cell_ids_encoding: morton`) now emits `dggs: {name: "morton", coordinate: "morton", resolution: "exact", ...}` (no `indexing_scheme` key) and appends the self-declared convention entry (#305, permanent UUID) to the `zarr_conventions` list. Nested stores byte-unchanged.
- [x] Phase 2 — **default flip**: `cell_ids` no longer written by default (template member + coords column both keyed off the flag); `output.grid.emit_cell_ids: true` escape hatch for transition stores (gridlook demos until the TS morton decode lands — espg/gridlook#1 Phases 4/6).
- [x] Phase 4 — **raster path joins the flip** (espg-ruled 2026-07-21, overruling this PR's earlier "distinct design change / S2 PR #219 follow-on" scope-out — the D16 premise is ONE default cell coordinate everywhere): the `(time, cells)` raster writer stores `morton` (packed u64 words) as its sole cell coordinate, the raster group/leaf specs carry the SAME morton-declared dggs attrs block as the spatial path (one reader contract), and raster `cell_ids` follows the same lifecycle — behind `emit_cell_ids`, retired with the same knob, never a separate schedule. No schema blocker surfaced — the flip was mechanical (`_raster_members`, `raster_group_spec`/`raster_leaf_spec` attrs, `emit_raster_leaf_template`, `write_raster_coords`; commit `03b3b51`).
- [x] Phase 3 — **removal + knob retirement**: `cell_ids_encoding` retires with a pointed error (#135 inverts: morton stored, NESTED derived at read via moczarr fabrication); shipped configs drop the `cell_ids` coordinate declarations; fixtures/tests regenerate; docs (`docs/morton_arrow.md`, `docs/design/architecture.md`) updated to the morton-only default. **`.github/` note:** `.github/scripts/bench_objects.py` needed NO change — its expected-count model derives from `grid.spec()`/`grid.shard_spec()` by design, so dropping the cell_ids member flows through automatically; only the test-side expected constants updated (`tests/test_benchmark_objects.py`, `tests/test_benchmark.py`: per-shard 4→3 arrays, hive per-leaf 12→10 objects).

## Branch state after the main merge (espg-directed, merge commit `ddd53fd`)

`origin/main` merged in (merge commit, no rebase): the branch is now **current with the full finalized spec surface** — #312/#305 conventions, #316/#299 product roots + semantic-core hasher + `has_run`, D16–D24 — with the raster phases the remaining in-flight work of this PR. **One conflicted file**: `tests/test_benchmark.py` (hive object-count assertion) — both sides changed it and the effects COMPOSE (this branch: one fewer array per leaf, 3 arrays; main/#299: +`aggregation.yaml` optional root object) — resolved to the observed merged model (`per_shard 10/10, metadata 4/min 1, total 11–14`), verified by running `bench_objects.expected_object_counts` against the merged tree. Everything else auto-merged with both intents preserved (main's D19 `pipeline.type` text AND this branch's O10 encoding-carried rewrite both survive in `sparse_coverage.md`). Full suite post-merge: 2562 passed, 12 skipped. Pre-existing (not merge-related): `test_runner.py`'s `./out.zarr` cwd-scribbling can flake under random test ordering — test-isolation debt flagged, untouched here.

## Post-review supersession (2026-07-21)

- **O10 final ruling (espg, via the mortie PR #118 review)**: the `resolution` declaration is **deleted** — kind is carried by the packed-word encoding itself (area words exact at every order; point ids are points; no metadata field; the 29→24 clip is a viewer-side transient cast, not spec semantics). Commit `ea35021` removes the field emission + `RESOLUTION_*` constants and rewrites the O10 record in `docs/design/sparse_coverage.md`; the order-29 decimal parse tie-break is pinned on the mortie spec page §4. The optional decimal kind suffix `p` (espg-ruled) is mortie-side only; zagg strings are all unmarked area words and unchanged.

## Scope notes

- ~~Raster stores are out of scope~~ **Overruled by espg (2026-07-21)** — raster is IN scope; see Phase 4 above.
- The `emit_cell_ids` hatch emits the legacy NESTED `cell_ids` array *in addition to* morton; the dggs attrs stay morton-declared. The hatch exists only until the gridlook morton path lands.

## Testing

- Phase 3: default store attrs are morton-declared for every HEALPix store; retired-knob pointed error; hatch emits NESTED only; `signature()` swaps `cell_ids_encoding` for `emit_cell_ids`; full suite green locally (2517 passed, 12 skipped, lambda-build excluded as below).
- Phase 2: default flip green across 13 test files (2524 passed at that commit); new `conftest.nested_ids()` derives NESTED index positions from morton (the same read-side fabrication moczarr performs).
- Phase 1: attrs assertions in `tests/test_grids.py` (`TestCellIdsEncoding`) — morton-declared block shape, convention-entry list, nested-store byte-stability; full suite green locally (2519 passed, 12 skipped; `tests/test_lambda_build.py` excluded — its function-build test fails on this machine for an environmental reason (local pip resolves against Python 3.10, pre-existing on `main`), unrelated).
- `ruff check` / `ruff format --check` clean (pre-existing `N818` in `src/zagg/registry.py` on `main` excluded, flagged in #312 too).

## Questions for review

1. *(resolved — espg ruled raster IN scope; implemented as Phase 4.)*
2. **Post-retirement attrs for hatch stores** (phase 3): a store written with `emit_cell_ids: true` keeps the morton-declared dggs attrs (cell_ids is a legacy extra array, not the declared coordinate). Confirm that's the intended transition-store shape for gridlook.

### DIFF
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -294,26 +294,23 @@ def validate_config(config: PipelineConfig) -> None:
                 f"output.grid.layout must be 'fullsphere' (got {layout!r}; "
                 "the deprecated 'dense' layout was removed — issue #88)"
             )
-        # Optional cell_ids encoding (issue #135): "nested" (default, the DGGS
-        # standard) or "morton" (emit the packed morton words as cell_ids — a
-        # test/prototype capability). HEALPix-only: rectilinear grids have no
-        # cell_ids coordinate, so the knob would silently do nothing there.
-        encoding = grid.get("cell_ids_encoding")
-        if encoding is not None:
-            if encoding not in ("nested", "morton"):
-                raise ValueError(
-                    f"output.grid.cell_ids_encoding must be 'nested' or 'morton' (got {encoding!r})"
-                )
-            if grid["type"] != "healpix":
-                raise ValueError(
-                    "output.grid.cell_ids_encoding only applies to healpix grids "
-                    f"(grid type is {grid['type']!r})"
-                )
-        # Transition escape hatch (issue #304 phase 2): emit the legacy
+        # The issue #135 knob is retired (issue #304 phase 3, D16 — #135
+        # inverts): morton is the stored cell coordinate and NESTED ids are
+        # DERIVED at read (moczarr fabrication). A config still carrying the
+        # key gets a pointed error, never a silent ignore.
+        if grid.get("cell_ids_encoding") is not None:
+            raise ValueError(
+                "output.grid.cell_ids_encoding was retired (issue #304): morton is "
+                "the stored cell coordinate and NESTED ids are derived at read; "
+                "to keep writing the legacy NESTED cell_ids array for transition "
+                "stores set output.grid.emit_cell_ids: true"
+            )
+        # Transition escape hatch (issue #304 phase 2): emit the legacy NESTED
         # cell_ids array ALONGSIDE the morton coordinate. Default off — morton
         # is the only stored cell coordinate (D16); the hatch exists for
         # transition stores (browser-direct demos) until the gridlook morton
-        # path lands. HEALPix-only for the same reason as cell_ids_encoding.
+        # path lands. HEALPix-only: rectilinear grids have no cell_ids
+        # coordinate, so the flag would silently do nothing there.
         emit = grid.get("emit_cell_ids")
         if emit is not None:
             if not isinstance(emit, bool):
@@ -324,15 +321,14 @@ def validate_config(config: PipelineConfig) -> None:
                     f"(grid type is {grid['type']!r})"
                 )
         # The legacy output.grid.indexing_scheme key is descriptive only (the
-        # shipped configs carry "nested"); it does NOT select the cell_ids
-        # encoding. Reject any other value so a user reaching for it lands on the
-        # real knob instead of a silently-NESTED store.
+        # shipped configs carry "nested"); it never selected an encoding.
+        # Reject any other value so a user reaching for it fails loudly.
         legacy_scheme = grid.get("indexing_scheme")
         if legacy_scheme is not None and legacy_scheme != "nested":
             raise ValueError(
                 f"output.grid.indexing_scheme is descriptive and must be 'nested' "
-                f"(got {legacy_scheme!r}); to emit morton words as cell_ids set "
-                f"output.grid.cell_ids_encoding: morton"
+                f"(got {legacy_scheme!r}); the store's declared coordinate is morton "
+                f"(D16 — issue #304)"
             )
 
     # Validate the optional per-cell carrier (issue #132). Mirrors the worker's
@@ -2059,29 +2055,6 @@ def get_sharded(config: PipelineConfig, default: bool = False) -> bool:
     not cost the ~K-fold object blow-up).
     """
     return bool(config.output.get("grid", {}).get("sharded", default))
-
-
-def get_cell_ids_encoding(config: PipelineConfig) -> str:
-    """Return the HEALPix ``cell_ids`` coordinate encoding (issue #135).
-
-    ``"nested"`` (default) stores the standardized NESTED HEALPix cell IDs.
-    ``"morton"`` stores the packed morton words instead — the same ``uint64``
-    words the ``morton`` coordinate carries — opening test/prototype flows that
-    index by morton directly. Default behavior (key absent, explicit ``null``,
-    or ``"nested"``) is byte-identical to a pre-flag run.
-
-    Parameters
-    ----------
-    config : PipelineConfig
-
-    Returns
-    -------
-    str
-        ``"nested"`` (default) or ``"morton"``.
-    """
-    # A present-but-null key (YAML ``cell_ids_encoding:``) must fall back to the
-    # default too — the same treatment ``from_config`` gives a null ``layout``.
-    return config.output.get("grid", {}).get("cell_ids_encoding") or "nested"
 
 
 def get_emit_cell_ids(config: PipelineConfig) -> bool:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -309,6 +309,20 @@ def validate_config(config: PipelineConfig) -> None:
                     "output.grid.cell_ids_encoding only applies to healpix grids "
                     f"(grid type is {grid['type']!r})"
                 )
+        # Transition escape hatch (issue #304 phase 2): emit the legacy
+        # cell_ids array ALONGSIDE the morton coordinate. Default off — morton
+        # is the only stored cell coordinate (D16); the hatch exists for
+        # transition stores (browser-direct demos) until the gridlook morton
+        # path lands. HEALPix-only for the same reason as cell_ids_encoding.
+        emit = grid.get("emit_cell_ids")
+        if emit is not None:
+            if not isinstance(emit, bool):
+                raise ValueError(f"output.grid.emit_cell_ids must be a boolean (got {emit!r})")
+            if emit and grid["type"] != "healpix":
+                raise ValueError(
+                    "output.grid.emit_cell_ids only applies to healpix grids "
+                    f"(grid type is {grid['type']!r})"
+                )
         # The legacy output.grid.indexing_scheme key is descriptive only (the
         # shipped configs carry "nested"); it does NOT select the cell_ids
         # encoding. Reject any other value so a user reaching for it lands on the
@@ -2068,6 +2082,19 @@ def get_cell_ids_encoding(config: PipelineConfig) -> str:
     # A present-but-null key (YAML ``cell_ids_encoding:``) must fall back to the
     # default too — the same treatment ``from_config`` gives a null ``layout``.
     return config.output.get("grid", {}).get("cell_ids_encoding") or "nested"
+
+
+def get_emit_cell_ids(config: PipelineConfig) -> bool:
+    """Whether to keep writing the legacy ``cell_ids`` array (issue #304).
+
+    Default ``False`` (the D16 flip): ``morton`` is the only stored cell
+    coordinate; NESTED ids are derived at read (moczarr fabrication). ``True``
+    is the transition escape hatch — the ``cell_ids`` array is written in
+    ADDITION to ``morton`` (values per ``cell_ids_encoding``) so legacy
+    browser-direct consumers keep working until the gridlook morton decode
+    lands; the dggs attrs are unaffected by the hatch.
+    """
+    return bool(config.output.get("grid", {}).get("emit_cell_ids", False))
 
 
 def get_store_path(config: PipelineConfig) -> str | None:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -2063,7 +2063,7 @@ def get_emit_cell_ids(config: PipelineConfig) -> bool:
     Default ``False`` (the D16 flip): ``morton`` is the only stored cell
     coordinate; NESTED ids are derived at read (moczarr fabrication). ``True``
     is the transition escape hatch — the ``cell_ids`` array is written in
-    ADDITION to ``morton`` (values per ``cell_ids_encoding``) so legacy
+    ADDITION to ``morton`` (holding the NESTED encode) so legacy
     browser-direct consumers keep working until the gridlook morton decode
     lands; the dggs attrs are unaffected by the hatch.
     """
--- a/src/zagg/grids/healpix.py
+++ b/src/zagg/grids/healpix.py
@@ -14,7 +14,6 @@ from zagg.config import (
     default_config,
     get_agg_fields,
     get_aoi_mask,
-    get_cell_ids_encoding,
     get_emit_cell_ids,
     get_output_signature,
     output_field_signature,
@@ -121,20 +120,6 @@ class HealpixGrid:
         self.sharded = bool(sharded)
         self.layout = layout
         self.config = config or default_config("atl06")
-        # cell_ids coordinate encoding (issue #135): "nested" (default, the DGGS
-        # standard) or "morton" (emit the packed morton words as cell_ids — a
-        # test/prototype capability). Default is byte-identical to a pre-flag run.
-        # Re-validated here (not only in validate_config) because both coords_of
-        # (which interprets this string for the hatch's cell_ids VALUES) and
-        # _dggs_attrs (which keys the morton-declared block on it) consume it: an
-        # unvalidated third value would write NESTED values while declaring a
-        # different block — a mis-decode for consumers.
-        self.cell_ids_encoding = get_cell_ids_encoding(self.config)
-        if self.cell_ids_encoding not in ("nested", "morton"):
-            raise ValueError(
-                f"Unknown cell_ids_encoding: {self.cell_ids_encoding!r} "
-                "(expected 'nested' or 'morton')"
-            )
         # D16 default flip (issue #304 phase 2): the legacy cell_ids array is
         # written only under the explicit transition hatch; morton is the one
         # stored cell coordinate. Both the template member (_group_spec) and
@@ -416,8 +401,8 @@ class HealpixGrid:
         it is stored as ``uint64`` on disk via :mod:`zagg.grids.morton` and is
         the ONLY cell coordinate written by default (D16, issue #304). The
         ``output.grid.emit_cell_ids: true`` transition hatch adds the legacy
-        ``cell_ids`` array — NESTED ``uint64``, or the packed morton words under
-        ``cell_ids_encoding: morton`` (issue #135).
+        NESTED ``uint64`` ``cell_ids`` array (the issue #135 encoding knob was
+        retired in issue #304 phase 3 — morton stored, NESTED derived).
         """
         return self.coords_of(self.children(shard_key))
 
@@ -432,10 +417,7 @@ class HealpixGrid:
         children = np.asarray(children)
         coords: dict = {"morton": to_morton_array(children)}
         if self.emit_cell_ids:
-            if self.cell_ids_encoding == "morton":
-                coords["cell_ids"] = np.asarray(children, dtype=np.uint64)
-            else:
-                coords["cell_ids"] = self.encode_cell_ids(children)
+            coords["cell_ids"] = self.encode_cell_ids(children)
         return coords
 
     # ── identity / nesting ───────────────────────────────────────────────
@@ -460,18 +442,16 @@ class HealpixGrid:
         """Canonical fingerprint of the grid's defining parameters.
 
         The full fingerprint: the spatial layout (:meth:`spatial_signature`)
-        plus the Option-B output-field set and the ``cell_ids_encoding``
-        (issue #135 — mixed encodings must never co-aggregate; ``nests_with``
-        keys on both). The shard-map reuse guard keys on the spatial part
-        only (#89).
+        plus the Option-B output-field set and the ``emit_cell_ids`` hatch
+        state (issue #304 — the hatch changes the leaf schema, so mixed
+        states must never co-aggregate; ``nests_with`` keys on both). The
+        shard-map reuse guard keys on the spatial part only (#89). (The
+        issue #135 ``cell_ids_encoding`` field was retired in issue #304
+        phase 3.)
         """
         return {
             **self.spatial_signature(),
             "output_fields": output_field_signature(self.config),
-            "cell_ids_encoding": self.cell_ids_encoding,
-            # The transition hatch (issue #304 phase 2) changes the leaf
-            # schema (an extra cell_ids array), so it is part of the full
-            # fingerprint — mixed hatch states must never co-aggregate.
             "emit_cell_ids": self.emit_cell_ids,
         }
 
@@ -481,18 +461,13 @@ class HealpixGrid:
         Any two HEALPix grids nest (the nested hierarchy subdivides 4-for-1 at
         every order), provided they declare the same Option-B output-field set
         (issue #29) — co-aggregated grids must produce the same scalar/vector
-        schema — and the same ``cell_ids_encoding`` (issue #135): NESTED ids
-        and morton words are different id spaces, so a consumer joining
-        co-aggregated products on ``cell_ids`` would silently mismatch.
-        Cross-family (e.g. rectilinear) never nests.
+        schema — and the same ``emit_cell_ids`` hatch state (issue #304): one
+        store carrying an extra cell_ids array while the other does not is a
+        schema mismatch. Cross-family (e.g. rectilinear) never nests.
         """
         if not isinstance(other, HealpixGrid):
             return False
-        if self.cell_ids_encoding != other.cell_ids_encoding:
-            return False
         if self.emit_cell_ids != other.emit_cell_ids:
-            # The hatch changes the leaf schema (issue #304 phase 2): one
-            # store carries a cell_ids array, the other does not.
             return False
         return output_field_signature(self.config) == output_field_signature(other.config)
 
@@ -714,28 +689,20 @@ class HealpixGrid:
             },
             "compression": "none",
         }
-        if self.cell_ids_encoding == "morton":
-            # D16 / issue #304 phase 1: a morton-declared store uses the
-            # DISTINCT grid name "morton" with the typed `morton` coordinate —
-            # never name: "healpix" + indexing_scheme: "morton", which a
-            # scheme-blind reader silently misreads as NESTED (garbage
-            # renders); an unknown grid name makes it hard-reject with a
-            # diagnostic instead, and matches moczarr's xdggs registration.
-            # The self-declared convention entry (issue #305; permanent UUID)
-            # rides the zarr_conventions LIST alongside the generic dggs
-            # entry, so a future upstream registry entry can coexist too.
-            return {
-                "zarr_conventions": [dggs_entry, dict(MORTON_CONVENTION)],
-                "dggs": {"name": "morton", **common, "coordinate": "morton"},
-            }
+        # D16 / issue #304 phase 3: every HEALPix aggregation store is
+        # morton-declared — the DISTINCT grid name "morton" with the typed
+        # `morton` coordinate. Never name: "healpix" + indexing_scheme:
+        # "morton", which a scheme-blind reader silently misreads as NESTED
+        # (garbage renders); an unknown grid name makes it hard-reject with a
+        # diagnostic instead, and matches moczarr's xdggs registration. The
+        # self-declared convention entry (issue #305; permanent UUID) rides
+        # the zarr_conventions LIST alongside the generic dggs entry, so a
+        # future upstream registry entry can coexist too. A hatch store
+        # (emit_cell_ids) carries the legacy NESTED array as an EXTRA member;
+        # the declared coordinate stays morton.
         return {
-            "zarr_conventions": [dggs_entry],
-            "dggs": {
-                "name": "healpix",
-                "indexing_scheme": "nested",
-                **common,
-                "coordinate": "cell_ids",
-            },
+            "zarr_conventions": [dggs_entry, dict(MORTON_CONVENTION)],
+            "dggs": {"name": "morton", **common, "coordinate": "morton"},
         }
 
 
--- a/src/zagg/grids/healpix.py
+++ b/src/zagg/grids/healpix.py
@@ -27,7 +27,12 @@ from zagg.grids.base import (
     vector_array_spec,
     vlen_dtype_warning_suppressed,
 )
-from zagg.grids.morton import RESOLUTION_EXACT, morton_decimal, to_morton_array
+from zagg.grids.morton import (
+    MORTON_CONVENTION,
+    RESOLUTION_EXACT,
+    morton_decimal,
+    to_morton_array,
+)
 
 HEALPIX_BASE_CELLS: int = 12
 # Reference order at which ``assign`` resolves points before ``cells_of`` /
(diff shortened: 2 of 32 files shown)


tests/test_grids.py
```diff
@@ -1024,9 +1024,78 @@ def test_morton_encoding_roundtrips_through_zarr(self):
...
+    def test_dggs_attrs_declare_resolution_exact(self):
+        # O10 discriminator (issue #305): grid-derived cell coordinates are
+        # "exact" by construction, under every encoding — "point" is reserved
+        # for location-derived id paths that don't exist in zagg outputs.
+        from zagg.grids.morton import RESOLUTION_EXACT
+
+        for enc in (None, "nested", "morton"):
+            assert self._grid(enc)._dggs_attrs()["dggs"]["resolution"] == RESOLUTION_EXACT
+
+    def test_dggs_attrs_reach_written_store(self):
```
espg, line 1093: 🤖 *from Claude (review)*

**F1 (minor) — no written-store test for the morton attrs block.** The three new tests all assert `_dggs_attrs()`'s in-memory return value; the round-trip guard just below (`test_dggs_attrs_reach_written_store`) — which exists precisely because "that store path is what #305 readers key on" — still exercises only the default (nested) grid. Since #304's whole subject is what a morton-declared store *writes*, the morton branch (name `"morton"`, `coordinate: "morton"`, the 2-entry `zarr_conventions` list) is never pinned past `spec()->emit_template->to_zarr`. Same serialization path as nested, so risk is low, but a morton variant here (or a sibling) would close the loop and catch e.g. the 2-element list or the morton coordinate name not surviving the JSON round-trip.

tests/test_grids.py
```diff
@@ -1024,9 +1024,78 @@ def test_morton_encoding_roundtrips_through_zarr(self):
...
+
+        attrs = self._grid("morton")._dggs_attrs()
+        assert attrs["dggs"]["coordinate"] == "morton"
+        assert attrs["dggs"]["name"] == "morton"
+        assert MORTON_CONVENTION in attrs["zarr_conventions"]
+        names = [c["name"] for c in attrs["zarr_conventions"]]
+        assert names == ["dggs", "morton-dggs"]
+
+    def test_nested_attrs_unchanged_by_flip(self):
+        # The default store's attrs block is byte-identical in shape to the
```
espg: 🤖 *from Claude (review)*

**F2 (nit) — "byte-identical in shape" overstates it; the key *order* changed.** Refactoring the nested block onto `**common` reorders the dggs dict keys: pre-flip was `name, refinement_level, indexing_scheme, resolution, spatial_dimension, ellipsoid, coordinate, compression`; post-flip is `name, indexing_scheme, refinement_level, resolution, spatial_dimension, ellipsoid, compression, coordinate`. The key *set* and *values* are identical (and every real zarr.json consumer parses order-independently, so this is harmless), but the serialized JSON bytes are not — so "byte-identical" reads as a stronger guarantee than the test checks (it only asserts name/coordinate/len). Suggest softening to "same key set and values".

espg:
🤖 *from Claude (review)*

**F3 (nit) — stale comment at `src/zagg/grids/healpix.py:127` (outside the phase-1 diff, so noting here).** The `__init__` comment describes `_dggs_attrs` as recording "the recorded indexing_scheme". After the phase-1 flip that is only true for the nested/default branch — a morton-declared store's `_dggs_attrs` records the grid `name`/`coordinate` and deliberately emits *no* `indexing_scheme` (correct: the dggs v1 schema requires `indexing_scheme` only when `name == "healpix"`; the old `name: "healpix"` + `indexing_scheme: "morton"` shape actually violated that enum, so the flip improves conformance). The comment's real point (both `coords_of` and `_dggs_attrs` branch on `cell_ids_encoding`) still holds; just the "recorded indexing_scheme" gloss is now nested-only.

tests/test_grids.py
espg: 🤖 *from Claude*

Fixed in 92c2011. Added `test_morton_dggs_attrs_reach_written_store` in `TestCellIdsEncoding`, mirroring `test_dggs_attrs_reach_written_store` (MemoryStore + emit_template + open_group(path="8")) but built from `self._grid("morton")`. It asserts the WRITTEN group attrs carry dggs name "morton", coordinate "morton", resolution "exact", no `indexing_scheme` key, and a 2-entry `zarr_conventions` list whose names are `["dggs", "morton-dggs"]`.

tests/test_grids.py
espg: 🤖 *from Claude*

Fixed in a0ea8b6. Reworded the comment in `test_nested_attrs_unchanged_by_flip` to say the block keeps the "same key set and values (key order is not part of the contract; JSON attrs parse order-independently)", dropping the "byte-identical in shape" phrasing that overstated the guarantee.

espg:
🤖 *from Claude*

Re the F3 finding (PR-thread comment 5025987100):

> stale `__init__` comment in src/zagg/grids/healpix.py (~line 127) still says _dggs_attrs records "the recorded indexing_scheme" from cell_ids_encoding — post-flip the morton branch has no indexing_scheme.

Fixed in c34bb80. Updated the comment to describe the post-flip roles: coords_of interprets the string for the hatch's cell_ids VALUES, and _dggs_attrs keys the morton-declared block on it (no longer any recorded indexing_scheme for morton).

src/zagg/grids/healpix.py
```diff
@@ -566,9 +586,21 @@ def _shard(arr):
 
         members = {}
         for name, meta in self.config.aggregation.get("coordinates", {}).items():
+            # D16 default flip (issue #304 phase 2): a declared cell_ids
+            # coordinate is honored only under the emit_cell_ids transition
+            # hatch — otherwise the member is skipped so the template never
+            # carries an array the writer will not fill. Keyed off the same
+            # flag as coords_of, so template and writes cannot disagree.
+            if name == "cell_ids" and not self.emit_cell_ids:
```
espg, line 565: 🤖 *from Claude (review)*

**F4 (major) — the default (D16) store's `dggs.coordinate` names a `cell_ids` array that is no longer on disk.**

Phase 2's headline (this very hunk's comment) is that the writer and the template "cannot disagree" because both key off `emit_cell_ids`. But the template has a third half — the DGGS attrs — and those still disagree. With the default (`emit_cell_ids=false`, `cell_ids_encoding=nested`), `_group_spec` writes **only** `morton`, yet `_dggs_attrs()` (healpix.py:730-738, unchanged) still emits:

```
"dggs": {"name": "healpix", "indexing_scheme": "nested", ..., "coordinate": "cell_ids"}
```

So every default store advertises `coordinate: cell_ids` (a NESTED-id array) that it does not contain — and `morton` is *not* a drop-in substitute (packed words, `base+1` prefix, not NESTED ids). `test_nested_attrs_unchanged_by_flip` pins this as intended.

Contrast phase 1's deliberate choice for the morton encoding: it flipped `name`→`morton` + `coordinate`→`morton` precisely so a scheme-blind reader hard-rejects instead of misreading. By that same logic the default store is now only readable through moczarr's read-side `cell_ids` fabrication — a generic xdggs open resolves `dggs.coordinate → cell_ids` and `KeyError`s.

Please confirm this is intended (stores now *require* the moczarr fabrication layer to be self-describing) and, if so, document the on-disk vs logical-coordinate split — because as written it contradicts the "template and writes cannot disagree" guarantee this commit sells. Option: point `coordinate` at `morton` in the default case too (keeping `name: healpix` / `indexing_scheme: nested` for the fabrication hint), so the named coordinate is one that actually exists on disk.

src/zagg/grids/healpix.py
```diff
@@ -416,14 +429,13 @@ def coords_of(self, children) -> dict:
...
-        else:
-            cell_ids = self.encode_cell_ids(children)
-        return {
-            "morton": to_morton_array(children),
-            "cell_ids": cell_ids,
-        }
+        coords: dict = {"morton": to_morton_array(children)}
+        if self.emit_cell_ids:
+            if self.cell_ids_encoding == "morton":
+                coords["cell_ids"] = np.asarray(children, dtype=np.uint64)
```

**F5 (minor) — hatch + `cell_ids_encoding: morton` writes a `cell_ids` array that byte-duplicates `morton`.**

Under `emit_cell_ids: true` with `cell_ids_encoding: morton`, this branch sets `coords["cell_ids"] = np.asarray(children, uint64)` — the same packed morton words already stored (as `uint64`) under the `morton` coordinate. The store then carries two identical uint64 arrays under different names.

The hatch exists for legacy browser-direct consumers that want the NESTED `cell_ids` array, so pairing it with the morton encoding is pointless duplication with no consumer. Consider either rejecting the `emit_cell_ids=true` + `cell_ids_encoding=morton` combination in `validate_config` (the way other no-op knob combos are fenced), or documenting that the combination is intentional #135 compatibility. Not blocking — just a footgun with no guard.

src/zagg/config.py
```diff
@@ -309,6 +309,20 @@ def validate_config(config: PipelineConfig) -> None:
...
+        # Transition escape hatch (issue #304 phase 2): emit the legacy
+        # cell_ids array ALONGSIDE the morton coordinate. Default off — morton
+        # is the only stored cell coordinate (D16); the hatch exists for
+        # transition stores (browser-direct demos) until the gridlook morton
+        # path lands. HEALPix-only for the same reason as cell_ids_encoding.
+        emit = grid.get("emit_cell_ids")
+        if emit is not None:
+            if not isinstance(emit, bool):
+                raise ValueError(f"output.grid.emit_cell_ids must be a boolean (got {emit!r})")
+            if emit and grid["type"] != "healpix":
```
espg, line 318: 🤖 *from Claude (review)*

**F7 (nit) — gating asymmetry vs `cell_ids_encoding` on non-healpix grids.**

`cell_ids_encoding` (line 307) rejects *any* value on a non-healpix grid, including the no-op `nested`. `emit_cell_ids` here rejects only when `emit` is truthy — `emit_cell_ids: false` on a rect grid passes validation silently. This is defensible (false is the default no-op, harmless) and I'd leave it as-is, but the two healpix-only knobs now reject differently for the same "declared on a rect grid" mistake. If you want parity, either reject `emit_cell_ids` presence on non-healpix regardless of value, or relax `cell_ids_encoding` to only reject `morton` there. Flagging for a conscious choice, not a fix.

espg:
🤖 *from Claude (review)*

**F6 (minor) — docs now describe the pre-flip default.**

Phase 2 flips the *default*, but the narrative docs still state cell_ids is written by default:

- `docs/morton_arrow.md:41` — "`cell_ids` stays the standardized **NESTED HEALPix** id (`uint64`) by default." After D16 the default store emits `morton` only; the NESTED `cell_ids` array is now the `emit_cell_ids: true` hatch. That section (and its `dggs.indexing_scheme` claim at :43) needs the flip folded in.
- `docs/design/architecture.md:95` and `:186` list `cell_ids` as a store array in the on-disk tree diagrams — now conditional on the hatch.

These describe behavior the phase-2 flip changed (not the phase-3 knob retirement), so they read wrong as of this commit. Cheap to fix in this PR.

espg:
🤖 *from Claude (review)*

**F8 (nit) — `emit_cell_ids` × `aoi_mask` combination is untested.**

The new hatch tests (`TestCellIdsEncoding`) never turn on `output.aoi_mask`, and the aoi_mask bench test never sets the hatch, so `_group_spec` emitting *both* the synthesized/declared `cell_ids` member **and** the `aoi_mask` member in one template goes unexercised. The two are independent branches so this is low-risk, but one combined round-trip (or a `spec().members` assertion carrying both) would close the gap the reviewer flagged.

src/zagg/grids/healpix.py
espg: 🤖 *from Claude*

Resolved by phase 3 (9467637). `_dggs_attrs()` is now unconditional: every HEALPix store — default and hatch alike — is morton-declared (`name: "morton"`, `coordinate: "morton"`, **no** `indexing_scheme` key, both `zarr_conventions` entries `["dggs", "morton-dggs"]`). There is no longer a store that advertises a `cell_ids` coordinate it does not contain; the default store self-describes as morton and a scheme-blind reader hard-rejects rather than misreading. `test_dggs_attrs_are_morton_declared` and `test_dggs_attrs_reach_written_store` pin both the in-memory and the written-store attrs (the latter for both `emit=None` and `emit=True`).

src/zagg/grids/healpix.py
espg: 🤖 *from Claude*

Resolved by phase 3 (9467637). The issue #135 `cell_ids_encoding` knob is retired: any config still carrying it now fails `validate_config` with a migration pointer (config.py:301-307), so the `emit_cell_ids: true` + `cell_ids_encoding: morton` duplication path no longer exists. The hatch always emits the legacy NESTED `uint64` array — `coords["cell_ids"] = encode_cell_ids(children)` — never the packed morton words, so it can never byte-duplicate the `morton` coordinate.

src/zagg/config.py
espg: 🤖 *from Claude*

Leaving as-is by design. Phase 3 changed the comparison landscape: `cell_ids_encoding` no longer "rejects any value on a non-healpix grid" — it is a **retired** knob that errors on any presence, any grid (config.py:301). So the two are not the same kind of check anymore. `emit_cell_ids` is a live knob whose `false` is the universal no-op default (morton-only) valid on every grid; only a `true` value changes the leaf schema, and that is what is healpix-gated (config.py:318). Rejecting `emit_cell_ids: false` on a rect grid would reject the default behavior, which reads as a pointless footgun rather than a guard. Kept minimal.

espg:
🤖 *from Claude*

Re **F6** (docs describe the pre-flip default): resolved by phase 3 (9467637). Verified every spot the finding named:

- `docs/morton_arrow.md` — the section is now `## `cell_ids`: morton-only storage (D16)` and states "**`morton` is the only stored cell coordinate**", NESTED ids "**derived at read**", and the `emit_cell_ids: true` hatch as the transition escape. The stale "cell_ids stays the standardized NESTED HEALPix id by default" / `dggs.indexing_scheme` claim is gone; the only remaining `indexing_scheme` mention (:67) correctly notes the descriptive config key must stay `nested`.
- `docs/design/architecture.md` — both on-disk tree diagrams now condition cell_ids on the hatch: `(cell_ids only under emit_cell_ids — issue #304)` (:97) and `NESTED cell_ids derived at read — issue #304` (:160).

A docs grep for `cell_ids.*by default` / `standardized.*nested` comes back empty. No new commit needed.

espg:
🤖 *from Claude*

Re **F8** (`emit_cell_ids` × `aoi_mask` combination untested): fixed in 30bf8cd. Added `test_hatch_with_aoi_mask_emits_both_members` to `TestCellIdsEncoding` (tests/test_grids.py) — it builds a hatch grid (`emit_cell_ids: true`) whose config also sets `output.aoi_mask: true` and asserts `spec().members` carries **all three** independent members in one template: `morton` (always-present coordinate), `cell_ids` (the #304 hatch member) and `aoi_mask` (the #101 strict-AOI bool member, dtype checked). Closes the gap where the two independent `_group_spec` branches were never exercised together. Full suite green: `pytest tests/test_grids.py tests/test_config.py -q` → 357 passed.

espg:
🤖 *from Claude (review)*

Phase 3 (`9467637`) adversarial review — the D16 breaking store flip. Verified: config retirement error + `emit_cell_ids` accessor/validation, `_dggs_attrs` now uniformly morton-declared (name/coordinate "morton", no `indexing_scheme`, both convention entries, resolution exact), `coords_of` hatch always NESTED, `signature()`/`nests_with` re-gated on `emit_cell_ids`, stripped YAMLs parse clean, and `TestCellIdsKnobs`/`TestCellIdsEncoding`/`TestSpatialSignature` cover the new surface. Local `pytest tests/test_config.py tests/test_grids.py tests/test_benchmark_objects.py tests/test_shardmap.py` → 443 passed / 1 skipped; ruff check + format clean on the touched files.

Confirmed-clean (no findings), noting for the record:
- **No in-tree silent-misread.** zagg has no read path that parses `dggs`/`indexing_scheme`/`cell_ids` back off a written store — the only construction site is `HealpixGrid._dggs_attrs`, now uniformly morton. (The raster path writes `cell_ids` with `attributes={}` and no dggs block — a separate product, untouched and out of scope.) moczarr is the reader; zagg is read-side agnostic, so there is nothing to "keep readable."
- **`spatial_signature()` keeping `indexing_scheme: "nested"` is coherent.** The HEALPix hierarchy *is* NESTED-numbered; morton is a bijective repack derived from it, so "nested" is the grid's structural identity independent of the stored coordinate name — and holding it stable across the flip is exactly what D19 ("additive, not breaking") wants for pre/post-flip products to share identity. `semantic_hash` is still a passthrough (`None`) today, not yet computed from this, so the coherence is forward-looking.
- **Unvalidated-config lingering-knob is harmless now.** The grid no longer reads `cell_ids_encoding` at all, so an unvalidated config carrying it changes nothing (morton is always the coordinate; the hatch is always NESTED). No construction-time guard needed; `validate_config` fires the pointed error.

Findings (all minor, all stale-reference cleanup the phase missed — none block):

**F9 (minor)** — `src/zagg/config.py:2066`, `get_emit_cell_ids` docstring still says the hatch writes `cell_ids` "values per `cell_ids_encoding`". That knob is retired *in this phase*; the hatch now always writes the NESTED encode. Reads as if the retired knob still selects the values — suggest "written in ADDITION to `morton` (NESTED `uint64`) …".

**F10 (minor)** — `docs/design/generalized_grids.md:295`, the grid table lists HealpixGrid coord arrays as `cell_ids (uint64), morton (int64)`. Two staleness bugs: `cell_ids` is no longer a default coordinate (hatch-only now), and `morton` is `uint64`, not `int64` — the exact `int64`→`uint64` correction this phase applied to `architecture.md` was missed here.

**F11 (minor)** — `docs/design/schema.md:70`, the "Output Columns" table lists `cell_ids | uint64 | HEALPix cell ID` as a default output column and `morton | int64`. Same post-D16 staleness: `cell_ids` is hatch-only and `morton` is `uint64`. The phase's "docs updated" scope touched `morton_arrow.md`/`architecture.md` but left these two design-doc tables describing the pre-flip schema.

espg:
🤖 *from Claude*

Phase-3 review findings folded — all three were stale-reference cleanups, one commit each:

- **F9** (`src/zagg/config.py` `get_emit_cell_ids` docstring): dropped the retired `cell_ids_encoding` reference; now reads that the hatch writes the `cell_ids` array holding the NESTED encode. → `192ffdc`
- **F10** (`docs/design/generalized_grids.md` coord table): HealpixGrid coord arrays updated to `morton (uint64)` as the only default coordinate, `cell_ids (uint64)` marked hatch-only (issue #304). → `2fdff09`
- **F11** (`docs/design/schema.md` Output Columns): `morton` now listed uint64 as the default coordinate; `cell_ids` (uint64) marked NESTED hatch-only. → `36acf3c`

Pre-push green: `ruff check`, `ruff format --check`, `pytest tests/test_config.py` (249 passed).

espg:
🤖 *from Claude*

Status: #312 merged (thanks!) — the `Blocked by #312` dependency is discharged: `origin/main` merged into this branch (e6f913e; #312 came in rebase-merged, conflicts resolved taking this branch's post-flip versions), the `blocked` label removed, and the espg-directed `resolution: mixed` third value (from the mortie PR #118 review) mirrored here as `RESOLUTION_MIXED` in `zagg.grids.morton` + constants-pin test (5d84f93 — originally noted on the #312 thread, which merged before it could ride there; the stray recreated branch was removed). Normative `mixed` text lands on the mortie spec page §4. This PR now carries the one remaining breaking store change (moczarr goldens coordination: espg/moczarr#11).

espg:
🤖 *from Claude*

**O10 supersession (espg final ruling, 2026-07-21 — via the mortie PR #118 review): the `resolution` declaration is deleted; kind is encoding-carried.** Commit ea35021:

- Removed the `resolution: "exact"` emission from `_dggs_attrs` and the `RESOLUTION_EXACT/POINT/MIXED` constants (#312 introduced the field + exact/point; 5d84f93 here mirrored `mixed`; all superseded). Point-vs-area kind rides the packed word's suffix encoding itself — area words are exact cells at every order incl. 29, point ids are points — so readers never consult a declaration and the attrs block carries none.
- The 29→24 clip left the spec: it is a viewer-side transient float64-safety cast (display-layer detail), not encoding semantics.
- The order-29 decimal parse-overlap tie-break (a parsed string always yields the area word) is pinned normatively on the mortie spec page §4 with a bit-level golden test (espg/mortie PR #118, commit 462ffb1).
- `docs/design/sparse_coverage.md` O10 entry rewritten to the final ruling (same commit); attrs tests updated to assert the field's ABSENCE.

Note: espg has additionally ruled an optional decimal **kind suffix `p`** for point ids (render/interchange only; paths never carry it; unmarked strings keep the area tie-break — fully backward compatible). zagg is unaffected — every string zagg renders or parses is an area word, unmarked and unchanged (verified by grep) — the change lands mortie-side (spec §2/§4 + a repr/parse implementation PR, linked from mortie#62/PR #118).

espg:
🤖 *from Claude (review)*

Fresh-context adversarial review of ea35021 ("fold review: kind is encoding-carried, resolution field removed", the O10 supersession) — narrow scope, this commit only. **No findings.**

Verified:
1. **Removal complete.** `RESOLUTION_EXACT/POINT/MIXED` gone from `morton.py` (definitions + `__all__`); no importer remains anywhere in the tree (`grep -rn RESOLUTION_ --include=*.py` empty). `healpix.py` drops the import and the `resolution` attr, replaced by a comment. No lingering `resolution discriminator` / `clip rule` / stale `RESOLUTION_*` references that contradict the ruling. The unrelated `resolution: cell|chunk` output-resolution keys, coverage-`encoding` discriminators, and multi-resolution/resolution-axis prose are correctly untouched. D16 tail (line ~593) rewritten to the final ruling; the D16 body clip sentence (line 585) is consistent with it (already frames the clip as browser-float64 safety = the viewer-side transient cast; only point ids clipped, area words untouched).
2. **O10 rewrite accurate to the ruling.** Area words exact at every order incl. 29 (nothing unrepresentable); point ids are points; no metadata field (readers key on the packed-word suffix region); 29→24 clip is a viewer-side transient cast, not spec semantics; order-29 decimal parse tie-break pinned normatively on the mortie spec page §4 (parsed string ⇒ area word), golden-pinned in `test_spec_page.py`. Companion identity/UUID ruling preserved. `morton_decimal` docstring (order-29 point parses back to the area word) is consistent with the pinned tie-break.
3. **moczarr coordination already corrected.** espg/moczarr#11 carries a correction comment (2026-07-20, "Correction to the attrs half of my coordination comment above") that retracts the earlier `resolution: exact|point` advertisement and states the encoding-carried ruling + the emitted attrs shape. No further correction warranted.
4. **Test coverage present for both surfaces.** `test_dggs_attrs_carry_no_kind_field` asserts `"resolution" not in _dggs_attrs()["dggs"]` (return value); `test_dggs_attrs_reach_written_store` asserts `"resolution" not in attrs["dggs"]` off the written MemoryStore group. The UUID-pin test correctly drops the removed constants and the three-value assertion.

Attrs block now: `dggs: {name, coordinate: "morton", refinement_level, spatial_dimension, ellipsoid, compression}` + two-entry `zarr_conventions` — coherent, no kind field.

tests/test_raster_pipeline.py
```diff
@@ -881,6 +875,63 @@ def test_flat_template_unchanged_by_refactor(self, tmp_path):
...
+            raster_group_spec(grid, cfg, 2).attributes,
+            raster_leaf_spec(grid, cfg, 2).members[grid.group_path].attributes,
+        ):
+            assert attrs == grid._dggs_attrs()
+            assert attrs["dggs"]["name"] == "morton"
+            assert attrs["dggs"]["coordinate"] == "morton"
+            names = [c["name"] for c in attrs["zarr_conventions"]]
+            assert names == ["dggs", "morton-dggs"]
+
+    def test_written_leaf_attrs_round_trip(self, tmp_path):
```
espg, line 922: 🤖 *from Claude (review)*

**F14 — nit (test parity):** `test_written_leaf_attrs_round_trip` round-trips the dggs attrs on the *written leaf* store, but nothing round-trips them on the *written flat* store. The flat path is exactly where `raster_group_spec` previously carried `attributes={}` and now carries `grid._dggs_attrs()`, landing on the group via `spec.to_zarr(store, grid.group_path, ...)` in `emit_raster_template`. `test_dggs_attrs_match_spatial_contract` only checks the *spec's* `.attributes`, not a written store. I verified by inspection that `to_zarr` does write the group attrs at `grid.group_path`, so this is low-risk — but the leaf got a written-store assertion and the flat write (the one that changed from `{}`) did not. Consider a one-liner: emit a flat template and assert the written group's `dggs.coordinate == "morton"`.

espg, line 922: 🤖 *from Claude (review)*

**F14 — nit (test parity):** `test_written_leaf_attrs_round_trip` round-trips the dggs attrs on the *written leaf* store, but nothing round-trips them on the *written flat* store. The flat path is exactly where `raster_group_spec` previously carried `attributes={}` and now carries `grid._dggs_attrs()`, landing on the group via `spec.to_zarr(store, grid.group_path, ...)` in `emit_raster_template`. `test_dggs_attrs_match_spatial_contract` only checks the *spec's* `.attributes`, not a written store. I verified by inspection that `to_zarr` does write the group attrs at `grid.group_path`, so this is low-risk — but the leaf got a written-store assertion and the flat write (the one that changed from `{}`) did not. Consider a one-liner: emit a flat template and assert the written group's `dggs.coordinate == "morton"`.

espg:
🤖 *from Claude (review)*

**F13 — minor (stale doc):** `docs/hive_layout.md:357` still describes the raster hive leaf as carrying **`cell_ids`** coords:

> each carrying leaf-local `time` (int64 microseconds, CF attrs) and `cell_ids` coords plus one `(T_leaf, cells_per_shard)` array per configured band

After this commit (03b3b51) the default raster leaf/flat store writes **`morton`** (packed u64 words) as the sole cell coordinate; `cell_ids` (NESTED) rides only the `emit_cell_ids` transition hatch — the same D16 contract the code docstrings and `docs/design/architecture.md` (lines 97/160) now state. This is the one raster-schema doc the flip left describing the old coord name. Suggest updating line 357 to `morton` (with the `cell_ids`-under-`emit_cell_ids` caveat) so `hive_layout.md`'s raster section matches the shipped schema.

(Not blocking; `docs/quickstart.md:134`'s `cell_ids:` is an *aggregation-config coordinate* declaration, out of scope for this raster phase — flagging only so it isn't conflated.)

espg:
🤖 *from Claude*

> **F13 — minor (stale doc):** `docs/hive_layout.md:357` still describes the raster hive leaf as carrying **`cell_ids`** coords [...] Suggest updating line 357 to `morton`

Fixed in 2f7130d. `docs/hive_layout.md` now describes the raster hive leaf as carrying `morton` (packed u64 words) as the **sole** cell coordinate, with `cell_ids` (NESTED) riding only the `emit_cell_ids` transition hatch (issue #304). Grepped the whole file for other stale `cell_ids` mentions — line 357 was the only one; `docs/quickstart.md:134` is an aggregation-config coordinate declaration and out of scope as you noted.

tests/test_raster_pipeline.py
espg: 🤖 *from Claude*

Fixed in 1c4e917. Added `test_written_flat_attrs_round_trip` in `TestRasterMortonCoordinate`, mirroring the leaf round-trip test: `emit_raster_template` to a `MemoryStore`, `zarr.open_group(path=grid.group_path)`, and assert the WRITTEN group attrs carry `dggs.coordinate == "morton"` (plus `cell_ids` absent). This pins the flat write — the path that changed from `attributes={}` to `grid._dggs_attrs()` — at the store level, matching the leaf.

espg: 🤖 *from Claude*

Fixed in 1c4e917. Added `test_written_flat_attrs_round_trip` in `TestRasterMortonCoordinate`, mirroring the leaf round-trip test: `emit_raster_template` to a `MemoryStore`, `zarr.open_group(path=grid.group_path)`, and assert the WRITTEN group attrs carry `dggs.coordinate == "morton"` (plus `cell_ids` absent). This pins the flat write — the path that changed from `attributes={}` to `grid._dggs_attrs()` — at the store level, matching the leaf.

espg:
> The emit_cell_ids hatch emits the legacy NESTED cell_ids array in addition to morton; the dggs attrs stay morton-declared. The hatch exists only until the gridlook morton path lands.


...we need to mark that depreciated for removal...

PR #315, opened 2026-07-20
Run-record write via worker invoke (D8): mode=stats transport (issue #313)

### LINKED ISSUE #313: Run-record write violates D8 orchestrator-no-write: route via worker invoke like the root MOC
🤖 *from Claude*

## The violation

D8 (`docs/design/sparse_coverage.md`): **"the orchestrator has no S3 write access"** — it's why the end-of-run root MOC is written by a fire-and-forget worker Event invoke from the dispatcher's completion list, never by the dispatcher. PR #302's run-level parquet write is dispatcher-side — the first orchestrator-originated PUT into a store — and the benchmark CI role is correctly denying it ([failing run](https://github.com/englacial/zagg/actions/runs/29767161250/job/88437075657): `403 AccessDenied` on `stats_{...}.parquet`, fail-open per design, followed by the object-count model rightly flagging the missing objects).

Espg-ratified direction: **fix the code, not the policy.** The orchestrator-no-write invariant is what makes the IAM story closed — all store writes flow through worker-role prefix grants, so no artifact we add (sidecars, sub-maps, overviews, product roots) ever needs a new grant. A CFN grant to the CI role was considered and rejected.

## The fix

Route the run-record write through the root-MOC pattern on the **Lambda dispatch paths** (spatial, raster, temporal): a fire-and-forget worker Event invoke (a `mode: "stats"`-style handler branch) performs the parquet PUT under the worker role. **Local dispatch paths keep writing directly** (user credentials; no invariant). Fail-open semantics preserved end-to-end; failure rows still included; timestamp-first naming (D20) unchanged; the object-count model needs no change.

Payload note for the implementer: rows × thousands of shards can exceed the 256 KB async-invoke payload — pass rows inline when small, else pass a pointer (the per-shard result envelopes already sit under the status prefix, which the worker role can read; the invoke can carry run metadata + the status-prefix location and let the worker assemble the rows). Implementer's judgment, stated in the PR body.

## Not in scope

The same benchmark run's per-shard 11-vs-12 sidecar gap is under separate triage (leading hypothesis: deployed fleet function code predates PR #302, so workers don't write sidecars yet) — do not paper over it here.

espg:
Closes #313

The issue #297 run-level parquet was the first orchestrator-originated PUT into a store, violating D8's orchestrator-no-write invariant (`docs/design/sparse_coverage.md`) — the benchmark CI's invoke-only role correctly 403'd it. Fix the code, not the policy (espg-ratified on the issue): on the **Lambda dispatch paths** the write moves behind a fire-and-forget worker Event invoke — the root-MOC (`mode: "coverage"`) pattern — as a new `mode: "stats"` handler branch; **local dispatch paths keep writing directly** (user credentials, no invariant). Fail-open at every step, failure rows preserved, timestamp-first D20 naming unchanged, object-count model untouched.

## Transport choice (stated per the issue's payload note)

**Async Event invoke per the D8 standing rule** (espg refinement, 2026-07-22): the run-parquet worker invoke is fire-and-forget `InvocationType="Event"`, exactly like the root-MOC `mode:"coverage"` leg — the dispatcher never blocks on it, fail-open, retries-0 (the dispatch clients run `retries: {max_attempts: 0}` and `_dispatch_run_stats` adds no retry loop; the seam test pins `InvocationType == "Event"`). Built this way from the start via the coverage pattern; nothing to flip.

**Fire → verify → re-fire hardening** (espg addendum; commit `013654c`): failure rows exist only in the dispatcher's run report — a dropped retries-0 Event invoke would lose them permanently (success rows stay rebuildable from the leaf sidecars). After firing, the dispatcher does a bounded READ-ONLY existence poll on the pinned D20 key (HEAD every 2 s, ≤ 20 s — reads were never forbidden; the dispatcher already polls result objects) and re-fires the Event invoke exactly once if the parquet hasn't appeared, then gives up fail-open with a logged notice. Worst case ≤ 2 × 20 s of end-of-run wall, zero writes (D8 intact); a non-positive window is the single-fire escape hatch (and keeps unit harnesses hermetic). Tests pin: object-present → one invoke; dropped-first → exactly one identical re-fire that lands; both-dropped → exactly two invokes, announced path retained, nothing written.

**Hybrid, size-gated, behind one seam** (`runner._dispatch_run_stats`):

- **Inline** when the serialized rows fit the async budget: the event carries `rows` directly. Threshold: rows JSON ≤ 200 KB (the existing `_ASYNC_PAYLOAD_CAP_BYTES = 250 KB` Event-invoke gate minus ~50 KB headroom for the rest of the event — creds block, paths). At ~0.5–1 KB/row that is roughly 200–400 shards; typical benchmark runs ride inline.
- **Pointer** above the threshold (async runs): the event carries `rows_from` — the run's `.status/` prefix — and the worker (whose role can read the store side) LISTs the issue-#151 mirrored result envelopes and assembles the success rows from their `body["stats"]` records. **Every dispatcher-built row rides inline next to the pointer** — failure rows (a timed-out shard mirrored no envelope) and stale-worker fallback success rows (a 200 envelope with no record; review fold `c5bce65`) cannot be assembled worker-side. Two acknowledged losses in pointer mode, stated rather than papered over: dispatcher `retries` counts are not in the envelopes (pointer-assembled success rows carry `retries: null`), and a shard whose envelope landed but whose poll timed out dispatcher-side may appear as both an inline failure row and an assembled success row (documented on the seam; the run record is best-effort telemetry, not load-bearing truth).
- **Sync-invocation runs** (`invocation="sync"`, legacy) have no status prefix; oversized row sets there log a fail-open skip. Sync is the legacy small-run path, so this corner is theoretical.

The dispatcher pins the D20 object key at dispatch (`timestamp` rides in the event), so `summary["run_stats_path"]` still names the real object — its semantics on lambda paths become "dispatched target, best-effort" (the invoke is fire-and-forget), mirroring the root MOC's semantics; local paths keep path-on-success/None-on-failure.

## Phases

- [x] Phase 1 — worker side: `mode: "stats"` handler branch (inline + pointer transports, pinned-key write, fail-open), `telemetry.write_run_parquet(timestamp=...)`, `telemetry.rows_from_status` (worker-side envelope assembly); tests against local stores
- [x] Phase 2 — dispatcher side: `_dispatch_run_stats` transport seam (size-gated inline/pointer, pinned D20 key, fail-open), wired into the spatial and raster Lambda paths (local paths untouched); seam unit tests + spatial/raster harness integration (raster dispatches the failure rows BEFORE its all-failed raise, pinned by test)

## Temporal note

The temporal Lambda path (`_run_lambda_events`) writes **no run record today** — the #297 wiring never covered it (its fan-out unit is an event, not a shard; its `store_path` is a tabular object, not a store root; its rows would need an event-keyed schema). Nothing to move behind the invoke there. When a temporal run record lands (follow-up), it inherits this PR's `mode: "stats"` transport as-is.

## Testing

- Phase 2: seam units (inline event shape + announced path == pinned key; pointer keeps only failure rows inline; oversized sync run skips fail-open; invoke failure fail-open; empty rows no invoke), spatial summary carries the dispatched path, raster all-failed run still dispatches its failure rows.
- Phase 1: inline rows written at the dispatcher-pinned key (exact D20 name asserted), pointer assembly from mirrored envelopes (stale-worker envelope skipped; inline failure rows merged), empty-rows no-op, fail-open 500 path.
- Full suite green locally; pre-existing environment-only items (lambda-build test on this machine's pip, `N818` in `registry.py`) unchanged.

## Questions for review

1. **Per-shard 11-vs-12 sidecar gap**: not touched here per the issue's scope note (separate triage, fleet-code-lag hypothesis). This PR doesn't interact with it — the sidecar writes are per-shard worker writes and unchanged; only the run-parquet transport moves.
2. **Pointer-mode `retries: null` + the double-row corner** (envelope landed after dispatcher poll deadline): accepted as stated above; if either matters for cost forensics, the alternative is the dispatcher chunking inline rows across several `mode:"stats"` invokes — say the word and it's a small change on the seam.

### DIFF
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -2209,6 +2209,7 @@ def _dispatch_run_stats(
     result_prefix=None,
     output_creds_event=None,
     summary=None,
+    inline_rows=None,
 ) -> str | None:
     """Fire-and-forget worker-invoke run-record write (issue #313, D8).
 
@@ -2218,9 +2219,12 @@ def _dispatch_run_stats(
     Event invoke exactly like the root ``coverage.moc``. Transport is
     size-gated (the seam the PR #315 body describes): rows inline under
     :data:`_RUN_STATS_INLINE_CAP_BYTES`; above it the run's async status
-    prefix is sent as a pointer (``rows_from``) with only the failure rows
-    inline (no envelope exists to assemble them from); an oversized SYNC run
-    (no prefix) logs a fail-open skip. The D20 object key is pinned here
+    prefix is sent as a pointer (``rows_from``) with only the rows the worker
+    cannot reassemble from the mirrored envelopes riding inline. Those are
+    given by ``inline_rows`` when the caller knows them (the spatial path:
+    failure rows PLUS stale-worker fallback-success rows — issue #313); absent,
+    the failure rows alone are kept inline (the raster path builds no fallback
+    rows). An oversized SYNC run (no prefix) logs a fail-open skip. The D20 object key is pinned here
     (``timestamp`` rides the event), so the returned/announced
     ``run_stats_path`` names the real object — best-effort: the invoke is
     fire-and-forget, mirroring the root MOC's semantics. Fail-open
@@ -2247,14 +2251,18 @@ def _dispatch_run_stats(
         if len(json.dumps(rows)) <= _RUN_STATS_INLINE_CAP_BYTES:
             event["rows"] = rows
         elif result_prefix is not None:
-            failures = [r for r in rows if not r.get("success")]
-            if len(json.dumps(failures)) > _RUN_STATS_INLINE_CAP_BYTES:
+            inline = (
+                inline_rows
+                if inline_rows is not None
+                else [r for r in rows if not r.get("success")]
+            )
+            if len(json.dumps(inline)) > _RUN_STATS_INLINE_CAP_BYTES:
                 logger.warning(
-                    f"run stats failure rows alone exceed the inline budget "
-                    f"({len(failures)} rows); skipping run record (fail-open, issue #313)"
+                    f"run stats inline rows alone exceed the inline budget "
+                    f"({len(inline)} rows); skipping run record (fail-open, issue #313)"
                 )
                 return None
-            event["rows"] = failures
+            event["rows"] = inline
             event["rows_from"] = result_prefix
         else:
             logger.warning(
@@ -2277,7 +2285,7 @@ def _dispatch_run_stats(
         return None
 
 
-def _lambda_result_rows(results, *, run_id=None) -> list:
+def _lambda_result_rows(results, *, run_id=None) -> tuple[list, list]:
     """Run-parquet rows from the lambda dispatch's per-cell result dicts.
 
     Success rows come from the envelope-ridden worker record
@@ -2286,13 +2294,25 @@ def _lambda_result_rows(results, *, run_id=None) -> list:
     a failure row (error, duration until failure, retry count). ``run_id``
     stamps the dispatcher-built fallback/failure rows; envelope records
     already carry it (the worker copies it from the event).
+
+    Returns ``(rows, inline_rows)``. ``inline_rows`` is the subset that MUST
+    ride inline on the pointer transport (issue #313): the pointer path has
+    the worker re-derive success rows from the mirrored envelopes via
+    :func:`~zagg.telemetry.rows_from_status`, which reads only ``body["stats"]``
+    — so any row NOT backed by an envelope record (failure rows, and the
+    stale-worker fallback-success rows built here) would be dropped worker-side
+    and has to be sent inline alongside the pointer.
     """
     from zagg.telemetry import build_record, failure_record, flatten_record
 
     rows = []
+    inline_rows = []
     for r in results:
         body = r.get("body") or {}
         record = body.get("stats")
+        # Envelope-backed rows are re-derivable worker-side (rows_from_status);
+        # everything else the pointer path can't rebuild, so it rides inline.
+        rideable = record is not None
         if record is None:
             if r.get("status_code") == 200 and not r.get("error"):
                 # Stale deployed worker (no record in the envelope): derive one
@@ -2309,8 +2329,11 @@ def _lambda_result_rows(results, *, run_id=None) -> list:
                     duration_s=r.get("lambda_duration") or r.get("wall_time"),
                     run_id=run_id,
                 )
-        rows.append(flatten_record(record, retries=r.get("retries")))
-    return rows
+        row = flatten_record(record, retries=r.get("retries"))
+        rows.append(row)
+        if not rideable:
+            inline_rows.append(row)
+    return rows, inline_rows
 
 
 def _resolve_source_credentials(config) -> dict:
@@ -3370,15 +3393,17 @@ def _run_lambda(
     # failure rows from the dispatch results the RunReport accumulated. The
     # PUT itself rides a fire-and-forget worker invoke — the dispatcher may
     # hold an invoke-only role with no S3 write access.
+    stats_rows, stats_inline_rows = _lambda_result_rows(report.results, run_id=run_id)
     _dispatch_run_stats(
         state["lambda_client"],
         function_name,
         store_path,
-        _lambda_result_rows(report.results, run_id=run_id),
+        stats_rows,
         run_id=run_id,
         result_prefix=result_prefix,
         output_creds_event=output_creds_event,
         summary=summary,
+        inline_rows=stats_inline_rows,
     )
     logger.info(
         f"Done: {report.cells_with_data} cells, {report.total_obs:,} obs, {report.cells_error} errors, {wall_time:.1f}s"
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -1402,11 +1402,15 @@ class RasterStrategy:
             )
             for key, error, body in stats_failures
         ]
-        run_stats_path = _write_run_stats(
+        # D8 worker-invoke transport (issue #313): mirrors the spatial path.
+        run_stats_path = _dispatch_run_stats(
+            client,
+            function_name,
             store_path,
             rows,
             run_id=run_id,
-            store_kwargs=_output_store_kwargs(output_creds_event, region),
+            result_prefix=result_prefix,
+            output_creds_event=output_creds_event,
         )
         if cells and errors == len(cells):
             raise RuntimeError(f"all {errors} raster shard(s) failed; last error: {last_error}")
@@ -2192,6 +2196,84 @@ def _write_run_stats(store_path, rows, *, run_id, store_kwargs, summary=None) ->
         return path
     except Exception as e:
         logger.warning(f"run stats parquet write failed (fail-open, issue #297): {e}")
+        return None
+
+
+def _dispatch_run_stats(
+    lambda_client,
+    function_name,
+    store_path,
+    rows,
+    *,
+    run_id,
+    result_prefix=None,
+    output_creds_event=None,
+    summary=None,
+) -> str | None:
+    """Fire-and-forget worker-invoke run-record write (issue #313, D8).
+
+    The Lambda-path twin of :func:`_write_run_stats`: the orchestrator has no
+    S3 write access (D8 — the invariant that keeps every store write behind
+    worker-role prefix grants), so the parquet PUT rides a ``mode="stats"``
+    Event invoke exactly like the root ``coverage.moc``. Transport is
+    size-gated (the seam the PR #315 body describes): rows inline under
+    :data:`_RUN_STATS_INLINE_CAP_BYTES`; above it the run's async status
+    prefix is sent as a pointer (``rows_from``) with only the failure rows
+    inline (no envelope exists to assemble them from); an oversized SYNC run
+    (no prefix) logs a fail-open skip. The D20 object key is pinned here
+    (``timestamp`` rides the event), so the returned/announced
+    ``run_stats_path`` names the real object — best-effort: the invoke is
+    fire-and-forget, mirroring the root MOC's semantics. Fail-open
+    everywhere: telemetry must never fail a run whose data landed.
+    """
+    from datetime import datetime, timezone
+
+    from zagg.telemetry import run_parquet_key
+
+    if summary is not None:
+        summary["run_stats_path"] = None
+    if not rows:
+        return None
+    try:
+        timestamp = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%SZ")
+        event = {
+            "mode": "stats",
+            "store_path": store_path,
+            "run_id": run_id,
+            "timestamp": timestamp,
+        }
+        if output_creds_event is not None:
+            event["output_credentials"] = output_creds_event
+        if len(json.dumps(rows)) <= _RUN_STATS_INLINE_CAP_BYTES:
+            event["rows"] = rows
+        elif result_prefix is not None:
+            failures = [r for r in rows if not r.get("success")]
+            if len(json.dumps(failures)) > _RUN_STATS_INLINE_CAP_BYTES:
+                logger.warning(
+                    f"run stats failure rows alone exceed the inline budget "
+                    f"({len(failures)} rows); skipping run record (fail-open, issue #313)"
+                )
+                return None
+            event["rows"] = failures
+            event["rows_from"] = result_prefix
+        else:
+            logger.warning(
+                f"run stats rows ({len(rows)}) exceed the async payload budget and this "
+                f"sync run has no status prefix to point at; skipping (fail-open, issue #313)"
+            )
+            return None
+        lambda_client.invoke(
+            FunctionName=function_name,
+            InvocationType="Event",
+            Payload=json.dumps(event),
+        )
+        path = f"{store_path.rstrip('/')}/{run_parquet_key(run_id, timestamp)}"
+        if summary is not None:
+            summary["run_stats_path"] = path
+        logger.info(f"Dispatched run stats write ({len(rows)} rows, fire-and-forget): {path}")
+        return path
+    except Exception as e:
+        logger.warning(f"run stats dispatch failed (fail-open, issue #313): {e}")
         return None
 
 
@@ -3283,14 +3365,19 @@ def _run_lambda(
     }
     if profile:
         summary["worker_phase_max"] = worker_phase_max
-    # Run-level stats parquet (issue #297 phase 3): success rows straight off
-    # the async result envelopes (no second S3 listing), failure rows from the
-    # dispatch results the RunReport accumulated.
-    _write_run_stats(
+    # Run-level stats parquet (issue #297 phase 3; D8 worker-invoke transport,
+    # issue #313): success rows straight off the async result envelopes,
+    # failure rows from the dispatch results the RunReport accumulated. The
+    # PUT itself rides a fire-and-forget worker invoke — the dispatcher may
+    # hold an invoke-only role with no S3 write access.
+    _dispatch_run_stats(
+        state["lambda_client"],
+        function_name,
         store_path,
         _lambda_result_rows(report.results, run_id=run_id),
         run_id=run_id,
-        store_kwargs=_output_store_kwargs(output_creds_event, region),
+        result_prefix=result_prefix,
+        output_creds_event=output_creds_event,
         summary=summary,
     )
     logger.info(
@@ -3428,6 +3515,12 @@ _MANIFEST_CHECK_JOIN_TIMEOUT_S = 30.0
 # Lambda's raw RequestEntityTooLargeException does; the realistic trigger is a
 # large strict-AOI ``aoi_payload`` (issue #101).
 _ASYNC_PAYLOAD_CAP_BYTES = 250 * 1024
+
+#: Inline-rows budget for the mode="stats" run-record invoke (issue #313):
+#: the Event cap above minus ~50 KB headroom for the rest of the event
+#: (credentials block, paths). Above this the dispatcher sends the status
+#: prefix pointer instead (failure rows stay inline — they have no envelope).
+_RUN_STATS_INLINE_CAP_BYTES = 200 * 1024
 
 # The result poller owns its retrying at the loop level (a fetch every
 # _ASYNC_POLL_INTERVAL_S until the deadline), so its store gets a short
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -1411,6 +1411,7 @@ class RasterStrategy:
             run_id=run_id,
             result_prefix=result_prefix,
             output_creds_event=output_creds_event,
+            store_kwargs=_output_store_kwargs(output_creds_event, region),
         )
         if cells and errors == len(cells):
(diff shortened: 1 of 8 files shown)


src/zagg/telemetry.py
```diff
@@ -364,7 +374,7 @@ def write_run_parquet(
 
     if not rows:
         raise ValueError("write_run_parquet requires at least one row")
-    key = run_parquet_key(run_id)
+    key = run_parquet_key(run_id, timestamp)
```
espg, line 401: 🤖 *from Claude (review)*

**Event-sourced `timestamp` (and `run_id`) flow unvalidated into an object KEY.** This diff makes `timestamp` a new caller-supplied input that lands directly in `run_parquet_key` → `f"stats_{ts}_{run_id}.parquet"`, with no charset check. A `/` or `..` in either component escapes the store root: on `write_run_parquet` local dispatch, `os.path.join(tmp, key)` then targets a non-existent subdir and the temp write raises; on the store PUT the composed key traverses out of the intended prefix. Previously `timestamp` was always the server-generated `%Y%m%dT%H%M%SZ` form, so this is a new injection surface, not the pre-existing `run_id` one.

This file already has the precedent for exactly this — `sidecar_name_for_leaf` calls `validate_label(stem)` (below, ~line 480) specifically so an embedded `/` path escape RAISES rather than passing through as a malformed key. Mirror it here: validate `run_id` against the frozen opaque charset (`[0-9A-Za-z-]`) and `timestamp` against the D20 timestamp grammar before composing the key, so a hostile/buggy dispatcher value fails loud instead of writing outside the store root under the worker role.

src/zagg/telemetry.py
```diff
@@ -375,6 +385,49 @@ def write_run_parquet(
...
+    import logging
+
+    import obstore
+
+    from zagg.store import open_object_store
+
+    logger = logging.getLogger(__name__)
+    store = open_object_store(status_prefix, **(store_kwargs or {}))
+    rows = []
+    for batch in obstore.list(store):
```

**Recursive `obstore.list` diverges from the repo's prefix-listing convention and will descend into subdirs.** Both existing prefix walks in the codebase use the delimiter form — `obstore.list_with_delimiter` (`coverage.py:246`, `hive.py:249`) — which returns only the objects immediately under the prefix and hands common-prefixes back for deliberate recursion. `obstore.list` here is fully recursive: it flattens every object at any depth under `{store}.status/{run_id}/`. Today the layout is flat (`{run_id}/{shard_label}.json`, `runner.py:1173/1289`) so it happens to work, but any nested key the async channel ever gains under that prefix would silently get parsed as a candidate envelope. Switch to `obstore.list_with_delimiter(store)` for the immediate-children semantics you actually want and to match the two precedents.

src/zagg/telemetry.py
```diff
@@ -375,6 +385,49 @@ def write_run_parquet(
...
+        for meta in batch:
+            key = meta["path"]
+            if not key.endswith(".json"):
+                continue
+            try:
+                envelope = _json.loads(bytes(obstore.get(store, key).bytes()))
```

**Serial per-envelope GETs are the pointer transport's scaling risk on a 900 s worker.** This branch is reached precisely for large runs (the ones too big for the inline budget — hundreds to thousands of shards), and it issues one blocking `obstore.get` per envelope with no concurrency. At a few thousand shards × a small-JSON S3 round-trip each, this can consume a meaningful slice of the 900 s wall — and if it overruns, the worker times out and the run record is lost for exactly the largest runs, the ones you most want recorded. Rows accumulating in memory is fine (~KB each); the serial round-trip count is the pressure. Consider bounded-concurrency reads (obstore async get / gather with a cap) or at least state the shard-count ceiling this is expected to hold under, so the fail-open ceiling is a known number rather than an implicit one.

tests/test_lambda_handler.py
```diff
@@ -1444,6 +1444,91 @@ def test_flat_event_still_consolidates(self, handler_mod, monkeypatch, tmp_path)
...
+
+        # Mirror two per-shard result envelopes the way _write_result does.
+        prefix = tmp_path / "out.zarr.status" / "runid2"
+        prefix.mkdir(parents=True)
+        for shard in (11, 12):
+            rec = build_record(shard_key=shard, metadata={"total_obs": 1, "duration_s": 2.0})
+            envelope = {"statusCode": 200, "body": json.dumps({"stats": rec})}
+            (prefix / f"{shard}.json").write_text(json.dumps(envelope))
+        # A stale-worker envelope without a record is skipped, not fatal.
+        (prefix / "13.json").write_text(json.dumps({"statusCode": 200, "body": "{}"}))
```
espg, line 1496: 🤖 *from Claude (review)*

**Test pins the no-record skip but not the "unparsable objects are skipped with a warning" claim.** `rows_from_status`' docstring promises two skip paths — an envelope without a `stats` record (stale worker) AND an object that fails to parse — but the suite only exercises the first (this `13.json` has a valid-JSON envelope with an empty `body`). The truly-unparsable path (malformed-JSON object, or a non-`.json` object under the prefix) — the one that hits the `logger.warning(...)` except branch — is asserted-in-docstring but not pinned. Add a bad-bytes `.json` (e.g. `not json`) under the prefix and assert it's skipped without failing the write, so a future refactor of the try/except can't silently turn a skip into a fatal.

src/zagg/runner.py
```diff
@@ -2195,6 +2199,84 @@ def _write_run_stats(store_path, rows, *, run_id, store_kwargs, summary=None) ->
...
+            "store_path": store_path,
+            "run_id": run_id,
+            "timestamp": timestamp,
+        }
+        if output_creds_event is not None:
+            event["output_credentials"] = output_creds_event
+        if len(json.dumps(rows)) <= _RUN_STATS_INLINE_CAP_BYTES:
+            event["rows"] = rows
+        elif result_prefix is not None:
+            failures = [r for r in rows if not r.get("success")]
```

Pointer mode silently drops the stale-worker **success** rows — an inline↔pointer asymmetry that is not in the PR body’s two acknowledged losses, and it is exactly the fleet-lag case issue #313 targets.

`_lambda_result_rows` builds a *success* row via `build_record(body)` for a 200 whose envelope carries no `stats` block (a worker predating #302 — the stale-fleet path). Here the pointer branch keeps only `failures = [r for r in rows if not r.get("success")]` inline, so that row is excluded; worker-side `rows_from_status` then skips the same envelope (`body.get("stats")` is `None`). Net: inline mode has a row for that shard, pointer mode has none. And pointer mode only triggers on *large* runs — precisely where a lagging fleet under-counts the very shards whose fleet-lag motivates this PR.

Options: (a) also ride these dispatcher-built fallback-success rows inline next to the failures — they are already assembled orchestrator-side, no envelope needed, same as the failure rows; or (b) document it as a third accepted loss alongside `retries: null` and the double-row corner. Note the raster path builds no such fallback (it only rows `body["stats"]`), so this is spatial-only.

src/zagg/runner.py
```diff
@@ -2195,6 +2199,84 @@ def _write_run_stats(store_path, rows, *, run_id, store_kwargs, summary=None) ->
...
+            if len(json.dumps(failures)) > _RUN_STATS_INLINE_CAP_BYTES:
```

Untested branch: the failure-rows-alone-exceed-budget guard (this `if` down through its `return None`) is the one `_dispatch_run_stats` path `TestDispatchRunStats` never exercises. The suite covers inline, pointer-with-failures-fitting, oversized-sync (no prefix), invoke-failure, and empty rows — but not the case where the *failure* rows themselves overflow the inline cap with a `result_prefix` present. A one-liner pins it: tiny `_RUN_STATS_INLINE_CAP_BYTES`, pass `result_prefix`, assert `path is None` and `client.events == []`.

src/zagg/runner.py
```diff
@@ -2195,6 +2199,84 @@ def _write_run_stats(store_path, rows, *, run_id, store_kwargs, summary=None) ->
...
+        timestamp = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%SZ")
+        event = {
+            "mode": "stats",
+            "store_path": store_path,
+            "run_id": run_id,
+            "timestamp": timestamp,
+        }
+        if output_creds_event is not None:
+            event["output_credentials"] = output_creds_event
+        if len(json.dumps(rows)) <= _RUN_STATS_INLINE_CAP_BYTES:
```
espg, line 2253: 🤖 *from Claude (review)*

Minor / likely leave-by-design: in the pointer branch this `json.dumps(rows)` serializes the **entire** row set (potentially multi-MB, thousands of rows) solely to reject inline, after which only `failures` and the final `event` are serialized and the full-row string is discarded — so oversized runs pay a redundant full serialization. It is one-time and orchestrator-side (not on a worker’s 900 s budget), so probably fine; if you would rather not pay it, gate on a cheap size estimate (e.g. `len(rows)` heuristic) or reuse a single `rows` serialization for the inline path. Flagging per the PR’s transport-cost note.

src/zagg/telemetry.py
espg: 🤖 *from Claude*

Fixed in 17eb50c (ruff-format follow-up 4ca8dab). `run_parquet_key` now validates both components before composing the key, mirroring `sidecar_key`'s `validate_label` precedent: `timestamp` against the exact D20 grammar `^[0-9]{8}T[0-9]{6}Z$` and `run_id` against the opaque charset `^[0-9A-Za-z-]{1,64}$`, each raising `ValueError` on violation. A `/` or `..` in either now fails loud instead of writing outside the store root. Added `test_run_parquet_key_rejects_traversal` (traversal attempts on both components raise).

src/zagg/telemetry.py

Fixed in 1258e60. Switched to `obstore.list_with_delimiter(store)` and iterate `listing["objects"]` — the immediate-children semantics the `coverage.py:246` / `hive.py:249` prefix walks use, so a future nested key under the status prefix can no longer get flattened into a candidate envelope.

src/zagg/telemetry.py

Fixed in e7ec4bc. The per-envelope GETs now fan out over a bounded `ThreadPoolExecutor(max_workers=min(16, len(keys)))` (the repo's `ThreadPoolExecutor`/`pool.map` pattern from `runner.py`), so the large-run branch no longer serializes thousands of S3 round-trips against the 900 s budget. Rows still accumulate in memory (~KB each); the skip/fail-open semantics are unchanged (each envelope maps to a record-or-`None`, `None` filtered out). Docstring now states why this branch is pooled.

tests/test_lambda_handler.py
espg: 🤖 *from Claude*

Fixed in 3235a6b. The pointer-mode test now drops a truly-unparsable `14.json` (`not json` bytes — hits the `except`/`logger.warning` branch) and a non-`.json` `manifest.txt` (filtered before any GET) under the prefix, and still asserts the write succeeds at the same row count (3, shard keys `{2, 11, 12}`). A future refactor of the try/except that turned a skip into a fatal would now fail this test.

src/zagg/runner.py

Fixed in c5bce65 (option (a) — ride the dispatcher-built fallback-success rows inline next to the failures). `_lambda_result_rows` now returns `(rows, inline_rows)`, where `inline_rows` is exactly the subset the pointer path cannot rebuild worker-side: any row NOT backed by an envelope `body["stats"]` record — i.e. failure rows PLUS the stale-worker fallback-success rows built here. `_dispatch_run_stats` gained an `inline_rows=None` kwarg; the pointer branch sends `inline_rows` when provided and otherwise keeps the old `not success` failure-filter (the raster caller at runner.py:1406 builds no fallback rows, so its default is correct). The inline-budget guard applies to the `inline` set unchanged. Tests: `TestRunStatsRows` now unpacks and asserts `inline_rows` shard keys are `[2, 3]` (fallback-success shard 2 + failure shard 3, NOT the envelope-backed shard 1); new `test_pointer_transport_rides_inline_rows` asserts a `success=True` fallback row (shard 999) rides inline despite the success flag.

src/zagg/runner.py

Fixed in 575cc9c. Added `test_pointer_inline_rows_over_budget_skips_fail_open` to `TestDispatchRunStats`: tiny `_RUN_STATS_INLINE_CAP_BYTES` (8), a `result_prefix` present, and `_rows(n_ok=5, n_fail=3)` so the inline-required rows overflow the cap on their own. Asserts `path is None` and `client.events == []` (fail-open skip, no invoke), plus `summary["run_stats_path"] is None`. That pins the `if len(json.dumps(inline)) > _RUN_STATS_INLINE_CAP_BYTES: return None` branch that the suite previously never exercised.

src/zagg/runner.py
espg: 🤖 *from Claude*

Leaving by design. The gate has to reject an oversized inline payload for correctness — Lambda async invoke caps the request body at 256 KB, so `_dispatch_run_stats` must know the EXACT serialized size of `rows` before choosing inline vs. pointer. A `len(rows)` heuristic can under-estimate (row widths vary with error strings, retry counts, null vs. populated fields) and would risk sending an over-budget inline payload that the invoke then bounces — trading a cheap serialization for a dropped run record. And the cost is genuinely one-time and orchestrator-side (not on any worker's 900 s budget), fired once per run after all shards land. So the safe read is: pay the single exact serialization at the seam. No code change; flagging as accepted per your leave-by-design note.

src/zagg/runner.py
```diff
@@ -2195,7 +2200,152 @@ def _write_run_stats(store_path, rows, *, run_id, store_kwargs, summary=None) ->
...
+        store = open_object_store(store_path, **(store_kwargs or {}))
+    except Exception as e:
+        logger.warning(f"run stats verify skipped (cannot open store read-only): {e}")
+        return False
+    deadline = time.time() + window_s
+    while True:
+        try:
+            obstore.head(store, key)
+            return True
+        except Exception:
```
espg, line 2345: 🤖 *from Claude (review)*

**Persistent read failure is indistinguishable from a genuine worker drop, and costs the full 2×window every run.** Every per-HEAD exception is swallowed here (`except Exception: pass`), so a dispatcher that can *open* the store but gets 403/absent on the pinned key produces only the generic "still absent after the re-fire" notice — a reader can't tell "worker never wrote it" (drop, the case re-fire targets) from "I structurally cannot see it" (a permission/visibility gap the re-fire can never fix). This matters because the run parquet lands at the store **root** (`stats_<ts>.parquet`), whereas the proven read path (`_output_store_kwargs` → async result poller) only demonstrates read access under the result/`.status/` prefix. If the dispatcher's read grant is prefix-scoped, the root-key HEAD 403s on *every* run → the poll burns the full window, re-fires pointlessly, burns it again (40s of end-of-run wall by default), forever, with no log explaining why. Fix: log the first head exception once (distinguishably from a 404/absent), and/or treat a persistent auth error as unrecoverable so the poll short-circuits instead of grinding the whole window twice.

tests/test_runner.py
```diff
@@ -3475,6 +3475,229 @@ def test_agg_reads_handoff_from_config_on_local(self, monkeypatch, atl06_config)
...
+
+    def _verify_dispatch(self, monkeypatch, tmp_path, drop_first):
+        from zagg import runner
+
+        monkeypatch.setattr(runner, "_RUN_STATS_VERIFY_WINDOW_S", 0.5)
+        monkeypatch.setattr(runner, "_RUN_STATS_VERIFY_INTERVAL_S", 0.01)
+        root = tmp_path / "out"
+        root.mkdir()
+        client = self._VerifyClient(root, drop_first=drop_first)
+        path = runner._dispatch_run_stats(client, "fn", str(root), self._rows(), run_id="rid")
```
espg: 🤖 *from Claude (review)*

**The re-fire tests never pass a `summary`, so the "path set once on attempt 1, not clobbered on re-fire" invariant is unpinned.** `_verify_dispatch` calls `_dispatch_run_stats(client, "fn", str(root), self._rows(), run_id="rid")` with no `summary=` — but the summary write is the one piece of loop state deliberately guarded by `if attempt == 1` inside the two-iteration fire loop. A regression that moved the `summary["run_stats_path"] = path` set out from under that guard (re-setting it on the re-fire, or leaving it None on the dropped-first path) would pass every current test. Add a `summary={}` to at least the `drop_first=1` case and assert `summary["run_stats_path"] == path` after the re-fire — that pins the "set once, on the first fire, survives the re-fire" contract the diff's structure depends on.

src/zagg/runner.py
```diff
@@ -2195,7 +2200,152 @@ def _write_run_stats(store_path, rows, *, run_id, store_kwargs, summary=None) ->
...
+                f"sync run has no status prefix to point at; skipping (fail-open, issue #313)"
+            )
+            return None
+        key = run_parquet_key(run_id, timestamp)
+        path = f"{store_path.rstrip('/')}/{key}"
+        payload = json.dumps(event)
+        # Fire -> verify -> re-fire (the async-per-D8 hardening): the Event
+        # invoke is retries-0 and failure rows exist only in this run report,
+        # so verify the parquet landed (read-only HEAD poll on the pinned
+        # key) and re-fire ONCE if not. Bounded — the dispatcher never
```

**"the dispatcher never meaningfully blocks" undersells the new hot-path cost — this is not just a 2×window worst case.** With verification on, *every successful* lambda run now blocks at its tail until the async worker actually lands the parquet: Lambda cold-start + envelope LIST/assemble + PUT, then the dispatcher's next ≤2s HEAD tick catches it. Previously this leg returned instantly (pure fire-and-forget). So the typical added wall isn't zero — it's one worker round-trip on the critical path, on every run, not only the drop case. The PR body's "≤2×20s" framing covers the pathological tail, but this inline comment's "never meaningfully blocks" reads as if the common case is free. Suggest softening the comment to name the typical cost (one worker-write round-trip added to the run tail) so the tradeoff is visible where the code lives, not only in the PR body.

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed in 12b81e1. The per-HEAD poll now splits its except: `(FileNotFoundError, NotFoundError)` stays the silent "not there yet" path the re-fire targets, while any *other* exception (e.g. a prefix-scoped read grant that 403s HEADs at the store root) logs once per poll — `run stats verify cannot read {key}: {e} — treating as absent` — so a structural read gap is visible in the log instead of hiding behind the generic still-absent notice. Kept it simple: warn-once, still treat as absent, no auth classification beyond not-FileNotFoundError (obstore's error taxonomy varies by backend). Left the short-circuit-on-persistent-auth option out — that would need reliable auth-vs-transient classification I don't have here; the warning is enough to surface the gap for @espg.

tests/test_runner.py
espg: 🤖 *from Claude*

Fixed in b0dded6. `_verify_dispatch` now threads a `summary=` through to `_dispatch_run_stats`, and both re-fire tests pass one: `test_dropped_first_invoke_refires_once` asserts `summary["run_stats_path"] == path` after the re-fire (the path is set once on attempt 1 and survives, not clobbered), and `test_both_invokes_dropped_gives_up_fail_open` asserts it's already set even when verification never confirmed the object. A regression moving the set out from under `if attempt == 1` — re-setting on the re-fire or leaving it None on the dropped-first path — now fails.

src/zagg/runner.py

Fixed in b7f99b5. Dropped "the dispatcher never meaningfully blocks" and named the real cost in the comment: every successful run now waits at its tail for one async worker round-trip (cold start + envelope assemble + PUT), caught by the next <=2s HEAD tick — seconds of added wall on the common path where fire-and-forget returned instantly — with the up-to-2x-window worst case kept for the drop tail. The tradeoff is now visible where the code lives, not only in the PR body.

Issue #308, opened 2026-07-20
Explicit-list re-template append (D15) is unreachable through the manifest overwrite guard


espg:
🤖 *from Claude*

## The conflict

D15 (`docs/design/sparse_coverage.md`, temporal block) declares the explicit-range-list schedule's append story as `append_policy: "re-template"` — appending a window outside the declared list rewrites the manifest ("a rare, single-writer, template-time operation"). `build_manifest` encodes exactly that (`src/zagg/hive.py:197-199`).

But the re-template path is unreachable as coded:

- The windows list rides the **frozen** `temporal` key (`_FROZEN_MANIFEST_KEYS`, `hive.py:309`), so a grown list is a frozen-key mismatch.
- `validate_manifest(..., overwrite=True)` refuses a frozen-key change whenever the digit tree already has base-shaped children (`hive.py:244-257` — the D2 old-order-masquerade guard).
- A store being re-templated *by definition* has children (that's why it's being appended to).

So: `overwrite=False` → "does not match this run" refusal; `overwrite=True` → "digit tree already has shard data" refusal. The documented D15 append path for explicit-list stores cannot be executed through `ensure_manifest`. (Generative schedules are unaffected — their appends never touch the manifest, which is why this went unnoticed.)

## Options

1. **Monotonic-growth carve-out (recommended)**: a `temporal.windows` change where the old list is a strict subset of the new one — and every *other* frozen key is equal — is safe by construction: existing leaves remain legal under the grown schedule, so the D2 masquerade hazard doesn't apply. `validate_manifest` special-cases it (with or without `overwrite`); shrinking or otherwise editing the list still refuses. Keeps D15's single-writer re-template semantics, one narrow relaxation.
2. **Documented manual override**: delete/replace the manifest object by hand before the append run. Works today, but violates the write-once spirit and leaves a window where the store has no manifest — worst option.
3. **Drop re-template**: let explicit-list stores accept out-of-schedule windows and have the §7 sweep update the declared list. Rejected lean — it moves accruing state into the manifest by the back door, exactly what D15 was designed to prevent.

Under D19 product roots the same logic applies per product tree, unchanged.

Two sibling edges from the same trace are implementation notes on #299 (spec-compat on append across the `/2`→`/3` bump; `path_grouping` frozen-key normalization) — they're writer-side requirements, not code conflicts, so they ride that issue rather than this one.

PR #316, opened 2026-07-20
Named product roots + semantic-core hash (issue #299)

### LINKED ISSUE #299: Template-hash leaf naming: multi-product hive store (breaking, pre-1.0)

espg:
Closes #299. Implements the **final** form of the issue — named product roots + semantic-core hash — per the governing record: the [Plan v2 comment](https://github.com/englacial/zagg/issues/299#issuecomment-5025685368) and [Scope revision 3](https://github.com/englacial/zagg/issues/299#issuecomment-5024857268) (which supersede the original leaf-basename plan), plus D19 rev 2, D24, and §3 in the merged `docs/design/sparse_coverage.md`. **Additive, not breaking**: bare single-product stores stay byte-identical and fully valid; there is no old-store error path.

## Phases

- [x] Phase 1 — **semantic-core canonicalization + hash** (`src/zagg/semantics.py`): the output-defining subset only (aggregation block minus the `handoff` carrier; `data_source` semantics minus `reader`/`driver`/`read_plan`/`anonymous`; grid **type + indexing scheme** — never orders/chunking/worker/store knobs) → sorted-key compact JSON → **full sha256** (git-style: full digest compared; 12-hex `semantic_fingerprint` for display, 48-bit rationale from the [thread](https://github.com/englacial/zagg/issues/299#issuecomment-5017334704)). Formalizes the #89 `spatial_signature`/`output_field_signature` seam. §8.3 property tests: syntactic edits never change the hash, packaging-knob edits never change it, semantic edits always do (`tests/test_semantics.py`).
- [x] Phase 2 — **named product roots (additive)**: product-name grammar validation (URL-safe lowercase `[a-z0-9_-]+`, base-component exclusion `-?[1-6]`), `product_root()` path construction, root-form discrimination by content (manifest ⇒ bare store; name-shaped prefixes ⇒ product directory), product listing.
- [x] Phase 3 — **manifest additions**: `semantic_hash` as a frozen key (absent-tolerant for pre-#299 stores), `path_grouping` frozen-key comparison with absent⇒1 normalization on both sides, the canonical core written as `{name}/aggregation.yaml` (derived convenience — the manifest hash is truth, D9 cache class), append-path spec-compat (writer honors the existing manifest's declared spec; the PR #307 seam is already spec-keyed).
- [x] Phase 4 — **`has_run` dedup API**: computed leaf path + `semantic_hash` + sidecar catalog identity (`granules_sha256`); catalog-grown shard reports **"stale"**, never "hit"; O11 per-array content hash is the verifier when present.
- [x] Phase 5 (verification) — **`template_hash` → `semantic_hash` rename**: the deferral from the PR #309 review thread was already discharged by the PR #309 fold itself ([thread](https://github.com/englacial/zagg/pull/309#discussion_r3616135096), fix `13b54e9` — `estimate_cost_usd(semantic_hash=...)`); a full-tree grep on this branch confirms **zero** `template_hash` references remain. Nothing to rename; recorded here for provenance.

## Implementation notes

- **`.github/` touch (flagged prominently, per the PR #302 precedent)**: `.github/scripts/bench_objects.py` hive object-count model gains the fail-open `aggregation.yaml` root object (metadata ceiling +1; measured-side classification includes `hive.AGGREGATION_CORE_NAME`). No workflow files touched.
- **Lambda worker parity (deployment/aws — NOT touched, per §1)**: the local runner writes `aggregation.yaml` (rides `ensure_manifest(config=...)`) and stamps `semantic_hash` into stats records; the Lambda worker's handler (`deployment/aws/lambda_handler.py`) does neither yet — its sidecars keep `semantic_hash: null` (which `has_run` conservatively reports as "stale", never a false hit). Wiring the handler needs an issue that names it, per the repo's infra rule — flagged under Questions.
- `has_run` lives in a new small module `src/zagg/dedup.py` (and the hasher in `src/zagg/semantics.py`) rather than growing `hive.py`, which already exceeds the ~1000-line discussion threshold.
- The spec-compat seam: `ensure_manifest` now RETURNS the manifest in effect (the accepted existing one on append), the runner threads its `spec` into sidecar naming (`write_sidecar(spec=...)` — the #307 seam), and `hive.append_spec` documents the rule for the future `/3` writer. For `/1`–`/2` stores the emitted names are byte-identical.

## Interactions

- **Independent of PR #314** (#304 attrs/cell_ids flip): based off `main`; the overlap is small (grid signature surfaces) — whichever merges second integrates by merge. (#312/#305 has since MERGED; #314 is unblocked and carries the writer flip.)
- **#308 re-template conflict**: untouched here; the frozen-key changes are confined to the two new keys with explicit absent-side normalization, so `validate_manifest`'s existing behavior on pre-#299 manifests is unchanged (tests pin appends to old manifests still pass the precheck).
- **D24 cross-order guard**: flagged rather than forced — see Questions.
- moczarr churn from this issue is **root-level product discovery only**; no golden-vector regeneration ([issue record](https://github.com/englacial/zagg/issues/299#issuecomment-5025685368)).

## Testing

Per-phase test files noted in the checklist; full `pytest` run locally before each push (`tests/test_lambda_build.py` excluded — pre-existing environmental failure on this machine, local pip resolves against Python 3.10; also flagged on #312/#314). Pre-existing `N818` in `src/zagg/registry.py` on `main` excluded from ruff runs, flagged not fixed (§4).

## Questions for review

1. *(folded during review)* The product-name grammar gained the espg-ruled **192-char length cap** (mirrored from the mortie spec page ruling) — `PRODUCT_NAME_MAX` in `hive.py`.
1. *(resolved — espg-ruled 2026-07-21)* **`pipeline.type` is IN the semantic core** (commit `d689348`): output-defining — a temporal/event engine over an identical aggregation block is a different product; absent normalizes to `spatial` on both sides (same discipline as `path_grouping` absent⇒1), so existing configs hash stably; golden pin re-pinned; D19 doc parenthetical updated in the same commit.
2. **Rectilinear grids** — resolved conservatively during review fold (F1): rect cores now fold in `crs`/`resolution`/`bounds` (D24's resolution-axis exclusion is a HEALPix/morton composability argument that doesn't extend to rect; over-discriminating is safe, colliding is not). Relax if D24-style polymorphism ever lands for rect.
3. *(confirmed — espg, 2026-07-21)* **D24 cross-order guard** stays with the #304-family writer work; not in this PR.

### DIFF
--- a/docs/design/sparse_coverage.md
+++ b/docs/design/sparse_coverage.md
@@ -643,8 +643,13 @@ rollup leaves all leaf reads intact.
   **`semantic_hash`** verifies it — sha256 over the canonicalized
   **output-defining subset only**: the `aggregation` block (functions +
   params + dtypes + fills + ragged kinds), the `data_source` semantics
-  (dataset/product, groups, coordinates, variables, filters), and the
-  grid *type + indexing scheme*. Excluded as packaging: cell order (a
+  (dataset/product, groups, coordinates, variables, filters), the
+  grid *type + indexing scheme*, and the **pipeline type** (`spatial` |
+  `temporal` | `event`, absent ⇒ `spatial` — espg-ruled on the PR #316
+  review, 2026-07-21: a temporal engine over the same aggregation block is
+  a different product; the original list omitted it only because the
+  temporal path wasn't in frame, and pre-merge was the free moment to add
+  it — no store carries a hash yet). Excluded as packaging: cell order (a
   resolution axis — D24), parent/shard order, `chunk_inner`/`sharded`,
   worker size, streaming mode (merge-vs-spill lands `np.isclose` and
   shares one store, with the actual mode recorded per-run), and read
--- a/src/zagg/dedup.py
+++ b/src/zagg/dedup.py
@@ -1,0 +1,139 @@
+"""Per-shard dedup status: ``has_run`` (issue #299 phase 4, D19).
+
+Answers "has this exact product already produced this shard?" from durable
+store state, for the estimate/skip flows (#298) to consume. Three signals, in
+strictly increasing cost:
+
+1. **The commit stamp** (D4): a leaf whose root attrs lack the stamp — or no
+   leaf at all — is a plain ``"miss"`` (debris is invisible).
+2. **The D20 stats sidecar**: the recorded ``semantic_hash`` (intended
+   identity, D19) and ``granules_sha256`` (catalog identity — the output is
+   ``f(template, shard, catalog snapshot)``, and ATL03 is a living
+   collection). A stamped leaf whose sidecar is missing, records a different
+   or absent ``semantic_hash``, or hashes a different granule set is
+   ``"stale"`` — present but not provably this product over this catalog;
+   a catalog-grown shard is stale, **never** a hit.
+3. **The O11 content hashes**, surfaced (not recomputed) when the sidecar
+   carries them: the *verifier* — "intended identical" (semantic hash) vs
+   "actually byte-identical" (per-array decoded-value hashes) — for callers
+   that go on to compare two stores.
+
+Statuses are conservative by construction: every ambiguity degrades toward
+recompute (``stale``/``miss``), never toward a false ``hit`` — a wrong "skip"
+silently ships wrong data; a wrong "recompute" costs one shard's work.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from zagg.hive import read_commit, read_manifest, shard_leaf_path
+from zagg.store import open_store
+from zagg.telemetry import granules_sha256, read_sidecar
+
+logger = logging.getLogger(__name__)
+
+#: ``has_run`` statuses: complete + verified for THIS product and catalog
+#: snapshot / present but unverifiable-or-outdated / not (completely) there.
+STATUSES = ("hit", "stale", "miss")
+
+
+def shard_status(
+    store_root: str,
+    shard_key,
+    *,
+    semantic_hash: str,
+    granule_ids=None,
+    window: str | None = None,
+    spec: str | None = None,
+    **store_kwargs,
+) -> dict:
+    """Dedup status of ONE (shard, window) leaf; see :func:`has_run`.
+
+    ``granule_ids`` is the shard's CURRENT catalog snapshot in the same id
+    space the sidecars record (resolved granule URLs on the aggregation
+    path, STAC item ids/datetimes for raster — cf.
+    :func:`zagg.telemetry.granules_sha256`); ``None`` skips the catalog
+    check (identity match alone then gates the hit).
+    """
+    leaf = shard_leaf_path(store_root, int(shard_key), window=window)
+    stamp = read_commit(open_store(leaf, **store_kwargs))
+    if stamp is None:
+        return {"status": "miss"}
+    sidecar = read_sidecar(leaf, spec=spec, **store_kwargs)
+    if sidecar is None:
+        # Stamped but unverifiable (pre-#297 leaf, or a lost fail-open PUT):
+        # never a hit — the leaf may be any product/catalog vintage.
+        return {"status": "stale", "reason": "no stats sidecar"}
+    detail: dict = {
+        "semantic_hash_match": sidecar.get("semantic_hash") == semantic_hash,
+        "catalog_match": None,
+    }
+    if content := sidecar.get("content_hashes"):
+        # O11 verifier, surfaced when recorded (never recomputed here).
+        detail["content_hashes"] = content
+    if granule_ids is not None:
+        detail["catalog_match"] = sidecar.get("granules_sha256") == granules_sha256(granule_ids)
+    if not detail["semantic_hash_match"]:
+        return {"status": "stale", "reason": "semantic_hash mismatch or unrecorded", **detail}
+    if detail["catalog_match"] is False:
+        return {"status": "stale", "reason": "catalog grown/changed", **detail}
+    return {"status": "hit", **detail}
+
+
+def has_run(
+    store_root: str,
+    config,
+    shards,
+    *,
+    window: str | None = None,
+    spec: str | None = None,
+    **store_kwargs,
+) -> dict:
+    """Per-shard dedup status for a prospective run (issue #299 phase 4).
+
+    Parameters
+    ----------
+    store_root : str
+        The PRODUCT root (apply :func:`zagg.hive.effective_store_root` /
+        :func:`zagg.hive.product_root` first for multi-product stores).
+    config : PipelineConfig
+        The prospective run's config; its ``semantic_hash`` is the identity
+        compared against each sidecar.
+    shards : mapping or iterable
+        ``{shard_key: granule_ids}`` (current catalog snapshot per shard —
+        the same id space the sidecars record) or a bare iterable of shard
+        keys (catalog check skipped).
+    window : str, optional
+        Window label for windowed stores (one status per (shard, window)).
+    spec : str, optional
+        The store's naming spec for sidecar keys; default: read once from
+        the manifest (``None`` on a manifest-less root = the ``/1`` legacy
+        names).
+
+    Returns
+    -------
+    dict
+        ``{int(shard_key): {"status": "hit"|"stale"|"miss", ...detail}}``.
+    """
+    from zagg.semantics import semantic_hash as _semantic_hash
+
+    want = _semantic_hash(config)
+    if spec is None:
+        spec = (read_manifest(store_root, **store_kwargs) or {}).get("spec")
+    items = shards.items() if hasattr(shards, "items") else ((k, None) for k in shards)
+    return {
+        int(key): shard_status(
+            store_root,
+            key,
+            semantic_hash=want,
+            granule_ids=ids,
+            window=window,
+            spec=spec,
+            **store_kwargs,
+        )
+        for key, ids in items
+    }
+
+
+__all__ = ["STATUSES", "has_run", "shard_status"]
--- a/src/zagg/dedup.py
+++ b/src/zagg/dedup.py
@@ -55,6 +55,17 @@ def shard_status(
     path, STAC item ids/datetimes for raster — cf.
     :func:`zagg.telemetry.granules_sha256`); ``None`` skips the catalog
     check (identity match alone then gates the hit).
+
+    .. warning::
+       The catalog check is an EXACT ``granules_sha256`` comparison, so
+       ``granule_ids`` must be the same id space the writer recorded — a
+       caller that supplies a different space (bare short-names vs resolved
+       URLs, sorted vs dispatch order) makes EVERY stamped shard report
+       ``stale`` on ``catalog_match`` rather than a genuine catalog change.
+       A ``"catalog grown/changed"`` stale carries both
+       ``granules_sha256_recorded`` and ``granules_sha256_current`` so a
+       mis-spaced caller can see the mismatch is systematic (every shard
+       differs), not a real snapshot drift.
     """
     leaf = shard_leaf_path(store_root, int(shard_key), window=window)
     stamp = read_commit(open_store(leaf, **store_kwargs))
@@ -73,11 +84,21 @@ def shard_status(
         # O11 verifier, surfaced when recorded (never recomputed here).
         detail["content_hashes"] = content
     if granule_ids is not None:
-        detail["catalog_match"] = sidecar.get("granules_sha256") == granules_sha256(granule_ids)
+        recorded = sidecar.get("granules_sha256")
+        current = granules_sha256(granule_ids)
+        detail["catalog_match"] = recorded == current
     if not detail["semantic_hash_match"]:
         return {"status": "stale", "reason": "semantic_hash mismatch or unrecorded", **detail}
     if detail["catalog_match"] is False:
-        return {"status": "stale", "reason": "catalog grown/changed", **detail}
+        # Surface both digests so a systematic id-space mismatch (every shard
+        # differs) is distinguishable from a genuine per-shard catalog change.
+        return {
+            "status": "stale",
+            "reason": "catalog grown/changed",
+            "granules_sha256_recorded": recorded,
+            "granules_sha256_current": current,
+            **detail,
+        }
     return {"status": "hit", **detail}
 
 
--- a/src/zagg/hive.py
+++ b/src/zagg/hive.py
@@ -441,8 +441,10 @@ def write_semantic_core(store_root: str, config, **store_kwargs) -> None:
     A derived convenience next to the manifest (D9 cache class): sorted-key,
     deterministic YAML of the output-defining subset, so product names stay
     human-inspectable without a registry. FAIL-OPEN — the manifest's
-    ``semantic_hash`` is the truth and the sweep can regenerate this; a
-    failed PUT must not fail the run.
+    ``semantic_hash`` is the truth, and the core is regenerable straight from
+    the config (rewriting it is a single :func:`write_semantic_core` call), so
+    a failed PUT must not fail the run. The D22 pyramid sweep is the intended
+    owner of the rewrite once it lands; there is no regenerator today.
     """
     import obstore
     import yaml
--- a/src/zagg/hive.py
+++ b/src/zagg/hive.py
@@ -100,27 +100,44 @@ ROOT_COVERAGE_NAME = "coverage.moc"
 AGGREGATION_CORE_NAME = "aggregation.yaml"
 
 #: Product-name grammar (D19; normative on the mortie spec page §6.5):
-#: URL-safe lowercase alphanumerics plus ``-``/``_``. The base-component
-#: exclusion (``-?[1-6]``, :func:`_is_base_component`) is checked separately —
-#: a name shaped like a hive base component would make the walker's child
-#: classification ambiguous at a multi-product store root.
+#: URL-safe lowercase alphanumerics plus ``-``/``_``, at most
+#: :data:`PRODUCT_NAME_MAX` chars. The base-component exclusion (``-?[1-6]``,
+#: :func:`_is_base_component`) is checked separately — a name shaped like a
+#: hive base component would make the walker's child classification ambiguous
+#: at a multi-product store root.
 _PRODUCT_NAME_RE = re.compile(r"^[a-z0-9_-]+$")
+
+#: Product-name length cap (D19; espg ruling mirrored from the mortie spec page
+#: §6.5). Derivation: a POSIX filename component is 255 bytes; the immutable-
+#: provenance decoration reserves 13 (``+`` + a 12-hex digest) ⇒ a 242-byte
+#: hard ceiling. 192 sits 50 under that, and leaves the S3 total-key and
+#: PATH_MAX budgets comfortable (the name also nests under the store prefix and
+#: the digit tree). The grammar is single-byte per char, so char count == byte
+#: count here.
+PRODUCT_NAME_MAX = 192
 
 
 def validate_product_name(name: str) -> str:
     """Validate a D19 product name; returns it.
 
     Grammar (normative on the mortie spec page §6.5): one or more of
-    ``[a-z0-9_-]``, and never matching the morton base-component grammar
-    ``-?[1-6]`` — the root-form discrimination
-    (:func:`classify_store_root`) depends on names and digit components
-    being disjoint. Names are URL-safe by construction (they appear in
-    gridlook deep-links and web paths).
+    ``[a-z0-9_-]``, at most :data:`PRODUCT_NAME_MAX` characters, and never
+    matching the morton base-component grammar ``-?[1-6]`` — the root-form
+    discrimination (:func:`classify_store_root`) depends on names and digit
+    components being disjoint. Names are URL-safe by construction (they appear
+    in gridlook deep-links and web paths).
     """
     if not isinstance(name, str) or not _PRODUCT_NAME_RE.match(name):
         raise ValueError(
             f"product name {name!r} does not match the D19 grammar ([a-z0-9_-]+, "
             f"lowercase; normative on the mortie spec page)"
+        )
+    if len(name) > PRODUCT_NAME_MAX:
+        raise ValueError(
+            f"product name {name!r} is {len(name)} chars; the D19 grammar caps names "
+            f"at {PRODUCT_NAME_MAX} (normative on the mortie spec page §6.5 — a POSIX "
+            f"255-byte filename component less the 13-byte immutable-provenance "
+            f"decoration)"
         )
     if _is_base_component(name):
         raise ValueError(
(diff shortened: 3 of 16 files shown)


src/zagg/semantics.py
```diff
@@ -0,0 +1,127 @@
...
+    grid_type = grid_cfg.get("type", "healpix")
+    core: dict = {
+        "aggregation": _without(config.aggregation, AGGREGATION_PACKAGING_KEYS),
+        "data_source": _without(config.data_source, DATA_SOURCE_PACKAGING_KEYS),
+        "grid": {"type": grid_type},
+    }
+    if grid_type == "healpix":
+        # The one indexing scheme zagg writes (the morton store convention
+        # rides D16 attrs; the underlying cell tiling is HEALPix NESTED).
+        core["grid"]["indexing_scheme"] = "nested"
```
espg: @/tmp/f1.md

src/zagg/semantics.py
```diff
@@ -0,0 +1,127 @@
...
+AGGREGATION_PACKAGING_KEYS = ("handoff",)
+
+#: Display length of :func:`semantic_fingerprint` (12 hex = 48 bits; the
+#: birthday bound puts same-store collision odds around 1e-8 at 1e4 products
+#: — recorded rationale on the issue #299 thread).
+FINGERPRINT_HEX = 12
+
+
+def _without(mapping: dict, keys: tuple[str, ...]) -> dict:
+    return {k: v for k, v in (mapping or {}).items() if k not in keys and v is not None}
```
espg: @/tmp/f2.md

src/zagg/semantics.py
```diff
@@ -0,0 +1,127 @@
...
+    Deterministic given the config's semantics: two configs differing only in
+    packaging knobs (orders, chunking, worker size, read machinery, carrier)
+    map to the same core. The returned structure is JSON-serializable plain
+    data (the YAML loader guarantees it).
+    """
+    grid_cfg = (config.output or {}).get("grid", {}) or {}
+    grid_type = grid_cfg.get("type", "healpix")
+    core: dict = {
+        "aggregation": _without(config.aggregation, AGGREGATION_PACKAGING_KEYS),
+        "data_source": _without(config.data_source, DATA_SOURCE_PACKAGING_KEYS),
```
espg, line 120: @/tmp/f3.md

tests/test_semantics.py
```diff
@@ -0,0 +1,149 @@
...
+            "  variables: {h_li: /p}\n"
+            "  groups: [gt1l]\n"
+            "  reader: h5coro\n"
+        )
+        ca = PipelineConfig(**a)
+        cb = PipelineConfig(**b)
+        assert canonical_semantic_json(ca) == canonical_semantic_json(cb)
+        assert semantic_hash(ca) == semantic_hash(cb)
+
+    def test_packaging_knobs_never_change_hash(self):
```
espg, line 80: @/tmp/f4.md

src/zagg/semantics.py
espg: 🤖 *from Claude*

Fixed in b283239. Non-healpix grids now fold the spatially-defining params — `crs`, `resolution`, `bounds` — into the grid sub-dict when present, so two rect products differing only in CRS or resolution no longer collide on `semantic_hash`. HEALPix is left exactly as-is (`type` + `indexing_scheme: "nested"`; the hardcode is correct — the config `indexing_scheme` key is descriptive-only per `validate_config`). D24 resolution-axis exclusion is a HEALPix/morton composability argument that does not extend to rect; over-discriminating is safe, colliding is not. `test_grid_family_is_semantic` updated to assert the params flow through, and a new `test_rect_resolution_changes_hash` pins that changing resolution (or CRS) moves the hash. The PR body Question 2 stands for espg to relax if D24-style polymorphism ever lands for rect.

src/zagg/semantics.py
espg: 🤖 *from Claude*

Fixed in 491c1af. Null pruning is now recursive: a new `_prune_nulls(obj)` drops every `None`-valued key from every dict at any depth, applied to the whole core (`_without` no longer carries its own top-level-only null filter). Lists are recursed but never pruned by value — list entries are positional, so a `None` element is content, not an absent key. This closes the §8.3 hole where a nested YAML explicit-null (e.g. `quality_filter.value:`) hashed differently from the key being absent. Docstring updated; `test_null_packaging_values_drop_out` is joined by `test_nested_null_values_drop_out`, which pins that `quality_filter.value: None` hashes equal to `value` absent, while `value: 0` still differs.

src/zagg/semantics.py
espg: 🤖 *from Claude*

Declining, by design. List order (groups/filters) is preserved as declared. Canonicalizing away list order would require knowing per-schema which lists are sets vs sequences, and over-discrimination here is safe: the worst case is that a reordered-but-equivalent config recomputes a product it could have reused — never a false dedup hit that serves stale/wrong values. Reordering `groups` or `filters` is also a deliberate edit, not a syntactic one like whitespace or key order. Leaving this standing for espg if he wants set-semantics for specific keys later.

tests/test_semantics.py
espg: 🤖 *from Claude*

Both added in 676eabc. (a) `test_packaging_knobs_never_change_hash` now sets `cfg.output["grid"]["emit_cell_ids"] = True` and asserts the hash is unchanged — the core reads only the grid type/scheme so it drops out trivially, but the exclusion is now pinned. (b) `test_golden_hash_pin` builds a fully-inline `PipelineConfig` (fixed dicts, no `default_config`) and hard-codes the full 64-hex digest `c17f7523ffb21e7b5a33d092f4ab12eaf805b1f88117f592132805458f354f86`, computed after the F1/F2 fold so it is the post-fold value. The comment explains it catches accidental canonicalization drift (key ordering, null pruning, packaging-key sets, JSON separators).

src/zagg/hive.py
```diff
@@ -297,9 +430,51 @@ def ensure_manifest(store_root: str, manifest: dict, *, overwrite: bool = False,
...
+        obstore.put(
+            open_object_store(store_root, **store_kwargs),
+            AGGREGATION_CORE_NAME,
+            payload.encode(),
+        )
+    except Exception as e:
+        logger.warning(f"{AGGREGATION_CORE_NAME} write failed (fail-open, D9 cache class): {e}")
+
+
+def append_spec(existing: dict | None, manifest: dict) -> str:
```

F5 (minor) — `append_spec` in src/zagg/hive.py is uncalled dead code: the runner already realizes its logic via `ensure_manifest`'s returned manifest (§4 no speculative abstraction); only its own unit test exercises it.

*(comment body repaired — the original post's file-reference didn't interpolate; the fold reply below addresses it.)*

src/zagg/hive.py
```diff
@@ -297,9 +430,51 @@ def ensure_manifest(store_root: str, manifest: dict, *, overwrite: bool = False,
     if existing is not None and not overwrite:
```
espg, line 447: 🤖 *from Claude (review)*

F6 (minor) — a pre-#299 / lambda-primary store never self-upgrades to carry `semantic_hash`: append returns `existing` with no re-PUT, so the D19 identity guard stays permanently off for such stores. Deliberate absent-side tolerance, but the durability limit is worth recording — is an opportunistic backfill in scope?

*(comment body repaired — the original post's file-reference didn't interpolate; the fold reply below addresses it.)*

src/zagg/hive.py
```diff
@@ -297,9 +430,51 @@ def ensure_manifest(store_root: str, manifest: dict, *, overwrite: bool = False,
...
     return manifest
 
 
+def write_semantic_core(store_root: str, config, **store_kwargs) -> None:
+    """PUT the canonical semantic core as ``aggregation.yaml`` (D19).
+
+    A derived convenience next to the manifest (D9 cache class): sorted-key,
+    deterministic YAML of the output-defining subset, so product names stay
+    human-inspectable without a registry. FAIL-OPEN — the manifest's
+    ``semantic_hash`` is the truth and the sweep can regenerate this; a
```

F7 (nit) — `aggregation.yaml` is written only on a store's first manifest PUT; no §7-sweep regenerator exists in the tree, so the docstring's "the sweep can regenerate this" is aspirational.

*(comment body repaired — the original post's file-reference didn't interpolate; the fold reply below addresses it.)*

src/zagg/dedup.py
```diff
@@ -0,0 +1,139 @@
...
+#: snapshot / present but unverifiable-or-outdated / not (completely) there.
+STATUSES = ("hit", "stale", "miss")
+
+
+def shard_status(
+    store_root: str,
+    shard_key,
+    *,
+    semantic_hash: str,
+    granule_ids=None,
```
espg, line 46: 🤖 *from Claude (review)*

F8 (minor) — `has_run`'s `granule_ids` id-space is a silent all-`stale` trap: a #298 consumer passing raw ids instead of resolved URLs gets permanent recompute with no error (degrade-safe but only guarded by docstring).

*(comment body repaired — the original post's file-reference didn't interpolate; the fold reply below addresses it.)*

tests/test_dedup.py
```diff
@@ -0,0 +1,155 @@
...
+        # degrades to recompute, never to a false hit.
+        cfg = _cfg()
+        _write_leaf(str(tmp_path), cfg, sidecar_hash="absent")
+        status = shard_status(
+            str(tmp_path), WORD, semantic_hash=semantic_hash(cfg), granule_ids=GRANULES
+        )
+        assert status["status"] == "stale"
+
+
+class TestHasRun:
```
espg, line 127: 🤖 *from Claude (review)*

F9 (nit) — coverage gaps: windowed `has_run` untested; no product-root end-to-end through `agg()` (only `effective_store_root` unit-tested).

*(comment body repaired — the original post's file-reference didn't interpolate; the fold reply below addresses it.)*

espg:
@f10.txt

src/zagg/hive.py

Deleted in 72b2ce1 — `append_spec` + `test_append_spec_honors_existing` are gone. It had no callers, was not in `__all__`, and the runner already realizes the same result via the manifest `ensure_manifest` returns: the returned-manifest IS the seam. When the `/3` writer lands, reinstating a one-line `(existing or manifest)["spec"]` helper (or reading `.spec` off the returned manifest inline) is trivial, so this is not a load-bearing abstraction to keep speculatively (§4, no speculative abstraction).

src/zagg/hive.py
espg: 🤖 *from Claude*

Leaving by design (no code change). The append precheck is read-only by contract: `validate_manifest` reads the existing manifest and refuses on a frozen-key mismatch, but never mutates it — a resume must not rewrite frozen keys, and `semantic_hash` is a frozen key. An opportunistic self-upgrade of a pre-#299 store (backfilling the missing hash on read) would be a write on a path deliberately kept write-free, and it would race concurrent resumes with no compare-and-swap. The natural home for a backfill is the D22/§7 sweep, which already owns authoritative rewrites of derived state. Standing for espg if he wants that sweep to backfill `semantic_hash`.

src/zagg/hive.py

Fixed in bedca72. The `write_semantic_core` docstring no longer claims a regenerator exists: it now says the core is regenerable straight from the config (rewriting is a single `write_semantic_core` call) and names the D22 pyramid sweep as the intended owner of that rewrite *once it lands* — "there is no regenerator today." The `semantic_hash` in the manifest remains the truth; the YAML core stays a fail-open D9 cache.

src/zagg/dedup.py
espg: 🤖 *from Claude*

Strengthened in 86d7370. Two parts:

1. **Docstring warning** on `shard_status` — the catalog check is an EXACT `granules_sha256` comparison, so a caller passing the wrong id space (bare short-names vs resolved URLs, sorted vs dispatch order) makes EVERY stamped shard report `stale` on `catalog_match`, not a genuine catalog change.
2. **Visible mismatch** — a `"catalog grown/changed"` stale now carries `granules_sha256_recorded` and `granules_sha256_current` in the detail, so a mis-spaced caller can see the mismatch is systematic (identical delta on every shard) rather than real snapshot drift. `test_catalog_growth_is_stale_never_hit` now pins that the two digests are present and differ.

tests/test_dedup.py
espg: 🤖 *from Claude*

Added the windowed unit test in fa81615 — `TestHasRun::test_windowed_leaf_status`: it writes a `window="2025"` leaf (stamp + legacy `spec=None` sidecar), asserts `zagg.telemetry.sidecar_key` keys the legacy windowed name `stats_2025.json`, then checks `has_run(..., window="2025")` is a `hit` while `has_run(..., window="2024")` is a `miss` (windows are disjoint leaves).

Declining the product-root end-to-end-through-`agg()` case as covered-by-unit-tests: the product-root arithmetic is already pinned in `TestProductRoots` (`effective_store_root`, `product_root`, `shard_leaf_path` under a product root all byte-identical to bare), and `has_run` takes the resolved product root as `store_root` — wiring a full `agg()` run through Lambda/local dispatch here would be a heavy integration harness for a path whose two halves are each unit-tested. Standing for espg if he wants the full e2e anyway.

espg:
🤖 *from Claude*

Re F10 (`deployment/aws/invoke_lambda.py` bypasses `effective_store_root`): leaving by design, no code change here. That file is under `deployment/aws/`, which the project conventions forbid touching in a fold that the issue does not name (§1). The right vehicle is the future follow-up already flagged in the PR body — the one that names the Lambda handler for `semantic_hash`/product-root parity — where the handler and its invoke shim get the product-root wiring together. Standing for espg to route onto that issue.

espg:
🤖 *from Claude*

**espg ruling folded: product-name length cap of 192 (8b571e0).** Mirrored from the mortie PR #118 review. `validate_product_name` now rejects names longer than 192 chars with a message citing the mortie spec §6.5; added a `PRODUCT_NAME_MAX = 192` constant carrying the derivation (POSIX 255-byte filename component − 13 reserved for the `+`+12-hex immutable-provenance decoration ⇒ 242 hard ceiling; 192 sits 50 under, S3 total-key/PATH_MAX comfortable). The grammar comment/docstring note the cap, and `TestProductRoots::test_name_length_cap` pins the boundary (192 accepts, 193 rejects).

espg:
🤖 *from Claude*

Question 1 resolved per the espg ruling (2026-07-21): **`pipeline.type` joined the semantic core** — commit d689348. `spatial`|`temporal`|`event` with absent ⇒ `spatial` normalization on both sides (the `path_grouping` absent⇒1 discipline), so every existing config hashes stably while a temporal engine over the same aggregation block becomes a distinct product, as ruled. Canonicalization tests extended (absent == explicit default; temporal ≠ spatial), the F4 golden pin re-pinned to the new digest, and the D19 parenthetical in `docs/design/sparse_coverage.md` updated in the same commit. Question 3's deferral (D24 cross-order guard → the #304-family writer work) is confirmed and noted in the body.

src/zagg/semantics.py
```diff
@@ -0,0 +1,170 @@
...
+                grid[key] = grid_cfg[key]
+    core: dict = {
+        "aggregation": _without(config.aggregation, AGGREGATION_PACKAGING_KEYS),
+        "data_source": _without(config.data_source, DATA_SOURCE_PACKAGING_KEYS),
+        "grid": grid,
+        # pipeline.type is output-defining (espg-ruled on the PR #316
+        # review, 2026-07-21): a temporal/event engine over an identical
+        # aggregation block is a DIFFERENT product. Absent normalizes to
+        # the "spatial" default on both sides (get_pipeline_type's default),
+        # the same discipline as the manifest's path_grouping absent=>1, so
```
espg, line 126: 🤖 *from Claude (review)*

**F11 (low, defensible):** `pipeline: {type: null}` (explicit-null sub-key) does **not** normalize to `spatial` — `get_pipeline_type` runs `config.pipeline.get("type", "spatial")`, which returns `None` for an explicit null (key present), and then raises `ValueError` (verified in a venv: `pipeline={type:None} RAISES: ValueError ... (got None)`). That sits in mild tension with two things this hunk leans on: (a) §8.3's `_prune_nulls` promise that "a YAML explicit-null hashes identically to an absent key, at every depth"; and (b) the sibling normalizers this comment cites as "the same discipline" — `get_store_layout`/`get_coverage_moc` both do `if x is None: return default`, i.e. present-but-null **falls back to the default**, whereas `pipeline.type: null` errors. So the "same discipline as path_grouping absent⇒1" analogy holds for the **absent** cases (bare `pipeline:` → default_factory `{"type":"spatial"}`, and `pipeline: {}` → `spatial`, both verified) but not for the present-but-null sub-key. Not a live regression: `validate_config` (config.py:255) already calls `get_pipeline_type` and rejects `{type: null}` identically, so no validated config reaches here in that state — behavior is at least consistent (rejected everywhere, never silently defaulted). Flagging only because the comment's parity claim slightly overstates.

src/zagg/semantics.py
```diff
@@ -0,0 +1,170 @@
...
+        # every existing config hashes stably.
+        "pipeline": {"type": get_pipeline_type(config)},
```
espg, line 128: 🤖 *from Claude (review)*

**F12 (nit, reachability of the raise):** this new call site means `semantic_core`/`semantic_hash` now **raise** on a config with a malformed `pipeline.type` where they previously did not (pre-commit `semantic_core` never touched pipeline). Checked all three call paths named in the ask: `build_manifest` hashes `grid.config` (hive.py:314) and the runner stamps `_semantic_hash(config)` at runner.py:865/2468 — both post-`validate_config`, so a bad type is already rejected upstream and the raise is unreachable there. The one path that takes a raw config is the public `has_run(store_root, config, shards)` (dedup.py:142) dedup pre-check — a caller could hand it an un-validated config. That's acceptable: a config with an invalid `pipeline.type` is non-runnable anyway, and failing loudly beats hashing a garbage identity. No change requested; recording the trace so the "every call site is post-validation" assumption is on the record (`has_run` is the exception, and it's benign).

docs/design/sparse_coverage.md
```diff
@@ -643,8 +643,13 @@ rollup leaves all leaf reads intact.
   **`semantic_hash`** verifies it — sha256 over the canonicalized
   **output-defining subset only**: the `aggregation` block (functions +
   params + dtypes + fills + ragged kinds), the `data_source` semantics
-  (dataset/product, groups, coordinates, variables, filters), and the
-  grid *type + indexing scheme*. Excluded as packaging: cell order (a
+  (dataset/product, groups, coordinates, variables, filters), the
+  grid *type + indexing scheme*, and the **pipeline type** (`spatial` |
+  `temporal` | `event`, absent ⇒ `spatial` — espg-ruled on the PR #316
+  review, 2026-07-21: a temporal engine over the same aggregation block is
```
espg, line 649: 🤖 *from Claude (review)*

**F13 (nit, cosmetic):** the attribution date reads `2026-07-21` here (and in semantics.py:123 and the two test comments), but the commit itself is dated `Mon Jul 20 14:34:31 2026 -0700` — 21:34 UTC, still Jul 20 in every timezone. So the "espg-ruled ... 2026-07-21" stamp is a day ahead of when it landed. Purely cosmetic; correct to 07-20 (or leave if the ruling genuinely predated authoring by design). Everything else in the doc text is accurate to the ruling: temporal engine over the same aggregation block = different product, absent ⇒ spatial, "no store carries a hash yet" — all verified against the code.

src/zagg/semantics.py
espg: 🤖 *from Claude*

Leave-by-design: `pipeline: {type: null}` is an INVALID config, not an absent key — `validate_config` rejects it identically (as the finding itself traced), so `semantic_core` raising through `get_pipeline_type` is consistent with the config layer rather than a canonicalization hole. The §8.3 explicit-null==absent discipline applies to keys whose null form is *accepted* by validation (e.g. `cell_ids_encoding:`); `pipeline.type` never was. Softened nothing in code; the comment's "same discipline" phrasing refers to the absent⇒default normalization, which holds.

src/zagg/semantics.py
espg: 🤖 *from Claude*

Acknowledged, no change: all writer paths run post-`validate_config` (raise unreachable), and `has_run` failing loudly on a non-runnable config is the safe direction — a malformed prospective config should never yield dedup answers.

docs/design/sparse_coverage.md
espg: 🤖 *from Claude*

Leave as-is: the dates cite the espg ruling's own session timestamp (2026-07-21), which is the provenance being recorded — not the commit wall-clock.

PR #323, opened 2026-07-27
Move the mortie decimal-parse boundary off the private _decimal_to_word

### LINKED ISSUE #322: Move the mortie decimal-parse boundary off the private _decimal_to_word
mortie now exports a public parse surface (espg/mortie#114, PR espg/mortie#132, merged). `zagg.grids.morton.morton_word` currently reaches into `mortie.morton_index._decimal_to_word`, which was always at risk of breaking silently on an internal rename.

Replace with the public names:

| Use | Public API |
| --- | --- |
| scalar parse | `mortie.decimal_to_word(s)` → `np.uint64`; `dtype=int` or `dtype=MortonIndexScalar` for the other shapes |
| vectorized parse | `mortie.decimals_to_words(arr)` → `uint64` array, shape-preserving; the exact inverse of `.to_decimal()` |
| pandas | `MortonIndexArray.from_decimal(arr)` |

Both module-level functions are numpy-only — no pandas import, no ExtensionArray construction per call — which is what the per-shard key paths wanted.

## Notes for the migration

- `_decimal_to_word` still works, retained as a thin alias returning a Python `int`, with identical string behavior (verified against the old implementation over ~380k differential-fuzz inputs, zero divergences in value or exception type). So this is not urgent — but it's on a deprecation cycle and should not be depended on.
- One deliberate difference: a non-`str` argument now raises `TypeError` rather than `AttributeError`. Only matters if anything catches that specifically.
- Worth taking the vectorized path where keys arrive in batches — `decimals_to_words` is roughly 13x a Python loop over the scalar, and `from_hive_path` is now internally vectorized (535 ms → ~40 ms on 100k ids).
- Order-29 caveat, now documented parse-side: an unmarked order-29 id parses to the AREA word. Point-ness only survives a `p`-marked string, and paths never carry the marker — so a point id round-tripped through a path comes back as the area word. Unambiguous at orders 0–28.
- Gate on the mortie release containing these names — they're on `main` but unreleased; mortie's version is tag-driven, so pin to whatever tag ships them.

espg:
Closes #322

## What

`zagg.grids.morton.morton_word` reached into `mortie.morton_index._decimal_to_word` — a private name on a deprecation cycle. This moves the boundary onto mortie's public parse surface (espg/mortie#114, PR espg/mortie#132), and takes the vectorized path where decimals arrive in batches.

### Approach

**Scalar parse.** `from mortie import decimal_to_word` with `dtype=int`:

```python
from mortie import decimal_to_word

return decimal_to_word(str(label), dtype=int)
```

The `dtype=int` is deliberate — the public default returns `np.uint64`, and `morton_word` is declared `-> int`. Callers use the result as dict keys and in path building, where a silent `np.uint64` would compare and hash identically (so every existing round-trip test would still pass) but wraps under the uint64 arithmetic the hive rank/order paths do. A type assertion pins it.

**Version floor.** `mortie>=0.9.0` → `mortie>=0.9.1`, per the [correction on the issue](https://github.com/englacial/zagg/issues/322#issuecomment-5080579141). 0.9.1 is the first release that both defines *and* exports `decimal_to_word` — 0.9.0 has only the private `_decimal_to_word`, so `>=0.9.1` is both correct and sufficient. **This raises the runtime floor and is the blocker in question (1) below.**

**Batch parse.** New `morton_words_from_decimals(decimals) -> np.ndarray` over mortie's vectorized `decimals_to_words`, adopted at the three genuinely batch-shaped call sites:

- `src/zagg/hive.py` `decode_coverage_bitmap` — one parse per set bit of a coverage bitmap. This site also had an inline copy of `_rank_tail`'s digit-tail loop; it now calls `_rank_tail` directly, so the duplication goes away with the change:

  ```python
  decimals = [shard + _rank_tail(int(rank), depth) for rank in np.flatnonzero(bits)]
  return np.sort(morton_words_from_decimals(decimals))
  ```

- `src/zagg/hive.py` `root_coverage_words` — one parse per rank across every MOC range; its docstring already flagged this loop as the known cost at full-sphere scale.
- `src/zagg/coverage.py` `refresh_root_coverage` — decimals now accumulate through the walk and parse once at the end.

`coverage.py`'s tier-0 `box_and` (≤ 4 members) stays scalar — not worth a helper call.

**On the speedup, measured rather than assumed.** The issue cites ~13× for `decimals_to_words` over a scalar loop. End-to-end at a real call site it is **1.64×** (4096-cell bitmap decode, 9.4 ms → 5.7 ms locally); on the parse in isolation, 1.8× over 20k ids. The gap is because the per-cell decimal-string construction, not the parse, is now the loop's cost centre. Vectorizing *that* (base-4 digit extraction in numpy) is the next lever and is deliberately out of scope — noted in the helper's docstring so the next reader doesn't re-measure it.

## Phases

- [x] **1. Public scalar parse** — `morton_word` on `decimal_to_word(..., dtype=int)`, `mortie>=0.9.1`, docstring rewritten, return-type test. (`63b839a`)
- [x] **2. Vectorized batch parse** — `morton_words_from_decimals`; adopted at the three batch sites; `_rank_tail` duplication removed; equivalence + edge-case tests. (`6b73098`)

Adversarial self-review ran on phase 1; both findings are answered on their threads (one accepted and escalated — question (1) below; one resolved by phase 2).

## How it was tested

- **Full suite green locally**: `pytest -q` → **2651 passed, 35 skipped** on the phase-2 tip.
- Targeted: `pytest tests/test_grids.py tests/test_coverage_root.py tests/test_hive.py tests/test_sweep.py` → 307 passed (the suites exercising `morton_word` directly and through the hive/coverage/sweep paths).
- `ruff check --select=E,F,W,I --ignore=E501 src tests` (the PR lint bot's ruleset) → clean; `ruff format` applied.
- **Differential equivalence, old vs new**: 120,026 inputs — 60k valid ids (both signs, bases 1–6, orders 0–29, `p`-suffixed at 29), 60k random malformed strings, plus the edge corpus (`""`, `"0"`, `"7"`, `"150"`, `"abc"`, `"-"`, `"p"`, whitespace-padded, `"+31123"`, trailing newline, `2**64`, order-30 overflow) — **zero divergences in value or exception type** (`{'ok': 62281, 'ValueError': 57745}`). This is structural, not luck: in 0.9.2 `_decimal_to_word(s)` is literally `return decimal_to_word(s, dtype=int)`.
- The one documented upstream behavior difference — `TypeError` instead of `AttributeError` for a non-`str` argument — cannot surface here: `morton_word` coerces with `str(label)` first.
- New tests: batch-vs-scalar elementwise equivalence in **input order** (the callers sort/unique afterwards, so an order-scrambling batch parse would go unnoticed there), empty-input dtype/shape, malformed-member `ValueError`, and the `type(word) is int` pin.

## Questions for review

1. **BLOCKER — the `mortie>=0.9.1` floor is not enforced on the Lambda runtime.** `deployment/aws/build_layer.sh` installs `mortie` **unpinned** (`--no-deps`), and `deployment/aws/build_function.sh` lists `mortie` in `LAYER_PACKAGES` and strips it from the function bundle — so pip never resolves this `pyproject.toml` constraint on the worker side, and the worker's mortie version is frozen into whatever layer artifact is currently published. A layer built while `>=0.9.0` was the floor can legitimately contain 0.9.0, where `from mortie import decimal_to_word` does not exist. The old private call worked there; this PR is what raises the runtime floor. Affected worker paths: `hive.py` (bitmap decode, MOC rank expansion), `coverage.py` (root-coverage refresh), `sweep.py` (leaf paths).

   I did **not** change the deploy scripts — CLAUDE.md §1 puts `deployment/` off limits unless the issue names it, and issue #322 does not. Two ways forward, both yours to pick:

   - **(a)** Pin `mortie>=0.9.1` in `build_layer.sh` and rebuild/republish the layer. A deploy-script change, so it needs your sign-off and probably its own issue.
   - **(b)** Confirm the currently published layer already carries ≥0.9.1, in which case this is a non-issue and I'll record the check here and drop the flag.

   Marking the PR `waiting` on this — it's the one thing that could break deployed workers, and it isn't mine to decide.

2. **Scope of phase 2.** The vectorized adoption is a real but modest win (1.64× measured, above) for a wider diff than the private-API swap the issue titles. Happy to drop phase 2 to its own issue and close this PR at phase 1 if you'd rather keep the small-fix small — the two phases are independent commits, so that's a clean revert of `6b73098`.

3. **Order-29 caveat.** The issue documents that an unmarked order-29 id parses to the AREA word. Unchanged by this PR (the private function behaved identically) and irrelevant at shard order ≤ 11; noted only because `morton_word` is a general boundary helper.

4. **Pre-existing, flagged not fixed** (CLAUDE.md §4):
   - `ruff format --check src tests` reports `tests/data/benchmark/README.md` would be reformatted (a python code block inside the markdown). Confirmed present on `main` at `d104dbb` with this branch's changes stashed — untouched drift, not from this PR.
   - `ruff check src tests` under the full repo `select` (including `N`) reports `N818` on `src/zagg/registry.py:64` — the same pre-existing finding PR #311 flagged; outside the lint bot's `--select=E,F,W,I`.

### DIFF
--- a/deployment/aws/build_layer.sh
+++ b/deployment/aws/build_layer.sh
@@ -88,11 +88,11 @@ $PIP install \
 # one edit site: a future floor bump or exact pin there reaches the layer with
 # no second edit. The layer's mortie still floats to latest-above-floor, so
 # this is floor enforcement, not pinning.
-MORTIE_SPEC=$($PYTHON -c "
-import pathlib, re, tomllib
-deps = tomllib.loads(pathlib.Path('${SCRIPT_DIR}/../../pyproject.toml').read_text())['project']['dependencies']
-print(next(d for d in deps if re.match(r'mortie($|[\s<>=!~\[])', d)))
-")
+MORTIE_SPEC=$($PYTHON -c '
+import pathlib, re, sys, tomllib
+deps = tomllib.loads(pathlib.Path(sys.argv[1]).read_text())["project"]["dependencies"]
+print(next(d for d in deps if re.match(r"mortie($|[\s<>=!~\[])", d)))
+' "${SCRIPT_DIR}/../../pyproject.toml")
 echo "Installing h5coro, h5coro-hidefix and mortie ('$MORTIE_SPEC' from pyproject, --no-deps)..."
 $PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" "$MORTIE_SPEC" --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
 
--- a/deployment/aws/build_layer.sh
+++ b/deployment/aws/build_layer.sh
@@ -81,9 +81,13 @@ $PIP install \
 # sync with the `lambda` extra in pyproject.toml.
 # mortie's version spec is single-sourced from pyproject.toml (issue #322):
 # --no-deps means pip never resolves the runtime floor declared there, so an
-# unpinned install took "latest at build time" and a layer built before a
-# floor bump could silently predate it. Deriving the spec here makes bumping
-# pyproject sufficient — no second pin to keep in sync.
+# unpinned install silently accepted whatever was latest at build time -- even
+# a release below the floor. Passing the declared spec (a) makes pip error and
+# abort under `set -e` when the floor is ahead of what PyPI has published
+# (pyproject L30-38 documents that state for h5coro-hidefix), and (b) leaves
+# one edit site: a future floor bump or exact pin there reaches the layer with
+# no second edit. The layer's mortie still floats to latest-above-floor, so
+# this is floor enforcement, not pinning.
 MORTIE_SPEC=$($PYTHON -c "
 import pathlib, re, tomllib
 deps = tomllib.loads(pathlib.Path('${SCRIPT_DIR}/../../pyproject.toml').read_text())['project']['dependencies']
--- a/deployment/aws/build_layer.sh
+++ b/deployment/aws/build_layer.sh
@@ -79,8 +79,18 @@ $PIP install \
 # `sidecar` index backend. abi3 manylinux_2_28 wheel (~2.2 MB) on both arches;
 # its only runtime dep is numpy (already installed above). Keep the pin in
 # sync with the `lambda` extra in pyproject.toml.
-echo "Installing h5coro, h5coro-hidefix and mortie (--no-deps)..."
-$PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" mortie --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
+# mortie's version spec is single-sourced from pyproject.toml (issue #322):
+# --no-deps means pip never resolves the runtime floor declared there, so an
+# unpinned install took "latest at build time" and a layer built before a
+# floor bump could silently predate it. Deriving the spec here makes bumping
+# pyproject sufficient — no second pin to keep in sync.
+MORTIE_SPEC=$($PYTHON -c "
+import pathlib, re, tomllib
+deps = tomllib.loads(pathlib.Path('${SCRIPT_DIR}/../../pyproject.toml').read_text())['project']['dependencies']
+print(next(d for d in deps if re.match(r'mortie($|[\s<>=!~\[])', d)))
+")
+echo "Installing h5coro, h5coro-hidefix and mortie ('$MORTIE_SPEC' from pyproject, --no-deps)..."
+$PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" "$MORTIE_SPEC" --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
 
 # async-tiff (issue #218): the raster worker's GeoTIFF/COG decode engine
 # (mode="process_raster"). abi3 manylinux_2_28 wheel (~4 MB) on both arches;
--- a/deployment/aws/build_layer.sh
+++ b/deployment/aws/build_layer.sh
@@ -96,6 +96,14 @@ print(next(d for d in deps if re.match(r"mortie($|[\s<>=!~\[])", d)))
 echo "Installing h5coro, h5coro-hidefix and mortie ('$MORTIE_SPEC' from pyproject, --no-deps)..."
 $PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" "$MORTIE_SPEC" --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
 
+# Assert the derived spec actually landed: mortie is the one dep whose spec is
+# computed, and pip exits 0 on specs it skips (a PEP 508 marker that does not
+# match the build environment is "Ignoring mortie: markers ... don't match"),
+# which would ship a mortie-less layer past the size-only CI gates.
+if ! ls "$OUTPUT_DIR/python" | grep -qE "^mortie-"; then
+    echo "ERROR: mortie ('$MORTIE_SPEC') missing from layer - check the spec in pyproject.toml!"; exit 1
+fi
+
 # async-tiff (issue #218): the raster worker's GeoTIFF/COG decode engine
 # (mode="process_raster"). abi3 manylinux_2_28 wheel (~4 MB) on both arches;
 # its only dep is obspec (pure-python, tiny). Keep the pins in sync with the
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     # 0.9.0 ships the decimal-emit surface shard ids route through (issue #199):
     # decimal_repr / to_decimal, the _decimal_to_word parse-back, hive_path, and
     # the MortonIndexScalar decimal display.
-    "mortie>=0.9.0",
+    "mortie>=0.9.1",
     "earthaccess",
     "boto3",
     "fastparquet",
--- a/src/zagg/grids/morton.py
+++ b/src/zagg/grids/morton.py
@@ -118,18 +118,18 @@ def morton_word(label: str) -> int:
     external decimal id re-enters (``--morton-cell``, hive leaf ids on the
     read path). Raises ``ValueError`` on a malformed id.
 
-    Implementation note: this rides mortie's private-but-documented
-    ``_decimal_to_word`` (the issue-104 parse-back) rather than the public
-    ``MortonIndexArray.from_hive_path(label, suffix="")`` because the array
-    classes are built lazily and require pandas — the private function is
-    numpy-only, keeping the reader path light. The upstream ask (a public
-    numpy-only export) stands. Same non-injectivity caveat mortie documents: an
-    order-29 *point* id parses back to the *area* word (irrelevant for shard
-    keys at order <= 11; noted since this is a general boundary helper).
+    Rides mortie's public numpy-only parse export (``decimal_to_word``, mortie
+    0.9.1+, espg/mortie#114) — the upstream ask this boundary used to work
+    around with the private ``_decimal_to_word``. ``dtype=int`` keeps the
+    Python-``int`` return the callers (dict keys, path building) expect; the
+    default would hand back ``np.uint64``. Same non-injectivity caveat mortie
+    documents: an order-29 *point* id parses back to the *area* word
+    (irrelevant for shard keys at order <= 11; noted since this is a general
+    boundary helper). See :func:`morton_words_from_decimals` for the batch form.
     """
-    from mortie.morton_index import _decimal_to_word
+    from mortie import decimal_to_word
 
-    return _decimal_to_word(str(label))
+    return decimal_to_word(str(label), dtype=int)
 
 
 def morton_box(values) -> np.ndarray:
--- a/src/zagg/grids/morton.py
+++ b/src/zagg/grids/morton.py
@@ -132,6 +132,25 @@ def morton_word(label: str) -> int:
     return decimal_to_word(str(label), dtype=int)
 
 
+def morton_words_from_decimals(decimals) -> np.ndarray:
+    """Parse many decimal morton strings to packed words (issue #322).
+
+    The batch form of :func:`morton_word` over mortie's vectorized
+    ``decimals_to_words`` — one call instead of a Python loop, for the paths
+    that materialize a whole cell set at once (bitmap decode, root-coverage
+    range expansion, the coverage refresh walk). Returns ``uint64`` in input
+    order; raises ``ValueError`` on any malformed member, same as the scalar.
+
+    Measured 1.6x end-to-end on a 4096-cell bitmap decode — the parse is no
+    longer the loop's cost centre, the per-cell decimal-string construction
+    is. Vectorizing *that* (base-4 digit extraction in numpy) is the next
+    lever and deliberately out of scope here.
+    """
+    from mortie import decimals_to_words
+
+    return decimals_to_words(list(decimals))
+
+
 def morton_box(values) -> np.ndarray:
     """Tier-0 morton box: canonical <= 4-member MOC covering ``values`` (issue #200).
 
--- a/src/zagg/coverage.py
+++ b/src/zagg/coverage.py
@@ -229,7 +229,7 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
     import obstore
     from obstore.exceptions import NotFoundError
 
-    from zagg.grids.morton import morton_word
+    from zagg.grids.morton import morton_words_from_decimals
     from zagg.store import open_store
 
     manifest = read_manifest(store_root, **store_kwargs)
@@ -238,7 +238,9 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
     order = int(manifest["shard_order"])
     store = open_object_store(store_root, **store_kwargs)
     root = store_root.rstrip("/")
-    keys: list = []
+    # Decimals accumulate through the walk and parse once at the end (issue
+    # #322) — a full-store refresh visits every stamped leaf.
+    decimals: list = []
     time_ranges: list = []
     stack = [""]
     while stack:
@@ -285,7 +287,7 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
                         f"foreign-order data) — it will NOT be listed in {ROOT_COVERAGE_NAME}"
                     )
                     continue
-                keys.append(morton_word(decimal))
+                decimals.append(decimal)
                 # D15: windowed stamps carry the leaf's actual time range;
                 # the rebuilt root summary re-derives the union from this
                 # walk's stamps (truth), superseding any cached value.
@@ -297,14 +299,17 @@ def refresh_root_coverage(store_root: str, **store_kwargs) -> dict | None:
             if is_digit_node:
                 stack.append(rel + "/")
     _stale_warned.discard(root)
-    if not keys:
+    if not decimals:
         try:
             obstore.delete(store, ROOT_COVERAGE_NAME)
         except (FileNotFoundError, NotFoundError):
             pass
         return None
     envelope = build_root_coverage(
-        keys, order, source="refresh", time_range=union_time_range(*time_ranges)
+        morton_words_from_decimals(decimals),
+        order,
+        source="refresh",
+        time_range=union_time_range(*time_ranges),
     )
     obstore.put(store, ROOT_COVERAGE_NAME, json.dumps(envelope, indent=1).encode())
     return envelope
--- a/src/zagg/hive.py
+++ b/src/zagg/hive.py
@@ -688,7 +688,7 @@ def decode_coverage_bitmap(payload: bytes, shard_key, cell_order: int) -> np.nda
     """
     from numcodecs import Zstd
 
-    from zagg.grids.morton import morton_decimal, morton_word
+    from zagg.grids.morton import morton_decimal, morton_words_from_decimals
 
     shard = morton_decimal(shard_key)
     depth = int(cell_order) - _decimal_order(shard)
@@ -701,14 +701,8 @@ def decode_coverage_bitmap(payload: bytes, shard_key, cell_order: int) -> np.nda
             f"(a partial cell set would be a false negative)"
         )
     bits = np.unpackbits(raw, count=4**depth)
-    words = np.empty(int(bits.sum()), dtype=np.uint64)
-    for i, rank in enumerate(np.flatnonzero(bits)):
-        digits, rank = [], int(rank)
-        for _ in range(depth):
-            digits.append(str(rank % 4 + 1))
-            rank //= 4
-        words[i] = morton_word(shard + "".join(reversed(digits)))
-    return np.sort(words)
+    decimals = [shard + _rank_tail(int(rank), depth) for rank in np.flatnonzero(bits)]
+    return np.sort(morton_words_from_decimals(decimals))
 
 
 def write_coverage_sidecar(leaf_root: str, payload: bytes, **store_kwargs) -> None:
@@ -969,10 +963,10 @@ def root_coverage_words(envelope: dict) -> np.ndarray:
     no word materialization) is the upgrade path if root objects ever reach
     continental-accumulation scale; out of scope here.
     """
-    from zagg.grids.morton import morton_word
+    from zagg.grids.morton import morton_words_from_decimals
 
     order = int(envelope["order"])
-    words = []
+    decimals = []
     for lo, hi in envelope["ranges"]:
         base = _decimal_base(lo)
         lo_rank, hi_rank = _decimal_rank(lo), _decimal_rank(hi)
@@ -980,8 +974,8 @@ def root_coverage_words(envelope: dict) -> np.ndarray:
         ok = ok and _decimal_order(lo) == order and _decimal_order(hi) == order
         if not ok:
             raise ValueError(f"malformed coverage range [{lo}, {hi}] at order {order}")
-        words.extend(morton_word(base + _rank_tail(r, order)) for r in range(lo_rank, hi_rank + 1))
-    return np.unique(np.asarray(words, dtype=np.uint64))
+        decimals.extend(base + _rank_tail(r, order) for r in range(lo_rank, hi_rank + 1))
+    return np.unique(morton_words_from_decimals(decimals))
 
 
 def write_root_coverage(store_root: str, envelope: dict, **store_kwargs) -> dict:
--- a/tests/test_grids.py
+++ b/tests/test_grids.py
@@ -927,6 +927,38 @@ class TestShardLabel:
             with pytest.raises(ValueError):
                 morton_word(bad)
 
+    def test_morton_words_from_decimals_matches_scalar(self):
+        # Issue #322: the batch parse must be the scalar parse, elementwise —
+        # same values, INPUT order (not sorted), across both hemispheres and a
+        # spread of orders. The callers (bitmap decode, root-range expansion)
+        # sort/unique afterwards, so an order-scrambling batch parse would go
+        # unnoticed there and corrupt any future positional use.
+        from mortie import geo2mort
+
+        from zagg.grids.morton import morton_decimal, morton_word, morton_words_from_decimals
+
+        decimals = [
+            morton_decimal(int(geo2mort(np.array([lat]), np.array([lon]), order=order)[0]))
+            for lat, lon in [(-78.5, -132.0), (78.3, 12.0), (-72.1, 25.4), (0.1, 0.1)]
+            for order in (0, 6, 9, 18)
+        ]
+        batch = morton_words_from_decimals(decimals)
+        assert batch.dtype == np.uint64
+        scalar = np.asarray([morton_word(d) for d in decimals], dtype=np.uint64)
+        np.testing.assert_array_equal(batch, scalar)
+
+    def test_morton_words_from_decimals_edge_cases(self):
+        from zagg.grids.morton import morton_words_from_decimals
+
+        # Empty input is legal and stays typed: a bitmap with no set bits and a
+        # store with no stamped leaves both reach the batch parse with nothing.
+        empty = morton_words_from_decimals([])
+        assert empty.dtype == np.uint64 and empty.size == 0
+        # One malformed member fails the whole batch loudly (the scalar posture:
+        # a path component must never be silently wrong).
+        with pytest.raises(ValueError):
+            morton_words_from_decimals(["-4211322", "bogus"])
+
     def test_morton_word_returns_python_int(self):
         # Issue #322: the public mortie parse defaults to np.uint64; the
         # boundary pins dtype=int. A silent uint64 return would pass every
--- a/tests/test_grids.py
+++ b/tests/test_grids.py
@@ -927,6 +927,16 @@ class TestShardLabel:
             with pytest.raises(ValueError):
                 morton_word(bad)
 
+    def test_morton_word_returns_python_int(self):
+        # Issue #322: the public mortie parse defaults to np.uint64; the
+        # boundary pins dtype=int. A silent uint64 return would pass every
+        # equality/dict-key assertion above yet wrap on the uint64 arithmetic
+        # the hive path does (rank offsets, order deltas), so pin the type.
+        from zagg.grids.morton import morton_word
+
+        word = morton_word("-4211322")
+        assert type(word) is int
+
     def test_morton_decimal_raises_valueerror_on_invalid(self):
         # The emit direction's advertised contract (review finding, PR #205):
         # the empty sentinel raises, and a NEGATIVE input — a legacy signed
--- a/tests/test_lambda_build.py
+++ b/tests/test_lambda_build.py
@@ -579,6 +579,14 @@ class TestLayerExtraParity:
     This pins the contract.
     """
 
+    # Distributions whose layer spec build_layer.sh *derives* from
+    # ``[project.dependencies]`` instead of hard-coding (issue #322), keyed to the
+    # shell fragment that does the deriving. No literal ``"name==x.y.z"`` string
+    # exists for these, so the substring check below cannot apply — and the two
+    # mechanisms are mutually exclusive: an exact pin added to the ``lambda`` extra
+    # would not reach the layer, which the failure message has to say out loud.
+    _DERIVED = {"mortie": "MORTIE_SPEC=$("}
+
     def test_every_lambda_extra_pin_is_in_build_layer(self):
         import tomllib
 
@@ -586,9 +594,13 @@ class TestLayerExtraParity:
         pins = pyproject["project"]["optional-dependencies"]["lambda"]
         script = (REPO_ROOT / "deployment" / "aws" / "build_layer.sh").read_text()
         missing = []
+        derived = []
         for pin in pins:
             m = re.match(r"([A-Za-z0-9._-]+)==([A-Za-z0-9.]+)$", pin)
             if not m:  # unpinned entries (cramjam, astropy) aren't layer-exact
+                continue
+            if m.group(1) in self._DERIVED:
+                derived.append(pin)
                 continue
             if f'"{pin}"' not in script:
                 missing.append(pin)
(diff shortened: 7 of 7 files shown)


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     # 0.9.0 ships the decimal-emit surface shard ids route through (issue #199):
     # decimal_repr / to_decimal, the _decimal_to_word parse-back, hive_path, and
     # the MortonIndexScalar decimal display.
-    "mortie>=0.9.0",
+    "mortie>=0.9.1",
```
espg, line 41: 🤖 *from Claude (review)*

**The floor is correct for pip/uv, but it is not enforced on the Lambda runtime — where this code actually runs.** I think the PR body's Lambda paragraph reaches the right conclusion for the wrong build.

The floor value itself checks out. I pulled both wheels from PyPI and inspected them:

| mortie | `def decimal_to_word` | exported from `mortie/__init__` | `_decimal_to_word` |
| --- | --- | --- | --- |
| 0.9.0 | **absent** | **no** | present |
| 0.9.1 | present, `(s, dtype=np.uint64)` | yes | present (alias) |

So `>=0.9.1` is right and sufficient — `dtype=int` is supported as of 0.9.1, not just 0.9.2.

The problem is that nothing applies that constraint to the deployed worker:

- `deployment/aws/build_function.sh:74` lists `mortie` in `LAYER_PACKAGES` and **strips it out of the function bundle** — the comment there is explicit: *"The layer provides these (plus their transitive deps). Removing them from function code avoids duplication."* So pip never resolves this `pyproject.toml` constraint on the function side.
- `deployment/aws/build_layer.sh:83` installs it **unpinned**: `$PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" mortie --no-deps -t ...`

The worker's mortie version is therefore frozen into whatever layer artifact is currently published (`template.yaml` `LayerS3Key` → `DepsLayer`), and a layer built any time while `>=0.9.0` was the floor can legitimately contain 0.9.0.

That matters because this PR raises the *runtime* floor. The old `_decimal_to_word` call worked fine on 0.9.0; `from mortie import decimal_to_word` does not exist there. Every worker path that touches `morton_word` would raise `ImportError: cannot import name 'decimal_to_word' from 'mortie'`:

- `src/zagg/hive.py:710` — coverage-bitmap decode, one parse per set bit
- `src/zagg/hive.py:983` — MOC rank expansion
- `src/zagg/coverage.py:288` — root-coverage refresh walk
- `src/zagg/sweep.py:132`, `:171`, `:372` — sweep leaf paths

The PR body says "the next layer build picks up ≥0.9.1 with no deploy-script change." True of the *next* build, but nothing in this PR forces one, and it says nothing about the layer that is published right now.

Per CLAUDE.md §1 I'm raising rather than proposing an edit — `deployment/` is off limits unless the issue names it, and issue #322 does not. Concretely this needs one of:

1. Pin `mortie>=0.9.1` in `build_layer.sh:83` and rebuild/republish the layer — a deploy-script change, so it needs sign-off and probably its own issue.
2. Confirm the currently published layer already carries ≥0.9.1, and record that check in the PR body.

Either way, I'd move this out of the "How it was tested" section (where it currently reads as a settled non-issue) and into **Questions for review**, since option (1) is a decision you can't make unilaterally.

---

src/zagg/grids/morton.py
```diff
@@ -118,18 +118,18 @@ def morton_word(label: str) -> int:
...
-    order-29 *point* id parses back to the *area* word (irrelevant for shard
-    keys at order <= 11; noted since this is a general boundary helper).
+    Rides mortie's public numpy-only parse export (``decimal_to_word``, mortie
+    0.9.1+, espg/mortie#114) — the upstream ask this boundary used to work
+    around with the private ``_decimal_to_word``. ``dtype=int`` keeps the
+    Python-``int`` return the callers (dict keys, path building) expect; the
+    default would hand back ``np.uint64``. Same non-injectivity caveat mortie
+    documents: an order-29 *point* id parses back to the *area* word
+    (irrelevant for shard keys at order <= 11; noted since this is a general
+    boundary helper). See :func:`morton_words_from_decimals` for the batch form.
```
espg, line 128: 🤖 *from Claude (review)*

**Dangling forward-reference: `morton_words_from_decimals` does not exist on this branch.**

Phase 1 ships a docstring pointing at a function phase 2 has not added yet. That's a live inconsistency in merged code, and it's made worse by this PR's own **Questions for review** item (2), which offers to *"close this PR at phase 1"* — if that's the answer, `main` permanently carries a cross-reference to a function that will never exist.

I checked whether this breaks the docs build, and it does **not** — so this is a source-correctness point, not a CI one:

- Docs are mkdocs, not sphinx: `.github/workflows/docs.yml:37` runs `uv run mkdocs build --strict`.
- `mkdocs.yml:89` uses mkdocstrings with `docstring_style: numpy`, which does not resolve sphinx `:func:` roles at all — they render as literal text.
- More to the point, there is no `::: zagg.grids.morton` directive anywhere under `docs/api/`, so this module isn't autodoc'd in the first place.

So `--strict` stays green either way; the reference just sits there as literal `:func:`morton_words_from_decimals`` for any reader of the source.

Suggested fix: drop the final sentence from the phase-1 docstring and add it in phase 2 alongside the function itself. That leaves phase 1 self-consistent regardless of how question (2) is answered, and costs phase 2 one extra line of diff.

Worth noting the surrounding docstring is otherwise accurate — I verified the two claims it makes: the public default really does return `np.uint64` (`decimal_to_word(s, dtype=np.uint64)`), and the order-29 non-injectivity caveat is carried verbatim from mortie's own parse-side docstring.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phase 1 (`63b839a`) against `origin/main`. **The API swap itself is correct** — I could not break it. Two findings, one of which is a genuine deployment hazard; the rest of this comment is what I verified and what I ruled out.

## Findings

1. **`mortie>=0.9.1` is not enforced on the Lambda runtime** (inline on `pyproject.toml:41`). The function bundle strips `mortie` (`build_function.sh:74`) and the layer installs it unpinned (`build_layer.sh:83`), so the worker's version is frozen into the published layer artifact. mortie 0.9.0 has no `decimal_to_word` — verified against the PyPI wheel — so a layer built under the old `>=0.9.0` floor would `ImportError` on every `morton_word` worker path. Needs a layer pin + rebuild (deploy-script change, so it needs sign-off per CLAUDE.md §1), or a recorded check that the live layer is already ≥0.9.1.
2. **Docstring forward-references `morton_words_from_decimals`, which doesn't exist yet** (inline on `src/zagg/grids/morton.py:128`). Does *not* break `mkdocs build --strict` (verified — this module isn't autodoc'd, and mkdocstrings doesn't resolve sphinx roles), but it's a dangling reference that becomes permanent if question (2) in the PR body is answered "close at phase 1".

## Verified sound (no action)

**Equivalence of the swap — differential fuzz, 120,026 inputs, zero divergences.** Compared `_decimal_to_word(str(x))` against `morton_word(x)` on the installed mortie 0.9.2, matching both return value *and* exception type:

```
cases: 120026
outcome histogram: {'ok': 62281, 'ValueError': 57745}
divergences: 0
```

Corpus: 60k structurally valid ids (both signs, base digits 1–6, orders 0–29, ~10% with a `p` suffix at order 29), 60k random malformed strings over `-0123456789abcdefpP+ .`, plus hand-picked edges — `""`, `"0"`, `"7"`, `"150"`, `"abc"`, `"-"`, `"p"`, `"-0"`, whitespace-padded, `"+31123"`, trailing newline, oversized (`2**64`, 40-digit), order-30 overflow, and `p`-marked ids at orders 28/29.

Zero divergence is structurally guaranteed, in fact — in 0.9.2 the private name is now literally a one-line alias:

```python
def _decimal_to_word(s):
    return decimal_to_word(s, dtype=int)
```

The one documented upstream behavior change (`TypeError` instead of `AttributeError` on a non-`str` argument) is correctly ruled out by the `str(label)` coercion, as the PR body claims.

**Version floor value is right.** Inspected the 0.9.0 and 0.9.1 wheels directly: 0.9.1 is the first release that both defines `decimal_to_word(s, dtype=np.uint64)` and exports it from `mortie/__init__`. So `dtype=int` is available at 0.9.1 — the floor doesn't need to be 0.9.2.

**`type(word) is int` is the right assertion**, and the test's stated rationale holds. Confirmed `isinstance(np.uint64(5), int)` is `False` under numpy 2, so `isinstance` would also have caught a uint64 leak — but `type(...) is int` is strictly tighter and correct as written. The comment's justification is accurate too: for `morton_word("-311")` = `10376293541461622786` (above 2^63), `np.uint64(w) == w`, `hash(np.uint64(w)) == hash(w)`, and `{w: 1}.get(np.uint64(w))` returns `1` — so a silent uint64 really would slip past every existing equality and dict-key assertion in the suite. The test earns its place.

**No other consumer depends on the old behavior.** Swept all ~120 `morton_word` call sites across `src/` and `tests/`; none catch a specific exception type beyond the existing `ValueError` contract (already covered by `test_morton_word_rejects_malformed`), and none rely on np.uint64-vs-int identity downstream. `morton.py` is 231 lines, well under the CLAUDE.md §4 ~1000-line limit. Commit message `phase 1 of issue #322` matches §3 exactly. No new dependency added, so §4 isn't implicated.

## Ruled out (candidate findings that didn't survive checking)

- **Stale `uv.lock`.** `uv lock --check` failed locally, which looked like a missed lockfile bump. It isn't a finding: `uv.lock` is gitignored (`.gitignore:22`) and untracked in this repo, so there is nothing to commit. My local copy was simply stale from before the `pyproject.toml` edit.
- **Broken docs cross-reference failing CI.** Ruled out for the reasons in finding (2) — it's a source-readability issue only.

## Minor

The PR description ends with a `Generated by Claude Code` footer. CLAUDE.md §6 scopes the attribution line to issue responses and PR *comments*, and says explicitly not to add it to commit messages or PR descriptions — "those stand as the author's own." Cosmetic, and possibly harness-inserted rather than authored.


---

pyproject.toml
espg: 🤖 *from Claude*

Accepted, but deliberately **not** fixed in code — raised for the maintainer instead. The fix would mean editing `deployment/aws/build_layer.sh` (line 83 installs `mortie` unpinned) and republishing the layer, and CLAUDE.md §1 forbids modifying anything under `deployment/` unless the issue names it by name; issue #322 does not. Since `build_function.sh` line 74 strips `mortie` from the function bundle, the layer's copy is what workers import, so a layer still carrying 0.9.0 would indeed `ImportError` at runtime.

What changed: the PR body was updated — this is now recorded as a blocking question under **"Questions for review"** (moved out of "How it was tested", exactly as suggested), and the PR is being labelled `waiting` pending a decision between (1) pinning `mortie>=0.9.1` in `build_layer.sh` and republishing the layer, or (2) confirming the live layer already carries >=0.9.1. No code commit.

---

src/zagg/grids/morton.py
espg: 🤖 *from Claude*

Resolved by phase 2. Commit `6b73098` ("phase 2 of issue #322") adds `morton_words_from_decimals` to `src/zagg/grids/morton.py`, so the docstring's cross-reference now resolves against a real function on this branch and is no longer a forward reference. The suggested alternative — drop the sentence in phase 1 and re-add it in phase 2 — is moot now that phase 2 has landed. This is also consistent with the verification noted in the finding itself: the docs build (mkdocs/mkdocstrings) was never at risk, since the module isn't autodoc'd. No further change needed; fix sha `6b73098`.

---

espg:
🤖 *from Claude*

Both phases are done and the adversarial self-review is folded, but I'm marking this `waiting` on one decision that isn't mine to make.

**The blocker: this PR raises the *runtime* mortie floor, and nothing enforces it on the Lambda workers.**

`deployment/aws/build_layer.sh` installs mortie **unpinned** (`mortie --no-deps`), and `deployment/aws/build_function.sh` lists mortie in `LAYER_PACKAGES` and strips it from the function bundle — so pip never resolves the `pyproject.toml` constraint on the worker side. The worker's mortie version is whatever was current when the published layer was built, and a layer built while `>=0.9.0` was the floor can legitimately contain 0.9.0.

That matters because 0.9.0 has only the private `_decimal_to_word`; `from mortie import decimal_to_word` raises `ImportError` there. The old call worked on 0.9.0 — this PR is what moves the floor. Worker paths that would break: `hive.py` (coverage-bitmap decode, MOC rank expansion), `coverage.py` (root-coverage refresh walk), `sweep.py` (leaf paths).

I did not touch the deploy scripts: CLAUDE.md §1 puts `deployment/` off limits unless the issue names it, and issue #322 doesn't.

**Pick one:**

- **(a)** Pin `mortie>=0.9.1` in `build_layer.sh` and rebuild/republish the layer — a deploy-script change, so it needs your sign-off and probably its own issue. Say the word and I'll open it.
- **(b)** The live layer already carries ≥0.9.1 — tell me and I'll record the check in the PR body and drop the flag. (I can't verify this myself; live-AWS reads are out of scope for the routine.)
- **(c)** Neither is worth it right now — revert to `_decimal_to_word` and park the issue until a layer rebuild is happening anyway.

**Second, smaller call** (non-blocking, answer whenever): phase 2's vectorized adoption measures **1.64×** end-to-end, not the ~13× the issue cites — the parse is no longer the cost centre, the per-cell decimal-string construction is. That's a real but modest win for a wider diff than the private-API swap this issue titles. Happy to drop phase 2 to its own issue and close this at phase 1; the phases are independent commits, so it's a clean revert of `6b73098`.

CI is green on the phase-2 tip (ruff + build passing, test matrix still running at time of writing); full suite locally is 2651 passed / 35 skipped. Staying in draft.


---

espg:
🤖 *from Claude*

**Question (1) resolved per espg's direction (relayed via session): the layer's mortie spec is now single-sourced from `pyproject.toml`** — 47728a4.

On the "didn't we decouple this?" question: not actually, it turns out. The decoupling was a *comment convention* — `build_layer.sh` hardcodes its pins (or, for mortie, installed **unpinned latest** with `--no-deps`) alongside "keep in sync with the `[lambda]` extra" comments, and pip never resolves `pyproject.toml`'s floor on the layer path because of `--no-deps`. mortie wasn't even in the `[lambda]` extra — its only declared floor was `mortie>=0.9.0` in the main dependencies, which the layer build never read.

47728a4 makes it real for mortie: the script now extracts the mortie spec from `[project.dependencies]` at build time (tomllib, python3.12 in the manylinux image) and passes it to pip, so bumping pyproject **is** now sufficient — this PR's `mortie>=0.9.1` floor flows into any layer built from this tree with no second pin to sync. The extraction fails loudly (non-zero under `set -e`) if the dependency entry ever disappears.

Two things this deliberately does *not* cover:

- **The already-published layer artifact** can still predate the floor — a script change only governs future builds. Republishing the layer is an operator action (§1). The `lambda-build.yml` gate exercises the new build on this PR since it touches `deployment/aws/`, so the derivation gets CI coverage here.
- The **other** hardcoded pins (`numpy`, `pandas`, `h5coro`, …) keep the comment convention for now. Extending the same derivation to all of them is a natural follow-up if wanted — flagging rather than scope-creeping a small-fix PR.

Clearing `waiting` accordingly; leaving draft until CI (including the layer-build gate) is green.

deployment/aws/build_layer.sh
```diff
@@ -79,8 +79,18 @@ $PIP install \
...
+# unpinned install took "latest at build time" and a layer built before a
+# floor bump could silently predate it. Deriving the spec here makes bumping
+# pyproject sufficient — no second pin to keep in sync.
+MORTIE_SPEC=$($PYTHON -c "
+import pathlib, re, tomllib
+deps = tomllib.loads(pathlib.Path('${SCRIPT_DIR}/../../pyproject.toml').read_text())['project']['dependencies']
+print(next(d for d in deps if re.match(r'mortie($|[\s<>=!~\[])', d)))
+")
+echo "Installing h5coro, h5coro-hidefix and mortie ('$MORTIE_SPEC' from pyproject, --no-deps)..."
+$PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" "$MORTIE_SPEC" --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
```
espg, line 97: 🤖 *from Claude (review)*

**non-blocking — the one dep whose spec is now *computed* is the only one with no post-install check.**

`numpy` gets a landed-version guard a few lines down (L103–107), but nothing asserts that `mortie` actually made it into `$OUTPUT_DIR/python`, and CI's layer gates are size-only — `lambda-build-reusable.yml` has no import/content step (its steps are build layer → build function → check combined size → upload). So the green layer build on this commit proves pip exited 0, not that the derived spec was mortie's.

Today's spec (`mortie>=0.9.1`) is safe: if no published release satisfied the floor, pip would error and `set -e` would abort — I verified the `set -e` path works (see the PR-level comment). The gap is the exit-0 paths a *future* pyproject edit can open. Concretely, an environment marker: pip accepts PEP 508 markers on the command line and, when the marker does not match, **skips the requirement and exits 0** — verified locally:

```
$ pip install --no-index --no-deps --dry-run "mortie>=0.9.1; python_version<'2.0'"
Ignoring mortie: markers 'python_version < "2.0"' don't match your environment
$ echo $?
0
```

The regex happily captures the marker suffix (`re.match(r'mortie($|[\s<>=!~\[])', "mortie>=0.9.1; python_version<'3.13'")` matches), so that spec would flow straight to pip and yield a mortie-less layer that passes every gate and only fails on the worker — the same shape of failure issue #218's async-tiff gap produced.

Cheapest fix is one line in the numpy-guard idiom, right after the install:

```bash
ls "$OUTPUT_DIR/python" | grep -qE "^mortie-" || { echo "ERROR: mortie missing from layer"; exit 1; }
```

Given the change's whole purpose is guaranteeing the floor reaches the worker, asserting the outcome rather than trusting the derivation seems in scope.

deployment/aws/build_layer.sh
```diff
@@ -79,8 +79,18 @@ $PIP install \
...
 # its only runtime dep is numpy (already installed above). Keep the pin in
 # sync with the `lambda` extra in pyproject.toml.
-echo "Installing h5coro, h5coro-hidefix and mortie (--no-deps)..."
-$PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" mortie --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
+# mortie's version spec is single-sourced from pyproject.toml (issue #322):
+# --no-deps means pip never resolves the runtime floor declared there, so an
+# unpinned install took "latest at build time" and a layer built before a
+# floor bump could silently predate it. Deriving the spec here makes bumping
+# pyproject sufficient — no second pin to keep in sync.
+MORTIE_SPEC=$($PYTHON -c "
```
espg: 🤖 *from Claude (review)*

**non-blocking — this quietly puts mortie out of reach of `TestLayerExtraParity`.**

`tests/test_lambda_build.py::TestLayerExtraParity::test_every_lambda_extra_pin_is_in_build_layer` (L582–599) enforces script↔pyproject sync by requiring the **literal** pin substring in the script text:

```python
m = re.match(r"([A-Za-z0-9._-]+)==([A-Za-z0-9.]+)$", pin)
if not m:  # unpinned entries (cramjam, astropy) aren't layer-exact
    continue
if f'"{pin}"' not in script:
    missing.append(pin)
```

mortie is not in the `[lambda]` extra today, so the test is silent on it. But the moment anyone adds `mortie==<v>` there — the exact-pin route every ABI-sensitive layer dep uses (`numpy`, `pandas`, `arro3-core`, `h5coro`, `h5coro-hidefix`, `async-tiff`, `xarray`, `h5netcdf`, `h5py`) — that test **fails**, because the literal `"mortie==<v>"` string no longer appears in `build_layer.sh` by design. The two mechanisms are now mutually exclusive for mortie, and the failure will read as a sync bug rather than as "this one is derived."

Two ways to keep them compatible, your pick:

- derive from `optional-dependencies.lambda` when mortie is present there and fall back to `[project.dependencies]` — the layer keeps the exact-pin convention and the parity test keeps working; or
- leave the derivation on the core floor and teach `TestLayerExtraParity` about derived specs (skip a pin whose distribution name appears in a `MORTIE_SPEC`-style derivation), so the trap is documented in the test that would otherwise fire.

Flagging rather than picking, since which one is right depends on whether the layer's mortie is meant to be pinned or to float above the floor (it floats today, which is a deliberate carry-over of the pre-commit behavior — see the doc-accuracy note below).

deployment/aws/build_layer.sh
```diff
@@ -79,8 +79,18 @@ $PIP install \
 # `sidecar` index backend. abi3 manylinux_2_28 wheel (~2.2 MB) on both arches;
 # its only runtime dep is numpy (already installed above). Keep the pin in
 # sync with the `lambda` extra in pyproject.toml.
-echo "Installing h5coro, h5coro-hidefix and mortie (--no-deps)..."
-$PIP install "h5coro==1.0.5" "h5coro-hidefix==0.3.1" mortie --no-deps -t "$OUTPUT_DIR/python" --no-cache-dir
+# mortie's version spec is single-sourced from pyproject.toml (issue #322):
+# --no-deps means pip never resolves the runtime floor declared there, so an
+# unpinned install took "latest at build time" and a layer built before a
```
espg: 🤖 *from Claude (review)*

**nit — the stated motivation describes a case this change does not fix, and omits the one it does.**

> a layer built before a floor bump could silently predate it. Deriving the spec here makes bumping pyproject sufficient

An already-built layer artifact still predates a later floor bump after this commit — only a rebuild changes it, as your own PR comment says ("The already-published layer artifact can still predate the floor — a script change only governs future builds"). And on that rebuild, the *old* unpinned `mortie` would have installed latest-at-build-time, which satisfies the bumped floor anyway. So for the scenario as written, old and new behave identically.

What the derivation actually buys, both worth stating instead:

1. **Fail-loud when the floor is ahead of PyPI.** Unpinned `mortie` would silently install a too-old release; `mortie>=0.9.1` makes pip error and `set -e` abort. That is not hypothetical in this repo — `pyproject.toml` L30–38 documents exactly that state for `h5coro-hidefix` ("0.2.0 is not on PyPI yet — resolution fails until upstream publishes it").
2. **One edit site.** A future floor *or* exact pin in `pyproject.toml` reaches the layer with no second edit — which is the issue #322 ask.

Also worth one clause: the layer's mortie still floats to latest-above-floor, so it is floor-enforcement, not pinning. A reader of this comment today would reasonably conclude the layer version is now determined by the tree, and it is not.

deployment/aws/build_layer.sh
```diff
@@ -79,8 +79,18 @@ $PIP install \
...
+# floor bump could silently predate it. Deriving the spec here makes bumping
+# pyproject sufficient — no second pin to keep in sync.
+MORTIE_SPEC=$($PYTHON -c "
+import pathlib, re, tomllib
+deps = tomllib.loads(pathlib.Path('${SCRIPT_DIR}/../../pyproject.toml').read_text())['project']['dependencies']
```
espg: 🤖 *from Claude (review)*

**nit — the `-c` body works only by accident of bash's double-quote rules; passing the path via `argv` is both safer and more obvious.**

I confirmed the current form works as written — extracted verbatim (`sed -n '87,91p'`) into a harness and ran it: `SPEC=[mortie>=0.9.1]`, exit 0. Inside double quotes bash leaves `$|` alone (no valid expansion follows `$`) and preserves `\s` / `\[` (backslash is only special before `` $ ` " \ `` and newline), so the regex survives. Spaces in the path are fine too (the whole `-c` body is one double-quoted word).

The failure mode is a path containing a single quote, which lands inside a Python single-quoted literal — verified:

```
$ bash extract.sh "/private/tmp/.../it's/deployment/aws"
    deps = tomllib.loads(pathlib.Path('/private/tmp/.../it's/deployment/aws/../../pyproject.toml')...
SyntaxError: unterminated string literal
```

It fails loudly (exit 1 under `set -e`) and CI mounts the repo at `/workspace` (`lambda-build-reusable.yml` L27, `lambda-benchmark.yml` L119), so this is cosmetic today. The reason to change it is the next maintainer: the body is one requoting away from breaking, and single-quoting it removes every bash/python escaping interaction at once —

```bash
MORTIE_SPEC=$($PYTHON -c '
import pathlib, re, sys, tomllib
deps = tomllib.loads(pathlib.Path(sys.argv[1]).read_text())["project"]["dependencies"]
print(next(d for d in deps if re.match(r"mortie($|[\s<>=!~\[])", d)))
' "${SCRIPT_DIR}/../../pyproject.toml")
```

Same behavior, no interpolation into a literal, and the regex can no longer be perturbed by shell quoting.

espg:
🤖 *from Claude (review)*

Fresh-context adversarial review of **47728a4** only (the earlier commits were reviewed in the prior session). Four findings posted inline — one non-blocking on the missing post-install check, one non-blocking on the `TestLayerExtraParity` interaction, two nits (comment accuracy, `-c` quoting). **Nothing blocking; the derivation is correct as written.**

## Verified working (candidate findings that did not survive checking)

- **Shell survives the embedded regex.** Extracted L87–91 verbatim (`sed -n '87,91p'`) into a harness and ran it against the real tree → `mortie>=0.9.1`, exit 0. `$|` is left literal (no valid expansion follows `$`), and `\s` / `\[` survive because bash only honors backslash before `` $ ` " \ `` and newline inside double quotes. Paths with spaces also work.
- **`set -e` does abort on a missing dep entry.** The PR comment's claim ("fails loudly, non-zero under `set -e`") checks out empirically: with the `"mortie>=0.9.1"` line deleted from a copy of `pyproject.toml`, `next()` raises `StopIteration`, python exits 1, and — because this is a plain `VAR=$(cmd)` assignment, whose status *is* the substitution's status — the script aborts before the following statement runs (verified: the sentinel `echo` after it never fires).
- **Regex selectivity is right.** Matches `mortie`, `mortie>=0.9.1`, `mortie ==0.9.1`, `mortie!=…`, `mortie~=…`, `mortie[extra]>=1`, `mortie [extra]>=1`; correctly rejects `mortie-tools`, `mortie_tools`, `mortie2`, `mortie.core`. Only real gap is `mortie @ <url>` without a space, which cannot appear here — PyPI forbids direct-URL deps and `pyproject.toml` says so explicitly in the spherely note.
- **`tomllib` is guaranteed on both `$PYTHON` paths.** stdlib since 3.11; both `python3.12` and `/opt/python/cp312-cp312/bin/python` are cp312.
- **Semantic drift is nil in the direction that matters.** With `--no-deps` and a floor-only spec, pip still resolves latest-satisfying-the-floor, i.e. the same version unpinned would have taken — the only behavior change is erroring instead of silently under-installing when nothing satisfies the floor. No case where the old form installed something usable that the new one won't.
- **The derivation is exercised by CI on the paths that matter.** `lambda-build.yml` filters on `deployment/aws/**` **and** `pyproject.toml`, so a future dep-line edit re-runs the extraction; `publish.yml` reuses the same build at release. (What CI does *not* do is check the layer's contents — that is inline finding (1).)
- **`build_function.sh` needs no change.** mortie stays in `LAYER_PACKAGES` (L74) and is stripped from the function bundle; the function install list never pulls it, so layer-provided remains correct.

## Flagged, not a finding

`deployment/LAMBDA_DEPLOYMENT.md` L113 still documents a manual layer recipe with `h5coro==0.0.8 mortie` (unpinned). That drift is pre-existing and already wrong on h5coro (`1.0.5` today), so per CLAUDE.md §4 I'm flagging rather than folding it into this PR.

deployment/aws/build_layer.sh
espg: 🤖 *from Claude*

Fixed in 539de09 — added the post-install presence guard in the numpy-guard idiom, right after the h5coro/hidefix/mortie install:

```bash
if ! ls "$OUTPUT_DIR/python" | grep -qE "^mortie-"; then
    echo "ERROR: mortie ('$MORTIE_SPEC') missing from layer - check the spec in pyproject.toml!"; exit 1
fi
```

The comment above it names the exit-0 path you verified (a non-matching PEP 508 marker → "Ignoring mortie: markers ... don't match" → exit 0), so the reason the guard exists is on the page rather than in the thread. Verified both directions against a fake `$OUTPUT_DIR/python`: absent → error + exit 1, present (`mortie-0.9.1.dist-info`) → exit 0. The guard runs before the bloat/cache strip, so nothing has removed the dist-info by then, and it prints the derived spec in the failure so the next reader sees *what* pip was handed.

deployment/aws/build_layer.sh
espg: 🤖 *from Claude*

Fixed in 9224443, taking your option (2) — the test now knows about derived specs — because option (1) does not actually clear the trap: the parity check matches the *literal* `"mortie==<v>"` substring, and a derived spec never puts that string in the script no matter which table it was derived from. So deriving from `optional-dependencies.lambda` would leave the same red test with the same misleading message.

What landed keeps the failure (it is a real conflict — an exact pin in the extra genuinely would not reach the layer) but makes it self-explaining instead of reading as a sync bug:

```python
_DERIVED = {"mortie": "MORTIE_SPEC=$("}
...
assert not derived, (
    f"{derived} is pinned in the lambda extra, but build_layer.sh derives that "
    "spec from [project.dependencies] (issue #322), so the exact pin never "
    "reaches the layer. Either move the pin to [project.dependencies], or "
    "replace the derivation in build_layer.sh with a literal pin."
)
```

Plus a second test, `test_derived_specs_still_come_from_project_dependencies`, so the exemption cannot rot: it asserts the derivation fragment is still in the script (if it goes, `_DERIVED` must shrink and the pin check covers mortie again) and that `[project.dependencies]` still declares a mortie spec (otherwise `next(...)` in the script raises and the build dies on an empty spec).

Verified the trap message by temporarily inserting `"mortie==0.9.1"` into the `lambda` extra:

```
E  AssertionError: ['mortie==0.9.1'] is pinned in the lambda extra, but build_layer.sh
   derives that spec from [project.dependencies] (issue #322), so the exact pin never
   reaches the layer. ...
```

Reverted after; `TestLayerExtraParity` is 2 passed on the branch as pushed.

On the doc-accuracy note you raised at PR level — `deployment/LAMBDA_DEPLOYMENT.md:113` is inside "Option A: Build on Apple Silicon (manual)", a hand-rolled alternative recipe, not the command this commit changed. It is drifted well past mortie (`h5coro==0.0.8` vs the script's `1.0.5`, plus `pyarrow`/`boto3`/`earthaccess`, all since removed from the layer), so reconciling just its mortie token would leave a recipe that is still wrong. Flagging it for a follow-up rather than scope-creeping this small-fix PR.

deployment/aws/build_layer.sh
espg: 🤖 *from Claude*

Fixed in 21e2a08 — the comment now claims only what the change does, and says out loud that the layer still floats:

```
# mortie's version spec is single-sourced from pyproject.toml (issue #322):
# --no-deps means pip never resolves the runtime floor declared there, so an
# unpinned install silently accepted whatever was latest at build time -- even
# a release below the floor. Passing the declared spec (a) makes pip error and
# abort under `set -e` when the floor is ahead of what PyPI has published
# (pyproject L30-38 documents that state for h5coro-hidefix), and (b) leaves
# one edit site: a future floor bump or exact pin there reaches the layer with
# no second edit. The layer's mortie still floats to latest-above-floor, so
# this is floor enforcement, not pinning.
```

Both of your points are in it, the "a layer built before a floor bump" sentence is gone, and the last clause closes the misreading that the tree now determines the layer's mortie version.

deployment/aws/build_layer.sh
espg: 🤖 *from Claude*

Fixed in c792f48 — took your snippet verbatim in shape: single-quoted `-c` body, path in via `sys.argv[1]`, `tomllib.loads(pathlib.Path(sys.argv[1]).read_text())`. No shell interpolation reaches the Python source now, so the regex cannot be perturbed by requoting.

Verified end-to-end by extracting the new fragment (`sed -n '91,95p'`) into a harness with `PYTHON=python3.12`:

- normal worktree path → `SPEC=[mortie>=0.9.1]`, exit 0
- path containing a single quote (`.../it's dir/deployment/aws`) → `SPEC=[mortie>=0.9.1]`, exit 0 — the `SyntaxError` you reproduced is gone

`bash -n deployment/aws/build_layer.sh` clean.

Issue #265, opened 2026-07-17
HHDC (Hyper Height Data Cube) , Photon Flux


espg:
The main (current) funder for zagg development is NASA sliderule-- and sliderule has a primary customer: STV (surface topography vegetation). That community has recently been working on hyper height data cubes, or HHDC (see pdf references in local directory ~/goddard_lidar/ ). Our mandate with zagg is to enable generation of these HHDC's in a way that is computationally tractable and with reasonable cost. Have a look also at ~/waveform_benchmarks for additional helpful context; that folder and notebook derives from example input data provided on the diffusion modelers side (i.e., what they've been feeding to pytorch pipelines); the data itself is from https://github.com/Anfera/hhdcExample .

STV doesn't exist yet as a mission; it's a mission concept, and the community is 'mock'-ing the data that we'd see from such a mission by combining sensors that do exist: ICESat-2 , GEDI, etc. What these mission have in common is that they measure photon flux as a function of height: 

- ICESat-2 measures photon counts, at range bins; these are single photon detection events
- GEDI measures a waveform; this is energy spectrum (voltage over time) that's normalized to digitized laser waveform
- Airborne sensors typically will measure discrete 'points' as a point cloud

All of the measurements can be converted to and from-- in principle, although not by us (yet). We aren't as concerned with point clouds (since we're dealing with satellites), but they're still a good example of conversion: the airborne sensor 'senses' a waveform, and then uses thresholds to determine which 'peaks' in the waveform are work assigning 'points' to. So that a direct conversion, and moving back is simply a weighted convolution.

Similarly (and much more relevant for what we're doing) photon counts can be convolved in one of two ways: a.) convolution to a waveform, and, b.) binning to range bins (the reference ~/goddard_lidar/remotesensing-17-03588.pdf and others in that folder go over this). As a more general mathematical concept, we'd also like a third option: c.) a photon probability distribution.

For our ICESat-2 measurements, we're encoding them as tdigests, since this is a computationally tractable way to deal with the data. What we need, are a few functions to deal with this data:

  1. Something that can read and emit reconstructed original measurements; the ICESat-2 is sparse, and tdigest doesn't begin to create merged clusters until a threshold above that for a single laser pulse... so in places that don't have repeat coverage and/or crossovers, we in general should be able to grab the original photon profile within the gridcell (especially when we're encoding the optional morton location indices per photon).
[...]

PR #324, opened 2026-07-29
Located-flip prep: centroid-ancestors singleton fast path + provisional located target

espg:
Refs #265 (HHDC located-flip prep), refs #87 (located channel)

## What

Prepares the located-config flip discussed on #265: removes the per-photon-shaped Python loop from the located build path, and adds the provisional benchmark target that A/Bs the located channel's overhead against the live matrix before any flip.

### Phase 1 — singleton fast path for `_centroid_ancestors`

`_centroid_ancestors` reduced **every** centroid's member locations through `mortie.common_ancestor` in a Python loop. Below the compression knee every centroid is a singleton, so the loop ran once per *photon* — the same O(n) Python-loop shape issue #279 removed from the build path. Measured on the live CI store (`s3://sliderule-public/zagg-bench/tdigest_healpix_o9_hive.zarr`, pinned NEON shard): 99.56% of non-empty cells are all-singleton, so the located flip would have reintroduced ~3.3M loop iterations per shard on the agg phase.

Singletons now copy their sole member's word through vectorized (already the documented semantics: "a single member returns itself with its point kind preserved"); the loop covers only multi-member centroids (~0.4% of cells on the pinned shard). Measured **228×** on the all-singleton regime (100k centroids: 79 ms → 0.3 ms). The helper now takes `starts` as the ndarray `_compress` already produces (all three call sites dropped `.tolist()`).

`common_ancestor` also *validates* its input (the empty sentinel `0` — the configs' fill value — and any bad base-cell prefix raise `ValueError`), so the copy-through would have let invalid words pass silently for singleton centroids while still raising for multi-member ones. The fast path keeps the guarantee by validating the whole singleton set in one `mortie.validate_morton` kernel pass (array-shaped, ~8 ms per 200k words — negligible against the loop it replaces), so validity no longer depends on the compression regime.

### Phase 2 — provisional located target

`tdigest_healpix_o9_hive_located` in `provisional_targets` (PR-tree-only, runnable via `/benchmark --target`, per the issue #130 mechanism): byte-identical to the live sidecar arm except `location: leaf_id` on the ragged field, so the delta against `tdigest_healpix_o9_hive_sidecar` on the same pinned shard isolates the located channel's overhead **in the metrics the harness records** — wall/runtime, `gb_seconds`/cost, the worker phase split, peak memory, and store object counts. `bench_metrics.build_record` has no store-byte column, so bytes are *not* part of the series; per-object sizes stay observable out-of-band via S3 LIST. For reference (not measurement): the `h_tdigest` payload array is unchanged and the location words ride a separate **equal-width** vlen sibling (one `uint64` per centroid vs the `float32` `(k, 2)` payload's 8 B), so total ragged bytes are ~2× plus vlen framing. Agg CPU is covered by phase 1. Caveat recorded in the target's `_comment`: a located config is non-mergeable (`validate_streaming` rejects located ragged → `SpillAggregator._mergeable = False`), so the located arm is single-block-spill only where the unlocated arm folds across blocks — not reachable on the pinned o9 NEON shard, but a real asymmetry on heavier shards. The expected-object model derives from the template spec (`bench_objects._member_layouts`), so the `h_tdigest_locations` companion array auto-joins the tripwire counts — no harness change needed.

## Phases

- [x] 1. Singleton fast path in `_centroid_ancestors` + equivalence tests
- [x] 2. Provisional located config + target entry

## How it was tested

- `pytest tests/test_tdigest.py` → 98 passed (4 new tests: fast path vs the all-loop reference on mixed singleton/multi-member partitions — including a below-order-29 area word spliced in as a singleton member, the mixed-order input `merge_tdigests` feeds the helper — all-singleton verbatim round-trip, end-to-end build equivalence across the compression knee, and a singleton fill-value word still raising).
- `pytest tests/test_benchmark.py tests/test_benchmark_objects.py tests/test_benchmark_shardmap.py` → green (targets.json + new config parse; expected-count model).
- New config validated via `zagg.config.load_config` (resolves `location: leaf_id`).
- `ruff check` / `ruff format --check` clean on touched files.
- Full suite: 2649 passed; **3 pre-existing failures, flagged not fixed** (CLAUDE.md §4): `tests/test_runner.py::TestConsolidationGate` (2, reproduce on clean `main` at `d104dbb` with this branch's changes stashed) and `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` (build-environment test, present in the pre-branch lastfailed cache).

## Questions for review

- The provisional target pins `index_backend: sidecar` (the watched line per #272). If you want the A/B against the inline arm too, it's one more provisional entry — held off to keep the on-demand surface minimal.


tests/data/benchmark/targets.json
```diff
@@ -84,6 +84,17 @@
...
+    "tdigest_healpix_o9_hive_located": {
+      "_comment": "Located-channel A/B probe (issues #265/#87): byte-identical to the live sidecar arm except ``location: leaf_id`` on the ragged field, so the delta vs tdigest_healpix_o9_hive_sidecar on the same pinned shard isolates the located channel's wall/cost/byte overhead (expected: ragged payload roughly 2-3x raw before zstd; agg CPU covered by the _centroid_ancestors singleton fast path). Run on demand via /benchmark --target ahead of any live-matrix flip to the located config; deliberately NOT in the every-merge matrix until that flip is decided.",
```
espg: 🤖 *from Claude (review)*

**non-blocking (doc accuracy, committed file).** The `_comment` claims the delta vs `tdigest_healpix_o9_hive_sidecar` "isolates the located channel's wall/cost/**byte** overhead" (and the PR body repeats it), but the harness records no store-byte metric. `bench_metrics.build_record` emits `runtime_s`, `gb_seconds`, `cost_per_shard_usd`, `ephemeral_cost_usd`, the phase split, `max_memory_mb`, and `objects_total`/`objects_expected`/`objects_per_shard`/`objects_mismatch` — there is no size/`ContentLength` column anywhere in `.github/scripts/` (`grep -rn "store_bytes\|total_bytes\|ContentLength\|Size" .github/scripts/*.py` → no hits). So this A/B measures wall + cost + **object count**, and the byte delta has to be read off the store by hand afterwards.

Two ways out: drop "byte" from the claim and say object-count + wall/cost, or say explicitly that the byte delta is an out-of-band `s3 ls` on the two stores.

Secondary, same sentence: "ragged payload roughly 2-3x raw before zstd". The payload array is untouched — the locations ride a *separate* `h_tdigest_locations` vlen array (8 B/centroid beside the payload's 2×float32 = 8 B/centroid), so the *payload* is 1× and total ragged bytes ≈ 2× plus per-cell vlen framing. Worth phrasing as "total ragged bytes ~2x (a second, equal-width vlen array)" so the expectation the run is checked against is the right one.

tests/data/benchmark/targets.json
```diff
@@ -84,6 +84,17 @@
...
+      "config": "configs/atl03_tdigest_healpix_o9_hive_located.yaml",
+      "shardmap": "healpix_o9",
+      "aggregator": "tdigest",
+      "grid_type": "healpix",
+      "grid_size": "o9",
+      "index_backend": "sidecar",
+      "streaming_mode": "spill",
```
espg, line 95: 🤖 *from Claude (review)*

**non-blocking (A/B validity caveat worth recording in the `_comment`).** The two arms are *not* interchangeable under a spill block-threshold crossing, so "byte-identical to the live sidecar arm except `location: leaf_id`" understates one behavioral difference.

`SpillAggregator.__init__` derives mergeability from the streaming validator:

```python
try:
    validate_streaming(config)
    self._mergeable = True
except ValueError:
    self._mergeable = False
```

and `validate_streaming` rejects located ragged outright (`src/zagg/processing/streaming.py:109-116`, "located ragged fields (location: ...) cannot stream yet"). So for this config `_mergeable = False`, and `_close_block` raises:

```python
if not self._mergeable:
    raise SpillOverflowError(
        f"spill block hit the {self.block_bytes:,}-byte threshold but the "
        f"config carries reducers with no merge law, ...")
```

i.e. the located arm is **single-block-exact only**, while `tdigest_healpix_o9_hive_sidecar` folds across blocks fine. On the pinned o9 NEON shard at `memory 4096 + extra_disk` the block threshold (`_default_block_bytes` → the 0.45×free-`/tmp` cap, several GB) is far above the shard's spilled column volume, so this should not fire in practice — but it means (a) the arm cannot be reused as-is on a heavy shard (88S, CONUS) without changing meaning, and (b) if it ever does fire the failure reads as "reducers with no merge law", which is misleading for a t-digest whose reducer *does* have a merge law (it is the location channel that cannot cross a block). One sentence in the `_comment` recording "single-block spill only; a block crossing aborts (located is non-mergeable)" would keep the next reader out of that trap.

Related latent hazard, not this PR's to fix: `SpillAggregator._chunk_outputs_merged` returns `{}` as its `ragged_locations` element (`src/zagg/processing/spill.py:777`), and `worker.py` then falls back to the 2-tuple ragged contract (`name in ragged_locs` is False), which makes `write.py` skip the `{field}_locations` array entirely (`_ragged_entry` → `locations_list is None`). That silent-drop path is only unreachable *because* of the `_mergeable` gate above — if the gate is ever loosened for located fields, locations vanish silently rather than raising.

src/zagg/stats/tdigest.py
```diff
@@ -173,13 +173,22 @@ def _centroid_ancestors(locations: np.ndarray, starts: list[int], n: int) -> np.
...
     from mortie import common_ancestor
 
-    bounds = [*starts, n]
+    starts = np.asarray(starts, dtype=np.int64)
+    sizes = np.diff(starts, append=n)
     out = np.empty(len(starts), dtype=np.uint64)
-    for j in range(len(starts)):
-        out[j] = common_ancestor(locations[bounds[j] : bounds[j + 1]])
+    single = sizes == 1
+    out[single] = locations[starts[single]]
```
espg: 🤖 *from Claude (review)*

**non-blocking (behavioral drift the equivalence tests cannot see).** The copy-through is value-identical to `common_ancestor` for every *valid* singleton word — I verified against mortie 0.9.1 that a lone word is returned verbatim for order-29 point words and for area words at orders 0/1/4/10/18/25/28/29 (matching the documented "Only a single-element input is returned unchanged"). But `common_ancestor` also **validated** the word, and the fast path drops that:

```
>>> common_ancestor(np.array([0], dtype=np.uint64))
ValueError: moc_min: input contains an empty or invalid word (0)
```

(same for 1, 2, … — any non-word bit pattern). After this change an invalid/empty word (0 is the natural one — it is the coordinate `fill_value` throughout the benchmark configs) **silently propagates** into the `h_tdigest_locations` output when its centroid is a singleton, and still raises when the same word lands in a multi-member centroid. So the error behavior now depends on the compression regime: garbage passes through below the knee and aborts the shard above it, which is a nasty debugging asymmetry.

It is not purely hypothetical: `location` need not be `leaf_id` — `config.py:489` accepts any `data_source.variables` column (`if location != "leaf_id" and location not in ds_vars: raise`), and `build_tdigest` only validates the *dtype* (uint64), never word validity. A mis-declared uint64 data column therefore reaches here.

Cheapest fix that keeps the 160-228× win: validate once, vectorized, instead of per centroid — e.g. reject `locations == 0` (or the mortie-invalid predicate) once at the top of `_centroid_ancestors`, which is O(n) numpy and costs nothing next to the old loop. At minimum, note in the docstring that singleton words are no longer validated.

tests/test_tdigest.py
```diff
@@ -255,6 +255,52 @@ def test_without_locations_returns_bare_array(self):
...
+class TestCentroidAncestorsFastPath:
+    """The singleton fast path matches the all-loop reference (issue #265)."""
+
+    def test_mixed_singleton_and_merged_partitions(self):
+        from zagg.stats.tdigest import _centroid_ancestors
+
+        locs = _point_words(20, seed=6)
+        # Partition mixing singletons with 2-, 3- and 4-member centroids,
+        # including singleton runs at both ends.
+        starts = np.array([0, 1, 3, 4, 5, 8, 12, 16, 17, 18, 19], dtype=np.int64)
```
espg, line 287: 🤖 *from Claude (review)*

**nit / test gap.** Every word in all three new tests comes from `conftest.point_words`, i.e. order-29 **point** words. But the case the `_centroid_ancestors` docstring specifically calls out — "Mixed-order input is fine (a below-order-29 mean-morton from a prior merge folds with fresh order-29 points)" — is exactly the one the merge call sites feed: `merge_tdigests` / `merge_tdigests_kway` pass `combined_locs` built from *build-side* location vectors, whose multi-obs entries are already collapsed **area** words. A singleton centroid in a merge therefore copies an area word through the new fast path, and nothing in this class pins that.

I checked it holds today (mortie 0.9.1 returns a lone area word verbatim at every order I tried, kind bits preserved), so this is a coverage note, not a bug. One extra member in the existing mixed partition closes it, e.g. splice `common_ancestor(locs[:4])` (or a `clip2order` word) in as one of the singleton slots and keep the same reference comparison — then the equivalence test covers the mixed-order singleton the docstring promises, and a future mortie change to singleton normalization fails loudly instead of silently diverging from the copy-through.

src/zagg/stats/tdigest.py
espg: 🤖 *from Claude*

Fixed in 2947163. You are right that the copy-through dropped `common_ancestor`'s validation, and the regime-dependence is the real bug: an invalid word would raise for multi-member centroids and pass silently for singletons.

Checked mortie 0.9.0 for a cheap public predicate before falling back to a zero-check, and there is one: `mortie.validate_morton` looks scalar in its docstring but is array-shaped by construction — it does `np.atleast_1d(np.asarray(morton, dtype=np.uint64))` and pushes the whole array through `_rust_mort2nested`, which raises `ValueError` on **any** invalid element (`tools.py:218`). One call validates the entire singleton set:

```python
single = sizes == 1
singles = locations[starts[single]]
if singles.size:
    validate_morton(singles)
out[single] = singles
```

So this restores *full* mortie validity, not just fill-value/zero detection — which matters, because a zero-check alone would **not** have been honest here. Probing mortie directly: `0`, `1`, `2`, `13`, and `0xFFFFFFFFFFFFFFFF` all raise (`Morton index cannot be zero` from the kernel decode, whatever the word), while `1<<63` and `1<<62` are valid. Malformed nonzero words are rejected by mortie and would have sailed through a zero-check.

Cost is negligible against the loop it replaces: 200k words validate in ~8 ms in one Rust pass, vs the ~79 ms/100k Python `common_ancestor` loop the fast path removes.

Regression test added (`test_singleton_fill_value_word_still_raises`): an all-singleton partition with `locs[2] = 0` raises `ValueError`.

tests/test_tdigest.py
espg: 🤖 *from Claude*

Fixed in 32590c8. `test_mixed_singleton_and_merged_partitions` now splices a below-order-29 area word in as the singleton member at index 4 (a singleton under that partition, `starts` 4 → 5) — exactly the mixed-order input `merge_tdigests` hands the helper:

```python
area = np.uint64(common_ancestor(locs[:4]))
assert infer_order_from_morton(int(area)) < 29
locs[4] = area
...
assert got[starts.tolist().index(4)] == area
```

The `infer_order_from_morton(...) < 29` assert pins that it really is an area word (so the case cannot silently degrade into another point-word test), and the verbatim-copy assert is checked alongside the existing full comparison against the all-loop reference.

tests/data/benchmark/targets.json
espg: 🤖 *from Claude*

Fixed in 65df320 (targets.json `_comment`) plus the matching PR-body sentence. Both points confirmed:

1. **No byte metric.** `bench_metrics.build_record` (`.github/scripts/bench_metrics.py:236-297`) emits runtime/wall/`gb_seconds`/cost/phase-split/`max_memory_mb` and the `objects_total` / `objects_expected` / `objects_per_shard` / `objects_mismatch` object columns — no store-byte column anywhere. The `_comment` now claims only what the harness records, and notes per-object sizes stay observable out-of-band via S3 LIST rather than in the series.
2. **Byte framing corrected.** The `h_tdigest` payload array is unchanged; the location words ride the separate `h_tdigest_locations` uint64 vlen sibling (`write.py:625-663`), and the payload is `dtype: float32` with `inner_shape: [2]` in the new config — 8 B/centroid on both sides, so equal-width and total ragged bytes ~2× plus vlen framing, not "payload 2-3x raw".

New text: "...isolates the located channel's overhead in the metrics the harness actually records -- wall/runtime, gb_seconds/cost, worker phase split, peak memory, and store object counts (bench_metrics.build_record has NO store-byte column; per-object sizes are observable out-of-band via S3 LIST). Byte framing, for reference rather than measurement: the h_tdigest payload array is unchanged and the location words ride a separate equal-width vlen sibling (uint64 per centroid vs the float32 (k, 2) payload's 8 bytes), so total ragged bytes are ~2x plus vlen framing."

tests/data/benchmark/targets.json
espg: 🤖 *from Claude*

Fixed in 8e25b37 — claim verified in the code before folding, and it holds:

- `validate_streaming` (`src/zagg/processing/streaming.py:93`) appends a problem for any ragged field with `sig["location"] is not None` ("located ragged fields ... cannot stream yet"), so a located config raises `ValueError`.
- `SpillAggregator.__init__` (`src/zagg/processing/spill.py:432-439`) sets `_mergeable` by try/except around that same validator → `False` for a located config.
- `_close_block` (`spill.py:580`) raises `SpillOverflowError` when `not self._mergeable`, so the arm is exact single-block only and aborts at the first crossing.

The target `_comment` now records it: "A/B asymmetry to keep in mind: a located config is NOT mergeable (validate_streaming rejects located ragged fields, so SpillAggregator._mergeable is False), which makes this arm single-block-spill only -- a block crossing raises SpillOverflowError where the unlocated sidecar arm folds block-by-block. Not reachable on this pinned o9 NEON shard (spill volume sits far below the block threshold), so the A/B is apples-to-apples here, but the two arms are NOT comparable on a heavier shard until the streaming state threads per-cell locations (issue #87)."

Same caveat added to the PR body's phase-2 paragraph so it is visible without opening the JSON.

PR #329, opened 2026-07-29
fix LAMBDA_DEPLOYMENT.md layer build drift (issue #325)

### LINKED ISSUE #325: LAMBDA_DEPLOYMENT.md Option A (Apple Silicon manual build) has drifted from build_layer.sh
🤖 *from Claude*

Surfaced during the PR #323 review-fold ([context](https://github.com/englacial/zagg/pull/323#discussion_r3678347578)): the manual-build recipe in `deployment/LAMBDA_DEPLOYMENT.md` ("Option A: Build on Apple Silicon (manual)", starting at line 98) has drifted far from what `deployment/aws/build_layer.sh` actually installs, and would produce a layer that neither matches production nor imports correctly:

- `h5coro==0.0.8` vs the script's `h5coro==1.0.5`
- `pyarrow` — removed from the layer entirely (issue #130; the worker's Arrow carrier is `arro3-core`, which the recipe omits)
- `boto3` — the script explicitly strips it (Lambda runtime provides it)
- `earthaccess`, `astropy`, `requests` — not in the layer build at all
- missing everything added since: `h5coro-hidefix`, `xarray`/`h5netcdf`/`h5py` (issue #220), `async-tiff`/`obspec` (issue #218), `fastparquet`/`cramjam` pins, the numpy `<2.3` constraint and arm64 page-alignment source build, the 250 MB gate, and (post PR #323) the pyproject-derived mortie spec
- the recipe also strips `*.dist-info`, which `build_layer.sh` does not (and which breaks `importlib.metadata` consumers)

Fixing one token would leave a still-wrong recipe, so the ask is a reconciliation with a decision first:

**(a)** Rewrite Option A to mirror `build_layer.sh` — but it duplicates a script that already runs fine in the manylinux container via Docker on Apple Silicon (Option B territory), so it will drift again; or
**(b)** Delete Option A and point at the container path (`docker run … quay.io/pypa/manylinux_2_28_aarch64 … ./build_layer.sh arm64`, exactly what `lambda-build-reusable.yml` does), keeping one normative build path.

Lean: (b) — the script is the single source of truth for layer contents (reinforced by PR #323's pyproject-derived mortie spec), and hand-maintained parallel recipes are how this drift happened.

espg:
Closes #325

Docs-only PR: reconciles `deployment/LAMBDA_DEPLOYMENT.md` with what `deployment/aws/build_layer.sh` / `build_function.sh` actually do, per the ruling on the issue — **option (b)**, delete the hand-maintained recipe and keep one normative containerized build path, **podman commands first, docker equivalents second** ([espg's approval + podman-first amendment](https://github.com/englacial/zagg/issues/325#issuecomment-5124137711)). No scripts or workflows are touched.

## Checklist
- [x] #325 — replace drifted "Option A" manual layer recipe with the containerized `build_layer.sh` path; fix the other statements the scripts contradict

## What changed
- **"Rebuilding the ARM64 Layer" → "Rebuilding the Layer"**: the drifted "Option A: Build on Apple Silicon (manual)" pip recipe (`h5coro==0.0.8`, `pyarrow`/`boto3`/`earthaccess`/`astropy`, blanket `*.dist-info` strip, hardcoded layer ARN) is deleted, along with the equally-historical "Option B" (macOS-runner pip cross-compile "key findings", contradicted by CI's arch-matched ubuntu runners) and "Option C" (x86_64 cross-compile "current approach"). Replaced by a single normative section: run `./build_layer.sh {arm64,x86_64}` inside the arch-matched `quay.io/pypa/manylinux_2_28_{aarch64,x86_64}` image — podman first, docker equivalent after. The commands mirror the CI invocation verbatim (mount → `/workspace`, `-w /workspace/deployment/aws`, `yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh <arch>`): [`lambda-build-reusable.yml` L27–35 (x86_64)](https://github.com/englacial/zagg/blob/main/.github/workflows/lambda-build-reusable.yml#L27-L35) and [L108–116 (arm64)](https://github.com/englacial/zagg/blob/main/.github/workflows/lambda-build-reusable.yml#L108-L116). Podman/docker differ only in the binary name here; a note covers the `:z` SELinux mount label needed on SELinux-enforcing Linux hosts (not on macOS `podman machine`) and that x86_64 is emulated on Apple Silicon.
- **"What's in the layer vs function code"**: layer package list updated to what `build_layer.sh` installs today (adds `arro3-core`, `xarray`/`h5netcdf`/`h5py`, `h5coro-hidefix`, `async-tiff`/`obspec`; notes `boto3` is explicitly stripped and `pyarrow` → `arro3-core` per issue #130). "Current Config" layer line now defers to the script instead of pinning `h5coro==0.0.8`.
- **"Deploying Updated Function Code"**: the manual function-zip recipe stripped *all* `*.dist-info`, which `build_function.sh` explicitly avoids for `zarr`/`pydantic_zarr` (stripping breaks `importlib.metadata` → `PackageNotFoundError` in `ArraySpec.from_zarr` — see [`build_function.sh` L101–106](https://github.com/englacial/zagg/blob/main/deployment/aws/build_function.sh#L101-L106)); same drift class as Option A, so it now points at the script, keeping only the `aws lambda update-function-code` deploy step.
- **"Build Infrastructure"**: script blurb now says manylinux container via podman/docker and cross-references the rebuild section.

Left alone: the Size Budget table and other historical figures already disclaimed by the header note ("trust the scripts ... over the numbers here"), and the "CI/CD Workflow Design" section, which is explicitly marked historical.

## How tested
Docs-only. `uvx codespell deployment/LAMBDA_DEPLOYMENT.md` is clean; the podman/docker commands were checked token-for-token against the workflow steps linked above (image names, mount target, workdir, and the `yum install zip` + `chmod` preamble `build_layer.sh` relies on when run from a fresh checkout inside the container).

## Questions for review
- The x86_64-on-Apple-Silicon note says emulation (Rosetta/qemu in the podman machine) is required — that matches the script's `uname -m` arch guard, but I couldn't execute a container build here; flag if the local podman-machine setup differs.

### DIFF
--- a/deployment/LAMBDA_DEPLOYMENT.md
+++ b/deployment/LAMBDA_DEPLOYMENT.md
@@ -81,8 +81,9 @@ the prompt). See
 [docs/deployment/lambda.md](../docs/deployment/lambda.md) for the parameter
 table and overrides.
 
-`deploy.sh` (below) is the maintainer path for *in-place updates* to an
-already-deployed function and does not create the role/function/bucket.
+`deploy.sh` is the maintainer path for *in-place updates* to an already-deployed
+function and does not create the role/function/bucket — see
+[Rebuilding the Layer](#rebuilding-the-layer) below for what it consumes.
 
 ---
 
@@ -144,8 +145,19 @@ The zip lands in `deployment/layers/lambda_layer_<arch>.zip`. The script
 enforces the 250 MB unzipped limit itself; CI additionally gates the combined
 layer + function size. CI builds both arches this same way on arch-matched
 runners (`lambda-build.yml` → `lambda-build-reusable.yml`), and release zips
-come from `publish.yml` calling the same reusable workflow — deploys consume
-those via `stand_up.sh` (above).
+come from `publish.yml` calling the same reusable workflow.
+
+**What consumes the zip:** nothing in the repo deploys
+`deployment/layers/lambda_layer_<arch>.zip` — it is for local verification and
+size-checking. To actually put a layer on a function:
+
+- **in-place maintainer update** — `deployment/aws/deploy.sh`. Note it pulls the
+  **CI artifact**, not your local build (`gh run download --name
+  "lambda-layer-${ARCH}"`), then `publish-layer-version` (staging through S3
+  when the zip is >50 MB) and `update-function-configuration`.
+- **full standup, or a fresh account/region** — cut a release and use
+  `stand_up.sh`, which deploys the release zips staged on the distribution
+  bucket (above), not local ones.
 
 ---
 
--- a/deployment/LAMBDA_DEPLOYMENT.md
+++ b/deployment/LAMBDA_DEPLOYMENT.md
@@ -21,7 +21,7 @@ x86_64 / py3.12 is available for local/testing parity.
 ### Current Config
 - **Runtime**: python3.12
 - **Architecture**: arm64 (default; x86_64 also supported)
-- **Layer**: `zagg-deps-{arch}` (py3.12, pyproj/odc-geo for rectilinear grids, h5coro==0.0.8)
+- **Layer**: `zagg-deps-{arch}` (py3.12; contents defined by `build_layer.sh` — see below)
 - **Function code**: `lambda_handler.py` + `zagg/` package + obstore/zarr/pydantic-zarr/pyyaml
 - **Role**: created by `template.yaml` (CloudFormation-auto-named; the template
   sets no `RoleName`), scoped least-privilege to the `OutputBucketName` bucket
@@ -33,11 +33,12 @@ x86_64 / py3.12 is available for local/testing parity.
 
 ### What's in the layer vs function code
 
-**Layer** (built by `build_layer.sh`):
-numpy, pandas, fastparquet, cramjam, shapely, pyproj, odc-geo, affine, cachetools,
-h5coro, mortie, and their transitive deps. (Corrected: `earthaccess` is
-orchestrator-only and **not** in the layer; `boto3` is provided by the Lambda
-runtime; `astropy`/`s3fs`/`pyarrow` are not installed by `build_layer.sh`. The old
+**Layer** (built by `build_layer.sh` — the script is the single source of truth):
+numpy, pandas, arro3-core, fastparquet, cramjam, xarray, h5netcdf, h5py, shapely,
+pyproj, odc-geo, affine, cachetools, h5coro, h5coro-hidefix, mortie, async-tiff,
+obspec, and their transitive deps. (`earthaccess` is orchestrator-only and
+**not** in the layer; `boto3` is provided by the Lambda runtime and explicitly
+stripped; `pyarrow` was replaced by `arro3-core` — issue #130. The old
 `xagg-dependencies:1`/222MB figure is historical.)
 
 **Function code** (built by `build_function.sh`):
@@ -83,124 +84,79 @@ already-deployed function and does not create the role/function/bucket.
 
 ---
 
-## Rebuilding the ARM64 Layer
+## Rebuilding the Layer
 
-### Why
+### Why arm64
 ARM64 Lambda is 20% cheaper ($0.0000133334 vs $0.0000166667 per GB-second). At ~90,000
 GB-seconds per full run, this saves ~$0.60/run. Over many runs it adds up.
 
-### What needs to happen
+### The build (containerized — the one normative path)
 
-1. Build a new layer with all deps compiled for `manylinux2014_aarch64` + `cp312`
-2. The layer must include the same packages as `xagg-dependencies:1` but for ARM64/py3.12
-3. Deploy the function with architecture `arm64` and runtime `python3.12`
+`deployment/aws/build_layer.sh` is the single source of truth for layer
+contents (package set, pins, numpy page-alignment build, bloat strip, 250 MB
+gate). Do not hand-maintain a parallel pip recipe — the script must run inside
+an arch-matched `manylinux_2_28` container (cp312), exactly as CI does in
+`.github/workflows/lambda-build-reusable.yml`.
 
-### Option A: Build on Apple Silicon (manual)
-
-On an Apple Silicon Mac:
+With podman (what we use locally), from the repo root:
 
 ```bash
-# Create build directory
-mkdir -p /tmp/layer_build/python
+# arm64 (native on Apple Silicon)
+podman run --rm \
+  -v "$(pwd)":/workspace \
+  -w /workspace/deployment/aws \
+  quay.io/pypa/manylinux_2_28_aarch64 \
+  bash -c "yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh arm64"
 
-# Install deps targeting Lambda's manylinux environment
-pip install \
-  --platform manylinux2014_aarch64 \
-  --target /tmp/layer_build/python \
-  --implementation cp \
-  --python-version 3.12 \
-  --only-binary=:all: \
-  numpy==2.2.6 pandas==2.2.3 h5coro==0.0.8 mortie earthaccess \
-  boto3 fastparquet pyarrow shapely pyproj odc-geo cramjam astropy requests
-
-# Trim bloat
-find /tmp/layer_build -type d -name '__pycache__' -exec rm -rf {} +
-find /tmp/layer_build -type d -name '*.dist-info' -exec rm -rf {} +
-find /tmp/layer_build -type d -name 'tests' -exec rm -rf {} +
-
-# Check size (must be <250MB unzipped when combined with function code)
-du -sh /tmp/layer_build/
-
-# Zip
-cd /tmp/layer_build && zip -qr /tmp/lambda_layer_arm64.zip python/
-
-# Publish
-aws lambda publish-layer-version \
-  --layer-name zagg-deps-arm64 \
-  --compatible-runtimes python3.12 \
-  --compatible-architectures arm64 \
-  --zip-file fileb:///tmp/lambda_layer_arm64.zip \
-  --region us-west-2
-
-# Update function
-aws lambda update-function-configuration \
-  --function-name process-shard \
-  --runtime python3.12 \
-  --layers "arn:aws:lambda:us-west-2:429435741471:layer:zagg-deps-arm64:1" \
-  --region us-west-2
-
-# Then update code with arm64 arch
-aws lambda update-function-code \
-  --function-name process-shard \
-  --zip-file fileb:///tmp/lambda_function.zip \
-  --architectures arm64 \
-  --region us-west-2
+# x86_64 (on Apple Silicon this is emulated — needs Rosetta/qemu in the podman machine)
+podman run --rm \
+  -v "$(pwd)":/workspace \
+  -w /workspace/deployment/aws \
+  quay.io/pypa/manylinux_2_28_x86_64 \
+  bash -c "yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh x86_64"
 ```
 
-### Option B: CI/CD on GitHub Actions with macOS Apple Silicon (recommended)
+Docker equivalents are identical apart from the binary name:
 
-GitHub provides free macOS Apple Silicon runners for public repos. This is the best
-option because Linux ARM64 runners have had issues building some of our deps.
-
-Key findings:
-- `macos-15` runners use M1 Apple Silicon (ARM64), 3 CPUs, 7GB RAM
-- Free for public repos (englacial/zagg is public), $0.062/min for private
-- Docker is NOT available on macOS ARM64 runners (Apple Virtualization limitation)
-- `pip install --platform manylinux2014_aarch64 --only-binary=:all:` works from macOS
-  to cross-compile Lambda layers — no Docker needed
-- All our deps have pre-built `manylinux2014_aarch64` wheels on PyPI
-
-The layer + function are now built in CI by `.github/workflows/lambda-build.yml` (both arches, with the combined 250MB size gate); the originally-planned `deploy-lambda.yml` was never added. The manual/CI options below are historical design notes.
-
-### Option C: Cross-compile from x86_64 (current approach for testing)
-
-Works for packages with pre-built `manylinux_2_17_x86_64` wheels. Download with:
 ```bash
-pip download --python-version 311 --platform manylinux_2_17_x86_64 \
-  --only-binary :all: --no-deps <package> -d /tmp/wheels
+docker run --rm \
+  -v "$(pwd)":/workspace \
+  -w /workspace/deployment/aws \
+  quay.io/pypa/manylinux_2_28_aarch64 \
+  bash -c "yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh arm64"
 ```
-Then unzip wheels into the build directory. This is what we're doing now for testing.
+
+On an SELinux-enforcing Linux host, add `:z` to the podman volume mount
+(`-v "$(pwd)":/workspace:z`) so the container can write the zip back; it is
+not needed on macOS (`podman machine`).
+
+The zip lands in `deployment/layers/lambda_layer_<arch>.zip`. The script
+enforces the 250 MB unzipped limit itself; CI additionally gates the combined
+layer + function size. CI builds both arches this same way on arch-matched
+runners (`lambda-build.yml` → `lambda-build-reusable.yml`), and release zips
+come from `publish.yml` calling the same reusable workflow — deploys consume
+those via `stand_up.sh` (above).
 
 ---
 
 ## Deploying Updated Function Code (no layer change)
 
-When only `lambda_handler.py` or `zagg/` package code changes (no new deps):
+When only `lambda_handler.py` or `zagg/` package code changes (no new deps),
+build with the script (it owns the function dep list, the layer-overlap strip,
+and the dist-info handling — zarr/pydantic_zarr keep theirs, stripping them
+breaks `importlib.metadata`):
 
 ```bash
-# Build zip
-rm -rf /tmp/lambda_build && mkdir -p /tmp/lambda_build
-cp deployment/aws/lambda_handler.py /tmp/lambda_build/
-cp -r src/zagg /tmp/lambda_build/zagg
-
-# Add deps not in layer (skip native ones if already unpacked)
-pip install --target /tmp/lambda_build --no-deps \
-  zarr pydantic-zarr pyyaml pydantic typeguard typing_inspect annotated-types
-
-# For obstore (native): download correct wheel and unzip
-pip download --python-version <VER> --platform <PLAT> --only-binary :all: \
-  --no-deps obstore -d /tmp/wheels
-unzip -qo /tmp/wheels/obstore-*.whl -d /tmp/lambda_build
-
-# Clean and zip
-find /tmp/lambda_build -type d -name '*.dist-info' -exec rm -rf {} +
-find /tmp/lambda_build -type d -name '__pycache__' -exec rm -rf {} +
-cd /tmp/lambda_build && zip -qr /tmp/lambda_function.zip .
+# Build zip (auto-detects arch and Python). Native deps (obstore) resolve for
+# the host, so run on Linux with the function's arch/Python — CI uses
+# arch-matched ubuntu runners; on a Mac, run inside a Linux container as with
+# the layer build.
+deployment/aws/build_function.sh
 
 # Deploy
 aws lambda update-function-code \
   --function-name process-shard \
-  --zip-file fileb:///tmp/lambda_function.zip \
+  --zip-file fileb://deployment/builds/lambda_function_arm64_py312.zip \
   --region us-west-2
 ```
 
@@ -249,7 +205,7 @@ deps from the layer into the function code (or vice versa).
 ## Build Infrastructure
 
 ### Scripts
-- `deployment/aws/build_layer.sh [x86_64|arm64]` — Lambda layer build (runs in an arch-matched Docker container)
+- `deployment/aws/build_layer.sh [x86_64|arm64]` — Lambda layer build (runs in an arch-matched manylinux container via podman/docker — see "Rebuilding the Layer" above)
 - `deployment/aws/build_function.sh` — function code build (handler + zagg + non-layer deps)
 
 ### CI/CD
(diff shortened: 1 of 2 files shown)


deployment/LAMBDA_DEPLOYMENT.md
```diff
@@ -33,11 +33,12 @@ x86_64 / py3.12 is available for local/testing parity.
 
 ### What's in the layer vs function code
 
-**Layer** (built by `build_layer.sh`):
-numpy, pandas, fastparquet, cramjam, shapely, pyproj, odc-geo, affine, cachetools,
-h5coro, mortie, and their transitive deps. (Corrected: `earthaccess` is
-orchestrator-only and **not** in the layer; `boto3` is provided by the Lambda
-runtime; `astropy`/`s3fs`/`pyarrow` are not installed by `build_layer.sh`. The old
+**Layer** (built by `build_layer.sh` — the script is the single source of truth):
```
espg: 🤖 *from Claude (review)*

**non-blocking** — "the script is the single source of truth" is a new sweeping claim that the script itself contradicts, which is the drift class this PR exists to remove. `build_layer.sh` derives mortie's spec *from* `pyproject.toml` (L91-99, issue #322 / PR #323 — `deps = tomllib.loads(...)["project"]["dependencies"]`, then a hard assert that a `mortie-` dir landed), and it says "Keep the pin(s) in sync with the `lambda` extra in pyproject.toml" four separate times (L64, L69, L80, L109) for arro3-core, xarray/h5netcdf/h5py, h5coro-hidefix and async-tiff. So pins are co-owned with the `lambda` extra, and mortie's floor is *owned* by `pyproject.toml`. Suggested wording: "the script is the normative build entry point; pins are kept in sync with the `lambda` extra in `pyproject.toml`, and mortie's spec is read from it".

deployment/LAMBDA_DEPLOYMENT.md
```diff
@@ -83,124 +84,79 @@ already-deployed function and does not create the role/function/bucket.
...
+`deployment/aws/build_layer.sh` is the single source of truth for layer
+contents (package set, pins, numpy page-alignment build, bloat strip, 250 MB
+gate). Do not hand-maintain a parallel pip recipe — the script must run inside
+an arch-matched `manylinux_2_28` container (cp312), exactly as CI does in
+`.github/workflows/lambda-build-reusable.yml`.
 
-### Option A: Build on Apple Silicon (manual)
-
-On an Apple Silicon Mac:
+With podman (what we use locally), from the repo root:
```
espg: 🤖 *from Claude (review)*

**non-blocking** — "the one normative path" is true only inside this file, and the surviving competitor is in the docs this file's own header defers to as canonical. `docs/deployment/arm64.md` still:

- requires **Docker Desktop** as the build environment (L21-25, `brew install --cask docker` at L37-41) — docker-first, i.e. the opposite of the podman-first ruling this PR implements;
- presents the build as a bare host command with **no container wrapper at all** (L43-47: `bash deployment/aws/build_layer.sh arm64`), which is exactly the "pip recipe on the Mac" drift being deleted here;
- verifies the built layer by importing `earthaccess`, `zarr`, `obstore`, `pydantic_zarr` out of it (L78-91) — none of which `build_layer.sh` installs into the layer (earthaccess is orchestrator-only; the other three are function-code deps, stripped from the layer by the `LAYER_PACKAGES` logic on the other side).

Editing a second doc is a scope change, so I'm flagging rather than asking for it silently: either extend this PR to `arm64.md` (a maintainer call) or leave a follow-up note, otherwise the canonical page keeps contradicting the new normative section.

deployment/LAMBDA_DEPLOYMENT.md
```diff
@@ -83,124 +84,79 @@ already-deployed function and does not create the role/function/bucket.
...
+  -w /workspace/deployment/aws \
+  quay.io/pypa/manylinux_2_28_aarch64 \
+  bash -c "yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh arm64"
+
+# x86_64 (on Apple Silicon this is emulated — needs Rosetta/qemu in the podman machine)
+podman run --rm \
+  -v "$(pwd)":/workspace \
+  -w /workspace/deployment/aws \
+  quay.io/pypa/manylinux_2_28_x86_64 \
+  bash -c "yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh x86_64"
```
espg: 🤖 *from Claude (review)*

**non-blocking** — the x86_64 invocation has no platform selector, and it is not a CI mirror. CI never builds cross-arch: x86_64 runs on `ubuntu-latest`, arm64 on `ubuntu-24.04-arm` (`lambda-build-reusable.yml` L21, L102), both native — so "exactly as CI does" (L98-99) covers only the arch-matched case, and this line is the one command in the section with no upstream precedent. On an arm64 host podman/docker have to be told to run an amd64 image; the repo's other container recipe does exactly that (`docs/deployment/arm64.md` L73: `docker run --rm --platform linux/arm64 ...`). Without it the failure is a confusing one rather than an obvious "exec format" error, because `build_layer.sh` L27-33 guards on `uname -m` first: `ERROR: building x86_64 layer on aarch64 machine`.

Concretely: add `--platform linux/amd64` to the x86_64 example and mention that guard message as the symptom when emulation isn't active — or drop the emulated recipe and point x86_64 at an x86_64 host / CI. (The PR body already flags this path as unexecuted; a wrong-but-normative command is the part worth de-risking.)

deployment/LAMBDA_DEPLOYMENT.md
```diff
@@ -83,124 +84,79 @@ already-deployed function and does not create the role/function/bucket.
...
+  -v "$(pwd)":/workspace \
+  -w /workspace/deployment/aws \
+  quay.io/pypa/manylinux_2_28_aarch64 \
+  bash -c "yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh arm64"
 ```
-Then unzip wheels into the build directory. This is what we're doing now for testing.
+
+On an SELinux-enforcing Linux host, add `:z` to the podman volume mount
+(`-v "$(pwd)":/workspace:z`) so the container can write the zip back; it is
+not needed on macOS (`podman machine`).
```
espg: 🤖 *from Claude (review)*

**nit** — the `:z` rationale is narrower than the failure it prevents. With SELinux enforcing, the container process cannot *read* the unlabeled bind mount either, so the build dies early — at `chmod +x build_layer.sh` / reading `../../pyproject.toml` for `MORTIE_SPEC` — not at "write the zip back" time. Also worth one clause noting `:z` relabels the mount **recursively with a shared label** (`:Z` is the private variant), since the mount here is the entire repo root, not a scratch dir.

deployment/LAMBDA_DEPLOYMENT.md
```diff
@@ -83,124 +84,79 @@ already-deployed function and does not create the role/function/bucket.
...
+
+The zip lands in `deployment/layers/lambda_layer_<arch>.zip`. The script
+enforces the 250 MB unzipped limit itself; CI additionally gates the combined
+layer + function size. CI builds both arches this same way on arch-matched
+runners (`lambda-build.yml` → `lambda-build-reusable.yml`), and release zips
+come from `publish.yml` calling the same reusable workflow — deploys consume
+those via `stand_up.sh` (above).
```
espg: 🤖 *from Claude (review)*

**non-blocking** — "Rebuilding the Layer" now dead-ends at a zip that nothing in the repo consumes. The deleted Option A block was also the file's *only* publish step (`main` L128 `aws lambda publish-layer-version`, L136 `aws lambda update-function-configuration`), and nothing replaces it. The `stand_up.sh` pointer doesn't close the gap: it deploys **release** zips from the distribution bucket (`s3://sliderule-public-cors/<minor>/`), never `deployment/layers/lambda_layer_<arch>.zip`. Nor does `deploy.sh`, which publishes the layer from the **CI artifact** — `gh run download --name "lambda-layer-${ARCH}"` (L46), then `publish-layer-version` with the >50 MB → S3 staging branch (L61-90) and `update-function-configuration` (L127).

One sentence closes it, e.g.: "the local zip is for verification / size-checking; to actually deploy a layer use `deploy.sh` (pulls the CI artifact, handles the >50 MB S3 upload and the function-config update) or cut a release and use `stand_up.sh`." Bonus: that also repairs L82's "`deploy.sh` (below)", which has had no "below" in this file since before this PR.

deployment/LAMBDA_DEPLOYMENT.md
```diff
@@ -83,124 +84,79 @@ already-deployed function and does not create the role/function/bucket.
...
-
-# Clean and zip
-find /tmp/lambda_build -type d -name '*.dist-info' -exec rm -rf {} +
-find /tmp/lambda_build -type d -name '__pycache__' -exec rm -rf {} +
-cd /tmp/lambda_build && zip -qr /tmp/lambda_function.zip .
+# Build zip (auto-detects arch and Python). Native deps (obstore) resolve for
+# the host, so run on Linux with the function's arch/Python — CI uses
+# arch-matched ubuntu runners; on a Mac, run inside a Linux container as with
+# the layer build.
+deployment/aws/build_function.sh
```
espg: 🤖 *from Claude (review)*

**non-blocking** — "on a Mac, run inside a Linux container as with the layer build" doesn't work with the layer build's image as written. `build_function.sh` detects the interpreter with bare `python3` (L38 → `ZIP_NAME` at L39) and installs with `pip3` (L57-61), with **no** cp312 fallback — unlike `build_layer.sh` L22, `PYTHON=$(command -v python3.12 || echo /opt/python/cp312-cp312/bin/python)`, a fallback that exists precisely because the manylinux images don't put cp312 on `PATH`. Run in `quay.io/pypa/manylinux_2_28_aarch64`, this script therefore picks up the image's default `python3`, yielding a mislabeled `lambda_function_arm64_py<other>.zip` (which `stand_up.sh` L96 wouldn't find, it expects `..._py312.zip`) or a pip failure, since `zarr>=3.1.5` requires-python ≥3.11. That's also why CI runs this one on the runner with `actions/setup-python@v6` 3.12 rather than in the container (`lambda-build-reusable.yml` L37-45).

Fix either way: name the interpreter (prepend `/opt/python/cp312-cp312/bin` to `PATH` inside the container, or use a `python:3.12` image), or drop the container suggestion and keep only the "Linux host with the function's arch/Python" framing.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the docs reconciliation. The layer package list now matches `build_layer.sh` exactly (all 18 direct installs, in order), the `boto3`-is-stripped and `pyarrow`→`arro3-core` notes check out, the podman/docker arm64 invocation is token-for-token the CI step, the zip paths (`deployment/layers/lambda_layer_<arch>.zip`, `deployment/builds/lambda_function_arm64_py312.zip`) match the scripts and `stand_up.sh`, and the `dist-info` claim matches `build_function.sh` L104-107. Six findings inline — none of them dispute option (b) or the podman-first framing; four are accuracy/completeness in the new prose, one is a cross-doc contradiction that undercuts the "one normative path" claim, one nit.

espg:
> The x86_64-on-Apple-Silicon note says emulation (Rosetta/qemu in the podman machine) is required — that matches the script's uname -m arch guard, but I couldn't execute a container build here; flag if the local podman-machine setup differs.

this seems unlikely to me... i think apple is ending (has ended?) rosetta x86 translation support in new versions of osx... also, I don't know when/why people would be using osx to compile the x86 images. Production is arm64 ; so that's why we have instructions for deriving that.

Seems like we can cut the x86 rosetta instructions...

deployment/LAMBDA_DEPLOYMENT.md
espg: 🤖 *from Claude*

Folded in e618918 (plus 2ca3cfa for the rewrap). Both overclaims are gone — the L36 parenthetical and the "single source of truth for layer contents" line in "The build". They now read:

- L36: "the normative build entry point; its pins are co-owned with the `lambda` extra in `pyproject.toml`, and mortie's version spec is read from `pyproject.toml` at build time — issue #322"
- "The build": "the normative build entry point for the layer (package set, numpy page-alignment build, bloat strip, 250 MB gate). Its pins are co-owned with the `lambda` extra in `pyproject.toml` — the script says "keep in sync" at each one — and mortie's spec is read out of `pyproject.toml` directly (issue #322), so a floor bump there reaches the layer with no second edit."

Note "pins" also came out of the parenthesized list of what the script owns, since that was the part that was actually false. Verified against `build_layer.sh` L91-99 (the `tomllib` read + the `^mortie-` assert at L103-105) and the four keep-in-sync comments at L64, L69, L80, L109.

deployment/LAMBDA_DEPLOYMENT.md
espg: 🤖 *from Claude*

Folded — extended the PR to `docs/deployment/arm64.md` in 6ee66aa. It is in `docs/` and is squarely the drifted-docs problem issue #325 describes, so reconciling it is in scope rather than a scope change. Kept the diff small and left the page's reason for existing (Apple-Silicon *why*: image choice, page alignment, glibc) intact. All three contradictions you listed:

1. **docker-first** → the Build Environment row is now "Container runtime | podman (what we use; Docker Desktop / OrbStack / Colima also work)", and step (1) is `brew install podman` + `podman machine init && podman machine start`.
2. **bare-host build** → step (2) is the same `podman run ... manylinux_2_28_aarch64 ... ./build_layer.sh arm64` invocation as the normative section, prefaced with "`build_layer.sh` is not a bare-host command; it needs the manylinux cp312 toolchain and refuses to run on a non-`aarch64` machine".
3. **wrong import list** → dropped `zarr`, `pydantic_zarr`, `obstore`, `earthaccess`; added `shapely`, `xarray`, `mortie`. Prefaced with why: the first three are `build_function.sh` deps stripped from the layer, `earthaccess` is orchestrator-only, and "expecting any of those here will fail a good layer". Also switched that verify command from `docker` to `podman` (keeping `--platform linux/arm64`).

Plus a top `!!! note` admonition pointing at LAMBDA_DEPLOYMENT.md's "Rebuilding the Layer" as normative with an explicit "if the two disagree, that section wins", so future drift resolves in one direction. Two adjacent staleness items in the same class came along: the "Slow build" tip no longer assumes Docker Desktop, and the "Why This Works" row "x86 runners need QEMU for ARM" now notes CI avoids that with native `ubuntu-24.04-arm` runners.

The one thing I left standing: `arm64.md`'s manual `publish-layer-version` recipe (with its `zagg-layer-arm64` name, which is not the deployed `<FunctionName>-deps`). Rather than delete it I prefixed it with a pointer to `deploy.sh` / `stand_up.sh` and framed it as "the manual equivalent, for a one-off layer of your own" — see the finding-5 thread, which is where that gap actually got closed.

deployment/LAMBDA_DEPLOYMENT.md
espg: 🤖 *from Claude*

Folded in de1f0f7, taking the **second** of your two options — dropped the emulated recipe rather than adding `--platform linux/amd64`. That was @espg's call on the PR thread ([comment](https://github.com/englacial/zagg/pull/329#issuecomment-5124259523)): Rosetta x86 translation is being wound down in current macOS, production is arm64, and there is no reason to compile x86 images on a Mac. So the wrong-but-normative command is gone entirely instead of being made correct-but-unexecuted.

What replaced it:

- the podman block is now arm64-only, prefaced "Production is **arm64**, and Apple Silicon runs `linux/arm64` containers natively, so the local recipe is the arm64 one";
- then: "**x86_64 is not a local-Mac path.** It exists for testing parity only; build it on an x86_64 Linux host (same command with `manylinux_2_28_x86_64` / `build_layer.sh x86_64`) or just take CI's `lambda-layer-x86_64` artifact. Don't try to emulate it on Apple Silicon."

Your "not a CI mirror" point is folded separately (2ca3cfa): the overreaching "exactly as CI does" is gone, replaced by the arch guard stated as the reason there is no cross-arch build — "it hard-fails on a mismatch (`ERROR: building <arch> layer on <machine> machine`, `build_layer.sh` L27-33) — so there is no cross-arch build. CI builds each arch on a native runner (`ubuntu-latest` / `ubuntu-24.04-arm`, `lambda-build-reusable.yml`)." Both runner labels verified at L21 and L102.

deployment/LAMBDA_DEPLOYMENT.md
espg: 🤖 *from Claude*

Folded in d723a1d. Both halves — the failure is now described at the right point in the build, and the recursive/shared-label caveat is stated:

> On an SELinux-enforcing Linux host, add `:z` to the podman volume mount (`-v "$(pwd)":/workspace:z`). Without it the container cannot *read* the unlabeled mount either, so the build dies at the top (`chmod +x build_layer.sh`, or reading `../../pyproject.toml` for `MORTIE_SPEC`) rather than at zip-write time. Note `:z` relabels the mount **recursively with a shared label** — and the mount here is the whole repo root, not a scratch dir; `:Z` is the private-label variant. Neither is needed on macOS (`podman machine`).

Your read-path claim checks out against the invocation: `chmod +x build_layer.sh` is the first thing the `bash -c` does, and `MORTIE_SPEC` is read from `${SCRIPT_DIR}/../../pyproject.toml` at `build_layer.sh` L91-95 — both well before any write to `deployment/layers/`.

deployment/LAMBDA_DEPLOYMENT.md
espg: 🤖 *from Claude*

Folded in 2aec671 — took your suggested sentence and expanded it slightly into an explicit block, since the dead-end was the substantive gap:

> **What consumes the zip:** nothing in the repo deploys `deployment/layers/lambda_layer_<arch>.zip` — it is for local verification and size-checking. To actually put a layer on a function:
> - **in-place maintainer update** — `deployment/aws/deploy.sh`. Note it pulls the **CI artifact**, not your local build (`gh run download --name "lambda-layer-${ARCH}"`), then `publish-layer-version` (staging through S3 when the zip is >50 MB) and `update-function-configuration`.
> - **full standup, or a fresh account/region** — cut a release and use `stand_up.sh`, which deploys the release zips staged on the distribution bucket (above), not local ones.

The "pulls the CI artifact, not your local build" framing is deliberate — that mismatch is the trap the deleted Option A block hid, and it verifies at `deploy.sh` L46/L50 (`gh run download`), L73/L83 (the two `publish-layer-version` branches) and L127 (`update-function-configuration`). `stand_up.sh` L77 + L96-97 confirm it only ever names `lambda_layer_<arch>.zip` / `lambda_function_<arch>_py312.zip` on `s3://sliderule-public-cors`.

Dangling reference repaired in the same commit: L82 now reads "`deploy.sh` is the maintainer path for *in-place updates* … — see [Rebuilding the Layer](#rebuilding-the-layer) below for what it consumes", so the "below" resolves to a real anchor for the first time. And per the finding-2 thread, `docs/deployment/arm64.md`'s manual publish recipe is now prefixed with the same `deploy.sh` / `stand_up.sh` pointer so it reads as the one-off fallback rather than the route.

deployment/LAMBDA_DEPLOYMENT.md
espg: 🤖 *from Claude*

Folded in 210f646 — took both of your fixes: dropped the "as with the layer build" container suggestion, and named the interpreter for anyone who does containerize. The comment block above `build_function.sh` now reads:

> Build zip. Native deps (obstore) resolve for the host, so run on Linux with the function's arch and Python 3.12. The script auto-detects both from the ambient `python3`/`uname -m` and bakes them into the zip name — it has no cp312 fallback (unlike `build_layer.sh`), so a non-3.12 `python3` silently yields a mislabeled `lambda_function_<arch>_py<other>.zip` that `stand_up.sh` won't find. Unlike the layer, this does NOT run in the manylinux image, whose default `python3` is not cp312; CI runs it on the arch-matched ubuntu runner under `actions/setup-python` 3.12. If you use a container, use one whose `python3` IS 3.12 (e.g. `python:3.12`) or prepend `/opt/python/cp312-cp312/bin` to `PATH`.

Your diagnosis verified end to end: bare `python3` at `build_function.sh` L38 feeding `ZIP_NAME` at L39, `pip3` at L57, no fallback anywhere in the file, against `build_layer.sh` L22's `command -v python3.12 || echo /opt/python/cp312-cp312/bin/python`. The silent-mislabel failure mode is the one worth calling out in the doc, since `stand_up.sh` L96-97 hardcodes `..._py312.zip` — I made that the explicit consequence rather than just saying "use 3.12". CI's split is as you describe: layer in the manylinux container (`lambda-build-reusable.yml` L25-34 / L106-115), function on the runner after `setup-python` (L37-45 / L118-126).

espg:
🤖 *from Claude*

Cut, in de1f0f7. You are right on both counts — Rosetta x86 translation is being wound down (deprecated in macOS 27, and the manylinux x86_64 image would need qemu in the podman VM regardless), and there was no scenario justifying an x86 build on a Mac. Production is arm64; that is the recipe that should exist.

The whole emulated block is gone. What is there now:

```bash
podman run --rm \
  -v "$(pwd)":/workspace \
  -w /workspace/deployment/aws \
  quay.io/pypa/manylinux_2_28_aarch64 \
  bash -c "yum install -y zip && chmod +x build_layer.sh && ./build_layer.sh arm64"
```

prefaced with "Production is **arm64**, and Apple Silicon runs `linux/arm64` containers natively, so the local recipe is the arm64 one", and followed by a short note that keeps x86_64 discoverable without pretending it is a Mac path:

> **x86_64 is not a local-Mac path.** It exists for testing parity only; build it on an x86_64 Linux host (same command with `manylinux_2_28_x86_64` / `build_layer.sh x86_64`) or just take CI's `lambda-layer-x86_64` artifact. Don't try to emulate it on Apple Silicon.

That kept `build_layer.sh x86_64` documented for the arch it actually runs on (CI builds it natively on `ubuntu-latest`, `lambda-build-reusable.yml` L21) while removing the command I could not execute — so the only build instruction in the file is now one that runs on the machine it is written for. The review bot had independently flagged the same line ([thread](https://github.com/englacial/zagg/pull/329#discussion_r3678583151)), offering either `--platform linux/amd64` or dropping it; your call settled which. The section also no longer claims to mirror CI — it states the `uname -m` arch guard (`build_layer.sh` L27-33) as the reason there is no cross-arch build at all.

Issue #327, opened 2026-07-29
zagg.client v2 transport: Event invoke + status-object future resolver


espg:
🤖 *from Claude*

Refs #265 ([ratified shape](https://github.com/englacial/zagg/issues/265#issuecomment-5123966118)); depends on #326 (the public futures API this swaps the resolver under).

Fleet-scale transport for the `zagg.client` futures: switch dispatch to Lambda **`Event`** invocation (fire-and-forget, returns immediately) and resolve futures by polling per-shard **status objects** in the store — the mechanism half-exists (runs already write a `<store>.zarr.status/` prefix). Properties v1 can't offer:

- no held connection per in-flight shard (thousands of shards without a thread pool pinned open)
- a run survives client disconnect and is **reattachable by run id** (a new `RunHandle` re-binds to the status prefix)
- completion/error polling is an S3 GET; worker error payloads land in the status object instead of an invoke response

Scope: worker-side status-object write on completion/error (small, worker-owned — dispatcher-never-writes preserved), the polling resolver behind the same `Future` surface, reattach-by-run-id, and backoff/poll-interval policy. Public API from #326 unchanged.

espg:
🤖 *from Claude*

Implementation plan. The public API is frozen by #326/PR #333 — v2 swaps the resolver under the same `Run`/`RunHandle`/`Future` surface; v1 (thread-pool over sync invokes) stays the default until you flip it.

## Phases (branch `claude/327-event-transport`)

1. **Worker-side per-shard status objects.** On completion or caught failure, the worker writes one tiny JSON to `<store>.zarr.status/run-<run_id>/shard-<key>.json` — status, phase timings, error payload, and an **attempt id**. Written by the worker itself, so D8 stays intact. Gated by an event flag the v2 dispatch sets (`status_object: true`), default off — `agg`'s sync path and the v1 client pay nothing.
2. **Dispatch manifest via the setup invoke.** Reattach needs the expected-shard set, and the dispatcher never writes — so the existing worker setup invoke (the hive ping/manifest write) is extended to also record the run manifest (run_id, shard keys, config hash) as a status object. D8 preserved by construction.
3. **Event dispatch + polling resolver.** Client transport switch: `InvocationType="Event"` (202, no held connection); a single poller thread resolves all futures from **one LIST of the status prefix per poll tick** (not per-shard GETs — one request at fleet scale), backoff 2 s → 15 s cap. Futures/`as_completed`/`status()`/`progress()` behave identically to v1.
[...]

PR #333, opened 2026-07-29
zagg.client facade: notebook API, v1 shard futures, tqdm progress (issue #326)

### LINKED ISSUE #326: zagg.client facade: notebook API, v1 shard futures, tqdm progress
🤖 *from Claude*

Refs #265 ([ratified shape](https://github.com/englacial/zagg/issues/265#issuecomment-5123966118), [approval](https://github.com/englacial/zagg/issues/265#issuecomment-5124011280)).

A notebook-first library API owning the composition that `.github/scripts/run_benchmark.py` and the dispatch scratch scripts do imperatively today: config-load → shardmap → dispatch → harvest. CI scripts become thin wrappers so the notebook API cannot drift from what CI tests. This is also where the (substantial) rework of the out-of-date #169 lands.

Scope:
- `zagg.client` facade: `Run.from_config(...)`, `run.dispatch() -> handle`, per-shard `concurrent.futures.Future`s (`.done()/.result()/.exception()`), `handle.as_completed()`, `handle.status()` aggregate counts.
- **v1 transport**: `ThreadPoolExecutor` over the existing synchronous `RequestResponse` invokes — zero worker-side change; per-shard results/errors surface through the future. Fine at NEON scale; cost is one held connection per in-flight shard.
- tqdm progress = `tqdm.auto` over `as_completed(...)`; **tqdm goes in the existing `analysis` extra** (espg-ratified — no new extra, never in `lambda`).
- Dispatcher-never-writes invariant untouched: futures observe worker-written state only.

v2 transport (Event invoke + status-object resolver) is deliberately out of scope — tracked separately so the public API here is designed to survive that swap.

espg:
Closes #326. Refs #265 ([ratified API shape](https://github.com/englacial/zagg/issues/265#issuecomment-5123966118), [espg sign-off, incl. tqdm into the existing `analysis` extra](https://github.com/englacial/zagg/issues/265#issuecomment-5124017044)).

## What

`zagg.client` — a notebook-first facade owning the config-load → shardmap → dispatch → harvest composition that `.github/scripts/run_benchmark.py` and the dispatch scratch scripts do imperatively today, with **per-shard `concurrent.futures.Future`s** as the public contract:

```python
from zagg.client import Run

run = Run.from_config("cfg.yaml", shardmap="shardmap.json", store="s3://bucket/out.zarr")
handle = run.dispatch()                    # returns after the setup handshake
for fut in handle.progress():              # tqdm.auto over as_completed()
    r = fut.result()                       # worker envelope (timings) / raises ShardError
handle.status()                            # {"pending": n, "ok": n, "failed": n}
```

Public surface: `Run.from_config(config_path_or_dict_or_PipelineConfig, shardmap=, store=, function_name=, region=, lambda_client=, **knobs)`, `Run.dispatch(shard_keys=None, max_workers=None) -> RunHandle`; `RunHandle.futures` (dict `shard_key -> Future`), `.as_completed()`, `.status()`, `.results(return_exceptions=False)`, `.raise_first()`, `.progress()` (tqdm, optional import), `.wait()`, `__len__`. Worker error payloads surface as `ShardError` from `future.result()` with the full per-shard result dict attached (`.payload`: status_code, body, timings, retries).

## Approach

- **v1 transport** (as ratified on #265): a `ThreadPoolExecutor` over the **existing synchronous `RequestResponse` invokes** — each shard is one `runner._invoke_lambda_cell` call and the pool future over it *is* the shard's future. Zero worker-side change; payloads are built with the same runner helpers `_run_lambda`'s `_cell_work` uses (`_safe_label`, `_resolve_urls`, `_clamped_data_source`, aoi payload map, leaf submap, `run_id`/`invoked_by` threading), so cell events stay byte-compatible with the `agg` path.
- **Dispatcher-never-writes (D8) untouched**: setup handshake (hive ping + fire-and-forget manifest write / flat template write), finalize backstop, and the fail-open coverage.moc / run-stats / sweep rollups all remain **worker invokes**; a background finisher thread runs that tail after the last shard settles, and `handle.wait()` surfaces a finalize failure. The client holds futures; it never PUTs.
- **boto3 client injection**: `from_config(lambda_client=...)` takes any invoke-shaped client (tests use an in-memory stub — no moto, no network). Default `None` builds a per-dispatch client mirroring `_run_lambda`'s botocore config (960 s read timeout, pool sized to the fan-out) and resolves STS `invoked_by`; injected clients skip the STS probe (fail-open, #297 semantics).
- **tqdm** is imported only inside `RunHandle.progress()`; `import zagg.client` works without it (regression-tested). Added `tqdm>=4.66` to the **existing `analysis` extra** — never `lambda` — per the espg sign-off linked above.
- **v1 scope gate**: temporal, raster, and windowed configs are refused with a `NotImplementedError` pointing at `zagg.runner.agg` / `zagg.notebook.run` (windowed fan-out units are `(shard, window)` pairs, not shards, so per-shard future keys would collide).

## Phases

- [x] Phase 1: `src/zagg/client.py` (`Run` / `RunHandle` / `ShardError`, v1 sync transport, worker-invoke post-run tail) + `tqdm` in the `analysis` extra
- [x] Phase 2: stub-Lambda-client unit tests (`tests/test_client.py`, 28 tests)
- [x] Adversarial-review fold (7 findings): tail joined when the harvest iterator drains, whole tail guarded with the failure surfaced, failure rows for non-`ShardError` shard failures, differential `TestRunnerParity` against `runner.agg`, run-stats verify-window timing test, `zagg.dispatch.BENIGN_ERRORS` as one public home
- [x] Phase 3: **driver fix in `agg`'s Lambda path** — `_run_lambda` threads the resolved `driver` into `_cell_work`, replacing the hardcoded `_resolve_urls(records, "s3")` ([espg ruling](https://github.com/englacial/zagg/pull/333#issuecomment-5124576186))
- [x] Phase 4: **concurrency preflight, option (a)** — `dispatch()` sizes the pool with `compute_available_workers` when `max_workers` is omitted and no client is injected; any probe failure degrades to `_DEFAULT_MAX_WORKERS` with a `RuntimeWarning` naming the denial
- [x] Phase 5: **finalize warns immediately** — fail-open + recorded + re-raise as before, plus a `RuntimeWarning` at the moment of failure naming the store and the idempotent remedy
- [ ] Follow-up (separate PR, deliberately not here): refactor `.github/scripts/run_benchmark.py` / the dispatch scratch scripts into thin wrappers over the facade so the notebook API cannot drift from what CI tests
- [ ] v2 transport (issue #327, out of scope): `Event` invoke + status-object resolver behind this same public surface — no held connections, reattachable runs

## How it was tested

- `pytest tests/test_client.py -v` — **44 tests**, all against a stub Lambda client: fan-out (one sync `RequestResponse` invoke per shard, no `result_url`), the complete cell-event key set, hive ping→setup ordering + finalize/stats/coverage tail, flat setup + finalize gating, `as_completed` resolution order, error payload → `ShardError` (incl. `FunctionError` timeout), `status()` counts with a gated in-flight snapshot, benign "No granules found" counted ok, `results(return_exceptions=)`, `raise_first()`, shard-subset selection by label and raw key, config/shardmap loading (dict / path / `PipelineConfig`), scope-gate refusals, and no-tqdm import safety. Added by the review fold and the phase 3-5 rulings:
  - `TestRunnerParity` — differential against `runner.agg(backend="lambda", invocation="sync")` on the same stub, parametrized over `driver: s3` / `driver: https`, comparing whole cell events field-by-field.
  - `TestConcurrencyPreflight` — probe sizes the pool, denial degrades with a `RuntimeWarning`, explicit `max_workers` skips the probe, injected client skips the probe.
  - Tail behavior — draining (`as_completed` / `progress` / `results` / `raise_first`) dispatches the whole tail without `wait()`; a tail exception surfaces and still shuts the pool down; finalize failure warns immediately while the tail is still in flight and still re-raises; a non-`ShardError` shard failure still gets a run-stats failure row; `TestTailTiming` exercises the run-stats verify window.
- `pytest tests/test_runner.py -v` — plus `TestDriverPassthroughLambda` (3 tests): the Lambda path resolves s3 hrefs by default, https from the `driver=` kwarg, and https from the config key alone.
- Full suite: `uv sync --extra test && uv run pytest --ignore=tests/test_lambda_build.py` → **2671 passed, 34 skipped**. `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` still fails as it does on `main` (unrelated to this PR; not touched per conventions).
- `ruff check src tests` / `ruff format --check src tests`: clean for the changed files. Pre-existing on `main` and left alone: `N818` on `src/zagg/registry.py:64` (the PR lint bot runs without `N`), and the mypy hook's pre-existing noise (e.g. missing `yaml` stubs, `runner.py` union-attr notes) — `src/zagg/client.py` itself is mypy-clean.

## Questions for review

1. ~~**Fan-out width / no concurrency probe**~~ — **RESOLVED** ([espg](https://github.com/englacial/zagg/pull/333#issuecomment-5124576186), option (a)): the probe now runs by default and degrades gracefully. When `max_workers` is omitted **and** no custom `lambda_client` was injected, `dispatch()` runs `agg`'s `compute_available_workers` (FD ceiling ∩ account headroom, requested at the work size) to size the pool; **any** probe failure — `AccessDenied` in a future narrow-permission cryocloud deployment, missing CloudWatch perms, a throttle — falls back to `_DEFAULT_MAX_WORKERS` with a `RuntimeWarning` naming the denial, so a run the caller may dispatch never fails on an optional preflight read. Explicit `max_workers=` wins verbatim and skips the probe; an injected client skips it too (stub-client seam). Phase 4, `e51acfc`.
2. **`dispatch()` returns after the setup handshake** (one short synchronous worker invoke; hive fail-fast-pings first) rather than backgrounding setup into the futures. This keeps the #252 fail-fast property — an incompatible store refuses before any shard is paid for — at the cost of "immediately" meaning "after ~one invoke round-trip". Acceptable reading of the ratified sketch?
3. **Post-run tail scope**: the finisher thread mirrors `agg`'s full lambda tail (finalize backstop always on hive — no overlapped manifest checker, so one extra idempotent invoke vs `_run_lambda`; coverage.moc; run-stats record; sweep trigger — all fail-open worker invokes). Alternative was to skip the rollups in v1 and leave client-run stores thinner than agg-run stores; went with parity.
4. **Benign outcomes**: shards returning `"No granules found"` / `"No data after filtering"` resolve normally and count `ok` in `status()` (matching `_run_lambda`'s error accounting, which counts them neither with-data nor error). Flag if they should be a third status bucket instead.

5. ~~**Two deliberate divergences from the byte-compatible-events claim**~~ — **RESOLVED** ([espg ruling](https://github.com/englacial/zagg/pull/333#issuecomment-5124576186)); both were folded into this PR, so nothing here is outstanding:
   - **`data_source.driver` — fixed in `agg`, not worked around in the facade** ("yes, this should be fixed — and we might as well fix it here"). `runner._run_lambda` now takes the resolved `driver` and `_cell_work` selects the href with it, replacing the hardcoded `_resolve_urls(records, "s3")`; `agg` forwards the same value it already gave `_run_local`. Use case per the ruling: a STAC catalog supplying an **https substitute** for an s3 object this account cannot or should not read (requester-pays, cross-region) — s3 dominates, but https must work as an explicit override, and the worker already builds `HTTPDriver` from the same config key (`processing/worker.py:250-254`). `TestRunnerParity` drops its carve-out: it is now parametrized over `driver: s3` **and** `driver: https` and asserts the full cell event equal field-by-field on both (`run_id` popped, since it is a fresh uuid per dispatch), plus that the href actually moved on each. Runner-side coverage in `TestDriverPassthroughLambda` (kwarg, config key, and the s3 default). Phase 3, `8520fb5`.
   - **Finalize failure — the middle option** (asked: "is there a fix or recommendation... fatal, or try block with a warning?"). Kept fail-open + recorded + re-raising, and added an immediate `RuntimeWarning` at the moment finalize fails: it names the store path, states that shard outputs are all written and only the root manifest may be stale, and that the backstop is **idempotent** so a re-run (or re-dispatched finalize) stamps it. The tail's rollups still run and `wait()` / a drained harvest iterator still re-raise, so nothing is silent and nothing is lost. Phase 5, `bb9fa27`. Making `agg` itself run the rollups before exiting nonzero (rather than dying pre-tail) remains out of scope here — flag if you want it filed.

---
Authored by a Claude Code session working issue #326.

### DIFF
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -13,6 +13,15 @@ shardmap → dispatch → harvest — as library API::
         r = fut.result()                       # worker envelope / raises ShardError
     handle.status()                            # {"pending", "ok", "failed"}
 
+Draining the harvest iterator to exhaustion (``as_completed`` / ``progress`` /
+``results`` / a clean ``raise_first``) also joins the post-run tail — the
+finalize backstop and the coverage/stats/sweep rollup invokes — so the loop
+above IS the whole run and :meth:`RunHandle.wait` is only needed for explicit
+control (a timeout, or a handle harvested some other way). Abandoning a handle
+without draining it is the one flow that truncates the tail: the finisher is a
+daemon thread, so an undrained handle cannot wedge interpreter exit (review
+finding, PR #333).
+
 **v1 transport** (ratified on issue #265): a ``ThreadPoolExecutor`` over the
 existing synchronous ``RequestResponse`` invokes — each shard is one
 :func:`zagg.runner._invoke_lambda_cell` call and its pool future is the
@@ -109,7 +118,12 @@ class RunHandle:
     A background finisher thread waits for every shard, then runs the same
     worker-invoke tail as ``agg``'s lambda path: the finalize backstop, plus
     the fail-open coverage/stats/sweep rollup invokes (D8: all worker-side).
-    :meth:`wait` joins it and surfaces a finalize failure.
+    Draining any of the harvest iterators (:meth:`as_completed`,
+    :meth:`progress`, :meth:`results`, a clean :meth:`raise_first`) joins that
+    finisher before returning and surfaces a finalize failure, so a completed
+    harvest loop means a completed run; :meth:`wait` is the explicit form (and
+    the only one taking a timeout). Only a handle nobody drains truncates the
+    tail — the finisher is a daemon so that case cannot hang exit.
     """
 
     def __init__(self, futures: dict[int, Future], *, store_path: str):
@@ -129,8 +143,15 @@ class RunHandle:
         )
 
     def as_completed(self, timeout: float | None = None) -> Iterator[Future]:
-        """Yield each shard future as it completes (``concurrent.futures`` order)."""
-        return as_completed(self.futures.values(), timeout=timeout)
+        """Yield each shard future as it completes (``concurrent.futures`` order).
+
+        Exhausting this iterator joins the post-run tail (see :meth:`wait`) and
+        re-raises a finalize failure, so the documented harvest loop finishes a
+        complete run without calling :meth:`wait` (review finding, PR #333).
+        Breaking out early skips that join — call :meth:`wait` if you do.
+        """
+        yield from as_completed(self.futures.values(), timeout=timeout)
+        self._join_tail()
 
     def status(self) -> dict[str, int]:
         """Non-blocking snapshot: ``{"pending": n, "ok": n, "failed": n}``."""
@@ -150,19 +171,31 @@ class RunHandle:
         With ``return_exceptions=False`` (default) the first failed shard's
         :class:`ShardError` propagates (key order); with ``True`` failures
         appear as the exception object in the mapping instead — the
-        ``asyncio.gather`` convention.
+        ``asyncio.gather`` convention. Either way the post-run tail is joined
+        first (see :meth:`as_completed`), so a shard failure does not truncate
+        the rollups.
         """
         out: dict[int, Any] = {}
+        first_exc: BaseException | None = None
         for key, fut in self.futures.items():
-            if return_exceptions:
-                exc = fut.exception()
-                out[key] = exc if exc is not None else fut.result()
-            else:
+            exc = fut.exception()
+            if exc is None:
                 out[key] = fut.result()
+            else:
+                out[key] = exc
+                if first_exc is None:
+                    first_exc = exc
+        self._join_tail()
+        if first_exc is not None and not return_exceptions:
+            raise first_exc
         return out
 
     def raise_first(self) -> None:
-        """Block until the first failure (raising it) or all shards complete."""
+        """Block until the first failure (raising it) or all shards complete.
+
+        A clean run drains :meth:`as_completed`, so it joins the post-run tail;
+        raising on a failure leaves the tail in flight (:meth:`wait` joins it).
+        """
         for fut in self.as_completed():
             exc = fut.exception()
             if exc is not None:
@@ -193,11 +226,27 @@ class RunHandle:
         reader-facing schema, D6); the coverage/stats/sweep rollups are
         fail-open by design and never raise here. Raises ``TimeoutError``
         when the tail is still running at ``timeout``.
+
+        Draining a harvest iterator already joins the tail, so ``wait`` is for
+        explicit control.
         """
         if self._finisher is not None:
             self._finisher.join(timeout)
             if self._finisher.is_alive():
                 raise TimeoutError(f"run tail still in flight after {timeout}s")
+        self._raise_tail_error()
+
+    def _join_tail(self) -> None:
+        """Join the post-run finisher, then re-raise what it recorded.
+
+        Idempotent (joining a finished thread is a no-op), so draining a
+        harvest iterator and then calling :meth:`wait` is safe.
+        """
+        if self._finisher is not None:
+            self._finisher.join()
+        self._raise_tail_error()
+
+    def _raise_tail_error(self) -> None:
         if self._finalize_error is not None:
             raise self._finalize_error
 
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -225,7 +225,8 @@ class RunHandle:
         """Block until every shard AND the post-run tail finish.
 
         Re-raises a failed finalize backstop (the manifest is required
-        reader-facing schema, D6); the coverage/stats/sweep rollups are
+        reader-facing schema, D6 — also warned at the moment it fails, so it is
+        visible before this join); the coverage/stats/sweep rollups are
         fail-open by design and never raise here, but an exception in the
         tail's own plumbing does surface (it would otherwise vanish into the
         finisher thread). Raises ``TimeoutError`` when the tail is still
@@ -776,8 +777,10 @@ class Run:
         flat only under ``consolidate_metadata``), then the fail-open rollups:
         root ``coverage.moc``, the run-stats record, and the sweep trigger —
         each a fire-and-forget worker invoke (D8), each swallowed on failure
-        exactly like ``_run_lambda``'s tail. A finalize failure is recorded on
-        the handle and re-raised from :meth:`RunHandle.wait`.
+        exactly like ``_run_lambda``'s tail. A finalize failure warns
+        immediately (``RuntimeWarning`` — notebook-visible while the tail is
+        still running), is recorded on the handle, and re-raises from
+        :meth:`RunHandle.wait` / a drained harvest iterator.
 
         Every shard contributes exactly one result dict, so every shard gets a
         run-stats row: a worker envelope, a :class:`ShardError` payload, or —
@@ -831,6 +834,23 @@ class Run:
         except Exception as e:
             handle._finalize_error = e
             logger.warning(f"finalize invoke failed (surfaced via handle.wait()): {e}")
+            # Warn AT THE FAILURE, not only when the harvest loop finally joins
+            # (espg ruling on PR #333, middle option): the tail goes on and the
+            # error still re-raises from wait()/drain, but a notebook shows this
+            # in-stream immediately instead of leaving the window between
+            # finalize and the join silent. Every shard's data is written — only
+            # the root manifest is stale — and the backstop is idempotent, so
+            # the remedy is a re-run, not a re-compute.
+            warnings.warn(
+                f"finalize (manifest backstop) failed for {self.store}: {e}. "
+                f"Shard outputs are written; the store's root manifest may be "
+                f"stale until finalize runs again. It is idempotent — re-run "
+                f"the dispatch (or re-dispatch finalize) to stamp the manifest. "
+                f"This error also re-raises from handle.wait() / draining the "
+                f"harvest iterator.",
+                RuntimeWarning,
+                stacklevel=2,
+            )
 
         ok_results = [r for r in results if r.get("status_code") == 200 and not r.get("error")]
         if layout == "hive" and get_coverage_moc(self.config):
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -73,11 +73,10 @@ from zagg.config import (
     load_config_from_dict,
     validate_config,
 )
-from zagg.dispatch import LAMBDA_ARCH, max_cost_usd
+from zagg.dispatch import BENIGN_ERRORS, LAMBDA_ARCH, max_cost_usd
 from zagg.grids import from_config as grid_from_config
 from zagg.grids.morton import morton_word
 from zagg.hive import effective_store_root
-from zagg.notebook import _BENIGN_ERRORS
 
 logger = logging.getLogger(__name__)
 
@@ -679,7 +678,7 @@ class Run:
         error = result.get("error")
         if result.get("status_code") == 200 and not error:
             return result
-        if error in _BENIGN_ERRORS:
+        if error in BENIGN_ERRORS:
             # "Nothing to do" is a normal outcome, not a failure — matching
             # _run_lambda's error counting (neither with_data nor error).
             return result
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -718,18 +718,36 @@ class Run:
         each a fire-and-forget worker invoke (D8), each swallowed on failure
         exactly like ``_run_lambda``'s tail. A finalize failure is recorded on
         the handle and re-raised from :meth:`RunHandle.wait`.
+
+        Every shard contributes exactly one result dict, so every shard gets a
+        run-stats row: a worker envelope, a :class:`ShardError` payload, or —
+        for a transport-level exception that never produced an envelope (the
+        payload-cap ``ValueError``, an fd-exhaustion re-raise, a botocore error
+        escaping the retry loop) — a synthesized failure dict, which
+        ``_lambda_result_rows`` turns into a ``failure_record`` row rather than
+        dropping the shard from telemetry (review finding, PR #333).
         """
         from zagg import runner
 
         futures = list(handle.futures.values())
         futures_wait(futures)
         results = []
-        for fut in futures:
+        for key, fut in handle.futures.items():
             exc = fut.exception()
             if exc is None:
                 results.append(fut.result())
             elif isinstance(exc, ShardError):
                 results.append(exc.payload)
+            else:
+                results.append(
+                    {
+                        "shard_key": key,
+                        "status_code": None,
+                        "body": {},
+                        "error": f"{type(exc).__name__}: {exc}",
+                        "retries": None,
+                    }
+                )
         layout = get_store_layout(self.config)
         try:
             if layout == "hive":
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -1,0 +1,734 @@
+"""Notebook-first dispatch facade: per-shard futures over Lambda (issue #326).
+
+:class:`Run` owns the composition that ``.github/scripts/run_benchmark.py``
+and the dispatch scratch scripts do imperatively today — config-load →
+shardmap → dispatch → harvest — as library API::
+
+    from zagg.client import Run
+
+    run = Run.from_config("atl03_tdigest_healpix_o9_hive.yaml",
+                          shardmap="shardmap.json", store="s3://bucket/out.zarr")
+    handle = run.dispatch()                    # returns after the setup handshake
+    for fut in handle.progress():              # tqdm over as_completed
+        r = fut.result()                       # worker envelope / raises ShardError
+    handle.status()                            # {"pending", "ok", "failed"}
+
+**v1 transport** (ratified on issue #265): a ``ThreadPoolExecutor`` over the
+existing synchronous ``RequestResponse`` invokes — each shard is one
+:func:`zagg.runner._invoke_lambda_cell` call and its pool future is the
+shard's :class:`concurrent.futures.Future`, wrapping exactly what the worker
+returns today (timings, error payloads). Zero worker-side change; the cost is
+one held connection per in-flight shard and the client staying alive. The v2
+transport (``Event`` invoke + status-object resolver, issue #327) swaps the
+resolver under this same public surface.
+
+The dispatcher-never-writes invariant (D8) is untouched: every store write —
+template setup, shard output, finalize, coverage/stats/sweep rollups — rides a
+worker invoke; this module only holds futures over those invokes.
+
+Scope (v1): the spatial point path on the ``lambda`` backend. Temporal,
+raster, and windowed configs are refused with a pointer to
+:func:`zagg.runner.agg` / :func:`zagg.notebook.run`, which already run them.
+
+tqdm is an *optional* import (``analysis`` extra, espg-ratified on issue
+#265): only :meth:`RunHandle.progress` touches it, so importing this module —
+and everything else in zagg — works without it.
+"""
+
+from __future__ import annotations
+
+import logging
+import threading
+import uuid
+from concurrent.futures import Future, ThreadPoolExecutor, as_completed
+from concurrent.futures import wait as futures_wait
+from dataclasses import asdict
+from typing import Any, Iterator
+
+from zagg.config import (
+    PipelineConfig,
+    get_child_order,
+    get_consolidate_metadata,
+    get_coverage_moc,
+    get_driver,
+    get_handoff,
+    get_output_endpoint_url,
+    get_output_region,
+    get_parent_order,
+    get_pipeline_type,
+    get_store_layout,
+    get_store_path,
+    get_sweep,
+    get_windowing,
+    load_config,
+    load_config_from_dict,
+    validate_config,
+)
+from zagg.dispatch import LAMBDA_ARCH, max_cost_usd
+from zagg.grids import from_config as grid_from_config
+from zagg.grids.morton import morton_word
+from zagg.hive import effective_store_root
+from zagg.notebook import _BENIGN_ERRORS
+
+logger = logging.getLogger(__name__)
+
+#: Fan-out width when ``dispatch(max_workers=...)`` is not given. Deliberately
+#: modest: the v1 sync transport holds one connection per in-flight shard and
+#: this path runs no account-concurrency probe (unlike ``agg``'s preflight,
+#: which clamps against account limits before sizing the pool) — fine at
+#: notebook/NEON scale; pass ``max_workers`` explicitly for bigger fan-outs.
+_DEFAULT_MAX_WORKERS = 64
+
+
+class ShardError(RuntimeError):
+    """A shard's worker invoke failed; raised from that shard's ``future.result()``.
+
+    Carries the full per-shard result dict (``payload``) the dispatch loop
+    accumulates on the ``agg`` path — ``{shard_key, status_code, body,
+    wall_time, lambda_duration, error, retries, ...}`` — so timings and the
+    worker's error body survive into the exception.
+    """
+
+    def __init__(self, message: str, *, shard_key: int, label: str, payload: dict):
+        super().__init__(message)
+        self.shard_key = shard_key
+        self.label = label
+        self.payload = payload
+
+
+class RunHandle:
+    """Live handle over one dispatched fan-out: per-shard futures + rollup.
+
+    ``futures`` maps the shardmap's shard key (int) to that shard's
+    :class:`~concurrent.futures.Future`. A future resolves to the worker's
+    result dict; a failed shard raises :class:`ShardError` from ``.result()``.
+    Benign no-work outcomes (``"No granules found"`` / ``"No data after
+    filtering"``) resolve normally and count as ``ok`` — matching how the
+    ``agg`` path counts them (neither data nor error).
+
+    A background finisher thread waits for every shard, then runs the same
+    worker-invoke tail as ``agg``'s lambda path: the finalize backstop, plus
+    the fail-open coverage/stats/sweep rollup invokes (D8: all worker-side).
+    :meth:`wait` joins it and surfaces a finalize failure.
+    """
+
+    def __init__(self, futures: dict[int, Future], *, store_path: str):
+        self.futures = futures
+        self.store_path = store_path
+        self._finisher: threading.Thread | None = None
+        self._finalize_error: BaseException | None = None
+
+    def __len__(self) -> int:
+        return len(self.futures)
+
+    def __repr__(self) -> str:
+        s = self.status()
+        return (
+            f"RunHandle({len(self)} shards: {s['pending']} pending, "
+            f"{s['ok']} ok, {s['failed']} failed)"
+        )
+
+    def as_completed(self, timeout: float | None = None) -> Iterator[Future]:
+        """Yield each shard future as it completes (``concurrent.futures`` order)."""
+        return as_completed(self.futures.values(), timeout=timeout)
+
+    def status(self) -> dict[str, int]:
+        """Non-blocking snapshot: ``{"pending": n, "ok": n, "failed": n}``."""
+        counts = {"pending": 0, "ok": 0, "failed": 0}
+        for fut in self.futures.values():
+            if not fut.done():
+                counts["pending"] += 1
+            elif fut.exception() is not None:
+                counts["failed"] += 1
+            else:
+                counts["ok"] += 1
+        return counts
+
+    def results(self, *, return_exceptions: bool = False) -> dict[int, Any]:
+        """Block until every shard completes; return ``{shard_key: result}``.
+
+        With ``return_exceptions=False`` (default) the first failed shard's
+        :class:`ShardError` propagates (key order); with ``True`` failures
+        appear as the exception object in the mapping instead — the
+        ``asyncio.gather`` convention.
+        """
+        out: dict[int, Any] = {}
+        for key, fut in self.futures.items():
+            if return_exceptions:
+                exc = fut.exception()
+                out[key] = exc if exc is not None else fut.result()
+            else:
+                out[key] = fut.result()
+        return out
+
+    def raise_first(self) -> None:
+        """Block until the first failure (raising it) or all shards complete."""
+        for fut in self.as_completed():
+            exc = fut.exception()
+            if exc is not None:
+                raise exc
+
+    def progress(self, **tqdm_kwargs) -> Iterator[Future]:
+        """:meth:`as_completed` under a ``tqdm.auto`` bar (one tick per shard).
+
+        tqdm is optional (the ``analysis`` extra, espg-ratified on issue
+        #265); without it this raises ``ImportError`` with the install hint —
+        :meth:`as_completed` is the bar-free equivalent.
+        """
+        try:
+            from tqdm.auto import tqdm
+        except ImportError as e:
+            raise ImportError(
+                "RunHandle.progress() needs tqdm (`pip install 'zagg[analysis]'`, "
+                "issue #326); handle.as_completed() is the bar-free equivalent"
+            ) from e
+        kwargs: dict[str, Any] = {"total": len(self), "desc": "shards", "unit": "shard"}
+        kwargs.update(tqdm_kwargs)
+        return tqdm(self.as_completed(), **kwargs)
+
+    def wait(self, timeout: float | None = None) -> None:
+        """Block until every shard AND the post-run tail finish.
+
+        Re-raises a failed finalize backstop (the manifest is required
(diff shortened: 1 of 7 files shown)


src/zagg/client.py
```diff
@@ -0,0 +1,734 @@
...
+                max_workers=workers,
+            )
+
+        handle = RunHandle(futures, store_path=self.store)
+        finisher = threading.Thread(
+            target=self._post_run,
+            args=(handle, client, pool, config_dict, dataset, output_creds_event, run_id),
+            name="zagg-client-finish",
+            daemon=True,
+        )
```
espg, line 661: 🤖 *from Claude (review)*

**blocking** — the post-run tail is silently truncated at process exit, so the "agg-parity post-run tail" claim does not hold for the flow the module docstring and PR body document.

The finisher is `daemon=True` and nothing joins it except `RunHandle.wait()`. The documented flow never calls `wait()`:

```python
handle = run.dispatch()
for fut in handle.progress():      # module docstring lines 11-14, and the PR body
    r = fut.result()
handle.status()
```

The last future resolving is exactly the moment the finisher wakes, so the user's loop exits and the interpreter tears down while the tail is mid-flight. Probed against a stub Lambda client (0.25 s per tail invoke — a conservative stand-in for a real round trip), recording every `mode`:

```
# documented flow, process then exits
MODES: ['ping', 'setup', None, None, None, 'finalize']

# identical run, with handle.wait() added
MODES: ['ping', 'setup', None, None, None, 'finalize', 'coverage', 'stats']
```

`coverage` and `stats` were never dispatched. `finalize` only survived because it is first in the tail; a slower finalize round trip loses the D6-required manifest backstop too. This bites hardest on the phase-3 follow-up in the PR body — converting `.github/scripts/run_benchmark.py` and the dispatch scratch scripts to this facade — because those are exactly the short-lived processes that exit right after the harvest loop.

Options, increasing intrusiveness:

1. `atexit.register` a bounded join of the finisher inside `dispatch()` (keeps `wait()` optional and the notebook non-blocking).
2. Make the finisher non-daemon so `threading._shutdown` joins it (simplest; blocks interpreter exit for the tail's duration, which is the semantics `agg` already has).
3. Keep it as-is but make `wait()` part of the documented contract — in the module docstring example, the PR body example, and a note that skipping it drops the rollups.

Whichever way, the class docstring's "A background finisher thread waits for every shard, then runs the same worker-invoke tail as `agg`" (lines 109-112) should stop reading as unconditional.

src/zagg/client.py
```diff
@@ -0,0 +1,734 @@
...
+                        envelope,
+                        output_creds_event=output_creds_event,
+                    )
+            except Exception as e:
+                logger.warning(f"root coverage.moc dispatch failed (fail-open, D9): {e}")
+
+        # Run-record + sweep, both fail-open (issues #297/#313/#300): sync
+        # transport -> no status prefix, so oversized row sets skip inside
+        # _dispatch_run_stats exactly like agg's invocation="sync" path.
+        stats_rows, stats_inline = runner._lambda_result_rows(results, run_id=run_id)
```
espg, line 884: 🤖 *from Claude (review)*

**non-blocking** — `_post_run`'s body is only partly guarded, and an exception in the unguarded part kills the finisher while `handle.wait()` still reports success, destroying the error channel `wait()` exists to be.

`get_store_layout` (652), the `ok_results` comprehension (676), `_lambda_result_rows` (705) and `_output_store_kwargs` (714) all sit outside every `try`. (`_dispatch_run_stats` is internally fail-open, so it's fine.) Probed by forcing `_lambda_result_rows` to raise:

```
RuntimeError: tail blew up          # -> threading.excepthook, stderr only
wait() returned CLEANLY despite tail exception
modes: ['ping', 'setup', None, None, None, 'finalize', 'coverage']
```

`stats` (and `sweep`) skipped, the thread dead, `pool.shutdown` on line 734 never reached, and `wait()` returns `None` because `_finalize_error` is unset and `is_alive()` is False. The caller has no way to learn the tail died.

Suggest wrapping the whole `_post_run` body in `try/except BaseException` → record on the handle (a separate `_tail_error`, so `wait()` keeps re-raising the D6-load-bearing finalize failure distinctly from a fail-open rollup crash), with `finally: pool.shutdown(wait=False)`.

Related divergence worth a line of PR-body prose either way: `agg` calls `executor.finalize()` unguarded (runner.py:3445), so a finalize failure is run-fatal there and the coverage/stats/sweep tail never runs; here finalize failure is caught (672-674) and the rollups proceed. Probably the better behavior, but it's a divergence from the parity claim and isn't called out.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the `zagg.client` facade. I read `client.py` in full against `runner._run_lambda` / `_cell_work` / `_invoke_lambda_*`, ran the new suite (28 passed), and ran two out-of-band probes against a stub Lambda client.

**Verified clean** (no findings — recording so it isn't re-litigated):

- **D8 holds.** No client-side store write anywhere in `client.py`. The "event manifest write" is `runner._invoke_lambda_setup_async`, which is `InvocationType="Event"` (runner.py:4501-4505) — a worker invoke. The only client-side S3 touch in the whole tail is the *read-only* `obstore.head` inside `runner._await_run_stats_object`, reached via `_dispatch_run_stats`, exactly as on the `agg` path.
- **Retry parity holds.** `dispatch()`'s `retry=LAMBDA_RETRY` is never applied at loop level (`zagg.dispatch.dispatch` just calls `future.result()`; `LambdaExecutor.submit` does no retry) — every retry lives inside `_invoke_lambda_cell(max_retries=...)`, which the facade passes through. No lost retry pass.
- **Scope gates are real.** The raster gate is character-for-character the runner's own re-route predicate (`kind == "spatial" and (config.data_source or {}).get("reader") == "raster"`, runner.py:314); the windowing gate matches runner.py:3065-3067; `agg(events=...)` named in the temporal message exists (runner.py:144, 172).
- **`aoi_payload` is not a divergence** despite looking like one: runner omits the kwarg when `aoi_by_shard` is empty specifically to keep the event byte-identical, and `_invoke_lambda_cell` omits the key on `aoi_payload=None` (runner.py:4871) — so the facade's unconditional `aoi_by_shard.get(key)` builds the same event.
- **Benign classification is faithful.** `notebook._BENIGN_ERRORS == ("No granules found", "No data after filtering")` is exactly the tuple runner.py:3405 inlines, and the coverage/sweep `ok_results` gate uses the same `status_code == 200 and not error` predicate, so benign shards are excluded from the MOC/sweep in both paths. A real failure cannot be misclassified `ok` — the worker would have to return the literal benign string.
- **tqdm** import is strictly function-local; `tqdm>=4.66` matches the `analysis` extra's `>=` neighbors.
- 734 lines, and I found no speculative abstraction — everything on the public surface is in the ratified sketch or a thin harvest convenience.

Seven findings inline: 1 blocking, 4 non-blocking, 2 nits.

src/zagg/client.py
```diff
@@ -0,0 +1,734 @@
...
+        """One synchronous shard invoke; the pool future over this IS the API.
+
+        Mirrors ``_run_lambda``'s ``_cell_work`` payload construction (labels,
+        granule-worker clamp, leaf sub-map) minus the async result channel —
+        v1 is the legacy sync ``RequestResponse`` path by design.
+        """
+        from zagg import runner
+
+        label = runner._safe_label(self.grid, shard_key)
+        granule_urls = runner._resolve_urls(records, self.driver)
```
espg, line 694: 🤖 *from Claude (review)*

**non-blocking** — cell events are *not* byte-compatible with `_cell_work` for a `driver: https` config: this passes `self.driver` where the Lambda path hardcodes `"s3"`.

`runner._cell_work`, runner.py:3224:

```python
granule_urls = _resolve_urls(records, "s3")
```

`agg` resolves `driver` (runner.py:392) but forwards it only to `_run_local` (runner.py:429) — `_run_lambda` has no `driver` parameter at all, so on the Lambda path `data_source.driver` is ignored for URL selection and `_resolve_urls` always picks the `s3` href. The facade honors it, so the same shardmap + config yields different `granule_urls` through the two paths.

Both directions are defensible and I can't tell which is intended:

- The facade looks *more* correct: `processing/worker.py:250-254` selects `webdriver.HTTPDriver` from `config.data_source.driver`, and `processing/read.py:19-26` passes an https URL through unrewritten — so `agg`'s Lambda path hands `s3://` strings to an HTTP driver for a `driver: https` config, which looks like a latent bug on `agg`'s side.
- But the PR's central claim is byte-compatible events, and this is the one field where they demonstrably differ.

Either pin it to `"s3"` for strict parity and raise the `agg` https question separately, or keep the fix and say so explicitly in the PR body ("one deliberate divergence: the facade honors `data_source.driver`, which `_run_lambda` drops") plus a test. Today it is undocumented and untested in both directions.

src/zagg/client.py
```diff
@@ -0,0 +1,734 @@
...
+        from zagg import runner
+
+        futures = list(handle.futures.values())
+        futures_wait(futures)
+        results = []
+        for fut in futures:
+            exc = fut.exception()
+            if exc is None:
+                results.append(fut.result())
+            elif isinstance(exc, ShardError):
```
espg, line 802: 🤖 *from Claude (review)*

**non-blocking** — a shard that fails with a non-`ShardError` exception contributes no row to the run-stats parquet, so it vanishes from telemetry while `status()` still counts it `failed`.

The `elif isinstance(exc, ShardError)` means anything else — `_invoke_lambda_cell`'s payload-cap `ValueError` (runner.py:4917), `raise_for_fd_exhaustion`'s re-raise (runner.py:4992), or any botocore error escaping the retry loop — is dropped from `results` and therefore from `_lambda_result_rows`. `agg` has no equivalent hole: `zagg.dispatch.dispatch` re-raises such an exception run-fatally, so the failure is loud rather than absent.

Suggest an `else` branch synthesizing a failure-shaped dict (`{"shard_key": key, "status_code": None, "body": {}, "error": str(exc), "retries": None, ...}`) so `_lambda_result_rows`' `failure_record` path picks it up. That needs the shard key, which means iterating `handle.futures.items()` rather than `.values()` on line 646 — worth doing regardless, since the current loop discards the key it would need.

tests/test_client.py
```diff
@@ -0,0 +1,388 @@
...
+        handle.results()
+        (_, _, event) = stub.cell_events()[0]
+        assert event["shard_key"] == _WORDS[1]
+        assert event["granule_urls"] == [_rec(3)["s3"]]
+        assert event["store_path"] == _STORE
+        assert event["s3_credentials"]["accessKeyId"] == "AK"
+        assert event["config"]["output"]["grid"]["parent_order"] == 6
+        assert event["handoff"] == "arrow"
+        assert event["run_id"]
+        assert event["submap"]["granules"] == [_rec(3)]
```
espg: 🤖 *from Claude (review)*

**non-blocking** — this is the only test pinning the PR's central claim, and it asserts 8 hand-picked keys rather than the event, so it would keep passing under a real divergence.

Concretely: it passes today *despite* the `driver` divergence I flagged at `client.py:577`, because `default_config("atl06")` resolves to `driver: "s3"` and `_rec(n)` carries both hrefs. Flip the fixture config to `driver: https` and line 247 is the only assertion that would move — nothing else in the suite notices that the facade and `_cell_work` now build different events.

Two ways to make the claim self-enforcing:

1. Assert the full key **set** (`assert set(event) == {...}`) alongside the values you already check, so an added/dropped event key fails loudly.
2. Better: a differential test — drive the same config + shardmap through `runner.agg(backend="lambda", invocation="sync", ...)` against the same stub client and assert the facade's cell event equals the runner's. That's the only construction that cannot drift, and it's what makes issue #326's "CI scripts become thin wrappers so the notebook API cannot drift from what CI tests" actually enforceable rather than aspirational.

tests/test_client.py
```diff
@@ -0,0 +1,388 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _no_stats_verify(monkeypatch):
+    # The post-run tail's run-stats invoke verifies the parquet landed with a
+    # read-only poll (issue #313); zero window skips it so the finisher thread
+    # never opens a (fake) store.
+    from zagg import runner
+
+    monkeypatch.setattr(runner, "_RUN_STATS_VERIFY_WINDOW_S", 0)
```
espg, line 77: 🤖 *from Claude (review)*

**nit** — zeroing the verify window is right for the suite, but it means no test sees the tail's real latency, and `wait()`'s docstring gives callers nothing to pick a `timeout` from.

Shipped default is `_RUN_STATS_VERIFY_WINDOW_S = 20.0` (runner.py:3805), and `_dispatch_run_stats` runs a fire → verify → re-fire loop, so the stats leg alone can spend ~40 s on a *fully successful* run before failing open. Probed with a stub client at the default window: `handle.wait(timeout=30)` raised `TimeoutError: run tail still in flight after 30s` on a run where all three shards succeeded — i.e. the obvious "give it 30 seconds" choice reports a false timeout.

Suggest a sentence on `RunHandle.wait` (client.py:189-196) noting the tail's floor is dominated by the run-stats verify window and that `timeout` should exceed ~2x it, plus one test at a small non-zero window so the timing path is exercised at all rather than only at 0.

src/zagg/client.py
```diff
@@ -0,0 +1,734 @@
...
+    get_windowing,
+    load_config,
+    load_config_from_dict,
+    validate_config,
+)
+from zagg.dispatch import LAMBDA_ARCH, max_cost_usd
+from zagg.grids import from_config as grid_from_config
+from zagg.grids.morton import morton_word
+from zagg.hive import effective_store_root
+from zagg.notebook import _BENIGN_ERRORS
```
espg: 🤖 *from Claude (review)*

**nit** — importing a private name across module boundaries, and it makes `zagg.client` depend on `zagg.notebook` for one 2-tuple.

The same tuple is also inlined literally at runner.py:3405 (`elif error not in ("No granules found", "No data after filtering")`), so there are now three call sites and two definitions of the benign set. Promoting it to a shared home (`zagg.dispatch`, or a module-level constant in `zagg.runner` that both `notebook` and `client` import) would leave one definition and drop the private-name import. Purely cosmetic — the values match, so there is no behavioral finding here.

src/zagg/client.py
espg: 🤖 *from Claude*

Fixed in c32de03 — took a fourth option so the documented flow is correct **by default**: exhausting a harvest iterator joins the tail, and `wait()` stays for explicit control (a timeout, or a handle harvested some other way).

- `as_completed()` is now a generator — it yields every future, then joins the finisher and re-raises a recorded finalize failure. `progress()` (which wraps it) and `raise_first()` (on a clean run) inherit that for free. `results()` collects every outcome first, joins, *then* raises the first `ShardError`, so a shard failure no longer truncates the rollups either.
- Kept the finisher `daemon=True` and weighed it against your option (2): with the join on the happy path the only truncating flow left is a handle nobody drains, and daemon means that case cannot wedge interpreter exit. A non-daemon thread would instead make an abandoned notebook handle block exit for the tail's full duration with no way out — the more surprising of the two. Noted in the module docstring.
- Docs updated where it read as unconditional: the module-docstring example now says the loop *is* the whole run, the class docstring names which iterators join, and `as_completed` / `results` / `raise_first` each say what they do and don't join.

Regression tests mirror your probe, with a `mode_delay` stub (0.05 s per tail invoke, so a missing join fails deterministically rather than racily):

- `TestFutures::test_draining_joins_the_post_run_tail`, parametrized over `as_completed` / `progress` / `results` / `raise_first`, asserts `stub.modes()[-3:] == ["finalize", "coverage", "stats"]` with no `wait()` call anywhere in the test.
- `test_break_out_of_the_loop_leaves_the_tail_to_wait` pins the documented escape hatch (break early → `wait()` is still required).

src/zagg/client.py
espg: 🤖 *from Claude*

Fixed in 4c77714, as suggested. `_post_run` is now a thin guard around a new `_run_tail` holding the previous body:

```python
try:
    self._run_tail(handle, client, config_dict, dataset, output_creds_event, run_id)
except BaseException as e:
    handle._tail_error = e
    logger.warning(f"post-run tail failed (surfaced via handle.wait()): {e}")
finally:
    pool.shutdown(wait=False)
```

`_tail_error` is kept separate from `_finalize_error` exactly as you suggested, and `_raise_tail_error()` (used by both `wait()` and the drain-join from the sibling finding) raises finalize first — it is the D6-load-bearing failure, strictly more important than a crashed fail-open leg. `wait()`'s docstring no longer claims the tail never raises beyond finalize.

Tests:

- `test_tail_exception_surfaces_and_still_shuts_the_pool_down` forces `_lambda_result_rows` to raise, then asserts the `RuntimeError` surfaces from **both** `wait()` and a plain `results()` drain, and that the pool refuses new work (`submit` → `RuntimeError: cannot schedule new futures after shutdown`) — i.e. the `finally` ran.
- `test_finalize_failure_wins_over_a_tail_crash` records both failures and pins the precedence.

The related `agg`-vs-facade finalize divergence you flagged is now written up as PR-body **question 5** (second bullet) rather than left as loose prose, so it lands in front of a human decision.

src/zagg/client.py
espg: 🤖 *from Claude*

Confirmed, and **kept as-is by design** — the facade goes on honoring `data_source.driver`. Reproducing `_cell_work`'s hardcoded `"s3"` would be copying a bug for the sake of a parity claim: the worker picks `webdriver.HTTPDriver` from that same config key (`processing/worker.py:250-254`) and `processing/read.py:19-26` passes an https url through unrewritten, so `agg`'s Lambda path is the side that's inconsistent with its own worker.

What changed instead:

- d367a96 — `_shard_work`'s docstring now names the divergence, why it is the more correct behavior, and where it is pinned.
- 450132f — `TestRunnerParity::test_driver_https_is_the_one_deliberate_divergence` drives a `driver: https` config through **both** paths and asserts the facade sends the `https` href, `agg` sends the `s3` href, and that `granule_urls` is the *only* differing field (the rest of the event still compares equal). Untested in both directions is fixed.
- PR body **question 5** now carries it as an explicit decision for review: fix `_run_lambda` to match (a potential separate issue) or pin the facade to `"s3"` for strict parity. Not filing that issue from here — that's a side-effecting call for the repo owner, not this fold.

src/zagg/client.py
espg: 🤖 *from Claude*

Fixed in e8baf3a, taking your suggestion including the `.items()` change:

```python
for key, fut in handle.futures.items():
    exc = fut.exception()
    if exc is None:
        results.append(fut.result())
    elif isinstance(exc, ShardError):
        results.append(exc.payload)
    else:
        results.append({
            "shard_key": key,
            "status_code": None,
            "body": {},
            "error": f"{type(exc).__name__}: {exc}",
            "retries": None,
        })
```

`_lambda_result_rows` routes that through `failure_record`, so every shard contributes exactly one row. The two visibility channels were already intact — `results(return_exceptions=True)` maps `fut.exception()` whatever its type, and `raise_first()` / `result()` re-raise it as itself — so the hole was telemetry-only; the test asserts all three.

`test_transport_exception_still_gets_a_failure_row` uses a stub that raises `OSError(EMFILE, "Too many open files")` for one shard (so `raise_for_fd_exhaustion`'s re-raise escapes the retry loop for real, rather than being simulated) and asserts: `result()` raises the `OSError`, `results(return_exceptions=True)` maps it, `status()` is `{pending: 0, ok: 2, failed: 1}`, and the captured run-stats rows cover all three shard keys with the dropped one carrying `success is False` and the error string.

tests/test_client.py
espg: 🤖 *from Claude*

Fixed in 450132f — did **both**, since they catch different drift.

(1) `test_cell_event_payload_shape` now asserts the complete key set (`assert set(event) == {...}` over all 11 keys) plus full values for `submap` and `s3_credentials` instead of spot fields, so an added or dropped event key fails loudly.

(2) The differential test you asked for, `TestRunnerParity::test_cell_events_match_the_runner_lambda_path`: the same config + shardmap goes through `runner.agg(backend="lambda", invocation="sync", ...)` against a `StubLambdaClient` (injected by patching `boto3.Session`, with `compute_available_workers` / `_get_function_timeout_s` / `get_nsidc_s3_credentials` pinned, and `hive.read_manifest` stubbed so the #274 manifest checker stays read-free), then through the facade against a second stub; both event sets are keyed by `shard_key` and compared **whole**:

```python
mine.pop("run_id") and theirs.pop("run_id")   # fresh uuid per dispatch by construction
assert mine == theirs
```

It passes today, so the byte-compatibility claim is now enforced rather than asserted — and it's the construction that makes issue #326's "CI scripts become thin wrappers" checkable. The `driver` hole you found is covered by the sibling `test_driver_https_is_the_one_deliberate_divergence`, which flips the config to `driver: https` and pins `granule_urls` as the single differing field (see the reply on `client.py:577` and PR-body question 5).

tests/test_client.py
espg: 🤖 *from Claude*

Fixed in 6771405 — did both halves.

Docstring: `RunHandle.wait` now says to size `timeout` off the tail's own floor, names `runner._RUN_STATS_VERIFY_WINDOW_S` (20 s shipped), explains the fire → verify → re-fire loop that can spend ~2x it on a fully successful run, and warns that anything under ~40 s risks the false `TimeoutError` you hit.

Test: new `TestTailTiming::test_wait_timeout_must_clear_the_stats_verify_window` re-enables the window at 0.2 s (100x scaled) *over* the autouse fixture and fakes `_await_run_stats_object` with a `time.sleep(window)`-then-`False` stand-in, so the timing path runs without opening a store. It asserts your exact failure mode — `wait(timeout=window/4)` raises `TimeoutError: still in flight` on a run where every shard succeeded — then that `wait(timeout=10)` is clean and `stub.modes().count("stats") == 2` (fire, then re-fire).

Kept the autouse zero-window fixture for the rest of the suite (a real 20 s tail per test is not payable); its comment now points at `TestTailTiming` as the place the window is exercised.

src/zagg/client.py

Fixed in b69398f — promoted to `zagg.dispatch.BENIGN_ERRORS`, which leaves one definition and three importers:

- `zagg/dispatch.py` defines it next to the other dispatch-wide constants. That module imports nothing from `zagg`, and `notebook`, `runner`, and `client` all already import from it, so this adds no new edge to the import graph and no cycle risk.
- `notebook._BENIGN_ERRORS` is gone (its one use now reads `BENIGN_ERRORS`); `runner.py:3405`'s inlined literal now reads `BENIGN_ERRORS` too; `client.py` drops the private cross-module import.

Values unchanged, so no behavioral delta — `tests/test_notebook.py` and the full suite stay green.

espg:
> Fan-out width / no concurrency probe: dispatch() runs no account-concurrency preflight (unlike agg's compute_available_workers probe, which needs CloudWatch+Lambda read perms and would defeat stub-client injection). Default max_workers=64, clamped to the shard count, documented as notebook/NEON-scale. OK, or should the probe run opportunistically when no client is injected?

not sure I understand; we must have permissions to launch the lambdas, so whoever is running this in a notebook would presumably also have read permissions for cloudwatch / lambda. This might change in the future if we were to do a cryocloud scoped deployment (i.e., in that case, we wouldn't have perms, but the lambda would, and we'd have narrow perms to deploy the worker). "OK, or should the probe run opportunistically when no client is injected?" -- what does this mean? I feel like we want to get the concurrency info preflight if possible; so what are our options for doing that?

> dispatch() returns after the setup handshake (one short synchronous worker invoke; hive fail-fast-pings first) rather than backgrounding setup into the futures. This keeps the https://github.com/englacial/zagg/issues/252 fail-fast property — an incompatible store refuses before any shard is paid for — at the cost of "immediately" meaning "after ~one invoke round-trip". Acceptable reading of the ratified sketch?

yes, this is fine

> Post-run tail scope: the finisher thread mirrors agg's full lambda tail (finalize backstop always on hive — no overlapped manifest checker, so one extra idempotent invoke vs _run_lambda; coverage.moc; run-stats record; sweep trigger — all fail-open worker invokes). Alternative was to skip the rollups in v1 and leave client-run stores thinner than agg-run stores; went with parity.

ok.

> Benign outcomes: shards returning "No granules found" / "No data after filtering" resolve normally and count ok in status() (matching _run_lambda's error accounting, which counts them neither with-data nor error). Flag if they should be a third status bucket instead.

seems fine.

> data_source.driver is honored here; _run_lambda drops it. runner._cell_work hardcodes _resolve_urls(records, "s3") (runner.py:3224) — agg resolves driver but forwards it only to _run_local (runner.py:429), so _run_lambda has no driver parameter at all. The facade passes the resolved driver, so a driver: https config yields different granule_urls through the two paths (the one field where the events differ; everything else is asserted equal field-by-field in TestRunnerParity). Honoring it looks correct rather than novel: the worker selects webdriver.HTTPDriver from config.data_source.driver (processing/worker.py:250-254) and processing/read.py:19-26 passes an https url through unrewritten — i.e. agg's Lambda path hands s3:// strings to an HTTP driver for a driver: https config, which reads as a latent bug on agg's side. Decision needed: should _run_lambda be fixed to match (a separate issue — not filed from this PR), or should the facade pin "s3" for strict parity?

yes, this should be fixed-- and we might as well fix it here. The use case is that we may have s3 objects that we don't want to parse (i.e., different region, requestor pays bucket), but whose stac catalog defines https objects that we would like to substitute. In other words, even if the *vast* majority of times the lambdas are using s3, they should be *capable* of https as an ingest path if that is what is explicitly supplied as an override.

> A failed finalize is fatal on agg, fail-open-and-recorded here. agg calls executor.finalize() unguarded (runner.py:3445), so a finalize failure is run-fatal and the coverage/stats/sweep tail never runs; the facade catches it, records it on the handle (re-raised from wait() / from draining the harvest iterator), and lets the rollups proceed. Better behavior in a notebook, but it is a divergence from the parity claim.

so... is there a fix or recommendation for this? Do we want this be fatal, or should it be in a try block with a warning rather than an error? what are your thoughts, and the trade space here?

espg:
🤖 *from Claude*

Answers on the two open items, and both fixes are landing in this PR now (driver fix + concurrency preflight; finalize gets the middle option below unless you object).

## Concurrency preflight — the options

What "opportunistically when no client is injected" meant: probe by default, but skip the probe when the caller hands us a custom `lambda_client` (that's the test seam — a stub client shouldn't need to fake `GetAccountSettings`). The real option space:

- **(a) Probe with graceful degrade** ← doing this. When `max_workers` isn't explicitly passed, run `agg`'s existing `compute_available_workers` probe; on *any* failure (AccessDenied in a future narrow-perm cryocloud deployment, missing CloudWatch perms, throttle) fall back to the documented default with a visible warning naming what was denied. Explicit `max_workers=` always wins and skips the probe. This gives you preflight sizing today (notebook users have the perms, as you say) and degrades to exactly the current behavior in the scoped-deployment future — no code change needed when that lands.
- (b) Probe always, hard-fail on denial — strictest, but breaks the cryocloud case outright.
- (c) No probe, knob only — what v1 shipped; leaves fleet sizing to the user.

## Finalize: fatal vs fail-open — the trade space

What finalize *is* matters here: the hive finalize backstop is an **idempotent worker invoke** that stamps the manifest/commit. If it fails, the store's data objects are all written but the root manifest is stale — readers see the pre-run manifest until finalize is re-run. Nothing is corrupted; it's re-invokable.

- **Fatal (agg today)**: operator can't miss it, but it also *skips the rollups* — you lose the coverage/stats/sweep telemetry that would help diagnose, and in a notebook it kills the cell after all shards were already paid for. Worst of both in interactive use.
- **Fail-open + recorded (facade as shipped)**: rollups run, error re-raises from `wait()`/drain — but the user learns about it only at the end of the drain, and a user who abandons the handle could miss it entirely.
- **Middle — warn immediately, continue, still re-raise** ← recommendation, implementing: `warnings.warn` fires the moment finalize fails (notebook-visible in-stream, names the store and says "re-run: idempotent"), the tail still runs, and the recorded error still re-raises at `wait()`/drain exactly as now. Belt and braces; no silent path left.
- For `agg` itself: arguably it should run the rollups *then* exit nonzero rather than dying pre-tail, but that's an `agg` behavior change out of this PR's scope — flag if you want it as an issue.

## Driver fix

Per your ruling (STAC catalogs supplying https substitutes for unreadable s3 objects — requester-pays, cross-region): fixing `_run_lambda`'s path in this PR — `_cell_work` gets the resolved `driver` threaded through instead of the hardcoded `"s3"`, so `agg`'s Lambda path and the facade produce identical events again, both honoring `driver: https` as the explicit override. The parity test drops its carve-out and asserts full equality.

PR #334, opened 2026-07-29
Signal/noise strata + packed composition field (issue #321)

### LINKED ISSUE #321: Carry ATL03 signal confidence through aggregation: signal/noise split + t-digest encoding for signal-weighted photon flux
🤖 *from Claude*

Design/scoping issue capturing an investigation into how to carry **ATL03 photon signal confidence** through zagg's t-digest aggregation, so that a **signal-weighted photon-flux-vs-height distribution** can be produced per grid cell. Refs #265 (HHDC / photon flux — this is a sub-concern of that product), #281 (declarative t-digest), #217 (chunk-scoped digest streaming). No code yet; this is a decision-space writeup with current leans for review.

All numbers below were measured on real v007 granules (Greenland RGT 1336 across seasons, and 61 NEON passes) with `h5py`; ATBD/data-dict citations are from the v007 NSIDC docs.

---

## 1. Background findings (established)

**`signal_conf_ph` is a per-surface categorical, not a probability.** Shape `[N_photons, 5]`, int8; the 5 columns are independent signal-finding passes per surface type (order: land, ocean, sea_ice, land_ice, inland_water). Domain `{-2 possible_tep, -1 not_considered, 0 noise, 1 buffer, 2 low, 3 med, 4 high}`.

- The ordinal is misleading — there is a **hard break between 1 and 2**. The ATBD's own signal predicate is `signal_conf_ph > 1` (so `{0,1}` = background, `{2,3,4}` = signal).
- `1 = buffer` is **not** a weak-signal tier: it is the vertical padding collar added so the signal set spans ≥20 m (ATBD §5.1.1) — background photons hugging the surface.
- `-1 = not_considered` means the per-surface mask didn't run there.
- v007 also ships `weight_ph` (uint16 0–65535, DDA/RBF photon density = a genuinely continuous per-photon score) and experimental `signal_class_ph` (categorical, thresholded from `weight_ph`).

**Confidence is not reconstructable from an aggregated digest.** It is a Poisson-outlier density classification in along-track × vertical histograms at ~20 m rate, with SNR tiers `SNRbin = Hz/μ_bg` where `μ_bg` comes from the **external atmospheric-histogram stream**, plus DEM gating, per-surface tuning, and slant histogramming (ATBD §5.1.4.2). Aggregation destroys the along-track axis, the background rate, and slope — so the signal/background decision **must happen at ingest**; it cannot be recomputed post-aggregation.

**Signal fraction is driven by solar background, not surface or latitude — it ranges ~11–91%.** Same Greenland ice track across seasons:

| regime | solar elev | background | signal % (of evaluated) |
|---|---|---|---|
| night (Dec) | −33° to −40° | 0.01 MHz | ~91% |
| day (Jun) | +30° to +47° | 7–10 MHz | ~20% (up to 85% noise) |

NEON mixed passes: **40% signal / 60% background** in aggregate, bimodally spread 3%→99.6% per pass. Two earlier assumptions are therefore **retracted**: (a) "background is small, store just a count" — false in daylight, where background is the majority; (b) "signal is ~99% high, so tiny low/med lanes" — the high-fraction of signal drops to ~43% in bright daylight, with low+med climbing to ~50%.

**No clean single-surface label exists in the product.** `geolocation/surf_type` `[n_seg, 5]` is per-**segment** and **multi-hot** ("which types are possible", 0/1) — ~49% of segments flag 3 types at once. There is no per-photon surface type and no NASA-provided "primary" surface. So any per-cell surface choice is **ours to define** (a priority rule keyed on the multi-hot mask), broadcast segment→photon via the existing `ph_index_beg`/`segment_ph_cnt` link.

**The multi-column split confounds — quantified.** Using "any column > 1" vs "the land_ice column > 1", the two disagree on 2% (day) to 10% (night) of photons. ~80–86% of conflicts are soft (buffer in one surface, signal `≥2` in another — the same near-surface photon under different pad windows); ~18% of night conflicts are hard (`high` in one column, `noise` in another) ≈ 1.8% of all photons. The confound only exists because the mask flags multiple surfaces (96% concordance across land/land_ice/inland_water).

**DDA vs categorical (cross-tab, both regimes).** `signal_class_ph ≥ 1` vs categorical `> 1`: agreement 91–97%, but phi (Matthews) only 0.36–0.76 — best in daylight (0.69–0.76), i.e. where signal-weighting matters. The DDA is systematically **~7% more inclusive** (picks up the buffer-collar photons), one-directional. `weight_ph` tracks categorical confidence well in daylight (Pearson r = 0.886, monotonic noise→high) but weakly at night (r = 0.585). The DDA is surface-agnostic (one call per photon → no confound) but **experimental**.

## 2. Decision space (with current leans)

**(A) How to select signal photons at ingest.**
- **(A1) Union: `any active column > 1`** — grab all photons any surface calls signal. **← current lean.** It is an idempotent OR (order-independent), decision-free (no primary-surface rule), and reversible (a union can always be narrowed downstream; a committed single-column subset cannot be widened). Cost: includes the ~1.8% hard-contradiction + cross-surface buffer-collar photons (defensible — a real return some tuning found).
- (A2) Commit to one surface column (the product's surface). Cleaner/tighter per NASA convention, but requires the self-defined primary-surface rule and is not reversible. Recoverable as a special case of (A1) if the presence field is kept.
- (A3) DDA `signal_class_ph ≥ 1` or a `weight_ph` threshold. Confound-free single call, good daytime correlation, but experimental and ~7% looser than categorical.

**(B) Digest structure.** **Lean: two strata per cell — a signal digest and a noise/background digest** (partition at `>1`). On-demand merge reconstructs the total ("photon flux") distribution; each digest's total weight is the **exact** stratum photon count, so the signal/noise count split is free and needs no separate field. Merging two strata is a single deterministic 2-way t-digest merge (the non-associativity only bites the spatial fan-out, not this recombination). Open sub-choice: full noise digest vs. a scalar noise count when only its magnitude is wanted.

**(C) Companion 64-bit field.** **Lean: per-surface (and/or per-level low/med/high) composition, stored as quantized fractions, merged as digest-weighted means.** Because the digest already holds the exact total, the field only needs composition, not magnitude — so fractions (bounded `[0,1]`) **never saturate** regardless of cell size, and `f_merged = (N_A·f_A + N_B·f_B)/(N_A+N_B)` is an order-independent monoid using the digest weight. Counts-on-demand = `fraction × N_signal`; presence is the degenerate `fraction > 0`. ~5 surfaces × 8 bits = 40 bits, fits with room.
  - Note the per-surface counts **overlap** (multi-hot) — they are marginals, `Σ > N_signal`, and cannot split the digest into per-surface sub-distributions. They also carry little information away from surface boundaries (96% concordance). Alternatives considered: OR-presence semilattice only (cheapest, provenance without counts); 3-bit-per-column ordinal bounding interval; full per-(surface,level) presence bitset (5×7 = 35 bits, lossless single-obs).

**(D) What stays out of scope.** Reconstructing a per-surface **height distribution** requires per-surface *strata* (K digests), not a packable field — deferred unless a product needs land_ice-only-vs-land distributions distinctly.

## 3. Open questions for review

1. Is the aggregation weighting **between cells** (weight each cell by its signal content) or **within cell** (reshape the height PDF)? Between-cell is fully served by the companion field + digest weights; within-cell tunable weighting needs the two strata in (B).
2. Confirm the selection lean **(A1) union** over (A2) commit / (A3) DDA. If a specific product wants tighter surface fidelity, (A2)/(A3) can layer on.
3. Is per-surface composition (C) worth 40 bits given 96% concordance, or is OR-presence provenance enough for v1?
4. Threshold: keep `> 1` (ATBD-standard) as the signal cut, or expose it (e.g. `== 4` high-only) as a config knob?

## References
- ATL03 v007 User Guide, Data Dictionary, and ATBD r006 (NSIDC / GSFC) — §5.1 photon classification, §5.1.4.2.3.5 confidence assignment, `surf_type`/`signal_conf_ph`/`weight_ph`/`signal_class_ph` definitions.
- Related: #265 (HHDC / photon flux — parent product), #281 (declarative t-digest), #217 (chunk-scoped digest streaming), #279/#209 (t-digest merge order / ragged sharding).

espg:
Closes #321. Refs #265 (HHDC parent — the 3-state mask upgrade), #87 (located channel interplay).

## What

Implements the fully ratified #321 design ([rulings](https://github.com/englacial/zagg/issues/321#issuecomment-5123797965), [encoding spec](https://github.com/englacial/zagg/issues/321#issuecomment-5123865321)): per-cell **signal/noise stratification** of ATL03 photons committed at ingest (selection A1 — union `signal_conf > 1` over all five surface columns), as two disjoint t-digest fields whose total weights are exact stratum counts, plus the **packed uint64 composition field** (`zagg-composition/1`: five per-surface + three level lanes as quantized fractions of N_signal, presence-floor quantization), plus the threshold recorded in **store attrs**.

## Phases

- [x] 1. `column:` selector on `data_source.variables` — the mapping form `{path, column}` mirrors the structured-filter `column`, so the five `signal_conf_ph` surfaces become scalar columns off one deduped dataset read (`read.py::_variable_specs`/`_select_column`, both read paths; validated in `config._validate_ds_variables`)
- [x] 2. Reducers — `build_tdigest_where` (mask-stratified digest, delegates to `build_tdigest`, located channel passes through) and `zagg.stats.composition` (`pack_composition`, `unpack/counts_from/merge_composition`; presence floor: nonzero counts never quantize to 0, so presence is exact at every N; counts exact for N ≤ 254; merge = digest-weighted lane mean, re-quantized with the same floor)
- [x] 3. Integer scalar coercion — scalar fields with integer dtypes coerce via `int()` (the old unconditional `float()` corrupts packed words above 2**53) and empty cells take the declared integer `fill_value` (uint64 can't hold NaN); end-to-end cell-path tests including a >2**53 word round-trip. The integer declaration is **verified, not trusted**: a reducer returning a fractional value into an integer field raises naming the field, its dtype, and the value rather than letting `int()` silently truncate it (integral floats like `np.float64(3.0)` still pass); the same fail-fast applies to a non-numeric `fill_value` on an integer field
- [x] 4. Streaming/spill posture — strata + composition are **pooled / single-block-spill only** (exact via the pooled replay); the merge surface's existing gates reject both loudly, now pinned by tests. Deliberate: the streaming fold rebuilds digests from the raw source column (`(source, delta)` only), so whitelisting `build_tdigest_where` would have silently produced *unstratified* digests; the composition fold needs paired-N state that isn't wired. Fold-law follow-up flagged below.
- [x] 5. Field-level store attrs — `aggregation.variables.{name}.attrs` merges config-declared provenance onto the field's array spec in both grids (reserved `ragged` key rejected; JSON-serializability validated), plus the shipped `atl03_tdigest_strata_healpix.yaml` template
- [x] 6. Narrative doc — `docs/signal_strata.md` (+ mkdocs nav)

## How it was tested

- New: `tests/test_composition.py` (16 — exact recovery below the knee, presence floor through 20-deep merge chains, merge-vs-pooled bounded error + exact presence, strata disjointness/exhaustiveness, NaN alignment with digest weights, threshold knob); `TestStratifiedAggregation`, `TestVariableColumnSelector` (test_processing), `TestFieldAttrsTemplate` (test_grids), config validation + shipped-config tests (test_config), streaming/spill rejection pins (test_streaming/test_spill).
- Full suite: **2,680+ passed; the only 3 failures are the known pre-existing ones** (`TestConsolidationGate` ×2, `TestFunctionBuild::test_function_build_succeeds` — all reproduce on clean `main` at `d104dbb`, flagged not fixed per §4).
- `ruff check` / `ruff format --check` clean on all touched files.

## Notes for review

- The merged composition word can sit 1 LSB off the pooled word (re-quantization); the contract is bounded count error + exact presence, and the tests assert exactly that, not word equality.
- A signal photon's level lane is its **strongest** per-surface confidence — the union-consistent choice; documented in the module and doc.
- Any new aggregation key shifts `semantic_hash` — intended (new product identity).
- `config.py` was already ~2.4× the 1000-line ceiling before this PR; additions were kept minimal rather than splitting (pre-existing, flagged).

## Questions for review

1. **Fold-law follow-up scope**: multi-block spill for strata needs a masked per-block fold (re-evaluating the `where` expression over spilled columns) and the composition fold needs per-block N from the signal digest parts. Worth its own issue now, or wait until a stratified product actually hits a multi-block shard? (The located config is single-block anyway, and the ratified plan is one schema epoch = located + strata.)
2. The shipped template is **unlocated** strata; the epoch config (located + strata) composes by adding `location: leaf_id` to both digest fields. Ship that variant too, or leave composition to the epoch-flip decision?

### DIFF
--- a/docs/signal_strata.md
+++ b/docs/signal_strata.md
@@ -17,7 +17,10 @@ is **committed at ingest** and carried through the store in three pieces:
    of quantized fractions of the signal stratum — five per-surface lanes
    (`signal_conf_ph` column order: land, ocean, sea_ice, land_ice,
    inland_water) and three low/med/high lanes (a signal photon's *strongest*
-   per-surface confidence, 2/3/4).
+   per-surface confidence, 2/3/4). The level lanes are **absolute** — always
+   `conf == 2/3/4`, never renumbered against the threshold — so a product
+   committing a higher threshold ships empty lower lanes rather than shifted
+   ones, and one lane layout serves every product.
 3. **Store attrs** recording the commitment: each stratum array carries
    `stratum` + `signal_threshold`, and the composition array carries the
    versioned `composition` block (`spec`, `lanes`, `of` — the digest whose
--- a/docs/signal_strata.md
+++ b/docs/signal_strata.md
@@ -74,6 +74,18 @@ streaming state. Heavy multi-block shards need the fold-law follow-up
 ## What the mask channel gains
 
 With the strata in place, the HHDC reader's occupancy mask (issue #265)
-upgrades from 2-state to 3-state from store contents alone: **unobserved**
-(both strata empty), **observed with zero signal returns** (noise occupied,
-signal empty — the laser crossed the cell), and **observed with signal**.
+upgrades from 2-state to 3-state from store contents alone. State the rule
+against `count`, which the template already writes:
+
+- **unobserved** — `count == 0`: no observation fell in the cell.
+- **observed with zero signal returns** — `count > 0` and the signal stratum
+  is empty: the laser crossed the cell and every photon was background.
+- **observed with signal** — the signal stratum is non-empty.
+
+`count` rather than "both strata empty" because the strata are keyed to
+*finite* heights: `build_tdigest_where` and `pack_composition` both drop
+non-finite `source` rows (that is what makes the `N_signal` alignment exact),
+so a cell whose every observation carried a non-finite height would report
+`count > 0`, both strata empty, `composition == 0`. ATL03 `h_ph` is not
+nullable, so the shipped template cannot produce that cell — but the `count`
+form is the robust rule for any source and costs nothing to prefer.
--- a/docs/signal_strata.md
+++ b/docs/signal_strata.md
@@ -1,0 +1,73 @@
+# Signal/noise strata and the composition field
+
+Issue #321: ATL03 photon signal confidence cannot be reconstructed after
+aggregation (the classifier needs the along-track axis, the atmospheric
+background stream, and per-surface tuning), so the signal/background decision
+is **committed at ingest** and carried through the store in three pieces:
+
+1. **Two disjoint t-digest fields** per cell — `h_tdigest_signal` and
+   `h_tdigest_noise` — split by the ATBD signal predicate: a photon is signal
+   when **any** `signal_conf_ph` surface column clears the threshold
+   (default `>= 2`, i.e. the ATBD `> 1`; selection rule A1 — the union is
+   idempotent and narrowable downstream). Each stratum digest's total weight
+   is the **exact** stratum photon count. The total photon-flux distribution
+   is recovered at read by one deterministic two-way `merge_tdigests`.
+2. **One packed `uint64` composition word** per cell
+   (`zagg.stats.composition`, spec `zagg-composition/1`): eight 8-bit lanes
+   of quantized fractions of the signal stratum — five per-surface lanes
+   (`signal_conf_ph` column order: land, ocean, sea_ice, land_ice,
+   inland_water) and three low/med/high lanes (a signal photon's *strongest*
+   per-surface confidence, 2/3/4).
+3. **Store attrs** recording the commitment: each stratum array carries
+   `stratum` + `signal_threshold`, and the composition array carries the
+   versioned `composition` block (`spec`, `lanes`, `of` — the digest whose
+   total weight is `N_signal` — and `threshold`). Readers bind to these,
+   never to config conventions.
+
+The shipped template is `zagg/configs/atl03_tdigest_strata_healpix.yaml`. The
+five confidence columns are read from the single 2-D `signal_conf_ph` dataset
+via the per-variable `column` selector (`{path: ..., column: k}` — the
+variable analogue of the structured-filter `column`); the shared path is
+still read once.
+
+## Quantization: the presence floor
+
+Lanes quantize as `k = round(255 * c / N)` **except any nonzero count
+quantizes to at least 1**. Consequences:
+
+- `lane > 0` means "this flag occurred" **exactly, at every N**, through
+  arbitrary merge chains.
+- Count recovery `round(k * N / 255)` is exact whenever `N <= 254` — the
+  entire below-compression-knee regime (measured 99.56% of non-empty cells
+  on the live NEON store, full-mission pooling).
+- Above that, counts are within `±N/510` (plus `O(N/510)` per
+  re-quantizing merge); presence stays exact.
+- A cell with one signal photon has lanes in `{0, 255}` — the lanes *are*
+  that photon's flags.
+
+The per-surface lanes are overlapping marginals (`surf_type` is multi-hot):
+they do not sum to 255, and they cannot split the height distribution per
+surface (decision D: strata stay signal/noise, never per-surface).
+
+## Merge law
+
+`merge_composition(word_a, n_a, word_b, n_b)` folds lanes as the
+digest-weighted mean `(n_a·lane_a + n_b·lane_b)/(n_a + n_b)`, re-quantized
+with the same presence floor — an order-independent monoid whose `n` inputs
+come from the signal digests' total weights.
+
+## Operational caveat
+
+Strata and composition run on the **pooled path and the single-block spill
+regime** (exact, via the pooled replay). The streaming merge surface rejects
+them: its fold rebuilds digests from the raw source column and would silently
+ignore the `where` mask, and the composition fold is not wired into the
+streaming state. Heavy multi-block shards need the fold-law follow-up
+(issue #321).
+
+## What the mask channel gains
+
+With the strata in place, the HHDC reader's occupancy mask (issue #265)
+upgrades from 2-state to 3-state from store contents alone: **unobserved**
+(both strata empty), **observed with zero signal returns** (noise occupied,
+signal empty — the laser crossed the cell), and **observed with signal**.
--- a/docs/signal_strata.md
+++ b/docs/signal_strata.md
@@ -24,7 +24,10 @@ is **committed at ingest** and carried through the store in three pieces:
    total weight is `N_signal` — and `threshold`). Readers bind to these,
    never to config conventions.
 
-The shipped template is `zagg/configs/atl03_tdigest_strata_healpix.yaml`. The
+The shipped template is `zagg/configs/atl03_tdigest_strata_healpix.yaml` —
+**located strata is the default**: both digest fields carry `location:
+leaf_id`, so each centroid stores its order-29 morton word (an exact photon
+position at weight 1) in the `{field}_locations` sibling arrays. The
 five confidence columns are read from the single 2-D `signal_conf_ph` dataset
 via the per-variable `column` selector (`{path: ..., column: k}` — the
 variable analogue of the structured-filter `column`); the shared path is
--- a/docs/signal_strata.md
+++ b/docs/signal_strata.md
@@ -21,11 +21,13 @@ is **committed at ingest** and carried through the store in three pieces:
    `conf == 2/3/4`, never renumbered against the threshold — so a product
    committing a higher threshold ships empty lower lanes rather than shifted
    ones, and one lane layout serves every product.
-3. **Store attrs** recording the commitment: each stratum array carries
-   `stratum` + `signal_threshold`, and the composition array carries the
+3. **Store attrs** recording the commitment: each stratum's *payload* array
+   carries `stratum` + `signal_threshold`, and the composition array carries the
    versioned `composition` block (`spec`, `lanes`, `of` — the digest whose
    total weight is `N_signal` — and `threshold`). Readers bind to these,
-   never to config conventions.
+   never to config conventions. The located sibling
+   (`{field}_locations`) carries no user attrs: it is addressed through its
+   payload array, which holds the pair's provenance.
 
 The shipped template is `zagg/configs/atl03_tdigest_strata_healpix.yaml` —
 **located strata is the default**: both digest fields carry `location:
--- a/src/zagg/configs/atl03_tdigest_strata_healpix.yaml
+++ b/src/zagg/configs/atl03_tdigest_strata_healpix.yaml
@@ -21,6 +21,11 @@
 # ``signal_conf_ph`` dataset via the per-variable ``column`` selector (one
 # read, five columns -- the dataset path is deduped with the TEP filter's).
 #
+# Both strata carry the located channel (issue #87, espg ruling on PR #334:
+# located strata IS the default, not a variant): each centroid stores its
+# order-29 morton word (exact photon position at weight 1) in the row-aligned
+# ``{field}_locations`` sibling arrays.
+#
 # Strata + composition are pooled / single-block-spill only for now: the
 # streaming merge surface rejects them (the fold would ignore the ``where``
 # mask), so heavy multi-block shards need the fold-law follow-up (issue #321).
@@ -80,6 +85,7 @@ aggregation:
       kind: ragged
       function: "zagg.stats.tdigest.build_tdigest_where"
       source: h_ph
+      location: leaf_id
       inner_shape: [2]
       params:
         delta: 256
@@ -96,6 +102,7 @@ aggregation:
       kind: ragged
       function: "zagg.stats.tdigest.build_tdigest_where"
       source: h_ph
+      location: leaf_id
       inner_shape: [2]
       params:
         delta: 256
--- a/src/zagg/configs/atl03_tdigest_strata_healpix.yaml
+++ b/src/zagg/configs/atl03_tdigest_strata_healpix.yaml
@@ -17,6 +17,11 @@
 # below-compression-knee regime). The committed threshold is recorded in the
 # field's ``attrs`` so readers bind to store metadata, not this file.
 #
+# The low/med/high lanes are ABSOLUTE (conf == 2/3/4), never renumbered against
+# ``threshold``: at the default threshold 2 all three carry counts, and a
+# variant committing a higher threshold simply ships empty lower lanes rather
+# than shifted ones -- one lane layout for every product.
+#
 # The five ``signal_conf_*`` columns are read from the single 2-D
 # ``signal_conf_ph`` dataset via the per-variable ``column`` selector (one
 # read, five columns -- the dataset path is deduped with the TEP filter's).
--- a/src/zagg/configs/atl03_tdigest_strata_healpix.yaml
+++ b/src/zagg/configs/atl03_tdigest_strata_healpix.yaml
@@ -1,0 +1,138 @@
+# ATL03 photon-height t-digest, SIGNAL/NOISE STRATIFIED (issue #321).
+#
+# Two disjoint per-cell t-digests split at ingest by the ATBD signal predicate
+# -- ``signal_conf_ph > 1`` on ANY surface column (selection rule A1: the
+# union is idempotent, decision-free, and narrowable downstream; a committed
+# single-surface subset is not widenable). Each stratum's digest total weight
+# is the exact stratum photon count; the total photon-flux distribution is
+# recovered at read by ONE deterministic two-way merge (``merge_tdigests``).
+# Classification is committed here because signal confidence is NOT
+# reconstructable post-aggregation (issue #321 background).
+#
+# The ``composition`` field packs eight u8 lanes of quantized fractions of the
+# signal stratum into one uint64 per cell (``zagg.stats.composition``,
+# spec ``zagg-composition/1``): five per-surface lanes (signal_conf_ph column
+# order) + low/med/high lanes. Quantization carries a presence floor, so
+# ``lane > 0`` is exact at every N; counts recover exactly for N <= 254 (the
+# below-compression-knee regime). The committed threshold is recorded in the
+# field's ``attrs`` so readers bind to store metadata, not this file.
+#
+# The five ``signal_conf_*`` columns are read from the single 2-D
+# ``signal_conf_ph`` dataset via the per-variable ``column`` selector (one
+# read, five columns -- the dataset path is deduped with the TEP filter's).
+#
+# Strata + composition are pooled / single-block-spill only for now: the
+# streaming merge surface rejects them (the fold would ignore the ``where``
+# mask), so heavy multi-block shards need the fold-law follow-up (issue #321).
+# Grid geometry matches atl03_tdigest_healpix.yaml (o11 shard / o13 inner
+# chunk / o19 cells).
+data_source:
+  reader: h5coro
+  driver: s3
+  groups: [gt1l, gt1r, gt2l, gt2r, gt3l, gt3r]
+  coordinates:
+    latitude: "/{group}/heights/lat_ph"
+    longitude: "/{group}/heights/lon_ph"
+  variables:
+    h_ph: "/{group}/heights/h_ph"
+    signal_conf_land: {path: "/{group}/heights/signal_conf_ph", column: 0}
+    signal_conf_ocean: {path: "/{group}/heights/signal_conf_ph", column: 1}
+    signal_conf_sea_ice: {path: "/{group}/heights/signal_conf_ph", column: 2}
+    signal_conf_land_ice: {path: "/{group}/heights/signal_conf_ph", column: 3}
+    signal_conf_inland_water: {path: "/{group}/heights/signal_conf_ph", column: 4}
+  filters:
+    - dataset: "/{group}/heights/signal_conf_ph"
+      column: 0                  # possible_tep is uniform across columns
+      op: ne
+      value: -2
+  base_level: photons
+  levels:
+    photons:
+      path: "/{group}/heights"
+      coordinates: {latitude: lat_ph, longitude: lon_ph}
+      link: null
+    segments:
+      path: "/{group}/geolocation"
+      coordinates: {latitude: reference_photon_lat, longitude: reference_photon_lon}
+      link:
+        to: photons
+        index_beg: "/{group}/geolocation/ph_index_beg"
+        count: "/{group}/geolocation/segment_ph_cnt"
+        index_base: 1
+  read_plan:
+    spatial_index: segments
+    pad: 1
+
+aggregation:
+  coordinates:
+    morton:
+      dtype: uint64
+      fill_value: 0
+  variables:
+    count:
+      function: len
+      source: h_ph
+      dtype: int32
+      fill_value: 0
+    # The signal predicate below and the composition ``threshold`` MUST agree:
+    # both commit the same ``conf >= 2`` cut (i.e. the ATBD ``> 1``).
+    h_tdigest_signal:
+      kind: ragged
+      function: "zagg.stats.tdigest.build_tdigest_where"
+      source: h_ph
+      inner_shape: [2]
+      params:
+        delta: 256
+        where: >-
+          (signal_conf_land >= 2) | (signal_conf_ocean >= 2) |
+          (signal_conf_sea_ice >= 2) | (signal_conf_land_ice >= 2) |
+          (signal_conf_inland_water >= 2)
+      dtype: float32
+      fill_value: 0
+      attrs:
+        stratum: signal
+        signal_threshold: 2
+    h_tdigest_noise:
+      kind: ragged
+      function: "zagg.stats.tdigest.build_tdigest_where"
+      source: h_ph
+      inner_shape: [2]
+      params:
+        delta: 256
+        where: >-
+          ~((signal_conf_land >= 2) | (signal_conf_ocean >= 2) |
+          (signal_conf_sea_ice >= 2) | (signal_conf_land_ice >= 2) |
+          (signal_conf_inland_water >= 2))
+      dtype: float32
+      fill_value: 0
+      attrs:
+        stratum: noise
+        signal_threshold: 2
+    composition:
+      function: "zagg.stats.composition.pack_composition"
+      source: h_ph
+      dtype: uint64
+      fill_value: 0
+      params:
+        conf_land: signal_conf_land
+        conf_ocean: signal_conf_ocean
+        conf_sea_ice: signal_conf_sea_ice
+        conf_land_ice: signal_conf_land_ice
+        conf_inland_water: signal_conf_inland_water
+        threshold: 2
+      attrs:
+        composition:
+          spec: zagg-composition/1
+          lanes: [land, ocean, sea_ice, land_ice, inland_water, low, med, high]
+          of: h_tdigest_signal        # N_signal = this digest's total weight
+          threshold: 2
+
+output:
+  grid:
+    type: healpix
+    indexing_scheme: nested
+    parent_order: 11             # shard / dispatch unit: 4^(19-11) = 256x256 cells
+    chunk_inner: 13              # inner Zarr chunk: 4^(19-13) = 64x64 cells (K=16)
+    sharded: true
+    child_order: 19              # leaf cell resolution (~10 m)
+  store_layout: hive
--- a/src/zagg/grids/base.py
+++ b/src/zagg/grids/base.py
@@ -59,6 +59,22 @@ RAGGED_SPEC = "zagg-ragged/1"
 RAGGED_ZSTD_LEVEL = 3
 
 
+def apply_field_attrs(spec, meta: dict):
+    """Merge a variable's config-declared ``attrs`` onto its array spec.
+
+    ``aggregation.variables.{name}.attrs`` (issue #321) records field-level
+    provenance in the store — e.g. the composition field's lane order and the
+    signal threshold the split was committed at — so readers bind to metadata
+    rather than reconstructing config conventions. Declared keys are merged
+    over the spec's own attributes; the reserved ``ragged`` block cannot be
+    overridden (enforced at config validation).
+    """
+    attrs = meta.get("attrs")
+    if not attrs:
+        return spec
+    return spec.with_attributes({**dict(spec.attributes or {}), **attrs})
+
+
 def ragged_locations_name(field_name: str) -> str:
     """On-disk array name of a located ragged field's uint64 channel (issue #87).
 
--- a/src/zagg/grids/base.py
+++ b/src/zagg/grids/base.py
@@ -65,9 +65,17 @@ def apply_field_attrs(spec, meta: dict):
     ``aggregation.variables.{name}.attrs`` (issue #321) records field-level
     provenance in the store — e.g. the composition field's lane order and the
     signal threshold the split was committed at — so readers bind to metadata
-    rather than reconstructing config conventions. Declared keys are merged
-    over the spec's own attributes; the reserved ``ragged`` block cannot be
-    overridden (enforced at config validation).
+    rather than reconstructing config conventions.
+
+    Attrs land on the **payload array only**. A located ragged field's sibling
+    uint64 channel (:func:`ragged_locations_name`) is addressed *through* the
+    payload array, so it deliberately carries no user attrs — the provenance of
+    the pair lives on the field the config named (review finding, PR #334).
+
+    Declared keys are merged **over** the spec's own attributes, so a config
+    could in principle shadow spec-owned metadata; the reserved ``ragged``
+    block is the one key protected from that (enforced at config validation),
+    which is a denylist of exactly the one populated spec attr in tree today.
     """
     attrs = meta.get("attrs")
     if not attrs:
--- a/src/zagg/processing/aggregate.py
+++ b/src/zagg/processing/aggregate.py
@@ -9,6 +9,7 @@ DAG stays acyclic.
 
 import logging
 import os
+from functools import lru_cache
 from typing import Any
 
 import numpy as np
@@ -49,6 +50,21 @@ def _rss_log(stage: str) -> None:
     """Opt-in per-stage RSS trace (set ``ZAGG_PROFILE_RSS=1``) for #130 diagnostics."""
     if os.environ.get("ZAGG_PROFILE_RSS"):
         logger.info(f"  [rss] {stage:34s} {_rss_mb():7.0f} MB")
+
+
+@lru_cache(maxsize=256)
+def _compile_param_expr(expression: str):
+    """Cache the code object for a params expression (review finding, PR #334).
+
+    A params value that names columns is evaluated **per cell**, so passing the
+    source string to ``eval`` recompiles it once per cell per field. Measured on
+    the issue #321 strata config's 5-term ``where`` predicate: 18.9 us per
+    ``eval(str)`` vs 3.8 us per ``eval(code)`` — i.e. ~2.5 s of pure
+    recompilation per o11 shard (65,536 cells x two ``where`` fields). Compiling
+    in ``"eval"`` mode is semantically identical to ``eval`` on the string; the
+    bounded cache keys on the expression source, of which a config has a handful.
+    """
+    return compile(expression, "<zagg-param>", "eval")
 
 
 def _field_sentinel(meta: dict) -> float:
@@ -296,7 +312,7 @@ def _eval_chunk_precompute(config: PipelineConfig, pooled: dict[str, np.ndarray]
                     resolved_params[pkey] = pooled[pval]
                 elif isinstance(pval, str) and any(c in pval for c in pooled):
                     ns = {"__builtins__": {}, "np": np, "numpy": np, **pooled}
-                    resolved_params[pkey] = eval(pval, ns)  # noqa: S307
+                    resolved_params[pkey] = eval(_compile_param_expr(pval), ns)  # noqa: S307
                 else:
                     resolved_params[pkey] = pval
             value = resolve_function(meta["function"])(values, **resolved_params)
@@ -471,7 +487,8 @@ def calculate_cell_statistics(
                     "numpy": np,
                     **cell_data,
                 }
-                resolved_params[pkey] = eval(pval, ns)  # noqa: S307
+                # Cached code object, not the source string: this runs per cell.
+                resolved_params[pkey] = eval(_compile_param_expr(pval), ns)  # noqa: S307
             else:
                 resolved_params[pkey] = pval
 
--- a/src/zagg/processing/aggregate.py
+++ b/src/zagg/processing/aggregate.py
@@ -449,7 +449,7 @@ def calculate_cell_statistics(
                         f"requires a scalar result (declare 'kind: vector' to store an "
                         f"array per cell)"
                     )
-                result[name] = float(out)
+                result[name] = _coerce_scalar_value(out, sig)
             continue
 
         values = cell_data[source]
@@ -520,9 +520,23 @@ def calculate_cell_statistics(
         elif sig["kind"] == "ragged":
             result[name] = _coerce_ragged_value(out, sig)
         else:
-            result[name] = float(out)
+            result[name] = _coerce_scalar_value(out, sig)
 
     return result
+
+
+def _coerce_scalar_value(out, sig: dict):
+    """Coerce one scalar field result per its declared dtype.
+
+    Float dtypes keep the pre-#29 ``float()`` round-trip byte-for-byte. An
+    integer dtype coerces via ``int()`` instead: a float64 round-trip corrupts
+    exact integers above 2**53, which a packed 64-bit lane word (the issue
+    #321 composition field) legitimately exceeds.
+    """
+    dtype = sig.get("dtype")
+    if dtype is not None and np.issubdtype(np.dtype(dtype), np.integer):
+        return int(out)
+    return float(out)
 
 
 def _empty_cell_value(meta: dict):
@@ -551,7 +565,14 @@ def _empty_cell_value(meta: dict):
     if sig["kind"] == "vector":
         dtype = np.dtype(sig["dtype"]) if sig["dtype"] is not None else np.dtype("float32")
         return np.full(sig["trailing_shape"], _field_sentinel(meta), dtype=dtype)
-    return 0 if meta.get("function") in ("len", "count") else np.nan
+    if meta.get("function") in ("len", "count"):
+        return 0
+    # An integer-dtype scalar cannot hold NaN; empty cells take the field's
+    # declared fill_value (issue #321 — e.g. the packed composition word's 0).
+    dtype_str = sig["dtype"]
+    if dtype_str is not None and np.issubdtype(np.dtype(dtype_str), np.integer):
+        return int(meta.get("fill_value", 0))
+    return np.nan
 
 
 def _coerce_field_value(value, sig: dict) -> np.ndarray:
--- a/src/zagg/processing/aggregate.py
+++ b/src/zagg/processing/aggregate.py
@@ -79,6 +79,27 @@ def _field_sentinel(meta: dict) -> float:
     return np.nan if fill_value == "NaN" else fill_value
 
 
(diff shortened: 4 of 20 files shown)


espg:
> 3. Integer scalar coercion — scalar fields with integer dtypes coerce via int() (the old unconditional float() corrupts packed words above 253) and empty cells take the declared integer fill_value (uint64 can't hold NaN); end-to-end cell-path tests including a >253 word round-trip

I don't understand this-- when does this happen? we don't want to write anything for empty cells... so, when does this hit?

> Fold-law follow-up scope: multi-block spill for strata needs a masked per-block fold (re-evaluating the where expression over spilled columns) and the composition fold needs per-block N from the signal digest parts. Worth its own issue now, or wait until a stratified product actually hits a multi-block shard? (The located config is single-block anyway, and the ratified plan is one schema epoch = located + strata.)

explain this to me, i don't understand-- whats happening, whats new, what the problem, why is it a problem?

> The shipped template is unlocated strata; the epoch config (located + strata) composes by adding location: leaf_id to both digest fields. Ship that variant too, or leave composition to the epoch-flip decision?

yes, fold that it here too; it will be the new default, not just a variant

src/zagg/processing/read.py
```diff
@@ -239,6 +239,36 @@ def _read_segment_broadcasts(
     return out
 
 
+def _variable_specs(variables: dict) -> dict[str, tuple[str, int | None]]:
```
espg, line 242: 🤖 *from Claude (review)* — **blocking**: the `{path, column}` mapping form crashes the `inline` index backend.

`_variable_specs` normalizes the two entry forms here, but it is not the only consumer of `data_source.variables`. `src/zagg/index/inline.py:549` still assumes every value is a path-template string:

```python
paths += [tmpl.format(group=group) for tmpl in data_source["variables"].values()]
```

Reproduced against this branch:

```python
from zagg.index.inline import InlineIndex
ds = {"coordinates": {"latitude": "/{group}/heights/lat_ph",
                      "longitude": "/{group}/heights/lon_ph"},
      "variables": {"h_ph": "/{group}/heights/h_ph",
                    "conf_land": {"path": "/{group}/heights/signal_conf_ph", "column": 0}}}
InlineIndex()._prebuild_group_maps(fake_h5, "gt1l", ds)
# AttributeError: 'dict' object has no attribute 'format'
```

So the whole `column:` selector — and therefore the shipped `atl03_tdigest_strata_healpix.yaml` — is unusable the moment `data_source.index: {backend: inline}` is set, which is the compiled/pooled read arm the fleet benchmarks run on. It fails inside `_prebuild_group_maps`, i.e. before the first byte of data is read, so every granule in every shard dies.

Fix is to feed the same normalizer:

```python
from zagg.processing.read import _variable_specs
paths += [tmpl.format(group=group)
          for tmpl, _ in _variable_specs(data_source["variables"]).values()]
```

`_prebuild_group_maps` wants the *dataset paths*, and the `dict.fromkeys(paths)` dedup below already collapses the five conf columns to one chunk map — so this is behaviour-preserving for the string form. `tests/test_index.py::TestInlineReadGroup` exists but has no mapping-form case; worth one there so the coupling stays pinned.

Grepped for other consumers: `read.py:491`/`802` are wired by this PR, `catalog/extract.py:77` is a different `variables` mapping, and `semantic_core` passes `data_source` through opaquely (so the mapping form hashes fine and shifts `semantic_hash` as intended). `inline.py:549` is the only other site.

src/zagg/stats/composition.py
```diff
@@ -0,0 +1,147 @@
...
+    "merge_composition",
+    "pack_composition",
+    "unpack_composition",
+]
+
+COMPOSITION_SPEC = "zagg-composition/1"
+
+#: Lane order, LSB byte first: five surfaces (``signal_conf_ph`` column
+#: order), then the three confidence levels of the union-signal stratum.
+LANES = ("land", "ocean", "sea_ice", "land_ice", "inland_water", "low", "med", "high")
```
espg, line 59: 🤖 *from Claude (review)* — **non-blocking**: nothing pins the *byte order* of the packed word, only its self-consistency.

`_pack` and `unpack_composition` are exact mutual inverses, and every assertion in `tests/test_composition.py` goes through `unpack_composition` / `counts_from_composition`. So flipping the layout to MSB-first — `word |= lane << (8 * (7 - i))` in both functions — passes all 16 tests while breaking the ratified "bytes 0–4 / lanes 5–7" contract for any reader that implements the spec independently (moczarr, per the O3 standalone-reader work). `test_composition_word_survives_above_2_53` only pins that *some* high byte is populated, which both layouts satisfy.

One golden-word assertion closes it, reusing the case already in `test_single_photon_lanes_are_exact_flags`:

```python
conf = np.array([[4, -1, 0, 3, 1]])
word = pack_composition(np.array([10.0]), **_conf_kwargs(conf), threshold=2)
assert word == 0xFF000000FF0000FF   # land(b0), land_ice(b3), high(b7) — LSB-first
```

Verified that is the value this implementation produces; the MSB-first variant gives `0xFF0000FF000000FF`, so the assertion discriminates.

src/zagg/configs/atl03_tdigest_strata_healpix.yaml
```diff
@@ -0,0 +1,138 @@
...
+        conf_ocean: signal_conf_ocean
+        conf_sea_ice: signal_conf_sea_ice
+        conf_land_ice: signal_conf_land_ice
+        conf_inland_water: signal_conf_inland_water
+        threshold: 2
+      attrs:
+        composition:
+          spec: zagg-composition/1
+          lanes: [land, ocean, sea_ice, land_ice, inland_water, low, med, high]
+          of: h_tdigest_signal        # N_signal = this digest's total weight
```
espg, line 139: 🤖 *from Claude (review)* — **non-blocking**: the single invariant the whole field rests on is enforced only by a YAML comment.

The ratified spec ties every lane to `N_signal` = *the signal digest's total weight*, and `counts_from_composition(word, N)` is only meaningful if the `N` a reader pulls from `h_tdigest_signal` is the `N` `pack_composition` divided by. Two independent things must agree for that to hold, and neither is checked:

1. the `threshold: 2` param (line 122) vs the `>= 2` literals hard-coded in both `where` expressions (lines 86-92, 102-108) — the comment above `h_tdigest_signal` saying they MUST agree is the entire enforcement;
2. `of: h_tdigest_signal` — free-form text. A typo, or a later rename of the digest field, leaves readers dereferencing a field that does not exist; point it at the noise digest and they silently report inverted counts.

Failure mode is silent-wrong-answer, not a crash: lanes stay in range, presence stays plausible, only the recovered counts are wrong — by the signal/noise ratio.

Cheapest guards, both alongside the `attrs` validation you already added in `config.py`:

- when `attrs.composition.of` is present, require it to name another `aggregation.variables` entry of `kind: ragged` — catches (2) outright, two lines;
- when `attrs.composition.threshold` is present, require it to equal the field's own `params.threshold` — catches the attrs-vs-params half of (1).

The `where`-vs-`threshold` half is genuinely hard to validate in general (the `where` is an arbitrary expression), so at minimum state the coupling in `pack_composition`'s docstring, where it travels with the function rather than with this one template. If you'd rather make it structural the alternative is for the composition field to *emit* the predicate and the digest fields to consume it — that's a design change, so flagging rather than asking.

src/zagg/stats/composition.py
```diff
@@ -0,0 +1,147 @@
...
+    signal = (conf >= threshold).any(axis=1)
+    n = int(signal.sum())
+    if n == 0:
+        return 0
+    csig = conf[signal]
+
+    counts = np.empty(8, dtype=np.int64)
+    counts[:_SURFACES] = (csig >= threshold).sum(axis=0)
+    strongest = csig.max(axis=1)
+    for i, level in enumerate((2, 3, 4)):
```
espg, line 157: 🤖 *from Claude (review)* — **non-blocking**: the low/med/high lanes silently degenerate when `threshold != 2`.

`strongest = csig.max(axis=1)` over the *signal* rows is `>= threshold` by construction, so:

- `threshold=3` ⇒ the `low` lane is structurally always 0;
- `threshold=4` ⇒ `low` **and** `med` are always 0.

`test_threshold_knob` asserts exactly that (`lanes[5] == 0 and lanes[6] == 0` at `threshold=4`) — it pins the degenerate behaviour as expected rather than flagging it. Defensible as a definition ("levels of the union stratum"), but the ratified spec names the lanes as `conf == 2/3/4`, so a reader binding to `LANES` gets three lanes of which one or two are unusable, with no in-band signal beyond the `threshold` attr.

Related, and a real if ATL03-impossible hole: if any conf value exceeded 4, `strongest` matches none of `(2, 3, 4)` and the three level lanes sum to less than `n` while the surface lanes stay populated — a silently non-exhaustive level partition. `signal_conf_ph` is documented `-2..4` so it cannot fire on ATL03, but the reducer is a general public function with a `threshold` knob.

Suggest one of: reject `threshold not in (2, 3, 4)` for the composition field at config validation; or assert `counts[5:].sum() == n` and clamp `strongest` into `[2, 4]`. Either way the docstring at lines 96-105 should say the level lanes are threshold-relative — it currently reads as an unconditional 2/3/4 decomposition.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the six phases. Verified against the ratified [rulings](https://github.com/englacial/zagg/issues/321#issuecomment-5123797965) and [encoding spec](https://github.com/englacial/zagg/issues/321#issuecomment-5123865321), and by running code rather than reading it.

**What holds up (checked, no finding):**

- **Encoding is spec-exact.** Lane order LSB-first = `signal_conf_ph` column order then low/med/high; the per-surface lanes are `(csig >= threshold).sum(axis=0)` over the *signal* rows (overlapping marginals, correct); level lanes are the *strongest* per-surface conf, the union-consistent reading the PR body flags. Presence floor `max(1, round(255c/N))` matches. No uint64 hazard in `_pack`: the widest lane is `255 << 56 = 0xFF00000000000000` = 1.837e19 < 2**64-1, and `_quantize` clips before the `astype(np.uint64)`.
- **N alignment is exact.** The one way this could have gone wrong — `pack_composition` gating on `np.isfinite` while `build_tdigest` drops only NaN — does not apply: `build_tdigest` also gates on `np.isfinite` (`tdigest.py:245`), so ±inf and NaN drop from both sides identically. `build_tdigest_where` masks then delegates, so the digest weight is `|where ∧ finite|` and `N_signal` is `|signal ∧ finite|` — the same set at matching thresholds. Float conf columns carrying NaN also agree (`astype(np.int64)` of NaN → INT64_MIN, `>= 2` False; `NaN >= 2` False in the `where` expression).
- **Column selector ordering and blast radius.** Both read paths select the column *before* `mask_spatial` and before `keep_mask`, matching the pre-PR order; filter datasets go through `filter_paths`/`_predicate_mask` and are untouched, so filter-only 2-D datasets (`test_2d_signal_conf_filter`, `test_atl03_shipped_template_2d_signal_conf`) are unaffected. Grepped every shipped config and test: **no existing config declares a 2-D dataset as a `data_source.variables` entry**, so the new N-D-without-`column` raise breaks nothing — and a 2-D variable could never have reached a carrier anyway (`pd.DataFrame` / `Table.from_pydict` reject a 2-D column), so this converts a late confusing failure into an early clear one. Seg-broadcast columns bypassing the selector is right: they are 1-D by construction.
- **Integer coercion is behaviour-preserving.** `int(out)` and the old `float(out)` → `stats_arrays[key][i] = value` both truncate toward zero, so no rounding change; and the NaN case is not a regression — numpy 2 already raised `ValueError: cannot convert float NaN to integer` on the assignment, so the new failure is the same failure, earlier. Confirmed `stats_arrays[key][i] = <python int>` round-trips a `> 2**53` uint64 word exactly, which is the fix's whole point.
- **`attrs` plumbing.** `with_attributes` is `type(self)(**{**self.model_dump(), "attributes": ...})` — a full model round-trip — so I checked it does not perturb the rest of the spec: built the sharded HEALPix template with and without `attrs` and diffed `model_dump()` minus `attributes` for both a ragged and a uint64 scalar field: **identical** (sharding codec, `data_type`, `fill_value` all survive). Collision surface is nil in practice — dense agg-field specs carry `{}` attributes today, and `ragged` (the only populated key) is rejected at validation. All 6 wiring sites are the complete set: `get_agg_fields` is iterated exactly once per grid (`healpix.py:582`, `rectilinear.py:630`), 3 branches each.
- **Streaming/spill posture is as claimed.** `build_tdigest_where` is absent from `_TDIGEST_FUNCTIONS`, so it takes the "ragged function … has no merge law" branch; `pack_composition` takes "scalar function … is not mergeable" — both messages match the new pins. `SpillAggregator` sets `_mergeable = False`, routes to `_chunk_outputs_exact`, which replays `_pool_chunk_columns` → `_aggregate_chunk_cells` (the pooled path verbatim), and `_close_block` raises `SpillOverflowError` on threshold. On the spill schema: it is pinned by the first `append`'s whole column dict, so **all** read columns including the five conf columns spill and replay — no missing-column hazard at the pooled replay.
- **Shipped config.** YAML `>-` folds both `where` strings to single lines the params sandbox accepts (`~` / `|` on numpy bool arrays are fine); the `config.py:465-473` params-as-expression check passes via the substring branch; `test_shipped_strata_config_validates` confirms. Geometry comments check out: o11→o19 is 4^8 = 256×256 cells/shard, o13→o19 is 4^6 = 64×64 cells/chunk, K = 4^2 = 16. The TEP-filter comment matches `atl03_tdigest_healpix.yaml`'s existing wording.
- **Test suite.** Ran it: `test_composition` / `test_streaming` / `test_spill` 98 passed; `test_processing` / `test_config` / `test_grids` 587 passed; the rest 725 passed / 27 skipped with the single pre-existing `test_lambda_build` failure. `ruff check` / `format` clean on everything touched (the one `N818` hit is pre-existing in `registry.py`; `_version.py` is generated).
- Lane order *is* well covered against permutation — `test_exact_count_recovery_below_knee` compares all eight lanes against independently computed marginals over random confs, which catches any swap among the five surfaces or the three levels. See the one remaining gap below.

One blocking finding, then correctness and robustness gaps.

src/zagg/configs/atl03_tdigest_strata_healpix.yaml
```diff
@@ -0,0 +1,138 @@
...
+    # The signal predicate below and the composition ``threshold`` MUST agree:
+    # both commit the same ``conf >= 2`` cut (i.e. the ATBD ``> 1``).
+    h_tdigest_signal:
+      kind: ragged
+      function: "zagg.stats.tdigest.build_tdigest_where"
+      source: h_ph
+      inner_shape: [2]
+      params:
+        delta: 256
+        where: >-
```
espg, line 97: 🤖 *from Claude (review)* — **non-blocking (perf)**: this is the first shipped config to put a string expression on the *per-cell* params path, and that path recompiles the expression on every call.

`aggregate.py:474` resolves an expression-valued param with a bare `eval(pval, ns)` on the **source string**, inside the per-cell loop — so CPython recompiles this 5-term predicate once per cell, per field. Measured on this branch with a 40-photon cell:

```
eval(str)  18.9 us      eval(compile(...))  3.8 us      → 80% of the cost is compilation
```

At this config's geometry (o11 shard / o19 leaves = 65,536 cells) with two `where` fields that is **≈2.5 s per shard of pure recompilation**, against ≈0.5 s for the same work with a cached code object. Not fatal against the 900 s budget — read is still the wall — but it's ~2 s/shard of nothing at fleet scale, and free to remove:

```python
@functools.lru_cache(maxsize=256)
def _compiled(expr: str):
    return compile(expr, "<param>", "eval")
...
resolved_params[pkey] = eval(_compiled(pval), ns)  # noqa: S307
```

The same site at `aggregate.py:299` is `chunk_precompute` and runs once per chunk, so it doesn't matter there. Strictly the mechanism is pre-existing rather than introduced here — but this PR is what moves it onto the hot path for a product intended to run the fleet, so it reads as in scope. Happy for it to become a follow-up issue instead if you'd rather keep the diff tight.

src/zagg/processing/aggregate.py
```diff
@@ -551,7 +565,14 @@ def _empty_cell_value(meta: dict):
...
         dtype = np.dtype(sig["dtype"]) if sig["dtype"] is not None else np.dtype("float32")
         return np.full(sig["trailing_shape"], _field_sentinel(meta), dtype=dtype)
-    return 0 if meta.get("function") in ("len", "count") else np.nan
+    if meta.get("function") in ("len", "count"):
+        return 0
+    # An integer-dtype scalar cannot hold NaN; empty cells take the field's
+    # declared fill_value (issue #321 — e.g. the packed composition word's 0).
+    dtype_str = sig["dtype"]
+    if dtype_str is not None and np.issubdtype(np.dtype(dtype_str), np.integer):
+        return int(meta.get("fill_value", 0))
```
espg: 🤖 *from Claude (review)* — **nit**: this `fill_value` default disagrees with `_field_sentinel`'s, and the `"NaN"` string reaches `int()`.

`_field_sentinel` (line 62 of this file) defaults `fill_value` to the string `"NaN"`; this new branch defaults it to `0`. Both are individually reasonable — an integer field cannot hold NaN — but the divergence is the kind of thing that bites later, and the string is unhandled:

```python
_empty_cell_value({"function": "np.max", "source": "h", "dtype": "uint64", "fill_value": "NaN"})
# ValueError: invalid literal for int() with base 10: 'NaN'
```

Not a regression — such a config was already unusable, since the `np.full(shape, np.nan, dtype=uint64)` preallocation in `_aggregate_chunk_cells` raises too — so this is about message quality and the drift, not behaviour. Either route through `_field_sentinel` and special-case NaN-for-integer there, or reject `fill_value: "NaN"` on an integer dtype at config validation so the failure names the config key instead of surfacing as a stray `int()` parse error mid-shard.

Separately, and worth recording since I chased it: I verified `int(out)` is **not** a rounding change — the old `float(out)` followed by `stats_arrays[key][i] = value` truncated toward zero exactly as `int()` does, and the all-NaN-into-integer case already raised `ValueError: cannot convert float NaN to integer` on the assignment under numpy 2. So phase 3 is behaviour-preserving apart from the intended `> 2**53` fix, which I confirmed round-trips exactly.

docs/signal_strata.md
```diff
@@ -0,0 +1,73 @@
...
+them: its fold rebuilds digests from the raw source column and would silently
+ignore the `where` mask, and the composition fold is not wired into the
+streaming state. Heavy multi-block shards need the fold-law follow-up
+(issue #321).
+
+## What the mask channel gains
+
+With the strata in place, the HHDC reader's occupancy mask (issue #265)
+upgrades from 2-state to 3-state from store contents alone: **unobserved**
+(both strata empty), **observed with zero signal returns** (noise occupied,
```
espg: 🤖 *from Claude (review)* — **non-blocking**: the 3-state mask derivation is only sound when no observation has a non-finite `source`.

"**unobserved** (both strata empty)" is the load-bearing claim for the issue #265 mask upgrade, but both digests *and* the composition drop non-finite heights (`build_tdigest` gates on `np.isfinite`, `pack_composition` likewise — which is exactly what makes the N alignment exact, so this is a consequence of a correct decision, not a bug). A cell all of whose photons carry a non-finite `h_ph` therefore reports `count > 0`, both strata empty, `composition == 0` — and reads as **unobserved** under this rule, when the laser did cross it.

ATL03 `h_ph` is not nullable so this should not fire on the shipped template, but the doc is written as the reader-facing contract for the mask channel and states the derivation against store contents generally. One sentence fixes it — either note the assumption, or state the rule against `count` (`count == 0` ⇒ unobserved; `count > 0` ∧ signal stratum empty ⇒ observed-with-zero-signal), which is robust to non-finite sources and uses a field the template already writes.

While there: `tests/test_composition.py::test_disjoint_strata_weights_sum_to_finite_count` pins `n_sig + n_noise == finite count` at the reducer level, but `TestStratifiedAggregation::test_strata_are_disjoint_and_exhaustive` asserts `n_sig + n_noise == result["count"] == 40` on all-finite data — so the cell-path test would still pass if the two diverged. One NaN-height row in that fixture, asserting `n_sig + n_noise < count`, would pin the relationship the doc depends on.

src/zagg/grids/base.py
```diff
@@ -59,6 +59,22 @@
...
+    """Merge a variable's config-declared ``attrs`` onto its array spec.
+
+    ``aggregation.variables.{name}.attrs`` (issue #321) records field-level
+    provenance in the store — e.g. the composition field's lane order and the
+    signal threshold the split was committed at — so readers bind to metadata
+    rather than reconstructing config conventions. Declared keys are merged
+    over the spec's own attributes; the reserved ``ragged`` block cannot be
+    overridden (enforced at config validation).
+    """
+    attrs = meta.get("attrs")
```
espg, line 80: 🤖 *from Claude (review)* — **nit**: the located sibling gets no config attrs, which starts to matter at the epoch flip.

Both grids apply `apply_field_attrs` to the ragged payload spec only; the `members[located] = ragged_array_spec(element_dtype="uint64", **rag_kw)` line right after it is left bare. So under the located + strata epoch config (PR-body question 2), `h_tdigest_signal_locations` carries the `ragged` block but no `stratum` / `signal_threshold`, while its sibling does.

I read this as intentional — provenance belongs on the payload array and the locations channel is addressed *through* it via `ragged_locations_name` — but `docs/signal_strata.md` says "each stratum array carries `stratum` + `signal_threshold`", which a reader could take to include the companion. One clause in this docstring stating that attrs land on the payload array only would remove the ambiguity before the epoch flip makes it live.

Worth a second clause too: user attrs are merged *over* the spec's own, so the reserved-key check in `config.py` is the only thing keeping a config from shadowing spec-owned metadata. Nil surface today (I confirmed dense agg-field specs carry `{}` attributes, and `ragged` is the only populated key anywhere in the templates), but the guard is a denylist of one — if a future spec grows an attribute, configs silently win rather than erroring.

src/zagg/stats/composition.py
```diff
@@ -0,0 +1,147 @@
...
+        counts[_SURFACES + i] = int((strongest == level).sum())
+    return _pack(_quantize(counts, n))
+
+
+def merge_composition(word_a: int, n_a: int, word_b: int, n_b: int) -> int:
```
espg, line 162: 🤖 *from Claude (review)* — **non-blocking (test gap)**: nothing exercises the merge law with `n` actually sourced from a signal digest.

The ratified law is `(N_A·f_A + N_B·f_B)/(N_A + N_B)` **with `N` taken from the signal digest weights**, and both this docstring and the `of: h_tdigest_signal` attr say so. But every caller in `tests/test_composition.py` passes an `n` the test computed itself (`merge_composition(wa, n_a, wb, n_b)` with `n_a, n_b = 40, 60` and a fixture rigged so every photon is signal, i.e. `N == n`). `test_nan_heights_align_with_signal_digest_weight` is the only test that touches a real digest weight, and it never reaches `merge_composition`. So the half of the law that can silently rot — "the `n` is the digest's total weight, not the cell's observation count" — is unpinned, and there is no in-tree caller to pin it either (grepped: tests and the doc only).

A test in the shape of the eventual read-side fold would cover it end to end:

```python
d_a = build_tdigest_where(vals_a, where=sig_a)
d_b = build_tdigest_where(vals_b, where=sig_b)
n_a, n_b = int(d_a[:, 1].sum()), int(d_b[:, 1].sum())
merged = merge_composition(pack_composition(vals_a, **kw_a), n_a,
                          pack_composition(vals_b, **kw_b), n_b)
# vs pack_composition over the concatenation, recovered at n_a + n_b
```

with some non-signal *and* some NaN-height rows in each block, so `n != len(vals)` on both axes — precisely the configuration where passing the wrong `n` still yields plausible-looking lanes. Given the fold-law follow-up in question 1 will build on this monoid, pinning it now is cheap insurance.

The law itself checks out as written: identity `(0, 0)` short-circuits correctly, `(w, n)` against a zero-signal cell `(0, 0)` returns `w` (right — a cell with no signal photons contributes nothing to `N_signal`), symmetry holds, and the floor `((la > 0) | (lb > 0)) & (k == 0)` matches the ratified "nonzero input lane ⇒ nonzero output lane".

espg:
🤖 *from Claude*

**(3) Done — located strata is now the shipped default**, not a variant: `ae851c2` adds `location: leaf_id` to both digest fields in `atl03_tdigest_strata_healpix.yaml` (each centroid carries its order-29 morton word in the `{field}_locations` siblings; exact photon positions at weight 1). Doc updated; the expected-object tripwire picks up the two sibling arrays automatically (template-derived). Config validates and the template emits both siblings.

**(1) When the integer-coercion/empty-cell change hits.** Nothing new gets written for empty cells — the write behavior is unchanged. Two separate paths:

- The `int()` coercion hits every **populated** cell of an integer-dtype scalar field. The reducer returns the packed uint64 word; the old code passed every scalar result through `float(out)` before storing, and float64 can't represent integers above 2⁵³ exactly — so a composition word with any of bytes 7–8 populated would silently corrupt. This is the load-bearing fix.
- The "empty cell" part is about cells **inside a chunk that has data**. Dense arrays are written as whole chunk slabs (one 64×64 inner chunk = 4,096 slots): if any cell in the chunk has photons, all 4,096 slots are written, and the empty slots hold the fill value — exactly as today (`count` holds 0, `h_mean` holds NaN in those slots). The per-cell loop calls `calculate_cell_statistics` for every child cell including the empty ones, and `_empty_cell_value` used to return `np.nan` for any non-count scalar — but NaN cannot be assigned into a uint64 slab. Without the fix, the **first chunk containing a single photon would crash on its first empty slot**. Chunks with no data at all are still never written (the zarr `fill_value` metadata covers them) — unchanged.

**(2) The fold-law question, from the top.**

- *What's happening:* on the spill path, when a shard's photon volume exceeds the in-memory block budget, the worker closes a "block" and must **fold** per-cell partial results across blocks — a cell whose photons landed in two blocks gets its two partial digests merged.
- *The catch:* that fold layer never runs the config's reducer. It rebuilds digests itself from the raw spilled height column using only `(source, delta)` — `params` are ignored. So `h_tdigest_signal`'s `where` mask would be dropped on the fold: both strata would come back as **identical, unstratified digests of all photons — silently wrong data**. The composition word likewise has no fold wired (its merge law needs each block's N_signal).
- *What this PR does:* nothing regresses — the existing mergeability validator already classifies these reducers as non-mergeable, which yields exactly the right two behaviors, now pinned by tests: (a) if the whole shard fits one spill block, the pooled machinery replays over the complete spilled columns at the end — **exact**, byte-identical to the non-spill path; (b) if a shard is heavy enough to need a second block, the worker raises `SpillOverflowError` and aborts **loudly** instead of writing wrong strata.
- *Why it's only a future problem:* the block budget is GBs — NEON-class shards never cross it; only the heaviest 88S-crossover-belt shards would. And the located channel (now the default) already has this same single-block posture, so located-strata is one consistent posture, not two.
- *The follow-up work:* teach the spill fold to re-evaluate the `where` expression over each block's spilled columns (all read columns are spilled, so the inputs exist at fold time) and fold composition with per-block N taken from the signal-digest parts. That's what would let stratified products run pole-scale multi-block shards. Question 1 on the PR stands: file that as an issue now, or when a stratified product first needs a shard that heavy?

src/zagg/processing/read.py
espg: 🤖 *from Claude* — fixed in 21361ed. `_prebuild_group_maps` now normalizes through the same helper:

```python
from zagg.processing.read import _level_coord_paths, _variable_specs
...
paths += [
    tmpl.format(group=group)
    for tmpl, _ in _variable_specs(data_source["variables"]).values()
]
```

No import cycle to work around — the function already did a local `from zagg.processing.read import _level_coord_paths`, so this is the same import line.

Two things worth recording:

1. **Trigger is narrower than stated, but real.** `_prebuild_group_maps` is called only under `if self.write_back:` ([inline.py:628-634](https://github.com/englacial/zagg/blob/5518da3/src/zagg/index/inline.py#L628-L634)), so plain `index: {backend: inline}` reads did not crash — `backend: inline` + `write_back: true` (the manifest-building arm) did. Verified both ways: the new test's `write_back=True` case fails with `AttributeError: 'dict' object has no attribute 'format'` at `inline.py:549` on the pre-fix tree and passes after; the `write_back=False` case passes either way.
2. **One more consumer the grep missed** — not a `.format(` call, which is why it did not show up: `window_time_filters` ([config.py](https://github.com/englacial/zagg/blob/5518da3/src/zagg/config.py#L2303)) pulled `variables[time_field]` expecting a bare path, and `_validate_windowing` only checks *membership* in `variables`. A mapping-form `time_field` therefore validated and then died on the "validate_config should have rejected this configuration" guard. Fixed in the same commit by carrying the selector into both predicates (a structured filter takes the same `column` key), so a packed time column filters at base rate like any other; the string form's filter pair stays byte-identical (`column` is emitted only for the mapping form).

Tests: `TestInlineReadGroup::test_mapping_form_variables_read` (parametrized over `write_back`, asserting inline == hierarchical rows and that the two selectors dedup to **one** chunk map for `signal_conf_ph`), plus `TestWindowedCellConfig::test_mapping_form_time_field_carries_its_column`.

src/zagg/stats/composition.py
espg: 🤖 *from Claude* — added in 07e35d5, with the literal recomputed from the spec rather than copied:

```python
def test_golden_word_pins_lsb_first_byte_order(self):
    conf = np.array([[4, -1, 0, 3, 1]])
    word = pack_composition(np.array([10.0]), **_conf_kwargs(conf), threshold=2)
    assert word == 0xFF000000FF0000FF
```

Derivation check: that row is signal (land 4), `n == 1`, so counts `[1,0,0,1,0,0,0,1]` quantize to lanes `[255,0,0,255,0,0,0,255]`; lane *i* at bits `8i..8i+8` puts `0xFF` in bytes 0, 3, 7 → `0xFF000000FF0000FF`. The MSB-first layout gives `0xFF0000FF000000FF`, so the assertion discriminates — confirmed both numerically before hardcoding.

src/zagg/configs/atl03_tdigest_strata_healpix.yaml
espg: 🤖 *from Claude* — both cheap guards added in 6c55eee (`_validate_composition_attrs`, called from the `attrs` branch of the agg-variables loop), and the third half is stated in the docstring rather than parsed.

Enforced now, at load:

- `attrs.composition.of` must name a **declared** `aggregation.variables` entry of `kind: ragged` — catches the typo and the later-rename case outright. Also rejects `of` pointing at the composition field itself.
- `attrs.composition.threshold` must equal the field's own `params.threshold` — the attrs-vs-params half.

Five tests in `tests/test_config.py`: accepted, unknown `of`, non-ragged `of`, self-referential `of`, threshold mismatch, plus a non-mapping `composition` attr staying inert.

**Left by design: the `where`-vs-`threshold` half.** Agreed it is not robustly checkable — `where` is an arbitrary expression, so a regex would both miss real drift (`> 1` is the same cut as `>= 2`, as is `>= 2.0`, as is a `signal_conf_ph`-derived helper column) and reject valid predicates. Rather than ship a half-check that gets trusted, it is now stated in two places that travel with the code:

- `pack_composition`'s docstring, as a named **caller invariant**, including the failure mode ("nothing raises… counts wrong by the signal/noise ratio") and why the reducer cannot check it (it sees the conf columns, not the digest's predicate);
- `_validate_composition_attrs`'s docstring, recording *why* the third coupling is absent from the validator so a future reader does not assume it is covered.

The structural alternative you sketched — composition *emits* the predicate, digest fields consume it — is a design change to the params surface; not folding it here, leaving it standing for review as a follow-up on issue #321 alongside the fold-law work.

src/zagg/stats/composition.py
espg: 🤖 *from Claude* — documented in 95cbb96, no behavior change: the degeneracy is the ratified design, and the PR now says so in the three places a reader lands.

The lanes are **absolute, not threshold-relative**: lanes 5–7 always mean `strongest conf == 2 / 3 / 4`, so a reader binding to `LANES` never has to renumber against `threshold`. The consequence of raising the threshold is therefore *empty* lanes, not shifted ones — a high-only product carries the same lane layout as the default one, with `low`/`med` structurally 0. Stated in `pack_composition`'s docstring, the module docstring, the shipped config's header comment, and `docs/signal_strata.md`. The `test_threshold_knob` assertion is now labelled "Documented behavior, not a surprise" with the reason inline, so it no longer reads as pinning an accident.

Also documented rather than defended: confidences are ATL03's `-2..4`, so `strongest` always lands in `(2, 3, 4)` for a signal photon and the three level lanes partition the signal stratum exactly. A source exceeding 4 would fall outside all three (`counts[5:].sum() < n`) — that is out of contract for `zagg-composition/1` and now says so in the docstring.

**Declining both suggested behavior changes**, since each contradicts the ratified spec or changes the public surface:

- *reject `threshold not in (2, 3, 4)` at config validation* — the field's `threshold` is a general cut on `signal_conf_ph`; whitelisting it is a scope decision, not a bug fix. Left standing for review.
- *clamp `strongest` into `[2, 4]`* — silently folds an out-of-contract value into the `high` lane, which is worse than the currently-visible `counts[5:].sum() < n`.

src/zagg/configs/atl03_tdigest_strata_healpix.yaml
espg: 🤖 *from Claude* — folded in 0d224e0, agreeing it is in scope: this PR is what moves that path onto the fleet.

```python
@lru_cache(maxsize=256)
def _compile_param_expr(expression: str):
    return compile(expression, "<zagg-param>", "eval")
...
resolved_params[pkey] = eval(_compile_param_expr(pval), ns)  # noqa: S307
```

Applied at both params-eval sites (`calculate_cell_statistics` and `chunk_precompute`) since it is the same one-line call and the helper's docstring carries your measurement (18.9 µs → 3.8 µs, ≈2.5 s/shard at o11/o19 with two `where` fields). `compile(..., "eval")` is semantically identical to `eval` on the string, and the cache keys on the source, of which a config has a handful.

Test pins the *mechanism*, not the timing — `test_where_expression_compiles_once_across_cells` runs five cells through the strata config and asserts `_compile_param_expr.cache_info()` is `currsize == 2, misses == 2, hits == 8`, so a regression to `eval(str)` fails rather than just getting slower.

Left alone deliberately: the sibling per-cell `expression` path (`config._eval_expression_raw`, called once per cell for `kind: scalar`/`vector` expression fields) has the identical shape but is untouched by this PR and used by every config in tree, so caching it belongs in its own change rather than riding a strata PR. Noting it here so it is on the record.

docs/signal_strata.md
espg: 🤖 *from Claude* — restated in e09379f, taking the `count`-based form you suggested (robust to non-finite sources, and `count` is a field the template already writes):

- **unobserved** — `count == 0`
- **observed with zero signal returns** — `count > 0` and the signal stratum empty
- **observed with signal** — signal stratum non-empty

with a paragraph explaining *why* `count` rather than "both strata empty": the strata are keyed to **finite** heights (`build_tdigest_where` and `pack_composition` both drop non-finite rows — the thing that makes the `N_signal` alignment exact), so an all-non-finite cell would report `count > 0`, both strata empty, `composition == 0`. Noted that ATL03 `h_ph` is not nullable so the shipped template cannot produce that cell, while the `count` form is the right rule for any source.

Also took the "while there" test suggestion: `TestStratifiedAggregation::test_non_finite_heights_leave_count_above_the_strata` NaNs three of the 40 heights and asserts `n_sig + n_noise == 37 < result["count"] == 40`, so the cell-path relationship the doc depends on is pinned and the all-finite test above it cannot drift into implying the stronger claim.

src/zagg/stats/composition.py
espg: 🤖 *from Claude* — added in 5a31a69, in the shape you sketched: `test_n_is_the_signal_digest_weight` splits 100 rows into 40/60 blocks with **both** non-signal rows and non-finite heights (`values[::13] = np.nan`), asserts `0 < n < len(block)` on each so both drops actually bite, and sources `n` the way a reader would — `int(build_tdigest_where(values[sl], where=signal[sl])[:, 1].sum())`.

It then pins four things:

- digest weights are additive: `pooled_digest[:, 1].sum() == n_a + n_b`;
- merged vs pooled counts within ±1 and presence exactly equal (same contract as `test_merge_matches_pooled_recompute`);
- the pooled counts equal the **raw** per-surface counts over `signal & isfinite(values)`, so the recovered numbers are grounded in the rows rather than in another quantization;
- the negative control: `counts_from_composition(pooled, len(values))` — the observation count as divisor — is **not** equal to the correct recovery, so passing the wrong `n` fails the test rather than looking plausible.

That is the half that could silently rot, and it now fails loudly. Thanks for the note that the law itself (identity, zero-signal passthrough, symmetry, presence floor) checks out — that matches what the existing tests assert.

src/zagg/processing/aggregate.py
espg: 🤖 *from Claude* — hardened in ac54b6e. The divergence is gone and the string sentinel now fails naming the config key:

```python
def _integer_fill(meta: dict, dtype) -> int:
    fill = meta.get("fill_value", 0)
    if isinstance(fill, str):
        raise ValueError(
            f"integer field (dtype {np.dtype(dtype)}) declares fill_value {fill!r}: an "
            f"integer array cannot hold a non-numeric sentinel — declare a numeric "
            f"fill_value (e.g. 0) or a float dtype"
        )
    return int(fill)
```

Both integer paths route through it — the empty *scalar* cell (was `int(meta.get("fill_value", 0))`) and the `vector` padding branch (was `np.full(..., _field_sentinel(meta), dtype=uint64)`, which raised numpy's `cannot convert float NaN to integer` instead) — so the two agree with each other and with `_field_sentinel` on where the value comes from, and the docstring cross-references it as "`_field_sentinel` for an integer-dtype field". `test_integer_field_rejects_the_nan_sentinel` covers both paths plus a declared numeric fill coming through on both.

Related, from the same neighborhood: 5518da3 turns `_coerce_scalar_value`'s integer branch from trust-the-declaration into verify — `int`/`np.integer` pass through (so a `uint64` word above `2**53` is still exact, per your round-trip note), floats pass only when they carry no fractional part, and a fractional value raises naming the field, its dtype, and the value instead of silently truncating `2.7 → 2`.

And thanks for chasing the `int(out)`-vs-`float(out)` rounding question — recording your conclusion here so it is not re-litigated later: not a rounding change (the old `float()` + array assignment truncated toward zero identically), and the all-NaN-into-integer case already raised under numpy 2.

src/zagg/grids/base.py
espg: 🤖 *from Claude* — read the same way you did (intentional), and documented in f00764a so the ambiguity does not survive to the epoch flip. Two clauses added to `apply_field_attrs`:

> Attrs land on the **payload array only**. A located ragged field's sibling uint64 channel (`ragged_locations_name`) is addressed *through* the payload array, so it deliberately carries no user attrs — the provenance of the pair lives on the field the config named.

> Declared keys are merged **over** the spec's own attributes, so a config could in principle shadow spec-owned metadata; the reserved `ragged` block is the one key protected from that, which is a denylist of exactly the one populated spec attr in tree today.

`docs/signal_strata.md` is corrected to match — "each stratum's *payload* array carries `stratum` + `signal_threshold`", with the sibling's absence stated explicitly rather than left to inference.

**Left standing (no behavior change here):** turning the reserved-key denylist into an allowlist, or erroring when a config key collides with a spec-owned attr. It is nil surface today (confirmed: `ragged` is the only populated spec attribute anywhere in the templates), and flipping merge precedence is a config-contract decision rather than a fix — worth its own issue if a future spec grows an attribute, which is exactly the trigger the docstring now records.

PR #337, opened 2026-07-30
Provisional strata epoch benchmark target (issue #321)

espg:
Refs #321 (merged in PR #334), refs #265 (epoch A/B), follows the PR #324 pattern.

## What

The provisional benchmark arm for the strata/composition epoch config: `tdigest_healpix_o9_hive_strata` — the located strata trio + packed composition at o9 benchmark geometry, otherwise identical to the live sidecar arm (sidecar / spill / 4 GB-disk / same pinned densest NEON shard).

The A/B ladder this completes:

| arm | measures |
|---|---|
| `tdigest_healpix_o9_hive_sidecar` (live) | baseline |
| `tdigest_healpix_o9_hive_located` (PR #324) | + located channel |
| `tdigest_healpix_o9_hive_strata` (this) | + strata split + composition |

so strata-vs-located isolates the split+composition cost and strata-vs-live gives the total epoch delta — the numbers that gate the live-matrix flip to the epoch config.

Provisional-target mechanics (issue #130): PR-tree-only, runnable by explicit `--target`; not in the every-merge matrix. The expected-object model auto-derives the 7-member leaf (two digests + two location siblings + composition + count + morton). Located ⇒ non-mergeable ⇒ single-block spill, same caveat as the located arm.

## How it was tested

- Config validates (`load_config` + `validate_config`); template emits all 7 members.
- `pytest tests/test_benchmark.py tests/test_benchmark_objects.py tests/test_benchmark_shardmap.py` → 99 passed.
- targets.json parses.

## Questions for review

- **Runtime gate**: the run needs the deployed `process-shard` function at zagg ≥ 0.39.0 (`build_tdigest_where` / `pack_composition`); noted in the target `_comment`. espg is cutting 0.39.0 — the first dispatch waits for the fleet to carry it. **This is also a merge-before-run arm**: the only route that forwards a target key is `lambda-benchmark.yml`'s `workflow_dispatch` `targets` input, and that harness resolves `targets.json` + `configs/` from main's tree, so this entry is only dispatchable once it lands on main. (`/benchmark` can't reach it either way: `lambda-benchmark-command.yml` base-pins the harness to `main` and its `run-benchmark` `with:` block never forwards `targets`, so it always runs the live matrix.) So: merge first, then dispatch against a ≥ 0.39.0 fleet.


tests/data/benchmark/targets.json
```diff
@@ -95,6 +95,17 @@
       "streaming_mode": "spill",
       "worker": {"memory": 4096, "extra_disk": true}
     },
+    "tdigest_healpix_o9_hive_strata": {
+      "_comment": "Strata/composition epoch probe (issues #321/#265, PR #334): the located strata trio + packed composition at benchmark geometry, otherwise identical to the live sidecar arm. A/B ladder on the same pinned shard: live sidecar arm (baseline) -> ..._located (PR #324: + located channel) -> this (+ strata split + composition), so this-vs-located isolates the split cost and this-vs-live gives the total epoch delta. Located => non-mergeable => single-block spill (same caveat as the located arm). Run on demand via /benchmark --target or workflow_dispatch targets input; NOT in the every-merge matrix until the epoch flip is decided. Requires the deployed function at zagg >= 0.39.0 (build_tdigest_where / pack_composition).",
```
espg: 🤖 *from Claude (review)*

**[medium] the `/benchmark --target` route named here cannot reach this entry — only the `workflow_dispatch` `targets` input can.** Two independent reasons, both checkable in the workflows:

1. `lambda-benchmark-command.yml` **base-pins the harness to `main`** and overlays only `src/zagg` from the PR head:

```yaml
      - name: Checkout base harness
        uses: actions/checkout@v5
        with:
          ref: main
```

so `targets.json` + `configs/` come from `main`. While this PR is open, `run_benchmark._resolve_target` is reading a manifest with no `tdigest_healpix_o9_hive_strata` in it (`known = set(manifest.get("targets", {})) | set(manifest.get("provisional_targets", {}))`, `.github/scripts/run_benchmark.py:367`).

2. Even post-merge, that workflow **never forwards the typed targets**. `.github/actions/run-benchmark/action.yml:9` defines a `targets` input and `lambda-benchmark.yml:184` passes `targets: ${{ github.event.inputs.targets }}`, but the `/benchmark` command workflow's `with:` block (`lambda-benchmark-command.yml:154-165`) omits `targets` altogether — so it always runs `targets: ""`, i.e. the live matrix. The gate step even advertises the syntax (`` `/benchmark <targets>` still fires ``) while discarding the parsed remainder.

Net effect: an operator following this `_comment` pays for a live-matrix run and gets no strata numbers. The route that actually works pre-merge is `workflow_dispatch` on this branch with the `targets` input (`lambda-benchmark.yml:169` is a bare checkout, so it carries the branch's manifest + config). Suggest narrowing the claim to that route.

This is pre-existing wording, not introduced here — `tdigest_healpix_o9_hive_located`'s `_comment` and the block-level `provisional_targets._comment` make the same `/benchmark --target` claim — so fixing the workflow is out of scope for this PR, but it's worth capturing as a follow-up rather than propagating the claim to a third entry.

**[low] two internal wording inconsistencies in this same `_comment`:**

- "the located strata trio + packed composition" double-counts: the config header defines "the ratified strata trio" as signal digest + noise digest + composition, so "+ packed composition" makes it four.
- "so this-vs-located isolates the split cost" drops composition, contradicting "+ strata split + composition" two clauses earlier (the config header's version, "isolates the split+composition cost", is the correct one).

tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_strata.yaml
```diff
@@ -0,0 +1,140 @@
+# ATL03 t-digest, SIGNAL/NOISE STRATIFIED epoch config -- o9 HIVE benchmark arm.
+#
+# The issue #321/#265 schema-epoch candidate at benchmark geometry: identical to
+# atl03_tdigest_healpix_o9_hive.yaml in data_source read shape and output grid,
```
espg: 🤖 *from Claude (review)*

**[low] "identical … in `data_source` read shape" is true of the HDF5 reads but not of the carrier — worth stating precisely, since attributing cost is this arm's whole job.**

The dedup claim on line 15 checks out: `_variable_specs` normalizes the `{path, column}` form and `_read_paths_pooled` is fed `dict.fromkeys(...)` over the union of variable + filter paths (`src/zagg/processing/read.py:543-549` on the full-read path, `:900-909` on the planned path), so `signal_conf_ph` is fetched exactly once — the same single read the TEP filter already paid for. No extra bytes off S3. ✔

What does change is everything downstream of the read: `_select_column` materializes five `int8` columns into `data_dict`, which flow through `_concat_and_group` into `col_arrays` and are therefore written to and read back from the spill block (`SpillAggregator.flush` → `SpillBlock.append`, `src/zagg/processing/spill.py:550`). Per photon the block goes from cells(8) + leaf_id(8) + h_ph(4) = 20 B to 25 B.

Scale check, so this reads as a note and not a concern: the pinned shard's recorded `total_obs` is 3,329,701 (`origin/benchmarks:series.parquet`, latest `tdigest_healpix_o9_hive_sidecar` rows), so that's ~17 MB extra against a default `block_bytes` of ~2.7 GB (`min(0.8·0.75·4096 MB·256/3, 0.45·6144 MB)`) — i.e. ~33× headroom, and the "single-block spill is fine on this shard" claim on line 23 holds comfortably. Suggest "identical in HDF5 reads (path deduped) — the carrier gains five int8 columns" so the ladder's cost attribution is stated rather than implied.

tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_strata.yaml
```diff
@@ -0,0 +1,140 @@
...
+        index_beg: "/{group}/geolocation/ph_index_beg"
+        count: "/{group}/geolocation/segment_ph_cnt"
+        index_base: 1            # ATL03 ph_index_beg is 1-based per the v3 dict
+  read_plan:
+    spatial_index: segments
+    pad: 1                       # one segment of padding on each side per #43
+
+aggregation:
+  coordinates:
+    cell_ids:
```
espg: 🤖 *from Claude (review)*

**[nit] `cell_ids` is dead config, and it's the one undocumented divergence from the shipped strata config this header claims to mirror.**

The `cell_ids` array was retired by issue #304: it is now written only when `output.grid.emit_cell_ids: true` (`src/zagg/config.py:301-320`, `get_emit_cell_ids` at `:2199`), which this config does not set. Nothing reads `aggregation.coordinates.cell_ids` — confirmed by the template itself, which emits 7 members with no `cell_ids` among them:

```
morton, count, h_tdigest_signal, h_tdigest_signal_locations,
h_tdigest_noise, h_tdigest_noise_locations, composition
```

(`grid.shard_spec().members` on this config; `expected_object_counts(..., store_layout="hive", coverage_moc=True)` → `per_shard_min == per_shard_max == 19`, `exact: True`, so the PR body's 7-member claim and the exact #215 tripwire both check out.)

It is inherited verbatim from `atl03_tdigest_healpix_o9_hive.yaml`, so keeping it does minimize the diff against the arm being A/B'd — that's a defensible call and inert either way. But the shipped `atl03_tdigest_strata_healpix.yaml` declares `morton` only, so the header's "mirroring the shipped `atl03_tdigest_strata_healpix.yaml` at o9 instead of o11" has an extra the reader can't account for. Either drop it (byte-identical output) or name it as inherited-from-the-base-arm alongside the geometry difference. Leaving it as-is is fine too — flagging so the choice is explicit rather than accidental.

tests/data/benchmark/targets.json
espg: 🤖 *from Claude*

Fixed in ab39538 — the `_comment`'s run route now names only `workflow_dispatch`, and says the merge has to land first:

> RUN ROUTE: the lambda-benchmark.yml ``workflow_dispatch`` ``targets`` input, AFTER this lands on main -- the /benchmark slash command cannot reach a provisional target, because lambda-benchmark-command.yml base-pins the harness (targets.json + configs/) to main and its run-benchmark ``with:`` block never forwards ``targets``, so it always runs the live matrix.

Both of your reasons confirmed against the tree: `lambda-benchmark-command.yml` checks out `ref: main` for the harness, and its `with:` block (`:154-165`) passes `event`/`commit`/`ref-label`/`pr-number`/`worker-note` and no `targets`, while `lambda-benchmark.yml:184` does pass `targets: ${{ github.event.inputs.targets }}`. So `/benchmark` on this PR would have bought a live-matrix run and zero strata numbers.

Also folded the merge-before-run fact into the PR body's runtime-gate bullet, since it changes the operator sequence (merge → cut/deploy ≥ 0.39.0 → dispatch), not just the wording.

Both wording nits fixed in the same commit: "the located strata trio (signal digest + noise digest + packed composition)" (no longer four things), and "isolates the split + composition cost" (now matches the config header and the "+ strata split + composition" clause above it).

Left standing for @espg, out of scope here as you noted: the two pre-existing `_comment`s making the same `/benchmark --target` claim (`provisional_targets._comment` at `:86` and `tdigest_healpix_o9_hive_located` at `:88`), and the underlying gap — the `/benchmark` gate step advertises `` `/benchmark <targets>` `` and parses the remainder but the workflow discards it. That's either a one-line `targets:` forward in `lambda-benchmark-command.yml` (which then also needs the PR-head overlay to cover `tests/data/benchmark/`, since the harness is base-pinned) or dropping the advertised syntax. Worth its own issue rather than a drive-by in this PR.

tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_strata.yaml
espg: 🤖 *from Claude*

Fixed in 427b02f — took your suggested split, with the numbers you measured. Line 4 now claims only what's true of the reads ("identical to atl03_tdigest_healpix_o9_hive.yaml in HDF5 reads and output grid"), and the conf-column paragraph states the carrier cost outright instead of implying it:

```yaml
# The five conf columns ride the per-variable ``column`` selector off the same
# 2-D dataset the TEP filter already reads, and the read paths are deduped
# (_read_paths_pooled is fed dict.fromkeys over the variable+filter union), so
# signal_conf_ph is fetched exactly once -- no extra bytes off S3. What does grow
# is the carrier: the five int8 columns materialize into data_dict and ride the
# spill block, taking it from cells(8) + leaf_id(8) + h_ph(4) = 20 B/photon to
# 25 B/photon -- ~17 MB at this shard's recorded 3.33M photons against a
# ~2.7 GB block_bytes budget, i.e. cosmetic, but the ladder attributes it here.
```

Agreed on the framing: this is a note, not a concern — ~33× headroom, so the single-block-spill claim further down stands unchanged. Keeping it in the header matters because attributing the strata-vs-located delta is the arm's only job, and 5 B/photon is a real (if tiny) component of it.

tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_strata.yaml
espg: 🤖 *from Claude*

Leaving as-is by design, and making the choice explicit here rather than in the file: `cell_ids` stays because it is inherited verbatim from `atl03_tdigest_healpix_o9_hive.yaml`, and minimizing the diff against the arm being A/B'd is the point of this config. Dropping it would make this the one arm in the ladder whose `aggregation.coordinates` block differs from the baseline's — a divergence a future reader would have to re-derive as inert, which is a worse trade than an extra dead key.

Your analysis is right on all counts and is what makes it safe to leave: it is dead config (post-#304 the array is written only under `output.grid.emit_cell_ids: true`, which none of the three arms set), the output is byte-identical either way, and the template still emits exactly the 7 members you listed. Critically it is *identical across all three ladder arms*, so it cancels out of every delta the A/B measures — which is the property that matters for the comparison.

Not naming it in the header either, deliberately: the header's job is to explain what this arm changes relative to the baseline, and this changes nothing. Standing for @espg if you'd rather the whole ladder drop it — that's a three-config edit (baseline is a live per-merge arm), so out of scope for a provisional-target PR.

espg: 🤖 *from Claude*

Done ladder-wide per espg's ruling — `cell_ids` dropped from all three arms (base `atl03_tdigest_healpix_o9_hive.yaml`, `_located`, `_strata`) in f42cf26, so the arms stay identical and the dead declaration stops confusing readers. All three configs validate; template members confirmed 3/4/7 with no `cell_ids`; benchmark tests 99 passed. Note: a few non-ladder configs still carry the dead block (`..._o9_hive_k8.yaml`, older benchmark configs, some shipped templates) — a repo-wide purge is a natural `small-fix` if wanted rather than widening this PR.

PR #338, opened 2026-07-30
Guard agg's Lambda finalize: warn, retry once (5 s), run tail, re-raise (issue #335)

### LINKED ISSUE #335: agg Lambda path: finalize failure is fatal pre-tail — losing run-stats/rollups; guard-and-defer?
🤖 *from Claude*

Refs #326 / PR #333 (where the divergence was surfaced and espg asked for this issue: [ruling](https://github.com/englacial/zagg/pull/333#issuecomment-5124576186)).

## What happens today

On `agg`'s Lambda path, `executor.finalize()` is called **unguarded** (`runner.py:3445`). If that invoke fails, the exception is run-fatal *before the post-run tail executes* — so a finalize failure also skips:

- the `coverage.moc` rollup,
- **the run-stats record** (the failed run becomes invisible to the series/benchmark tooling — there is no row saying it happened, let alone that it failed),
- the sweep trigger.

## What finalize actually is

The hive finalize backstop is an **idempotent worker invoke** that stamps the store's root manifest/commit. On failure, nothing is corrupted: every shard's data objects are already written; the store root simply still points at the **pre-run manifest**, so readers (`moczarr.open_hive`) see the *old, internally consistent* product until finalize is re-run. Remediation is a single re-invoke.

## Why this is worth a decision

The current behavior couples two things that don't have to be coupled: *"the operator must notice"* (correct — batch runs should exit nonzero) and *"the tail must not run"* (costly — it discards exactly the telemetry that would help diagnose, and hides the run from run-stats). PR #333's client facade already decouples them for the notebook context (fail-open, immediate `warnings.warn`, tail runs, error re-raises at drain — espg-ratified). This issue is whether `agg` should follow.

## Options

**(a) Status quo** — fatal pre-tail. Simplest; operator can't miss it. Costs: no run-stats row, no coverage rollup, no sweep, harder postmortem; a cron/CI consumer sees "process died" with less to go on.

**(b) Match the facade exactly** — warn + continue + surface at the end as a recorded error but exit 0 unless the caller checks. Wrong for batch: `agg` is the CLI/automation path and must exit nonzero on a failed finalize.

**(c) Guard-and-defer (recommended)** — wrap finalize in a try; on failure: emit the warning immediately (same message contract as the facade: store path + "idempotent, re-invoke to stamp"), **run the full tail** (coverage, run-stats — which can then record the finalize failure in the run summary — and sweep), then **exit nonzero** by re-raising the recorded finalize error after the tail completes. Batch semantics preserved, telemetry preserved, run-stats finally records the failure mode.

**(d) (c) + one automatic retry** — since the backstop is idempotent, retry finalize once (short backoff) before declaring failure; most transient Lambda-side failures (throttle, cold-start timeout) self-heal, and the retry is one cheap invoke. Recommended as (c)'s companion unless there's a reason a failed finalize should never be retried in-run.

## Decision inputs

- Consumers between failure and remediation see a stale-but-consistent manifest, not corruption — so deferring the crash by one tail's duration adds no data-integrity risk.
- The run-stats gap in (a) is arguably the strongest argument for change: today the benchmark/series machinery cannot distinguish "finalize failed" from "run never happened".
- Exit-code contract: (c)/(d) keep `agg` nonzero on failure — CI and cron behavior unchanged from (a) in the success/failure signal, only richer in what got written first.

**Recommendation: (d)** — guard, warn, retry once, run tail (recording the failure in run-stats), re-raise nonzero.

espg:
Closes #335

Implements the espg-ratified **(d) + (c)** design ([approval](https://github.com/englacial/zagg/issues/335#issuecomment-5124725663), [plan](https://github.com/englacial/zagg/issues/335#issuecomment-5124735689), [backoff ruling](https://github.com/englacial/zagg/issues/335#issuecomment-5124739854)): `agg`'s Lambda-path `executor.finalize()` is no longer run-fatal pre-tail. On failure it now warns immediately, retries **once** after a **fixed 5 s** backoff (no knobs), runs the full post-run tail (coverage, run-stats — recording the failure — and sweep), then re-raises the original exception so `agg` still exits nonzero.

## What changed

- **`_finalize_with_retry(finalize, *, store_path, reraise_note, retries=1, backoff_s=None)`** (module-level in `runner.py`): guards **any zero-arg finalize callable**, so it is the single implementation for `agg`, the raster driver and the `zagg.client` facade. A retryable failure warns the diagnosis only (`... Retrying in 5 s.`); the terminal failure warns the full contract (store path + "Shard outputs are written; the store's root manifest may be stale until finalize runs again. It is idempotent — re-run the dispatch (or re-dispatch finalize) to stamp the manifest.") plus the caller's `reraise_note`, the one clause that varies. Sleeps `_FINALIZE_BACKOFF_S` (5 s, module constant so tests can collapse it) between attempts, validates `retries >= 0` (a negative value would skip finalize entirely and read as success), and **returns** the last exception instead of raising — the caller owns failure semantics. Success path untouched: no sleep, no warning.
- **Call sites**: `_run_lambda`'s tail (`executor.finalize`), **`RasterStrategy`'s Lambda driver** (`runner.py`'s `_invoke_lambda_finalize`, previously unguarded and run-fatal pre-tail — raster is not scoped out of store/telemetry parity), and **`client._run_tail`**'s finalize block. `finalize_s` still brackets the phase on the aggregation path.
- **Run-stats (durable)**: the summary dict gains **`finalize_error`** — always present, `None` on success, `"{Type}: {msg}"` on failure — on both the aggregation and raster Lambda summaries, **and it now reaches a durable consumer**: `_dispatch_run_stats` puts it on the `mode="stats"` event, the worker's `_handle_stats` forwards it to `write_run_parquet`, and it lands as one run-level parquet column (constant down the rows, like `run_id`). The write stays worker-side, D8 intact. That is what lets a postmortem distinguish "finalize failed" from "run never happened" off the parquet alone.
- **Re-raise after the tail, with D6 precedence**: the stored exception is re-raised (original object, traceback preserved) right before `return summary`, after coverage/stats/sweep dispatched exactly as on success. The whole tail sits in a `try` whose only job is precedence — a tail exception raised after a finalize failure is logged (`exc_info=True`) and chained as `__cause__`, never allowed to displace the finalize error, matching PR #333's ratified `RunHandle._raise_tail_error` behavior. With finalize clean a tail exception propagates exactly as before. CI/cron failure signal unchanged, only richer in what got written first.

## Phases

All landed as one coherent commit (the plan's five phases are one small, interlocking change):

- [x] Retry helper (`_finalize_with_retry`, importable)
- [x] Guard `agg`'s call site (warn → 5 s → one retry → record, don't raise yet)
- [x] Tail runs + run-stats records `finalize_error`
- [x] Re-raise after the tail (nonzero exit preserved)
- [x] Tests (success-path unchanged, retry-heals, both-fail full-tail + re-raise, exactly-one-retry)
- [x] **Adversarial-review fold** (6 findings, one commit each): durable run-stats record, client adoption of the shared helper, D6 tail precedence, raster Lambda parity, retry-imminent warning, `retries` validation

## Reconciliation with PR #333 — resolved here

PR #333 merged three minutes after this branch's first commit, so **#338 is the branch landing second** and adopting the helper in the client is this PR's work, not a follow-up. Done in `ed3fb7b`: `client._run_tail`'s finalize block calls `runner._finalize_with_retry` and the duplicated five-line warning text is deleted, satisfying the plan's phase-1 criterion ("one shared helper; `agg` and the client cannot drift on retry/warn semantics"). The client's recorded-error / `wait()` re-raise semantics and `_raise_tail_error`'s finalize-first precedence are unchanged.

**Integrated `main` by merge, not rebase** (`c810a65`): the phase commit was already pushed, so a rebase would have required a force-push, which CLAUDE.md §1 forbids. The merge was clean — main's `client.py`, `BENIGN_ERRORS` and `driver=` threading all outside the finalize/tail region.

One consequence worth naming: the client path now retries with the shared 5 s backoff. It costs nothing interactively — the tail already runs on the daemon finisher thread, so the backoff delays the *join*, not the notebook cell, and `wait()`'s documented timeout floor is already ~40 s for the run-stats verify window. Tests set `runner._FINALIZE_BACKOFF_S = 0`.

## How it was tested

- New `TestFinalizeGuard` in `tests/test_runner.py` (hive-path drive mirroring `TestManifestChecker._hive_finalize_calls`, tail invokes stubbed into an ordered event log, `time.sleep` patched):
  1. success path: one finalize attempt, tail order unchanged, no sleep, no finalize warning, `finalize_error is None`;
  2. fail-once-retry-succeeds: warning fired, `sleep(5)` observed, no raise, `finalize_error is None`;
  3. both attempts fail: two warnings (store path + message contract asserted), full tail observed (`coverage`/`stats`/`sweep` events), run-stats summary carries `finalize_error == "RuntimeError: invoke failed"`, and the **original** exception object re-raises after the tail;
  4. helper unit test: exactly one retry (`finalize.call_count == 2`), `sleeps == [5]`, error returned not raised.
- `TestSummaryKeysByteIdentical._LAMBDA_KEYS` extended with `finalize_error` (lambda path only).
- **Review-fold tests** (each finding gets its own): `TestDispatchRunStats.test_finalize_error_rides_the_event` (the value on the wire, fake client asserting the JSON payload); `TestStatsMode.test_finalize_error_from_event_lands_in_the_parquet` in `tests/test_lambda_handler.py` (real worker handler → parquet read-back); `TestWriteRunParquet.test_finalize_error_is_a_run_level_column`; `TestFinalizeGuard.test_finalize_error_wins_over_a_later_tail_crash` + `test_tail_crash_alone_still_propagates` (D6 precedence, both directions, `__cause__` and log asserted); `test_retries_validated_not_silently_open`; `TestRasterFinalizeGuard` in `tests/test_raster_runner.py` (success + defer-through-tail-then-re-raise, `_FakeLambdaClient` responder style); `tests/test_client.py::test_finalize_retry_heals_and_stays_clean`, with the existing client finalize tests updated for the two-attempt warning.
- Post-fold: `uv run pytest tests/test_runner.py tests/test_client.py tests/test_telemetry.py tests/test_lambda_handler.py tests/test_raster_runner.py tests/test_sweep.py tests/test_benchmark.py -q` → **621 passed, 17 skipped**. Full suite: **2768 passed, 35 skipped, 1 failed** (the same pre-existing `TestFunctionBuild::test_function_build_succeeds`, which fails locally on a pip/python-version mismatch in the build script's environment, unrelated to this diff). `ruff check` (CI's `--select=E,F,W,I --ignore=E501`) and `ruff format --check` clean on every touched file.
- `uv run pytest tests/test_runner.py -q`: **220 passed**. `pytest tests/test_benchmark.py tests/test_full_aoi_series.py -q`: 98 passed, 22 skipped.
- Full suite (`uv run pytest -q`): **2712 passed, 35 skipped, 1 failed** — the failure is the known pre-existing `TestFunctionBuild::test_function_build_succeeds` (unrelated to this diff; the two known TestConsolidationGate failures did not reproduce in this run).
- `ruff check src tests` and `ruff format --check` on the touched files: clean.
- Consumer check for the new summary key: `.github/scripts/bench_metrics.py` reads the summary via `summary.get(...)` (additive keys ignored), and `.github/scripts/update_series.py` reindexes to `bench_metrics.RECORD_COLUMNS` (`update_series.py:31`), so a new summary key never reaches the parquet series schema.

## Questions for review

1. **Summary-key and parquet-column shape** (judgment call): `finalize_error` is **always present, `None` on success**, in the summary, on the `mode="stats"` event, and as a parquet column — following the `run_stats_path` precedent in `_write_run_stats` ("keep the summary key set deterministic", issue #297), so a `duckdb`/Athena query can filter on the column across every run's parquet without a schema union. If omit-on-success is preferred, it's a small change.
2. **Raster all-failed precedence** (review-fold judgment call): on the raster path the existing all-shards-failed `RuntimeError` still raises **ahead** of the deferred finalize error, on the reasoning that it is the run's own root cause and a stale manifest is moot when no shard wrote anything (the finalize failure is durable regardless — the run-stats write precedes that raise). Say if finalize should win there too.
3. **Not folded, offered as a follow-up**: the review also suggested surfacing `finalize_error` in the bench series (`bench_metrics.build_record` + an appended `RECORD_COLUMNS` entry). Declined for now — `build_record` reads *summary* keys, and the summary never reaches a caller on the failing path, so the column would be null for exactly the runs it describes. The parquet is the postmortem surface; a bench-series column is a separate ask.
4. Pre-existing, unrelated to this diff (flagged, not fixed per conventions): codespell `statics ==> statistics` at `src/zagg/runner.py:1633` (present on `origin/main`); `ruff format --check` reformat of `tests/data/benchmark/README.md`; mypy errors in untouched regions (`runner.py:2126/2131/2212`, `notebook.py:271`, and others across followed imports); and the 3 known test failures (TestConsolidationGate ×2, TestFunctionBuild::test_function_build_succeeds).

https://claude.ai/code/session_01HW3smGgEzF9LVWhXcbNv5s

### DIFF
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -3060,6 +3060,11 @@ def _finalize_with_retry(
     ``reraise_note``, the one clause of the warning that differs between
     callers. The success path is untouched: no sleep, no warning.
 
+    Two warning shapes, one per attempt kind: a retryable failure warns the
+    diagnosis only (the remediation is not yet true — the retry may heal it),
+    and the terminal failure carries the full stale-manifest contract plus
+    ``reraise_note``.
+
     ``finalize`` is any zero-arg callable, so this is the ONE implementation of
     the warn/retry/record contract for every dispatch path: ``_run_lambda``
     (``executor.finalize``, then the full tail with the error recorded in
@@ -3077,15 +3082,29 @@ def _finalize_with_retry(
         except Exception as e:
             error = e
             retrying = attempt < retries
-            warnings.warn(
+            # A retryable attempt gets the DIAGNOSIS ONLY (review finding): the
+            # full stale-manifest remediation would be false the moment the
+            # retry heals a throttle or a cold-start timeout — exactly the case
+            # (d) was approved for — and an operator trained to ignore this
+            # warning ignores the one that matters. The contract text lands on
+            # the terminal failure, where it is true.
+            message = (
                 f"finalize (manifest backstop) failed for {store_path}: {e}. "
+                f"Retrying in {backoff_s} s."
+                if retrying
+                else f"finalize (manifest backstop) failed for {store_path}: {e}. "
                 f"Shard outputs are written; the store's root manifest may be "
                 f"stale until finalize runs again. It is idempotent — re-run "
                 f"the dispatch (or re-dispatch finalize) to stamp the manifest. "
-                + (f"Retrying once in {backoff_s} s." if retrying else reraise_note),
-                RuntimeWarning,
-                stacklevel=2,
+                f"{reraise_note}"
             )
+            # stacklevel=2 attributes this to the dispatch call site inside zagg,
+            # not the user's frame — deliberate: no fixed level can reach user
+            # code from every caller (``agg`` is several frames up, and the
+            # client facade's tail runs on a finisher THREAD whose stack has no
+            # user frame at all), so the message names the store and the remedy
+            # instead of relying on attribution.
+            warnings.warn(message, RuntimeWarning, stacklevel=2)
             if retrying:
                 time.sleep(backoff_s)
     return error
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -1418,6 +1418,7 @@ class RasterStrategy:
                     timesteps_written += body["timesteps"]
 
         wall_time = time.time() - t0
+        finalize_error = None
         if store_layout == "hive":
             # Finalize-backstop-before-all-failed-raise, mirroring the local
             # backend (where ensure_manifest + root coverage run ahead of the
@@ -1428,17 +1429,24 @@ class RasterStrategy:
             # fires (natural gating); the finalize backstop still runs.
             #
             # Finalize backstop (issue #252 hybrid): idempotent ensure_manifest
-            # worker-side — required reader-facing schema (D6), so a failure
-            # raises, unlike the fail-open coverage dispatch below.
-            _invoke_lambda_finalize(
-                client,
-                function_name,
-                store_path,
-                output_creds_event=output_creds_event,
-                config_dict=config_dict,
-                dataset=dataset,
-                parent_order=int(grid.parent_order),
-                overwrite=overwrite,
+            # worker-side. Guarded exactly like the aggregation path (issue #335,
+            # same shared helper — raster is not scoped out of store/telemetry
+            # parity): warn, retry once, then DEFER so the tail below (coverage,
+            # the run-stats record that carries the failure, sweep) still runs
+            # and the error re-raises at the end.
+            finalize_error = _finalize_with_retry(
+                lambda: _invoke_lambda_finalize(
+                    client,
+                    function_name,
+                    store_path,
+                    output_creds_event=output_creds_event,
+                    config_dict=config_dict,
+                    dataset=dataset,
+                    parent_order=int(grid.parent_order),
+                    overwrite=overwrite,
+                ),
+                store_path=store_path,
+                reraise_note="This error re-raises after the post-run tail completes.",
             )
             if get_coverage_moc(config):
                 # Root coverage.moc (issue #200 phase 3): one fire-and-forget
@@ -1486,7 +1494,12 @@ class RasterStrategy:
             )
             for key, error, body in stats_failures
         ]
-        # D8 worker-invoke transport (issue #313): mirrors the spatial path.
+        # D8 worker-invoke transport (issue #313): mirrors the spatial path,
+        # deferred finalize failure included (issue #335) — this write is what
+        # makes it durable, and it happens BEFORE the all-failed raise below.
+        finalize_error_str = (
+            None if finalize_error is None else f"{type(finalize_error).__name__}: {finalize_error}"
+        )
         run_stats_path = _dispatch_run_stats(
             client,
             function_name,
@@ -1496,6 +1509,7 @@ class RasterStrategy:
             result_prefix=result_prefix,
             output_creds_event=output_creds_event,
             store_kwargs=_output_store_kwargs(output_creds_event, region),
+            finalize_error=finalize_error_str,
         )
         # End-of-run rollup sweep (issue #300): D8 worker-invoke transport,
         # mirroring the aggregation lambda path — one fire-and-forget
@@ -1516,6 +1530,11 @@ class RasterStrategy:
                     logger.info(f"Dispatched rollup sweep ({len(leaves)} leaves, fire-and-forget)")
             except Exception as e:
                 logger.warning(f"rollup sweep dispatch failed (fail-open, D9): {e}")
+        # Precedence note (issue #335): the all-failed verdict still raises
+        # FIRST — it is the run's own root cause, not a tail crash, and a stale
+        # manifest is moot when no shard wrote anything. The finalize failure is
+        # already durable (the run-stats write above) and is what re-raises when
+        # some shards did land.
         if cells and errors == len(cells):
             raise RuntimeError(f"all {errors} raster shard(s) failed; last error: {last_error}")
         # Worker telemetry rollup (issue #250): billed durations and peak RSS,
@@ -1562,6 +1581,10 @@ class RasterStrategy:
             "store_path": store_path,
             "backend": "lambda",
             "run_stats_path": run_stats_path,
+            # Guarded finalize (issue #335): always present, None on success —
+            # same key and same meaning as the aggregation lambda summary, so
+            # the two paths' summaries keep one shape for this field.
+            "finalize_error": finalize_error_str,
         }
         if profile:
             # Straggler-maxed stage seconds (+ the write bucket) and summed
@@ -1580,6 +1603,11 @@ class RasterStrategy:
             f"Done (lambda): {shards_with_data}/{len(cells)} shards, "
             f"{errors} errors, {wall_time:.1f}s"
         )
+        # Deferred finalize failure (issue #335): the tail ran with the error
+        # recorded in run-stats; re-raise the original so agg still exits
+        # nonzero, exactly as the aggregation lambda path does.
+        if finalize_error is not None:
+            raise finalize_error
         return summary
 
 
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -3013,39 +3013,48 @@ def _run_local(
     return summary
 
 
-def _finalize_with_retry(executor, *, store_path, retries=1, backoff_s=5) -> Exception | None:
-    """Guarded ``executor.finalize()``: warn at failure, retry, return the error.
+#: Fixed backoff between the finalize attempts (issue #335): not a user knob
+#: (espg ruling), a module constant so tests can collapse the wait.
+_FINALIZE_BACKOFF_S = 5
+
+
+def _finalize_with_retry(
+    finalize, *, store_path, reraise_note, retries=1, backoff_s=None
+) -> Exception | None:
+    """Guarded finalize backstop: warn at failure, retry, return the error.
 
     Issue #335 ((d)+(c), espg-ratified): the finalize backstop is an idempotent
     worker invoke — nothing is corrupted on failure (every shard's outputs are
     already written; only the root manifest may be stale), so a failure warns
-    immediately (same message contract as the PR #333 client facade, so the two
-    paths read identically) and retries once after a fixed 5 s backoff —
-    deliberately not knobs (espg ruling on issue #335). Returns ``None`` on
-    success or the last exception; the caller owns failure semantics
-    (``_run_lambda`` runs the full post-run tail with the error recorded in
-    run-stats, then re-raises so ``agg`` still exits nonzero). The success path
-    is untouched: no sleep, no warning.
+    immediately and retries once after a fixed 5 s backoff — deliberately not
+    knobs (espg ruling on issue #335). Returns ``None`` on success or the last
+    exception; the caller owns failure semantics and describes them in
+    ``reraise_note``, the one clause of the warning that differs between
+    callers. The success path is untouched: no sleep, no warning.
+
+    ``finalize`` is any zero-arg callable, so this is the ONE implementation of
+    the warn/retry/record contract for every dispatch path: ``_run_lambda``
+    (``executor.finalize``, then the full tail with the error recorded in
+    run-stats before it re-raises), ``RasterStrategy``'s Lambda driver, and the
+    :class:`zagg.client.Run` facade's finisher thread (records it on the handle
+    for :meth:`~zagg.client.RunHandle.wait`). Sharing it is the issue #335
+    phase-1 acceptance criterion — ``agg`` and the client cannot drift.
     """
     error = None
+    backoff_s = _FINALIZE_BACKOFF_S if backoff_s is None else backoff_s
     for attempt in range(retries + 1):
         try:
-            executor.finalize()
+            finalize()
             return None
         except Exception as e:
             error = e
             retrying = attempt < retries
-            tail = (
-                f"Retrying once in {backoff_s} s."
-                if retrying
-                else "This error re-raises after the post-run tail completes."
-            )
             warnings.warn(
                 f"finalize (manifest backstop) failed for {store_path}: {e}. "
                 f"Shard outputs are written; the store's root manifest may be "
                 f"stale until finalize runs again. It is idempotent — re-run "
                 f"the dispatch (or re-dispatch finalize) to stamp the manifest. "
-                f"{tail}",
+                + (f"Retrying once in {backoff_s} s." if retrying else reraise_note),
                 RuntimeWarning,
                 stacklevel=2,
             )
@@ -3519,7 +3528,11 @@ def _run_lambda(
     # run-stats recording the failure, sweep) runs as on success and the error
     # re-raises at the end so agg still exits nonzero.
     finalize_start = time.time()
-    finalize_error = _finalize_with_retry(executor, store_path=store_path)
+    finalize_error = _finalize_with_retry(
+        executor.finalize,
+        store_path=store_path,
+        reraise_note="This error re-raises after the post-run tail completes.",
+    )
     finalize_s = time.time() - finalize_start
     wall_time = time.time() - start_time
 
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -2315,6 +2315,7 @@ def _dispatch_run_stats(
     store_kwargs=None,
     summary=None,
     inline_rows=None,
+    finalize_error=None,
 ) -> str | None:
     """Fire-and-forget worker-invoke run-record write (issue #313, D8).
 
@@ -2334,6 +2335,12 @@ def _dispatch_run_stats(
     ``run_stats_path`` names the real object — best-effort: the invoke is
     fire-and-forget, mirroring the root MOC's semantics. Fail-open
     everywhere: telemetry must never fail a run whose data landed.
+
+    ``finalize_error`` rides the event (always present, ``None`` on success)
+    and lands as a run-level parquet column (issue #335): the guarded finalize
+    defers its failure through the tail, and this is the durable record a
+    postmortem reads — the alternative is a signal that lives only in the
+    orchestrator's ``RuntimeWarning``. The write stays worker-side (D8).
     """
     from datetime import datetime, timezone
 
@@ -2350,6 +2357,10 @@ def _dispatch_run_stats(
             "store_path": store_path,
             "run_id": run_id,
             "timestamp": timestamp,
+            # Always on the wire, None on success (issue #335) — same
+            # determinism rule as summary["run_stats_path"] below, so the
+            # parquet's column set does not vary run to run.
+            "finalize_error": finalize_error,
         }
         if output_creds_event is not None:
             event["output_credentials"] = output_creds_event
@@ -3696,6 +3707,10 @@ def _run_lambda(
         store_kwargs=_output_store_kwargs(output_creds_event, region),
         summary=summary,
         inline_rows=stats_inline_rows,
+        # The deferred finalize failure becomes DURABLE here (issue #335): the
+        # worker stamps it as the run parquet's run-level column, so the
+        # postmortem signal outlives this process's RuntimeWarning + exit code.
+        finalize_error=summary["finalize_error"],
     )
     # End-of-run rollup sweep (issue #300): the Lambda dispatcher never PUTs
     # (D8 standing rule), so the sweep rides ONE fire-and-forget mode="sweep"
--- a/src/zagg/runner.py
+++ b/src/zagg/runner.py
@@ -3535,246 +3535,272 @@ def _run_lambda(
     )
     finalize_s = time.time() - finalize_start
     wall_time = time.time() - start_time
+    # The whole post-run tail is inside this try purely for PRECEDENCE (issue
+    # #335 fold): whatever it raises must not displace a finalize failure. The
+    # legs themselves keep their own fail-open guards; this is the blanket that
+    # keeps the unguarded code BETWEEN them (row building, statistics, the log
+    # f-strings) from swallowing the D6 error the next refactor introduces.
+    try:
+        # End-of-run root coverage.moc (issue #200 phase 3; default-on for hive,
+        # O9): the dispatcher cannot PUT to S3, so it builds + serializes the MOC
+        # and posts ONE fire-and-forget worker invoke that GET-unions-PUTs it.
+        # Transport rationale (espg-requested, plan question 3) — serialized
+        # ranges IN the event vs the completion list via the status channel:
+        #   - Ranges (chosen): the dispatcher already holds the completion list
+        #     in memory, so building the MOC costs milliseconds, and the payload
+        #     is bounded by construction — spatially coherent coverage collapses
+        #     to a few-KB range list, far under Lambda's 256 KB async-invoke cap,
+        #     which a raw ~50k-key completion list would break. One hop, and no
+        #     read-back race against status objects still landing from retried
+        #     stragglers.
+        #   - Completion list via .status/ (rejected): payload size would be
+        #     run-independent and the artifact replayable from durable state, but
+        #     it costs the worker a LIST + N GETs, races in-flight status writes,
+        #     and its replayability is already owned by the §7 sweep's
+        #     authoritative rebuild — the leaves are the durable truth (D9).
+        # Fail-open everywhere: a failed build/invoke logs and the run result is
+        # untouched (the root MOC is a regenerable cache).
+        if get_store_layout(config) == "hive" and get_coverage_moc(config):
+            try:
+                from zagg.hive import build_root_coverage
+                from zagg.windows import union_time_range
 
-    # End-of-run root coverage.moc (issue #200 phase 3; default-on for hive,
-    # O9): the dispatcher cannot PUT to S3, so it builds + serializes the MOC
-    # and posts ONE fire-and-forget worker invoke that GET-unions-PUTs it.
-    # Transport rationale (espg-requested, plan question 3) — serialized
-    # ranges IN the event vs the completion list via the status channel:
-    #   - Ranges (chosen): the dispatcher already holds the completion list
-    #     in memory, so building the MOC costs milliseconds, and the payload
-    #     is bounded by construction — spatially coherent coverage collapses
-    #     to a few-KB range list, far under Lambda's 256 KB async-invoke cap,
-    #     which a raw ~50k-key completion list would break. One hop, and no
-    #     read-back race against status objects still landing from retried
-    #     stragglers.
-    #   - Completion list via .status/ (rejected): payload size would be
-    #     run-independent and the artifact replayable from durable state, but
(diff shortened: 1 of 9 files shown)


src/zagg/runner.py
```diff
@@ -3593,6 +3638,12 @@ def _accumulate(report, i, result):
         "setup_s": setup_s,
         "fanout_s": fanout_s,
         "finalize_s": finalize_s,
+        # Always-present, None on success (issue #335): the run-stats record can
+        # finally distinguish "finalize failed" from "run never happened", and
+        # the key set stays deterministic (the run_stats_path precedent).
+        "finalize_error": (
```

**blocking** — the new `finalize_error` key reaches **no durable consumer**, so phase 3's stated goal ("run-stats records the failure … the series tooling can finally distinguish 'finalize failed' from 'run never happened'") is not met end-to-end. I traced all three routes off this key:

1. **Run-stats parquet (Lambda path):** `_dispatch_run_stats` never puts `summary` on the wire — the event is `{mode, store_path, run_id, timestamp}` plus `rows`/`rows_from` (`runner.py:2340-2363`); `summary` is used only to set `summary["run_stats_path"]` (`runner.py:2334-2335`, `2391`). The worker's `_handle_stats` calls `write_run_parquet(event["store_path"], rows, …)` (`deployment/aws/lambda_handler.py:1020-1026`), and `write_run_parquet` takes rows only — one row per *shard*, no run-level record at all (`telemetry.py:381-389`). No column anywhere can carry a finalize failure.
2. **Bench series:** `bench_metrics.build_record` reads named summary keys and never `finalize_error`, and `update_series.py:31` reindexes to `RECORD_COLUMNS` (which omits it). The PR body cites this as proof the parquet is *unaffected* — true, but it is the same reason the series still cannot see the failure.
3. **In memory:** on the failure path `_run_lambda` raises at `runner.py:3734`, so the `summary` dict holding the key is never returned to any caller. The only thing that can observe it today is a test that monkeypatches `_dispatch_run_stats` (`tests/test_runner.py:4265`) — i.e. the assertion at `tests/test_runner.py:4283` proves a property no production consumer can read.

The PR's real win stands and is worth keeping: the tail now runs, so a finalize-failed run produces a stats parquet *at all* instead of vanishing. But "finalize failed" specifically is recorded nowhere durable, and as implemented the always-present key is dead weight.

Options: (a) put it on the wire — add `finalize_error` to the `mode="stats"` event and write a run-level record/sidecar; (b) surface it in the bench path — `build_record` + an appended `RECORD_COLUMNS` entry (the schema is documented append-only, `bench_metrics.py:79-104`); (c) keep the key but restate the claim in the PR body/docstrings as "the run is now *visible* in run-stats", drop "records the failure", and say plainly that the operator-visible signal is the RuntimeWarning plus the nonzero exit.

src/zagg/runner.py
```diff
@@ -2993,6 +2994,47 @@ def _accumulate(report, i, outcome):
     return summary
 
 
+def _finalize_with_retry(executor, *, store_path, retries=1, backoff_s=5) -> Exception | None:
```

**blocking** — **PR #333 merged at 2026-07-30T00:34:30Z, three minutes after this commit** (`origin/main` now contains `src/zagg/client.py`). The PR body's premise ("`client.py` only exists on the unmerged PR #333 branch … PR #333's `client._run_tail` adopts this same helper as a one-line follow-up on whichever branch merges second") therefore resolves *against* this PR: **#338 is the branch landing second**, so adopting the helper in `client._run_tail` is now this PR's work, not a follow-up on someone else's branch.

That matters because it is an explicit acceptance criterion of the ratified plan ([phase 1 / criteria](https://github.com/englacial/zagg/issues/335#issuecomment-5124735689)): "One shared helper; `agg` and the client cannot drift on retry/warn semantics." On `origin/main` today `client._run_tail` still (i) has **no retry** at all and (ii) duplicates the message verbatim across five lines (`src/zagg/client.py:815-853` on main — try at 815, `except` at 834, `warnings.warn` at 844-853). Merging #338 as-is ships exactly the divergence the plan set out to prevent, with the added trap that the two copies of the message contract can now drift independently.

Rebase onto main is otherwise clean here: main's only `runner.py` changes since the merge base (`driver=` threading, `BENIGN_ERRORS`) are outside the finalize/tail region. Suggested scope: rebase, have `client._run_tail`'s finalize block call `_finalize_with_retry(…, store_path=self.store)` and set `handle._finalize_error` from the return value, and delete the duplicated warn text. If instead the retry is deliberately *not* wanted in the notebook path (a 5 s stall in an interactive drain is a defensible reason), that is a design fork for @espg — say so in the body rather than leaving phase 1 silently unmet.

src/zagg/runner.py
```diff
@@ -3675,6 +3726,12 @@ def _accumulate(report, i, result):
         breakdown = ", ".join(f"{phase} {secs:.0f}s" for phase, secs in worker_phase_max.items())
         logger.info(f"Worker phases (max across cells): {breakdown}")
     _log_container_stats(container_stats)
+    # Deferred finalize failure (issue #335): the tail above ran to completion
+    # with the error recorded in run-stats; re-raise the original exception
+    # (traceback preserved on the stored object) so agg exits nonzero — the
+    # CI/cron failure signal is unchanged, only richer in what got written first.
+    if finalize_error is not None:
+        raise finalize_error
```
espg, line 3863: 🤖 *from Claude (review)*

**non-blocking** — **tail-error precedence diverges from #333's D6 choice, and the finalize error can be silently lost.** The re-raise is unconditionally last, so any exception raised anywhere between the guarded call (`runner.py:3490`) and here wins outright and `finalize_error` is discarded. The three *rollup dispatches* are fail-open, so this is not hypothetical-only from them (`_dispatch_run_stats` is wholly wrapped, `runner.py:2338`/`2407`; coverage and sweep have their own `except Exception` logs) — but the code *between* them is not guarded:

- `stats_rows, stats_inline_rows = _lambda_result_rows(report.results, run_id=run_id)` (`runner.py:3666`) is unguarded and walks arbitrary worker envelopes;
- the summary literal at `runner.py:3645` calls `str()` on the stored exception, so an exception whose `__str__` raises replaces the finalize error *and* skips the whole stats/sweep tail;
- `statistics.pstdev`, `_container_telemetry_summary`, and the log f-strings between 3600-3733 are likewise unguarded.

PR #333 (now on main) made the opposite call deliberately: `client.py` keeps `_tail_error` "distinct from the D6-load-bearing `_finalize_error`" and `RunHandle.wait()` raises `_finalize_error` **first**. Two sibling paths with opposite precedence for the same failure pair is the kind of thing that gets discovered during an incident. Either match D6 (compute+raise in a `try/finally`, or attach the tail error as `__context__`) or state the chosen precedence in the comment here so the divergence reads as deliberate. Note the diagnostic cost of losing it: the operator gets a stats/summary traceback with the actual root cause surviving only in the earlier `RuntimeWarning`.

src/zagg/runner.py
```diff
@@ -3440,9 +3482,12 @@ def _accumulate(report, i, result):
...
 
     # Consolidate metadata via Lambda (same rationale as setup -- avoids
-    # requiring orchestrator-side S3 access).
+    # requiring orchestrator-side S3 access). Guarded (issue #335): a failed
+    # finalize warns + retries once, then DEFERS — the tail below (coverage,
+    # run-stats recording the failure, sweep) runs as on success and the error
+    # re-raises at the end so agg still exits nonzero.
     finalize_start = time.time()
-    executor.finalize()
+    finalize_error = _finalize_with_retry(executor, store_path=store_path)
```

**non-blocking** — **the raster Lambda path is untouched and now diverges.** `RasterStrategy`'s Lambda driver calls `_invoke_lambda_finalize` **unguarded** at `runner.py:1425`, with a comment asserting the old semantics verbatim ("required reader-facing schema (D6), so a failure raises, unlike the fail-open coverage dispatch below"). So for `agg`'s raster target a finalize failure is still fatal *before* its coverage dispatch, its run-stats write (whose `run_stats_path` is folded into the raster summary at `runner.py:1556`), and the sweep — and that summary gains no `finalize_error` key, so the raster and spatial Lambda summaries now have different shapes for the same field.

Every argument from #335 applies identically (idempotent `ensure_manifest` backstop, shard outputs already written, remediation is one re-invoke), and this repo's standing precedent is that store/telemetry changes are not scoped away from raster. If it is deliberately out of scope, that belongs as an explicit line in the PR body plus a follow-up issue rather than silence; if it isn't, it is a two-line change here (`_finalize_with_retry` + defer to the existing raise site).

src/zagg/runner.py
```diff
@@ -2993,6 +2994,47 @@ def _accumulate(report, i, outcome):
...
+            return None
+        except Exception as e:
+            error = e
+            retrying = attempt < retries
+            tail = (
+                f"Retrying once in {backoff_s} s."
+                if retrying
+                else "This error re-raises after the post-run tail completes."
+            )
+            warnings.warn(
```

**nit** — **the first-attempt warning fires the full remediation contract even when the retry is about to heal it.** `test_retry_succeeds_warns_sleeps_no_raise` (`tests/test_runner.py:4249-4262`) pins exactly that: a run that ends completely clean still emits a `RuntimeWarning` telling the operator the store's root manifest "may be stale until finalize runs again … re-run the dispatch (or re-dispatch finalize) to stamp the manifest". For the transient case (throttle, cold-start timeout — the whole reason espg approved (d)) that statement is false by the time the run returns, and it will train operators to ignore the warning that matters.

Suggestion: on the retryable attempt emit only the diagnosis ("finalize attempt 1 failed for {store}: {e}. Retrying in 5 s."), and keep the full stale-manifest/idempotent contract for the terminal failure. The "same message contract as the #333 facade" argument only really binds the terminal message — #333 has no retry, so there is no precedent being broken.

Minor, same line: `stacklevel=2` resolves to `_run_lambda`'s own call line (the helper adds a frame the #333 inline `warnings.warn` does not have), so the attribution is zagg-internal either way — fine, just not the caller as the parameter usually implies.

src/zagg/runner.py
```diff
@@ -2993,6 +2994,47 @@ def _accumulate(report, i, outcome):
...
+    immediately (same message contract as the PR #333 client facade, so the two
+    paths read identically) and retries once after a fixed 5 s backoff —
+    deliberately not knobs (espg ruling on issue #335). Returns ``None`` on
+    success or the last exception; the caller owns failure semantics
+    (``_run_lambda`` runs the full post-run tail with the error recorded in
+    run-stats, then re-raises so ``agg`` still exits nonzero). The success path
+    is untouched: no sleep, no warning.
+    """
+    error = None
+    for attempt in range(retries + 1):
```
espg, line 3085: 🤖 *from Claude (review)*

**nit** — `retries` is unvalidated, and one bad value fails **silently open**: `retries=-1` makes `range(0)` empty, so the loop body never runs, `executor.finalize()` is **never called**, and the helper returns `None` — which the call site reads as success. `retries=0` behaves correctly (one attempt, no sleep, error returned). Since the helper is deliberately module-level and importable for other paths to adopt (see the #333 note above), a `retries = max(retries, 0)` or an assert is cheap insurance against a caller passing a computed value.

The "returns the LAST exception" contract is documented (`runner.py:3002-3005`), and both failures are warned individually, so a first-then-second divergence is at least visible in the log — no action needed there.

src/zagg/runner.py

Fixed via option (a) — the key now reaches a durable consumer end to end, worker-side (D8 intact). Commit `1e716ce`.

The chain you traced, closed at each hop:

1. **On the wire** — `_dispatch_run_stats` gained `finalize_error` and puts it in the `mode="stats"` event unconditionally (`None` on success, the same determinism rule as `summary["run_stats_path"]`, so the parquet's column set can't vary run to run).
2. **Worker write** — `_handle_stats` forwards `event.get("finalize_error")` to `write_run_parquet` (absent on a pre-#335 dispatcher → `None`, so an old dispatcher against a new worker is fine).
3. **Parquet** — `write_run_parquet` stamps it as one run-level column, constant down the rows like `run_id`/`n_shards` already are:
```python
    df = pd.DataFrame(rows)
    # Run-level (issue #335): constant down the column, None on a clean run.
    df["finalize_error"] = finalize_error
```
Always written, so `duckdb`/Athena can `WHERE finalize_error IS NOT NULL` across every run's parquet without a schema union. `object_encoding="utf8"` already pins all-null string columns (the `invoked_by` precedent), and the pyarrow round-trip is covered.
4. **Call site** — `_run_lambda` passes `finalize_error=summary["finalize_error"]`, i.e. the failure becomes durable *before* the re-raise. Same for the raster driver (see the parity thread).

Tests assert each hop separately rather than one monkeypatched seam: `TestDispatchRunStats.test_finalize_error_rides_the_event` (fake client, asserts the JSON payload), `TestStatsMode.test_finalize_error_from_event_lands_in_the_parquet` (real worker handler → real parquet read-back), `TestWriteRunParquet.test_finalize_error_is_a_run_level_column`, and `TestFinalizeGuard` now asserts the value handed to the dispatch (`captured["stats_finalize_error"]`) on both the clean and failed paths.

I did **not** take option (b) as well — `bench_metrics.RECORD_COLUMNS` reads *summary* keys for the benchmark series, and the summary is unreachable on the failing path (your point 3), so appending a column there would still be dead for exactly the runs it's meant to describe. The parquet is the postmortem surface. If you want the bench series to carry it too that's a follow-up worth its own issue.

Note this touches `deployment/aws/lambda_handler.py` — worker application code (the `mode="stats"` handler added by issue #313), not the infra/deploy scripts CLAUDE.md §1 fences off, and the D8 rule is why the write has to live there.

src/zagg/runner.py

Correct, and done — #338 lands second, so this is this PR's work. Commit `ed3fb7b`. Integrated main with **`git merge origin/main`** (commit `c810a65`), not a rebase: a rebase of the single phase commit would rewrite pushed history and force a `--force-with-lease`, which CLAUDE.md §1 forbids outright. The merge was clean; `client.py`, `dispatch.py`'s `BENIGN_ERRORS`, and the `driver=` threading all came across untouched.

`client._run_tail`'s finalize block now routes through the shared helper, and the duplicated five-line message text is deleted:

```python
        if layout == "hive" or get_consolidate_metadata(self.config):
            error = runner._finalize_with_retry(
                lambda: runner._invoke_lambda_finalize(
                    client, self.function_name, self.store,
                    output_creds_event=output_creds_event, **finalize_kwargs,
                ),
                store_path=self.store,
                reraise_note=(
                    "This error also re-raises from handle.wait() / draining the harvest iterator."
                ),
            )
            if error is not None:
                handle._finalize_error = error
                logger.warning(f"finalize invoke failed (surfaced via handle.wait()): {error}")
```

To make that possible the helper now takes **any zero-arg callable** instead of an executor, and `reraise_note` is the single clause that varies by caller — everything else (the warn text, the retry, the "return, don't raise" contract) is shared by construction. `_run_lambda` passes `executor.finalize`; the raster driver passes its own lambda.

Semantics preserved exactly: the error is still recorded on the handle, `wait()`/drain still re-raise it, and `_raise_tail_error`'s finalize-before-tail precedence is untouched. All 45 `tests/test_client.py` tests stay green, plus one new: `test_finalize_retry_heals_and_stays_clean` (transient failure → warned, `_finalize_error is None`, nothing re-raised, and the retry's invoke really lands).

On the interactive-stall concern you flagged as a possible design fork: I kept the retry, because the tail already runs on the daemon finisher thread — the 5 s delays the *join*, not the notebook cell. Only an explicit `wait()` that happens to overlap the backoff pays it, and `wait()`'s documented floor is already ~40 s for the run-stats verify window, so 5 s is inside the existing budget. Noted in the `_run_tail` docstring and the PR body. Tests set `runner._FINALIZE_BACKOFF_S = 0` (new module constant — same pattern as `_RUN_STATS_VERIFY_WINDOW_S`) so no suite pays it.

PR body's Reconciliation note updated to say this landed here.

src/zagg/runner.py
espg: 🤖 *from Claude*

Matched to D6 — the finalize error now wins. Commit `63327ad`.

The whole post-run tail sits inside one `try` whose only job is precedence, so the fix survives the refactor that introduces the *next* unguarded statement rather than patching the four you enumerated:

```python
    except Exception as tail_error:
        # D6 precedence (issue #335 fold), matching the PR #333 client facade:
        # the manifest-backstop failure is strictly more load-bearing than a
        # crashed tail leg, so a tail exception raised AFTER a finalize failure
        # is logged and chained onto it, never allowed to replace it. With
        # finalize clean the tail error propagates exactly as it always has.
        if finalize_error is None:
            raise
        logger.error(... exc_info=True)
        raise finalize_error from tail_error
```

Three properties, one test each in `TestFinalizeGuard`:

- `test_finalize_error_wins_over_a_later_tail_crash` — crashes `_lambda_result_rows` (your first example) after a failed finalize; asserts `excinfo.value is boom` (the finalize error), `excinfo.value.__cause__ is tail_boom` (your `__context__` suggestion, as an explicit `from`), `"tail blew up" in caplog.text`, and that the tail did get as far as `coverage` before dying.
- `test_tail_crash_alone_still_propagates` — with finalize clean a tail exception propagates unchanged (bare `raise`), so this adds no swallowing.
- The existing `test_both_attempts_fail_tail_runs_then_reraises` still pins the clean-tail case.

`git diff -w` is the useful view here: the real change is ~30 lines, the rest is the indent.

Two notes on what I deliberately did *not* change:

- The diagnostic cost you mention is now covered from the other side too: `finalize_error` reaches the run-stats parquet (the durability thread), so even a tail that crashes *after* the stats dispatch leaves the failure durably recorded — and when it crashes *before* it, the `logger.error` + the `__cause__` chain carry both errors.
- On the raster path the all-shards-failed `RuntimeError` still raises ahead of the deferred finalize error, and I left it that way with an explicit comment: that's the run's own root cause rather than a tail crash, and a stale manifest is moot when no shard wrote anything. Flagging it as a judgment call rather than an oversight.

src/zagg/runner.py

Fixed — raster is at parity, not scoped out. Commit `46963a5`.

`RasterStrategy`'s Lambda driver now routes through the same shared helper, and the stale comment asserting the old "a failure raises" semantics is rewritten:

```python
            finalize_error = _finalize_with_retry(
                lambda: _invoke_lambda_finalize(
                    client, function_name, store_path,
                    output_creds_event=output_creds_event, config_dict=config_dict,
                    dataset=dataset, parent_order=int(grid.parent_order), overwrite=overwrite,
                ),
                store_path=store_path,
                reraise_note="This error re-raises after the post-run tail completes.",
            )
```

Plus the rest of the parity you asked for: the failure rides the raster `_dispatch_run_stats` call (so it is durable before anything can raise), the summary gains the same always-present `finalize_error` key with the same meaning as `_run_lambda`'s, and the original exception re-raises after the tail so `agg` still exits nonzero.

New `TestRasterFinalizeGuard` in `tests/test_raster_runner.py`, written in the `_FakeLambdaClient` responder style of the surrounding hive-Lambda tests:
- `test_success_path_unchanged` — exactly one `finalize` event, `summary["finalize_error"] is None`, and `None` on the wire.
- `test_failure_defers_through_the_tail_then_reraises` — two `finalize` events (initial + one retry), `coverage` still fires, `stats["finalize_error"] == "RuntimeError: finalize down"`, and the `RuntimeError` re-raises.

One deliberate divergence, commented in place: the all-shards-failed `RuntimeError` still raises **before** the deferred finalize error. It is the run's own root cause rather than a tail crash, and a stale root manifest is moot when no shard wrote anything — and because the run-stats dispatch precedes it, the finalize failure is durable either way. Say the word if you'd rather finalize win there too.

src/zagg/runner.py

Both taken, one fixed and one answered. Commit `bf3658d`.

**The retryable warning is now diagnosis-only** — your exact objection: the stale-manifest statement is false by the time a healed run returns.

```python
            message = (
                f"finalize (manifest backstop) failed for {store_path}: {e}. "
                f"Retrying in {backoff_s} s."
                if retrying
                else f"finalize (manifest backstop) failed for {store_path}: {e}. "
                f"Shard outputs are written; ... It is idempotent — re-run "
                f"the dispatch (or re-dispatch finalize) to stamp the manifest. "
                f"{reraise_note}"
            )
```

The full contract lands only on the terminal failure, where it is true. `test_both_attempts_fail_tail_runs_then_reraises` now pins the split rather than looping the same assertions over both messages: `msgs[0].endswith("Retrying in 5 s.")` and `"Shard outputs are written" not in msgs[0]`, `msgs[1]` carries the remediation and ends with the caller's `reraise_note`. Both still name the store and the underlying error, so neither warning is cryptic. Your point that "same message contract as #333" only binds the terminal message is right — and it's now enforced structurally, since the terminal message is the *only* one either caller can emit.

**`stacklevel`: leaving it at 2, with the reasoning written down** rather than changed. I checked whether it can be made to point at a user frame and it can't, for either caller: from `_run_lambda` the user's frame is 3+ deep and depth varies by strategy, and the client facade's tail runs on the **finisher thread**, whose stack contains no user frame at all. Any fixed level is therefore wrong somewhere. So the comment now says so explicitly and leans on the message naming the store + the remedy instead of on attribution:

```python
            # stacklevel=2 attributes this to the dispatch call site inside zagg,
            # not the user's frame — deliberate: no fixed level can reach user
            # code from every caller (``agg`` is several frames up, and the
            # client facade's tail runs on a finisher THREAD whose stack has no
            # user frame at all), so the message names the store and the remedy
            # instead of relying on attribution.
```

src/zagg/runner.py
espg: 🤖 *from Claude*

Fixed, and validated rather than clamped. Commit `c771c1c`.

```python
    # Validated, not clamped (review finding): a negative ``retries`` would make
    # ``range(0)`` empty — finalize NEVER called and None returned, which every
    # call site reads as success. The helper is deliberately importable for
    # other paths to adopt, so a computed value must fail loudly. 0 is legal:
    # one attempt, no sleep.
    if retries < 0:
        raise ValueError(f"retries must be >= 0, got {retries}")
```

`max(retries, 0)` would have silently *skipped the retry* a caller asked for; the failure mode you found (never calling finalize and reporting success) is severe enough that a computed bad value should surface at the call, not degrade quietly. `retries=0` stays legal.

`test_retries_validated_not_silently_open` pins both halves: `-1` raises `ValueError` with `finalize.call_count == 0`, and `0` makes exactly one attempt with no backoff and still returns the error.

Agreed on the "returns the LAST exception" contract — no change there, and now that each attempt's warning differs by kind (see the retry-message thread) a first-then-second divergence is even easier to read in the log.

Issue #328, opened 2026-07-29
Narrative notebooks: query / write / read, each stage-timed


espg:
🤖 *from Claude*

Refs #265 ([ratified shape](https://github.com/englacial/zagg/issues/265#issuecomment-5123966118)); depends on #326 (write notebook consumes the client API) and on the tensor/mask emission work tracked on #265 (read notebook).

Three narrative notebooks, each timing its stage so end-to-end pipeline latency is tracked:

1. **Query** — CMR → catalog → shardmap construction (timed: CMR query, catalog build, shard assignment).
2. **Write** — Lambda dispatch for S3 staging via `zagg.client` (#326): futures + tqdm progress, per-shard timings harvested (timed: dispatch, fleet wall, per-phase worker splits).
3. **Read** — client-side pull and casting to HHDC tensors: `(tensor, mask, offset/gain, morton_id)` blocks (timed separately: fetch, decode, vertical rasterize — the stage split is what identifies whether S3 or numpy is the wall).

Conventions (§4): binder-runnable; prototype against `sliderule-public` (the read notebook can target the exact leaf verified on the #265 thread, doubling as a standing regression check on the lossless fraction); binder runs may later move to source.coop. Involves the substantial rework of the out-of-date #169.

espg:
🤖 *from Claude*

Dependency status first: **two of three notebooks are fully unblocked** — #326 merged (PR #333: `zagg.client`, futures, tqdm) covers the write notebook, and the query notebook never had a blocker. The read notebook's *contract* is settled (`(tensor, mask, (offset, gain), morton_id)`, implemented on PR #339) — only its execution waits on that merge. So planning and phases 1–2 can start now.

## Phases (branch `claude/328-notebooks`, one notebook per phase)

1. **Query notebook** — `notebooks/01_query_shardmap.ipynb`: CMR search → catalog → shardmap for the NEON SERC AOI, stage-timed (CMR query, catalog build, shard assignment). CMR search is anonymous, mortie is on PyPI — genuinely binder-runnable end to end.
[...]

PR #339, opened 2026-07-30
Spatially faithful tensor emission: rank_to_xy deinterleave, block assembly, mask channel

### LINKED ISSUE #336: Spatially faithful tensor emission: rank_to_xy deinterleave, block assembly, mask channel
🤖 *from Claude*

Refs #265 (parent — the "nested→2-D deinterleave / block / mask" follow-up bullet in the [distillation](https://github.com/englacial/zagg/issues/265#issuecomment-5123695622)); unblocked by espg/mortie#149 → **mortie 0.9.3** (`rank_to_xy`/`xy_to_rank`, normative spec §8). Downstream consumer: the #328 read notebook (which stays in #328).

## Problem

`readers/tdigest_tensor.py` maps the 1-D nested cell index to 2-D with a **row-major reshape** (`divmod(cell_id, side)`). Nested order within a chunk is a Z-order curve, so emitted `(side, side, n_bins)` tensors are spatially scrambled beyond 2-cell runs — an ML pipeline consuming them learns noise. This is the load-bearing gap for HHDC emission (#265).

## Scope

1. **mortie floor → `>=0.9.3`** in pyproject (single edit reaches the Lambda layer post-#323).
2. **Spatial faithfulness**: replace the row-major reshape with `mortie.rank_to_xy` at every 1-D→2-D seam in `readers/tdigest_tensor.py`; tensors and their morton indexing become healpy-`pix2xyf`-convention 2-D blocks (matching gridlook's rendering convention).
3. **128×128 block assembly**: emit at a requested block order (an o12 child = 4 × 64×64 inner chunks on the production geometry), assembling inner-chunk tensors into the caller's block size.
4. **Mask channel**: decode the hive leaf's `coverage.moc` occupancy sidecar into the same deinterleaved 2-D layout — completing the ratified reader contract **`(tensor, mask, offset/gain, morton_id)`**. 2-state today (observed-with-data / not), upgrading to 3-state automatically once the #334 strata land (noise-occupied ⇒ observed-no-signal).
5. **Golden spatial-fidelity tests** pinned to mortie spec §8's convention (adjacent-on-sphere ⇒ adjacent-in-tensor), plus S2 co-registration sanity (identical o19 axis ⇒ same deinterleave aligns raster and tensor).

Out of scope: percentile-surface emission, waveform/`PhotonFlux` objects, the moczarr migration (all tracked on #265), and the read-side vectorization of the vertical binning (separate #265 bullet).

espg:
Closes #336. Refs #265 (the "nested→2-D deinterleave / block / mask" follow-up bullet in the [distillation](https://github.com/englacial/zagg/issues/265#issuecomment-5123695622)).

Implements the [plan comment](https://github.com/englacial/zagg/issues/336#issuecomment-5124645379) exactly: the tensor readers stop row-major-reshaping the nested cells axis (a Z-order curve — the old `divmod(cell_id, side)` scrambled every emitted block spatially beyond 2-cell runs) and instead place each cell at the **bit deinterleave of its chunk-local nested rank** via mortie 0.9.3's `rank_to_xy`/`xy_to_rank` (espg/mortie#150; normative contract in mortie `docs/specification.md` §8, frozen for mortie 1.x).

## What changed

- **`src/zagg/readers/_layout.py`** (new): the `rank_to_rowcol` / `rowcol_to_rank` helper pair, pinning the orientation once — **row = y, col = x**, matching gridlook's `bit_combine(j, i)` texture convention (`gridlook/src/ui/grids/Healpix.vue`, `getUnshuffleIndex`: `texture[i*size + j] = data[bit_combine(j, i)]`), so an emitted tensor uploads to a gridlook texture with no further shuffle. The module docstring documents the choice against both mortie spec §8 and the gridlook file.
- **`src/zagg/readers/tdigest_tensor.py`**: every 1-D→2-D seam routes through the helper — `read_tensors` cell placement, and `read_raw_values` / `read_locations` now report the same deinterleaved `(row, col)` instead of a raw 1-D `cell_id` (one convention across the readers; invert with `rowcol_to_rank` for `read_cell`).
- **Block assembly**: `read_tensors(block_order=...)` assembles the `4**(chunk_order − block_order)` inner chunks of one block-order subtree into a single `(2**d, 2**d, n_bins)` tensor (o12 on production geometry → 128×128 from 4 inner chunks). The z-window and `fit` policy are reconciled **block-wide** (one shared offset/gain per block). `_chunk_word` now also validates that a block's written cells share the block-order ancestor, so a dense multi-shard axis cannot silently mis-assemble.
- **Mask channel**: `read_tensors` now yields the full ratified contract **`(tensor, mask, (offset, gain), morton_id)`**. The uint8 mask (same deinterleaved layout) is `0` unobserved / `1` observed-no-digest / `2` observed-with-data; states 1–2 decode the hive leaf's `coverage.moc` occupancy sidecar via the existing frozen `hive.decode_coverage_bitmap` (reused, not reimplemented) — one small sidecar object, no digest bytes. Stores without exact occupancy (every flat store; unstamped/box-only leaves) degrade to the 2-state `{0, 2}` mask, per D9. Today occupancy equals digest coverage so `1` does not occur; once the #334 strata land, noise-occupied cells appear as `1` automatically (data-driven — pinned by `test_noise_occupied_reads_as_observed_no_signal`).
- **`pyproject.toml`**: `mortie>=0.9.1` → `>=0.9.3` (single edit reaches the Lambda layer post-#323).
- **Docs**: new "Spatially faithful tensors (deinterleave, blocks, mask)" section in `docs/ragged_layout.md` cross-linking mortie spec §8 as the normative convention; `docs/quickstart.md` blurb; `notebooks/tdigest_reader_example.ipynb` updated to the new contract (all code cells re-verified to run).

## Phases

- [x] Phase 1 — mortie floor `>=0.9.3` + `readers/_layout.py` helper pair; all `divmod` seams routed through it (8ee6e71)
- [x] Phase 2 — fidelity goldens: spec-§8 vectors copied verbatim from mortie's merged `tests/test_rank_xy.py` (healpy 1.20.0 provenance, depths 6/8), synthetic 4-cell adjacency, rank↔(row,col) round-trips (26cebcb)
- [x] Phase 3 — `block_order=` assembly with block-wide window/fit (7932d4a)
- [x] Phase 4 — `coverage.moc` mask channel; contract `(tensor, mask, (offset, gain), morton_id)` (03430af)
- [x] Phase 5 — docs + notebook (16e4027)

## How it was tested

- New `tests/test_reader_layout.py` (30 tests): golden placement at depths 6 and 8 (vectors imported from mortie's spec-pinned suite, never re-derived), round-trips, store-driven placement through the production template + writers, block assembly (bit-identical tiles vs chunk-at-a-time reads, order-7/6 grouping + morton coarsening, block-wide window + block-wide `fit="raise"`, out-of-range `block_order`), and the mask channel on committed synthetic hive leaves (digest alignment, noise-occupied → state 1, `encoding: "full"`, unstamped degrade, corrupt-sidecar raise, cell-order mismatch raise).
- **Adjacency acceptance pin verified both ways**: `test_sphere_adjacent_cells_land_tensor_adjacent` and `test_golden_ranks_land_at_golden_xy` were run against a temporary local revert to the old `divmod` reshape — both fail there and pass on the new path (the revert was not committed).
- `tests/test_readers.py` updated to the new contract (placement expectations, `(row, col)` reporting, 4-tuple unpacking; sharded/flat parity now also compares masks and scales).
- Full suite locally: **2679 passed, 35 skipped, 1 failed** — the one failure is the known pre-existing `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, unrelated to this change (the two `TestConsolidationGate` failures previously seen on main did not reproduce). `ruff check` / `ruff format --check` clean on everything this PR touches.
- mortie **0.9.3 resolved from PyPI** (`uv lock --upgrade-package mortie`) — no local-checkout fallback was needed; the goldens run against the released wheel.

## Questions for review

- **Breaking reader API**: `read_tensors` now yields a 4-tuple and `read_raw_values`/`read_locations` report `(row, col)` instead of a 1-D `cell_id`. These are in-repo client read helpers (issue #79) with no other in-tree consumers beyond the example notebook (updated here); the #328 read notebook consumes the new contract downstream. Flagging in case any out-of-tree consumer needs a heads-up.
- **Pre-existing local lint noise, not fixed here** (per the don't-fix-unrelated rule): `ruff check src tests` flags `N818` on `src/zagg/registry.py:64` (`UnknownCapability`), and `ruff format --check` wants to reformat a python snippet inside `tests/data/benchmark/README.md` — both present on current `origin/main` and outside the PR-bot's `E,F,W,I` gate.
- Blocks with occupancy but **zero stored digests** (possible post-#334: all-noise blocks) are skipped by the sweep (there is no z-window to emit); the mask for such blocks is therefore never yielded. Called out in the docstring — happy to add a `yield`-empty-blocks mode later if the #334 consumer needs it.

### DIFF
--- a/docs/ragged_layout.md
+++ b/docs/ragged_layout.md
@@ -213,7 +213,13 @@ for tensor, mask, (offset, gain), morton in read_tensors(store, field):
   frozen `hive.decode_coverage_bitmap` convention, one small sidecar object,
   no digest bytes). A store without exact occupancy (every flat store, or a
   box-only/`full`-less missing sidecar) degrades to the 2-state `{0, 2}`
-  populated/not mask. Today occupancy equals digest coverage, so state `1`
+  populated/not mask, where `0` means only "no stored digest" and asserts
+  nothing about whether the cell was observed. **The mask does not carry which
+  regime it is in** — a degraded mask and a 3-state mask over a block with no
+  observed-but-empty cell are both `{0, 2}` — so check
+  `has_exact_occupancy(store)` before keying on `mask == 1`; without it, an
+  empty noise stratum on a degraded store reads as a genuine absence. Today
+  occupancy equals digest coverage, so state `1`
   does not occur; once the
   [issue #334](https://github.com/englacial/zagg/issues/334) signal strata
   land, noise-occupied cells appear as `1` automatically — the 3-state
--- a/docs/ragged_layout.md
+++ b/docs/ragged_layout.md
@@ -157,13 +157,64 @@ cannot drift.
   not re-fetched per inner chunk). Each read chunk is a square `(side, side)`
   block of cells (`side = isqrt(cells_per_chunk)`, 64 for the production
   `chunk_inner` configs), and its coverage-cell morton id is derived from the
-  sibling `morton` coordinate coarsened to the chunk order.
+  sibling `morton` coordinate coarsened to the chunk order. Cell placement
+  within the block is the spec-pinned bit deinterleave — see
+  [the tensor section below](#spatially-faithful-tensors-deinterleave-blocks-mask).
 - **Single-cell random access** (`read_cell`) — indexes the vlen array directly.
   On a sharded store that is exactly **2 ranged GETs** (the shard-index suffix,
   then the one inner chunk holding the cell), never the whole shard object. An
   out-of-range index raises `IndexError` naming the valid range (no negative-index
   wrap); an absent cell returns the zero-length `(0, *inner_shape)` array. Works
   on the `{field}_locations` sibling too.
+
+## Spatially faithful tensors (deinterleave, blocks, mask)
+
+The nested cells axis traces a **Z-order curve** within each chunk subtree, so
+a row-major reshape of the 1-D axis scrambles the 2-D block spatially beyond
+2-cell runs. The readers instead place each cell at the **bit deinterleave**
+of its chunk-local nested rank
+([issue #336](https://github.com/englacial/zagg/issues/336)):
+`mortie.rank_to_xy` / `xy_to_rank` (mortie ≥ 0.9.3), whose contract is
+normative in mortie's `docs/specification.md` **§8** and frozen for mortie 1.x
+— `x` gathers the rank's even bits, `y` its odd bits, origin `(0, 0)` at the
+subtree's south corner, equal to healpy's face-local `pix2xyf` (nest)
+convention. zagg pins **row = y, col = x** (`readers/_layout.py`), matching
+gridlook's `bit_combine(j, i)` texture convention
+(`gridlook/src/ui/grids/Healpix.vue`), so an emitted tensor uploads to a
+gridlook texture with no further shuffle. The golden placement vectors in
+`tests/test_reader_layout.py` are imported from mortie's spec-pinned test
+suite, never re-derived.
+
+`read_tensors` yields the full reader contract per block:
+
+```python
+for tensor, mask, (offset, gain), morton in read_tensors(store, field):
+    ...  # tensor: (side, side, n_bins); mask: (side, side) uint8
+```
+
+- **Blocks** — by default one block per read chunk; `block_order=` assembles
+  the `4**(chunk_order - block_order)` chunks of one block-order subtree into
+  a single `(2**d, 2**d, n_bins)` tensor (`d = cell_order - block_order`; an
+  order-12 block on production geometry is 128×128 from 4 inner chunks). The
+  z-window and `fit` policy are reconciled **block-wide**: one shared
+  `(offset, gain) = (z_lo, resolution)` per block, so bin `i` covers
+  `[offset + i*gain, offset + (i+1)*gain)` for every cell in the block.
+- **Mask** — the block's occupancy channel on the same deinterleaved layout:
+  `0` unobserved, `1` observed with no stored digest, `2` observed with data.
+  States 1/2 come from the hive leaf's `coverage.moc` occupancy sidecar
+  ([issue #200](https://github.com/englacial/zagg/issues/200); decoded by the
+  frozen `hive.decode_coverage_bitmap` convention, one small sidecar object,
+  no digest bytes). A store without exact occupancy (every flat store, or a
+  box-only/`full`-less missing sidecar) degrades to the 2-state `{0, 2}`
+  populated/not mask. Today occupancy equals digest coverage, so state `1`
+  does not occur; once the
+  [issue #334](https://github.com/englacial/zagg/issues/334) signal strata
+  land, noise-occupied cells appear as `1` automatically — the 3-state
+  upgrade is data-driven, not a reader flag.
+
+`read_raw_values` / `read_locations` report the same deinterleaved
+`(row, col)` per cell (invert with `readers._layout.rowcol_to_rank` to get a
+cells-axis index for `read_cell`).
 
 ## Issue #210 typed-dtype migration
 
--- a/docs/ragged_layout.md
+++ b/docs/ragged_layout.md
@@ -199,6 +199,13 @@ for tensor, mask, (offset, gain), morton in read_tensors(store, field):
   z-window and `fit` policy are reconciled **block-wide**: one shared
   `(offset, gain) = (z_lo, resolution)` per block, so bin `i` covers
   `[offset + i*gain, offset + (i+1)*gain)` for every cell in the block.
+  Coarser than the stored shard assembles too (the block-local index is still
+  the nested rank), but the memory bound is then per *block*: the block's
+  decoded digests are all held for the block-wide window, and the emitted
+  tensor grows 4× per coarser order (an order-6 block on the production
+  order-19 geometry would be 34 TB). `max_block_bytes=` (2 GiB default)
+  refuses that allocation with a pointed error instead of a bare
+  `MemoryError`.
 - **Mask** — the block's occupancy channel on the same deinterleaved layout:
   `0` unobserved, `1` observed with no stored digest, `2` observed with data.
   States 1/2 come from the hive leaf's `coverage.moc` occupancy sidecar
--- a/docs/ragged_layout.md
+++ b/docs/ragged_layout.md
@@ -213,8 +213,22 @@ for tensor, mask, (offset, gain), morton in read_tensors(store, field):
   upgrade is data-driven, not a reader flag.
 
 `read_raw_values` / `read_locations` report the same deinterleaved
-`(row, col)` per cell (invert with `readers._layout.rowcol_to_rank` to get a
-cells-axis index for `read_cell`).
+`(row, col)` per cell. `readers._layout.rowcol_to_rank` inverts that to the
+**chunk-local** nested rank (`0..4**depth - 1`) — *not* a `read_cell` key,
+which is a **global** cells-axis index; the chunk's start offset is the
+missing term, and a bare rank is always in range so `read_cell` would read
+the wrong cell without raising. `cell_index` composes the two:
+
+```python
+for morton, (row, col), values in read_raw_values(store, field):
+    cell = cell_index(store, field, morton, row, col)  # chunk_start + rank
+    assert (read_cell(store, field, cell)[:, 0] == values).all()
+```
+
+It resolves the offset from the sibling `morton` coordinate, searching only
+the array's stored spans (the populated chunks the sweep readers yield from)
+and no digest bytes. `morton_index` must be a read-chunk id — a coarser
+`block_order` block id names no single chunk and raises.
 
 ## Issue #210 typed-dtype migration
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,11 @@ dependencies = [
     # register from it). 0.2.0 is not on PyPI yet — resolution fails until
     # upstream publishes it.
     "h5coro-hidefix>=0.3.1",
-    # 0.9.0 ships the decimal-emit surface shard ids route through (issue #199):
+    # Floor 0.9.3 for the reader deinterleave (issue #336): rank_to_xy /
+    # xy_to_rank, the nested-rank ↔ face-local (x, y) pair the tensor readers
+    # place cells by (espg/mortie#150; normative convention in mortie's
+    # docs/specification.md §8, frozen for mortie 1.x). Earlier floors: 0.9.0
+    # ships the decimal-emit surface shard ids route through (issue #199) —
     # decimal_repr / to_decimal, the _decimal_to_word parse-back, hive_path, and
     # the MortonIndexScalar decimal display.
     "mortie>=0.9.3",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dependencies = [
     # 0.9.0 ships the decimal-emit surface shard ids route through (issue #199):
     # decimal_repr / to_decimal, the _decimal_to_word parse-back, hive_path, and
     # the MortonIndexScalar decimal display.
-    "mortie>=0.9.1",
+    "mortie>=0.9.3",
     "earthaccess",
     "boto3",
     "fastparquet",
--- a/src/zagg/readers/tdigest_tensor.py
+++ b/src/zagg/readers/tdigest_tensor.py
@@ -536,7 +536,7 @@ def read_tensors(
     the block's coverage-cell morton id (``side`` is 64 for the production
     ``chunk_inner`` configs at the default per-chunk block). Cell placement is
     the spatially faithful bit deinterleave of the cell's nested rank
-    (issue #336): ``tensor[row, col] = rank_to_rowcol(rank, depth)`` per the
+    (issue #336): ``row, col = rank_to_rowcol(rank, depth)`` per the
     :mod:`zagg.readers._layout` orientation contract (mortie spec §8; row 0 /
     col 0 at the block subtree's south corner, gridlook texture convention).
 
--- a/src/zagg/readers/tdigest_tensor.py
+++ b/src/zagg/readers/tdigest_tensor.py
@@ -62,6 +62,7 @@ from __future__ import annotations
 
 import math
 from collections.abc import Iterator
+from itertools import chain, groupby
 from typing import Literal
 
 import numpy as np
@@ -350,17 +351,12 @@ def _open_morton(store: Store, field: str, zarr_format):
         ) from e
 
 
-def _chunk_word(words: np.ndarray, field: str, start: int) -> int:
-    """A read chunk's coverage-cell morton id from its cells' morton words.
+def _cells_order(words: np.ndarray, field: str, start: int) -> int:
+    """HEALPix order of a chunk's written cell words (the cells-axis order).
 
-    ``words`` are the chunk's per-cell morton coordinates (packed uint64 area
-    words at the cell order; ``0`` is the unwritten fill). The chunk id is any
-    written cell's word coarsened to the chunk order — ``cell_order -
-    log4(cells_per_chunk)`` — the same parent cell the CSR layout named its
-    subgroups by.
+    ``words`` are per-cell morton coordinates (packed uint64 area words at the
+    cell order; ``0`` is the unwritten fill).
     """
-    from mortie import clip2order
-
     from zagg.grids.morton import morton_decimal
 
     written = words[words != 0]
@@ -370,9 +366,33 @@ def _chunk_word(words: np.ndarray, field: str, start: int) -> int:
             f"'morton' coordinate — the dense coordinate write did not cover it"
         )
     decimal = morton_decimal(int(written[0]))
-    cell_order = len(decimal) - (2 if decimal.startswith("-") else 1)
-    depth = (int(len(words)).bit_length() - 1) // 2  # log4(cells_per_chunk)
-    return int(clip2order(cell_order - depth, written[:1])[0])
+    return len(decimal) - (2 if decimal.startswith("-") else 1)
+
+
+def _chunk_word(words: np.ndarray, field: str, start: int) -> int:
+    """A read chunk's/block's coverage-cell morton id from its cells' words.
+
+    The id is the written cells' common ancestor at ``cell_order -
+    log4(len(words))`` — the same parent cell the CSR layout named its
+    subgroups by. Raises when the written cells do NOT share that ancestor:
+    the cells axis is then not nested-aligned over this span (a dense
+    multi-shard axis assembled at a block order coarser than the shard
+    order), and index arithmetic would place cells in the wrong subtree.
+    """
+    from mortie import clip2order
+
+    written = words[words != 0]
+    cell_order = _cells_order(words, field, start)
+    order = cell_order - (int(len(words)).bit_length() - 1) // 2  # log4(len)
+    ancestors = clip2order(order, written)
+    if np.any(ancestors != ancestors[0]):
+        raise ValueError(
+            f"cells of the block at {start} of {field!r} span more than one "
+            f"order-{order} ancestor — the cells axis is not nested-aligned over "
+            f"this span (a dense multi-shard axis assembles only at block orders "
+            f"at or finer than the shard order)"
+        )
+    return int(ancestors[0])
 
 
 def _tensor_side(arr, field: str) -> tuple[int, int]:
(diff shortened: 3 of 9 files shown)


docs/ragged_layout.md
```diff
@@ -157,14 +157,65 @@ cannot drift.
...
+  box-only/`full`-less missing sidecar) degrades to the 2-state `{0, 2}`
+  populated/not mask. Today occupancy equals digest coverage, so state `1`
+  does not occur; once the
+  [issue #334](https://github.com/englacial/zagg/issues/334) signal strata
+  land, noise-occupied cells appear as `1` automatically — the 3-state
+  upgrade is data-driven, not a reader flag.
+
+`read_raw_values` / `read_locations` report the same deinterleaved
+`(row, col)` per cell (invert with `readers._layout.rowcol_to_rank` to get a
+cells-axis index for `read_cell`).
```
espg: 🤖 *from Claude (review)*

**[medium] `rowcol_to_rank` does not produce a `read_cell` index.** This line tells the reader to "invert with `readers._layout.rowcol_to_rank` to get a cells-axis index for `read_cell`", but the two indices are different things:

- `rowcol_to_rank(row, col, depth)` returns the **chunk-local** nested rank `0..4**depth - 1` (its own docstring: "Chunk-local nested rank at a `(row, col)` tensor position").
- `read_cell(store, field, cell)` documents `cell` as "the cell's **global** position on the array's cells axis" and range-checks against `arr.shape[0]`.

So the documented recipe silently returns the wrong cell for every chunk except the one at cells-axis offset 0 — and it will not raise, because a chunk-local rank is always in range. Worse, the composition is not currently possible from the yielded tuple at all: `read_raw_values`/`read_locations` yield `(morton_index, (row, col), values)` and never expose the chunk/block start, so there is no way to recover `chunk_start + rowcol_to_rank(...)`.

Either drop the `read_cell` clause here (and in `read_raw_values`'s docstring, `src/zagg/readers/tdigest_tensor.py:703-704`), or make the composition possible — e.g. yield the block start alongside, or document deriving it from `morton_index` via mortie.

src/zagg/readers/tdigest_tensor.py
```diff
@@ -366,21 +370,136 @@ def _chunk_word(words: np.ndarray, field: str, start: int) -> int:
...
+    """
+    from mortie import clip2order
+
+    written = words[words != 0]
+    cell_order = _cells_order(words, field, start)
+    order = cell_order - (int(len(words)).bit_length() - 1) // 2  # log4(len)
+    ancestors = clip2order(order, written)
+    if np.any(ancestors != ancestors[0]):
+        raise ValueError(
+            f"cells of the block at {start} of {field!r} span more than one "
```
espg, line 406: 🤖 *from Claude (review)*

**[medium] this guard's message asserts a constraint the code does not have, and the branch looks unreachable.** The message says "a dense multi-shard axis assembles only at block orders at or finer than the shard order". I ran the K=16 sharded fixture (`_sharded_store`, order-6 shard, order-8 chunks, whole-sphere order-12 cells axis) at block orders both finer and **coarser** than the shard order:

```
bo 8 -> 2 blocks [(16, 16, 128), (16, 16, 128)]
bo 7 -> 2 blocks [(32, 32, 128), (32, 32, 128)]
bo 6 -> 1 blocks [(64, 64, 128)]
bo 5 -> 1 blocks [(128, 128, 128)]   # coarser than the shard order
bo 4 -> 1 blocks [(256, 256, 128)]   # coarser than the shard order
```

Coarser-than-shard blocks assemble fine and get correct ids, because on a nested-ordered cells axis the block-local index *is* the nested rank by construction (I verified the composition identity numerically for several depth pairs: `rank_to_xy(R*4**d2 + r, d1+d2) == (X*2**d2 + x, Y*2**d2 + y)`). The actual guard against a non-assemblable request is the `arr.shape[0] % block_cells` check above, not this one.

Which leaves the `np.any(ancestors != ancestors[0])` branch unreachable on any well-formed store (nothing in the suite covers it — `grep "span more than one" tests/` is empty). Per §4 (no dead code, every behavioral change tested): either drop it, or pin it with a test that writes a deliberately misordered `{group}/morton` array — and either way fix the message, which currently misdescribes the supported envelope.

src/zagg/readers/tdigest_tensor.py
```diff
@@ -454,37 +609,75 @@ def read_tensors(
...
+        block_cells = 4**block_depth
+        if int(arr.shape[0]) % block_cells:
+            raise ValueError(
+                f"{field!r} has {int(arr.shape[0])} cells — not a whole number of "
+                f"order-{block_order} blocks ({block_cells} cells each), so the cells "
+                f"axis cannot assemble at this block order"
+            )
+    block_side = 1 << block_depth
+
+    for block, group in groupby(chain([first], chunks), key=lambda c: c[0] // block_cells):
```
espg, line 720: 🤖 *from Claude (review)*

**[medium] `block_order` breaks the module's documented memory envelope, and nothing bounds it from below.** Two compounding effects, neither reflected in the docs:

1. `cells` accumulates the **decoded digests of the whole block** before `chunk_z_range` can run (the block-wide window needs all of them). `_iter_populated_chunks`'s docstring pins the accepted bound at "one span's decoded payload — ~141 MB at the o8 t-digest scale". A block coarser than the stored-object span covers `4**(span_order - block_order)` spans, so the held cost is that bound times the number of spans, not one span. As shown in the sibling comment, coarser-than-shard block orders *do* assemble, so this is reachable, not hypothetical.
2. The output allocation is `4**block_depth * n_bins` elements. The only validation is `0 <= block_order <= chunk_order`, so on the production geometry (order-19 cells) `block_order=6` asks for `np.zeros((8192, 8192, 128), uint32)` ≈ 34 TB and dies with a bare `MemoryError` — no pointed error naming `block_order` as the cause, unlike every other bad-input path in this module.

Suggest (a) restating the memory bound in the `read_tensors` / `_iter_populated_chunks` docstrings in terms of a *block* rather than a span, and (b) a pointed guard on the emitted block size (or at least on `block_depth`) so the footgun fails with a message instead of an allocator.

src/zagg/readers/tdigest_tensor.py
```diff
@@ -31,10 +31,13 @@
...
-``cell_id`` within the chunk, ``side = isqrt(cells_per_chunk)`` — 64 for the
-production ``chunk_inner`` configs); its coverage-cell morton id is derived
-from the sibling ``morton`` coordinate. Random access to one cell
+Each read chunk is a square ``(side, side)`` block of cells
+(``side = isqrt(cells_per_chunk)`` — 64 for the production ``chunk_inner``
+configs); its coverage-cell morton id is derived from the sibling ``morton``
+coordinate. A cell's 2-D position is the **bit deinterleave** of its nested
+rank within the chunk (``readers._layout.rank_to_rowcol`` — mortie spec §8 /
+gridlook orientation; issue #336), never a row-major reshape: nested order is
+a Z-order curve, so ``divmod(rank, side)`` would scramble the block spatially. Random access to one cell
```

**[low] a stale "row-major" comment survives inside the file whose whole point is retiring that framing.** `_iter_populated_chunks` still closes with (`src/zagg/readers/tdigest_tensor.py:332`, unchanged by this PR so not inline-commentable):

> `cell_pos` is the cell's **row-major position** within the chunk — the same index the writer placed it at.

That value is the cells-axis **nested rank** that every new seam feeds to `rank_to_rowcol`; describing it as a row-major position is exactly the confusion this PR exists to remove, and it is the docstring a reader hits first when tracing `pos` → `rank`. (Same class as the stale-comment fold on issue #300, d104dbb.)

**[low] also on this line:** it is 104 chars and merges two sentences ("…scramble the block spatially. Random access to one cell"). `E501` is ignored so ruff will not flag it, but it is the only >100-char line in the module and it breaks the paragraph's wrap; `git diff` shows it as one long added line where the rest of the hunk wraps at ~78.

pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     # 0.9.0 ships the decimal-emit surface shard ids route through (issue #199):
     # decimal_repr / to_decimal, the _decimal_to_word parse-back, hive_path, and
     # the MortonIndexScalar decimal display.
-    "mortie>=0.9.1",
+    "mortie>=0.9.3",
```
espg, line 45: 🤖 *from Claude (review)*

**[low] the floor moved but the comment explaining it did not.** The three lines directly above still read "0.9.0 ships the decimal-emit surface shard ids route through (issue #199): decimal_repr / to_decimal, the _decimal_to_word parse-back, hive_path…" — nothing records that the floor is now 0.9.3 for `rank_to_xy`/`xy_to_rank` (espg/mortie#150, spec §8), which is the *only* reason this PR touches pyproject at all.

That matters more here than in a normal dep block: `deployment/aws/build_layer.sh:94` derives the layer's mortie spec by regexing this exact line out of `pyproject.toml`, so this comment is the single place the floor's rationale lives for anyone auditing the Lambda layer.

(The PR body's claim that the single edit reaches the layer post-#323 checks out — `build_layer.sh` reads the spec from pyproject and hard-fails if no `mortie-*` lands in the layer. And `uv.lock` is gitignored (`.gitignore:22`), so there is correctly nothing to commit there; I confirmed a fresh `uv sync` in a clean worktree resolves mortie 0.9.3 from PyPI and the new goldens pass against it.)

src/zagg/readers/tdigest_tensor.py
```diff
@@ -398,17 +517,34 @@ def read_tensors(
...
+    time, visiting only the STORED objects. For each populated block it trims
     the per-cell tails, derives a fixed z-window (see :func:`chunk_z_range`),
     rasterizes every populated cell's digest into ``n_bins`` counts (see
     :func:`rasterize_cell`), and emits the ``(side, side, n_bins)`` tensor with
-    the chunk's coverage-cell morton id (``side`` is 64 for the production
-    ``chunk_inner`` configs).
+    the block's coverage-cell morton id (``side`` is 64 for the production
+    ``chunk_inner`` configs at the default per-chunk block). Cell placement is
+    the spatially faithful bit deinterleave of the cell's nested rank
+    (issue #336): ``tensor[row, col] = rank_to_rowcol(rank, depth)`` per the
```

**[low] the snippet in this docstring is not valid as written.** `` tensor[row, col] = rank_to_rowcol(rank, depth) `` reads as assigning a `(row, col)` pair *into* the tensor. The code does the opposite:

```python
row, col = rank_to_rowcol(rank, block_depth)
tensor[row, col, :] = counts.astype(out_dtype)
```

Suggest `` `row, col = rank_to_rowcol(rank, depth)` `` (the notebook already phrases it that way: "placing cell rank `r` at `rank_to_rowcol(r, depth)`").

src/zagg/readers/tdigest_tensor.py
```diff
@@ -366,21 +370,136 @@ def _chunk_word(words: np.ndarray, field: str, start: int) -> int:
...
+    return "bitmap", hive.decode_coverage_bitmap(payload, shard, cell_order)
+
+
+def _block_mask(words: np.ndarray, occupancy: tuple | None, block_depth: int) -> np.ndarray:
+    """The block's ``(side, side)`` uint8 occupancy base (values 0/1).
+
+    ``1`` marks a cell the leaf's occupancy sidecar records as observed;
+    the caller upgrades digest-bearing cells to ``2``. Without occupancy
+    truth the base stays ``0`` everywhere (mask degrades to the 2-state
+    populated/not — never a wrong answer).
```

**[low] "never a wrong answer" will stop being true the moment #334 lands, and the consumer has no way to tell.** The 3-state semantics assign `0` the meaning **unobserved** (`read_tensors` docstring: "`0` = unobserved"). On the D9 degrade the base stays `0` everywhere, so a flat/unstamped store post-#334 reports a noise-occupied cell as `0` — i.e. *asserts* "not observed" about a cell that was observed. That is a false negative on occupancy, the one thing the sibling `decode_coverage_bitmap` posture explicitly refuses ("refusing to zero-pad or truncate — a partial cell set would be a false negative"). Calling it "the 2-state populated/not" is fine; calling it "never a wrong answer" overstates it.

The underlying gap is that the yielded 4-tuple carries **no discriminator** for which semantics apply. A consumer keying on `mask == 1` (the documented "observed, no signal" stratum — precisely the #334 consumer this channel exists for) silently gets an empty set on a degraded store, indistinguishable from a genuine "no noise-occupied cells in this block". `test_unstamped_store_degrades_to_two_state` pins the degrade but nothing lets a caller *detect* it.

Options: surface occupancy availability (a 5th tuple element / a small dataclass), or reserve a documented sentinel for "unknown" so `0` keeps meaning "observed-negative" only when it is actually known. Either way, at minimum soften the "never a wrong answer" claim and state in the `read_tensors` Yields block that a caller cannot distinguish the two mask regimes from the yielded value.

tests/test_reader_layout.py
```diff
@@ -0,0 +1,374 @@
...
+
+def _put_object(store, key, payload):
+    """PUT raw bytes into a zarr store (the MemoryStore sidecar write)."""
+    from zarr.core.buffer import default_buffer_prototype
+    from zarr.core.sync import sync
+
+    sync(store.set(key, default_buffer_prototype().buffer.from_bytes(payload)))
+
+
+class TestMaskChannel:
```
espg, line 388: 🤖 *from Claude (review)*

**[low] no test crosses the mask channel with `block_order`.** `TestBlockAssembly` runs on `_sharded_store`, which is never stamped — so every block-assembly test sees `occupancy is None` and a mask that is `{0, 2}` by construction. `TestMaskChannel` only ever reads the default per-chunk block (`_read_one` unpacks a single tuple). The one place the two meet is `test_chunk_order_block_matches_default`, where `block_order == chunk_order` means no assembly happens.

So the interaction that phases 3 and 4 jointly introduce — `_block_mask(words, occupancy, block_depth)` slicing a leaf-wide bitmap against a *multi-chunk* `words` window, with the deinterleave at `block_depth` rather than `depth` — is unpinned.

I built the missing case out-of-tree (a stamped 16-chunk leaf: `chunk_inner=8`, order-6 shard, order-12 cells, digests at ranks 11/300/1000/4000, extra occupancy at 12/301/2500) and it is **correct** at every block order:

```
block_order 7 -> block 0: ones (2,2) (6,19)  twos (3,1) (6,18) (30,24)
block_order 6 -> ones (2,2) (6,19) (40,26)   twos (3,1) (6,18) (30,24) (60,48)
```

which matches the hand-deinterleave of ranks 12→(2,2), 301→(6,19), 2500→(40,26) and 11→(3,1), 300→(6,18), 1000→(30,24), 4000→(60,48). So this is a missing pin, not a bug — but it is the one seam where a future refactor could regress the mask silently, since `block_depth` and `depth` coincide in every test that currently touches occupancy.

docs/ragged_layout.md
espg: 🤖 *from Claude*

Fixed both halves — the doc recipe **and** the API gap it exposed (04b8374).

The global index is genuinely underivable from the yielded tuple: the chunk start is not in it, and a chunk word → cells-axis index mapping needs the axis layout a store-scoped reader never sees. So the fold adds the minimal accessor:

```python
for morton, (row, col), values in read_raw_values(store, field):
    cell = cell_index(store, field, morton, row, col)  # chunk_start + rank
    assert (read_cell(store, field, cell)[:, 0] == values).all()
```

`cell_index` resolves the offset from the sibling `morton` coordinate and searches only the array's STORED spans (the populated chunks the sweep readers yield from) — one small `morton` slice per span, no digest bytes. `morton_index` must be a read-chunk id; a coarser `block_order` block id names no single chunk and raises with that stated.

Doc/docstring: `docs/ragged_layout.md` now says `rowcol_to_rank` gives the **chunk-local** rank (`0..4**depth - 1`), *not* a `read_cell` key, spells out why the wrong-cell read is silent (a chunk-local rank is always in range), and shows the composition above; `read_raw_values`' Yields block says the same and points at `cell_index`.

Tests (`TestCellIndexComposition`, `tests/test_reader_layout.py`): round-trips every yielded `(row, col)` across two shards through `cell_index` → `read_cell` and compares to `read_raw_values`' exact means; pins that the bare chunk-local rank returns a silently **empty** digest instead of raising (the failure mode the old recipe had); pins the coarse-block-id and out-of-block-position raises.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Agreed on both counts — the restriction was false and the message misdescribed the envelope. Fixed in ef77ec3.

- **Message + docstring corrected.** `_chunk_word` now states that any span of a nested-ordered cells axis shares its ancestor by construction, *including a block coarser than the stored shard* (with the composition identity you verified, `rank_to_xy(R*4**d2 + r, d1+d2) == (X*2**d2 + x, Y*2**d2 + y)`), and that the axis-length check in `read_tensors` is what rejects an untileable block order. The raise now reads "…the 'morton' coordinate is not in nested order over this span, so a cells-axis index does not name a position in the block's subtree".
- **Kept the guard, and pinned it** rather than dropping it: `test_morton_not_in_nested_order_raises` grafts a cell from the other shard's subtree over a written `12/morton` word and asserts the raise, so it is no longer unreachable.
- **Pinned the behaviour you measured out-of-tree**: `test_block_coarser_than_the_shard_assembles` is parametrized on `block_order` 5 and 4 (128×128 and 256×256), asserting the emitted id is `clip2order(block_order, shard6)` and that the whole order-6 shard tensor lands *bit-identical* at `64 * rank_to_rowcol(shard_rank_in_parent, 6 - block_order)` with nothing outside that tile. Reproduced in-tree: bo 5 → shard at (0, 0); bo 4 → shard at (0, 128), rank 4 of 16 children.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Both parts done in 67aee17.

**(b) the guard.** `read_tensors` now takes `max_block_bytes: int = 2 * 1024**3` and computes the emitted tensor size *before* allocating:

```
block_order=4 emits a 256×256×128 uint32 block tensor = 33554432 bytes (0.03 GiB),
over the 1024-byte max_block_bytes limit; use a finer block_order, fewer n_bins,
or raise max_block_bytes deliberately
```

It is checked against the REQUESTED `n_bins` (the `fit` policy only ever shrinks it), so it is a true upper bound; the escape hatch is explicit. `test_block_tensor_over_max_block_bytes_raises` pins the message (block order, shape/dtype, byte count, limit) and that the same block passes under the default cap.

**(a) the docs, honestly.** `read_tensors` gains a "the memory bound is per BLOCK, not per stored object" paragraph naming both terms: (1) the block's decoded digests, held because the block-wide window needs all of them, = `4**(span_order - block_order)` × one span's payload (~141 MB at o8), and (2) the `4**block_depth * n_bins` tensor, 34 TB for an order-6 block on the production order-19 geometry. It states plainly that `max_block_bytes` bounds term (2) only — term (1) follows the block order asked for. `_iter_populated_chunks` now scopes its "~141 MB" to *this generator* and points at the block-level accumulation; `docs/ragged_layout.md`'s Blocks bullet carries the same, plus that coarser-than-shard does assemble (per the sibling thread).

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Both fixed in b550f2b.

- `_iter_populated_chunks`: `cell_pos` is now described as "the cell's chunk-local NESTED RANK — the same index the writer placed it at, and the index the readers deinterleave to a tensor position (`rank_to_rowcol`, issue #336); it is never a row-major position (nested order is a Z-order curve)."
- The 104-char module-docstring line is rewrapped at the paragraph's ~78 columns, splitting the two merged sentences ("…scramble the block spatially." / "Random access to one cell…"). `awk 'length > 100'` over the module is now empty.

pyproject.toml
espg: 🤖 *from Claude*

Recorded in cfedc26 — the comment now leads with the reason this PR touches `pyproject.toml` at all and keeps the older rationale below it:

```
# Floor 0.9.3 for the reader deinterleave (issue #336): rank_to_xy /
# xy_to_rank, the nested-rank ↔ face-local (x, y) pair the tensor readers
# place cells by (espg/mortie#150; normative convention in mortie's
# docs/specification.md §8, frozen for mortie 1.x). Earlier floors: 0.9.0
# ships the decimal-emit surface shard ids route through (issue #199) — ...
"mortie>=0.9.3",
```

The spec line itself is untouched, and `build_layer.sh`'s extractor parses `[project].dependencies` via `tomllib` (not raw lines), so the layer still derives `mortie>=0.9.3` — verified by running its exact snippet against the edited file.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Fixed in cc7b6fb — now `` `row, col = rank_to_rowcol(rank, depth)` ``, matching the code and the notebook's phrasing.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Fair — the claim overstated it, and the missing piece was a discriminator. Folded in 3dc299f with the smallest **API-stable** option: a per-store predicate, `has_exact_occupancy(store)`, plus honest wording. The ratified 4-tuple is unchanged.

- `True` — the leaf carries exact `coverage.moc` occupancy, so `0` really means *unobserved* and `1` is reported wherever it occurs.
- `False` — no exact occupancy (no stamp, i.e. every flat store; a box-only envelope; or a bitmap envelope whose sidecar is gone); the mask degrades to 2-state and `0` means only "no stored digest here".

Why this over the two alternatives you listed: a 5th tuple element (or dataclass) changes the issue #336 contract for every consumer to fix a per-*store* fact that never varies block to block; a reserved "unknown" sentinel changes the degraded mask's *values* (`{0, 2}` → `{sentinel, 2}`), which is the documented degrade and a wire-visible break for anyone keying on the current values. The predicate adds no per-block cost and is a pure addition.

It shares a new `_coverage_occupancy` helper with `_load_occupancy`, so the predicate cannot report a regime the mask build does not produce (pinned: a stamped leaf whose sidecar object is deleted reports `False` *and* reads back `{0, 2}`).

Wording: `_block_mask` drops "never a wrong answer" and says `0` "asserts NOTHING about whether the cell was observed"; `read_tensors`' Yields block states that **the yielded mask does not say which regime it is in** — a degraded mask and a 3-state mask over a block with no observed-but-empty cell are both `{0, 2}` — so a consumer keying on `mask == 1` must check `has_exact_occupancy` first or it reads an empty stratum as a genuine absence. `docs/ragged_layout.md`'s mask bullet carries the same. `test_has_exact_occupancy_discriminates_the_two_regimes` pins the identical `{0, 2}` value sets with opposite semantics.

tests/test_reader_layout.py
espg: 🤖 *from Claude*

Pinned in c01f163 — `TestMaskBlockCrossing` builds exactly the case you did out-of-tree (stamped 16-chunk leaf: `chunk_inner=8`, order-6 shard, order-12 cells, digests at 11/300/1000/4000, extra occupancy at 12/301/2500) and reproduces your numbers verbatim:

```
bo 8 -> ones [(2,2)] [(6,3)] [] []        twos [(3,1)] [(6,2)] [(14,8)] [(12,0)]
bo 7 -> ones [(2,2), (6,19)]              twos [(3,1), (6,18), (30,24)]  | block 2: [] / [(28,16)]
bo 6 -> ones [(2,2), (6,19), (40,26)]     twos [(3,1), (6,18), (30,24), (60,48)]
```

`test_mask_deinterleaves_at_the_block_depth` is parametrized over `block_order` 8/7/6 with those exact position sets (so `block_depth` 4/5/6 against a leaf-wide bitmap, `block_depth != depth` in two of the three), and asserts the mask shape per block. One thing the crossing exposed and the test now records: only POPULATED blocks are yielded, so at `block_order=7` the leaf block holding *only* occupancy cell 2500 is never emitted — `test_block_mask_states_agree_with_the_per_chunk_read` pins that the whole-leaf block's state-2 count equals the per-chunk sweep's total while its state-1 count is that total **+1** (cell 2500, which no populated chunk contains).

PR #343, opened 2026-07-30
v2 Event transport: always-on status objects, polling resolver, reattach, agg event mode (issue #327)

### LINKED ISSUE #327: zagg.client v2 transport: Event invoke + status-object future resolver

espg:
Closes #327. Refs #326 (built directly on the merged #333 client facade).

Implements the v2 Event transport per the ratified plan on #327 — the [implementation plan](https://github.com/englacial/zagg/issues/327#issuecomment-5124738952) plus its three amendments ([EventInvokeConfig correction](https://github.com/englacial/zagg/issues/327#issuecomment-5124759868), [always-on ratification](https://github.com/englacial/zagg/issues/327#issuecomment-5124783802), [prefix-delete hygiene + fail-open](https://github.com/englacial/zagg/issues/327#issuecomment-5124806827)) — and the session ruling that `agg` grows `invocation="event"` (2026-07-29).

## What this does

Workers now record every per-unit outcome as a small JSON **status object** at `<store>.status/run-<run_id>/shard-<decimal-key>.json` — always-on (every run, every dispatcher, every status branch including the caught-error 500) and **fail-open** (a status PUT failure logs and never affects the shard result). On top of that channel:

- **`Run.dispatch(transport="event")`** — fire-and-forget `InvocationType="Event"` invokes (202), all futures resolved by **one poller thread doing one LIST of the run prefix per tick** (backoff 2 s → 15 s cap, reset on progress). No held connections; `as_completed`/`status()`/`progress()`/`wait()`/tail semantics identical to v1. `max_retries` migrates to the **three-class retry policy** below (amendment (b) as folded from review); dispatch is **paced** to the concurrency window because the fleet's `MaximumEventAgeInSeconds: 60` silently drops invokes under backpressure (amendment (a′)).
- **Drop detection** — a shard with no status object after `function timeout + 60 s max event age + 90 s margin` resolves the distinct **`failed-unknown`** outcome (`payload["outcome"]`, counted failed by `status()`); a drop re-fires exactly once (class (b) below).
- **Dispatch manifest** — the existing setup invoke (hive fire-and-forget setup / flat sync setup) carries a `run_manifest` block; the **worker** writes `run-<id>/manifest.json` (run_id, shard keys as decimal strings, `semantic_hash`, `dispatched_at`, dataset identity, and the setup event's own config folded in). Size-gated on the hive async setup (dropped over the 256 KB cap, never fatal).
- **`Run.attach(store, run_id, ...)`** — rebuilds a `RunHandle` from manifest + current statuses; observe-only (no granule records → no re-dispatch); drop deadline anchored at the manifest's `dispatched_at`; the post-run tail runs **only if not already recorded** — the `mode="stats"` worker writes a `tail.json` marker when the tail's record leg completes **and** finalize succeeded (a run carrying #335's `finalize_error` withholds it, so a reattached handle re-runs the idempotent backstop), checked read-only.
- **`agg(..., invocation="event")`** — same blocking call-shape, summary shape, and exceptions as sync; Event dispatch + the same shared poller underneath (`_cell_work` blocks on the shard's future, so the accumulator/summary/tail are untouched and the pool width provides the pacing). Raster/temporal keep refusing anything but async/sync. **The benchmark harness and CI stay on sync** (their metrics come from invoke responses; `run_benchmark.py` untouched).
- **Tripwire** — `bench_objects` gains `_is_status_object` (generalizing the `_is_run_parquet` telemetry filter, ratified amendment): anything under a `.status` prefix segment is excluded from every bucket of the #215/#240 object model.

D8 is intact end to end: every write (status objects, manifest, tail marker) is worker-side off an invoke; the client/poller side only LISTs and GETs — asserted by an explicit audit test that counts every PUT during a full v2 run.

## Status object schema (as implemented)

```json
{
  "schema_version": 1,
  "status": "ok" | "failed" | "no_data",
  "attempt_id": "<uuid4 hex, fresh per execution>",
  "run_id": "<dispatcher run id>",
  "shard": "<decimal shard key>",
  "window": null | "<window label>",
  "timings": { "read": 1.0, "aggregate": 0.5, "write": 0.2 },
  "error": null | "<error string>",
  "status_code": 200,
  "body": { "...": "the full response envelope body" }
}
```

`body` is one field beyond the ratified minimum — see Questions.

## Retry policy (three fault classes)

Folded from the PR review, which found that re-dispatching on **any** non-ok status diverged
from a recorded decision and could bill `max_retries` full executions for an unrecoverable
shard. Function errors never return to an Event caller, so v1's single `max_retries` loop is
split by what actually failed:

| class | trigger | behavior |
|---|---|---|
| (a) deterministic worker failure | a `failed` **status object** | never re-fires; resolves immediately |
| (b) queue drop | no status object by the drop deadline (`failed-unknown`) | re-fires **exactly once** (`_MAX_DROP_REFIRES = 1`) |
| (c) invoke-layer fault | `client.invoke` itself raised (throttle, read timeout) | retries to `max_retries` with `2**attempts` s backoff |

- **(a)** A status object exists ⇒ the worker ran to completion and reported its own error ⇒
  deterministic for that shard. This is the rule the sync transport already follows: *"Every
  `FunctionError` on a synchronous invoke is deterministic for a given shard … so none are
  retried"* (#119). Re-firing buys a guaranteed second failure at full cost.
- **(b)** The one class where a re-fire can actually succeed, since an invoke that aged out of
  the async queue never ran. Bounded at one re-fire because a drop and an OOM/timeout death are
  indistinguishable from the client, so the exposure is **at most 2 × function timeout** billed
  for a shard that may be unrecoverable. That bound is the point: #151 pinned
  `MaximumRetryAttempts: 0` precisely so async fan-out is never re-billed at 900 s per service
  retry, and an unbounded client-side re-dispatch would reintroduce it from the outside.
  `max_retries <= 1` opts out entirely, keeping the knob's "no retries" meaning.
- **(c)** `max_retries` in its original v1 meaning. Backoff is a **not-before timestamp** on the
  entry, not a `time.sleep` — one thread serves every shard, so sleeping would stall the LISTs,
  and an unstamped re-queue burned the whole budget in the same dispatch pass (measured: 5
  invokes in 0.3 ms), turning a transient `TooManyRequestsException` into a wholesale run
  failure. Throttling is the canonical fleet-scale fault, so this is the class that most needed
  the fix.

**@espg's to veto** — this is a policy choice, not a bug fix, and it changes what a `max_retries`
value means on the event transport (it no longer governs worker-reported failures at all). Say
the word and any of the three classes can move; the code keeps them in three separate methods
(`_worker_failed` / `_drop_failed` / `_invoke_faulted`) so a change is one method deep.

## Phases

- [x] Phase 1 — worker-side status objects (always-on, fail-open) + tripwire exclusion (`4688c17`)
- [x] Phase 2 — dispatch manifest via the setup invoke (`95ef221`)
- [x] Phase 3 — client Event transport + `StatusPoller` + `_build_cell_event`/`_cell_payload` extraction (`007f89c`)
- [x] Phase 4 — drop detection: `failed-unknown` + budget participation, fake-clock tests (`2194771`)
- [x] Phase 5 — `Run.attach` + tail-completion marker (`fc0267f`)
- [x] Phase 6 — `agg(invocation="event")` + `rows_from_status` dual-shape (`324f9e1`)
- [x] Phase 7 — cross-cutting tests: D8 write audit, dual-shape rows (`65025b8`)
- [x] Module-ceiling split + typing hygiene + a test-ordering fix (`ab411f4`, `f06a506`, `3851a2a` — the tqdm test in `test_client.py` reload()s `zagg.client` mid-suite, so the transport tests resolve `ShardError` at call time)
- [x] Merge of `origin/main` (`142c4d7` — main advanced with the issue #335 guarded-finalize tail mid-implementation; the `_dispatch_run_stats` overlaps keep BOTH sides: `finalize_error` (#335) and `tail_status_url` + the event-run stats pointer (#327), applied inside #335's new tail `try:` structure)
- [x] Adversarial-review fold — 8 findings, one commit each (`862e23a`, `009a0f6`, `a8dd7b1`, `7ce6c5a`, `3818c81`, `5f448d5`, `7876f4b`, `de583d8`): the retry-policy split above, the throttle backoff, the event-run run-record pointer, the post-shutdown `register` guard, the withheld tail marker on a failed finalize, the schema/attempt-id read guards, the corrected per-tick LIST cost + same-tick slot refill, and the windowed-key + attach-D8 coverage

## How it was tested

New `tests/test_client_transport.py` (46 tests): a stub Lambda whose Event "workers" write status objects into an obstore `MemoryStore` **through the real `build_shard_status`**, so the worker and client halves cannot drift. Covers: poller resolution, event/sync payload parity (byte-identical modulo InvocationType), each retry class (a `failed` status resolving with zero re-dispatches, a drop re-firing exactly once at `max_retries=5`, a throttle backing off on an asserted 2/4/8 s schedule), attempt-id dedupe (a duplicate execution's object is acted on once) and its falsy-id guard, an unknown `schema_version` warning, post-shutdown `register` refusal, same-tick slot refill, pacing window (`max_in_flight=1` provably holds), drop classification with a fake clock + budget participation, attach mid-run/post-run/failed/stale/missing-manifest/windowed-refusal, agg event↔sync parity (same events modulo InvocationType, same tail mode sequence, same summary counts), a windowed 2-window event run pinning the per-window status key across both halves of the channel, the D8 PUT audit on **both** the dispatch and attach paths, and `rows_from_status` over both object shapes.

Plus: 9 handler tests for the status write (every branch incl. caught-error 400/500, window suffix, no-run-id no-op, raising store → shard result unaffected — the ratified fail-open test), 4 for the dispatch manifest, 4 for the tail marker; 6 client tests for the `run_manifest` block (client/agg block equality, size gate); 2 tripwire tests.

Local gates: `pytest tests/test_client.py tests/test_runner.py tests/test_processing.py tests/test_benchmark_objects.py tests/test_lambda_handler.py tests/test_client_transport.py` green; `ruff check` + `ruff format --check` clean on every touched file; mypy clean on the new module. Full suite green except pre-existing/environmental failures: `TestFunctionBuild::test_function_build_succeeds` (known), and `test_full_aoi_benchmark`/`test_raster_benchmark`/`test_shardmap`/`test_sources`/`test_viz` items that need `pyarrow` (catalog extra absent in this env). The known `TestConsolidationGate` pair passes locally.

## Questions for review

1. **Status objects carry the full envelope `body`** — one field beyond the ratified `{status, attempt_id, run_id, shard, timings, error}` schema. Without it the event transport cannot reproduce sync's results (`total_obs`/`cells_with_data` in the summary, `body["stats"]` for the run parquet and sweep leaves, `time_range` for coverage). The body is a few KB (granule ids ride as `granules_sha256`, never as lists). Flag if you'd rather enumerate fields.
2. **Shard naming uses the raw decimal key** (`shard-<int>`), not the issue #199 morton label — the caught-error handler path must name the object without constructing a grid. The manifest lists the same decimal strings, so the join is consistent; labels remain in messages/`ShardError`.
3. **Drop margin** is 90 s (`_DROP_MARGIN_S`, mirroring `_ASYNC_POLL_MARGIN_S`), so the deadline is `timeout + 60 + 90`. Poll-tick LIST pagination rides `obstore.list`'s internal pagination (one logical LIST per tick).
4. **Manifest size gate**: on the hive async setup the `run_manifest` block is dropped when it would push the event over the 256 KB cap (roughly 10k+ shards) — attach then degrades to statuses only. Keeping it would need a different transport for the manifest (another worker invoke); deferred.
5. **Both modules are over the §4 ~1000-line ceiling** and I am raising it rather than splitting further on my own. `client.py` is 1148 lines (912 pre-#327; the poller/transport/attach machinery already lives in the new `src/zagg/client_transport.py` — the split the plan pre-authorized), and `client_transport.py` is now **1099** lines, up from 854 because the review fold added the retry-policy rationale, the corrected per-tick cost paragraph, and the read-guard docstrings. Both overages are prose, not logic: `ruff`'s own count of code lines is well under. Three options — (a) leave it, since the growth is documentation of ratified decisions; (b) split the worker half (`build_shard_status`/`write_*`, ~250 lines) into `zagg/status_objects.py`, leaving the poller and reattach in `client_transport.py`; (c) move the retry-policy prose to a docs page and cross-reference it. Your call — I did not want a module split to ride in on a review fold.
6. **Payload-cap behavior differs slightly on the event transport**: the 256 KB check runs for ALL shards before anything fires (fail-whole-run early) instead of mid-fan-out per shard.
7. **Scope notes**: local-backend runs write no status objects (no worker handler in that path); `Run.attach` covers unwindowed spatial runs (windowed units are (shard, window) pairs the manifest does not enumerate) — `agg(invocation="event")` does handle windowed configs (status keys carry the window suffix; a 2-window run is now covered locally per the review, still untested at fleet scale).
8. **`deployment/aws/lambda_handler.py` was modified** — the issue's phase 1 explicitly requires the worker-side status write, and the additions are three thin, doubly-fail-open delegations into `zagg.client_transport` (the #297/#300/#313 precedent for handler changes). No infra files (template.yaml, build/standup scripts) touched, nothing run against live AWS.

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -154,6 +154,10 @@ the dispatcher at end of run, like coverage mode):
     "rows_from": str (optional) -- the run's async status prefix; the worker
         assembles success rows from the mirrored result envelopes when the
         row set exceeds the async payload budget,
+    "tail_status_url": str (optional, issue #327) -- where to ALSO write the
+        run's tail-completion marker on success (the v2 status prefix's
+        tail.json); a reattached handle (Run.attach) then skips a tail that
+        already ran. Fail-open; absent -> no write, byte-identical.
     "output_credentials": dict (optional, same shape as process mode),
 }
 
@@ -645,6 +649,16 @@ def _write_dispatch_manifest(event: Dict[str, Any]) -> None:
         logger.warning(f"dispatch manifest write failed (fail-open, issue #327): {e}")
 
 
+def _write_tail_status(event: Dict[str, Any]) -> None:
+    """Tail-completion marker off the stats event (issue #327), doubly fail-open."""
+    try:
+        from zagg.client_transport import write_tail_status
+
+        write_tail_status(event, _output_store_kwargs(event))
+    except Exception as e:
+        logger.warning(f"tail status write failed (fail-open, issue #327): {e}")
+
+
 def _write_result(result_url: str, response: Dict[str, Any], event: Dict[str, Any]) -> bool:
     """Write the response envelope to ``result_url`` as JSON (issue #151).
 
@@ -1063,6 +1077,7 @@ def _handle_stats(event: Dict[str, Any]) -> Dict[str, Any]:
         if not rows:
             # A pointer prefix that yielded nothing (and no inline rows):
             # nothing to persist — not an error (fail-open telemetry).
+            _write_tail_status(event)
             return {
                 "statusCode": 200,
                 "body": json.dumps({"ok": True, "mode": "stats", "rows": 0}),
@@ -1074,6 +1089,10 @@ def _handle_stats(event: Dict[str, Any]) -> Dict[str, Any]:
             timestamp=event.get("timestamp"),
             store_kwargs=_output_store_kwargs(event),
         )
+        # Tail-completion marker (issue #327): the stats leg is the recorded
+        # end of the post-run tail, so a reattached handle can skip a tail
+        # that already ran. Fail-open; absent url -> no-op.
+        _write_tail_status(event)
         return {
             "statusCode": 200,
             "body": json.dumps({"ok": True, "mode": "stats", "rows": len(rows), "path": path}),
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -568,6 +568,12 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     # setup/finalize/extract bodies stay byte-identical (their consumers don't
     # aggregate container state).
     response = _attach_container_telemetry(response, telemetry)
+    # Per-shard status object (issue #327): always-on for every per-unit
+    # response carrying a run identity, every status branch (200/400/500,
+    # including the caught-error envelope) -- the v2 Event transport resolves
+    # futures from these instead of the invoke response. Fail-open by
+    # ratification: never affects the shard result.
+    _write_shard_status(event, response)
     # Async result channel (issue #151): on an Event invoke the return value is
     # discarded, so mirror the response envelope to the orchestrator-supplied
     # result_url for it to poll. Covers every branch (200 / 400 / 500) of both
@@ -592,6 +598,22 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
         if mirrored:
             _maybe_self_recycle()
     return response
+
+
+def _write_shard_status(event: Dict[str, Any], response: Dict[str, Any]) -> None:
+    """Always-on per-shard status object (issue #327), doubly fail-open.
+
+    The body lives in ``zagg.client_transport.write_shard_status`` (itself
+    fail-open); this wrapper additionally swallows an import/lookup failure so
+    a function zip missing the module cannot fail a shard either. Uses the
+    same output-store resolution as every other worker write.
+    """
+    try:
+        from zagg.client_transport import write_shard_status
+
+        write_shard_status(event, response, _output_store_kwargs(event))
+    except Exception as e:
+        logger.warning(f"shard status write failed (fail-open, issue #327): {e}")
 
 
 def _write_result(result_url: str, response: Dict[str, Any], event: Dict[str, Any]) -> bool:
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -84,6 +84,13 @@ writes the raster (time, cells) template instead, from a synchronous invoke):
         time coordinate, int64 microseconds since the epoch; the orchestrator
         owns the global timestep index and threads it here so the template
         write needs no S3 access from the dispatcher.
+    "run_manifest": dict (optional, issue #327) -- {"run_id", "shards"
+        (decimal shard-key strings), "semantic_hash", "dispatched_at",
+        "dataset"} dispatch identity: on a successful setup the worker writes
+        it (plus this event's "config") as
+        "<store>.status/run-<run_id>/manifest.json" -- what Run.attach
+        rebuilds a handle from (D8: the dispatcher never writes). Fail-open;
+        absent -> no write, byte-identical to pre-#327 events.
     "output_credentials": dict (optional, same shape as process mode),
 }
 
@@ -543,7 +550,13 @@ def lambda_handler(event: Dict[str, Any], context: Any) -> Dict[str, Any]:
     telemetry = _container_telemetry()
     mode = event.get("mode", "process")
     if mode == "setup":
-        return _handle_setup(event)
+        response = _handle_setup(event)
+        # Dispatch manifest (issue #327 phase 2): the run's shard set +
+        # identity, recorded by the WORKER off the same setup invoke (D8) so
+        # a run is reattachable by run id. Fail-open; absent block -> no-op.
+        if response.get("statusCode") == 200:
+            _write_dispatch_manifest(event)
+        return response
     if mode == "finalize":
         return _handle_finalize(event)
     if mode == "ping":
@@ -614,6 +627,22 @@ def _write_shard_status(event: Dict[str, Any], response: Dict[str, Any]) -> None
         write_shard_status(event, response, _output_store_kwargs(event))
     except Exception as e:
         logger.warning(f"shard status write failed (fail-open, issue #327): {e}")
+
+
+def _write_dispatch_manifest(event: Dict[str, Any]) -> None:
+    """Run dispatch manifest off the setup event (issue #327), doubly fail-open.
+
+    Mirrors ``_write_shard_status``: the body lives in
+    ``zagg.client_transport.write_dispatch_manifest`` (itself fail-open); this
+    wrapper additionally swallows an import failure so a function zip missing
+    the module cannot fail the setup either.
+    """
+    try:
+        from zagg.client_transport import write_dispatch_manifest
+
+        write_dispatch_manifest(event, _output_store_kwargs(event))
+    except Exception as e:
+        logger.warning(f"dispatch manifest write failed (fail-open, issue #327): {e}")
 
 
 def _write_result(result_url: str, response: Dict[str, Any], event: Dict[str, Any]) -> bool:
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -1045,6 +1045,7 @@ class Run:
         # the failure (espg ruling on PR #333, middle option) so a notebook sees
         # it in-stream rather than only at the join, retries once, and RETURNS
         # the error — the handle keeps this path's re-raise semantics.
+        finalize_error = None
         finalize_kwargs: dict = {}
         if layout == "hive":
             finalize_kwargs = {
@@ -1068,6 +1069,7 @@ class Run:
                 ),
             )
             if error is not None:
+                finalize_error = str(error)
                 handle._finalize_error = error
                 logger.warning(f"finalize invoke failed (surfaced via handle.wait()): {error}")
 
@@ -1117,6 +1119,12 @@ class Run:
             output_creds_event=output_creds_event,
             store_kwargs=runner._output_store_kwargs(output_creds_event, self.region),
             inline_rows=stats_inline,
+            # Same wire contract as runner's tail (issue #335): always present,
+            # None on success, so the parquet's column set does not vary — and
+            # it is what makes the worker WITHHOLD the tail marker on a failed
+            # finalize, so a reattached handle re-runs the idempotent backstop
+            # instead of reading "recorded" as "succeeded" (review, PR #343).
+            finalize_error=finalize_error,
             tail_status_url=f"{run_status_prefix(self.store, run_id)}/{TAIL_NAME}",
         )
         if layout == "hive" and get_sweep(self.config):
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -594,13 +594,25 @@ class Run:
         )
         run_id = uuid.uuid4().hex
 
+        # Dispatch manifest block (issue #327): rides the setup invoke so the
+        # WORKER records the run's shard set + identity at the status prefix
+        # (D8) — what :meth:`Run.attach` rebuilds a handle from.
+        from zagg.client_transport import build_run_manifest_block
+
+        md = self.catalog_data.get("metadata") or {}
+        run_manifest = build_run_manifest_block(
+            run_id,
+            [k for k, _ in cells],
+            self.config,
+            dataset={"short_name": md.get("short_name"), "version": md.get("version")},
+        )
+
         # Setup handshake, synchronous and load-bearing (worker-side writes
         # only — D8). Hive: fail-fast ping + fire-and-forget manifest write;
         # finalize below is the idempotent backstop (issue #252 hybrid). Flat:
         # the template write itself.
         dataset = None
         if get_store_layout(self.config) == "hive":
-            md = self.catalog_data.get("metadata") or {}
             dataset = {"short_name": md.get("short_name"), "version": md.get("version")}
             runner._invoke_lambda_ping(
                 client,
@@ -621,6 +633,7 @@ class Run:
                 parent_order=self._parent_order,
                 overwrite=self.overwrite,
                 output_creds_event=output_creds_event,
+                run_manifest=run_manifest,
             )
         else:
             runner._invoke_lambda_setup(
@@ -632,6 +645,7 @@ class Run:
                 overwrite=self.overwrite,
                 config_dict=config_dict,
                 output_creds_event=output_creds_event,
+                run_manifest=run_manifest,
             )
 
         aoi_by_shard = runner._aoi_payload_map(self.catalog_data)
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -644,9 +644,11 @@ class Run:
             tick — no held connections, and the run survives client exit
             (reattachable via :meth:`attach`). Futures, harvest iterators,
             ``status()``, and the post-run tail behave identically; the retry
-            policy migrates to status-driven re-dispatch (same
-            ``max_retries``), and a shard whose status object never lands
-            resolves ``failed-unknown``. Requires a deployed worker that
+            policy splits by fault class (see
+            :class:`~zagg.client_transport.StatusPoller`) — ``max_retries``
+            still bounds invoke-layer faults, a worker-reported failure is
+            terminal, and a shard whose status object never lands re-fires once
+            then resolves ``failed-unknown``. Requires a deployed worker that
             writes status objects (issue #327) and read access to the status
             prefix for the poll.
         """
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -112,6 +112,26 @@ class ShardError(RuntimeError):
         self.payload = payload
 
 
+def _fail_with_shard_error(entry, result: dict) -> None:
+    """v2 poller ``on_failed``: a terminal failure raises :class:`ShardError`.
+
+    Shared by :meth:`Run.dispatch` (event transport) and :meth:`Run.attach` so
+    a failed status, a spent retry budget, and the ``failed-unknown`` drop
+    outcome all surface from ``future.result()`` exactly like the v1 transport
+    — the payload is the same per-shard result dict, with ``outcome`` marking
+    a drop.
+    """
+    detail = result.get("error") or f"status {result.get('status_code')}"
+    entry.future.set_exception(
+        ShardError(
+            f"shard {entry.label}: {detail}",
+            shard_key=entry.shard_key,
+            label=entry.label,
+            payload=result,
+        )
+    )
+
+
 class RunHandle:
     """Live handle over one dispatched fan-out: per-shard futures + rollup.
 
@@ -461,6 +481,151 @@ class Run:
             source_credentials=source_credentials,
         )
 
+    @classmethod
+    def attach(
+        cls,
+        store: str,
+        run_id: str,
+        *,
+        function_name: str | None = None,
+        region: str = "us-west-2",
+        lambda_client=None,
+        output_credentials: dict | None = None,
+        output_endpoint_url: str | None = None,
+    ) -> RunHandle:
+        """Reattach to a dispatched run by id (issue #327 phase 5).
+
+        Rebuilds a :class:`RunHandle` from the run's dispatch manifest
+        (``<store>.status/run-<run_id>/manifest.json`` — the config, shard
+        set, and identity the setup invoke recorded worker-side) plus the
+        current status objects: shards already settled resolve on the first
+        poll tick, the rest resolve as their statuses land. A notebook kernel
+        restart or a second machine can adopt a running fleet this way — the
+        Event invokes are already in flight and owe the client nothing.
+
+        Observe-only by design: the manifest carries no granule records, so a
+        ``failed`` status resolves that shard's :class:`ShardError`
+        immediately (no re-dispatch), and a shard with no status object by
+        the drop deadline — anchored at the manifest's ``dispatched_at`` —
+        resolves ``failed-unknown``. The post-run tail runs only if not
+        already recorded (the run's ``tail.json`` marker, checked read-only);
+        when it does run it is the same worker-invoke tail as a live handle
+        (D8 intact — this method's own store traffic is reads).
+
+        Parameters
+        ----------
+        store : str
+            The run's effective store root exactly as dispatched
+            (``handle.store_path`` / the summary's ``store_path``).
+        run_id : str
+            The dispatch's run id (``manifest.json`` names it; it is also in
+            every cell event and stats record).
+        function_name, region, lambda_client, output_credentials,
+        output_endpoint_url
+            Same semantics as :meth:`from_config`; the function/client are
+            only used for the tail invokes when the tail is not yet recorded.
+        """
+        from zagg import client_transport, runner
+
+        output_creds_event = runner._build_output_creds_event(
+            output_credentials, output_endpoint_url, region
+        )
+        store_kwargs = runner._output_store_kwargs(output_creds_event, region)
+        prefix = client_transport.run_status_prefix(store, run_id)
+        manifest = client_transport.read_dispatch_manifest(prefix, store_kwargs)
+        if manifest is None:
+            raise ValueError(
+                f"no dispatch manifest at {prefix}/{client_transport.MANIFEST_NAME} — "
+                f"wrong store/run_id, a dispatcher predating issue #327, or a hive "
+                f"setup Event that was dropped (its manifest write is best-effort)"
+            )
+        if not manifest.get("config"):
+            raise ValueError(f"dispatch manifest for run {run_id} carries no config")
+        config = load_config_from_dict(manifest["config"])
+        validate_config(config)
+        if get_windowing(config) is not None:
+            raise NotImplementedError(
+                "Run.attach covers unwindowed spatial runs; windowed units are "
+                "(shard, window) pairs the manifest does not enumerate"
+            )
+        shards = [int(s) for s in manifest.get("shards") or []]
+        if not shards:
+            raise ValueError(f"dispatch manifest for run {run_id} lists no shards")
+
+        grid = grid_from_config(config)
+        catalog_data = {
+            "shard_keys": shards,
+            "granules": [[] for _ in shards],
+            "grid_signature": grid.spatial_signature(),
+            "metadata": manifest.get("dataset"),
+        }
+        run = cls(
+            config,
+            catalog_data,
+            store=store,
+            function_name=runner._resolve_function_name(config, function_name),
+            region=region,
+            lambda_client=lambda_client,
+            driver=get_driver(config),
+            handoff=get_handoff(config),
+            output_credentials=output_credentials,
+            output_endpoint_url=output_endpoint_url or get_output_endpoint_url(config),
+        )
+        client = lambda_client
+        if client is None:
+            import boto3
+            from botocore.config import Config
+
+            client = boto3.Session().client(
+                "lambda",
+                region_name=region,
+                config=Config(read_timeout=960, connect_timeout=10, retries={"max_attempts": 0}),
+            )
+
+        # Drop deadline anchored at the manifest's dispatched_at: attaching
+        # long after the fleet finished classifies status-less shards
+        # immediately instead of waiting a fresh window.
+        dispatched_at = None
+        try:
+            from datetime import datetime
+
+            dispatched_at = datetime.fromisoformat(manifest["dispatched_at"]).timestamp()
+        except (KeyError, TypeError, ValueError):
+            pass
+        timeout_s = runner._get_function_timeout_s(client, run.function_name)
+        poller = client_transport.StatusPoller(
+            lambda: client_transport.open_status_store(prefix, store_kwargs),
+            drop_timeout_s=client_transport.drop_timeout_s(timeout_s),
+            on_failed=_fail_with_shard_error,
+        )
+        futures: dict[int, Future] = {}
+        for key in shards:
+            label = runner._safe_label(run.grid, key)
+            futures[key] = poller.register(key, label, dispatch=None, dispatched_at=dispatched_at)
+        poller.start()
+
+        handle = RunHandle(futures, store_path=store)
+        finisher = threading.Thread(
+            target=run._post_run,
+            args=(
+                handle,
+                client,
+                poller,
+                asdict(config),
+                manifest.get("dataset"),
+                output_creds_event,
+                run_id,
+            ),
+            kwargs={
+                "tail_done_check": lambda: client_transport.tail_recorded(prefix, store_kwargs)
+            },
+            name="zagg-client-finish",
+            daemon=True,
+        )
+        handle._finisher = finisher
+        finisher.start()
+        return handle
+
     # -- shard selection ----------------------------------------------------
 
     def _resolve_key(self, key) -> int:
@@ -758,24 +923,12 @@ class Run:
         # the helper falls back to the template default (an injected stub
         # client need not answer get_function_configuration).
         timeout_s = runner._get_function_timeout_s(client, self.function_name)
-
-        def _failed(entry, result):
-            detail = result.get("error") or f"status {result.get('status_code')}"
-            entry.future.set_exception(
-                ShardError(
-                    f"shard {entry.label}: {detail}",
-                    shard_key=entry.shard_key,
-                    label=entry.label,
-                    payload=result,
-                )
-            )
-
         poller = client_transport.StatusPoller(
             lambda: client_transport.open_status_store(prefix, store_kwargs),
             drop_timeout_s=client_transport.drop_timeout_s(timeout_s),
             max_retries=self.max_retries,
             max_in_flight=workers,
-            on_failed=_failed,
+            on_failed=_fail_with_shard_error,
         )
         futures: dict[int, Future] = {}
         for key, records in cells:
@@ -900,6 +1053,7 @@ class Run:
         dataset: dict | None,
         output_creds_event: dict | None,
         run_id: str,
+        tail_done_check=None,
     ) -> None:
         """Finisher-thread entry point: run the tail, never lose its failure.
 
@@ -909,9 +1063,17 @@ class Run:
         distinct from the D6-load-bearing ``_finalize_error``) and ``closer``
         — the v1 invoke pool or the v2 status poller, both
         ``shutdown(wait=False)``-shaped — is shut down either way (review
-        finding, PR #333).
+        finding, PR #333). ``tail_done_check`` (issue #327, the attach path):
+        a zero-arg read-only predicate consulted AFTER every shard settles —
+        True means the tail was already recorded (the run's ``tail.json``
+        marker), so a reattached handle never re-runs it.
         """
         try:
+            if tail_done_check is not None:
+                futures_wait(list(handle.futures.values()))
+                if tail_done_check():
+                    logger.info(f"post-run tail already recorded for run {run_id}; skipping")
+                    return
             self._run_tail(handle, client, config_dict, dataset, output_creds_event, run_id)
         except BaseException as e:
             handle._tail_error = e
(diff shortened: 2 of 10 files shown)


src/zagg/client_transport.py
```diff
@@ -0,0 +1,857 @@
...
+        entry.attempts += 1
+        entry.dispatched_at = self._clock()
+        if entry.first_dispatched_at is None:
+            entry.first_dispatched_at = entry.dispatched_at
+        try:
+            dispatch()
+        except Exception as e:
+            # An invoke fault burns this attempt (throttles included): the
+            # retry budget re-fires it on a later tick, mirroring the sync
+            # transport's bounded transient retry.
```
espg: 🤖 *from Claude (review)* — **severity: medium-high (correctness/cost at exactly the scale this PR targets)**

This comment is not what the code does. `_attempt_failed` sets `entry.queued = True`, and `_dispatch_up_to_window`'s `while True` re-reads `queued` in the **same pass** — so there is no "later tick" and no backoff at all. Measured on this branch (`max_retries=5`, a `dispatch` that always raises `TooManyRequestsException`):

```
PROBE throttle: 5 invokes in 0.3 ms
  inter-attempt gaps (ms): [0.12, 0.03, 0.02, 0.02]
  result retries=4 error='Event invoke failed: TooManyRequestsException: Rate exceeded'
```

The whole retry budget is spent in a third of a millisecond. The sync transport it claims to mirror does the opposite for the *same* error strings — `runner.py:5381`, `time.sleep((2**attempt) + (time.time() % 1))` gated on `["TooManyRequestsException", "Rate exceeded", "Read timeout", ...]`.

Why this matters here specifically: Lambda `Invoke` throttling is the canonical fleet-scale fault, and it is the exact condition the pacing window exists to manage (amendment (a′)). On the first backpressure event every shard in flight burns its budget instantly and resolves `failed`, converting a transient throttle into a wholesale run failure.

Suggested fix — since this runs on the poller thread, sleeping is the wrong tool: stamp a `retry_not_before = self._clock() + 2**entry.attempts` on the entry and have `_dispatch_up_to_window` skip queued entries whose stamp is in the future. That keeps the thread free for LISTs and gives the same bounded transient retry the docstring promises.

Note `test_invoke_fault_burns_an_attempt_and_retries` passes either way — it only counts calls, never the interval — so the test doesn't pin the property the comment asserts.

src/zagg/client_transport.py
```diff
@@ -0,0 +1,857 @@
...
+            "timeout": False,
+            "granule_count": entry.granule_count,
+        }
+        if outcome is not None:
+            result["outcome"] = outcome
+        return result
+
+    def _attempt_failed(self, entry: _ShardEntry, result: dict) -> None:
+        """Status-driven retry (issue #327 amendment (b)): re-dispatch or resolve."""
+        if entry.dispatch is not None and entry.attempts < self._max_retries:
```
espg, line 1058: 🤖 *from Claude (review)* — **severity: medium-high (divergence from a ratified policy; 3× billing on unrecoverable shards)**

Re-dispatch fires on **any** non-ok status, including the handler's caught-error 500. The sync transport deliberately does the opposite, and it is a recorded decision: *"Every `FunctionError` on a synchronous invoke is deterministic for a given shard — a timeout, `Runtime.OutOfMemory`, or an exception that escaped the handler — so none are retried"* (`runner.py` `_invoke_lambda_cell`, issue #119). `max_retries` in v1 governs **only** the transient invoke-layer faults in the `except` block.

A `failed` status object by construction means the worker **ran to completion and reported an error** — precisely the deterministic case #119 excludes — yet here it costs up to `max_retries` full executions.

The drop path is the sharper version: an OOM- or timeout-killed worker writes no status (the module docstring says so at lines 39-40 — a drop and a death are indistinguishable from the client), so it lands as `failed-unknown` and `_dropped` → `_attempt_failed` re-dispatches it. Default `max_retries=3` ⇒ **3 × 900 s billed for a shard that cannot succeed**. That is the re-billing exposure `MaximumEventAgeInSeconds`/`MaximumRetryAttempts: 0` was pinned to close (#151, quoted in this file's own docstring at lines 36-38 — *"precisely so async fan-out is never re-billed at 900 s per service retry"*). The v2 client reintroduces it from the outside.

Amendment (b) ratified "same limit" — and the sync limit for a FunctionError is **zero**. Two concrete options:

1. Gate re-dispatch on the fault class: transport-level invoke exceptions retry (with the backoff from the adjacent finding); a `failed` **status** resolves immediately, like v1.
2. Keep re-dispatching `failed-unknown` (that is the drop case worth retrying) but cap it at 1, and/or make drop-vs-death distinguishable — e.g. the worker writing a `started` marker at handler entry, which turns "no status at all" into "never started ⇒ dropped ⇒ retry" vs "started, no status ⇒ died ⇒ don't".

If the current all-failures-retry behavior *is* the intent, please say so explicitly in the PR body so the cost is a recorded decision rather than an emergent one — this is the kind of thing that shows up as a surprise line on a full-AOI run.

src/zagg/client.py
```diff
@@ -875,9 +1066,13 @@ def _run_tail(
             except Exception as e:
                 logger.warning(f"root coverage.moc dispatch failed (fail-open, D9): {e}")
 
-        # Run-record + sweep, both fail-open (issues #297/#313/#300): sync
-        # transport -> no status prefix, so oversized row sets skip inside
-        # _dispatch_run_stats exactly like agg's invocation="sync" path.
+        # Run-record + sweep, both fail-open (issues #297/#313/#300): no #151
+        # envelope-mirror prefix on either transport, so oversized row sets
```
espg: 🤖 *from Claude (review)* — **severity: medium (silently loses the run record on exactly the large runs this transport is for)**

"no #151 envelope-mirror prefix on either transport" is true as written, but it draws the wrong conclusion for `transport="event"`: that transport *does* have a per-run prefix full of status objects, and this same PR taught `telemetry.rows_from_status` to read the #327 object shape.

The agg side does the right thing — `runner.py:3891`:

```python
result_prefix=result_prefix if result_prefix is not None else status_prefix,
```

The facade hardcodes `result_prefix=None` (line 1083). So for a `Run.dispatch(transport="event")` run whose rows exceed `_RUN_STATS_INLINE_CAP_BYTES`, `_dispatch_run_stats` falls into its last `else` and logs *"exceed the async payload budget and this sync run has no status prefix to point at; skipping"* — the run record is dropped entirely, while the byte-identical `agg(invocation="event")` run keeps it. Fleet-scale runs are the only ones that hit the cap, and they are the whole point of v2.

Fix: thread the transport (or just the status prefix) into `_post_run`/`_run_tail` and pass `run_status_prefix(self.store, run_id)` when the run used the event transport. Worth a test asserting `rows_from` is set on an event-transport tail with an oversized row set.

src/zagg/client_transport.py
```diff
@@ -0,0 +1,857 @@
...
+            future=Future(),
+            dispatch=dispatch,
+            granule_count=granule_count,
+        )
+        if dispatch is None:
+            entry.queued = False
+            entry.attempts = 1
+            entry.dispatched_at = dispatched_at if dispatched_at is not None else self._clock()
+            entry.first_dispatched_at = entry.dispatched_at
+        with self._lock:
```
espg, line 765: 🤖 *from Claude (review)* — **severity: medium (unbounded block is the worst failure mode for a blocking API)**

`register()` has no `self._stop` guard, so a registration that lands after `shutdown()` hands back a Future that **nothing will ever resolve** — the loop is gone, so there is no drop deadline and no error either. Verified on this branch:

```
PROBE register-after-shutdown: NEVER RESOLVES -> TimeoutError
```

(register after `shutdown(wait=True)`, then a matching status object put; `future.result(timeout=1)` times out.) In production `runner.py:3459` is `return fut.result()` with **no timeout**, so the agg pool thread blocks forever and `executor.shutdown()` never returns.

This is the asymmetry that makes it worth fixing rather than just noting: `_run`'s comment at lines 679-683 explicitly designs for registration **trailing** `start()` ("agg's pool registers shards as its threads reach them"), so the entry map is deliberately open-ended on one side — but the shutdown side has no matching guard. Today it's unreachable only because `_run_lambda`'s `finally` orders `executor.shutdown()` before `event_poller.shutdown()` and the executor drains its threads; that ordering is load-bearing and undocumented.

Cheap fix: in `register`, if `self._stop.is_set()`, resolve the entry immediately as a failure (or raise `RuntimeError`) instead of returning a dead Future. A one-line test pins it.

deployment/aws/lambda_handler.py
```diff
@@ -1029,6 +1095,10 @@ def _handle_stats(event: Dict[str, Any]) -> Dict[str, Any]:
             # failure becomes durable. Absent on a pre-#335 dispatcher -> None.
             finalize_error=event.get("finalize_error"),
         )
+        # Tail-completion marker (issue #327): the stats leg is the recorded
+        # end of the post-run tail, so a reattached handle can skip a tail
+        # that already ran. Fail-open; absent url -> no-op.
+        _write_tail_status(event)
```
espg, line 1101: 🤖 *from Claude (review)* — **severity: low-medium (attach skips the tail on precisely the runs that need it re-run)**

The marker is written whenever the stats leg completes — and the stats event also carries `finalize_error` (#335). So `tail.json` reads "recorded" for a run whose **finalize failed**, and `Run.attach` then skips the *entire* tail via `_post_run`'s `tail_done_check`, including the idempotent hive finalize / `ensure_manifest` backstop that a failed-finalize run is exactly the one needing another pass.

The docstrings are careful about this (`TAIL_NAME`: "when the tail's record leg completes"), but `attach`'s use of it isn't — it treats "recorded" as "succeeded". The reason to care is that reattaching after a failure is one of the headline use cases for `Run.attach`.

Two options: don't write the marker when `event.get("finalize_error")` is set (simplest, and the safe direction — attach re-runs an idempotent tail), or record `finalize_error` **in** the marker so `attach` can re-run just the finalize leg and still skip the rollups. Either way worth a test: manifest + statuses + a `tail.json` carrying a finalize error ⇒ attach fires `finalize`.

src/zagg/client_transport.py
```diff
@@ -0,0 +1,857 @@
...
+        """GET one listed status object; act on it once per attempt_id."""
+        import obstore
+
+        try:
+            obj = json.loads(bytes(obstore.get(self._store, entry.key).bytes()))
+        except Exception as e:
+            logger.warning(f"shard {entry.label}: status GET failed (retrying): {e}")
+            return False
+        attempt_id = obj.get("attempt_id")
+        if attempt_id in entry.consumed:
```
espg: 🤖 *from Claude (review)* — **severity: low (read-side robustness on the two fields the channel's forward-compat rests on)**

Two gaps, both verified empirically:

**(a) `schema_version` is write-only.** `STATUS_SCHEMA_VERSION` is stamped on every object and documented "bump on any key change" (line 87-88), but nothing ever reads it. A `{"schema_version": 99, "status": "ok", ...}` object is consumed as v1 in silence (`PROBE schema_version=99 accepted silently`). Since a newer worker against an older client is the *normal* fleet state during a layer roll, the version currently buys nothing on the one scenario it exists for. A `logger.warning` (or a `failed` resolution with a clear message) on an unrecognized major would make a future bump loud instead of subtly wrong — the same posture `telemetry.sidecar_key` takes on an unrecognized spec ("an unrecognized spec RAISES rather than silently defaulting").

**(b) A missing `attempt_id` poisons the dedupe permanently.** `obj.get("attempt_id")` returns `None`, `None` goes into `entry.consumed`, and every *subsequent* attempt's object is then also `None` and ignored. Verified: attempt 1 (no id, `failed`) → re-dispatch fires; attempt 2 (no id, `status: "ok"`) → never consumed, `done=False`; the shard then falls all the way through to `failed-unknown` after the full drop window, reported as a failure despite having succeeded. `build_shard_status` always sets the id, so this needs a foreign/hand-written/truncated object to trigger — but the fix is one line (treat a falsy `attempt_id` as always-new, or refuse the object with a log) and it removes a silent-wrong-answer mode.

src/zagg/client_transport.py
```diff
@@ -0,0 +1,857 @@
...
+
+    # -- resolution -----------------------------------------------------------
+
+    def _list_keys(self) -> set:
+        import obstore
+
+        if self._store is None:
+            self._store = self._store_factory()
+        keys: set = set()
+        for batch in obstore.list(self._store):
```
espg, line 934: 🤖 *from Claude (review)* — **severity: low (accuracy of the O(1) claim + three throughput notes)**

None of these block, but the acceptance criterion on #327 is phrased as *"Fleet-scale polling cost: O(1) LIST per tick regardless of shard count"*, and that is not what this is:

1. `_list_keys` re-lists the **whole** run prefix every tick and materializes every key into a set. At 5k shards that is ~5 LIST pages plus 5k path strings parsed per tick, forever — O(N/1000) requests and O(N) work, not O(1). Q3 acknowledges pagination "rides `obstore.list`'s internal pagination", which is true but is the thing that makes it not O(1). Worth restating in the docstring (line 570-572 makes the same O(1) claim) so nobody sizes a 50k-shard run off it. If it ever matters, `list_with_offset`/a lexicographic offset past the highest consumed key would trim the steady-state page count.
2. A shard awaiting re-dispatch keeps its **stale** object listed, so `_consume` re-GETs and re-JSON-parses it on every tick until the new attempt lands (`attempt_id in entry.consumed` → `return False`). Bounded by the failed-shard count, but it's a GET per failed shard per tick for the whole retry window.
3. `_tick` calls `_dispatch_up_to_window()` **before** it consumes, so a slot freed by a resolution is refilled only on the *following* tick — up to `_POLL_MAX_INTERVAL_S` of dead slot time, typically ~2 s since a consumption resets the interval. And `_dispatch_up_to_window` fires the whole batch **serially on the poller thread**, where v1 invoked from `workers` threads: a `max_in_flight`-wide opening burst is one thread's worth of sequential `invoke` round-trips (order 10 ms each ⇒ tens of seconds at a 1000-wide window), during which no LIST runs. Fine for 900 s shards; worth a line in the docstring so it isn't rediscovered as a mystery ramp.

src/zagg/runner.py
```diff
@@ -3371,6 +3416,47 @@ def _cell_work(payload):
...
+            def _fire(p=cell_payload):
+                state["lambda_client"].invoke(
+                    FunctionName=function_name, InvocationType="Event", Payload=p
+                )
+
+            fut = state["event_poller"].register(
+                int(shard_key),
+                label,
+                dispatch=_fire,
+                window=window["label"] if window is not None else None,
```
espg, line 3459: 🤖 *from Claude (review)* — **severity: low (test coverage on the one seam that can drift between the two halves)**

This is the **only** place the client half derives the window component of the status key, and the worker half derives it independently at `client_transport.py:159` (`window = (event.get("window") or {}).get("label")`). Those two expressions are the single spot where the channel's two halves can disagree on an object name — and a disagreement is silent: the client waits out the full drop window and reports `failed-unknown` for a shard that succeeded.

The suite deliberately closes this risk everywhere else (the stub worker calls the real `build_shard_status`, so key derivation can't drift for unwindowed units), but no test exercises a **windowed** event run: `test_shard_status_key_decimal_and_window` covers the helper, `test_windowed_unit_suffixes_the_window_label` covers the worker, and nothing joins them. Q7 flags "untested at fleet scale" — this part is testable *locally*: one `agg(invocation="event")` over a 2-window explicit schedule, asserting both units resolve.

Related coverage gap while you're here: `TestD8Audit` audits PUTs only on the `Run.dispatch` path. `Run.attach` is the one client path that can trigger a tail from a fresh process with no dispatch state, and its tests assert on `cell_events()`/`events == []` rather than on PUTs — the same `obstore.put` counter over `test_attach_mid_run_resolves_late_shards_and_runs_the_tail` would extend the D8 guarantee to the path most likely to grow a client-side write later.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the v2 Event transport. 8 inline findings below (2 medium-high, 2 medium, 4 low). Everything empirically checked against a read-only worktree of `142c4d7` (`uv sync --extra test`); `pytest tests/test_client_transport.py tests/test_client.py tests/test_lambda_handler.py tests/test_benchmark_objects.py` = 237 passed / 1 skipped, and `ruff check` / `ruff format --check` are clean on every file this PR touches (the two repo-wide offenders — `src/zagg/registry.py` N818, `tests/data/benchmark/README.md` — are pre-existing and outside CI's `--select=E,F,W,I`).

**What held up under attack** (recorded so it doesn't get re-litigated):

- **D8 is clean.** Hand-traced every obstore/boto call in `client_transport.py`: the write side is `write_shard_status` / `write_dispatch_manifest` / `write_tail_status`, all three reached only from `deployment/aws/lambda_handler.py` seams. The client side is `read_dispatch_manifest` (GET), `tail_recorded` (GET), `_list_keys` (LIST), `_consume` (GET) — reads only. `tail.json` is written by the **worker** (`mode="stats"`, off `tail_status_url`), never the client or the poller; the re-dispatch path only calls `client.invoke`. `TestD8Audit` patching `obstore.put` globally and asserting the exact PUT set is the right shape of test.
- **Prefix naming matches the ratified schema.** `run_status_prefix("s3://b/out.zarr", id)` → `s3://b/out.zarr.status/run-<id>` — i.e. the ratified `<store>.zarr.status/`, and a store *sibling*, consistent with #151's `.status/<run_id>/` (no collision: `run-` prefixed vs bare). `_is_status_object` correctly excludes `manifest.json` and `tail.json` as well as `shard-*.json`.
- **No other store walker sees the status prefix.** Checked every `list_with_delimiter` / `obstore.list` call site (`hive.py:190,210,392`, `coverage.py:248`, `sweep.py:881,935`, `telemetry.py:479`, `bench_objects.py:223`). The one that walks a *parent* of the store — the D19 multi-product discovery in `hive.classify_store_root` / `list_products` — is safe because `_is_valid_product_name("atl06.status")` is `False` (dots are outside the grammar); verified by running it.
- **The drop deadline keys off the right function.** `_get_function_timeout_s(client, run.function_name)` after `_resolve_function_name(config, function_name)`, and every worker in `template.yaml` shares `Timeout: !Ref Timeout`, so the base/4096-disk variants cannot diverge; the fallback `_DEFAULT_FUNCTION_TIMEOUT_S = 900` matches the template. `timeout + 60 + 90 = 1050 s` genuinely dominates the worst legal case (invoke queued 59 s + full 900 s run = 959 s), so a slow-but-alive shard is not falsely dropped.
- **Drop detection is not masked by a stale status object.** The `if entry.key in listed and self._consume(entry): ... elif <deadline>` shape still evaluates the `elif` when `_consume` returns `False`, so a re-dispatched attempt that is itself dropped still classifies — confirmed with a fake clock.
- **Re-dispatch double-writes are safe** (the Q the issue leaves implicit). Because the 1050 s deadline exceeds the worst legal in-flight case and `MaximumRetryAttempts: 0` bars service-side duplicates, the two executions are effectively *sequential*, and a sequential re-run is idempotent: hive rewrites the whole leaf (`overwrite=True` on the leaf template, `hive.py:1088`), flat rewrites the same absolute chunk keys with byte-identical content from the byte-identical event, and the sidecar / run-parquet row are last-writer-wins per shard. So the fail-open hole (a SUCCEEDED shard whose status PUT failed → re-dispatched) is *correct*, just double-billed — which is the substance of finding 2.
- **Payload parity.** `_invoke_lambda_cell` now delegates to `_build_cell_event`, so sync/async/event share one construction site and the key-presence rules cannot drift per-transport (the `aoi_payload` / `window` / `profile` / `handoff` / `invoked_by` variants are covered transitively by the existing event-shape tests). The differential test compares full event dicts modulo `run_id`, and the agg parity test additionally compares the tail's worker-invoke mode sequence and summary counts. The #335 merge is right: `runner._dispatch_run_stats` carries **both** `finalize_error=summary["finalize_error"]` and `tail_status_url` inside #335's new tail `try:` (the facade never passed `finalize_error` — that gap is pre-existing on `main`, not introduced here).
- **Fail-open is genuinely total on the worker side.** `_output_store_kwargs(event)` (which *can* raise `ValueError` on malformed `output_credentials`) is evaluated **inside** the wrapper's `try`, and the inner function has its own `try`; an import error, a creds failure, and a raising store are all covered by tests, including the ratified "raising store → shard result unaffected" one.

Module ceilings: `client.py` 1105 (Q5 — the split landed, residual is docstrings, fine by me); `client_transport.py` 857; `runner.py` is 5394 and this PR adds ~200 net — pre-existing and out of scope, but it is drifting further from §4 and is worth its own issue rather than another silent increment.

src/zagg/client_transport.py
espg: 🤖 *from Claude*

Folded in 009a0f6 — you were right that the comment described behavior the code didn't have, and the 0.3 ms probe is now impossible.

Throttles are fault class (c) of the three-class retry policy this PR now carries (`StatusPoller` docstring; the PR body flags the whole policy as espg's to veto). A faulted `invoke` re-arms the entry with a **not-before timestamp** rather than a sleep, exactly as suggested:

```python
delay = min(2.0**entry.attempts, _RETRY_BACKOFF_CAP_S)
with self._lock:
    entry.retry_not_before = self._clock() + delay
    entry.queued = True
```

and `_dispatch_up_to_window` filters on it, so the `while True` can no longer drain the budget in one pass:

```python
queued = [e for e in self._entries.values()
          if e.queued and not e.future.done() and e.retry_not_before <= now]
```

Two details worth naming: a parked entry is **not** counted against `max_in_flight` (it holds no slot), and `_run` clamps its wait to the soonest not-before (`_next_retry_wait`) so the idle-tick doubling can't add up to `_POLL_MAX_INTERVAL_S` to every re-fire. No jitter — the poller re-fires serially from one thread, so a synchronized throttle can't produce a simultaneous burst; noted in the constant's comment.

On your closing point that the old test didn't pin the property: `test_throttled_invoke_backs_off_instead_of_bursting` now drives a fake clock and asserts the exact re-fire instants (`[1000.0, 1002.0, 1006.0, 1014.0]` — 2 s, 4 s, 8 s), and `test_invoke_fault_burns_an_attempt_and_retries` was moved onto the same fake clock so it can't pass by accident.

src/zagg/client_transport.py
espg: 🤖 *from Claude*

Folded in 862e23a, taking your option (1) for the status case and your option (2) for the drop case — the policy is now explicit and split by fault class, and the PR body records it under "Retry policy" as espg's to veto.

**(a) A `failed` status object never re-fires.** The object exists ⇒ the worker ran to completion and reported its own error ⇒ deterministic for this shard, which is the case issue #119 already excludes on the sync transport. `_consume` routes it to `_worker_failed`, which resolves terminally. One execution billed, not `max_retries` of them.

**(b) A drop re-fires exactly once.** `_dropped` → `_drop_failed`, bounded by `_MAX_DROP_REFIRES = 1` (and by `max_retries <= 1`, which opts out entirely so the knob keeps its "no retries" meaning). Worst case is **2 × function timeout**, which is the bound I documented rather than left implicit:

> issue #151 pinned `MaximumRetryAttempts: 0` precisely so async fan-out is never re-billed at 900 s per service retry, and an unbounded client-side re-dispatch would reintroduce it.

I did not take the `started`-marker half of option (2). It would make drop-vs-death distinguishable, but it costs a second PUT per shard on the hot path and the bound above already caps the exposure at 2×; worth its own issue if you want the distinction. **Left standing for you.**

Tests: `test_failed_status_raises_shard_error_without_redispatch` (1 invoke, `retries == 0` at `max_retries=3`), `test_drop_refires_exactly_once_then_resolves` (`max_retries=5`, still exactly 2 executions), `test_drop_refire_recovers_the_shard`, and `test_dropped_invoke_refires_once_and_recovers` end-to-end through the stub's new `drop_times`.

src/zagg/client.py

Folded in a8dd7b1 — agreed, and the fix is the one you named: thread the prefix through instead of hardcoding `None`.

`_post_run` / `_run_tail` now take `result_prefix`; `dispatch()` sets it to `run_status_prefix(self.store, run_id)` when `transport == "event"` and leaves it `None` on sync (which genuinely has nothing to point at), and `attach_run` passes the same prefix it is already reading from. So the facade and `runner.py:3891` now make the same choice for the same reason.

The stale comment went with it — it now says the pointer is the status prefix on an event run and `None` on sync, rather than "no prefix on either transport".

Test: `test_oversized_rows_point_at_the_status_prefix` drops `_RUN_STATS_INLINE_CAP_BYTES` to 10 and asserts `rows_from == run_status_prefix(...)` with `rows == []`. That needed envelope-backed rows to be meaningful (otherwise `inline_rows` is the whole set and the record skips on the *other* branch), so the stub gained a `record=True` mode that rides a real `telemetry.build_record` — the deployed-worker shape, with the default staying the stale-worker shape the parity tests rely on.

src/zagg/client_transport.py
espg: 🤖 *from Claude*

Folded in 7ce6c5a — took the cheap fix you described.

`register` now refuses a post-shutdown registration and resolves it through the caller's own failure policy (`_resolve_failure`, so the client raises `ShardError` and agg records a result dict) with `outcome: "not-dispatched"`, instead of handing back a Future nothing owns. The check and the insert share the entry lock, and `shutdown` sets `_stop` under that same lock, so a registration can't interleave mid-insert:

```python
with self._lock:
    stopped = self._stop.is_set()
    if not stopped:
        self._entries[entry.key] = entry
```

Your point about the undocumented ordering is now written down where it matters — `register`'s docstring names the asymmetry (registration deliberately trails `start()`, so the shutdown side needed the matching guard) and records that `_run_lambda`'s `finally` draining the executor before `event_poller.shutdown()` is what kept it unreachable.

Two tests: `test_register_after_shutdown_resolves_instead_of_hanging` (resolves, and the entry isn't tracked, so a late status object can't revive it) and `test_register_after_shutdown_honors_the_failure_policy` (the client's `on_failed` applies to the refusal, so it surfaces as a failed shard rather than a silent ok).

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Folded in 3818c81 — took your first option (don't write the marker), for the reason you gave: attach re-running an idempotent tail is the safe direction, and it needs no new marker semantics.

`write_tail_status` now returns early when the stats event carries a `finalize_error`, and both docstrings (`TAIL_NAME` and the function) say the marker means "the tail's record leg completed **and** finalize succeeded". The run parquet still gets `finalize_error` as its run-level column, so nothing is lost for the postmortem.

One thing the finding surfaced that wasn't in it: on the **facade** path the fix would have been inert, because `Run._run_tail` never passed `finalize_error` to `_dispatch_run_stats` at all (only `runner.py`'s tail did). So the guard would have keyed off a field that was never on the wire from `Run.dispatch`. That's now threaded too — same always-present/`None`-on-success contract as runner's tail, which also closes an issue #335 parity gap between the CLI and the facade.

Tests: `test_finalize_error_withholds_the_marker` in `tests/test_lambda_handler.py` (parquet written, marker absent; then `finalize_error: None` writes it), and `test_attach_reruns_the_tail_after_a_failed_finalize` end-to-end — an event run whose finalize invoke fails leaves no `tail.json`, and the reattached handle fires `finalize` again.

src/zagg/client_transport.py
espg: 🤖 *from Claude*

Folded both in 5f448d5.

**(b) the dedupe poisoning** — fixed as you described: a falsy `attempt_id` is treated as always-new (and logged) rather than added to `consumed`, so it can't swallow every subsequent object. This was the half worth the commit on its own; a shard that succeeded being reported `failed-unknown` is the worst class of bug in this channel.

**(a) `schema_version`** — now read back, but I warned rather than resolved `failed`. The reasoning: a status object means the shard's data already landed, so failing it on a telemetry-schema mismatch would trade a silent-wrong for a loud-wrong. `telemetry.sidecar_key` can raise because an unrecognized spec means it cannot compute a key at all; here the object is very likely a superset of v1 and reads correctly. So the bump is loud (the log names both versions) without being fatal. If you'd rather it be terminal, say so and I'll flip it — noting the difference here so it's a recorded choice rather than an omission.

Tests: `test_unknown_schema_version_warns_and_still_resolves` (v99 → warning in `caplog`, future still resolves 200) and `test_missing_attempt_id_does_not_poison_the_dedupe` (`consumed` stays empty; a second id-less object is consumed again instead of ignored).

src/zagg/client_transport.py
espg: 🤖 *from Claude*

Folded in 7876f4b. Points (1) and (2) as documentation, point (3) partly as code.

**(1) the O(1) claim.** Corrected rather than defended. The `StatusPoller` docstring now has a "What 'one LIST per tick' actually costs" paragraph saying it is one LIST *call site* — O(N/1000) requests and O(N) key parsing, ~5 pages and 5k parsed paths per tick at 5k shards — quotes the acceptance criterion it contradicts, and names `list_with_offset` past the highest consumed key as the fix if it ever matters. The module docstring's version points at it too, so neither reads as O(1) anymore.

**(2) the stale-object re-GET.** Documented in the same paragraph. Worth noting it got much smaller in passing: with class (a) resolving immediately (#3679433912), the only shards holding a stale listed object are drop re-fires, which are capped at one.

**(3) dead slot time** — fixed in code, since it was cheap. `_tick` now resolves before dispatching:

```python
progressed = self._resolve_landed()
return self._dispatch_up_to_window() or progressed
```

so a slot freed by a resolution refills in the *same* tick. Harmless on the opening tick — every entry is still queued, so `_resolve_landed` finds no pending entry and skips the LIST entirely. `test_a_freed_slot_refills_in_the_same_tick` drives `_tick()` by hand (no threads) and asserts shard 2 fires in the tick that resolves shard 1.

The serial-dispatch-on-the-poller-thread ramp is documented, not changed — an opening burst of one thread's sequential invokes is a real ramp but bounded and fine for 900 s shards. **Left standing for you** if you want the dispatch batch moved onto a small pool.

src/zagg/runner.py
espg: 🤖 *from Claude*

Folded both coverage gaps in de583d8.

**The windowed seam.** `test_windowed_units_resolve_on_both_windows` runs one `agg(invocation="event")` over a 2-window explicit schedule (3 shards × 2 windows = 6 units), asserting all 6 resolve with `cells_error == 0` and — the part that actually joins the two halves — that the objects in the store are exactly:

```python
sorted(ct.shard_status_key(w, window=lbl) for w in _WORDS for lbl in ("w1", "w2"))
```

Since the stub writes through the real `build_shard_status` (worker-side derivation) and the runner registers with `window["label"]` (client-side derivation), a disagreement between `runner.py:3456` and `client_transport.py:159` now fails this test instead of surfacing as a `failed-unknown` for a shard that succeeded. `_agg` took an optional `config` to make this possible.

**The `TestD8Audit` gap.** `test_attach_makes_no_store_writes` puts the same `obstore.put` counter over the attach path with mid-run state (manifest + 3 statuses, no tail marker) so the whole tail actually runs, and asserts `puts == [ct.TAIL_NAME]` — the worker's marker riding the stats invoke, nothing else. Agreed this is the path most likely to grow a client-side write later.

src/zagg/client_transport.py
espg: 🤖 *from Claude*

Follow-up on the same finding: three docstrings still described the old single-budget policy after 862e23a (`runner.agg`'s `invocation` docs, the `_cell_work` comment, and `Run.dispatch`'s `transport` docs). Corrected in 6635f54 so the public API prose matches the code — no behavior change.

Issue #341, opened 2026-07-30
Benchmark run failures: schema-narrowing orphans crash the store walker; strata config group reads raise on Lambda (errors swallowed)


espg:
🤖 *from Claude*

Two defects surfaced by benchmark run 30503334617 (the first post-#337 ladder dispatch) — filed together per espg so one PR can fix both in a pass. The stale stores have been cleared operationally (espg-approved S3 delete), so the per-merge leg is unblocked; these are the underlying bugs.

## Bug A — schema-narrowing leaves orphan objects that crash the store walker

**Symptom**: after PR #337 removed the dead `cell_ids` coordinate from the benchmark configs, every dispatch against the pre-existing stores died worker-side with `Unhandled exception: No array found in store ... at path 19/cell_ids` (sidecar + located arms; the per-merge leg was red from #337's merge until the stores were cleared).

**Mechanism**: `run_benchmark` dispatches with `overwrite=True`, but overwrite only rewrites members the *new* template knows about — the old schema's `19/cell_ids/c/0` chunk object survives as an orphan. Something walking the leaf (store-walk during write or rollup) finds the orphan chunk, infers an array at `19/cell_ids`, and `zarr.open` fails on the missing `zarr.json`.

**Two-layer fix candidate**: (1) the walker should skip/inventory unknown array dirs instead of dying (resilience); (2) `overwrite=True` on a hive leaf should clear the leaf prefix (or at minimum diff old-vs-new members and delete retired ones) so schema narrowing is clean. Also worth answering: the **semantic-hash guard did not refuse the stale root** — the aggregation block changed, so `validate_manifest` should have rejected reuse before any worker ran; check whether `overwrite=True` bypasses the guard and whether that bypass should still hard-delete-then-write rather than write-over.

## Bug B — strata config: every overlapping group read raises on Lambda, errors swallowed

**Symptom**: the strata arm (fresh store, no orphans) ran 66.7 s and returned `"No data after filtering (121 group reads raised)"` — `read_errors: 121`, `obs=0`, only the setup manifest written. Status object: `s3://…strata.zarr.status/f64ed316…/4331422412.json`.

[...]

Issue #330, opened 2026-07-29
Convert docstrings to numpydoc standard (repo-wide sweep)


espg:
🤖 *from Claude*

Replicates for zagg what espg/mortie just did for its docstrings — the mortie trio is espg/mortie#135 (conversion sweep), espg/mortie#140 (mechanical enforcement), espg/mortie#133 (docs workflow, where espg ratified "numpy / numpydoc are the gold standard"). This issue is the zagg counterpart of espg/mortie#135. Filed at espg's request per session 2026-07-29; **deliberately left unlabelled** — per CLAUDE.md §5 an unlabelled issue is out of scope for the routine until a human applies `plan`/`implement`.

Companion issues: #331 "Enforce numpydoc validation in CI" (counterpart of espg/mortie#140, sequenced after this sweep) and #332 "Versioned documentation" (`discuss`, counterpart of the mike phase of espg/mortie#133).

## Why

zagg's published API reference already renders with **mkdocstrings configured for `docstring_style: numpy`** (`mkdocs.yml`, `plugins.mkdocstrings.handlers.python.options`). Docstrings that follow numpydoc render as structured parameter/return tables; freeform prose renders as an undifferentiated block. zagg today is mixed — some modules (e.g. `src/zagg/stats/tdigest.py`, `src/zagg/config.py`) already carry `Parameters`/`Returns` sections, most don't — so the API pages under `docs/api/` are visibly inconsistent.

CLAUDE.md §4 names docstrings as one of the three documentation forms but names no standard. As with mortie's CLAUDE.md §7 revision, the sweep should end with zagg's CLAUDE.md updated to say **numpydoc** explicitly, so the convention and the (future) linter agree.

## Current state (measured 2026-07-29)

AST census over `src/zagg/**/*.py`: counting `def`/`class` statements, those with any docstring, and those whose docstring contains a numpydoc section underline (`----------`).

**Totals: 900 defs+classes, 718 with docstrings, only 119 with numpydoc sections (~13% converted).** For calibration, running ruff's pydocstyle with `convention = "numpy"` (not currently enabled — zagg selects `E,F,W,I,N` only) reports just **157 `D` violations** — the same ruff-vs-numpydoc divergence measured in espg/mortie#135: well-formed prose satisfies ruff while being ~0% converted, so ruff cannot be the progress metric here either.

Largest-gap modules (defs / numpydoc blocks):

| module | defs+classes | with docstring | numpydoc blocks |
|---|---|---|---|
[...]

PR #344, opened 2026-07-30
Overview-zarr sweep family: pyramid generation at ancestor nodes (issue #201)

### LINKED ISSUE #201: Pyramid/overview generation sweep (second pass): overview zarrs, MOC regen, optional interop materialization
🤖 *from Claude*

Design record: [`docs/design/sparse_coverage.md`](https://github.com/englacial/zagg/blob/main/docs/design/sparse_coverage.md) §7 (D11/D12), tracked on #198. `discuss` — the write-time layout (#199) is settled; this second-pass process has genuinely open design questions.

## Settled framing (from the design discussion)

- Overview generation is a **separate post-process sweep**, never write-time — it aggregates across worker-shard boundaries, so shard workers can't produce it anyway.
- Overview zarrs live at **ancestor nodes** of the hive tree, explicitly marked (`role: overview` + source order + aggregation method in attrs). Never inferred from position: a shallow zarr may equally be *coarse source* in a sparse region.
- Full pyramid storage cost is geometric ~1/3 extra (4 children per order).
- The sweep is idempotent and non-load-bearing: it can fail or lag without corrupting anything. It also owns MOC regeneration and any optional derived interop artifacts (e.g. a `zarr.open`-able root hierarchy or consolidated index — the D12 escape hatch, generated as a *cache*, never required).

## To discuss

1. **Aggregation semantics per field**: mean/count compose trivially up-tree; t-digests merge natively; which fields degrade (e.g. located/ragged CSR fields — summarize, subsample, or omit at overview orders?).
2. **Level schedule**: every order up to base? Sparse levels (e.g. every 2–3 orders)? Declared in `morton_hive.json`'s pyramid section — who writes that declaration, template time or sweep time?
3. **Orchestration**: same Lambda fan-out machinery over ancestor cells (bottom-up levels, each level reading the one below), or a coarser local/batch job? Reuses the leaves-to-root construction argument from `docs/design/architecture.md`.
4. **Trigger**: manual CLI, end-of-run hook from the dispatcher, or scheduled?
5. **Walker/read interplay**: display readers may stop at an overview; analysis readers must not conflate it with source — is the `role` attr enough, or does the manifest also enumerate overview orders for zero-open filtering?

espg:
Closes #201

Implements the **overview-zarr family** of the unified second-pass sweep — the fourth D22 derived-artifact family, riding the #300 engine (d104dbb) — against the amended spec (PR #306: D16–D24 + O-resolutions) and the [implementation plan posted on the thread](https://github.com/englacial/zagg/issues/201#issuecomment-5025396182), with the on-thread rulings applied: all-time overviews as `all.zarr` (confirmed), display schedule every 2 orders, and the [ragged roster-field ruling](https://github.com/englacial/zagg/issues/201#issuecomment-5025509889) (option A per-field exclusion as the default, option B declared derived summary as the opt-in, [full option space](https://github.com/englacial/zagg/issues/201#issuecomment-5025519604)).

## What it does

For each **manifest-declared overview order k**, the sweep folds committed source leaves into an overview zarr at the ancestor digit node, at cell order `cell_order − (shard_order − k)` — constant depth, cells coarsen 4× per order of ascent, so the pyramid is the store's resolution axis, partially materialized (D24). Field inclusion is gated by the new **D24 composability-class helpers** (`zagg.semantics.field_composability`, derived from the aggregator merge-law set):

- **exact** (`len/count/sum/nansum/min/nanmin/max/nanmax`, `np.`-prefix normalized) — folds by merge law, byte-equal to direct aggregation at the coarser order;
- **approximate** (unlocated ragged t-digest, `(2,)` centroids) — the order-independent k-way merge (`merge_tdigests_kway`, issue #279), `np.isclose` class;
- **none** (everything else) — **excluded**: exists only at native resolution, absence declared in the pyramid block, loud warning at template time (the ruled D24 default).

Discovery is run-record + root `coverage.moc` — never a LIST; `moc` precedes `overview` in `DEFAULT_FAMILIES` so the root MOC is fresh in the same pass. Idempotence is the D22 generation stamp (merged-leaf count + max leaf stamp timestamp) **plus a content hash** over the folded decoded values (O11-style) — the same-second backstop the JSON families get from the engine's payload compare. Overviews are regenerable caches (D9): deleting every one leaves leaf reads green, and regeneration is byte-stable (the digest merge is permutation-independent).

## Store shape emitted (the moczarr#15 8b seam)

```
{product_root}/
  morton_hive.json                 pyramid block: template-declared schedule + per-field
                                   D24 classes/methods (incl. declared ABSENCE for
                                   class-none fields); sweep-populated `materialized`
  {base}/{d1}/.../{dk}/            ancestor node at a declared overview order k
    {window}.zarr/                 per-window overview (D23: overviews inherit window
                                   naming); `all.zarr` = unwindowed / all-time fold
      zarr.json                    root attrs: `role: "overview"` (never inferred from
                                   position, D11) + `zagg_overview` {spec, node, order,
                                   cell_order, source_shard_order, source_cell_order,
                                   window, per-field {class, method}, generation,
                                   content_hash, generated_at} + the D4 commit stamp
                                   (the FINAL write; window label + time-range union
                                   ride it on windowed stores)
      {cell_order}/                inner group, same structure as a source leaf
                                   (emitted through HealpixGrid.emit_shard_template so
                                   dtypes/fills/D18 ragged attrs/D16 dggs attrs cannot
                                   drift):
        morton                     uint64, full subtree words at the overview cell order
        {exact fields...}          dense arrays, merge-law folded
        {tdigest fields...}        vlen-bytes ragged, k-way merged
    overview.rollup.json           sweep bookkeeping (same closed name grammar as the
                                   other families' rollup objects): window inventory +
                                   per-window generation + content hash
```

Source leaves carry **no** `role` attr (absence = source); analysis readers reject on `role: "overview"`, display readers stop at it.

## Phases (mirroring the thread plan)

- [x] **Phase A — fold kernels + class gating**: `zagg.semantics.{EXACT_MERGE_LAWS, field_composability, composability_classes}`; `zagg.sweep_overview.{fold_dense, combine_dense, fold_digests, encode_digest, decode_digest}` (`phase 1`)
- [x] **Phase B — overview writer**: `sweep_overviews` + `OverviewFamily` (the engine's new whole-tree `sweep_store` hook), D23 naming, D11 attrs, D22 generation stamps, per-node envelope; `DEFAULT_FAMILIES` gains `overview` (`phase 2`)
- [x] **Phase C — manifest pyramid block**: `build_pyramid_block` wired into `build_manifest` (template-time declaration, default every-2 schedule), the `output.pyramid` knob + validation (`get_pyramid`/`_validate_pyramid`, incl. the option-B `summarize` declaration grammar), sweep-populated `materialized` actuals (`phase 3`)
- [x] **Phase D — §8.3 obligations**: rollup == direct at every declared order (byte-equal exact / `np.isclose` approximate), sweep idempotence, same-second staleness caught by the content-hash backstop, nothing-load-bearing + byte-stable regeneration, role-never-inferred (`phase 4`)
- [ ] **Phase E — orchestration scale-out**: deferred by the plan itself ("local pass first; Lambda fan-out per level only if measured wall-time demands it"). The family already reaches Lambda through the existing `mode="sweep"` worker invoke; per-level fan-out is unbuilt.
- [ ] **Phase F — seeded-reservoir overview fields (opt-in, espg-flagged)**: deferred. It gates on the option-B declaration grammar, which IS landed here (`output.pyramid.summarize` → pyramid block), but both B-generation and C-sampling act only on roster-kind ragged fields — and every ragged field shipped today is digest-kind (the ruling is explicitly forward-looking). The writer warn-skips declared summaries with a pointer to issue #265.

## How it was tested

- `tests/test_sweep_overview.py` — 63 tests: composability classes (shipped configs + synthetic metas), kernel laws (fill/NaN edges, int/float dtypes, two-hop composability, digest permutation stability), pyramid-block declaration + config grammar, and the writer end-to-end on real leaf zarrs (per declared order, windowed + all-time, coarser-than-shard accumulation, sibling contribution via the root MOC, debris invisibility, envelope inventory), plus the §8.3 obligations above.
- The issue #341 Bug A landmine: the fold opens leaf arrays **by name** — no member enumeration and no LIST anywhere on the overview path — so orphan array prefixes / foreign objects inside a leaf cannot crash it (tested with a junk array prefix + stray status-like object); a leaf missing a declared field contributes fill (schema evolution, tested); an unreadable or mixed-order leaf skips loudly (`failed` count), never fatally (tested). The no-LIST property also keeps the #327 `.zarr.status/` prefixes structurally out of reach.
- `tests/test_sweep.py` / `tests/test_hive.py` updated for the new default-family set and the manifest block.
- **`.github/scripts/bench_objects.py`** (the #240 object-count model, NOT workflow config — flagged here because §1 covers `.github/`): with `overview` in the default families, local runs now land overview zarrs in the store, which the model's hive walk pooled into `objects_other` and failed on. They now classify into their own `objects_overviews` bucket — the exact treatment the `*.rollup.json` bucket got in #300 (second-pass D9 caches, excluded from the audited write-path total; the #215 per-shard bypass guard is untouched, and a stray *id*-named zarr outside the dispatched leaf set still surfaces as a loud `other` finding). Classification unit test added.
- Full local suite: `uv run --extra test --extra catalog pytest` — 2900+ passed; the only failure is the known pre-existing `TestFunctionBuild::test_function_build_succeeds` (TestConsolidationGate did not reproduce locally).
- `ruff check` / `ruff format --check` clean on every touched file; the new code is mypy-clean (pre-existing findings in untouched files left alone, per §4).

## Questions for review

1. **RESOLVED (espg ruling, 2026-07-30): `overview` stays in `DEFAULT_FAMILIES`.** The condition was a config-level disable, which exists: `output.pyramid: false` in the yaml declares the overview family off per-store (the manifest pyramid block gets `orders: []` and the sweep no-ops, `declared: false`). Rather than switching the coded default, the fleet-scale operational configs opt out via that knob until Phase E sizing (b84f31b: the three `*_hive*` benchmark arms, `s2_neon_o9`, and the shipped `atl03_tdigest_strata_healpix`); local/small test configs keep the default.
2. **All-time fold defaults OFF** on windowed stores (`output.pyramid.all_time: true` opts in; a pre-existing `all.zarr` entry keeps being maintained either way). D13 recorded the all-time overview as "optional"; flag if it should default on. **This is now the only open question on the PR** — everything else below is resolved or informational.
3. **RESOLVED (spec-ratified): overview cell order = constant depth** (`cell_order − (shard_order − k)`): the arithmetic is now frozen in the PR #346 store spec §4.4 (implementation conforms to spec).
4. **Mixed-order source leaves** are skipped with a loud warning. The remaining gate is zagg-side and tracked in #347 (sweep fold semantics + the writer-side append guard) — the reader-side prerequisites are done (mortie#116 merged in 0.9.1; moczarr#8 closed via its PRs 22/25) and the skip's error text now points at #347 (553ca04, espg-confirmed). No writer emits heterogeneous cell orders yet; the overview writer itself is what makes the tree multi-order.
5. **D24 enforcement beyond this issue** is follow-up, not this PR: the mixed-order fold semantics + writer-side append guard are tracked in #347; recording the class into the semantic core rides with that work.
7. **RESOLVED (espg ruling, 2026-07-30): exact-fold NaN semantics** ([finding-3 thread](https://github.com/englacial/zagg/pull/344#discussion_r3680582363)): nan-skipping is the intended behavior — the `nan_policy: "skip"` declaration (pyramid block + overview attrs, `d211fff`) stands, plus a config-validation warning when a plain `min`/`max`/`sum` pairs a NaN fill (`818a788`; warning not error — the shipped atl06 config trips it by design and stays valid).
6. **Pre-existing local failures, not touched** (§4): ruff N818 on `src/zagg/registry.py` (`UnknownCapability`), codespell on `src/zagg/config.py:1095` ("unparseable"), and the known `TestFunctionBuild` failure.

Module sizes: the heavy lifting lives in the new `src/zagg/sweep_overview.py` (~830 lines); `src/zagg/sweep.py` grew ~18 lines to ~1034 (the family class + the `sweep_store` dispatch hook) — it was already at the ~1000 ceiling before this PR, and the split is issue #330's pending plan, deliberately not done here.

https://claude.ai/code/session_01HW3smGgEzF9LVWhXcbNv5s

### DIFF
--- a/src/zagg/semantics.py
+++ b/src/zagg/semantics.py
@@ -80,6 +80,15 @@ GRID_SPATIAL_KEYS = ("crs", "resolution", "bounds")
 #: direct aggregation at the coarser order, byte for byte (§8.3). Names are
 #: matched after :func:`_fold_function_name` normalization, so ``min``,
 #: ``np.min`` and ``numpy.min`` all key the same law.
+#:
+#: The plain and nan-aware variants share a law ON PURPOSE, and the exactness
+#: claim is against the **nan-skipping** reduction for both (review finding,
+#: issue #201): a leaf's stored NaN is the same bytes whether it is the fill
+#: sentinel or a NaN datum, so nothing downstream can tell them apart —
+#: :func:`zagg.sweep_overview.fold_dense` skips both, and the overview records
+#: :data:`zagg.sweep_overview.EXACT_NAN_POLICY` per field so a reader knows
+#: which reduction it actually got. A NaN-propagating ``min``/``max``/``sum``
+#: is unrecoverable at this seam, not merely unimplemented.
 EXACT_MERGE_LAWS = {
     "len": "sum",
     "count": "sum",
--- a/src/zagg/semantics.py
+++ b/src/zagg/semantics.py
@@ -74,6 +74,90 @@ FINGERPRINT_HEX = 12
 GRID_SPATIAL_KEYS = ("crs", "resolution", "bounds")
 
 
+#: D24 exact merge laws: aggregator name -> fold law. These are the reducers
+#: whose per-cell outputs compose EXACTLY across cell orders — count/sum by
+#: addition, min/max by extremum — so an up-tree fold of leaf values equals a
+#: direct aggregation at the coarser order, byte for byte (§8.3). Names are
+#: matched after :func:`_fold_function_name` normalization, so ``min``,
+#: ``np.min`` and ``numpy.min`` all key the same law.
+EXACT_MERGE_LAWS = {
+    "len": "sum",
+    "count": "sum",
+    "sum": "sum",
+    "nansum": "sum",
+    "min": "min",
+    "nanmin": "min",
+    "max": "max",
+    "nanmax": "max",
+}
+
+#: The three D24 composability classes, weakest first. ``exact`` folds byte-
+#: equal; ``approximate`` merges natively but order-dependently (t-digests —
+#: ``np.isclose`` equality, the merge-vs-spill epistemic class); ``none`` has
+#: no merge law and exists only at native resolution (per-field exclusion —
+#: the ruled D24 default, issue #201).
+COMPOSABILITY_CLASSES = ("none", "approximate", "exact")
+
+
+def _fold_function_name(name) -> str | None:
+    """Normalize an aggregator ``function`` name for merge-law lookup.
+
+    ``np.``/``numpy.`` prefixes strip (they resolve to the same callables in
+    :func:`zagg.config.resolve_function`); other dotted paths pass through
+    unchanged (their laws are keyed by full path, e.g. the t-digest builders).
+    """
+    if not isinstance(name, str):
+        return None
+    for prefix in ("np.", "numpy."):
+        if name.startswith(prefix):
+            return name.removeprefix(prefix)
+    return name
+
+
+def field_composability(meta: dict) -> str:
+    """The D24 composability class of one aggregation field's metadata.
+
+    Derived from the field's aggregator via the merge-law flags the #217
+    mergeable-reducer machinery already established (the same set
+    ``validate_streaming`` accepts, widened by the exact sum/min/max laws):
+
+    - ``exact`` — scalar ``function`` in :data:`EXACT_MERGE_LAWS`;
+    - ``approximate`` — an unlocated ragged t-digest field with the standard
+      ``(2,)`` centroid inner shape (merge is order-dependent; ``np.isclose``
+      equality class, cf. D24);
+    - ``none`` — everything else: expressions, vector fields, chunk-resolution
+      companions, located ragged (no streaming merge law for the location
+      channel yet), and any scalar reducer without an exact law (mean, std,
+      median, quantiles, ...).
+    """
+    from zagg.config import get_output_signature
+
+    sig = get_output_signature(meta)
+    if meta.get("expression") is not None or sig["resolution"] != "cell":
+        return "none"
+    function = _fold_function_name(meta.get("function"))
+    if sig["kind"] == "ragged":
+        from zagg.processing.streaming import _TDIGEST_FUNCTIONS
+
+        if (
+            sig["location"] is None
+            and meta.get("function") in _TDIGEST_FUNCTIONS
+            and tuple(sig["inner_shape"]) == (2,)
+        ):
+            return "approximate"
+        return "none"
+    if sig["kind"] == "scalar" and function in EXACT_MERGE_LAWS:
+        return "exact"
+    return "none"
+
+
+def composability_classes(config: PipelineConfig) -> dict[str, str]:
+    """``{field_name: composability class}`` for every aggregation field (D24)."""
+    from zagg.config import get_agg_fields
+
+    return {name: field_composability(meta) for name, meta in get_agg_fields(config).items()}
+
+
 def _without(mapping: dict, keys: tuple[str, ...]) -> dict:
     return {k: v for k, v in (mapping or {}).items() if k not in keys}
 
@@ -160,10 +244,14 @@ def semantic_fingerprint(digest: str) -> str:
 
 __all__ = [
     "AGGREGATION_PACKAGING_KEYS",
+    "COMPOSABILITY_CLASSES",
     "DATA_SOURCE_PACKAGING_KEYS",
+    "EXACT_MERGE_LAWS",
     "FINGERPRINT_HEX",
     "GRID_SPATIAL_KEYS",
     "canonical_semantic_json",
+    "composability_classes",
+    "field_composability",
     "semantic_core",
     "semantic_fingerprint",
     "semantic_hash",
--- a/src/zagg/sweep_overview.py
+++ b/src/zagg/sweep_overview.py
@@ -173,6 +173,116 @@ def fold_digests(cell_digests: list, *, delta: int, dtype="float32") -> bytes:
 
 
 # ---------------------------------------------------------------------------
+# The manifest pyramid block: template-time declaration (D22, per-family
+# schedules), sweep-populated actuals.
+# ---------------------------------------------------------------------------
+
+
+def build_pyramid_block(config, shard_order: int) -> dict:
+    """The manifest ``pyramid`` block: the template-time declaration (D22).
+
+    Declares the overview family's order schedule (``output.pyramid`` knob;
+    default every :data:`DEFAULT_SPACING` orders below the shard order —
+    espg-ratified) and each field's composability class + fold method (D24).
+    ``none`` fields are declared with their class ONLY — the recorded absence
+    that gives readers zero-open filtering (option A, the ruled default) —
+    and warned about loudly at template time, per the D24 ruling. The sweep
+    later adds ``materialized`` actuals; it never rewrites the declaration.
+    """
+    from zagg.config import get_agg_fields, get_pyramid
+    from zagg.semantics import EXACT_MERGE_LAWS, _fold_function_name, composability_classes
+
+    knob = get_pyramid(config)
+    if knob is None:  # output.pyramid: false — declared off
+        return {"spec": PYRAMID_SPEC, "overview": {"orders": []}}
+    spacing = int(knob.get("spacing") or DEFAULT_SPACING)
+    if knob.get("orders") is not None:
+        orders = sorted({int(k) for k in knob["orders"]}, reverse=True)
+    else:
+        orders = list(range(int(shard_order) - spacing, -1, -spacing))
+    agg = get_agg_fields(config)
+    fields: dict = {}
+    excluded = []
+    for name, cls in composability_classes(config).items():
+        meta = agg[name]
+        if cls == "exact":
+            fields[name] = {
+                "class": "exact",
+                "method": EXACT_MERGE_LAWS[_fold_function_name(meta.get("function"))],
+                "dtype": meta.get("dtype", "float32"),
+                "fill_value": _json_fill(meta.get("fill_value", "NaN")),
+            }
+        elif cls == "approximate":
+            inner = meta.get("inner_shape") or (2,)
+            fields[name] = {
+                "class": "approximate",
+                "method": TDIGEST_LAW,
+                "dtype": meta.get("dtype", "float32"),
+                "inner_shape": [int(inner)] if isinstance(inner, int) else [int(x) for x in inner],
+                "delta": int((meta.get("params") or {}).get("delta", 512)),
+            }
+        else:
+            fields[name] = {"class": "none"}
+            excluded.append(name)
+    if excluded and orders:
+        logger.warning(
+            f"pyramid: fields {excluded} are non-composable (D24 class 'none') and will "
+            f"exist ONLY at native resolution — excluded from every overview order (the "
+            f"per-field-exclusion default; declare a derived summary to opt in, issue #201)"
+        )
+    overview: dict = {
+        "spacing": spacing,
+        "orders": orders,
+        "all_time": bool(knob.get("all_time", False)),
+        "fields": fields,
+    }
+    if knob.get("summarize"):
+        overview["summarize"] = {str(k): dict(v) for k, v in knob["summarize"].items()}
+    return {"spec": PYRAMID_SPEC, "overview": overview}
+
+
+def _json_fill(fill_value):
+    """A JSON-safe fill token (zarr v3 uses the string ``"NaN"``)."""
+    if isinstance(fill_value, float) and np.isnan(fill_value):
+        return "NaN"
+    return fill_value
+
+
+def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
+    """Record materialized overview orders in the manifest pyramid block.
+
+    The one manifest key the sweep may touch (D11: the block is populated/
+    updated by the §7 sweep; it is excluded from the frozen resume keys, so
+    this RMW can never brick appends). Fail-open — the declaration readers
+    key on is untouched; ``materialized`` is a convenience actual.
+    """
+    import obstore
+
+    from zagg.hive import MANIFEST_NAME, _utcnow, read_manifest
+    from zagg.store import open_object_store
+
+    try:
+        fresh = read_manifest(store_root, **store_kwargs)
+        if fresh is None:
+            return False
+        block = fresh.setdefault("pyramid", {}).setdefault("overview", {})
+        known = set((block.get("materialized") or {}).get("orders") or [])
+        block["materialized"] = {
+            "orders": sorted(known | {int(k) for k in orders}),
+            "generated_at": _utcnow(),
+        }
+        obstore.put(
+            open_object_store(store_root, **store_kwargs),
+            MANIFEST_NAME,
+            json.dumps(fresh, indent=1).encode(),
+        )
+        return True
+    except Exception as e:
+        logger.warning(f"sweep[overview]: manifest pyramid update failed (fail-open): {e}")
+        return False
+
+
+# ---------------------------------------------------------------------------
 # Phase B: the overview writer — the family's whole-tree sweep.
 # ---------------------------------------------------------------------------
 
@@ -236,6 +346,7 @@ def sweep_overviews(store_root: str, manifest: dict, by_shard: dict, *, store_kw
         orders = [k for k in orders if k not in bad]
     candidates = _candidate_decimals(store_root, shard_order, by_shard, store_kwargs)
     store = open_object_store(store_root, **store_kwargs)
+    materialized: set[int] = set()
     for k in orders:
         nodes = sorted({_node_at(d, k) for d in by_shard})
         for node in nodes:
@@ -264,6 +375,8 @@ def sweep_overviews(store_root: str, manifest: dict, by_shard: dict, *, store_kw
                 )
                 if entry is not None:
                     entries[key] = entry
+            if entries:
+                materialized.add(k)
             fresh = {
                 "spec": _sweep().SWEEP_SPEC,
                 "family": "overview",
@@ -275,8 +388,14 @@ def sweep_overviews(store_root: str, manifest: dict, by_shard: dict, *, store_kw
                 import obstore
 
                 obstore.put(
-                    store, f"{_node_rel(node)}/{ENVELOPE_NAME}", json.dumps(fresh, indent=1).encode()
+                    store,
+                    f"{_node_rel(node)}/{ENVELOPE_NAME}",
+                    json.dumps(fresh, indent=1).encode(),
                 )
+    if counts["written"]:
+        counts["manifest_updated"] = _update_manifest_pyramid(
+            store_root, sorted(materialized), store_kwargs
+        )
     return counts
 
 
@@ -483,7 +602,9 @@ def _fold_node(
     for name, meta in fields.items():
         if meta["class"] == "exact":
             dtype = np.dtype(meta.get("dtype") or "float32")
-            slabs[name] = np.full(n_cells, _fill_scalar(meta.get("fill_value", "NaN"), dtype), dtype)
+            slabs[name] = np.full(
+                n_cells, _fill_scalar(meta.get("fill_value", "NaN"), dtype), dtype
+            )
         else:
             digests[name] = [[] for _ in range(n_cells)]
     if target_order >= shard_order:
--- a/src/zagg/sweep_overview.py
+++ b/src/zagg/sweep_overview.py
@@ -691,7 +691,8 @@ def _fold_node(
                     raise ValueError(
                         f"morton shape {morton.shape} is not the manifest cell_order "
                         f"{cell_order} subtree ({leaf_cells} cells); mixed-order source "
-                        f"leaves are unsupported this round (D24 reader gate)"
+                        f"leaves are unsupported this round (issue #347: sweep fold "
+                        f"semantics + the writer-side append guard)"
                     )
                 partials: dict = {}
                 cell_digests: dict = {}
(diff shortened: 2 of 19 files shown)


espg:
🤖 *from Claude*

Status: phases A–D of the [thread plan](https://github.com/englacial/zagg/issues/201#issuecomment-5025396182) landed (kernels, writer, pyramid block, §8.3 obligations); E/F deferred per the plan (rationale in the PR body). CI green on this push; the ruff bot posted no findings. Next run: the fresh-context adversarial self-review + fold cycle over the diff.

src/zagg/sweep_overview.py
```diff
@@ -0,0 +1,834 @@
...
+    from zarr import open_array
+
+    from zagg.grids.healpix import HealpixGrid
+    from zagg.grids.morton import morton_word
+    from zagg.hive import _utcnow, stamp_commit
+    from zagg.store import open_store
+    from zagg.windows import SCHEDULE_NONE_TOKEN, leaf_name_v3
+
+    target_order = cell_order - (shard_order - k)
+    basename = leaf_name_v3(None if key == SCHEDULE_NONE_TOKEN else key)
```
espg, line 837: 🤖 *from Claude (review)*

**[bug]** Overview basenames collide with the leaf-name grammar the full-tree walker uses, and `zagg.coverage.refresh_root_coverage` — the *sanctioned* D9 repair path — now mistakes every overview zarr for a source leaf. That walk classifies any `*.zarr` common prefix under a digit node as a leaf, reads its commit stamp (overviews carry one), then `split_leaf_name` + `_decimal_order`; it never consults `role`. Two outcomes, both reproduced on this branch with the PR's own fixtures:

- `_decimal_order("all") == 2`, so on a **shard_order-2** store the escape hatch dies outright: `refresh_root_coverage(root)` → `ValueError: malformed decimal Morton id 'all'` (the `morton_words_from_decimals(decimals)` call at the end of the walk is unguarded, unlike the two carve-outs above it).
- On a windowed store a legal window label made of `[1-4]` digits with length `shard_order+1` parses as a *valid* decimal. shard_order 3 + window `2411` rebuilds the root MOC as `[["2411","2411"], ["-3111","-3111"]]` — a phantom positive-base shard, which then feeds `_candidate_decimals` here and every other root-MOC consumer.

The PR body's "display readers stop at it" is not implemented anywhere: `role` is grepped for nowhere outside this new module. Fix: gate the walker on the D11 role attr (skip a `.zarr` whose root attrs carry `role`) and/or only treat a `.zarr` at exactly `shard_order` depth as a leaf; either way `morton_words_from_decimals` must not be able to raise out of the repair path.

src/zagg/sweep_overview.py
```diff
@@ -0,0 +1,834 @@
...
+        logger.warning(f"sweep[overview]: fold failed at node {node} window {key!r}; ({e})")
+        counts["failed"] += 1
+        return None
+    if fold is None:
+        counts["empty"] += 1
+        return existing_entry
+    if (
+        isinstance(existing_entry, dict)
+        and existing_entry.get("generation") == fold["generation"]
+        and existing_entry.get("content_hash") == fold["content_hash"]
```
espg, line 582: 🤖 *from Claude (review)*

**[correctness-risk]** Skip-if-current trusts the *sidecar envelope*, not the artifact, so a lost overview zarr is never regenerated — the D9 self-healing property the JSON families have for free. `_roll_node` compares only `existing_entry`'s `generation` and `content_hash`, both read out of `{node}/overview.rollup.json`, and returns `current` without ever probing `{node}/{basename}`:

```python
        and existing_entry.get("generation") == fold["generation"]
        and existing_entry.get("content_hash") == fold["content_hash"]
    ):
        counts["current"] += 1
```

The other families self-heal because `_read_rollup` reads the artifact itself; here the bookkeeping and the artifact are two objects. Reproduced on this branch: delete only the overview zarrs (leaving the envelopes) and re-sweep → `{"written": 0, "current": 2}` with zero `all.zarr` on disk. `test_deleting_every_overview_leaves_leaf_reads_green` hides this by unlinking the envelopes too, so the D9 claim is only tested for a both-deleted store.

Fix: on a generation+hash match, confirm the object is really there before counting it `current` — one `read_commit` GET of the overview root, which doubles as proof the D4 stamp landed (a torn prior write also currently reads as `current`). Then let the test delete only the zarrs.

src/zagg/semantics.py
```diff
@@ -74,6 +74,90 @@
...
+#: addition, min/max by extremum — so an up-tree fold of leaf values equals a
+#: direct aggregation at the coarser order, byte for byte (§8.3). Names are
+#: matched after :func:`_fold_function_name` normalization, so ``min``,
+#: ``np.min`` and ``numpy.min`` all key the same law.
+EXACT_MERGE_LAWS = {
+    "len": "sum",
+    "count": "sum",
+    "sum": "sum",
+    "nansum": "sum",
+    "min": "min",
```
espg, line 97: 🤖 *from Claude (review)*

**[correctness-risk]** The exact-class fold is nan-aware *unconditionally*, which is right for `nansum/nanmin/nanmax` but diverges from `sum/min/max` — and this table maps both onto the same law. `sweep_overview._is_missing` treats any NaN in a float slab as missing, so a leaf cell whose stored value is NaN *because the data contained a NaN* (`np.min` over a slice holding one) is dropped from the fold, where a direct aggregation at the coarser order returns NaN. The shipped config is exposed, not just a synthetic one:

```
composability_classes(default_config("atl06"))
  -> {"count": "exact", "h_min": "exact", "h_max": "exact", ...}
# h_min/h_max: function "min"/"max", fill_value "NaN"
```

So §8.3's "byte-equal to direct aggregation at the coarser order" and `fold_dense`'s "Exact by construction" hold only under the extra premise that no leaf value is ever NaN-with-data — which the PR body states unconditionally.

Fix, either way: (a) restrict `exact` to the nan-aware variants plus `len`/`count` (the store encoding makes the non-nan forms unrecoverable regardless, since NaN fill and NaN data are the same bytes), or (b) keep the widened set but say so — qualify the docstring and record the fold as nan-skipping in the pyramid block's `method`, so a reader of the overview knows which reduction it actually got.

src/zagg/sweep_overview.py
```diff
@@ -0,0 +1,834 @@
...
+    from zagg.windows import SCHEDULE_NONE_TOKEN
+
+    if not windowed:
+        return [(SCHEDULE_NONE_TOKEN, [None])]
+    labels = sorted(
+        {w for w in dirty_windows if w is not None}
+        | {key for key in entries if key != SCHEDULE_NONE_TOKEN}
+    )
+    work = [(w, [w]) for w in labels]
+    if decl.get("all_time") or SCHEDULE_NONE_TOKEN in entries:
```
espg, line 495: 🤖 *from Claude (review)*

**[bug]** An explicit window labeled `all` silently loses its per-window overview to the all-time fold. `SCHEDULE_NONE_TOKEN` is `"all"`, and nothing rejects it as an *explicit* label: `validate_label("all")` passes and `leaf_name("-311", "all")` → `-311_all.zarr` is a legal `/1`/`/2` leaf — only `leaf_name_v3` refuses the token. With windows `["2024", "all"]` and `all_time: true`, `_window_work` yields `("all", ["all"])` from `labels` and then appends `("all", labels)`; both items resolve to the same `all.zarr` basename and the same `entries["all"]` key, so the all-time fold overwrites the per-window overview. The next pass then drops it from `labels` (`{key for key in entries if key != SCHEDULE_NONE_TOKEN}`), so it is never regenerated — one window's overview is gone permanently with no warning.

Fix: reject `all` as an explicit window label at config validation (mirroring `leaf_name_v3`'s refusal, which already calls the collision out), or key the all-time entry on a token the explicit grammar cannot produce.

tests/test_sweep_overview.py
```diff
@@ -0,0 +1,821 @@
...
+        span = 4 ** (CELL_ORDER - SHARD_ORDER - (SHARD_ORDER - k))
+        pooled: dict[int, list] = {}
+        for (dec, row), obs in self.OBS.items():
+            shard_rank = {"-311": 0, "-312": 1}[dec]
+            cell = shard_rank * span + row // 4 ** (SHARD_ORDER - k)
+            pooled.setdefault(cell, []).extend(obs)
+        return pooled
+
+    @pytest.mark.parametrize("k", [1, 0])
+    def test_rollup_equals_direct_at_every_declared_order(self, tmp_path, k):
```
espg, line 768: 🤖 *from Claude (review)*

**[test-gap]** Nothing pins the overview `morton` array against the leaf cells actually folded into each slot. The writer derives slab positions from base-4 rank arithmetic (`start = _rel_rank(dec, node) * span`, then `i // fold_factor`) but *labels* them with `generate_morton_children(morton_word(node), target_order)` — two independent orderings that must agree, or every overview cell is attached to the wrong morton id while all the value assertions still pass. This oracle cannot catch that, because `_direct()` recomputes the writer's own formula:

```python
        cell = shard_rank * span + row // 4 ** (SHARD_ORDER - k)
```

and no test in the file ever reads the overview's `morton` array. I checked the invariant does hold today (order-3 overview at node `-3/1` labels `-3111..-3144` and each slot's `h_min` equals the min over the leaf cells whose morton parent matches), so this is a regression pin rather than a live bug. Cheap version: resolve `j` from the overview's own `morton` values (`morton_decimal(...)` of the leaf cell, truncated one digit) instead of from rank arithmetic, and assert `h_min[j]` against the pooled leaf minimum — plus that unclaimed slots stay fill.

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Fixed in `b6b8bbe`. The repair walk is now role-aware: a `.zarr` whose node depth is not the manifest `shard_order` gets one small `zarr.json` GET and is skipped when its root attrs carry `role` (D11 — authoritative, never inferred from position; the depth gate keeps that GET off every source leaf). Independently, a stamped leaf whose id is not a D1 morton decimal is now skipped with a warning at the collection site, so `morton_words_from_decimals` can no longer raise out of the escape hatch — `all.zarr` at shard-order depth included. Both of your reproductions are pinned: `test_repair_walk_ignores_overviews` (shard_order 2, `refresh_root_coverage` used to die on `malformed decimal Morton id 'all'`) and `test_repair_walk_ignores_digit_labeled_overviews` (shard_order 3 + window `2411`, root MOC used to gain the phantom `2411` shard), plus `test_non_decimal_stamped_leaf_skipped_with_warning` in `tests/test_coverage_root.py`. The #215/#327 carve-outs (unstamped debris silent, foreign-order and malformed-window leaves warned) are unchanged and still green.

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Fixed in `ac05bbc`. Skip-if-current now also requires `_overview_committed(...)` — one `read_commit` GET of the recorded overview root — so the generation+hash match is only trusted when the artifact is really there and stamped. Two new tests: `test_deleting_only_the_zarrs_regenerates_them` (deletes only the zarrs, keeps the envelopes: was `{written: 0, current: 2}` with nothing on disk, now `{written: 2, current: 0}` regenerating byte-identical hashes) and `test_torn_overview_write_is_not_current` (zarr present, commit stamp unlinked → rewritten, not counted current). Both fail on the parent commit and pass on the fix. I left `test_deleting_every_overview_leaves_leaf_reads_green` as-is so the both-deleted regime stays pinned too.

src/zagg/semantics.py
espg: 🤖 *from Claude*

Took your option (b) in `d211fff` — the premise is now stated and machine-readable rather than implicit, and I did not restrict the exact class. Reasoning, in your own parenthetical: the store encoding makes the NaN-propagating forms **unrecoverable**, not merely unimplemented (fill NaN and datum NaN are the same bytes at the leaf), so moving plain `min`/`max`/`sum` out of `exact` would not buy fidelity — it would drop `h_min`/`h_max` from every atl06 pyramid and leave `count` alone in the exact class. Instead: new `EXACT_NAN_POLICY = "skip"` in `sweep_overview`, recorded per field in the pyramid declaration (`build_pyramid_block`) and in every overview's root attrs (`_field_provenance`), so a reader of the overview knows the reduction it got; the §8.3 claim in the module docstring, `fold_dense`'s "exact by construction" line, and `EXACT_MERGE_LAWS`' docstring are all qualified to "byte-equal to a direct **nan-skipping** aggregation". Tests: `test_nan_datum_is_skipped_not_propagated` asserts the fold equals `np.nanmin`/`np.nansum` **and** that plain `np.min`/`np.sum` diverge (NaN), plus `test_nan_datum_skipped_even_under_a_numeric_fill` for the non-NaN-fill case. If you want the stricter classes anyway (exact = nan-aware variants + `len`/`count`, everything else `none`), that is a one-line change to `EXACT_MERGE_LAWS` — flagging it as yours to rule on since it changes which fields ship in the shipped-config pyramids.

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Fixed in `39fc6c7`, both halves. Primary: `_validate_windowing_windows` now rejects an explicit window labeled with `SCHEDULE_NONE_TOKEN`, mirroring `leaf_name_v3`'s refusal ("would alias the all-time artifact") — pinned by `test_explicit_reserved_all_label` in `tests/test_hive_windows.py`. Secondary, for manifests written before that guard: `_window_work` drops the per-window item for that label when the all-time fold is also scheduled and warns, then appends the all-time item — whose `fold_windows` still includes the label, so the leaf's data lands in `all.zarr` rather than being silently overwritten and lost. `test_window_named_all_folds_into_the_all_time_overview` covers it (windows `2019` + `all`, `all_time: true` → 2 written, warning logged, `all.zarr` count 3). I left the `all_time`-off case alone: there the single `("all", ["all"])` item is not a collision and writes correctly.

tests/test_sweep_overview.py
espg: 🤖 *from Claude*

Added in `0f15027`: `test_morton_labels_address_the_folded_cells[k=1,0]` resolves every slot from the overview's **own** `morton` value — `morton_decimal()` of each leaf cell truncated to the target order gives the overview cell id, and the assertion is `count[j] == len(pooled)` / `h_min[j] == min(pooled)` for the slot whose label matches, fill for every unclaimed slot, plus labels distinct, all under the node, and no populated cell left unlabelled. No rank arithmetic anywhere in the oracle. Confirmed it bites: reversing the `morton` write in `_write_overview` (`arr[:] = words[::-1]`) fails both parametrizations while `test_rollup_equals_direct_at_every_declared_order` still passes — exactly the gap you described.

espg:
🤖 *from Claude*

Status: adversarial self-review + fold cycle complete on the phase A–D diff. The reviewer posted [5 inline findings](https://github.com/englacial/zagg/pull/344#discussion_r3680579971); all 5 folded, one commit each (`b6b8bbe` repair walk skips overviews, `ac05bbc` skip-if-current probes the overview zarr, `d211fff` declared exact-fold nan policy, `39fc6c7` reserved `all` window label rejected, `0f15027` independent morton-label pin), with a reply + fix sha on every thread. CI green on `0f15027`; the ruff bot posted no findings. Also cross-checked the store shape against the frozen spec §4 on PR #346 — key-for-key conformant; the one post-spec addition (`nan_policy` on exact field entries, from the fold) is [flagged there for a §4 amendment](https://github.com/englacial/zagg/pull/346#issuecomment-5127860012).

Applying `waiting`: the ball is with espg — the [finding-3 ruling](https://github.com/englacial/zagg/pull/344#discussion_r3680582363) (keep `min/max/sum` exact with declared `nan_policy: skip`, vs reclassifying them out of `exact`) plus the PR-body questions, chiefly (1) `overview` staying in `DEFAULT_FAMILIES` ahead of fleet-scale sizing. Phases E/F remain deferred per the plan (E gates on measured fleet wall-time; F on roster-kind ragged fields existing).

espg:
🤖 *from Claude*

Acted on the in-session rulings: `553ca04` repoints the mixed-order skip's rationale at #347 (mortie#116 merged, moczarr#8 closed — no longer citable as gates; PR-body question (4) rewritten to match), and `b84f31b` opts the fleet-scale configs out of the overview sweep via the existing yaml knob `output.pyramid: false` (the three `*_hive*` benchmark arms, `s2_neon_o9`, shipped `atl03_tdigest_strata_healpix` — each commented "overview sweep opted out pending Phase E fleet sizing"). `DEFAULT_FAMILIES` keeps `overview` unchanged; PR-body question (1) is marked RESOLVED accordingly.

Re-applying `waiting` — still espg's court: the [finding-3 nan-policy ruling](https://github.com/englacial/zagg/pull/344#discussion_r3680582363) and the all-time-fold default (question (2)).


CI is running on `b84f31b` (kicked via close/reopen after a PR-event stall); the prior head `0f15027` was fully green and the two new commits are text/config-only.

src/zagg/semantics.py
espg: 🤖 *from Claude*

espg accepted option (b) as landed (in-session ruling: "this is correct behavior in my view, it just needs to be documented" — "(b) + the warning"): nan-skipping IS the intended semantics for the exact fold, and the `nan_policy: "skip"` declaration in the pyramid block and overview attrs stands. The one directed addition is now in `818a788`: config validation warns when an exact-class field pairs a plain `min`/`max`/`sum` with a NaN fill — "NaN data is indistinguishable from fill at read; the reduction behaves as its nan-form in overviews and any NaN-carrying result is unrecoverable — use the nan-form explicitly to silence." Warning, not error: the shipped atl06 config trips it by design and stays valid (tested, `TestNanAmbiguousReductionWarning`). This thread's stricter-class alternative is closed.

Issue #342, opened 2026-07-30
Freeze the O11 content-hash serialization: writer must match the moczarr reader's pinned recipe (keys, combined, vlen)


espg:
🤖 *from Claude*

Refs #340 (the spec home this belongs in), `docs/design/sparse_coverage.md` O11, and espg/moczarr#23 (whose review round forced the freeze).

## Why now

zagg's writer does **not** yet record `content_hashes` — the only in-tree surfaces are dead reads in `dedup.py`. moczarr's PR #23 shipped the O11 *verify* reader, and its review round made it pin reference choices that zagg's future writer must reproduce exactly, or every deployed verifier breaks / the conformance fixture regenerates. Recording them zagg-side **before** the writer ships is the whole point of this issue.

## The frozen contract (as pinned by the moczarr reader)

1. **Per-array keys**: path relative to the leaf root (e.g. `"8/morton"`).
2. **Combined hash**: sha256 over the **sorted** per-array hex digests joined by `\n`.
3. **Sidecar shape**: `{"arrays": {...}, "combined": ...}`; the flat form is accepted on read but the writer emits the structured form.
4. **Fixed-width arrays**: sha256 over decoded values, C-order, little-endian.
5. **Vlen/ragged arrays** (the part that MUST NOT be improvised — a naive `.tobytes()` hashes heap pointers, verified non-deterministic across processes): hash the decoded per-cell payloads **length-prefixed, in cell order** — the exact recipe is being finalized in moczarr PR #23's fold round; whatever it lands is normative and this issue's deliverable copies it verbatim.

## Deliverables

1. The contract lands as the **O11 section of #340's spec** (normative text + the golden fixtures gaining recorded `content_hashes` so writer and verifier prove conformance against the same bytes).
2. The zagg **O11 writer** (worker-side, in the stats-sidecar write path) implements it — keying, combined recipe, vlen recipe — with cross-checks against moczarr's verifier outputs on the shared fixture (both directions: zagg-written hashes verify in moczarr; moczarr's fixture hashes verify through zagg's hasher).
3. `dedup.py`'s dead reads reconciled with the real field.

## Acceptance

- Round-trip: a leaf written by zagg + hashed by the new writer returns `match: True` from moczarr's `verify_arrays` for every array including ragged digest + locations vlen arrays.
- The recipe text in #340's spec, moczarr's docstrings, and zagg's writer docstring are literally identical (one normative source, two citations).

espg:
🤖 *from Claude*

[...]

PR #27, opened 2026-06-11
Add bring-your-own-role path for IAM-constrained deploys; creds handling for external s3 bucket writes

espg:
template.yaml gains CreateExecutionRole/ExecutionRoleArn so the stack
can reference an admin-created execution role instead of minting one,
for accounts lacking iam:CreateRole (e.g. SSO power-user). stand_up.sh
gains CREATE_ROLE/ROLE_ARN. execution_role.yaml is a standalone template
for an admin to create the least-privilege role in its own stack;
EXECUTION_ROLE.md documents the workflow. Default stays CreateExecutionRole=true.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #346, opened 2026-07-30
Promote zagg-ragged/1 + digest + composition byte layouts to a normative spec (issue #340)

### LINKED ISSUE #340: Promote zagg-ragged/1 + digest + composition byte layouts to a normative spec (reader-migration gate)
🤖 *from Claude*

Refs #265 (the reader-migration gate bullet), #209 (the vlen-bytes layout), #87 (located channel), #321/PR #334 (composition).

Promote the byte-level store conventions from code comments to a normative spec document (the way mortie's `docs/specification.md` governs hive paths), so an independent reader (moczarr) can decode without importing zagg or reverse-engineering `grids/base.py`:

1. **`zagg-ragged/1`**: the vlen-bytes layout — one array per ragged field on the cells axis, cell payload = raw little-endian bytes of `(n, *inner_shape)`, `b""` fill, the versioned `ragged` attrs block (`spec`/`element`/`locations`), codec chain `[vlen-bytes, zstd(3)]`, ShardingCodec geometry, the located sibling naming + row alignment.
2. **Digest payload semantics**: `(k, 2)` float32 (mean, weight) centroids, sorted by mean; weight sums = observation counts; located words = order-29 point words / common ancestors (defer algebra to zagg, spec the bytes).
3. **`zagg-composition/1`**: the packed uint64 lane layout, presence-floor quantization, and the attrs block (`spec`/`lanes`/`of`/`threshold`) — normative text exists in `docs/signal_strata.md`; consolidate/cross-link.
4. Golden fixtures: a tiny committed leaf + expected decoded values, so a reader implementation can prove conformance without S3.

Home: `docs/` alongside `ragged_layout.md`/`signal_strata.md` (likely a promotion+consolidation of those two into explicitly versioned spec sections rather than a new doc). This issue is the named **gate for the moczarr reader adoption** (moczarr-side issue links here). Note the issue #210 typed-dtype migration supersedes the attrs seam eventually — the spec should state its own succession plan.

espg:
Closes #340.

Promotes the byte-level store conventions to a normative spec document, `docs/specification.md`, so an independent reader (moczarr) can decode a zagg store from the spec + committed conformance fixtures alone — no zagg import, no reverse-engineering of `grids/base.py`. Follows the [approved plan](https://github.com/englacial/zagg/issues/340#issuecomment-5125189863) on the issue thread ([all three leans approved](https://github.com/englacial/zagg/issues/340#issuecomment-5125505363): single spec doc, §4 written now with the in-flight marker, two-leaf fixture split) and the [ratified amendment](https://github.com/englacial/zagg/issues/340#issuecomment-5125245251) pinning `zagg-ragged/1` as the 1.0 wire contract.

Approach: `docs/specification.md` is the mortie-precedent normative page (MUST/SHOULD language, versioned `spec`-marker conformance, frozen revisions with explicit succession). `ragged_layout.md` and `signal_strata.md` become narrative companions that cite it — no duplicated normative text. Golden fixtures under `tests/data/spec/` are generated by zagg's real write path (`tools/generate_spec_fixtures.py`) and pinned by `tests/test_spec_conformance.py`, so the spec, the fixtures, and the shipping reader cannot drift apart silently; moczarr vendors the same fixtures for its parity gates (moczarr#19/#20).

## Phases

- [x] Phase 1 — spec home + skeleton: `docs/specification.md` (normative-language + conformance sections, TOC), mkdocs nav entry (412acc1)
- [x] Phase 2 — §1 `zagg-ragged/1`: layout, attrs block, codec chain, golden wire framing, storage geometries, succession clause; `ragged_layout.md` becomes the narrative companion (e167aff)
- [x] Phase 3 — §2 digest payload: `(k, 2)` float32 (mean, weight) sorted by mean, exact-count weights, located channel word semantics; algebra deliberately not specified (b86f86b)
- [x] Phase 4 — §3 `zagg-composition/1`: LSB-first lanes, presence floor, attrs block, normative merge law, golden word; `signal_strata.md` becomes the narrative companion (a0f0e85)
- [x] Phase 5 — §4 pyramid/overview declarations (ratified #201/D22-D24 grammar, implementation-in-flight marker) + §5 O11 content hashes (the #342 freeze: keys, vlen recipe, combined hash — verbatim from the merged espg/moczarr#23 verifier) (f458f99)
- [x] Phase 6 — §6 `zagg-ragged/2` typed-dtype full normative section (#210's ratified hybrid-C: `vlen-ndarray` name, byte-identity requirement, revision signaling) (e42840e)
- [x] Phase 7 — golden conformance fixtures: `tools/generate_spec_fixtures.py` (production write path — `ensure_manifest` + `process_and_write_hive`), committed `minimal` + `kitchen_sink` hive stores under `tests/data/spec/` (~124 KB) with `*.expected.json` goldens, `tests/test_spec_conformance.py` (38 tests), spec §7 (ef88694)
- [x] Phase 8 — wiring: `design/sparse_coverage.md` division-of-labor pointer, CLAUDE.md §4 spec-maintenance pointer (per the approved plan's phase 7 — flagged below), notes on moczarr#19/#20 (166e908)

## How it was tested

- `uv run pytest -q`: **2842 passed, 35 skipped** (one pre-existing env-dependent failure deselected — see below). The new `tests/test_spec_conformance.py` (38 tests) pins every normative byte claim two ways: through the shipping readers (`readers/tdigest_tensor`, `stats/composition`), and through **spec-text-only decoders** (struct + zstd + hashlib, no zagg read path) for the §1.4 framing, §1.5 shard index/sentinel, and §5 O11 recipes — proving the spec text alone suffices to decode the store (the moczarr acceptance criterion).
- `ruff check` / `ruff format --check` clean on everything this PR touches.
- Fixture edge cases committed: empty inner chunk (index sentinel), empty cells in populated chunks (`b""` fill), a noise-only cell with an empty `(0, 2)` signal payload, the §3.1 golden composition word `0xFF000000FF0000FF` from a single photon, merged (weight > 1) centroids with common-ancestor location words, exact stratum counts as digest weights.

## Notes / pre-existing failures (not touched here)

- `ruff check src tests` fails on main with N818 (`registry.py:64` `UnknownCapability`) — pre-existing, unrelated (N is not in the PR lint bot's `--select=E,F,W,I` gate).
- `ruff format --check src tests` flags `tests/data/benchmark/README.md` on main — pre-existing, unrelated.
- `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` fails in this local env on main too (`pip` in the layer build cannot resolve `zarr>=3.1.5`) — environment-dependent, unrelated.

## Questions for review

- §4 records the ratified pyramid grammar (`role`/`zagg_overview`/`zagg-pyramid/1`) that the in-flight #201 branch implements; per the approved plan, any divergence found while landing #201 resolves on the spec first.
- Phase 8 adds one bullet to **CLAUDE.md §4** (the spec-maintenance rule from the approved plan's wiring phase). CLAUDE.md edits are sensitive by convention — the bullet is deliberately minimal and shown in the diff; drop it if unwanted.
- §6.3 leaves one deliberate reservation: the located-sibling binding key under `/2` (where `locations` lives once the `ragged` attrs block is retired) is reserved for the `/2` implementation PR to pin in that section before any `/2` store exists.


espg:
🤖 *from Claude*

Status: all 8 phases landed (see the body checklist for shas). Continued from a prior session's unpushed phases 1-4 (same branch, commits 412acc1..a0f0e85 — the session stopped before pushing or opening this PR). Full local suite green (2842 passed / 35 skipped; three pre-existing env failures flagged in the body, untouched). Ruff CI check green, no lint-bot inline comments. Gate notes posted on [moczarr#19](https://github.com/espg/moczarr/issues/19#issuecomment-5127578028) and [moczarr#20](https://github.com/espg/moczarr/issues/20#issuecomment-5127578713). Left in draft pending the adversarial-review/fold cycle.

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the normative spec, at the higher bar the PR asks for. 16 inline findings: 2 medium-high, 8 medium, 1 low-medium, 5 low.

**What I ran** (not just read): `uv sync --extra test`; `uv run pytest tests/test_spec_conformance.py -q` → **38 passed**; `uv run python tools/generate_spec_fixtures.py` + `git status` to test the determinism claim; hand-decoded the committed shard objects per §1.4/§1.5; recomputed the §4.3 node/order arithmetic against `mortie.MortonIndexArray.order()`; recomputed the §3.1 golden word in both bit orders; diffed §5 against the merged `espg/moczarr` `src/moczarr/stats.py` (fetched via `gh api`); diffed §4 against `sweep_overview.py`/`hive.py` on `claude/201-pyramid-sweep` and against the three #201 ruling comments; grepped every spec constant (`RAGGED_SPEC`, `COMPOSITION_SPEC`, `LANES`, `PYRAMID_SPEC`, `OVERVIEW_SPEC`) for writer-side use.

**Skipped**: the full `pytest` suite (this PR touches no `src/` — confirmed via `gh pr diff --name-only`); the mkdocs build (not in the sync'd env, and `mkdocs.yml` isn't `strict`, so a bad anchor wouldn't fail CI either way).

### Held up under scrutiny

- **§1 against the shipping writer, line by line.** `grids/base.py:ragged_array_spec` matches §1.2 (`spec` + `element` + conditional `locations`), §1.3 (`[vlen-bytes, zstd(level=3, checksum=False)]`), and §1.5 (`sharding_indexed`, `chunk_shape == inner`, `index_codecs [bytes(little), crc32c]`, `index_location: end`) exactly. `fill_value: ""` ✓. The §1.2 reserved-key claim is real (`config.py:557`). The "sibling carries no user attrs" claim is real — `healpix.py:628` builds it *without* `apply_field_attrs`.
- **The spec-text-only decoders are the genuine article** for §1.4/§1.5. `TestWireFraming` imports only `struct`, `numcodecs.Zstd` and `numpy` — no `zagg.readers`, no `zarr` — parses the index suffix at `-(16*k+4):-4`, honours the `2^64-1` sentinel, walks `u32 count || (u32 len || payload)*`, and asserts `pos == len(raw)` so trailing-byte drift can't hide. I re-derived the same walk independently and it lands on the same payloads. `test_populated_chunks_only_no_ghost_payloads` closes the empty-cell-vs-`b""`-fill direction. This is the strongest part of the PR, and it does what the §7 acceptance criterion claims.
- **§2.2 is exactly right, including the subtle part.** Verified order 29 (`MortonIndexArray.from_latlon(..., points=True).order() == 29`) and that `_centroid_ancestors` copies a singleton's word through *unchanged* rather than round-tripping it through `common_ancestor` — so "a centroid of weight 1 carries its single member's word unchanged" is a real code guarantee, not an aspiration. Multi-member → `common_ancestor` ✓. §2.3's carve-out boundary is clean: bytes in, algebra out, with the "order-dependent / `np.isclose`" consequence stated rather than hidden.
- **§3.1's golden word is correct in both directions.** Confidences `[4,-1,0,3,1]` @ `threshold=2` → land + land_ice clear, strongest 4 ⇒ high → lanes `[255,0,0,255,0,0,0,255]` → LSB-first `0xFF000000FF0000FF`; and the MSB-first foil `0xFF0000FF000000FF` is right too, which is what makes the pin discriminating. §3.2's `np.rint` → presence-floor → clip and §3.4's "nonzero on **either** input" both match `_quantize` / `merge_composition`, including the `n <= 0` identity behaviour.
- **§5 is a faithful verbatim adoption** on every point I could diff: leaf-relative keys (`"8/morton"`), sha256 over decoded C-order LE bytes with big-endian byte-swapped first, the u64-LE length-prefixed vlen recipe with the injectivity rationale, and `sha256("\n".join(sorted(hex_digests)))` with names excluded. §5.3's structured-MUST / flat-SHOULD split and the "unverifiable ≠ tampered" posture both match `_recorded_hashes`. The one gap is the element-normalization function, filed above.
- **§4 matches the landed #344 grammar** on the things that matter: `zagg-overview/1` / `zagg_overview` / `role` (`OVERVIEW_SPEC` / `OVERVIEW_ATTR` / `ROLE_ATTR` ✓), `cell_order = source_cell_order - (source_shard_order - order)` ✓, `spacing` default 2 = `DEFAULT_SPACING` ✓, `orders` descending ✓, `zagg-pyramid/1` ✓, `all` as the reserved all-time token (`windows.SCHEDULE_NONE_TOKEN == "all"` ✓), the `summarize` shape ✓ — and the write order, template → arrays → `role`/provenance → stamp LAST, is precisely what `_write_overview` does, so "stamp presence certifies the `role` attr landed" is sound. The regenerable-cache and never-inferred-from-position rules track D11/D24 and the #201 rulings.
- **§6's byte-identity story is coherent**, and §6.2's "metadata-only rewrite" genuinely follows from the §1.4 pin actually being pinned. `vlen-ndarray` naming, the retirement-vs-`/1`-forever framing, and the missing-extension failure mode are all stated well.
- **§7's fixture claims check out empirically.** Geometry ✓ (shard 4 / chunk 5 / cell 6 → 16 cells, K = 4, 4 per chunk); chunk ordinal 2 absent from every shard index ✓; empty cells inside populated chunks ✓; single-photon golden cell ✓; noise-only cell with an empty `(0, 2)` signal payload and `n_signal = 0` ✓; merged centroids present (max weight 29 in `minimal` cell 15; 17 / 11 in `kitchen_sink`) so §2.2's common-ancestor path is genuinely exercised ✓. 124 KB, and both leaves are stamped and `role`-free, which pins "absence means source".
- **RFC-2119 discipline is clean.** Only the four declared keywords appear — no stray `SHALL` / `REQUIRED` / `SHOULD NOT` outside the declared vocabulary — 29 `MUST`s, and the Contract / *informative* marking is applied consistently. The de-duplication of `ragged_layout.md` / `signal_strata.md` into narrative-plus-pointer is done properly: I looked for leftover normative text and found one overclaim (filed), not a second copy of the byte layouts.
- **The CLAUDE.md §4 bullet and the mkdocs nav entry are correct as wiring.** The bullet sits with the other §4 quality rules, names both artifacts a wire change must touch, and matches what this PR actually built. `sparse_coverage.md`'s `../specification.md` relative link resolves. All in-body §-anchors resolve under python-markdown's slugify; only the TOC's `#4-pyramid-overview-declarations` differs on GitHub (which renders `4-pyramid--overview-declarations` for the `/` in the heading) — harmless on the docs site, mildly annoying for anyone reading the spec on GitHub, which is where moczarr will read it.

Nothing here blocks the direction. Both medium-highs are "the spec claims an enforcement that doesn't exist yet" rather than "the spec mis-states shipped bytes" — and I found no instance of the latter, which is the failure class that would actually hurt.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+```json
+"composition": {
+  "spec": "zagg-composition/1",
+  "lanes": ["land", "ocean", "sea_ice", "land_ice", "inland_water", "low", "med", "high"],
+  "of": "h_tdigest_signal",
+  "threshold": 2
+}
+```
+
+- **`lanes`** — the lane names in bit order (LSB byte first). For `/1` the
```
espg, line 377: 🤖 *from Claude (review)*

**medium-high** — the whole `composition` attrs block is author-declared config with zero writer-side derivation, so a valid zagg config can emit a store that violates this Contract silently

`§3.3` is marked **Contract** and tells readers they `MUST bind to it` — but `COMPOSITION_SPEC` and `LANES` are dead constants outside their own module:

```
$ grep -rn "COMPOSITION_SPEC\|LANES" src/zagg/ | grep -v stats/composition.py
(nothing)
```

The block reaches the store only through `aggregation.variables.composition.attrs`, merged verbatim by `grids/base.py:apply_field_attrs` (whose own docstring concedes "a config could in principle shadow spec-owned metadata; the reserved `ragged` block is the one key protected from that"). `config.py:_validate_composition_attrs` checks exactly two things — `of` names a declared `kind: ragged` field, and `threshold == params.threshold`. It never checks `spec`, and it never checks `lanes`.

Consequence: a config declaring `spec: "zagg-composition/1"` with `lanes` permuted (or truncated, or renamed) validates, writes, and passes the whole suite, while `pack_composition` keeps packing in `LANES` order. An external reader doing exactly what §3.3 mandates — bind to `lanes` in bit order — then mis-labels every lane. This is the opposite posture from `zagg-ragged/1`, where `ragged_array_spec` *stamps* `RAGGED_SPEC` writer-side and `config.py:557` reserves the key against shadowing.

Note the §7 fixture inherits the same circularity: `test_attrs_block` asserts the block equals a literal that came from `tools/generate_spec_fixtures.py:_config()`'s own hand-typed copy, not from `LANES`. Two ways out, either fine, but the spec shouldn't claim Contract without one: (a) derive `spec`/`lanes` writer-side from the module constants (mirroring `ragged_array_spec`) and say so here; or (b) state plainly that the block is author-declared, and extend `_validate_composition_attrs` to reject a `spec`/`lanes` that disagree with `COMPOSITION_SPEC`/`LANES` — the same "enforced at config validation" language §1.2 already uses.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+digest carries merged centroids (weight > 1) whose location words are
+common ancestors (§2.2).
+
+Each fixture ships a sibling **`{name}.expected.json`** recording the shard
+key, leaf path, geometry, every populated cell's decoded values (digest
+centroids, location words and composition words as decimal strings — JSON
+numbers cannot carry `uint64` faithfully), the per-stratum exact counts,
+and the §5 O11 `content_hashes` (per-array + combined). Expected values
+were computed from the generator's *inputs* — the arrays handed to the
+writers — never read back through a zagg reader, so the reader is pinned,
```

**medium-high** — the "never read back through a zagg reader, so the reader is pinned, not self-certified" claim is false for `content_hashes`, and the O11 golden is a duplicate of the code it tests

True for the cell values (`_build_cells` records the reducer *inputs*). Not true for `content_hashes`: `tools/generate_spec_fixtures.py:336` calls `_o11_hashes(str(out / leaf_rel))`, which opens the **written leaf** and reads `node[...]` back out. And `_o11_hashes` (lines 262–284) is a line-for-line duplicate of `tests/test_spec_conformance.py::TestContentHashes._hash_leaf` (lines 311–329) — same enumeration, same length prefix, same joiner. So `test_per_array_hashes_match_golden` asserts one copy of the recipe against the other copy's output over the same bytes: change the prefix width to 4, or drop the prefix, or switch the joiner, in *both* copies and every test still passes.

This is exactly the finding moczarr's own review round raised and fixed — `ba804e6`, "the combined-hash golden was self-referential; pinned by the `FROZEN_COMBINED_4111` literal plus a hand-rolled chain", and it then narrowed its docstring claim to "the per-array digests (which are)". §7 here re-broadens the claim past what the fixtures support.

Two fixes, and the second is the one that makes §5's "adopt it verbatim" checkable: (1) narrow the §7 sentence so it covers the decoded cell values only, and say the content hashes are computed from the written store; (2) pin at least one per-array digest and the combined digest as **literal hex constants** in `test_spec_conformance.py` (moczarr's `FROZEN_COMBINED_4111` pattern) so a recipe change fails a test rather than sailing through — which is also the only mechanism that would catch a future zagg↔moczarr divergence, since neither side's fixture can catch it (moczarr#23 called that out explicitly: "This deviation cannot be caught by the fixture (both sides are this implementation's choice)").

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+
+**Status: contract.** Source of truth in code: `zagg.stats.composition`
+(issue #321). Rationale and narrative:
+[`signal_strata.md`](signal_strata.md).
+
+A composition field is one dense **uint64** word per cell carrying eight
+8-bit lanes of quantized fractions of the cell's **signal stratum**
+(`N_signal` = the signal digest's total weight — magnitude lives in the
+digest, composition here). An empty signal stratum packs to `0` (the array's
+fill value).
```

**medium** — "An empty signal stratum packs to `0` (the array's fill value)" asserts a coupling nothing requires

`pack_composition` returning `0` for an empty stratum is code truth. But "(the array's fill value)" holds only because `configs/atl03_tdigest_strata_healpix.yaml` (and `_config()` in the fixture generator) happen to declare `fill_value: 0`. It is a config knob: nothing in `_validate_composition_attrs`, `apply_field_attrs`, or the template path constrains it.

A config declaring `fill_value: 1` for a composition field writes a store where every *unwritten* cell decodes as lanes `[1,0,0,0,0,0,0,0]` — i.e. a spurious `land` presence, which under §3.2 a reader is entitled to read as "this flag occurred **exactly**". Since §3 is Contract and readers key presence off `lane > 0`, make `fill_value: 0` a **MUST** here (and, per CLAUDE.md §4, back it with the config validation that already cross-checks `of`/`threshold`).

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+values — never stored object bytes, so codec and packaging changes
+(ShardingCodec inner chunks, compressor upgrades, §1.5 geometry) are
+invisible by construction, while any value change flips the hash (exact
+bytes, no float tolerance — interpretation pairs the hash with the recorded
+zagg version).
+
+### 5.1 Scope and keys
+
+**Contract.** The hash set covers **every named zarr array beneath the leaf
+root** — data fields, the ragged vlen payload arrays and their
```
espg, line 652: 🤖 *from Claude (review)*

**medium** — the enumeration-based hash scope makes intra-leaf debris load-bearing for verification

"**every named zarr array beneath the leaf root**" is implemented on both sides as `group.members(max_depth=None)` (moczarr `stats.py:304`; the generator at line 268; the test at line 320). That means the hash set is whatever happens to be *discoverable* under the leaf, not what the template declared — so any foreign array prefix inside a leaf adds a key, changes `combined`, and reports an intact leaf as mismatched.

That is not hypothetical in this repo. PR #344's own body: "the fold opens leaf arrays **by name** — no member enumeration and no LIST anywhere on the overview path — so orphan array prefixes / foreign objects inside a leaf cannot crash it (tested with a junk array prefix + stray status-like object)", plus the issue #341 Bug A class and the issue #327 `.zarr.status/` prefixes. §5 adopts precisely the posture #344 designed around. It also collides with §5.3's stated posture that verification distinguishes "nothing recorded" from a mismatch — debris produces a *mismatch*, the loud outcome, for a leaf whose declared arrays are all intact.

Either scope the hash set normatively to the template/manifest-declared arrays (name-driven, matching #344), or state the debris consequence here explicitly so a verifier can classify an extra-key delta as "extra array present" rather than "tampered". Silence is the one option that leaves two conforming implementations disagreeing.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+
+### 2.1 Centroid array
+
+**Contract.** A populated cell's decoded payload is a `(k, 2)` **float32**
+array of weighted centroids:
+
+- column 0 is the centroid **mean**; column 1 is the centroid **weight**
+  (the number of observations merged into it, ≥ 1);
+- rows MUST be sorted **ascending by mean**;
+- `sum(weights)` MUST equal the cell's **exact** observation count — the
```
espg, line 250: 🤖 *from Claude (review)*

**medium** — "MUST equal the cell's **exact** observation count" is unrepresentable above 2^24 in the float32 the same section mandates

§2.1 fixes the payload as `(k, 2)` **float32** and then requires `sum(weights)` to be the *exact* count. Integers are exact in float32 only to 2^24 = 16,777,216; above that both an individual centroid weight and the sum round. So the MUST is satisfiable only under an unstated bound.

At leaf cell orders that bound is comfortable. It is not comfortable under §4: §4.5's default schedule walks the pyramid down to `orders: [3, 1]`, and a cell in an order-1-node overview pools an enormous ATL03 photon population — the regime `docs/design/sparse_coverage.md`'s benchmark envelope already describes as obs-volume-bound. §4.4 says "§1–§3 of this page apply to overview arrays unchanged", so this MUST is inherited by exactly the arrays most likely to break it.

Bound the claim: `sum(weights)` is the exact count while that count is ≤ 2^24, and the nearest float32 above it. A reader recovering counts from weights (which §3.3 tells it to do for `N_signal`) needs that boundary stated, not inferred.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+  carry no `role` key: absence means source.** A reader MUST check `role` on
+  every zarr it opens at an overview-carrying order; analysis readers reject
+  or skip `role: overview` zarrs, display readers MAY stop at one.
+- **`zagg_overview`** — the versioned provenance block, present exactly when
+  `role` is `"overview"`:
+
+```json
+"zagg_overview": {
+  "spec": "zagg-overview/1",
+  "node": "-311",
```

**medium** — the §4.3 example is internally inconsistent — `"node": "-311"` is an order-2 node, not order 3

mortie's decimal grammar puts one digit per order after the base-cell digit, so a node string's order is `len(digits) - 1`. Verified against the shipping stack:

```
$ uv run python -c "from zagg.grids.morton import morton_word; import mortie, numpy as np; \
  print([(k, mortie.MortonIndexArray(np.array([morton_word(k)],dtype=np.uint64)).order()) \
         for k in ('11213','-311','-3111')])"
[('11213', 4), ('-311', 2), ('-3111', 3)]
```

So `{"node": "-311", "order": 3}` cannot both be true; the order-3 node is `-3111`. Every other number in the example is self-consistent (`cell_order = 13 - (5 - 3) = 11` ✓, and `source_shard_order: 5` + `spacing: 2` is exactly what `build_pyramid_block` yields as §4.5's `orders: [3, 1]`), which makes this the one value an implementer copying the example would calibrate their node parser against and get wrong.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+      "count":     {"class": "exact", "method": "sum", "dtype": "int32", "fill_value": 0},
+      "h_tdigest": {"class": "approximate", "method": "tdigest_kway",
+                    "dtype": "float32", "inner_shape": [2], "delta": 512},
+      "photon_ids": {"class": "none"}
+    }
+  }
+}
+```
+
+- **`orders`** — the ancestor orders that carry overviews (descending; empty
```
espg, line 570: 🤖 *from Claude (review)*

**medium** — §4.5's grammar doesn't admit the declared-off form #344 actually writes, so a reader binding `pyramid.overview.fields` KeyErrors on a pyramid-off store

The block is presented as `spacing` / `orders` / `all_time` / `fields` with only `summarize` marked "(optional)". The landed implementation on `claude/201-pyramid-sweep` (`sweep_overview.py:build_pyramid_block`) emits a strictly smaller object when the knob is off:

```python
    knob = get_pyramid(config)
    if knob is None:  # output.pyramid: false — declared off
        return {"spec": PYRAMID_SPEC, "overview": {"orders": []}}
```

No `spacing`, no `all_time`, no `fields`. §4.5 does anticipate the state ("empty = pyramid declared off") but not the *shape*, so a reader that trusts the grammar and reads `pyramid["overview"]["fields"]` — which §4.4 explicitly sells as "the zero-open way to know" a field's presence — raises on every pyramid-off store. Per the PR body, divergences resolve spec-first: state which keys are absent in the declared-off form (or require the full shape and let #344 fill it in), and say which keys a reader may assume present when `orders` is non-empty.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+
+The key words **MUST**, **MUST NOT**, **SHOULD**, and **MAY** are used as in
+RFC 2119. Text marked **Contract** is frozen for the revision that carries it;
+text marked *informative* explains or motivates and binds nothing.
+
+## Conformance
+
+- Every versioned convention on this page is signaled in store metadata by a
+  `spec` marker (`"zagg-ragged/1"`, `"zagg-composition/1"`, …) — the
+  coverage-envelope discipline. A conforming reader MUST strict-check the
```

**medium** — the global conformance rule (every convention carries a strict-checked `spec` marker) is contradicted by `zagg-ragged/2`, which deliberately has none

"Every versioned convention on this page is signaled in store metadata by a `spec` marker … A conforming reader MUST strict-check the marker and **fail loudly on an unknown or future revision**."

§1.6 and §6.1/§6.3 then define `zagg-ragged/2` as signaled **by data type**, with the `ragged` marker "retired (not written)". A reader that implements the conformance rule literally — no `spec` marker ⇒ unknown ⇒ fail loudly — rejects every conforming `/2` array; a reader that doesn't has no rule telling it when marker-absence is legal.

Give the rule its carve-out here: a convention is signaled either by a `spec` marker *or* by a registered data type, and marker-absence is only legal where a section names the dtype that replaces it (§6.1). Without that, the two normative statements are directly opposed, and §6 is the section a `/2` implementer reads first.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+### 6.3 Revision signaling
+
+**Contract** (restating §1.6 from the `/2` side):
+
+- An array whose zarr data type is `vlen-ndarray` **is** `zagg-ragged/2`;
+  the `ragged` attrs marker is retired on such arrays (not written). The
+  element declaration lives in the dtype configuration alone — a reader
+  MUST NOT require the attrs block on a `/2` array.
+- A located `/2` payload array still declares its sibling binding in
+  **metadata, never by naming convention** (the §1.2 rule survives the
```
espg, line 804: 🤖 *from Claude (review)*

**medium** — §6.3's reservation is not tight enough to keep the `/2` PR from misreading it — it doesn't say whether a residual `ragged` block is allowed

§6.1 and §6.3 both say the `ragged` attrs marker is "retired on such arrays (**not written**)". But §1.2 puts `locations` *inside* that block, and §6.3 simultaneously requires a located `/2` array to declare its sibling binding "in metadata, never by naming convention". Those two sentences leave the `/2` implementer a choice the spec means to reserve:

1. write `ragged: {"locations": "…"}` — a residual block with `spec`/`element` dropped ("the marker is retired" reads as the `spec` key), or
2. write a brand-new top-level attrs key and honour "not written" literally.

(1) also silently reintroduces the §1.2 problem the reservation exists to avoid: an array carrying a `ragged` block with no `spec` is, by §1.2's own words, "**not** a `zagg-ragged/1` array" that "a reader MUST refuse … with a pointed error" — so a `/1`-only reader hitting a `/2` located array would take the refuse path with a misleading message instead of the §6.3 bullet-4 "install `zarr-vlen-ndarray`" path. Say which of (1)/(2) the reservation covers, and whether the `ragged` key stays reserved-but-unwritten under `/2`. One sentence closes it, and it's cheaper now than after a `/2` store exists.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+}
+```
+
+  `spec` follows the conformance rule (strict-check, fail loudly on an
+  unknown revision). `node` is the ancestor's morton decimal string; `order`
+  its order; `cell_order` the overview's own cell order
+  (`source_cell_order - (source_shard_order - order)` — constant tree depth,
+  §4.4); `window` the §4.2 window key (`"all"` for the all-time fold);
+  `fields` the per-field composability class + fold method actually applied;
+  `generation` the D22 staleness stamp (merged-leaf count + max leaf commit
```
espg, line 499: 🤖 *from Claude (review)*

**medium** — "the fold method **actually applied**" doesn't match what #344 writes for `class: none` fields, and the spec is silent on whether excluded fields appear at all

§4.4 says `none` fields are "**absent**" from an overview. §4.3 then describes `fields` as "the per-field composability class + fold method actually applied", which reads as: included fields only. The landed writer disagrees on both halves (`sweep_overview.py`, `_write_overview`):

```python
                "fields": {
                    n: {"class": m["class"], "method": m.get("method", TDIGEST_LAW)}
                    for n, m in fields.items()
                },
```

`fields` is the full manifest map, so `none`-class fields **are** recorded — and since `build_pyramid_block` gives them `{"class": "none"}` with no `method`, the `.get(..., TDIGEST_LAW)` default stamps them `method: "tdigest_kway"`: a fold method that was emphatically not applied, on a field that isn't in the zarr. A reader cross-checking `zagg_overview.fields` against the arrays it can open sees a declared t-digest field that doesn't exist.

Resolve spec-first: state whether `fields` enumerates all declared fields (with `none` entries carrying **class only**, mirroring §4.5's "recorded absence") or only the materialized ones. Either is fine; the current wording plus the current code is neither.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+  dtype is byte-swapped to little-endian before hashing).
+- **Vlen (ragged) arrays**: an object-dtype array has no flat buffer
+  (`ndarray.tobytes()` on it would serialize per-process pointer
+  addresses). It is hashed instead as, over cells in flat C order:
+
+```text
+sha256( for each cell:  u64_le(len(payload)) || payload )
+```
+
+  where `payload` is the cell's decoded bytes — exactly the §1.4
```
espg, line 688: 🤖 *from Claude (review)*

**medium** — §5.2's "the cell's decoded bytes" under-specifies the element normalization moczarr's merged verifier pins — and §6.2 leans on exactly that gap

The keys, the u64-little-endian length prefix, the flat-C-order iteration, and the sorted-`\n`-joined-hex combined hash all match `espg/moczarr` `src/moczarr/stats.py` (`hash_arrays`, `combined_hash`) verbatim — I diffed them. One piece didn't come across: moczarr's `_element_bytes` (lines 228–253) is a normative element→bytes function, not a cast:

```python
    if element is None:
        return b""  # an unwritten vlen cell can decode as None, not b""
    if isinstance(element, bytes | bytearray | memoryview):
        return bytes(element)
    if isinstance(element, str):
        return element.encode()
    if isinstance(element, np.ndarray):
        values = np.ascontiguousarray(element)
        if values.dtype.byteorder == ">":
            values = values.astype(values.dtype.newbyteorder("<"))
        return values.tobytes()
    raise ValueError(...)   # "a digest that is silently wrong is worse here than no digest at all"
```

The ndarray branch exists specifically "for the typed `vlen-array<T>` / vlen-utf8 futures" — i.e. for §6. And §6.2 asserts "the §5 O11 vlen recipe is unaffected (it hashes decoded payload bytes, which are identical by construction)", which is only true if the spec says what this function says: a `/2` cell decodes to an **ndarray**, and its bytes are defined as C-contiguous, little-endian normalized. This repo's two copies (`_o11_hashes`, `_hash_leaf`) do `bytes(element)` with no contiguity or byte-order normalization and no raise, so they already diverge from the merged verifier on the `/2` path §6.2 promises is unaffected.

Add to §5.2: `None`/unwritten ⇒ zero-length; `bytes` ⇒ as-is; a typed (`/2`) ndarray element ⇒ C-contiguous little-endian bytes at the declared dtype; anything else ⇒ the recipe does not apply and a verifier MUST raise rather than hash.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+numbers cannot carry `uint64` faithfully), the per-stratum exact counts,
+and the §5 O11 `content_hashes` (per-array + combined). Expected values
+were computed from the generator's *inputs* — the arrays handed to the
+writers — never read back through a zagg reader, so the reader is pinned,
+not self-certified.
+
+**Conformance criteria for an external reader**: decode every ragged array
+per §1–§2 and the composition array per §3, reproducing the expected
+decoded values exactly (byte-exact float32/uint64 — no tolerance), and
+reproduce `content_hashes` per §5. zagg's own suite additionally decodes
```
espg, line 894: 🤖 *from Claude (review)*

**low-medium** — two of the fixtures' recorded expectations are never asserted against the store, and the §1.5 geometry a reader "MUST read either" of has no fixture at all

`*.expected.json` records `"morton"` and `"count"` for every populated cell, and the §7 criteria tell an external reader to "reproduc[e] the expected decoded values exactly". zagg's own suite never opens either array:

```
$ grep -n "morton\|count" tests/test_spec_conformance.py
224:            counts = {"h_tdigest": cell["count"]}          # reads the JSON, not the store
228:                    "h_tdigest_noise": cell["count"] - cell["n_signal"],
291:        (golden,) = [c for c in exp["cells"] if c["n_signal"] == 1 and c["count"] == 1]
351:        assert attrs["morton_hive_commit"]["complete"] is True
357:        manifest = json.loads(...)
```

So for `minimal/` — two data arrays plus `morton` — only `h_tdigest` is pinned. A writer change to the `morton` coordinate (ordering, dtype, nested-vs-ring) leaves the fixture's `morton` strings wrong and the suite green, while moczarr, whose whole addressing path keys on `morton`, breaks. Two direct asserts fix it.

Separately: §1.5 declares two geometries and says "a single reader code path **MUST** read either", but both fixtures are sharded (`sharded: True` in `_config()`). The 1-GET per-inner-chunk geometry — the one an external reader is most likely to get wrong, since the framing is identical but the object boundary isn't — has no golden. `tests/test_processing.py::TestRaggedVlenLayout` covers it zagg-side; §7's fixtures are what moczarr vendors. Either add a third tiny unsharded fixture, or scope §7's conformance claim to the sharded geometry explicitly.

`test_means_sorted_and_weights_are_exact_counts` has the same shape — it asserts §2.1 over `expected.json` only, under a class docstring that says "through the shipping reader". It's transitively sound via `test_read_cell_matches_expected`; the docstring just overstates it.

tools/generate_spec_fixtures.py
```diff
@@ -0,0 +1,356 @@
...
+            for element in values.ravel(order="C"):
+                payload = b"" if element is None else bytes(element)
+                digest.update(len(payload).to_bytes(8, "little"))
+                digest.update(payload)
+            hashes[key] = digest.hexdigest()
+            continue
+        if values.dtype.byteorder == ">":
+            values = values.astype(values.dtype.newbyteorder("<"))
+        hashes[key] = hashlib.sha256(values.tobytes()).hexdigest()
+    combined = hashlib.sha256("\n".join(sorted(hashes.values())).encode()).hexdigest()
```
espg, line 305: 🤖 *from Claude (review)*

**low** — regeneration is not a clean no-op — the `content_hashes.arrays` key order churns because `group.members()` order is nondeterministic

Ran the determinism claim:

```
$ uv run python tools/generate_spec_fixtures.py && git status --short tests/data/spec
 M tests/data/spec/kitchen_sink.expected.json
 M tests/data/spec/kitchen_sink/1/1/2/1/3/11213.zarr/zarr.json
 M tests/data/spec/kitchen_sink/morton_hive.json
 M tests/data/spec/minimal/1/1/2/1/3/11213.zarr/zarr.json
 M tests/data/spec/minimal/morton_hive.json
```

Good news first: every `c/0` chunk object is **byte-identical** across regeneration, so §1.4/§1.5's golden framing and zstd(level=3) determinism hold on this platform. The `zarr.json` / `morton_hive.json` churn is the commit stamps, which §7 already discloses.

The `kitchen_sink.expected.json` diff is the unexpected one — and it is *only* key reordering; all seven digests and `combined` are identical. `group.members(max_depth=None)` yields concurrently, so the insertion order into `hashes` varies per run. Two consequences: every future regeneration produces a noisy 4-line diff that reviewers have to read to confirm it's benign, and §7's "Regenerating the fixtures reproduces the same logical values" is doing more work than it looks like. `dict(sorted(hashes.items()))` here (or `sort_keys=True` on the dump) makes regeneration diff-clean for the one file that matters.

Incidental confirmation worth an informative sentence in §5.3: because `combined_hash` sorts the *digests*, the combined value is immune to this enumeration order — a real robustness property of the moczarr recipe.

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+{group}/morton              <- per-cell uint64 morton coordinate (zagg's standard
+                               HEALPix coordinate array; the chunk-identity source)
+```
+
+- Each populated cell's value MUST be the raw **little-endian** bytes of an
+  `(n, *inner_shape)` array (`n` varies per cell — e.g. `(k_centroids, 2)` for
+  a t-digest whose `inner_shape` is `(2,)`), C-order, in the declared element
+  dtype, independent of the producing machine.
+- Empty cells keep the `b""` fill (`fill_value: ""`); an inner chunk with no
+  ragged data MUST be omitted from disk entirely — the same sub-shard
```

**low** — "MUST be omitted from disk entirely" is only literally true in one of the two §1.5 geometries

In the sharded geometry — "every hive leaf and the sharded flat path", i.e. the default and the only one the §7 fixtures cover — an empty inner chunk is *not* omitted from disk; the shard object still contains its index slot, marked with the `2^64 - 1` sentinel. §1.5 says this correctly. §1.1 stating it as a flat MUST reads, in the section a reader implements first, as "the object is absent", which is the wrong recipe for the default geometry. Suggest "MUST be absent from the store — omitted as an object in the per-inner-chunk geometry, marked with the §1.5 sentinel when sharded".

docs/specification.md
```diff
@@ -0,0 +1,737 @@
...
+within `O(n/510)`. The `n` inputs come from the `of` digests' total weights
+(§3.3).
+
+## 4. Pyramid / overview declarations
+
+**Status: ratified design; implementation in flight
+([#201](https://github.com/englacial/zagg/issues/201)).** The decisions this
+section records are ratified (D11/D22–D24 in
+[`design/sparse_coverage.md`](design/sparse_coverage.md) and the
+[#201 rulings](https://github.com/englacial/zagg/issues/201)); the grammar
```

**low** — §4 calls the rulings "ratified" but cites the issue thread root, not the ruling permalinks

CLAUDE.md §6: "When referencing a discussion, **link the specific comment's permalink**, not just the thread." §4's status line and §4.5's `summarize` bullet both link bare `issues/201`. The section's normative content traces to three specific comments, and a future reader (or the #344 fold) has no way to check the spec against them without re-reading six comments:

- [`#issuecomment-5025459421`](https://github.com/englacial/zagg/issues/201#issuecomment-5025459421) — `all.zarr` + `role: overview` "confirmed"; display schedule "every 2" (the §4.5 `spacing` default)
- [`#issuecomment-5025509889`](https://github.com/englacial/zagg/issues/201#issuecomment-5025509889) — the D24 `none`-class ruling: per-field exclusion default, declared derived summary as the opt-in, "never the semantic core"
- [`#issuecomment-5025519604`](https://github.com/englacial/zagg/issues/201#issuecomment-5025519604) — the A/B/C/D option space; C (seeded subsampling) recorded as an espg-flagged opt-in phase, D rejected

I checked §4 against all three and the grammar matches. One nuance worth a word: §4.5 says seeded subsampling "is deferred", which matches the ruling comment, but the option-space comment records espg calling it "compelling — worth a phase as opt-in" (#344's Phase F). "Deferred as an opt-in phase, not rejected" is the accurate compression.

docs/signal_strata.md
```diff
@@ -38,31 +38,29 @@ via the per-variable `column` selector (`{path: ..., column: k}` — the
...
+ones, and one lane layout serves every product.
 
 ## Merge law
 
 `merge_composition(word_a, n_a, word_b, n_b)` folds lanes as the
-digest-weighted mean `(n_a·lane_a + n_b·lane_b)/(n_a + n_b)`, re-quantized
-with the same presence floor — an order-independent monoid whose `n` inputs
-come from the signal digests' total weights.
+digest-weighted mean, re-quantized with the same presence floor — an
+order-independent monoid whose `n` inputs come from the signal digests' total
```
espg: 🤖 *from Claude (review)*

**low** — the narrative companion asserts a stronger merge property than the spec it defers to

This line (rewritten in this PR) keeps "an **order-independent monoid**", while `specification.md` §3.4 — now the normative home — says only "symmetric and, **up to the bounded re-quantization error**, associative … fold order never affects presence, and affects counts only within `O(n/510)`". The spec's version is the true one; `merge_composition` re-quantizes at every fold, so `(a⊕b)⊕c` and `a⊕(b⊕c)` can differ by a lane count. `stats/composition.py`'s module docstring has the same overclaim ("order-independent monoid over `(word, n)` pairs") though its `merge_composition` docstring gets it right.

The page's own "Where the two disagree, the spec wins" banner covers the collision formally, but a reader who lands here first walks away thinking composition folds are byte-stable across orders — which is the one thing §3.4 is careful not to promise. "symmetric, and associative up to bounded re-quantization" costs four words.

espg:
🤖 *from Claude*

Spec-first discrepancy note from the #201 implementation (PR #344), per the divergences-resolve-on-§4-first rule stated in §4's status paragraph.

**What diverged:** an adversarial-review finding on PR #344 ([thread](https://github.com/englacial/zagg/pull/344#discussion_r3680582363)) showed the exact-class fold is unconditionally nan-*skipping*: with `fill_value: NaN`, a leaf cell's data-NaN and fill-NaN are the same bytes, so the NaN-propagating reductions (`min`/`max`/`sum` as opposed to their `nan*` forms) are unrecoverable at the leaf seam. The fold fix (`d211fff` on PR #344) keeps those functions in the `exact` class but **declares the policy**: exact field entries now carry `"nan_policy": "skip"` in both the manifest `pyramid` block (§4.5 grammar) and each overview's `zagg_overview.fields` entries (§4.3 grammar).

**What §4 should say:** the §4.3 and §4.5 example blocks predate this and show exact entries without `nan_policy`. Suggested amendment: add `"nan_policy": "skip"` to the exact-field entries in both examples, with one sentence in §4.5's `fields` bullet — exact entries record the NaN handling actually applied (`skip`: NaN is treated as missing, never propagated), because a NaN fill makes data-NaN indistinguishable from empty at the leaves.

The alternative (reclassifying plain `min`/`max`/`sum` out of `exact` when NaN-propagation can differ) is left standing on the PR #344 thread for a ruling; if that lands instead, §4 needs no change.

docs/specification.md
espg: 🤖 *from Claude*

Folded in b6c1506 — took both ways out, so the block now has the same enforcement posture as `zagg-ragged/1`.

Writer-side derivation (option (a)): `stats/composition.py` gains `COMPOSITION_ATTR` + `composition_attrs_block()`, and `grids/base.py:apply_field_attrs` runs every declared `composition` block through it, so `spec`/`lanes` in the store come from `COMPOSITION_SPEC`/`LANES` regardless of what the config typed (or omitted):

```python
    block = merged.get(COMPOSITION_ATTR)
    if isinstance(block, dict):
        merged[COMPOSITION_ATTR] = composition_attrs_block(block)
```

Config validation (option (b)): `_validate_composition_attrs` now rejects a declared `spec` that isn't `COMPOSITION_SPEC` and a declared `lanes` that isn't `list(LANES)`. Posture on permuted lanes is **reject, not honour** — `pack_composition` packs lane `i` at bits `8i..8i+7` in fixed `LANES` order and takes no lane-order parameter, so an author permutation could not be honoured; the spec text now states lane order is not a product knob and that a writer MUST reject a permuted/truncated/renamed value.

§3.3 records the split: `spec`/`lanes` writer-stamped (disagreement rejected at load), `of`/`threshold` author-declared and cross-checked. The constants are no longer dead outside their module.

Fixture circularity: `test_attrs_block` keeps its literal, and a new `test_attrs_block_binds_to_the_writer_constants` asserts the committed block's `spec`/`lanes` equal the module constants — so the generator's hand-typed copy is now machine-checked both at config load and against the store.

Tests: `test_composition_spec_must_match_the_convention`, `test_composition_lanes_must_match_lane_order`, `test_composition_declared_spec_and_lanes_accepted_when_they_agree` (test_config.py); `test_composition_lanes_stamped_when_config_omits_them` plus an updated `test_scalar_field_attrs` (test_grids.py).

docs/specification.md

Folded in f8e84f2, the moczarr `ba804e6` way: broke the circle with frozen literals rather than only narrowing the claim.

(1) `tests/test_spec_conformance.py` gains `FROZEN_COMBINED` (both fixtures' combined digests) and `FROZEN_ARRAYS` (one per-array digest per §5.2 element kind: a vlen payload array, a vlen `uint64` locations sibling, and two fixed-width arrays), asserted by `test_frozen_digests_pin_the_recipe`. A prefix-width / joiner / key-set change made consistently in both copies now fails that test.

(2) Added a genuinely independent leg: `test_vlen_digest_from_shard_bytes_only` recomputes each ragged array's digest from the **shard objects alone** — the §1.5 index recipe plus the §1.4 u32 framing (`struct` + `Zstd` + `hashlib`, no zarr, no helper shared with the generator) — with an absent inner chunk contributing its cells' zero lengths, which is the part that makes the digest cover the cell grid. The §1.4/§1.5 decode helpers moved to module level so this and `TestWireFraming` share one spec-text implementation.

(3) §7's claim is corrected, not re-broadened: the *decoded values* come from the generator's inputs; the `content_hashes` are stated to be computed from the written leaf (they could not be anything else) and pinned instead by the frozen literals plus the shard-bytes recomputation — with the espg/moczarr#23 point made explicit, that this is the only mechanism that can catch a zagg↔moczarr divergence.

docs/specification.md

Folded in f46415f — made it a MUST and backed it with validation, per CLAUDE.md §4.

§3 now reads: an empty signal stratum packs to `0`, and a composition array **MUST** declare `fill_value: 0` so an unwritten cell decodes to the same word, with your `fill_value: 1` ⇒ lanes `[1,0,0,0,0,0,0,0]` ⇒ spurious `land` presence spelled out as the reason.

`_validate_composition_attrs` gained the check next to the existing `of`/`threshold` cross-checks (it fires for any field carrying a `composition` block, and an *absent* `fill_value` is rejected too — the template's default is `"NaN"`, which is not a legal uint64 fill anyway):

```python
    fill = meta.get("fill_value")
    if fill != 0:
        raise ValueError(...)
```

Test: `test_composition_fill_value_must_be_zero` covers both `fill_value: 1` and absent. The shipped `atl03_tdigest_strata_healpix.yaml` and both fixture configs already declare `0`, so nothing in tree changes behavior — it is now enforced rather than coincidental.

docs/specification.md
espg: 🤖 *from Claude*

Folded in 550675a — took the "state the consequence explicitly" branch rather than rescoping the key set, because both shipped implementations enumerate (moczarr `stats.py:304` and both copies here), and narrowing the spec to template-declared arrays would make the merged verifier non-conforming.

§5.1 now says the scope is **discovery-based** and adds the normative classification rule the silence was leaving open: a **key-set difference is a distinct outcome from a digest mismatch**. A verifier MUST compare the per-array map first and report extra/missing keys as their own outcome ("extra array present" / "array missing"), reserving "mismatch" for a differing digest on a *shared* key — which lines up with §5.3's "nothing recorded ≠ tampered" posture instead of colliding with it. Debris classes are named (issue #341 Bug A, the issue #327 `.zarr.status/` prefixes) with the note that debris changes `combined`, so a `combined`-only comparison is what misreports.

Also recorded that hashing is the **only** path that enumerates: the read and fold paths open leaf arrays by name (no member enumeration, no LIST — #344), so debris is inert everywhere else. That keeps #344's design point visible in the spec rather than in tension with it.

docs/specification.md
espg: 🤖 *from Claude*

Folded in e83ab2e — bounded the claim rather than dropping it.

§2.1 now requires `sum(weights)` to equal the exact observation count **while that count is `<= 2^24` (16,777,216)**, and states what a reader gets above the bound: the nearest float32 values to the true counts, so count recovery (the §3.3 `N_signal` path) is exact at or below the bound and rounded above it. The sentence also points at where the bound actually bites (coarse overview orders, §4.4) instead of leaving it implicit.

§4.4's inheritance clause got the matching bound, so the section that says "§1–§3 apply unchanged" no longer silently re-imposes an unsatisfiable MUST on the arrays most likely to exceed it:

> §1–§3 of this page apply to overview arrays unchanged, **including §2.1's float32 exactness bound**: a coarse overview cell can pool more than `2^24` observations, and there `sum(weights)` is the nearest float32 to the true count rather than the count itself

docs/specification.md

Folded in 90affbe. Reproduced your check on this branch before editing:

```
$ uv run python -c "..."
11213 4
-311 2
-3111 3
```

The example's `node` is now `"-3111"` (order 3, matching the `order: 3` / `cell_order: 11` line), and the prose gained the rule an implementer would otherwise have to infer from the example:

> mortie's decimal grammar puts one digit per order after the base-cell digit, so `order` is always `len(digits) - 1` for the node string (`"-3111"` is order 3; the leading `-` marks a southern base cell and is not a digit)

Everything else in the example was already self-consistent, so nothing else moved.

docs/specification.md
espg: 🤖 *from Claude*

Folded in 7031ac8 — resolved spec-first, and the resolution ratifies the shape `build_pyramid_block` already writes, so #344 needs no change.

§4.5 now spells the declared-off form out as normative, with the reader binding rule that removes the KeyError:

> With the pyramid knob off the block is exactly `"pyramid": {"spec": "zagg-pyramid/1", "overview": {"orders": []}}` — `spacing`, `all_time`, `fields`, and `summarize` are **absent**, not empty. A reader MUST branch on `orders` first: an empty `orders` (or no `pyramid` block at all — pre-pyramid manifests) means no overview family exists and no other key of the block may be assumed. When `orders` is **non-empty**, `spacing`, `all_time`, and `fields` MUST all be present (`summarize` stays optional), so the zero-open field query of §4.4 is well-defined exactly when there is something to query.

That makes `orders` the only unconditionally bindable key and puts the bug where it belongs — a reader that binds `pyramid.overview.fields` without checking `orders` is the defect, not the writer. I checked the landed `sweep_overview.py:build_pyramid_block` on `claude/201-pyramid-sweep` against this text: the off branch returns `{"spec": PYRAMID_SPEC, "overview": {"orders": []}}` and the on branch always populates `spacing`/`orders`/`all_time`/`fields` (with `summarize` conditional), so it conforms as landed — no #344 comment needed for this one.

docs/specification.md

Folded in e0e94c9 — gave the rule the carve-out rather than leaving two opposed normative statements.

The Conformance bullet now reads that a convention is signaled "by a `spec` marker … **or, where a section names it, by the array's registered zarr data type**", and a second bullet bounds marker-absence:

> Marker-*absence* is legal only where a section names the data type that replaces the marker. Today there is exactly one such carve-out: the typed `vlen-ndarray` dtype **is** the `zagg-ragged/2` signal, and the `ragged` attrs marker is deliberately retired (not written) on those arrays (§1.6/§6.1) — so a reader MUST NOT treat a missing marker there as an unknown revision. Everywhere else absence is a hard failure: a `variable_length_bytes` array with no `ragged` block is not signaled at all and MUST be refused (§1.2).

So the `/2` implementer reading §6 first now finds the exemption stated in the rule that would otherwise reject their arrays, and the §1.2 refuse path stays sharp for the un-signaled case.

docs/specification.md
espg: 🤖 *from Claude*

Folded in 3e89901 — ruled for your option (2), and recorded *why*, so the `/2` PR inherits the reason and not just the rule.

§6.3's located bullet now says the sibling binding lives under a **new top-level attrs key**, not a residual `ragged` block, and that "retired" is literal (no `ragged` key written on a `/2` array). The rationale is the one you traced: a `ragged` block carrying only `locations` is, by §1.2's own words, an array with no well-formed `element` declaration, which a `/1`-only reader MUST refuse with "not a `zagg-ragged/1` array" — the misleading path instead of the actionable "install `zarr-vlen-ndarray`" one in bullet 4. Since a `/1`-only reader's zarr stack cannot open the `vlen-ndarray` dtype at all, option (2) puts it on the right error path by construction.

Also pinned the reservation half you asked about: the `ragged` key stays **reserved but unwritten** under `/2` (config-declared attrs still MUST NOT shadow it), and the new key's exact name remains for the `/2` implementation PR to pin in that section before any `/2` store exists.

docs/specification.md
espg: 🤖 *from Claude*

Folded in 8aa0f70 — resolved spec-first in the "materialized only" direction, which is the half of §4.4 that was already unambiguous (`none` fields are **absent** from the zarr).

§4.3 now states it outright:

> **`fields` enumerates the materialized fields only** — exactly the fields present as arrays in *this* overview … A `none`-class field is absent from the zarr (§4.4) and so MUST be absent from this map: its recorded absence lives in the manifest's `pyramid.overview.fields` (§4.5), which is the map that enumerates **every** declared field.

plus the reader consequence (the attrs map may be treated as the overview's variable list; every array it names MUST be openable; cross-checking it against the arrays present is a valid integrity check). §4.5 got the complementary sentence — a `none` entry carries **`class` only**, no `method` and no dtype/shape, precisely because stamping a default fold method would declare a t-digest array that does not exist — and closes with "this map is the **all-fields** view; `zagg_overview.fields` (§4.3) is the materialized subset".

On the code half: the snippet you quoted has since moved on `claude/201-pyramid-sweep`. `_write_overview` now stamps `{n: _field_provenance(m) for n, m in fields.items()}`, and that `fields` is the dict filtered upstream to composable classes:

```python
    fields = {
        n: dict(m)
        for n, m in (decl.get("fields") or {}).items()
        if isinstance(m, dict) and m.get("class") in ("exact", "approximate")
    }
```

so `none` fields no longer reach the attrs map and the `TDIGEST_LAW` default can no longer land on one. #344 conforms as landed, so no conformance note was needed there. One grammar alignment I did make while here: `_field_provenance` adds `nan_policy` to `exact` entries, so both §4.3 and §4.5 examples now show `"nan_policy": "skip"` and the text says entries MAY carry further fold provenance and readers MUST tolerate additional keys — otherwise the spec would have read as forbidding what the ratified implementation writes.

docs/specification.md
espg: 🤖 *from Claude*

Folded in d99a2e2 — ported the normalization into §5.2 as a normative table, and brought both in-tree copies up to it.

§5.2 now carries the element → bytes contract: `None` (an unwritten vlen cell may decode as `None`, not `b""`) ⇒ zero-length; `bytes`/`bytearray`/`memoryview` ⇒ as-is; `str` ⇒ UTF-8; a typed `/2` ndarray cell ⇒ **C-contiguous, little-endian** bytes at the declared element dtype; anything else ⇒ the recipe does not apply and a verifier MUST raise rather than hash. The "silently wrong digest is worse than no digest" reason is stated, as is the `None`/`b""` equivalence and why that is intended.

§6.2's promise is now grounded rather than assumed — it says the vlen recipe is unaffected "*this is exactly what §5.2's element→bytes normalization buys: a `/2` cell decodes to an ndarray, whose C-contiguous little-endian bytes are the `/1` cell's bytes*".

Code: `tools/generate_spec_fixtures.py` gained `_element_bytes` and `tests/test_spec_conformance.py` gained `TestContentHashes._element_bytes` (deliberately still two independent copies, mirroring the two-implementation posture), both implementing the full table including the raise. `test_element_bytes_normalization` covers `None`, `bytearray`/`memoryview`, `str`, a big-endian F-order ndarray normalizing to the little-endian C-order buffer, and the raise on an unsupported type.

docs/specification.md
espg: 🤖 *from Claude*

Folded in 590480e — took the asserts, and scoped §7's geometry claim rather than committing a third store.

New `TestCoordinateAndDense` reads both arrays off the store: `test_morton_coordinate_matches_expected` (dtype `uint64`, full 16-cell span, every populated cell's word equal to the recorded decimal string), `test_count_field_matches_expected` (dtype `int32`, every cell, populated and fill), and `test_morton_ascends_per_written_chunk_and_is_fill_where_absent`. That last one pinned something worth having in the spec: the coordinate array gets the same §1.5 sub-shard sparsity, so chunk ordinal 2's four `morton` slots are the `0` fill —

```
[1233986297899515910, …, 1235956622736490502, 0, 0, 0, 0, 1237363997620043782, …]
```

— which §7 now says explicitly ("a reader MUST NOT assume the coordinate is dense across a shard").

On the geometry gap: rather than add a third fixture, §7's conformance claim is now **scoped to the sharded geometry**, with the unsharded one named as pinned zagg-side by `tests/test_processing.py::TestRaggedVlenLayout` and the failure mode spelled out — "a reader that derives the stored span from the array's own metadata, as §1.5 requires, reads both from one code path; a reader that hard-codes the shard-index suffix passes §7 and still fails on an unsharded store". If you want the third golden committed instead, that is a small follow-up and I would rather it be your call than mine, since it adds a committed store to the contract.

`test_means_sorted_and_weights_are_exact_counts` keeps its `expected.json` scope with a comment recording that it is bound to the store transitively through `test_read_cell_matches_expected` — the docstring no longer implies more than the test does.

tools/generate_spec_fixtures.py
espg: 🤖 *from Claude*

Folded in 053161b. `_o11_hashes` now records the map key-sorted:

```python
    return {"arrays": dict(sorted(hashes.items())), "combined": combined}
```

Verified the fix does exactly what you measured and nothing more: regenerated into a scratch dir and compared the two `*.expected.json` against the committed ones — both `dict`-equal (all seven digests, `combined`, every cell), differing only in `arrays` key order. So the committed files were updated with the sorted rendering and the stores were left untouched, which keeps the commit a 6-line key reorder instead of a fixture-wide stamp churn. `test_recorded_arrays_map_is_key_sorted` guards it for both fixtures.

Took the informative sentence too — §5.3 now records the robustness property you spotted:

> Because it sorts the *digests*, `combined` is immune to the order in which the §5.1 enumeration happened to yield arrays — a real robustness property, and the reason two implementations agree without agreeing on traversal order. The recorded `arrays` map is a different matter: a writer SHOULD record it key-sorted so a regenerated record diffs cleanly.

docs/specification.md

Folded in f25b46b, close to your suggested wording:

> an inner chunk with no ragged data MUST be **absent from the store** — omitted as an object in the per-inner-chunk geometry, and marked absent with the §1.5 sentinel inside the shard object when sharded (the default). Either way object size scales with populated chunks only — the same sub-shard sparsity the dense arrays get.

So §1.1 no longer reads as "the object is absent" in the geometry that is both the default and the only one the §7 fixtures cover, and it points at §1.5 for the sentinel rather than restating it.

docs/specification.md

Folded in 9bfe363. §4's status line now cites the three ruling permalinks instead of the thread root (all three fetched and re-read before linking):

- [`#issuecomment-5025459421`](https://github.com/englacial/zagg/issues/201#issuecomment-5025459421) — `all.zarr` + `role: overview` confirmed; display schedule every 2 orders
- [`#issuecomment-5025509889`](https://github.com/englacial/zagg/issues/201#issuecomment-5025509889) — the D24 `none`-class ruling
- [`#issuecomment-5025519604`](https://github.com/englacial/zagg/issues/201#issuecomment-5025519604) — the A/B/C/D option space

and §4.5's `summarize` bullet now points at the ruling comment rather than the thread.

Took the nuance too — the option-space comment records "*espg: compelling — worth a phase as opt-in*", so the bullet now reads "deterministic seeded subsampling is **deferred as an opt-in phase, not rejected** — it shares this block's declaration grammar and stays declared/never-default when it lands", with the option-space permalink attached. Roster concatenation stays recorded as rejected.

docs/signal_strata.md
espg: 🤖 *from Claude*

Folded in 555874b — the four words, in both places.

`docs/signal_strata.md`: "symmetric, and associative **up to the bounded re-quantization error** (fold order never affects presence; it affects counts only within `O(n/510)`, so folds are not byte-stable across orders)".

`stats/composition.py`'s module docstring had the same overclaim and got the same treatment ("Merge law over `(word, n)` pairs — symmetric, and associative up to the bounded re-quantization error…"). `merge_composition`'s own docstring already stated it correctly, so it stands as-is, and the narrative page and the spec's §3.4 now say the same thing without relying on the "where the two disagree, the spec wins" banner.

espg:
🤖 *from Claude*

Scope addition from espg's review (in-session, 2026-07-31): the reader-consumption text should describe a **third access mode** — the sub-leaf subtree read — alongside the whole-store sweep and single-cell random access.

Concretely, in the `docs/ragged_layout.md` "consumes the layout two ways" passage (and the spec's reader-expectation text where it summarizes access modes):

1. State the layout consequence explicitly and citably: **a morton subtree of the cells axis is one contiguous cell span** — the cells axis is nested-ordered, so the order-`k` subtree below an ancestor at rank `r` occupies exactly `[r·4^d, (r+1)·4^d)`. This already follows from the nested-order requirement, but readers should be able to rely on it by citation rather than re-deriving it.
2. Note the access shape it licenses: on a sharded store a contiguous span read costs the shard-index suffix plus only the covering inner chunks — `read_cell`'s 2-GET pattern generalized to a span — so "everything below this morton node" never requires the whole-array sweep.

The zagg reader implementation is issue #351 (with a moczarr counterpart on the moczarr side); this PR only owes the spec/narrative statement, so the text can reference the span property as normative layout fact and keep the reader function itself informative/forward-looking.

Also cross-linking two issues filed out of the same review pass, since they touch the spec's neighborhood: #352 (per-field pyramid selection — §4's declaration grammar has room for it) and #353 (sweep telemetry, prerequisite for the pyramid benchmark runs). Neither changes this PR's scope beyond item (1)–(2) above; #342 has moved to `plan` for the §5 writer side.

espg:
🤖 *from Claude*

The scope addition above is now on the branch: 06dd3dd — spec §1.5 gains the normative **subtree-spans** clause (cells axis MUST be canonically nested-ordered, stated citably; the contiguous-span consequence and the MAY-read it licenses), and `docs/ragged_layout.md`'s read-paths section gains the third access mode as *(planned — issue #351)*. Text-only; no fixture bytes change, so the moczarr-side pins (`555874b` composition fixture, `166e908` spec vectors on moczarr PR #27) verify unchanged — they should simply re-confirm against the merge sha per their own recorded process.

PR #349, opened 2026-07-30
Module splits for the &gt;1000-line offenders (issue #330 PR-A)

espg:
Refs #330. PR-A of the consolidated two-PR plan ratified in https://github.com/englacial/zagg/issues/330#issuecomment-5124476349 ("Phase A can be its own PR, with a module per implementation commit") and detailed in https://github.com/englacial/zagg/issues/330#issuecomment-5124503244.

## What this is

The five `src/zagg` modules that exceed the CLAUDE.md §4 ~1000-line limit, split into packages — **one phase commit per module, every one a pure move behind a re-export facade**. No behavior change, no store-schema change, `semantic_hash` untouched.

**This PR does not convert any docstrings.** The numpydoc sweep is PR-B, which the ruling blocks on this PR merging — split first so the conversion never rewrites text that is about to relocate, and so `git log --follow` stays intact through a mechanical diff.

## Phases

- [x] **1 — `hive.py` → `hive/` package** (`f897e27`)
- [x] **2 — `processing/raster.py` → `processing/raster/` package** (`93d65b1`)
- [x] **3 — `sweep.py` → sibling modules** (`27cd554`)
- [x] **4 — `config.py` → `config/` package** (`d3375c8`)
- [ ] **5 — `runner.py` → `runner/` package — BLOCKED, needs a decision.** See "Questions for review" (1) and the [analysis comment](https://github.com/englacial/zagg/pull/349#issuecomment-5129979059). Phase 5 was written in the plan as "held until #333 merges"; #333 *has* merged, so that is no longer the blocker — the blocker is that the split relocates 34 of the 39 `zagg.runner` attributes the test suite monkeypatches, across 139 patch sites, with a silent-no-op failure mode.

Plus `4ab54c7`, a self-review fix reverting an unrelated `tests/data/benchmark/README.md` reformat my format command had swept in.

## Before / after

| module | before | after (largest resulting module) |
|---|---|---|
| `hive.py` | 1,393 | `hive/coverage.py` 550 |
| `processing/raster.py` | 1,266 | `processing/raster/__init__.py` 523 |
| `sweep.py` | 1,034 | `sweep_families.py` 498 |
| `config.py` | 2,747 | `config/validate.py` 722 |
| `runner.py` | 5,397 | *not split — blocked* |

Every module produced by phases 1-4 is under 1000 lines, asserted by a test rather than by inspection (`test_every_split_module_is_under_the_line_limit`). Headroom was deliberate: PR-B adds docstring lines to all of these, so no new module was left near the ceiling.

### Phase 1 — `hive.py` → `zagg/hive/`

Seam per the ratified plan: *path/manifest arithmetic vs coverage/rollup writers*.

| module | lines | holds |
|---|---|---|
| `hive/layout.py` | 183 | spec versions, the D19 product-name grammar, D2/D3 leaf-path arithmetic, the D5 node invariant, and the two shared helpers (`_read_json`, `_utcnow`) both halves use |
| `hive/manifest.py` | 382 | the §3/D6 `morton_hive.json` builder, the frozen-key resume guard, the D19 semantic core, root-form discrimination |
| `hive/coverage.py` | 550 | the D4 commit stamp, §4 tier-0 coverage + its bitmap sidecar (issue #200 phase 2), the store-root ranges MOC rollup (issue #200 phase 3) |
| `hive/__init__.py` | 414 | design docstring, facade re-exports, and the per-shard write path |

Dependency direction is one-way — `layout` ← `manifest` ← `coverage` — so no cycle and no lazy-import scaffolding.

`process_and_write_hive` stays on the facade, mirroring the plan's runner design ("facade keeps `agg()` + the local path"), because `tests/test_hive.py` patches the stamp helper on that module to assert leaf write ORDER:

```python
monkeypatch.setattr(hive, "stamp_commit", rec("stamp", hive.stamp_commit))
```

That only resolves if the call site reads `stamp_commit` out of the `zagg.hive` namespace.

### Phase 2 — `processing/raster.py` → `zagg/processing/raster/`

Seam per the plan: *decode engine / band-map+template / write path*.

| module | lines | holds |
|---|---|---|
| `raster/decode.py` | 541 | the async-tiff engine: process-wide obspec store cache, IFD georeferencing, per-asset tile fan-out, acquisition-time index |
| `raster/template.py` | 200 | the flat `(time, cell)` template, the hive leaf spec, the band-to-array map |
| `raster/write.py` | 83 | slab + coordinate writers (plain Zarr region assignment) |
| `raster/__init__.py` | 523 | docstring, facade re-exports, and the shard orchestration (`process_raster_shard`, `process_and_write_raster_hive`) |

The orchestration stays on the facade for the same monkeypatch-resolution reason: `tests/test_raster_pipeline.py` patches `sample_item_async`, `write_raster_leaf_slab` and even `time` on `zagg.processing.raster` and expects `process_raster_shard` / `process_and_write_raster_hive` to see them. Keeping those two callers in the facade namespace makes all of those patches keep working untouched.

### Phase 3 — `sweep.py` → sibling modules (not a package)

Seam per the plan: *the #300 family seams (stats / MOC / sub-shardmap)*.

| module | lines | holds |
|---|---|---|
| `sweep_families.py` | 498 | the `SweepFamily` contract, the four D22 families, the leaf sub-map helpers, the `FAMILIES` registry |
| `sweep_rollup.py` | 203 | the family-agnostic node fold: shard/interior rollup, generation + ancestor arithmetic, rollup object read/write |
| `sweep.py` | 374 | docstring, facade re-exports, `run_sweep`, discovery, the end-of-run hook, and `main()` |

**Siblings rather than a package, deliberately.** `zagg.sweep` is a CLI entry point — `python -m zagg.sweep` is named as the regeneration backstop in `runner.py` (two places) — and `python -m` on a *package* runs `__main__.py`, not `__init__.py`, so packaging it would have silently broken the CLI. Sibling modules also match the tree's existing convention: `zagg/sweep_overview.py` is already the split-out overview family. Verified post-split:

```
$ uv run python -m zagg.sweep --help
usage: sweep.py [-h] [--families FAMILIES] ...
```

### Phase 4 — `config.py` → `zagg/config/`

Seam per the plan: *loaders+dataclass / validation / accessors*. Validation and accessors each had to divide again to fit the limit — 2,747 lines needs at least three modules and the natural seams did not land evenly.

| module | lines | holds |
|---|---|---|
| `config/base.py` | 224 | `PipelineConfig`, the `TypedDict` block shapes, filter/worker vocabularies, shared leaf helpers |
| `config/expressions.py` | 237 | filter readers, level lookups, `resolve_function`, the sandboxed expression evaluators |
| `config/accessors.py` | 669 | the whole `get_*` family |
| `config/validate_source.py` | 245 | `data_source` block validators (temporal specs, raster, collections) |
| `config/validate_output.py` | 686 | `output` block validators (store layout, windowing, chunk precompute, output kinds, pyramid) |
| `config/validate.py` | 722 | `validate_config` plus the cross-block checks (variables, filters, levels, index, composition, worker) |
| `config/__init__.py` | 188 | docstring, facade re-exports, and the three loaders |

The loaders (`load_config`, `load_config_from_dict`, `default_config`) stay on the facade because that is what makes the package a DAG: a loader calls `validate_config`, which reads the dataclass the loader builds. Putting them in `base.py` would have forced a `base` to `validate` cycle and a lazy import inside a function body — which is not a pure move.

## Import compatibility

`tests/test_module_split_compat.py` freezes the pre-split surface of each split — the AST census of **every top-level name the module defined**, privates included — and asserts, per split:

1. every frozen name resolves on the facade;
2. `from FACADE import NAME` works for each;
3. `__all__` (where the module had one) stays within the frozen surface and every entry resolves — `zagg.hive.__all__` is byte-identical to pre-split, no widening;
4. facade names are the **same objects** as the submodules', not copies;
5. every module in the split is <= 1000 lines.

Frozen surface sizes: hive 50 names, raster 37, sweep 37, config 79.

Privates matter and are covered — `zagg/coverage.py` imports `_read_json` off `zagg.hive`; `zagg/sweep.py` and `zagg/sweep_overview.py` import `_utcnow` and the `_decimal_*` family; `tests/test_raster.py` imports `_geo_from_ifd` / `_store_and_path`. All are re-exported with the `x as x` redundant-alias marker so ruff sees a deliberate re-export rather than F401. **No `# noqa` and no `# type: ignore` anywhere in this PR.**

### Purity is verified mechanically, not by eye

Every phase was checked with an AST diff of the pre-split file against the post-split modules, comparing the exact source text of each top-level definition:

| phase | defs before -> after | text drift | new defs |
|---|---|---|---|
| 1 `hive` | 52 -> 52 | none | none |
| 2 `raster` | 37 -> 37 | none | none |
| 3 `sweep` | 35 -> 35 | none | none |
| 4 `config` | 80 -> 80 | none | none |

Not one character of any function, class or constant body changed in phases 1-4. The only name appearing in more than one module is `logger`, which each submodule now binds for itself.

### Incidental re-exports deliberately not preserved

The pre-split modules also leaked names they merely *imported*. Most (`json`, `re`, `logging`, `zarr`, `np`, …) are artifacts nothing references. Two were referenced and are handled explicitly rather than re-exported:

- `zagg.hive.union_time_range` (really `zagg.windows`') — `tests/test_hive_windows.py::TestRootTimeUnion` now imports it from `zagg.windows`, which those tests already named as the owner in their own `caplog` filter (`logger="zagg.windows"`).
- `zagg.hive.open_object_store` (really `zagg.store`') — `tests/test_raster_runner.py` **monkeypatched** it to redirect `s3://` writes. Re-exporting it would have been a trap: each submodule binds `open_object_store` at its own import, so a patch on the facade binding would silently miss the code that calls it and the test would pass while testing nothing. The test now patches each submodule binding.

Same rule applied to `zagg.processing.raster._STORE_CACHE` in phase 2 (`tests/test_raster.py` now patches `raster.decode`). The principle is recorded in `test_module_split_compat.py`'s docstring: **a facade re-export is read compat only; patch the module that owns the binding.**

## Testing

`uv sync --extra test`, then per phase: `uv run ruff check src tests`, `uv run ruff format --check src tests`, `uv run pytest -q`.

| phase | result |
|---|---|
| 1 | 3012 passed, 35 skipped |
| 2 | 3017 passed, 35 skipped |
| 3 | 3021 passed, 36 skipped |
| 4 | 3025 passed, 37 skipped |

Because the diffs are pure moves, the pre-existing suites are the real regression gate — `test_hive*`, `test_coverage*`, `test_raster*`, `test_sweep*`, `test_config`, `test_lambda_*` all pass unchanged in substance.

### Pre-existing failures on `main`, not from this PR

- `ruff check`: `N818` on `src/zagg/registry.py:64` — `UnknownCapability` wants an `Error` suffix.
- `ruff format --check`: `tests/data/benchmark/README.md`.
- `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` is env-dependent (pip resolution); it passed in this environment.

Left alone per CLAUDE.md §4. `4ab54c7` exists because my format command initially swept the README in — reverted so the diff stays scoped.

## Questions for review

**(1) Phase 5 (`runner.py`) needs a decision before it can land — this is what the `waiting` label is for.** The mechanical split is straightforward (9 modules, DAG verified). What is not straightforward: the test suite monkeypatches **39 distinct attributes of `zagg.runner` across 139 sites in 13 files**, and the split relocates the callers of **34 of those 39**. A `monkeypatch.setattr` against a name whose caller moved does not fail — it rebinds a name nothing reads, and the test goes green while asserting nothing. 17 of the 34 have callers in **two or more** resulting modules, so there is no single repoint target. Full analysis, numbers and four concrete options in [this comment](https://github.com/englacial/zagg/pull/349#issuecomment-5129979059); answering with a letter is enough to unblock.

**(2) Facade shape.** Phases 1, 2 and 4 keep the entry points (`process_and_write_hive`, `process_raster_shard`, the config loaders) on the facade rather than a pure re-export shim, for the monkeypatch-resolution and DAG reasons above. That follows the plan's runner precedent, but it means "facade" here is "entry point + re-exports". Confirm that shape, or say the word and they move to submodules with the affected tests repointed.

**(3) `__all__` left byte-identical.** `zagg.hive.__all__` omits names that are in practice public API and imported across the package — `effective_store_root`, `validate_product_name`, `product_root`, `classify_store_root`, `list_products`, `validate_manifest`, `write_semantic_core`, `PRODUCT_NAME_MAX`, `AGGREGATION_CORE_NAME`. Widening it is one line but it *is* a public-surface change, so it is out of scope for a pure-moves PR. Follow-up issue, or fold in here?

**(4) Naming collision by sight.** `zagg/hive/coverage.py` sits next to the top-level `zagg/coverage.py` (the reader-side module). Unambiguous by path, not by basename. `stamps.py` or `rollup.py` if the collision reads badly.

**(5) Stale rationale.** `zagg/coverage.py`'s docstring says it "Lives in its own module rather than `zagg.hive` to keep that file inside" the line limit — a rationale phase 1 partly retires. Left untouched as out of scope; say if it should be reworded in PR-B's `zagg/coverage.py` commit.

**(6) Process deviation.** CLAUDE.md §2's fresh-context adversarial review subagent could not run — no agent-spawning tool is exposed in this environment. Reviews on this PR are same-context self-reviews, which are weaker. Compensated with the mechanical AST purity checks above, but worth a human pass before merge. Detail in the first comment.


espg:
🤖 *from Claude*

**Process deviation to flag up front, before the phase-1 review below.**

CLAUDE.md §2 requires a fresh-context adversarial self-review after each phase push — a *separate* review subagent on an Opus-class model posting inline findings, then a *separate* fold subagent applying them. **This run has no agent-spawning tool exposed**, so neither subagent can be created. Per the CLAUDE.md preamble that is a conflict to surface rather than guess around, so: the phases proceed (the routine's own instruction is not to stop at phase 1), but the review below is a **same-context self-review, which is materially weaker than the mandated one** — it cannot catch what this context has already talked itself into. Worth a re-review before merge, or a re-run of this routine in an environment where the review subagent is available.

Compensating for the lost adversarial pass, phase 1 is verified **mechanically** rather than by reading: an AST census of the pre-split `hive.py` against the four post-split modules, comparing the exact source text of every top-level definition.

```
pre-split defs: 52   post-split defs: 52
MISSING: []
TEXT DRIFT (not a pure move): []
NEW DEFS (should be empty for a pure move): []
```

Zero characters of any function, class or constant changed — the move is provably pure, and the only name appearing twice is `logger`, which each submodule now binds for itself. That is the property the review would most want to confirm, and it is now checked by construction instead of by eye.

## Phase 1 self-review findings

**(1) Logger names changed — checked, no impact.** Warnings from `validate_manifest` and `write_root_coverage` now emit under `zagg.hive.manifest` / `zagg.hive.coverage` rather than `zagg.hive`. Every hive test asserting on those warnings uses an unfiltered `caplog.at_level("WARNING")` and reads `caplog.text` / `caplog.records`, so all still pass; and because both are children of `zagg.hive`, any `logger="zagg.hive"` filter would still capture them by propagation. No production code filters on logger name (`grep` for `getLogger`/`addFilter` across `src`).

**(2) The monkeypatch hazard is the real risk in this class of change, and it bit once.** A pure move relocates a *binding*, so any `monkeypatch.setattr(module, name, ...)` against a name the module imported keeps succeeding while silently patching nothing. `tests/test_raster_runner.py` did exactly this with `open_object_store`. It is fixed by patching each submodule binding, but the general lesson applies to phases 2-5: for each split, `grep` the tests for `setattr` against the module being split *before* moving code, not after. Doing that for phases 2-5.

**(3) `hive/coverage.py` at 550 lines is the largest new module** — comfortably under the limit, but it is the one carrying three separable concerns (leaf stamp, tier-0 bitmap, root ranges rollup). If PR-B's docstring conversion pushes it up materially, it splits again along those lines rather than growing.

**(4) Left standing for @espg** (not folded — items (1) and (2) of "Questions for review" are shape/scope decisions, not diff-scoped fixes): whether the facade should keep the write path at all, and whether `__all__` should be widened to the names that are already de-facto public. Both are recorded in the PR body.


---

espg:
🤖 *from Claude*

## Phase 5 (`runner.py`) is blocked on a decision — self-review of phases 2-4, then the question

### Phases 2-4 self-review

Same caveat as the first comment: no agent-spawning tool is exposed in this environment, so this is a same-context self-review rather than the fresh-context adversarial subagent CLAUDE.md §2 mandates. Compensated by verifying purity mechanically — an AST diff of each pre-split file against its post-split modules, comparing the exact source text of every top-level definition:

| phase | defs before → after | text drift | new defs | duplicated |
|---|---|---|---|---|
| 2 `raster` | 37 → 37 | none | none | none |
| 3 `sweep` | 35 → 35 | none | none | `logger` only |
| 4 `config` | 80 → 80 | none | none | `logger` only |

Findings folded during the phases rather than after (no separate fold subagent available either):

1. **An unrelated file rode along.** `uv run ruff format src/zagg tests` reformatted `tests/data/benchmark/README.md` — one of the pre-existing `main` failures I was explicitly not to fix — and it landed in the phase-2 commit. Reverted in `4ab54c7`. The lesson generalizes: `ruff format` on a whole tree is not diff-scoped, and a PR that claims "pure moves" has to prove it by what is *absent* from the diff too.

2. **The CLI would have broken silently in phase 3.** The obvious move was `sweep.py` → `sweep/` package, matching phases 1, 2 and 4. But `python -m` on a package runs `__main__.py`, not `__init__.py`, and `python -m zagg.sweep` is named as the regeneration backstop in `runner.py` (`runner.py:3910`, `runner.py:5094`). No test covers the CLI entry point, so packaging it would have gone green. Split into siblings instead — which also matches `zagg/sweep_overview.py`, the tree's existing precedent — and verified by hand.

3. **Phase 4 needed six modules, not three.** The ratified seam was *loaders+dataclass / validation / accessors*. Three modules cannot hold 2,747 lines under a 1,000-line ceiling, and the validators do not divide evenly: `validate_config` alone is 350 lines. Split validation by config block (`validate_source` for `data_source`, `validate_output` for `output`) and lifted the filter/expression readers out of the accessors. Largest resulting module is 722 lines, which leaves real headroom for PR-B's docstring growth. Flagging the deviation from the plan's module count explicitly rather than burying it.

4. **The monkeypatch hazard bit twice more** — `_STORE_CACHE` in phase 2, and it is what forced the facade shape in phases 1, 2 and 4. Which brings us to phase 5.

---

## Phase 5: the blocker

The mechanical split is not the problem. `runner.py` decomposes cleanly into nine modules with a verified DAG:

`common` → `dispatch` → `run_stats` → `lambda_control` → `lambda_transport` → `lambda_events` → `lambda_run` → `local` → `strategies_raster` → `strategies` → facade (`agg()`)

(nine rather than the plan's four because `RasterStrategy` is a single 924-line class and `_run_lambda` a single 835-line function — neither can be divided without editing a body, which would stop being a pure move.)

The problem is the test suite. **39 distinct attributes of `zagg.runner` are monkeypatched across 139 sites in 13 files**, and the split relocates the callers of **34 of the 39**.

This failure mode is silent. `monkeypatch.setattr(runner, "_invoke_lambda_cell", fake)` against a `_invoke_lambda_cell` whose only caller now lives in `runner/lambda_run.py` does not raise — it rebinds a name nothing reads. The test goes green having asserted nothing, and for a Lambda-dispatch suite that is exactly the coverage you cannot afford to lose quietly.

The worst cases have callers in **two or more** resulting modules, so there is no single repoint target — each patch site fans out:

| patched name | defined in | called from |
|---|---|---|
| `ThreadPoolExecutor` | *(module import)* | `lambda_events`, `lambda_run`, `local`, `strategies`, `strategies_raster` |
| `_get_function_timeout_s` | `lambda_control` | `lambda_events`, `lambda_run`, `strategies_raster` |
| `_result_fetcher` | `lambda_transport` | `lambda_events`, `lambda_run`, `strategies_raster` |
| `_resolve_function_name` | facade | `lambda_events`, `strategies`, `strategies_raster` |
| `_write_run_stats` | `run_stats` | `local`, `strategies_raster` |
| `_invoke_lambda_ping` / `_finalize` / `_sweep` / `_coverage` / `_setup_async` | `lambda_control` | `lambda_run`, `strategies_raster` |
| `open_store` | *(module import)* | `local`, `strategies_raster` |
| `process_shard`, `consolidate_metadata`, `write_dataframe_to_zarr`, `write_ragged_to_zarr` | *(module imports)* | `local` |

17 of the 34 are multi-module. Only 5 of the 39 are safe as-is (`agg`, `_process_and_write`, `_await_run_stats_object`, `_fetch_result`, `_FINALIZE_BACKOFF_S`).

Phases 1-4 hit the same hazard at a manageable scale — 5 patch sites total, each repointed with a comment naming why. At 139 sites with a silent failure mode and **no fresh-context adversarial review available to catch a mistake**, pushing on unilaterally is the wrong call. So: phases 1-4 are landed and green, phase 5 is left for a decision.

### Options

**(a) Repoint every patch site to the owning submodule.** Truthful and mechanical; multi-module names become a loop over the modules that call them (the shape already used in `tests/test_raster_runner.py` this PR). Cost: ~139 test edits, and every one that is wrong is invisible. Mitigation: land the runner split as its own set of phase commits with a guard test that asserts each patched name resolves to the *same object* on every module that calls it, so a stale target fails loudly.

**(b) Keep the heavily-patched surface on the facade.** Move only the cleanly separable pieces — `strategies_raster` (924 lines), `dispatch`, `run_stats` — and leave `_run_lambda`, `_run_local` and the `_invoke_lambda_*` family in `runner/__init__.py`. Fewest test edits, but `__init__.py` still lands around 2,800 lines: the limit is not met and phase 5's acceptance criterion fails. Would need a recorded exemption.

**(c) Split fully and convert the patches to `unittest.mock.patch` with string targets** (`patch("zagg.runner.lambda_run._invoke_lambda_cell")`), which raises `AttributeError` on a bad target instead of no-op'ing. Turns the silent failure into a loud one — the real fix for the hazard — but is a larger and more opinionated test-suite change than a "pure moves" PR should carry, and it is a testing-convention change, which per CLAUDE.md §6 is not mine to make.

**(d) Defer phase 5 to its own PR** after this one merges, so the runner split and its test rewiring are reviewable on their own rather than buried under four clean splits. The plan already anticipated sequencing pressure here (phase 5 was originally "held until #333 merges").

Leaning **(a) combined with (d)**: land phases 1-4 now, and do the runner split as a follow-on PR where (a)'s repointing plus the resolve-identity guard test is the whole diff and can be reviewed as such. That keeps this PR's "not one character changed" property intact and gives the risky part its own review surface.

Answering with a letter is enough to unblock.

### State

PR is draft, labelled `implement` + `waiting`. Phases 1-4 pushed and green (3,025 passed / 37 skipped at `d3375c8`). Nothing is half-applied — the working tree was clean when phase 5 stopped; `runner.py` is untouched on this branch.


---

espg:
🤖 *from Claude*

**Two offenders exist that the issue's census did not list — flagging, not acting.**

The ratified plan's offender table was measured on 2026-07-29 and named five modules. Re-measuring `src/zagg` at this branch's base (`92a9f54`) turns up two more that are already over the CLAUDE.md §4 limit:

| module | lines | origin |
|---|---|---|
| `src/zagg/client.py` | 1,148 | issue #326 / #333 |
| `src/zagg/client_transport.py` | 1,099 | issue #327 |

Both landed *after* the census was taken, which is why the plan does not mention them. `src/zagg/processing/read.py` at 996 remains the next-nearest and is still under.

Not touching them: adding modules to this PR's agreed scope is a side-effecting directive under CLAUDE.md §6, and the ruling that shaped this PR was explicitly about limiting churn ("this is too much churn as planned"). Adding two unplanned splits to a PR already carrying a blocked phase 5 would cut against that directly.

Three ways to take it, whichever you prefer:

1. **Fold into phase 5's follow-on.** They are `zagg.client`-family modules and phase 5 already has to touch `client.py`'s imports of runner internals, so the two splits would land next to each other.
2. **Separate small PR.** They are self-contained and the split is mechanical — probably two phase commits.
3. **Leave them.** 1,148 and 1,099 are 15% and 10% over; if the intent was the five worst offenders and not a repo-wide gate, they can wait.

Relevant to the acceptance criterion either way: issue #330 lists "Every `src/zagg` module <= 1,000 lines" as a condition, so whichever way this goes it should be recorded there — the criterion is not met by PR-A alone even once phase 5 lands.

Also worth noting for the phase-5 decision: `client.py` and `client_transport.py` growing past the limit within a day of the census is itself evidence that a **repo-wide guard test** would earn its keep. `tests/test_module_split_compat.py` already asserts the limit per split; widening it to walk all of `src/zagg` is a few lines. It is deliberately *not* in this PR — it would fail today on `runner.py` plus these two, so it belongs with the last split that makes it pass.


---

espg:
🤖 *from Claude*

**Run status — phases 1-4 landed, phase 5 waiting on the decision above.**

CI at `d3375c8`: `ruff` **pass**, docs `build` + `check` **pass**, Lambda `build-x86_64` and `build-arm64` **both pass**. `test (3.12)` and `test (3.13)` were still running when this run ended — the same commit is green locally on the full suite (**3025 passed, 37 skipped**, `uv run pytest -q`), and the only known env-dependent failure (`test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, pip resolution) is flagged in the PR body as pre-existing on `main`. Worth a glance at the matrix result; nothing in this diff is expected to move it, since every phase is a byte-identical relocation.

No ruff-bot inline comments to answer — the linter came back clean.

**State on stopping:** draft, `implement` + `waiting`. Working tree was clean; `runner.py` is untouched on this branch, so nothing is half-applied and the branch can be rebased or continued freely.

**To unblock:** a letter for phase 5 (a/b/c/d) from the [analysis comment](https://github.com/englacial/zagg/pull/349#issuecomment-5129979059). Questions (2)-(6) in the PR body are answerable at review time and do not block — none of them changes the code that is already here.


---

espg:
🤖 *from Claude*

## A sixth offender arrived today: `sweep_overview.py`

espg directed this here rather than to a standalone issue (2026-07-31; the standalone #364 is closed as a duplicate of this PR's scope).

`src/zagg/sweep_overview.py` was **901 lines** when this PR's phase list was drawn, so it was correctly out of scope then. Today it crossed:

- PR #360 (declare-pyramid retrofit, issue #358 — **merged** 23:27Z) took it to **1,194 lines**.
- PR #356 (writer-side O11 hashes, issue #342 — open) grows the same file again in its phase 4 (overview-leaf `content_hashes` sidecars) and is expected to push it **over 1,200**, the ceiling espg raised the trigger to today (CLAUDE.md §4, per PR #359).

espg's ruling on that crossing was: let it cross, and take the split as follow-up work — which is this PR.

## Proposed seam

The module already separates into two halves along its own section banners, so the cut is the one the file documents about itself:

| module | holds |
|---|---|
| kernels half | `fold_dense`, `combine_dense`, `fold_digests`, `encode_digest`/`decode_digest`, the merge-law constants (`EXACT_NAN_POLICY`, `TDIGEST_LAW`), `build_pyramid_block`, and the `declare_pyramid` retrofit path (config-derived declaration, `_semantic_guard`, `_validate_block_against_store`, `_field_drift`) |
| writer half | `sweep_overviews`, `OverviewFamily`, `ENVELOPE_NAME` and the per-node envelope, `_update_manifest_pyramid`, the D23 naming |

Same posture as phases 1–4: a pure move behind a re-export facade, no behavior change, no signature change, and `test_every_split_module_is_under_the_line_limit` extends to cover it for free.

## Sequencing

This one must land **after PR #356 merges** — that PR is actively editing the writer half (phase 4 sidecars, plus a phase 5 raster arm in flight), so splitting first would hand it a conflict across a relocation diff, exactly the failure mode this PR's "split first, then convert" ordering exists to avoid in the other direction.

Worth noting for whoever rebases this PR: two of its already-landed phases touch files that moved today — `hive.py` (phase 1) is being grown by PR #356's O11 wiring, and `sweep.py` (phase 3) was grown by PR #354's sweep telemetry (merged, in 0.41.0). Both are mechanical-move conflicts, not semantic ones, but they will need resolving at rebase.

espg:
🤖 *from Claude*

Still `waiting` on one thing: **a letter (a/b/c/d) for phase 5's `runner.py` split** — the four options are in [this comment](https://github.com/englacial/zagg/pull/349#issuecomment-5129979059) and the blocker is unchanged (the split relocates the callers of 34 of the 39 `zagg.runner` attributes the suite monkeypatches, across 139 sites, with a silent-no-op failure mode).

Checked this run and *not* acting on it: the sixth split ([`sweep_overview.py`](https://github.com/englacial/zagg/pull/349#issuecomment-5148315600), issue #364) named PR #356 as its sequencing dependency, and that dependency has cleared — PR #360 merged 2026-07-31T23:27Z and PR #356 at 23:53Z. So that phase is ready to start the moment phase 5 is answered; it is held only because the PR is `waiting`, and there has been no maintainer comment or push on this branch since the label went on (2026-07-30). Labels left as `implement` + `waiting`, PR left in draft, branch untouched at `d3375c8` — no merge of `main` performed, so the known mechanical-move conflicts in `hive.py` and `sweep.py` are still outstanding and will be resolved on the run that resumes this.


---

espg:
🤖 *from Claude*

**This PR has decayed past the point where it can be advanced mechanically, and I don't think the phase-5 letter is the decision that matters any more. Recommending it be redone against current `main`. Numbers below — this needs one call from you before any further commit is worth making.**

I came here to do the one piece of work that isn't blocked: the `sweep_overview.py` split you directed into scope on 2026-07-31, whose only stated dependency (PR #356 / PR #360) merged that same day. My last note here conceded it "is ready to start the moment phase 5 is answered; it is held only because the PR is `waiting`" — which is circular, so I went to start it. Measuring first is what stopped me.

### The branch is 54 days stale against the exact files it splits

Measured on `main` at `09ae3ee` against this branch's head `d3375c8`:

| module | size when this branch split it | on `main` now | commits on `main` touching it since `d3375c8` |
|---|---|---|---|
| `sweep_overview.py` | 901 | **2,724** | **67** |
| `config.py` | 2,747 | **3,864** | 50 |
| `hive.py` | 1,393 | **2,003** | 31 |
| `sweep.py` | 1,034 | **1,460** | 27 |
| `processing/raster.py` | 1,266 | **1,557** | 12 |

**187 commits on `main` have touched these five modules since this branch last moved.** `git merge-tree origin/main d3375c8` reports modify/delete conflicts on `config.py`, `hive.py` and `processing/raster.py` — main modified the file, this branch deleted it in favour of a package. Resolving those is not a merge; it is re-applying 600–1,100 lines of upstream change per module into a package layout built against a much smaller file. The purity property the PR rests on — "not one character of any function body changed", verified by AST diff — cannot survive that by hand.

And the sixth split is now moot as scoped: splitting the branch's **901-line** `sweep_overview.py` when `main` carries **2,724** lines would produce a seam against a file that no longer exists. The seam you posted is still the right seam; the file it described isn't the current one.

### The underlying problem also grew while this sat

`main` now has **11** modules over §4's 1,200-line cap, not the 5 this PR was scoped around:

```
5982  src/zagg/runner.py          2062  src/zagg/catalog/shardmap.py
3864  src/zagg/config.py          2003  src/zagg/hive.py
2724  src/zagg/sweep_overview.py  1557  src/zagg/processing/raster.py
1520  src/zagg/sweep_stage.py     1460  src/zagg/sweep.py
1361  src/zagg/processing/read.py 1252  src/zagg/client.py
1234  src/zagg/processing/spill.py
```

`runner.py` is at 5× the cap. That is the same list I raised as item (2) on issue #579.

### The call — one of three

- **(a) Redo PR-A against current `main`.** Close this PR, branch fresh, and re-derive each split from today's files with the same AST-purity verification. The seams in this PR's body stay valid as a design document; the diffs don't. Costs the four landed phase commits, keeps the method and the ratified seams. **My recommendation** — the mechanical-purity guarantee is the whole value of PR-A, and it's the one thing a hand-resolved merge/delete conflict destroys.
- **(b) Merge `main` in and repair here.** Keeps the PR number and history. Means hand-resolving three modify/delete conflicts by re-applying 187 commits' worth of upstream change into the new packages, then re-running the AST diff to prove nothing drifted. Doable, but the proof gets much weaker and the review burden lands on you.
- **(c) Park it deliberately and re-scope.** If the 11-module list means PR-A's scope was the wrong unit, say so and I'll put a fresh plan on issue #330 covering the current offenders in dependency order, with `runner.py` handled separately given the 139-monkeypatch-site problem.

Under any of the three, the phase-5 `runner.py` letter (a/b/c/d from [this comment](https://github.com/englacial/zagg/pull/349#issuecomment-5129979059)) is still yours to answer, and the 139-site silent-no-op hazard is unchanged by the staleness — that analysis holds. But it's no longer the thing blocking progress: (a)/(b)/(c) is.

I've pushed nothing. Choosing between these changes the PR's agreed scope, which isn't mine to do, so this stays `implement` + `waiting`.

### One caveat to retire

The body and an earlier comment treat the 3.12/3.13 matrix as still running. It completed green on `d3375c8` — 13 check runs, no failures. CI is not a concern here; staleness is.

---

PR #350, opened 2026-07-30
Narrative notebooks: query / write / read, each stage-timed (issue #328)

### LINKED ISSUE #328: Narrative notebooks: query / write / read, each stage-timed

espg:
Closes #328. Refs #265.

Three narrative notebooks that walk one pipeline end to end — **query → write → read** — each timing its own stages so end-to-end pipeline latency is tracked, per the [posted plan](https://github.com/englacial/zagg/issues/328#issuecomment-5124976420) and the [ratified shape](https://github.com/englacial/zagg/issues/265#issuecomment-5124017044) on #265. All three gates are open as of [today's status](https://github.com/englacial/zagg/issues/265#issuecomment-5128822038): PR #333 (`zagg.client` + futures + tqdm), PR #339 (tensor + mask emission), PR #343 (v2 event transport), all merged.

This is the substantial rework #328 says supersedes #169. That PR is `waiting` on @espg and is **not touched here**.

## What each notebook does

**`notebooks/01_query_shardmap.ipynb` — query.** CMR-STAC search → stac-geoparquet catalog → `ShardMap.build` on the production o9/o19 HEALPix grid, over the benchmark AOI and pin from `tests/data/benchmark/README.md` (NEON SERC AOP box, `2018-10-13 .. 2026-03-15`, ATL03 v007 / `NSIDC_CPRD`). Stages: **CMR query · catalog build · shard assignment**. Section 4 re-derives the densest shard with `bench_metrics.select_densest_shard` and compares it to the `targets.json` pin, so the notebook doubles as a drift check on the benchmark map. Fully anonymous — granule *metadata* search needs no Earthdata Login — but it does need a repo checkout (every input is resolved out of the tree); the intro says so.

**`notebooks/02_dispatch_fleet.ipynb` — write.** The `zagg.client` facade end to end: `Run.from_config(...)` → the pre-invoke cost ceiling → `dispatch()` → `handle.progress()` (tqdm over `as_completed`) → `handle.status()`, with `ShardError` surfacing and the `read`/`index`/`aggregate`/`write` worker splits summed across the fleet. Stages: **dispatch wall · fleet completion**, plus the per-phase worker table. Section 8 shows the v2 `transport="event"` + `Run.attach` variant (real-fleet only).

**Dual-mode.** `USE_REAL_FLEET = False` (default) injects a stub Lambda client defined in the notebook through the same `lambda_client=` seam `tests/test_client.py` uses, so the fan-out runs with no AWS account, no credentials and no cost. That exercises the real `Run`/`RunHandle`/`ShardError` paths, the real event construction and dispatch ordering, and the real post-run tail (the notebook prints which `mode` invokes fired: `ping`, `setup`, `finalize`, `coverage`, `stats`). It does **not** exercise the worker, IAM, or S3 — the notebook says so and tells you to read stub timings as a shape, not a measurement.

**`notebooks/03_read_tensors.ipynb` — read.** Decodes a t-digest product to the post-#339 reader contract `(tensor, mask, (offset, gain), morton_id)`. Stages: **fetch · decode · vertical rasterize**, done *by hand with public API only* so each is separately timed — `arr[base:base+4096]` for the I/O, `frombuffer(...).reshape(elem_shape)` for the ragged unpack (element read from the `ragged` attrs, never hardcoded), then `chunk_z_range` + `rasterize_cell` + `rank_to_rowcol`. Section 6 then checks the hand-assembled tensor **and** mask for exact equality against `read_tensors`, because a hand-decomposed split is only worth reading if it reproduces the shipping reader. Ends with bare-earth / canopy percentile surfaces and a CHM, plus one cell's reconstructed vertical profile.

**Dual-source.** `SOURCE = "synthetic"` (default) writes a 64×64 block of order-19 cells to a temp directory with zagg's own write path (`build_tdigest` + `write_ragged_to_zarr`), anchored at the real NEON SERC coordinates via `grid.assign`/`grid.shard_of`, with a photon column per cell (ground return + skewed canopy return) and a deliberate unobserved gap so the occupancy mask has both states. Writing through `rowcol_to_rank` means the reader's `rank_to_rowcol` has a real inverse to undo rather than a tautology. `SOURCE = "public"` opens a published hive leaf anonymously instead.

**Wiring (phase 4).** Binder badges + a README section for the series, `docs/notebooks.md` (new) with a mkdocs nav entry and `::: zagg.notebook.StageTimer`, a quickstart pointer, and `.binder/environment.yml` / `postBuild` prose updated to enumerate exactly which mode of each new notebook the badge covers. **No dependency changes** — tqdm was already added to the `analysis` extra by PR #333, per the [#265 ratification](https://github.com/englacial/zagg/issues/265#issuecomment-5124017044) (`analysis`, never `lambda`).

**`zagg.notebook.StageTimer`** is the shared helper the three notebooks time with: `with timer.stage(name):` accumulates (re-entering a name sums and counts calls — what the read leg's per-block split needs), `summary()` prints the table, `_repr_html_` renders it in Jupyter, `as_dict()` is the machine-comparable form. Landed in phase 1 rather than phase 4 so each notebook is self-consistent from its own commit.

## Phases

- [x] Phase 1 — `01_query_shardmap.ipynb` + `StageTimer` + its tests (358464e)
- [x] Phase 2 — `02_dispatch_fleet.ipynb` (b9bb47f)
- [x] Phase 3 — `03_read_tensors.ipynb` (1109dfb)
- [x] Phase 4 — wiring: Binder badges, README, `docs/notebooks.md` + nav, `.binder/` prose (73d74d4)

Adversarial-review folds (8 findings, one commit each): 4138927, efd8393, e8f25fa, 373282f, b759902, ed521c6, 7749379, 2ab78cb.

## Executed vs unexecuted cells

**Every notebook is committed with outputs cleared**, `execution_count: null`. That is this repo's convention — all nine pre-existing notebooks under `notebooks/` carry zero outputs, and #169 committed its notebooks the same way — and it is also forced here by the sandbox. Nothing is fabricated: no cell has an output it did not produce, because no cell has an output.

What was actually run, and what was not:

| Notebook | Path | Executed? |
| --- | --- | --- |
| 01 | anonymous CMR query | **No** — `cmr.earthdata.nasa.gov` is blocked by this environment's egress proxy (`403 Forbidden` on CONNECT). The path is the same anonymous CMR-STAC search the already-badged `shardmap_viewer.ipynb` runs. |
| 01 | everything else | **Yes**, offline against the committed fixtures: config → grid, `ShardMap` metadata keys, `select_densest_shard` vs the pin (matches: shard `5347395636851376137`, 66 granules), `to_json` round trip. |
| 02 | stub mode (the default, and what the badge covers) | **Yes** — executed top to bottom via `nbclient`, zero errors: 4 shards dispatched, 3 ok / 1 `ShardError` / 1 benign no-work counted as ok, the tqdm bar completing out of order, the phase split summing to 4.76 s, and the tail invokes `['ping', 'setup', 'finalize', 'coverage', 'stats']` all answered. |
| 02 | `USE_REAL_FLEET = True` | **No, deliberately** — CLAUDE.md §1 forbids dispatching to live AWS from a routine run. The posted plan expected committed outputs from a credentialed run; that is the one place this PR departs from it, and it needs an operator. |
| 03 | synthetic mode (the default, and what the badge covers) | **Yes** — executed top to bottom via `nbclient`, zero errors. Parity vs `read_tensors` exact for both tensor and mask; bare earth recovers the known terrain to 0.635 m MAE (0.597 m unbinned off the digests, against a 0.25 m binning floor), CHM to 0.859 m; both figures render. |
| 03 | `SOURCE = "public"` | **No** — anonymous access to `s3://sliderule-public/zagg-bench/...` could **not** be verified from here: an unsigned `obstore` client got `AccessDenied` on both LIST and GET, and an unauthenticated HTTPS GET of the store root returned `403`. See question (2). The block-discovery logic that branch uses was verified store-agnostically against the synthetic store (it returns exactly the offset the synthetic path computes independently). |

**No Binder badge covers an unexecuted path.** Each badge covers the default mode only, and `.binder/environment.yml` names which mode that is per notebook — the mistake caught by review on #169 is the one thing this PR is most careful about.

## How it was tested

- `uv run pytest -q`: **3036 passed, 9 skipped** (one pre-existing env-dependent failure deselected — below). New: `TestStageTimer` (8 tests) covering ordering, accumulation on re-entry, elapsed arithmetic, a raising stage still recording, `as_dict` being a snapshot copy, the summary table, the empty-timer path, and HTML escaping.
- `uv run ruff check src tests` / `ruff format --check src tests`: clean on everything this PR touches. CI's `ruff` job is green.
- Every notebook validated programmatically before each commit: `nbformat.validate`, every code cell `ast.parse`s, cell ids present/unique/stable across regenerations, zero outputs, `execution_count is None`.
- Notebooks 02 and 03 executed end to end via `nbclient` in their default modes (into scratch copies — the committed files stay cleared) and re-executed after every fold commit that touched them.
- `mkdocs build` runs clean over the new `docs/notebooks.md` and nav entry (the only errors are this environment's blocked `objects.inv` fetches, which predate this PR).
- `codespell` with the repo's `.pre-commit-config.yaml` flags: clean on the touched files.

## Pre-existing failures (not touched here)

- `ruff check src tests` — N818 on `src/zagg/registry.py:64` (`UnknownCapability`). Pre-existing on `main`, unrelated; `N` is not in the PR lint bot's `--select=E,F,W,I` gate, which is why CI's ruff job is green.
- `ruff format --check src tests` — flags `tests/data/benchmark/README.md`. Pre-existing on `main`, unrelated.
- `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` — fails in this local env on `main` too (the layer build's `pip` cannot resolve deps offline). Environment-dependent, unrelated.

## Questions for review

1. **A public knob for the run-stats verify window?** Notebook 2's stub path sets `runner._RUN_STATS_VERIFY_WINDOW_S = 0`, a private module global, in a document people copy from. It is convenience only — the poll is fail-open, so the stub run is correct without it; zeroing it just skips ~40 s of pointless polling (a 20 s window plus one re-fire) against a store that does not exist. Options: **(a)** leave as-is with the current "not a pattern to copy" comment; **(b)** add `Run.dispatch(..., verify_window_s=)`; **(c)** a module-level config knob on `zagg.client`. Adding API is a scope decision, so I left it at (a).

2. **The read leg's public store is unreachable, so the default is synthetic.** #328 wants the read notebook to target `s3://sliderule-public/zagg-bench/tdigest_healpix_o9_hive.zarr` and double as a standing regression on the lossless fraction. From this session that bucket is not anonymously readable (`AccessDenied` unsigned, on LIST and GET), so badging it would repeat #169's mistake. Options: **(a)** grant anonymous read on the `zagg-bench/` prefix and flip `SOURCE` to `"public"` by default; **(b)** publish the tdigest leaf to source.coop, which *is* verified anonymously readable from here (`englacial/zagg/benchmarks/` lists and reads fine unsigned) — note the existing `neon_tdigest.zarr` there has only `cell_ids`/`count`/`morton`, no digest field, so it would need a fresh publish; **(c)** leave synthetic as the default and keep the public branch as an opt-in for credentialed users, which is what is committed. (b) also gets the lossless-fraction regression the issue asks for, since it would be a real store on a Binder-reachable endpoint.

3. **Per-phase review/fold subagents could not be spawned in this environment** — no Agent/Task tool is exposed to this runner, so CLAUDE.md §2's "separate review subagent, then separate fold subagent, per phase" could not be followed literally. Rather than skip the gate, I ran one adversarial pass over the whole four-phase diff and posted it as a `COMMENT` review with 9 inline findings prefixed `🤖 *from Claude (review)*`, then folded them as 8 separate title-only commits with a reply on every thread (one declined with rationale). Flagging the deviation rather than letting it pass silently.

4. **Stub fidelity.** The stub worker returns a plausible `phase_timings` split (62/11/19/8) rather than replaying a recorded envelope from a real run. A recorded fixture would make the stub numbers real-ish but would go stale against worker changes and would need a fixture file. Happy to swap if you would rather the demo showed real proportions.


notebooks/03_read_tensors.ipynb
espg, line 1: 🤖 *from Claude (review)*

**Cell `open-public`: `cell_base = 0` is wrong for a hive leaf, and the branch is untested.**

```python
store = open_store(leaf, read_only=True, skip_signature=True)
field = f"{CHILD_ORDER}/h_tdigest"
cell_base = 0  # a leaf's cells axis starts at its own shard
```

The comment is right that a leaf's cells axis is scoped to its shard — but that axis is the **whole shard**, not one read chunk. On the production geometry (`parent_order: 9`, `child_order: 19`, `chunk_inner: 13`) a leaf holds `4**10 = 1,048,576` cells across **256** inner chunks of 4096. `cell_base = 0` therefore reads inner chunk 0, which for a sparse ICESat-2 shard is very likely empty — and the `fetch`/`decode` cells would sail through on empty bytes only for `chunk_z_range` to raise `"chunk has no populated cells with a finite quantile range"` two cells later, with nothing pointing at the cause.

This is exactly the class of defect the PR body already warns about (the public path could not be exercised from this session), which is why it should not be left as a guess. There is a public way to resolve it — `readers.cell_index`, which searches only the array's **stored** spans:

```python
from zagg.readers import cell_index, read_tensors

_t, _m, _s, block_morton = next(read_tensors(store, field))   # first POPULATED chunk
cell_base = cell_index(store, field, block_morton, 0, 0)      # rank 0 == (row, col) (0, 0)
```

That pass re-reads the block, so it must be called out as deliberately *outside* the timed split rather than folded into it.

---

espg, line 1: 🤖 *from Claude (review)*

**Cell `parity`: reaches into a submodule for a symbol the package re-exports.**

```python
from zagg.readers.tdigest_tensor import has_exact_occupancy
```

`has_exact_occupancy` is in `zagg/readers/__init__.py`'s `__all__` alongside `read_tensors`, `cell_index`, `chunk_z_range`, `rasterize_cell`, and both layout helpers. Every other import in this notebook comes from `zagg.readers`; this one alone points at the private module path. A worked example is where readers learn what the public surface *is*, so the inconsistency teaches the wrong thing.

---

espg, line 1: 🤖 *from Claude (review)*

**Cell `scene`: unguarded synthetic setup leaks wrong geometry into `SOURCE = "public"`.**

The `write` / `surfaces` ground-truth cells are correctly guarded with `if SOURCE == "synthetic"`, but the `scene` cell is not — so a `"public"` run still builds and discards a 64×64 terrain/canopy/observed scene, and, worse, defines

```python
PARENT_ORDER, CHILD_ORDER = 13, 19
```

which the published product **contradicts**: the pinned public store is `parent_order: 9`. `CHILD_ORDER` is load-bearing in the public branch (it builds `field = f"{CHILD_ORDER}/h_tdigest"`) while `PARENT_ORDER` is silently wrong there. Two constants of the same shape, one meaningful and one misleading, defined in a cell the mode switch says is synthetic-only.

Split them: geometry that both modes need (`SIDE`, `DEPTH`, `CHILD_ORDER`) belongs in the `source` cell next to the other configuration; `PARENT_ORDER` and the scene arrays belong inside the synthetic branch.

---

espg, line 1: 🤖 *from Claude (review)*

**Cell `profile-plot`: the hardcoded cell is only populated by construction in synthetic mode.**

```python
row, col = 20, 24  # the middle of the taller stand
```

That is the centre of the synthetic canopy blob. On a real block it is an arbitrary index that may well be an unobserved column — in which case the step plot is a flat zero line and both `axhline` calls draw at `nan`, i.e. a silently empty figure rather than an error. Pick the block's densest column from the tensor instead (`np.argmax(tensor.sum(axis=2))` unravelled), which reduces to roughly the same place in synthetic mode and is always meaningful.

---

notebooks/02_dispatch_fleet.ipynb
espg, line 1: 🤖 *from Claude (review)*

**Cell `dispatch`: a published example mutates a private module global, and the comment overstates the need.**

```python
from zagg import runner

runner._RUN_STATS_VERIFY_WINDOW_S = 0
```

Two problems, one of substance and one of framing.

*Substance*: this is the first place in the repo where a user-facing notebook writes to a `_`-prefixed module attribute. `tests/test_client.py` does the same thing, but it does it through `monkeypatch` in an autouse fixture — scoped, reverted, and in a test. Here it is a permanent mutation of an imported module in a document people copy from.

*Framing*: the comment implies the notebook needs this. It does not — the verify poll is fail-open (`_await_run_stats_object` treats read errors as absent and `_dispatch_run_stats` swallows the whole thing), so a stub run is **correct** without it; the setting only avoids ~40 s of pointless polling (a 20 s window plus one re-fire) on the finisher thread that the harvest loop then joins. Say that, so nobody reads it as a required incantation.

Whether the fix is a public knob (a `verify_window_s=` on `dispatch`, or a module-level config) is a scope/API decision, not a diff-scoped one — that belongs to @espg, under "Questions for review".

---

src/zagg/notebook.py
```diff
@@ -19,13 +22,122 @@
...
+            return f"{self.label}: no stages timed"
+        width = max(len(name) for name in self.stages)
+        width = max(width, len("stage"))
+        head = f"{'stage':<{width}}  {'calls':>5}  {'seconds':>9}  {'%':>6}"
+        lines = [f"{self.label} stage timings", head, "-" * len(head)]
+        for name, calls, seconds, pct in self._rows():
+            lines.append(f"{name:<{width}}  {calls:>5d}  {seconds:>9.3f}  {pct:>5.1f}%")
+        lines.append("-" * len(head))
+        lines.append(
+            f"{'total':<{width}}  {'':>5}  {self.total_s:>9.3f}  {100.0 if self.stages else 0:>5.1f}%"
```
espg: 🤖 *from Claude (review)*

Dead conditional. `summary()` returns at the top when `self.stages` is empty, so by the time this line runs `self.stages` is always truthy and the `else 0` branch is unreachable — the expression is a constant `100.0`. §4 is explicit about no dead code; it also makes the reader stop and check whether there is an empty case being handled here that the early return missed.

---

src/zagg/notebook.py
```diff
@@ -19,13 +22,122 @@
...
+    needs. Timing is ``time.perf_counter``; a stage that raises is still
+    recorded (the exception propagates unchanged). :meth:`as_dict` is the
+    machine-comparable form — the shape a later run-over-run series would key
+    on (wiring that up is out of scope here, issue #328 phase 4).
+    """
+
+    def __init__(self, label: str = "run"):
+        self.label = label
+        #: Insertion-ordered ``{stage: {"seconds": float, "calls": int}}``.
+        self.stages: dict[str, dict[str, float]] = {}
```
espg, line 70: 🤖 *from Claude (review)*

The annotation says both values are floats, but `calls` is an integer count — `entry["calls"] += 1` only type-checks because mypy promotes `int` to `float`. `as_dict()` hands this mapping straight to callers and is documented as "the machine-comparable form", so its value types are part of the contract; `_rows()` already has to write `int(e["calls"])` to get an integer back, which is the tell.

Not worth a `TypedDict` for two keys, but the docstring should state that `calls` is an `int` in the same mapping, so a consumer serialising `as_dict()` knows what it is getting.

---

notebooks/01_query_shardmap.ipynb
espg, line 1: 🤖 *from Claude (review)*

**The notebook silently requires a repo checkout, and the intro implies otherwise.**

Every input is resolved through `repo_file(...)` — the AOI geojson, the benchmark config, `targets.json`, the committed shard map — and cell `pin-check` goes further:

```python
sys.path.insert(0, repo_file(".github/scripts"))
import bench_metrics
```

Binder is a checkout so this is fine there, and `cost_reporting.ipynb` set the `find(rel)` precedent. But the intro sells the notebook on "the only network call is an anonymous CMR-STAC query", which reads like it would run against a plain `pip install zagg`. It would not: it dies at the first `repo_file` with `FileNotFoundError`, and the `.github/scripts` import would fail even in an sdist install. One sentence in the intro fixes it.

Worth a second look at the `bench_metrics` import specifically: pulling a CI helper into a user-facing narrative is defensible only because `select_densest_shard` is the single source of truth for the pin the cell is checking against — that rationale is in the code comment but not in the prose, where a reader would look for it.

---

tests/test_notebook.py
espg, line 1: 🤖 *from Claude (review)*

**`test_elapsed_tracks_real_time` / `test_as_dict_...` / `test_summary_table_...` patch the stdlib `time` module globally.**

```python
monkeypatch.setattr(notebook.time, "perf_counter", lambda: next(ticks))
```

`notebook.time` *is* `sys.modules["time"]`, so this replaces `time.perf_counter` process-wide for the duration of each test, not just for `StageTimer`. Anything else that reads the clock while the test runs — a library timeout, a warning throttle, pytest's own timing — sees the fake iterator, and when the iterator is exhausted it raises `StopIteration` from an unrelated call site. It is bounded by monkeypatch's teardown and by these tests being synchronous, but it is a wider blast radius than the test needs.

Flagging rather than prescribing: the alternatives (sleep-based assertions, or an injectable clock on `StageTimer`) each cost something — flakiness in one case, API surface for a test seam in the other.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial pass over all four phases (`358464e..73d74d4`), against CLAUDE.md §4 and #328's acceptance criteria.

**What holds up.** The Binder claims are scoped honestly — each badge covers a default path that genuinely needs no credentials, and neither `USE_REAL_FLEET = True` nor `SOURCE = "public"` is badged, which is the exact failure caught on PR #169. Notebook 3's parity check (hand-split tensor + mask vs `read_tensors`, asserted equal) is the right way to make a hand-decomposed timing split trustworthy rather than decorative, and writing the synthetic block through `rowcol_to_rank` so the reader's `rank_to_rowcol` has something real to invert keeps the round trip from being a tautology. `StageTimer` has tests covering accumulation, the raising-stage path, the empty case, and HTML escaping.

**The one substantive defect** is in notebook 3's `SOURCE = "public"` branch: `cell_base = 0` reads inner chunk 0 of a leaf that holds 256 of them, so the untested path will usually land on an empty chunk and surface as a confusing `chunk_z_range` error. Details inline, along with a public-API fix. The rest are consistency and framing issues — a submodule import for a re-exported symbol, synthetic geometry leaking into the public branch, a hardcoded plot cell, dead code and a loose annotation in `StageTimer`, and two places where the prose promises more than the code delivers (the repo-checkout requirement in notebook 1, the "needed" framing on the private-global mutation in notebook 2).

**Standing for @espg, not folded** (both imply a decision, not a diff fix):

1. The private-global mutation in notebook 2 — whether `zagg.client` should expose a public knob for the run-stats verify window instead.
2. The unreachable public store — the read leg's default is synthetic because anonymous access to `sliderule-public` could not be verified from this session. Options are in the PR body.

Nothing here blocks the phases; all four are complete.

---

notebooks/03_read_tensors.ipynb
espg: 🤖 *from Claude*

Folded in 7749379. The `cell_base = 0` guess is gone; the public branch now discovers a populated block:

```python
_t, _m, _s, block_morton = next(read_tensors(store, field))
cell_base = cell_index(store, field, block_morton, 0, 0)
```

with a comment spelling out why (a leaf's cells axis is the whole shard — 256 inner chunks at `parent_order: 9` / `chunk_inner: 13` — most of them empty) and stating that this pass re-reads the block and is therefore **setup, not a read stage**, deliberately outside the fetch/decode/rasterize split. A new markdown section 2b introduces it, so the public path is no longer an unannounced code cell.

The logic is store-agnostic, so I verified it against the synthetic store where the answer is known independently — `next(read_tensors(...))` + `cell_index(..., 0, 0)` returns exactly the `block_index(shard_key)[0] * cells_per_chunk` offset the synthetic path computes, and slicing at that offset recovers all 256 populated cells of a deliberately sparse block. The S3 leg itself is still unexercised from here; that stays under "Questions for review".

---

espg: 🤖 *from Claude*

Folded in b759902 — now `from zagg.readers import has_exact_occupancy, read_tensors`. Confirmed against `src/zagg/readers/__init__.py`: `has_exact_occupancy` is imported and listed in `__all__` there, so the package path is the supported one and every import in the notebook now points at it.

---

espg: 🤖 *from Claude*

Folded in ed521c6, taking the suggested split:

- `SIDE, DEPTH, CHILD_ORDER = 64, 6, 19` moved up to the `source` cell as shared block geometry, with a comment saying why `PARENT_ORDER` is *not* there (the published store's is 9).
- `PARENT_ORDER`, the SERC coordinates, the terrain/canopy/`observed` arrays and their summary print all moved inside `if SOURCE == "synthetic":` in the `scene` cell. `photon_column` and `PHOTONS_PER_CELL` stay at cell scope since they are pure helpers.
- The `write` cell had the same latent bug for a different reason — its `PipelineConfig(...)` referenced `PARENT_ORDER` *outside* the mode guard, so the fix above would have turned it into a `NameError` in public mode. The config construction moved inside the guard too.
- Section headings now read "synthetic mode only", and the section-1 intro says the read stages from section 3 on are shared.

Guarded with `if`, not `raise SystemExit` — the latter shuts the kernel down in some Jupyter versions. Re-executed the notebook end to end in synthetic mode: no errors, parity still exact.

---

espg: 🤖 *from Claude*

Folded in 2ab78cb, exactly as suggested:

```python
row, col = np.unravel_index(int(np.argmax(tensor.sum(axis=2))), (SIDE, SIDE))
```

The comment now says it picks the block's densest column so the cell is populated whatever the source is, noting that in synthetic mode that lands in one of the two stands. Re-executed: the figure still shows the ground spike with the canopy shoulder between the two percentile lines.

---

notebooks/02_dispatch_fleet.ipynb
espg: 🤖 *from Claude*

Framing folded in 373282f; the API question is left standing.

The comment now leads with "Convenience only — the stub run is CORRECT without this", spells out why (the verify poll is fail-open: `_await_run_stats_object` counts read errors as absent and `_dispatch_run_stats` swallows the whole thing, so a nonexistent store just reports "absent"), quantifies what it buys (~40 s of pointless polling — a 20 s window plus one re-fire — on the finisher thread the harvest loop joins), and says outright that writing a private module global "is not a pattern to copy".

Not folded: whether `zagg.client` should grow a public knob. That is an API addition, i.e. a side-effecting directive under §6, so it is not mine to take from a review comment — it is now question (1) under "Questions for review" in the PR body, with concrete options.

---

src/zagg/notebook.py
espg: 🤖 *from Claude*

Folded in 4138927. Correct — the early return above makes it unreachable, so the total row is now just:

```python
lines.append(f"{'total':<{width}}  {'':>5}  {self.total_s:>9.3f}  {100.0:>5.1f}%")
```

`test_empty_timer_renders_without_dividing_by_zero` already pins the early-return path, so the removed branch is covered by the case that made it dead. 38 tests still pass.

---

src/zagg/notebook.py
espg: 🤖 *from Claude*

Folded in efd8393, taking the documentation route you suggested over a `TypedDict`. The class docstring now carries:

> Value types in `stages` (and so in `as_dict`): `seconds` is a `float`, `calls` an `int`. They share one mapping whose annotation widens to `float` — mypy promotes `int` — so a consumer serializing the snapshot should read `calls` as the integer count it is.

and `as_dict`'s one-liner spells the shape out: `{"label", "stages": {name: {seconds: float, calls: int}}, "total_s"}`. The runtime was already correct (`0` and `+= 1` keep it an `int`; `_rows`'s `int(...)` is belt-and-braces for the `%d` format), so this is a contract-documentation fix, not a behavior change.

---

notebooks/01_query_shardmap.ipynb
espg: 🤖 *from Claude*

Both points folded in e8f25fa.

The intro now carries an explicit callout right under the Binder note:

> **Needs a repo checkout, not just an install.** Every input below is resolved out of the source tree — the AOI polygon, the benchmark config, `targets.json`, the committed shard map, and (section 4) the `bench_metrics` helper under `.github/scripts/`. Binder gives you a checkout, so it just works there; from a bare `pip install zagg` the first `repo_file(...)` raises `FileNotFoundError`. Clone the repo and run from it.

And the section 4 prose now carries the `bench_metrics` rationale that was only in a code comment: the rule is imported rather than reimplemented because it is the single source of truth the pin, the drift test, and the benchmark harness all agree on, and a second copy in a notebook is a second thing to drift — with a note that it is import-light (no AWS, no network) but does need the checkout.

---

tests/test_notebook.py
espg: 🤖 *from Claude*

Left by design — the analysis is right, but every alternative costs more than the exposure.

The blast radius is real but tightly bounded: `monkeypatch` reverts on teardown, the three tests are synchronous and do nothing but enter/exit `StageTimer.stage`, and each `ticks` iterator is sized to exactly the number of `perf_counter` calls the timer makes (2 per stage). Nothing else in those functions reads the clock. The `StopIteration`-from-elsewhere failure mode would need another clock reader inside the same test body.

Against that:

- **Sleep-based assertions** trade a deterministic test for a timing-sensitive one. The tests currently assert exact values (`2.0`, `25.0%` / `75.0%`), which is what makes them useful — with real sleeps they could only assert "greater than zero", i.e. roughly what `test_stages_record_in_order_with_calls` already covers without any patching.
- **An injectable clock on `StageTimer`** adds public API whose only consumer is the test suite. §4 is explicit about no speculative abstraction, and the class is deliberately ~80 lines.

It is also the established pattern here — `tests/test_client.py` monkeypatches `runner._RUN_STATS_VERIFY_WINDOW_S` in an autouse fixture, and `tests/test_notebook.py` already patches `builtins.input` two classes up in `test_temporal_backend_skips_gate`, which is a strictly wider global.

Standing for @espg if a test seam on `StageTimer` is wanted anyway — that would be an API decision, not a diff fix.

---

espg:
🤖 *from Claude*

Run complete. State: **all four phases landed, adversarial review posted and folded, `implement` + `waiting`, left in draft.**

**Phases** — 358464e (query notebook + `StageTimer` + tests), b9bb47f (write notebook), 1109dfb (read notebook), 73d74d4 (wiring: badges, README, `docs/notebooks.md` + nav, `.binder/` prose).

**Review + fold** — one adversarial pass over the whole four-phase diff, posted as a `COMMENT` review with 9 inline findings ([review](https://github.com/englacial/zagg/pull/350#pullrequestreview-4817910427)). 8 folded as separate title-only commits (4138927, efd8393, e8f25fa, 373282f, b759902, ed521c6, 7749379, 2ab78cb), 1 declined with rationale ([thread](https://github.com/englacial/zagg/pull/350#discussion_r3682027469)); every thread has a reply. The substantive one was notebook 3's public branch reading inner chunk 0 of a leaf that holds 256 — now resolved via `read_tensors` + `cell_index`, with the discovery logic verified store-agnostically against the synthetic store.

**CI** — `ruff`, `build`, `build / build-x86_64`, `build / build-arm64`, `test (3.13)` all green on 2ab78cb. `test (3.12)` is still `in_progress` ~35 min in, against 10 min for `test (3.13)` on the same commit; the full suite runs in 6.5 min locally on 3.12, so that reads as a slow runner rather than anything in this diff. Worth a glance before merge: [job](https://github.com/englacial/zagg/actions/runs/30536145868/job/90849828234).

**Why `waiting`** — four questions in the PR body need a call. Question (2) is the load-bearing one: #328's acceptance criteria say the read notebook consumes only the public store, but `s3://sliderule-public/zagg-bench/...` is not anonymously readable from this session (`AccessDenied` unsigned on both LIST and GET), so the committed default is the synthetic store and the public path is opt-in. Badging a path that cannot actually run anonymously is the exact failure caught on review of #169, so that stayed a question rather than a guess. Option (b) — publishing the tdigest leaf to source.coop, which *is* verified anonymously readable — would satisfy the criterion and give the standing lossless-fraction regression the issue asks for.

**Left in draft deliberately.** §5 would normally pair `waiting` + all-phases-complete with a draft→ready flip; this run was instructed to leave the PR in draft, and draft is the more conservative of the two, so I did not flip it. `gh pr ready` when you want it visible for review.

**Nothing else skipped.** No live AWS dispatch was run (§1), no CI config, `deployment/aws/`, or tags touched, no dependency added (tqdm was already in the `analysis` extra from PR #333), and #169 was not modified.

---

espg:
🤖 *from Claude*

Closing as superseded (espg-confirmed in-session 2026-08-18; the recommendation had been standing since before 08-17). The narrative-notebook work moved into the demo set (`00_tldr` / `01_query` / `02_write` / `03_read` / `06_paired` / `07_minimal`) being prototyped in the working tree; the public demo pass lands as its own PR once the release train ships and the CA stores are hosted on source-coop (anonymously reachable, per the §4 binder contract). The StageTimer work and phase commits here remain referenceable on branch `claude/328-notebooks` history if the demo pass wants to salvage any of it.

Issue #347, opened 2026-07-30
Overview sweep: fold mixed-order source leaves (D24) — replace the skip fence


espg:
The overview sweep (PR #344, `sweep_overview.py` — the leaf-fold loop) skips any source leaf whose `morton` array doesn't match the manifest `cell_order` subtree shape, with a loud warning citing a "D24 reader gate." That citation is stale: the reader-side gates are closed — espg/mortie#116 landed mixed-order morton support (mortie 0.9.1+, already zagg's floor), and moczarr#8 closed via moczarr PRs #22/#25. What actually remains is zagg-side and twofold:

1. **Sweep fold semantics for heterogeneous source orders** (this issue): folding a coarse leaf and a fine leaf into one ancestor overview node — per-leaf truth carried in morton words + MOC per the D24 ratification (regionally mixed cell orders as a first-class store state). The skip fence in `sweep_overview.py` is the code home; its error text now points here.
2. **Writer-side**: the D24 append guard still refuses heterogeneous cell orders, so no zagg writer can emit a mixed-order source tree today — which is why the fence is currently unreachable in practice and why this issue is not blocking PR #344.

Note the overview zarrs themselves make a tree multi-order, but re-sweeps exclude them by `role` (PR #344 fold commit b6b8bbe), not via this fence — this issue concerns mixed-order *source* leaves only.

Acceptance sketch: a synthetic mixed-order store (hand-built until the append guard relaxes) folds into correct ancestor overviews; the skip fence and its warning are removed; per-(shard, window) source-order sets (not a single per-product order) drive the fold, per the D24 uniqueness ruling recorded on issue #201.

Refs #201, PR #344. D24 ratification + the mortie/moczarr closures are linked above.

PR #354, opened 2026-07-31
Sweep telemetry: per-family durations + store-root run record (issue #353)

### LINKED ISSUE #353: Sweep telemetry: per-family duration_s in run_sweep summary, handler response, and a persisted record

## Motivation

The shard workers time their phases (`read`/`index`/`aggregate`/`write` → `metadata["phase_timings"]`) and those flow through `zagg.telemetry` into the run parquet. The sweep has nothing parallel: `run_sweep` returns per-family **counts** only (`written`/`current`/`empty`/`failed`), `sweep_after_run` logs one summary line, and the Lambda path is a fire-and-forget `mode="sweep"` Event invoke whose response is never captured — no run-record row at all. Benchmarking the overview/pyramid sweep (PR #344 Phase E measurement, and later the #347 mixed-order fold) is currently blind: CloudWatch billed duration and log lines are all that exists.

## Change (small, contained)

1. Time each family in `run_sweep` (`perf_counter` around the per-family runner) into the summary: `families.<name>.duration_s`, plus a total. For the overview family, also count bytes/objects read for the fold where already cheap to know.
2. Include the enriched summary in the `mode="sweep"` handler's response body (it already returns counts).
3. Persist it: one stats-style record at the store root so it lands where the cost rollups already read — same posture as the workers' `phase_timings` → run parquet.

This is a prerequisite for the pyramid-sweep benchmark runs (rather than benchmarking blind), so it should land before those.

## Benchmark-harness note

The dispatcher's sweep trigger is deliberately fire-and-forget; for measured runs, invoke `mode="sweep"` synchronously (RequestResponse) from a small driver script so the response is captured directly — or rely on the persisted record once (3) lands.

Refs #300, PR #344, #347.

espg:
Closes #353.

## What

Gives the rollup sweep the same telemetry posture the shard workers already have (`phase_timings` → run parquet), so the pyramid-sweep benchmark runs (PR #344 Phase E measurement, issue #347) are not limited to CloudWatch billed-duration inference:

1. **Per-family + total durations** — `run_sweep` times each family fold (`families.<name>.duration_s`) and the whole pass (`duration_s`), additive keys on the existing summary shape.
2. **A store-root run record** — each pass PUTs `sweep_stats_{ts}.json` (the summary itself, `spec: zagg-sweep/1`), timestamp-first like the run parquets (D20) so listing is chronological, and deliberately **outside** the `stats_*.parquet` glob `discover_leaves` scans, so a sweep record can never read as a shard run record. Fail-open (one warning, `record: null`) — telemetry, never truth (D9). `record=False` opts out.
3. **`mode="sweep"` handler response** — passes the pass total and the record key through, so a synchronous (RequestResponse) benchmark invoke captures timings directly. (`deployment/aws/lambda_handler.py` is named by issue #353.)

## Phases

- [x] Phase 1 — `run_sweep` durations + store-root record + tests (`phase 1`)
- [ ] Phase 2 — `mode="sweep"` handler response passthrough (`phase 2`)

## How it was tested

`tests/test_sweep.py::TestRecordAndTimings`: durations present per family and total (total ≥ every family), record written at root and equal to the summary (minus its own key), `record=False` writes nothing, record-write failure is fail-open (fold results untouched, one warning), and the record name is provably outside the `stats_.+\.parquet` discovery glob. Full `test_sweep.py` + `test_sweep_overview.py` + `test_telemetry.py` green locally; `ruff check` / `ruff format --check` clean.

## Questions for review

- **JSON vs parquet for the record**: the issue says "lands where the cost rollups already read." The run records are parquet; this PR writes JSON to keep the sweep record out of the `stats_*.parquet` discovery glob (and duckdb reads JSON fine). If the #296 cost rollups want a parquet row instead, that's a follow-up — flagging rather than deciding silently.
- **Overview fold bytes**: issue #353 said "where already cheap to know" — it is not (the fold would need byte accounting plumbed through `sweep_overview`'s leaf reads), so this PR records durations only. Left out deliberately.
- Same-second double-sweep overwrites the prior record (`timespec` seconds, like the D20 grammar) — accepted for telemetry; say the word if you want a nonce suffix.

### DIFF
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -1043,6 +1043,11 @@ def _handle_sweep(event: Dict[str, Any]) -> Dict[str, Any]:
                     "mode": "sweep",
                     "n_leaves": summary["n_leaves"],
                     "families": summary["families"],
+                    # issue #353: a synchronous benchmark invoke reads timings
+                    # straight off the response; the store-root record is the
+                    # durable copy (fail-open -> null).
+                    "duration_s": summary["duration_s"],
+                    "record": summary.get("record"),
                 }
             ),
         }
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -1025,14 +1025,25 @@ def _handle_sweep(event: Dict[str, Any]) -> Dict[str, Any]:
     logger.info(f"Sweep mode: folding rollups at {event.get('store_path')}")
     try:
         store_kwargs = _output_store_kwargs(event)
+        t0 = time.perf_counter()
         if event.get("leaves") is not None:
             leaves = [(int(key), window) for key, window in event["leaves"]]
         else:
             leaves = discover_leaves(event["store_path"], store_kwargs=store_kwargs)
         if not leaves:
+            # Same key set as the swept response: a benchmark driver reading
+            # body["duration_s"] must not KeyError on a no-work invoke (#353).
             return {
                 "statusCode": 200,
-                "body": json.dumps({"ok": True, "mode": "sweep", "n_leaves": 0}),
+                "body": json.dumps(
+                    {
+                        "ok": True,
+                        "mode": "sweep",
+                        "n_leaves": 0,
+                        "duration_s": time.perf_counter() - t0,
+                        "record": None,
+                    }
+                ),
             }
         summary = run_sweep(event["store_path"], leaves, store_kwargs=store_kwargs)
         return {
--- a/deployment/aws/lambda_handler.py
+++ b/deployment/aws/lambda_handler.py
@@ -1028,8 +1028,10 @@ def _handle_sweep(event: Dict[str, Any]) -> Dict[str, Any]:
         t0 = time.perf_counter()
         if event.get("leaves") is not None:
             leaves = [(int(key), window) for key, window in event["leaves"]]
+            discover_s = None  # the work set rode inline; nothing was derived
         else:
             leaves = discover_leaves(event["store_path"], store_kwargs=store_kwargs)
+            discover_s = time.perf_counter() - t0
         if not leaves:
             # Same key set as the swept response: a benchmark driver reading
             # body["duration_s"] must not KeyError on a no-work invoke (#353).
@@ -1041,6 +1043,7 @@ def _handle_sweep(event: Dict[str, Any]) -> Dict[str, Any]:
                         "mode": "sweep",
                         "n_leaves": 0,
                         "duration_s": time.perf_counter() - t0,
+                        "discover_s": discover_s,
                         "record": None,
                     }
                 ),
@@ -1056,8 +1059,13 @@ def _handle_sweep(event: Dict[str, Any]) -> Dict[str, Any]:
                     "families": summary["families"],
                     # issue #353: a synchronous benchmark invoke reads timings
                     # straight off the response; the store-root record is the
-                    # durable copy (fail-open -> null).
+                    # durable copy (fail-open -> null). ``duration_s`` is the
+                    # fold alone -- run_sweep's own span, which is exactly what
+                    # the store-root record carries -- and ``discover_s`` the
+                    # work-set derivation, null when the leaves rode inline.
+                    # Invoke wall-clock is their sum, not duration_s.
                     "duration_s": summary["duration_s"],
+                    "discover_s": discover_s,
                     "record": summary.get("record"),
                 }
             ),
--- a/src/zagg/sweep.py
+++ b/src/zagg/sweep.py
@@ -32,6 +32,10 @@ chain (shard nodes re-read their leaf sidecars each pass; interior nodes fold
 the freshly computed child payloads, so the rewrite cascades up). A second
 sweep over an unchanged tree recomputes but PUTs nothing.
 
+Each pass also PUTs its own small run record at the store root
+(``sweep_stats_{ts}.json`` — per-family durations + counts, issue #353); like
+every other sweep artifact it is regenerable telemetry, never truth.
+
 Rollup objects are JSON sidecars at digit nodes, named ``{family}.rollup.json``
 — deliberately DISTINCT from the leaf sidecar names (``stats.json`` /
 ``coverage.moc``): under D24 mixed-order stores a node can be leaf and
@@ -529,7 +533,14 @@ def get_family(name: str) -> SweepFamily:
     return cls()
 
 
-def run_sweep(store_root: str, leaves, *, families=None, store_kwargs: dict | None = None) -> dict:
+def run_sweep(
+    store_root: str,
+    leaves,
+    *,
+    families=None,
+    store_kwargs: dict | None = None,
+    record: bool = True,
+) -> dict:
     """One sweep pass: fold leaf artifacts up-tree for each family (D22).
 
     ``leaves`` is the run-record-derived work set — an iterable of
@@ -545,11 +556,20 @@ def run_sweep(store_root: str, leaves, *, families=None, store_kwargs: dict | No
     payload compare is the same-second backstop the generation stamp cannot see
     (module docstring). Returns a summary with
     per-family ``written`` / ``current`` (skip-if-current) / ``empty`` (no
-    artifact found) / ``failed`` (unmergeable, logged) counts.
+    artifact found) / ``failed`` (unmergeable, logged) counts, each carrying
+    its wall-clock ``duration_s`` (issue #353) plus a pass total — the sweep's
+    analogue of the workers' ``phase_timings``, so benchmark runs are not
+    limited to billed-duration inference.
+
+    Unless ``record=False``, the summary is also PUT at the store root as the
+    sweep's own run record (:func:`_write_sweep_record`, fail-open).
     """
+    import time
+
     from zagg.hive import MANIFEST_NAME, read_manifest
     from zagg.store import open_object_store
 
+    t0 = time.perf_counter()
     store_kwargs = dict(store_kwargs or {})
     manifest = read_manifest(store_root, **store_kwargs)
     if manifest is None:
@@ -566,14 +586,46 @@ def run_sweep(store_root: str, leaves, *, families=None, store_kwargs: dict | No
         "families": {},
     }
     for fam in fams:
+        fam_t0 = time.perf_counter()
         runner = getattr(fam, "sweep_store", None)
         if runner is not None:
-            summary["families"][fam.name] = runner(store_root, manifest, by_shard, store_kwargs)
+            result = runner(store_root, manifest, by_shard, store_kwargs)
         else:
-            summary["families"][fam.name] = _sweep_family(
+            result = _sweep_family(
                 store_root, store, fam, by_shard, shard_order, manifest.get("spec"), store_kwargs
             )
+        result["duration_s"] = time.perf_counter() - fam_t0
+        summary["families"][fam.name] = result
+    summary["duration_s"] = time.perf_counter() - t0
+    if record:
+        summary["record"] = _write_sweep_record(store, summary)
     return summary
+
+
+def _write_sweep_record(store, summary: dict) -> str | None:
+    """PUT the sweep's run record at the store root; its key, or ``None`` (#353).
+
+    One small JSON object per sweep pass — ``sweep_stats_{ts}.json``,
+    timestamp-first like the run parquets (D20) so a lexicographic listing is
+    chronological, and deliberately OUTSIDE the ``stats_*.parquet`` glob the
+    sweep's own run-record discovery scans (a sweep record must never read as
+    a shard run record). A second pass within the same wall-clock second
+    overwrites the first — acceptable for telemetry, which this is: fail-open
+    (one warning, ``None``), never truth, exactly like every other sweep
+    artifact (D9).
+    """
+    from datetime import datetime, timezone
+
+    import obstore
+
+    ts = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%SZ")
+    key = f"sweep_stats_{ts}.json"
+    try:
+        obstore.put(store, key, json.dumps({"spec": SWEEP_SPEC, **summary}, indent=1).encode())
+    except Exception as e:
+        logger.warning(f"sweep: run record write failed (fail-open, D9 — telemetry): {e}")
+        return None
+    return key
 
 
 def _normalize_leaves(leaves, shard_order: int):
--- a/src/zagg/sweep.py
+++ b/src/zagg/sweep.py
@@ -605,6 +605,14 @@ def run_sweep(
 def _write_sweep_record(store, summary: dict) -> str | None:
     """PUT the sweep's run record at the store root; its key, or ``None`` (#353).
 
+    The return is deliberately the bare store-root-relative key, NOT the joined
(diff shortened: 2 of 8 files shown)


deployment/aws/lambda_handler.py
```diff
@@ -1043,6 +1043,11 @@ def _handle_sweep(event: Dict[str, Any]) -> Dict[str, Any]:
                     "mode": "sweep",
                     "n_leaves": summary["n_leaves"],
```
espg, line 1058: 🤖 *from Claude (review)*

[bug] The `n_leaves == 0` early return (lines 1032-1036) does not carry the two keys this PR just added, so `mode="sweep"` now has two response shapes:

```python
return {"statusCode": 200, "body": json.dumps({"ok": True, "mode": "sweep", "n_leaves": 0})}
```

Why it matters: the comment five lines below says a synchronous benchmark invoke "reads timings straight off the response". A driver doing `json.loads(resp["Payload"])["body"] -> ["duration_s"]` gets a `KeyError` on exactly the no-work invoke — which is what a `discover: true` benchmark hits when the store has no run records yet, or when the previous pass already swept everything. It also makes "no leaves" indistinguishable from "old handler" for a consumer probing with `.get("duration_s")`.

Fix (one line + one test): return the same key set from the early branch —

```python
"body": json.dumps({"ok": True, "mode": "sweep", "n_leaves": 0, "duration_s": 0.0, "record": None}),
```

and cover it in `TestHandlerSweepResponse` with `event["leaves"] = []` (that path reaches `if not leaves` since the handler tests `is not None`).

deployment/aws/lambda_handler.py
```diff
@@ -1043,6 +1043,11 @@ def _handle_sweep(event: Dict[str, Any]) -> Dict[str, Any]:
...
                     "families": summary["families"],
+                    # issue #353: a synchronous benchmark invoke reads timings
+                    # straight off the response; the store-root record is the
+                    # durable copy (fail-open -> null).
+                    "duration_s": summary["duration_s"],
```
espg, line 1067: 🤖 *from Claude (review)*

[bug] `summary["duration_s"]` is `run_sweep`'s own span only, and it starts *after* the handler has already done the work on line 1031:

```python
leaves = discover_leaves(event["store_path"], store_kwargs=store_kwargs)
```

`discover_leaves` is a root LIST plus one GET + `fastparquet` read per `stats_*.parquet` run record (and, on a windowed store with pre-column records, one extra delimiter LIST per shard). On a store that has accumulated run records that is a real slice of the invoke — and it is unreported here and in the store-root record, both of which now read as "the sweep took N s". The failure mode is the one issue #353 is trying to end: a `discover: true` benchmark reports 3 s of sweep for a 40 s invoke and the CloudWatch billed duration disagrees with the artifact, with no way to tell which part moved.

Fix: time the handler body and report discovery separately rather than folding it into `duration_s` —

```python
t0 = time.perf_counter()
...
discover_s = time.perf_counter() - t0   # None when leaves rode inline
...
"duration_s": summary["duration_s"],      # the fold
"discover_s": discover_s,                 # the work set derivation
```

If you would rather leave discovery unmeasured this round, then at minimum say so in the comment above (`duration_s` is the fold, excluding `discover_leaves`), so the benchmark driver does not read it as invoke wall-clock.

tests/test_sweep.py
```diff
@@ -212,6 +212,69 @@ def test_deleting_every_rollup_leaves_leaf_reads_green(self, tmp_path):
...
+
+    def test_record_name_is_outside_the_run_parquet_glob(self, tmp_path):
+        # discover_leaves scans ``stats_.+\.parquet``; the sweep's own record
+        # must never read as a shard run record.
+        import re
+
+        _write_manifest(tmp_path)
+        _put_leaf(tmp_path, "-311")
+        summary = run_sweep(str(tmp_path), _leaf_refs("-311"))
+        assert not re.fullmatch(r"stats_.+\.parquet", summary["record"])
```
espg: 🤖 *from Claude (review)*

[test-gap] This pins the invariant against a **copy** of the regex, not against `discover_leaves` itself. The comment two lines up names `discover_leaves` as the thing being protected, but the test never calls it — so if the real glob ever loosens (or a future run-record reader picks up JSON at the root), this guard still goes green. The one assertion that would actually catch a sweep record reading as a shard run record is the round trip, and the helpers for it already live in this file (`_run_record` / `_row`, line 718).

Verified locally on this branch that this passes as written:

```python
_write_manifest(tmp_path)
_put_leaf(tmp_path, "-311")
_run_record(tmp_path, [_row("-311")])
summary = run_sweep(str(tmp_path), _leaf_refs("-311"))
# the record is now at the root the NEXT pass discovers from
assert discover_leaves(str(tmp_path)) == [(morton_word("-311"), None)]
```

Suggest keeping the `sweep_stats_...` name-grammar assertion on line 275 (it is a genuine grammar pin) and replacing the `stats_.+\.parquet` re-implementation on 274 with the `discover_leaves` call above. Worth noting the CLI makes this a live sequence, not a hypothetical: `main()` runs `discover_leaves` first and `run_sweep` second, so every manual pass discovers over the previous pass's record.

src/zagg/sweep.py
```diff
@@ -566,16 +586,48 @@ def run_sweep(store_root: str, leaves, *, families=None, store_kwargs: dict | No
...
+    import obstore
+
+    ts = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%SZ")
+    key = f"sweep_stats_{ts}.json"
+    try:
+        obstore.put(store, key, json.dumps({"spec": SWEEP_SPEC, **summary}, indent=1).encode())
+    except Exception as e:
+        logger.warning(f"sweep: run record write failed (fail-open, D9 — telemetry): {e}")
+        return None
+    return key
```
espg, line 636: 🤖 *from Claude (review)*

[nit] The record is returned as a bare store-relative key while the analogue this PR keeps citing returns a full path — `write_run_parquet` ends with

```python
return f"{store_root.rstrip('/')}/{key}"   # telemetry.py:442
```

and that joined value is what `summary["run_stats_path"]` carries through the runner. So a consumer holding only the `mode="sweep"` response body gets `"record": "sweep_stats_20260731T193122Z.json"` and has to re-join it with `store_path` itself, while the neighbouring "where is my run record" field is already absolute. Two fields, same job, two conventions.

Either return `f"{store_root.rstrip('/')}/{key}"` for parity (the docstring already sells this record as the run-parquet posture), or — if the bare key is deliberate, since it is what you would hand back to `obstore` against this same store — say that in the docstring; today it just says "its key", which reads as an oversight rather than a choice. Both call sites in the tests would move to a basename compare (`tests/test_sweep.py:233`, `:1104`).

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Fixed in f1c9a91. The `n_leaves == 0` branch now returns the same key set as the swept response — `duration_s` (measured, not the literal `0.0`, since the discovery path can spend real time before reaching the empty check) and `record: null`:

```python
if not leaves:
    # Same key set as the swept response: a benchmark driver reading
    # body["duration_s"] must not KeyError on a no-work invoke (#353).
    return {
        "statusCode": 200,
        "body": json.dumps(
            {
                "ok": True,
                "mode": "sweep",
                "n_leaves": 0,
                "duration_s": time.perf_counter() - t0,
                "discover_s": discover_s,
                "record": None,
            }
        ),
    }
```

(`discover_s` arrived in 802c7f5 for the adjacent finding; both no-work transports carry it.)

Covered by `TestHandlerSweepResponse::test_no_work_response_carries_the_same_keys`, which exercises both no-work transports — `leaves: []` (inline, empty) and the discovery path over a manifest-bearing store with no run records — and asserts `n_leaves == 0`, `duration_s >= 0.0`, `record is None` on each.

deployment/aws/lambda_handler.py
espg: 🤖 *from Claude*

Fixed in 802c7f5, taking the first option — discovery is measured and reported as its own field rather than folded into `duration_s`:

```python
t0 = time.perf_counter()
if event.get("leaves") is not None:
    leaves = [(int(key), window) for key, window in event["leaves"]]
    discover_s = None  # the work set rode inline; nothing was derived
else:
    leaves = discover_leaves(event["store_path"], store_kwargs=store_kwargs)
    discover_s = time.perf_counter() - t0
```

`discover_s` is `null` when the leaves rode inline (nothing was derived to charge for), a measured span on the `discover: true` path. The response comment now scopes both fields explicitly so a driver cannot read either as invoke wall-clock:

```python
# ``duration_s`` is the fold alone -- run_sweep's own span, which is
# exactly what the store-root record carries -- and ``discover_s`` the
# work-set derivation, null when the leaves rode inline.
# Invoke wall-clock is their sum, not duration_s.
```

The store-root record's semantics are unchanged, deliberately: it is written inside `run_sweep` (`_write_sweep_record`, called from the end of `run_sweep`), so it covers the fold pass and only the fold pass. It has no visibility into a discovery that happened in the handler above it, and giving it one would mean threading a handler-side timer through the library API for a number the response already carries. So: record = fold, response = fold + discovery, and the two `duration_s` values agree by construction.

Covered by `test_discovery_path_reports_its_own_span` (discovery over a store with a run record: `discover_s >= 0.0`), plus `discover_s is None` asserts on the inline path in `test_response_carries_durations_and_record` and the no-work test.

tests/test_sweep.py
espg: 🤖 *from Claude*

Adopted in 3c35b3a. The `stats_.+\.parquet` re-implementation is gone; the test now calls `discover_leaves` over a store holding both a run parquet and a sweep record, so it fails if the real glob ever loosens:

```python
_write_manifest(tmp_path)
_put_leaf(tmp_path, "-311")
_run_record(tmp_path, [_row("-311")])
summary = run_sweep(str(tmp_path), _leaf_refs("-311"))
# The record now sits at the root the NEXT pass discovers from.
assert discover_leaves(str(tmp_path)) == [(morton_word("-311"), None)]
assert re.fullmatch(r"sweep_stats_[0-9]{8}T[0-9]{6}Z\.json", summary["record"])
```

Kept the `sweep_stats_...` name-grammar assertion as you suggested — it is a genuine grammar pin (timestamp-first, D20), independent of what discovery happens to scan. Also folded your note about the CLI making this a live sequence into the comment, since that is the reason the round trip is the right shape here.

`pytest tests/test_sweep.py -q` green (63 passed).

src/zagg/sweep.py
espg: 🤖 *from Claude*

Taking the second option — bare key stays, docstring now states it as a choice (635fa72):

```
The return is deliberately the bare store-root-relative key, NOT the joined
``{store_root}/{key}`` that ``telemetry.write_run_parquet`` hands back: it
is the same currency as every other key this sweep writes (the rollups), so
it is what you hand straight back to ``obstore`` against this same store.
Joining would mean plumbing ``store_root`` in here, where only the store
handle is held. Callers holding the response body alone re-join it with the
``store_path`` they invoked with.
```

Reasoning for preferring the ruling over parity: `_write_sweep_record(store, summary)` holds only the store handle, so joining would mean adding a `store_root` parameter purely to decorate a return value — and the record key would then be the one sweep-written key in a different convention from every rollup key the same pass writes. The asymmetry with `run_stats_path` is real but it is the smaller one, and it is now documented rather than implied. Test call sites are unchanged (`tests/test_sweep.py:233`, `:1104` still compare the bare key), so no basename compare was needed.

Happy to flip to the joined form if you would rather have the two response fields match — say the word and it is a one-line change plus the two test sites.

espg:
🤖 *from Claude*

CI was red and my earlier "pre-existing at base" call was wrong — I compared against `f56d26b`, which is this PR's own phase-2 commit, so both sides of the comparison already carried the record write. Against the real base (`main`, 92a9f54) all four tests pass; they are regressions this PR introduced. Fixed in 80b1bbb, 92d4e3c, c240b13.

**Cause.** Local dispatchers run `sweep_after_run` in-process at end of run (default on for hive), so since phase 2 every local hive run PUTs one `sweep_stats_{ts}.json` at the store root. Four tests either model root object counts for cost purposes or enumerate the root listing exactly, and none knew about it. This is intended product behavior — sweep telemetry on real runs — so the fix extends the model and the expectations rather than suppressing the record.

**80b1bbb — object-count model (`.github/scripts/bench_objects.py` + `tests/test_benchmark_objects.py`).** The record was landing in `other`, which `object_count_mismatch` treats as an unconditional finding, so this would have tripped on every hive benchmark run, not just in tests. Added a `_is_sweep_record` predicate mirroring `_is_run_parquet`:

```python
def _is_sweep_record(key: str) -> bool:
    """A store-root sweep run record (issue #353): ``sweep_stats_*.json``."""
    return "/" not in key and key.startswith("sweep_stats_") and key.endswith(".json")
```

It classifies as metadata for the same reason the run parquet does — fixed, shard-independent, one object per run — and `metadata_max` gains `+1` on the hive branch. Ceiling only, floor stays 1: like the root MOC and the run parquet it is fail-open (D9 telemetry, one warning and `None`), and on the Lambda path the sweep is fire-and-forget, so it may legitimately be absent at measurement time. `metadata` 4 -> 5 in both tests, which also closes the 32 == 31 write-path gap in `test_hive_sharded_store_matches_model`.

**92d4e3c — root listings (`tests/test_coverage_root.py`, `tests/test_hive.py`).** Both now assert the record IS present rather than filtering it, since its presence on a default hive run is the contract: `test_flag_off_writes_nothing` gains a `sweep_stats_*.json` assertion and goes 4 -> 5 objects; `test_local_hive_writes_manifest_before_cells` enumerates it in the exact listing next to the run parquet.

**c240b13 — `tests/test_benchmark.py::test_hive_config_expected_counts_root_moc_optional`.** Not in the reported four; it surfaced once the model widened, since it pins the whole `expected_object_counts` dict for the committed hive arm. Metadata window `[1, 4] -> [1, 5]`, total `[12, 15] -> [12, 16]`; per-shard stays exact at 11, so the issue #215 bypass tripwire is untouched.

No `src/` behavior changed in any of the three — the record still writes exactly where it did.

**Local status.** The four named tests plus `tests/test_sweep.py` are green (67 passed), as is the full suite bar three environment-only failures I left alone: `test_lambda_build::test_function_build_succeeds` (sandbox has no network — `No matching distribution found for zarr>=3.1.5`; fails on `main` too) and the two `test_runner.py::TestConsolidationGate` cases, which fail only in my working copy because a stale gitignored `out.zarr/` at the repo root collides with the test's relative `./out.zarr` store — both pass in a clean `main` checkout.

**One flag for review.** 80b1bbb touches `.github/scripts/bench_objects.py`, which is outside the usual "don't touch CI" line. I took the fix direction as naming it, and the alternative — adjusting only the tests — would have left the shipped model mis-classifying the record on every real hive benchmark run. Happy to split it out if you would rather it land separately.

espg:
🤖 *from Claude*

Closing the flagged judgment call: espg **blessed** the `80b1bbb` touch to `.github/scripts/bench_objects.py` (in-session, 2026-07-31) — in-scope for issue #353, since the issue changed what a store contains and a store-content model that doesn't learn the new object emits false findings on every real hive benchmark run.

Status: both phases landed, all four review findings folded (plus the fifth latent `test_benchmark` pin), CI fully green on `c240b13` (ruff, build ×3, test 3.12 + 3.13). Marking ready for review.

PR #356, opened 2026-07-31
Writer-side O11 content hashes (issue #342)

### LINKED ISSUE #342: Freeze the O11 content-hash serialization: writer must match the moczarr reader's pinned recipe (keys, combined, vlen)

espg:
Closes #342

Writer-side O11 content hashes: the spec §5 recipe (per-array sha256 over decoded C-order little-endian bytes; vlen arrays as u64le length-prefixed payload bytes per cell in flat C order; combined = sha256 over the sorted per-array hex digests joined by `"\n"`) computed **at write, in-worker**, and recorded under `content_hashes` in the leaf's D20 stats sidecar per §5.3.

Implements the [plan](https://github.com/englacial/zagg/issues/342#issuecomment-5146680196) under the binding [ratification record](https://github.com/englacial/zagg/issues/342#issuecomment-5146967460): (1) sweep-written overview leaves also get §5 hashes; (2) the overview stamp's sweep-internal skip digest stays separate and untouched; (3) O11 is hive-only — the writer no-ops on non-hive layouts, documented; (4) hashes are computed from the STAGED arrays, guarded by a write-read parity test. The recipe itself is adopted verbatim from the moczarr reference (`moczarr.stats.hash_arrays` / `combined_hash`) and is not this PR's to adjust.

## Approach

- `src/zagg/content_hash.py` (new) holds ALL hash logic: `hash_array` (§5.2, dense + vlen with the element→bytes normalization and its raise gate), `hash_arrays(group, staged=...)` (discovery-based §5.1 scope via `group.members(max_depth=None)`; a staged in-memory value is hashed in place of a read-back), `combined_hash`, `content_hashes_record` (the §5.3 structured, key-sorted shape), and `StreamingArrayHash` (the incremental digest, phase 5).
- **Spatial hive leaves** — the leaf writers (`write_leaf_to_zarr` / `write_ragged_leaf_to_zarr` / `write_dataframe_to_zarr`) gained an optional `staged_out` sink: refs to the exact slabs written, no copies, on BOTH the sharded and the streaming/unsharded paths. `zagg.hive.process_and_write_hive` computes the record after the stamp on the success path (fail-open), returns it as `metadata["content_hashes"]`, and times it as the `hash` phase. `telemetry.build_record` rides it into the D20 sidecar for both dispatch backends; `merge` treats it equal-or-None.
- **Overview leaves** — `sweep_overview._write_overview` writes the same record from the folded slabs + morton words already in memory, as a sibling `{stem}.stats.json` PUT after the stamp.
- **Raster hive leaves (phase 5)** — this path never holds a whole band array (it streams one timestep at a time, issues #231/#232), so neither staging nor read-back applies. Instead each array carries a live `sha256` fed **row by row as written**: the concatenation of the written rows' decoded C-order LE bytes IS the array's full decoded byte sequence, so the incremental digest finalizes to exactly what `hash_array` over the assembled array would produce. Coordinates (`time`/`morton`/`cell_ids`) are written whole at template time and hash from their staged values.

  Both correctness traps are **enforced, not assumed** — §5 covers an array's full contents, and a digest that is silently wrong is worse than none:
  - *unwritten rows still count*: rows missing at finalize contribute their **fill** bytes (an array with a gap and no reconstructable fill is dropped);
  - *order is a precondition*: the sink delivers timesteps in **completion** order (`asyncio.as_completed`, and on worker threads at `write_buffer > 1`), so out-of-order rows are parked until their turn under a byte budget (`STREAM_PENDING_MAX_BYTES`, 64 MiB); a row written twice, a row outside the extent, a shape/dtype mismatch, or a parking overflow **invalidates** the digest. An invalidated array drops the **whole** leaf record (a partial map would read as "array missing" to a §5.1 verifier and its `combined` would be a digest of a subset), with one warning naming the array and the reason.

## Phases

- [x] Phase 1 — recipe port (`1f72cab`)
- [x] Phase 2 — worker wiring, staged source (`c9aefd7`)
- [x] Phase 3 — coverage completeness (`2051ae7`)
- [x] Phase 4 — overview leaves + §8.2 O11 addendum (`954d9b4`, `cb43209`)
- [x] Phase 5 — raster hive leaves, incremental streaming hash (`144e831`)
- [x] Spec: the overview sidecar naming grammar pinned in §5.3 (`e178868`)

## Review round

Adversarial self-review posted 4 inline findings; 3 folded — `2dc7f85` (deep-copy nested `content_hashes` on merge), `d160d2c` (warn on staged keys matching no array), `1bf700e` (stage dense arrays on the streaming leaf path too, pinned by an unsharded kitchen-sink test reproducing the committed sharded fixture's hashes byte-for-byte). The [naming finding](https://github.com/englacial/zagg/pull/356#discussion_r3693345107) was escalated and is now resolved by the first ruling below.

## How it was tested

- `uv run pytest -q`: **3153 passed**, 37 skipped, 1 failed — `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, environment-only (its `pip` resolves cp310 wheels in this sandbox), untouched by this PR.
- **The phase-5 acceptance gate**: a raster hive leaf written through the real streaming path, then re-hashed in the test with `content_hash.hash_arrays` over the leaf **read back from the store** — the streamed record equals the read-back per-array map and `combined` byte-for-byte (`TestRasterHiveContentHashes::test_streamed_hashes_match_a_full_read_back`). Plus 12 `StreamingArrayHash` unit tests: in-order == out-of-order == whole-array hash, unwritten rows == fill bytes, NaN fill, every precondition violation dropping the hash with its reason, the parking-budget overflow, and a threaded concurrent-update case.
- Spatial parity is pinned both ways: staged hashes == full read-back, and the worker-recorded record byte-matches the committed `kitchen_sink.expected.json` on **both** the sharded and unsharded paths.
- `ruff check` / `ruff format --check` clean on all touched files. One pre-existing repo-wide finding unrelated to this PR: `N818` on `src/zagg/registry.py::UnknownCapability` (CI's lint bot selects only `E,F,W,I`, so CI is unaffected).
- `origin/main` merged in (`f6d0178`) to clear a CONFLICTING state that was silently suppressing CI on this branch; the one conflict was `tests/test_benchmark_objects.py` (issue #353's metadata count vs this PR's overview-sidecar residue pin) and both assertions are kept.

## Module sizes

`hive.py` (1,446) and `sweep_overview.py` (934 here, crossing 1,200 once PR #360 lands) both exceed the old ceiling **by design**, per espg's standing ruling: no split in this PR. The mechanical splits attach to PR #349 (issue #330 PR-A), where the seam analysis is posted; `hive.py`'s split there is a known mechanical-move conflict to be resolved on that PR's rebase.

## Questions for review — resolved

- ~~Overview sidecar naming~~ — **ruled: keep the D23 grammar as shipped** (`{stem}.stats.json`), because per-window overview sidecars at one ancestor node would otherwise collide on a single `stats.json`. Now pinned in `docs/specification.md` §5.3 (`e178868`).
- ~~Module size~~ — **ruled: approved**, no split here (see above).
- ~~Overview sidecar record shape~~ — **ruled: keep `telemetry.build_record`.** The D20 schema is what the sweep's stats fold and the cost rollups already parse; a bespoke shape would need its own parser and would break `telemetry.merge`, and the null run-only fields are honest — the artifact had no run.
- ~~Raster hive leaves~~ — **ruled: wire them, no read-back** — landed as phase 5 above.

## Still standing

- The **write-seam deviation** stands as accepted: the plan named `processing/worker.py`, but the hive leaf write (and the only place the staged slabs exist) is `zagg.hive.process_and_write_hive`, shared by both dispatch backends — wiring there also keeps `deployment/aws/lambda_handler.py` untouched.
- The `.github/scripts/bench_objects.py` classifier does not yet bucket the new overview sidecars; `tests/test_benchmark_objects.py` pins them as the only unclassified residue and folds them out of the write-path total. That gap is being fixed separately under issue #362.

### DIFF
--- a/src/zagg/content_hash.py
+++ b/src/zagg/content_hash.py
@@ -125,7 +125,8 @@ def hash_arrays(group: Any, *, staged: Mapping[str, np.ndarray] | None = None) -
     for key, node in group.members(max_depth=None):
         if not isinstance(node, zarr.Array):
             continue
-        hashes[key] = hash_array(staged[key] if key in staged else node[...])
+        values = staged[key] if key in staged else node[...]
+        hashes[key] = hash_array(np.asarray(values))
     unused = set(staged) - set(hashes)
     if unused:
         logger.warning(
--- a/src/zagg/content_hash.py
+++ b/src/zagg/content_hash.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 
 import hashlib
 import logging
+import threading
 from typing import Any, Mapping
 
 import numpy as np
@@ -38,6 +39,16 @@ logger = logging.getLogger(__name__)
 
 #: Width of the §5.2 vlen recipe's per-cell length prefix (uint64, LE).
 VLEN_LENGTH_PREFIX = 8
+
+#: Byte budget for out-of-order rows ONE :class:`StreamingArrayHash` may hold
+#: (issue #342 phase 5). The raster sink receives timesteps in *completion*
+#: order (``asyncio.as_completed``), so a row arriving before its predecessors
+#: must be parked until its turn. The budget bounds that parking — at typical
+#: raster leaf geometry (tens of acquisitions x a shard's cells) the whole
+#: array is well under it, while a pathological leaf degrades to "no recorded
+#: hash" (§5.3 unverifiable, not tampered) instead of eating the streaming
+#: write path's memory bound (issues #231/#232).
+STREAM_PENDING_MAX_BYTES = 64 * 1024 * 1024
 
 
 def _element_bytes(element: object) -> bytes:
@@ -124,6 +135,130 @@ def hash_arrays(group: Any, *, staged: Mapping[str, np.ndarray] | None = None) -
             sorted(unused),
         )
     return hashes
+
+
+#: Sentinel for "the fill row has not been materialized yet".
+_UNSET = object()
+
+
+class StreamingArrayHash:
+    """§5.2 digest of an array assembled from leading-axis rows as they are written.
+
+    The incremental hash source for write paths that never hold the whole
+    array (issue #342 phase 5, the raster hive leaf): a ``(time, cells)``
+    band is written one timestep at a time, and the concatenation of those
+    rows' decoded C-order little-endian bytes IS the array's full decoded
+    byte sequence — so a live ``sha256`` fed row by row finalizes to exactly
+    the digest :func:`hash_array` would compute over the assembled array. No
+    read-back, no whole-array staging.
+
+    Two things the identity depends on, both ENFORCED here rather than
+    assumed — a digest that is silently wrong is worse than no digest:
+
+    - **Unwritten rows still count.** The recipe covers the array's full
+      logical contents, so a row that never gets a slab contributes its
+      **fill** bytes. Rows missing at :meth:`finalize` are filled from
+      ``fill_value``; an array with a gap and no reconstructable fill is
+      dropped instead of hashed.
+    - **Order.** Rows may arrive in any order (the raster sink runs on
+      ``asyncio.as_completed``), so out-of-order rows are parked until their
+      turn — bounded by :data:`STREAM_PENDING_MAX_BYTES`. A row written
+      twice, a row outside the declared extent, a shape/dtype mismatch, or a
+      parking overflow all **invalidate** the digest: :meth:`finalize`
+      returns ``None`` and :attr:`invalid_reason` says why.
+
+    Thread-safe: the raster sink hands slabs off to worker threads when
+    ``write_buffer > 1``.
+    """
+
+    def __init__(self, shape, dtype, fill_value=None):
+        self.shape = tuple(int(s) for s in shape)
+        self.dtype = np.dtype(dtype)
+        self.fill_value = fill_value
+        self.invalid_reason: str | None = None
+        self._row_shape = self.shape[1:]
+        self._n_rows = self.shape[0] if self.shape else 0
+        self._digest = hashlib.sha256()
+        self._next = 0
+        self._pending: dict[int, np.ndarray] = {}
+        self._pending_bytes = 0
+        self._fill_row: Any = _UNSET
+        self._lock = threading.Lock()
+
+    def _invalidate(self, reason: str) -> None:
+        if self.invalid_reason is None:
+            self.invalid_reason = reason
+        self._pending.clear()
+        self._pending_bytes = 0
+
+    def _row_bytes(self, values: np.ndarray) -> bytes:
+        values = np.ascontiguousarray(values)
+        if values.dtype.byteorder == ">":  # canonical form is little-endian
+            values = values.astype(values.dtype.newbyteorder("<"))
+        return values.tobytes()
+
+    def update(self, index: int, values) -> None:
+        """Record the row written at leading-axis ``index`` (any order)."""
+        with self._lock:
+            if self.invalid_reason is not None:
+                return
+            index = int(index)
+            values = np.asarray(values)
+            if values.shape != self._row_shape or values.dtype != self.dtype:
+                self._invalidate(
+                    f"row {index} is {values.shape}/{values.dtype}, expected "
+                    f"{self._row_shape}/{self.dtype}"
+                )
+                return
+            if not 0 <= index < self._n_rows:
+                self._invalidate(f"row {index} is outside the array's {self._n_rows} rows")
+                return
+            if index < self._next or index in self._pending:
+                self._invalidate(f"row {index} was written more than once")
+                return
+            if index > self._next:
+                self._pending[index] = values
+                self._pending_bytes += values.nbytes
+                if self._pending_bytes > STREAM_PENDING_MAX_BYTES:
+                    self._invalidate(
+                        f"out-of-order rows exceeded the {STREAM_PENDING_MAX_BYTES}-byte "
+                        f"parking budget"
+                    )
+                return
+            self._digest.update(self._row_bytes(values))
+            self._next += 1
+            while self._next in self._pending:  # drain rows this one unblocked
+                row = self._pending.pop(self._next)
+                self._pending_bytes -= row.nbytes
+                self._digest.update(self._row_bytes(row))
+                self._next += 1
+
+    def finalize(self) -> str | None:
+        """The array's §5.2 digest, or ``None`` when it cannot be stood behind."""
+        with self._lock:
+            if self.invalid_reason is not None:
+                return None
+            for index in range(self._next, self._n_rows):
+                row = self._pending.pop(index, None)
+                if row is not None:
+                    self._pending_bytes -= row.nbytes
+                    self._digest.update(self._row_bytes(row))
+                    continue
+                if self._fill_row is _UNSET:
+                    self._fill_row = (
+                        None
+                        if self.fill_value is None
+                        else np.full(self._row_shape, self.fill_value, dtype=self.dtype).tobytes()
+                    )
+                if self._fill_row is None:
+                    self._invalidate(
+                        f"row {index} was never written and the array declares no "
+                        f"fill_value to reconstruct it"
+                    )
+                    return None
+                self._digest.update(self._fill_row)
+            self._next = self._n_rows
+            return self._digest.hexdigest()
 
 
 def combined_hash(hashes: Mapping[str, str]) -> str:
--- a/src/zagg/content_hash.py
+++ b/src/zagg/content_hash.py
@@ -80,7 +80,7 @@ def hash_array(values: np.ndarray) -> str:
     return hashlib.sha256(values.tobytes()).hexdigest()
 
 
-def hash_arrays(group: Any) -> dict[str, str]:
+def hash_arrays(group: Any, *, staged: Mapping[str, np.ndarray] | None = None) -> dict[str, str]:
     """O11 per-array hashes of every named zarr array beneath ``group``.
 
     ``group`` is an open zarr v3 group at the leaf ROOT; the scope is
@@ -88,14 +88,25 @@ def hash_arrays(group: Any) -> dict[str, str]:
     whatever named arrays exist beneath the root — data fields, ragged vlen
     payloads and their locations siblings, ``morton``, every coordinate —
     keyed by the array's path relative to the leaf root (e.g. ``"8/morton"``).
+
+    ``staged`` maps array keys to the IN-MEMORY values the writer just wrote
+    (the ratified issue #342 hash source: the leaf write path already holds
+    each array's full slab, so hashing is a memory-bandwidth pass, no data
+    GETs). A staged value is hashed in place of a store read-back; enumerated
+    arrays with no staged value (e.g. ``resolution: chunk`` companions, which
+    are written per chunk-block rather than as one slab) fall back to reading
+    that array back. The two sources are pinned equal by the write-read
+    parity test in ``tests/test_content_hash.py`` — the CI guard for the
+    dtype-cast/fill drift class §5 exists to catch.
     """
     import zarr
 
+    staged = staged or {}
     hashes: dict[str, str] = {}
     for key, node in group.members(max_depth=None):
         if not isinstance(node, zarr.Array):
             continue
-        hashes[key] = hash_array(node[...])
+        hashes[key] = hash_array(staged[key] if key in staged else node[...])
     return hashes
 
 
--- a/src/zagg/content_hash.py
+++ b/src/zagg/content_hash.py
@@ -1,0 +1,119 @@
+"""O11 content hashes: the writer-side port of the spec §5 recipe (issue #342).
+
+The logical content hash of a hive leaf is **per-array sha256 over decoded
+values** — never stored object bytes, so codec/packaging changes are invisible
+by construction while any value change flips the hash (exact bytes, no float
+tolerance). The recipe is NOT this module's to adjust: the normative text is
+``docs/specification.md`` §5 and the reference implementation is
+``moczarr.stats.hash_arrays`` / ``combined_hash`` (espg/moczarr PR #23),
+adopted verbatim here and pinned byte-for-byte by the committed conformance
+fixtures (``tests/data/spec/*.expected.json``, spec §7).
+
+Recipe summary (§5.2/§5.3):
+
+- fixed-width arrays hash as their full decoded contents, raw C-order
+  **little-endian** bytes at the declared dtype;
+- vlen (object-dtype) arrays hash per cell in flat C order as
+  ``u64le(len(payload)) || payload`` — the length prefix makes the digest
+  injective and covers the cell grid, not just the payloads;
+- an element with no byte recipe RAISES (a silently wrong digest is worse
+  than none);
+- ``combined`` = sha256 over the SORTED per-array hex digests joined by
+  ``"\\n"``, hashed as ASCII (array names deliberately excluded).
+
+Hashes are recorded in the leaf's D20 stats sidecar under ``content_hashes``
+in the §5.3 structured shape (:func:`content_hashes_record`); absence reads
+as *unverifiable, not tampered*.
+"""
+
+from __future__ import annotations
+
+import hashlib
+from typing import Any, Mapping
+
+import numpy as np
+
+#: Width of the §5.2 vlen recipe's per-cell length prefix (uint64, LE).
+VLEN_LENGTH_PREFIX = 8
+
+
+def _element_bytes(element: object) -> bytes:
+    """One vlen cell's decoded payload as raw little-endian bytes (§5.2).
+
+    ``variable_length_bytes`` cells decode to :class:`bytes` (zagg's ragged
+    payloads and their ``{field}_locations`` siblings). ``None`` — an
+    unwritten cell may decode as ``None``, not ``b""`` — is zero-length;
+    ``str`` covers a vlen-utf8 future; an ndarray (a typed ``zagg-ragged/2``
+    cell, spec §6) normalizes to C-contiguous little-endian bytes at its
+    element dtype. Anything else RAISES rather than hashing a ``repr`` — a
+    digest that is silently wrong is worse than no digest at all.
+    """
+    if element is None:
+        return b""
+    if isinstance(element, (bytes, bytearray, memoryview)):
+        return bytes(element)
+    if isinstance(element, str):
+        return element.encode()
+    if isinstance(element, np.ndarray):
+        values = np.ascontiguousarray(element)
+        if values.dtype.byteorder == ">":
+            values = values.astype(values.dtype.newbyteorder("<"))
+        return values.tobytes()
+    raise ValueError(
+        f"vlen element of type {type(element).__name__} has no O11 byte recipe "
+        f"(expected bytes, str, ndarray, or None)"
+    )
+
+
+def hash_array(values: np.ndarray) -> str:
+    """The §5.2 hash of ONE decoded array (fixed-width or vlen/object dtype)."""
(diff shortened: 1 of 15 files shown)


src/zagg/sweep_overview.py
```diff
@@ -873,6 +876,36 @@ def _write_overview(
...
+        record = build_record(
+            shard_key=morton_word(node),
+            metadata={
+                "cells_with_data": int(populated.sum()),
+                "granule_count": int(fold["granule_count"]),
+                "content_hashes": content_hashes_record(hash_arrays(group, staged=staged)),
+            },
+            window=stamp_window,
+        )
+        write_sidecar(path, record, spec=SPEC_V3, **store_kwargs)
```
espg, line 906: 🤖 *from Claude (review)*

**[blocking]** The overview sidecar name is keyed to `SPEC_V3` unconditionally, but **no zagg writer ever emits a `morton-hive/3` manifest** — `build_manifest` writes `HIVE_SPEC` / `HIVE_SPEC_V2` only:

```python
# src/zagg/hive.py:310
"spec": HIVE_SPEC_V2 if windowing else HIVE_SPEC,
# src/zagg/hive.py:846
"spec": HIVE_SPEC if window is None else HIVE_SPEC_V2,
```

So every overview sidecar this PR writes lands under a naming grammar the store's own manifest spec does **not** select. That matters because sidecar naming is an explicitly spec-keyed writer/reader seam on both sides — `zagg.telemetry.sidecar_key(name, spec)` and `moczarr.stats.stats_sidecar_key(leaf_name, spec)` mirror each other on purpose, and moczarr's version *raises* on an unrecognized spec rather than defaulting ("a versioned naming bump must be a loud, deliberate change, or a writer/reader spec mismatch would key the wrong name and read as absent instead of failing"). A reader resolving these objects the documented way — manifest spec first — computes `stats.json` / `stats_{window}.json` at the node and finds nothing, while the actual objects sit at `{window}.stats.json`. zagg's own `sweep._sidecar_window(name, spec)` classifies them as `_NO_SIDECAR` on a `/1`/`/2` store for the same reason.

Note this is *not* the same as the overview `.zarr` basename: spec §4.2 makes `{window}.zarr` normative for overviews regardless of spec. There is no equivalent sentence anywhere for the overview *sidecar* name — `docs/specification.md` §5.3 only says "recorded in the leaf's D20 stats sidecar", and the §8.2 addendum in this PR does not mention naming. Per CLAUDE.md §4 ("any change to a wire format... must update `docs/specification.md` ... external readers decode from the spec + fixtures alone"), a new object name in the store needs the spec text in the same PR.

Suggested resolution, pick one:
1. Add the rule to `docs/specification.md` §4 (next to §4.2) — "an overview's D20 sidecar is `{stem}.stats.json` next to `{stem}.zarr`, spec-independent, because the overview basename is itself spec-independent" — and file the mirror on moczarr so `stats_sidecar_key` grows an overview branch. This makes the current code correct.
2. Key off `manifest["spec"]` like every other sidecar and skip the sidecar (fail-open, unverifiable-not-tampered) on legacy-spec stores, so no unaddressable object is ever written.
3. Hold overview sidecars until `/3` lands.

Option (1) looks right given decision (1) ratified overview scope, but it needs the spec commit here, not later.

src/zagg/hive.py
```diff
@@ -1343,6 +1353,41 @@ def _write_chunk(block_index, carrier, ragged):
     # and a no-data shard (no leaf) stays write-less.
     if not metadata.get("error") and "phase_timings" in metadata and "store" in box:
         metadata["phase_timings"]["write"] = _write_elapsed
+    if not metadata.get("error") and "store" in box:
+        # O11 content hashes (issue #342, spec §5): computed in-worker at
+        # write, from the STAGED arrays (the ratified source — the write path
+        # already holds every slab, so this is a memory-bandwidth pass; only
+        # per-chunk-block companions fall back to a read-back inside
```

**[should-fix]** This comment is factually wrong on the most common hive config, and the gap is the ratified staged-source decision (4), not a doc nit.

`sharded` is silently forced off whenever the leaf has one inner chunk — i.e. whenever `chunk_inner` is unset, which is the default:

```python
# src/zagg/grids/healpix.py:114
if self.chunks_per_shard <= 1:
    sharded = False
```

On that path `write_leaf_to_zarr` is never called, so the only writer that populates `staged` is `write_ragged_leaf_to_zarr` — **every dense array falls back to a read-back**, not just `resolution: chunk` companions. Measured on a stock `default_config()` + one ragged field at `HealpixGrid(4, 7, layout="fullsphere", config=cfg)` (no `chunk_inner`), driven through the real `process_and_write_hive`:

```
grid.sharded = False  chunks_per_shard = 1
STAGED : ['7/h_ragged']
READ BACK (not staged): ['7/count', '7/h_max', '7/h_mean', '7/h_min', '7/h_q25',
                         '7/h_q50', '7/h_q75', '7/h_sigma', '7/h_variance', '7/morton']
```

10 of 11 arrays read back. Correctness is fine (parity holds trivially — read-back vs read-back), but the ratified decision was "**(4) Hash source — staged arrays**", chosen explicitly over read-back "despite the extra GETs", and the D20 addendum this PR adds to `sparse_coverage.md` §8.2 states "the hash source is the **staged arrays** (no read-back GETs)". On an S3 leaf that is one GET + decode per dense array on the worker's write path, on the default config. No test covers this path's hash source either (`tests/test_content_hash.py` only drives sharded grids).

Suggested resolution, pick one:
1. Stage the streaming path too: give `_write_chunk` a leaf-shaped slab sink (lazily allocated per array at `grid.shard_slab_shape()`, filled at `leaf_block_index`) and hand it to `staged`, so the ratified property actually holds everywhere. Add a non-sharded parity test alongside `test_write_read_parity`.
2. If the read-back is accepted for K==1 leaves, say so precisely here and in the §8.2 addendum ("dense arrays on the unsharded streaming path and `resolution: chunk` companions read back"), and raise the deviation from decision (4) on issue #342 for the record rather than leaving the docs claiming no GETs.

src/zagg/content_hash.py
```diff
@@ -0,0 +1,130 @@
...
+    dtype-cast/fill drift class §5 exists to catch.
+    """
+    import zarr
+
+    staged = staged or {}
+    hashes: dict[str, str] = {}
+    for key, node in group.members(max_depth=None):
+        if not isinstance(node, zarr.Array):
+            continue
+        hashes[key] = hash_array(staged[key] if key in staged else node[...])
```
espg: 🤖 *from Claude (review)*

**[should-fix]** A staged key that matches no enumerated array is silently dropped, which makes total staging breakage **undetectable by every test in this PR**.

The lookup is one-directional: keys come from `group.members()`, and `staged` is consulted only when a key already exists in the enumeration. If the writer's key composition ever drifts from the enumeration's — `f"{grid.group_path}/{name}"` in `write.py` vs. the member path, a leading slash, a `group_path` change, a future nested group — then *nothing* in `staged` is ever used, every array is read back, and:

- `test_write_read_parity` still passes (read-back == read-back);
- `test_matches_pinned_fixture_hashes` still passes (the store is correct);
- `test_companion_field_enters_hash_set_via_readback_fallback` still passes;
- `test_staged_values_take_priority_over_the_store` still passes — it builds its `staged` dict by hand (`{"6/count": ...}`), so it proves the *mechanism*, never the *writer's* keys.

The one thing that silently disappears is the ratified no-GET property, which is the whole point of decision (4). Same class of blind spot as the unsharded path (separate comment on `hive.py`), but this one hides it even where staging is supposed to work.

Suggested resolution:
```python
    hashes: dict[str, str] = {}
    for key, node in group.members(max_depth=None):
        ...
    unused = set(staged) - set(hashes)
    if unused:
        raise ValueError(f"staged arrays not present under the leaf root: {sorted(unused)}")
```
(a raise is safe here — the hive caller is fail-open, so a key-drift bug degrades to "no record" rather than a silently-read-back record that looks healthy). Plus one test that pins the *writer's* staged key set: capture `staged_out` from `write_leaf_to_zarr` in the kitchen-sink run and assert it equals the leaf's dense + ragged array inventory (everything except `resolution: chunk` companions).

src/zagg/telemetry.py
```diff
@@ -64,6 +64,10 @@
     "run_id",
     "semantic_hash",
     "granules_sha256",
+    # O11 (issue #342): per-leaf by definition, so a multi-leaf rollup
+    # collapses it to None (absence = unverifiable, §5.3) while merge([r])
+    # stays r.
+    "content_hashes",
```
espg, line 70: 🤖 *from Claude (review)*

**[nit-but-real]** `content_hashes` is the first *nested* dict in `_EQ_OR_NONE_KEYS`, so the merge site's defensive copy no longer does what its comment claims:

```python
# telemetry.py:246
# Defensively copy dict values (``lambda``/``invoked_by``) so a
# rolled-up record never aliases a leaf's nested dict, mirroring
# build_record (issue #297).
out[key] = dict(first) if isinstance(first, dict) else first
```

`dict(first)` is shallow. For `{"lambda": ...}` / `{"invoked_by": ...}` (flat, scalar values) that is sufficient; for `{"arrays": {...}, "combined": ...}` it is not — `merge([r])["content_hashes"]["arrays"] is r["content_hashes"]["arrays"]`, exactly the aliasing the comment says it prevents. Nothing mutates it today, so this is not a live bug; it is an invariant the code states and no longer holds, and rollup records do outlive the leaf records they were folded from.

Suggested resolution: `out[key] = copy.deepcopy(first) if isinstance(first, dict) else first` (or special-case `content_hashes` through `content_hashes_record`), and extend `test_content_hashes_merge_equal_or_none` with `assert merge([a, b])["content_hashes"]["arrays"] is not a["content_hashes"]["arrays"]` so the next nested key added here can't reintroduce it.

src/zagg/hive.py

Fixed in 1bf700e — dense arrays are now staged on the streaming path too.

- `write_dataframe_to_zarr` takes an optional `staged_out` dict. In the non-companion branch it lazily allocates the array at its full shape (`np.full(array.shape, array.metadata.fill_value, dtype=values.dtype)` — mirroring `write_leaf_to_zarr`'s slab construction) and places the chunk's values at `tuple(slice(b * c, (b + 1) * c) for b, c in zip(chunk_idx, grid.chunk_shape))`, trailing (vector) dims spanning whole. The docstring pins it as the **hive leaf writers' staging seam** — flat-path callers must not pass it, since the region math assumes the leaf-local chunk grid.
- `hive.process_and_write_hive._write_chunk` passes `staged_out=staged`.
- The O11 comment block no longer claims companions are the only read-back — after this it actually is true: dense and ragged are staged on **both** the sharded and streaming paths; `resolution: chunk` companions are the one read-back fallback (written per chunk-block, never as a leaf slab).
- Memory: this re-adds the same O(dense leaf) term (~1.3 MB at production geometry) the sharded path already accepts — noted at the `staged` declaration, next to the existing acceptance comment above `chunk_results`.

Measured on your exact repro shape (kitchen-sink config, `sharded=False`), staged now covers everything:

```
STAGED       : ["6/composition", "6/count", "6/h_tdigest_noise", "6/h_tdigest_noise_locations",
                "6/h_tdigest_signal", "6/h_tdigest_signal_locations", "6/morton"]
READBACK ONLY: []
```

Test: `TestStreamingLeafStaging::test_unsharded_leaf_reproduces_pinned_hashes_from_staged_values` writes an UNSHARDED kitchen-sink leaf through `process_and_write_hive` (`_write_kitchen_sink` generalized with a `sharded=` parameter; a wrapper fake collects into a local `chunk_results` list and replays each `(block, carrier, ragged)` through the `write_chunk` callback) and asserts the hashes equal the committed sharded fixture's pinned `content_hashes` (decoded-value hashing is packaging-invariant — §5's point) AND a full read-back, plus caplog holds no `staged values for` warning (the finding-(3) mechanism) so the staged keys demonstrably landed.

Also verified out-of-band that the unsharded path keeps parity with a `vector` field + a `resolution: chunk` companion in the mix (all 13 arrays parity-true), so the trailing-dim placement is right beyond the kitchen-sink shape.

src/zagg/content_hash.py
espg: 🤖 *from Claude*

Fixed in d160d2c — `hash_arrays` now warns on staged keys that match no enumerated array:

```python
unused = set(staged) - set(hashes)
if unused:
    logger.warning(
        "staged values for %s match no array beneath the leaf root; those arrays "
        "were hashed from a store read-back instead (staged keys must be array "
        "paths relative to the leaf root, e.g. '8/morton')",
        sorted(unused),
    )
```

Digests stay correct (that array was read back), but total key-composition drift between the writer and the enumeration is now loud instead of invisible. Module logger added to `content_hash.py`.

Tests, both halves of the blind spot you named:

- unit — `TestWorkerWiring::test_unused_staged_key_warns`: `hash_arrays(group, staged={"nonexistent/key": np.zeros(1)})` logs the warning (caplog) and returns hashes equal to the unstaged call;
- integration — `test_companion_field_enters_hash_set_via_readback_fallback` now wraps the real `process_and_write_hive` call in `caplog.at_level(logging.WARNING, logger="zagg.content_hash")` and asserts `"staged values for" not in caplog.text`, pinning that the *writer's* `staged_out` keys all land (the thing the hand-built `{"6/count": ...}` staged dict could never prove). The new unsharded streaming test (finding on `hive.py`, fixed in 1bf700e) carries the same assertion.

src/zagg/telemetry.py
espg: 🤖 *from Claude*

Fixed in 2dc7f85 — the merge site now copies one level deeper, so the comment's invariant holds again for the first nested value:

```python
out[key] = (
    {k: dict(v) if isinstance(v, dict) else v for k, v in first.items()}
    if isinstance(first, dict)
    else first
)
```

Took the one-level-deeper form rather than `copy.deepcopy` — `content_hashes` is exactly two levels (`arrays` map of str→str, `combined` str), so this is sufficient and keeps `lambda`/`invoked_by` behavior byte-identical (flat scalar values, unaffected by the inner `isinstance` check). Comment updated to say what it now does and why (`content_hashes` nests an `arrays` map a shallow copy would still alias).

Test: `test_merge_does_not_alias_nested_content_hashes` asserts `merged["content_hashes"] == a["content_hashes"]`, `merged["content_hashes"] is not a["content_hashes"]`, and `merged["content_hashes"]["arrays"] is not a["content_hashes"]["arrays"]`.

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Leaving this one standing for espg — no code change, and deliberately no `docs/specification.md` change.

Why not folded: the fix you describe is a **spec decision**, not a diff-scoped fix. Nothing in `docs/specification.md` pins the *overview* sidecar name (§5/§7 pin the leaf's), so picking a rule here means writing normative text — and the standing constraint on issue #342 bars this PR from adjusting the spec (spec drift = stop and raise). It also needs a matching moczarr-side reader, since `moczarr.stats.stats_sidecar_key` raises rather than defaults, exactly as you quote. So it wants an espg ruling first.

Why the writer emits SPEC_V3 grammar today rather than the legacy name: legacy naming is a **self-clobber** at an overview node. An ancestor node can carry several windows' overviews, and legacy keys them all to the single `stats.json` — the last window written would overwrite every earlier one. The `{window}.stats.json` grammar is the only one of the two that keeps them distinct, so it is the writer's current best guess pending the ruling, not an assertion that `morton-hive/3` exists.

The parent session is posting the decision question on the PR thread with concrete options and applying `waiting`, so this blocks review-ready rather than getting silently resolved here.

espg:
🤖 *from Claude*

Two items need a ruling before this PR can go ready — applying `waiting`. All four phases are implemented, review findings (2)–(4) are folded (`2dc7f85`, `d160d2c`, `1bf700e`); finding (1) below is the one left standing.

**Overview sidecar naming ([the standing review finding](https://github.com/englacial/zagg/pull/356#discussion_r3693345107))** — the overview leaves' D20 sidecars are named with the D23 window-only grammar (`{stem}.stats.json`, i.e. `all.stats.json` / `2019.stats.json`), but the stores' manifests are `morton-hive/1`/`/2`, and sidecar naming is a spec-keyed seam on both sides (`telemetry.sidecar_key`, `moczarr.stats.stats_sidecar_key`). Nothing in `docs/specification.md` pins an overview sidecar name, and this PR is barred from touching the spec by the issue's standing constraint. Options:

(1) Keep the D23 grammar as shipped (my recommendation): overview basenames are already v3-named unconditionally, and per-window sidecars need per-window names — the legacy grammar would key every window's overview sidecar at a node to the single `stats.json` (self-clobber). Needs one spec sentence (§5.3 or §4) pinning "an overview leaf's sidecar is its basename stem + `.stats.json` regardless of the store spec", plus a moczarr reader helper (its `read_stats` is shard-arithmetic only today, so overview records are currently write-only either way).

(2) Drop the overview sidecar write from this PR (revert that part of phase 4) until the spec pins a name; overview zarrs would stay unverifiable-not-tampered for now.

**Module size (issue-#358 ruling)** — `hive.py` lands at 1,438 lines (1,393 pre-PR; +45 from the O11 wiring), which is at/over the 1,200 stop-and-raise line. The overage is pre-existing; options:

(1) Accept for this PR (delta is small and the hash logic itself lives in the new `content_hash.py`) and open a split issue — the natural cut is the per-shard write path (`process_and_write_hive` + the leaf write/stamp helpers, ~350 lines) into a `hive_write.py`, with re-exports so callers are untouched.

(2) Do that split inside this PR (blurs the O11 diff — not recommended).

Everything else is done: full suite green locally (3092 passed; known env-only `test_lambda_build`), fixture parity byte-exact, staged-source parity pinned sharded AND unsharded.

PR #357, opened 2026-07-31
Subtree span reads for the ragged vlen readers (issue #351)

### LINKED ISSUE #351: Readers: sub-leaf region read (read_subtree) — one morton subtree as a contiguous cell span

## Motivation

Within one store, the ragged readers expose exactly two granularities today: `read_cell` (2 ranged GETs) and the whole-array sweep (`read_tensors` / `read_raw_values` / `read_locations`, which LIST the array's stored `c/<ordinal>` objects via `_stored_chunk_spans` and visit every one). Region selection exists only at leaf granularity, one layer up: coverage-MOC/AOI intersection picks hive leaves (issue #200; moczarr's `open_hive(aoi=...)`), then each selected leaf is swept whole. There is no "everything below this morton node" *within* a leaf or flat store.

## Why it's cheap

The cells axis is nested-ordered, so an order-`k` subtree of the axis is **one contiguous cell span** `[r·4^d, (r+1)·4^d)`. On a sharded store a contiguous slice costs the shard-index suffix plus only the covering inner chunks — the same access shape `read_cell` already uses, generalized to a span. `cell_index` already does span-restricted searches over the `morton` sibling; no new store primitive is needed.

## Scope sketch

- `read_subtree(store, field, morton_word, ...)` (name open): the `read_tensors` contract restricted to the subtree below `morton_word`; span-restricted variants (or a shared span parameter) for `read_raw_values` / `read_locations`.
- Validate the requested word is an ancestor within the axis' subtree (same `clip2order` arithmetic as `_chunk_word`); a word outside coverage yields an empty result, not an error — consistent with the AOI-over-no-coverage posture.
- Only stored objects intersecting the span are fetched (intersect `_stored_chunk_spans` with the span before reading).

## Split (per espg)

1. **Spec/narrative text — a mod in PR #346**: the reader-consumption passages (`docs/ragged_layout.md` "consumes the layout two ways" and the spec's reader expectations) gain the third access mode, and the layout consequence worth stating explicitly: *a morton subtree of the cells axis is one contiguous span* (it follows from nested order, but readers should be able to rely on it by citation).
2. **Code — this issue**: the zagg reader implementation + tests.
3. **moczarr counterpart** — tracked on the moczarr side; its MOC-backed lazy index already models the row domain as rank-space interval sets, so a subtree select is largely native there.

Refs #346, #340, #209, #336, #200.

espg:
Closes #351

Span-restricted subtree reads for the ragged vlen readers: a `subtree=` keyword on the three sweep readers (`read_raw_values`, `read_locations`, `read_tensors`), implementing the normative spec §1.5 "Subtree spans" clause (`docs/specification.md`, landed on main in c0493ba). Binds to the [implementation plan](https://github.com/englacial/zagg/issues/351#issuecomment-5146756732), the [ratified decision record](https://github.com/englacial/zagg/issues/351#issuecomment-5146868655) (both currencies; sub-chunk refusal in `read_tensors` v1; warn+empty for well-formed out-of-domain words; `ValueError` for malformed/impossible words; docstrings state the empty-yield ambiguity), and the [§4 module-cap waiver](https://github.com/englacial/zagg/issues/351#issuecomment-5147013259) (option (1): `tdigest_tensor.py` may cross ~1000 lines for this feature; no pre-split).

## Approach

- **Span identity** (`readers/_layout.py`, pure arithmetic — no store access): a cell's axis position is its HEALPix NESTED id — chunks are keyed by the parent's nested id (`HealpixGrid.block_index` / `mort2healpix`) and the in-chunk position is the nested rank — so the order-`k` subtree below a word with nested id `h` occupies `[h·4^d, (h+1)·4^d)` (`d = c − k`) in fullsphere coordinates. A single-root power-of-four axis (a hive leaf) subtracts its root's span start, with the root derived from one written anchor word exactly as `_load_occupancy` derives the shard word. Verified numerically before writing: `generate_morton_children(w, k)` yields exactly the ascending nested-id range.
- **Currency normalization** (`normalize_subtree`): packed area word (int) or decimal string (str), both normalized at the top of the call; a parsed string always yields the area word (mortie spec §4 tie-break), per ratified fork (1).
- **Fetch restriction**: the span is intersected with `_stored_chunk_spans` before any payload read (`_iter_populated_chunks(arr, span)` clips each stored object's slice to the covering read chunks and skips non-overlapping objects without a fetch) — the §1.5 read plan: index suffix + covering inner chunks on sharded, covering chunk objects on flat. Sub-chunk restriction is a rank filter inside the one covering chunk (per-cell readers only, per ratified fork (2)).
- **Out-of-domain** (ratified fork (3)): a well-formed word disjoint from the axis warns once per reader call — `subtree {word} is outside this axis' order-{k} root {root} — yielding nothing` — then yields nothing; malformed words and words deeper than the cells-axis order raise `ValueError`. Reader docstrings state the empty-yield ambiguity explicitly; tests pin the exact warning text.

## Phases

- [x] Phase 1 — span arithmetic + currency normalization in `readers/_layout.py`; span∩stored-spans in `_iter_populated_chunks`; `subtree=` on `read_raw_values` / `read_locations` (any order down to a single cell); tests: golden equality vs independently-filtered whole sweep on flat, sharded, and hive-leaf stores, both currencies, warn+empty (exact text, once per call), malformed/too-deep raises, `subtree=None` regression.
- [x] Phase 2 — `subtree=` on `read_tensors`: span-restricted block assembly, `subtree_order ≤ block_order ≤ chunk_order`, sub-chunk refusal pointing at the per-cell readers, occupancy-mask blocks unchanged; GET accounting with a counting store; empty/out-of-domain edges; hive-leaf case.
- [x] Phase 3 — flip the `docs/ragged_layout.md` third-access-mode bullet from "(planned)" to the landed API (spec §1.5 untouched — already normative).

## How it was tested

`uv run ruff check` + `ruff format --check` on the touched files, and `uv run pytest tests/test_readers.py tests/test_reader_layout.py` (104 passed by phase 3 = 76 baseline + 28 new; 113 after the review-fold commits) per phase commit. Full suite before the final push: `uv run pytest -q` — **3090 passed, 35 skipped, 1 failed**; the one failure is `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, the known environment-only failure (pip has no network in the sandbox) — pre-existing and unrelated to this change (the fold pass also observed an intermittent pre-existing timing race in `tests/test_client_transport.py::TestStatusPoller::test_invoke_fault_burns_an_attempt_and_retries`, present on the base commit too). The golden filter in the new tests is deliberately independent of the reader's span arithmetic: each yielded cell resolves through `cell_index` + the stored `morton` coordinate, then coarsens with `mortie.clip2order` — the two paths cannot share a bug.

Note: `ruff check src tests` on this branch reports one **pre-existing** `N818` (`src/zagg/registry.py:64`, `UnknownCapability`) under the locked ruff 0.16.1 — present on clean main (c0493ba), unrelated to this change, so it is flagged here rather than fixed (per repo conventions §4). The touched files are clean. Final `tdigest_tensor.py` size: **1,140 lines** after the review folds (the §4 waiver projected ~1,055–1,070; the overshoot is the decision-record-mandated docstring text plus the reviewer's nested-placement invariant check and single-LIST plumbing — flagging the delta for transparency).

## Questions for review

1. **Ancestor-of-the-root words clip to the whole axis rather than warn+empty.** The decision record's out-of-domain example is a *sibling* branch ("433… outside … root 431…" — disjoint, the moczarr no-coverage posture), and the acceptance criterion is `subtree read ≡ whole sweep filtered to descendants of w` — for a word *above* a leaf's root, every stored cell IS a descendant of `w`, so this implementation returns the span∩axis intersection (the whole leaf) and reserves warn+empty for genuinely disjoint words. This keeps the golden equality universal over well-formed words (pinned in `test_ancestor_of_the_axis_root_clips_to_the_whole_axis`). If the intent was instead "any word not at-or-below the root warns and yields nothing", say so and phase 2 will flip it (one clamp + one test).
2. **Empty store**: with no stored objects there is no anchor word, so neither the axis root nor the cells-axis order is derivable — a syntactically valid `subtree` yields nothing silently (no domain warning, no too-deep check), while malformed words still raise at parse time. Pinned in `test_empty_store_yields_nothing_but_malformed_still_raises`; flag if the too-deep check should instead consult some other order source.

https://claude.ai/code/session_01JnmnBoL5eTCGPMEYUtHrRK

### DIFF
--- a/src/zagg/readers/_layout.py
+++ b/src/zagg/readers/_layout.py
@@ -78,7 +78,14 @@ def normalize_subtree(subtree) -> tuple[int, str, int]:
 
 
 def subtree_cell_span(
-    subtree, anchor: int, anchor_index: int, cell_order: int, n_cells: int, field: str
+    subtree,
+    anchor: int,
+    anchor_index: int,
+    cell_order: int,
+    n_cells: int,
+    field: str,
+    *,
+    stacklevel: int = 2,
 ) -> tuple[int, int]:
     """Cell-index span ``[start, stop)`` of ``subtree`` on a nested cells axis.
 
@@ -102,11 +109,17 @@ def subtree_cell_span(
 
     The span is returned intersected with the axis: a word at or above the
     axis root clips to the whole axis (every stored cell is its
-    descendant), and a well-formed word DISJOINT from the axis warns once —
-    naming the word and the axis root — and returns the empty span
-    ``(0, 0)``. Out-of-domain is a data answer (the ratified issue #351
-    fork (3)), so an empty read is ambiguous between "in-domain, nothing
-    stored" and "outside the domain" unless the warning is observed.
+    descendant), and a well-formed word DISJOINT from the axis warns — naming
+    the word and the axis root — and returns the empty span ``(0, 0)``.
+    Out-of-domain is a data answer (the ratified issue #351 fork (3)), so an
+    empty read is ambiguous between "in-domain, nothing stored" and "outside
+    the domain" unless the warning is observed — and DELIVERY of that warning
+    is the caller's ``warnings`` filter's call: the default ``default`` action
+    dedups on (message, category, module, lineno), so a second call with the
+    same word is silent in the same process. The guarantee is "at most once
+    per call" (never per cell or per chunk). ``stacklevel`` is the caller's
+    depth from here — the readers pass the depth that attributes the warning
+    to THEIR caller rather than to zagg's own frames.
     Raises ``ValueError`` for a malformed word, or one deeper than the
     cells-axis order (grammatically impossible — nothing stored could be
     its descendant).
@@ -153,7 +166,7 @@ def subtree_cell_span(
         warnings.warn(
             f"subtree {decimal} is outside this axis' order-{root_order} root "
             f"{morton_decimal(root)} — yielding nothing",
-            stacklevel=2,
+            stacklevel=stacklevel,
         )
         return 0, 0
     return start, stop
--- a/src/zagg/readers/_layout.py
+++ b/src/zagg/readers/_layout.py
@@ -78,7 +78,7 @@ def normalize_subtree(subtree) -> tuple[int, str, int]:
 
 
 def subtree_cell_span(
-    subtree, anchor: int, cell_order: int, n_cells: int, field: str
+    subtree, anchor: int, anchor_index: int, cell_order: int, n_cells: int, field: str
 ) -> tuple[int, int]:
     """Cell-index span ``[start, stop)`` of ``subtree`` on a nested cells axis.
 
@@ -91,6 +91,14 @@ def subtree_cell_span(
     hive leaf's shard subtree) subtracts its root's own span start, with the
     root derived from ``anchor`` — any WRITTEN cell word (the morton anchor
     slice the caller already read); no store bytes are touched here.
+
+    That identity is CHECKED, not assumed: ``anchor`` sits at ``anchor_index``
+    on the axis, so ``nested(anchor) - root_start`` must equal ``anchor_index``
+    or the axis is not in canonical nested placement, and this arithmetic
+    would hand back plausible-but-wrong cells silently — the single-root
+    geometry is INFERRED from ``4**depth == n_cells`` alone, and this is the
+    first reader that derives absolute placement instead of searching for it
+    (review, PR #357). A mismatch raises ``ValueError``.
 
     The span is returned intersected with the axis: a word at or above the
     axis root clips to the whole axis (every stored cell is its
@@ -119,14 +127,27 @@ def subtree_cell_span(
     lo = int(h) * 4**d
     n = int(n_cells)
     depth = (n.bit_length() - 1) // 2
-    if 4**depth != n:
-        # A fullsphere axis (12·4^c cells): the nested id IS the axis
-        # position, and every well-formed word's span lies inside it.
+    # A fullsphere axis (12·4^c cells) starts at 0 — the nested id IS the axis
+    # position, and every well-formed word's span lies inside it. A single-root
+    # power-of-four axis starts at its root's own span start.
+    single_root = 4**depth == n
+    root_order, root, root_start = cell_order - depth, 0, 0
+    if single_root:
+        root = int(clip2order(root_order, np.asarray([anchor], dtype=np.uint64))[0])
+        (rh,), _ro = mort2healpix(np.asarray([root], dtype=np.uint64))
+        root_start = int(rh) * 4**depth
+    (ah,), _ao = mort2healpix(np.asarray([anchor], dtype=np.uint64))
+    if int(ah) - root_start != int(anchor_index):
+        raise ValueError(
+            f"{field!r}'s cells axis is not in canonical nested placement: cell "
+            f"{int(anchor_index)} carries morton {morton_decimal(int(anchor))}, whose "
+            f"nested id puts it at axis position {int(ah) - root_start}. A subtree "
+            f"span is derived arithmetically (cell position == nested id minus the "
+            f"axis root's start), so this axis would yield the wrong cells"
+        )
+    if not single_root:
         return lo, lo + 4**d
-    root_order = cell_order - depth
-    root = int(clip2order(root_order, np.asarray([anchor], dtype=np.uint64))[0])
-    (rh,), _ro = mort2healpix(np.asarray([root], dtype=np.uint64))
-    lo -= int(rh) * 4**depth
+    lo -= root_start
     start, stop = max(lo, 0), min(lo + 4**d, n)
     if start >= stop:
         warnings.warn(
--- a/src/zagg/readers/_layout.py
+++ b/src/zagg/readers/_layout.py
@@ -21,13 +21,23 @@ i.e. the texture's slow axis ``i`` feeds ``bit_combine``'s second (odd-bit,
 — so a zagg tensor uploads to a gridlook texture with no further shuffle.
 ``tensor[0, 0]`` is the subtree's south corner; rows advance toward the
 north-west edge, columns toward the north-east edge.
+
+Subtree spans (issue #351)
+--------------------------
+The same nested ordering makes "everything below one morton node" a
+contiguous cell-index span (normative: zagg ``docs/specification.md`` §1.5).
+:func:`normalize_subtree` and :func:`subtree_cell_span` are the pure
+arithmetic behind the readers' ``subtree=`` keyword — no store access.
 """
 
 from __future__ import annotations
 
+import warnings
+
+import numpy as np
 from mortie import rank_to_xy, xy_to_rank
 
-__all__ = ["rank_to_rowcol", "rowcol_to_rank"]
+__all__ = ["rank_to_rowcol", "rowcol_to_rank", "normalize_subtree", "subtree_cell_span"]
 
 
 def rank_to_rowcol(rank, depth: int):
@@ -49,3 +59,80 @@ def rowcol_to_rank(row, col, depth: int):
     (scalars or arrays; values must be ``< 2**depth``).
     """
     return xy_to_rank(col, row, depth)
+
+
+def normalize_subtree(subtree) -> tuple[int, str, int]:
+    """Normalize a ``subtree=`` argument to ``(word, decimal, order)``.
+
+    Both ratified currencies (issue #351): a packed morton AREA word
+    (``int``) or a decimal morton string (``str``). A parsed string always
+    yields the area word (mortie's spec §4 parse tie-break), so no kind
+    flag is needed. Raises ``ValueError`` on a malformed member of either
+    currency (an unparseable string, an invalid packed word).
+    """
+    from zagg.grids.morton import morton_decimal, morton_word
+
+    word = morton_word(subtree) if isinstance(subtree, str) else int(subtree)
+    decimal = morton_decimal(word)  # raises on an invalid/negative packed word
+    return word, decimal, len(decimal) - (2 if decimal.startswith("-") else 1)
+
+
+def subtree_cell_span(
+    subtree, anchor: int, cell_order: int, n_cells: int, field: str
+) -> tuple[int, int]:
+    """Cell-index span ``[start, stop)`` of ``subtree`` on a nested cells axis.
+
+    The spec §1.5 subtree-span identity, resolved by pure arithmetic: a
+    cell's axis position is its HEALPix NESTED id (chunks are keyed by the
+    parent's nested id — ``HealpixGrid.block_index`` — and the in-chunk
+    position is the nested rank), so the order-``k`` subtree below a word
+    with nested id ``h`` occupies ``[h·4^d, (h+1)·4^d)`` (``d = cell_order
+    - k``) in fullsphere coordinates. A single-root power-of-four axis (a
+    hive leaf's shard subtree) subtracts its root's own span start, with the
+    root derived from ``anchor`` — any WRITTEN cell word (the morton anchor
+    slice the caller already read); no store bytes are touched here.
+
+    The span is returned intersected with the axis: a word at or above the
+    axis root clips to the whole axis (every stored cell is its
+    descendant), and a well-formed word DISJOINT from the axis warns once —
+    naming the word and the axis root — and returns the empty span
+    ``(0, 0)``. Out-of-domain is a data answer (the ratified issue #351
+    fork (3)), so an empty read is ambiguous between "in-domain, nothing
+    stored" and "outside the domain" unless the warning is observed.
+    Raises ``ValueError`` for a malformed word, or one deeper than the
+    cells-axis order (grammatically impossible — nothing stored could be
+    its descendant).
+    """
+    from mortie import clip2order, mort2healpix
+
+    from zagg.grids.morton import morton_decimal
+
+    word, decimal, order = normalize_subtree(subtree)
+    if order > cell_order:
+        raise ValueError(
+            f"subtree {decimal} is order {order} — deeper than {field!r}'s "
+            f"order-{cell_order} cells axis, so nothing stored can be its descendant "
+            f"(a subtree names an ancestor, at any order up to the cell order)"
+        )
+    d = cell_order - order
+    (h,), _o = mort2healpix(np.asarray([word], dtype=np.uint64))
+    lo = int(h) * 4**d
+    n = int(n_cells)
+    depth = (n.bit_length() - 1) // 2
+    if 4**depth != n:
+        # A fullsphere axis (12·4^c cells): the nested id IS the axis
+        # position, and every well-formed word's span lies inside it.
+        return lo, lo + 4**d
+    root_order = cell_order - depth
+    root = int(clip2order(root_order, np.asarray([anchor], dtype=np.uint64))[0])
+    (rh,), _ro = mort2healpix(np.asarray([root], dtype=np.uint64))
+    lo -= int(rh) * 4**depth
+    start, stop = max(lo, 0), min(lo + 4**d, n)
+    if start >= stop:
+        warnings.warn(
+            f"subtree {decimal} is outside this axis' order-{root_order} root "
+            f"{morton_decimal(root)} — yielding nothing",
+            stacklevel=2,
+        )
+        return 0, 0
+    return start, stop
--- a/src/zagg/readers/tdigest_tensor.py
+++ b/src/zagg/readers/tdigest_tensor.py
@@ -61,6 +61,12 @@ anchors a fixed ``n_bins * resolution`` window at the floor of that range
 counts via :func:`zagg.stats.tdigest.cdf_from_tdigest`, and decodes the hive
 leaf's ``coverage.moc`` occupancy sidecar (when present) into the uint8
 ``mask`` channel on the same deinterleaved layout.
+
+All three sweep readers accept ``subtree=`` (issue #351): the nested cells
+axis makes "everything below one morton ancestor" a contiguous cell span
+(spec §1.5 subtree spans), so a restricted read fetches only the stored
+objects the span covers — on a sharded store the index suffix plus the
+covering inner chunks, on the flat layout the covering chunk objects.
 """
 
 from __future__ import annotations
@@ -594,6 +600,7 @@ def read_tensors(
     fit: FitMode = "raise",
     dtype: TensorDtype = "uint32",
     block_order: int | None = None,
+    subtree: int | str | None = None,
     max_block_bytes: int = 2 * 1024**3,
     zarr_format: Literal[2, 3] = 3,
 ) -> Iterator[tuple[np.ndarray, np.ndarray, tuple[float, float], int]]:
@@ -660,7 +667,24 @@ def read_tensors(
     block_order : int, optional
         HEALPix order of the emitted blocks (default ``None`` — one block per
         read chunk). Must be at or coarser than the chunk order; a block is
-        assembled from whole read chunks with one shared z-window.
+        assembled from whole read chunks with one shared z-window. With a
+        ``subtree`` the composed bound is ``subtree_order <= block_order <=
+        chunk_order``, so blocks tile the subtree.
+    subtree : int or str, optional
+        Restrict the sweep to the read chunks below this morton ancestor —
+        packed area word or decimal string, as in :func:`read_raw_values`
+        (issue #351; spec §1.5): only stored objects overlapping the
+        subtree's cell span are fetched. The subtree must be at or coarser
+        than the READ-CHUNK order — a finer word raises, pointing at the
+        per-cell readers, because a sub-chunk block would re-derive its
+        z-window from fewer cells and stop being a slice of the chunk tensor
+        (the ratified v1 refusal; recover a sub-chunk region client-side as
+        a corner slice of the chunk tensor instead, keeping its shared
+        ``(offset, gain)``). A well-formed word disjoint from this axis
+        warns once per call (naming the word and the axis root) and yields
+        nothing — so an **empty yield is ambiguous** between "in-domain,
+        nothing stored" and "outside the domain"; the warning is the only
+        discriminator. Malformed / too-deep words raise ``ValueError``.
     max_block_bytes : int, optional
         Refuse a block whose emitted tensor would exceed this many bytes
         (default 2 GiB) — the guard on the ``block_order`` footgun. Raise it
@@ -701,7 +725,8 @@ def read_tensors(
         On an unknown ``dtype``/``fit``, a store missing the ragged element
         attrs or the ``morton`` sibling, an out-of-range ``block_order``, a
         block tensor over ``max_block_bytes``, a corrupt/misaligned occupancy
-        sidecar, or (with ``fit="raise"``) a block whose trimmed range
+        sidecar, a ``subtree`` finer than the read chunks (or malformed /
+        too-deep), or (with ``fit="raise"``) a block whose trimmed range
         overflows the fixed window.
     """
     if dtype not in _TENSOR_DTYPES:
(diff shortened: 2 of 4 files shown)


src/zagg/readers/tdigest_tensor.py
```diff
@@ -337,17 +348,54 @@ def _iter_populated_chunks(arr) -> Iterator[tuple[int, list]]:
...
+    nothing (a disjoint word — :func:`~zagg.readers._layout.subtree_cell_span`
+    already warned — or an empty store, which has nothing below any word).
+    The anchor is the first stored span's first read chunk of the ``morton``
+    coordinate — one small slice, no payload bytes (the dense coordinate
+    write covers every chunk of a stored span). A malformed ``subtree``
+    raises even on an empty store.
+    """
+    if subtree is None:
+        return None
+    spans = _stored_chunk_spans(arr)
```
espg, line 407: 🤖 *from Claude (review)*

[should-fix] Every `subtree=` read LISTs the array's data keys **twice**: `_subtree_span` calls `_stored_chunk_spans(arr)` here, and `_iter_populated_chunks(arr, span)` calls it again a few lines later (all three readers go through both). On a hive leaf that is one key and free; on the flat fullsphere path it is a paginated LIST per ~1000 shard objects - the one whole-array-scale operation left on this read path, in a feature whose framing is "never a whole-array sweep".

It is invisible to the new GET-accounting tests because they filter to `"/h_tdigest/c/" in g[0]` **GETs**, and `list_prefix` is not a GET.

Cheap fix: have `_subtree_span` return `(span, spans)` (or accept the spans), and let `_iter_populated_chunks` take an optional pre-listed `spans` and only list when it gets none. Three call sites, all in this module.

src/zagg/readers/tdigest_tensor.py
```diff
@@ -337,17 +348,54 @@ def _iter_populated_chunks(arr) -> Iterator[tuple[int, list]]:
...
+    if not spans:
+        normalize_subtree(subtree)  # malformed still raises; nothing to visit
+        return 0, 0
+    words = morton[spans[0][0] : spans[0][0] + int(arr.chunks[0])]
+    cell_order = _cells_order(words, field, spans[0][0])
+    anchor = int(words[words != 0][0])
```

[should-fix] The identity the whole feature rests on - *axis position == nested id minus axis-root start* - is checkable right here and never checked. At this line you hold both sides: the anchor's nested id (`mort2healpix(anchor)`) and the anchor's absolute axis index (`spans[0][0] + int(np.flatnonzero(words)[0])`). I verified the identity holds on all four fixture geometries (flat, sharded flat, leaf, sharded leaf), so a guard would be a pure invariant, not a behavior change.

Why it is worth two lines: `subtree_cell_span` infers "single-root axis" purely from `4**depth == n_cells`. That inference is sound for today's writers (`HealpixGrid` accepts only `layout="fullsphere"`, and a leaf axis is exactly one shard subtree) - but it is an inference, and when it is wrong the failure mode is *silently wrong cells*, not an error: a hypothetical K-shard axis with K a power of four would derive a bogus root from the anchor and hand back a plausible span. The existing guards do not cover it - `_chunk_word` only checks that cells **within** one span share an ancestor, and `cell_index` locates a chunk by **content** rather than arithmetic. This is the first reader that *derives* absolute placement instead of searching for it.

Suggested: raise a pointed `ValueError` when `nested(anchor) - root_start != anchor_index`, with one test that stamps a shifted `morton` coordinate and pins the message.

src/zagg/readers/tdigest_tensor.py
```diff
@@ -618,7 +667,24 @@ def read_tensors(
     block_order : int, optional
         HEALPix order of the emitted blocks (default ``None`` — one block per
         read chunk). Must be at or coarser than the chunk order; a block is
-        assembled from whole read chunks with one shared z-window.
+        assembled from whole read chunks with one shared z-window. With a
+        ``subtree`` the composed bound is ``subtree_order <= block_order <=
+        chunk_order``, so blocks tile the subtree.
```

[nit] This bound is not the implemented one when `subtree` is an **ancestor** of the axis root (the case Question 1 in the PR body keeps as clip-to-axis). `min_order` is derived from the CLIPPED span, so on a hive leaf with an order-6 root the effective floor is the ROOT order, not the subtree order. Reproduced on the `_sharded_leaf_store` fixture with an order-4 ancestor word:

```
block_order=4: RAISE ... must be between 6 and the chunk order 8
block_order=5: RAISE ... must be between 6 ...
block_order=6: 1 block
block_order=7: 2 blocks
block_order=8: 4 blocks
```

The behavior is correct - an order-5 block is 16,384 cells against a 4,096-cell axis, which the pre-existing `arr.shape[0] % block_cells` check would also reject, just with a different message. Only the sentence is wrong. Accurate: `max(subtree_order, axis_root_order) <= block_order <= chunk_order`, or in words "at or inside the visited span". `docs/ragged_layout.md` carries the same phrasing and wants the same tightening.

src/zagg/readers/_layout.py
```diff
@@ -49,3 +59,80 @@ def rowcol_to_rank(row, col, depth: int):
...
+    root_order = cell_order - depth
+    root = int(clip2order(root_order, np.asarray([anchor], dtype=np.uint64))[0])
+    (rh,), _ro = mort2healpix(np.asarray([root], dtype=np.uint64))
+    lo -= int(rh) * 4**depth
+    start, stop = max(lo, 0), min(lo + 4**d, n)
+    if start >= stop:
+        warnings.warn(
+            f"subtree {decimal} is outside this axis' order-{root_order} root "
+            f"{morton_decimal(root)} — yielding nothing",
+            stacklevel=2,
```
espg: 🤖 *from Claude (review)*

[nit] Two small things about this warning, both about what the docstrings promise:

(1) `stacklevel=2` attributes it to `_subtree_span` in `readers/tdigest_tensor.py` - zagg's own frame, never the caller's. Since the readers are generators the warning fires during the consumer's first `next()`, so reaching user code is ~4 frames up. Anything filtering by `module=`, or a human reading the printed `file:line`, gets pointed at the library.

(2) "warns once per reader call" holds only under `always`/`error` filters. Under Python's DEFAULT filter the `default` action dedups on (message, category, module, lineno), so the SECOND reader call in the same process with the same disjoint word is silent - a walker looping over leaves with one wrong word gets one warning for the whole process, not one per call. Both new tests set `simplefilter("always")`, so nothing pins the real-world behavior.

The ratified requirement (never one warning per cell or per chunk) is met either way; this is about the three docstrings saying "warns once per call" flatly. "At most once per call, subject to the caller's warning filters" would be exact.

tests/test_readers.py
```diff
@@ -561,6 +644,380 @@ def test_value_only_field_raises_clearly(self):
...
+        store = MemoryStore()
+        _grid_, shard6, _t = _sharded_store(store, populate={0, 3, 7, 12})
+        # Order-7 first child of the shard = inner chunks 0..3 (locals 0, 3
+        # populated). Per-chunk blocks AND one assembled order-7 block must
+        # both equal the filtered whole sweep at the same block_order.
+        sub = int(np.asarray(generate_morton_children(shard6, 7))[0])
+        per_chunk = self._filtered(list(read_tensors(store, "12/h_tdigest")), sub, 7)
+        assert len(per_chunk) == 2
+        self._assert_same(list(read_tensors(store, "12/h_tdigest", subtree=sub)), per_chunk)
+        assembled = self._filtered(list(read_tensors(store, "12/h_tdigest", block_order=7)), sub, 7)
```
espg, line 979: 🤖 *from Claude (review)*

[nit] Coverage gap: no test drives a span covering **more than one stored object**. Every subtree case selects inside one object (`_sharded_store`, `_sharded_leaf_store`, `_leaf_store`) or exactly one whole object (`subtree=_KEY_A` on the flat store), so the clip loop is exercised either as "skip" or as "clip within one object" - never both in one read - and `read_tensors` never assembles a block ACROSS stored objects under a subtree.

It works (I ran it before filing this), but nothing would catch a regression. Recipe on the flat K==1 fixture, using an order-5 parent of two sibling shards:

```python
parent = "112112"          # children at order 6: 1121121..1121124
# _write_shard(grid, store, "1121121", ...) and "1121122"
read_raw_values(store, "12/h_tdigest", subtree=parent)               # both shards' cells
read_tensors(store, "12/h_tdigest", subtree=parent)                  # 2 per-chunk blocks
read_tensors(store, "12/h_tdigest", subtree=parent, block_order=5)   # ONE (128, 128, 128) block
```

The last line is the interesting one: `min_order` lands at exactly 5, `_chunk_word` must resolve a block spanning two stored objects, and the emitted block is bit-identical to the same block from the unrestricted `block_order=5` sweep.

tests/test_readers.py
```diff
@@ -561,6 +644,380 @@ def test_value_only_field_raises_clearly(self):
...
+        with warnings.catch_warnings(record=True) as rec:
+            warnings.simplefilter("always")
+            out = list(read_raw_values(store, "12/h_tdigest", subtree=morton_word(_KEY_B)))
+        assert out == []
+        msgs = [str(w.message) for w in rec if "outside this axis" in str(w.message)]
+        assert msgs == [
+            f"subtree {_KEY_B} is outside this axis' order-6 root {_KEY_A} — yielding nothing"
+        ]
+
+    def test_ancestor_of_the_axis_root_clips_to_the_whole_axis(self):
```
espg, line 794: 🤖 *from Claude (review)*

[question] On Question 1 in the PR body (ancestor words clip to the axis instead of warn+empty) - no change requested; recording the reviewer read so the ruling has one place to land.

The two ratified clauses genuinely conflict for an ancestor: the decision record's "a well-formed word outside the axis root's subtree warns and yields nothing" reads literally on an ancestor (it is not *below* the root), while the acceptance criterion "subtree read is the whole sweep filtered to descendants of w" demands the whole axis (every stored cell IS a descendant of an ancestor). The implementation picks the acceptance criterion, which is the composable one - an AOI walker that coarsens to a common ancestor before descending never has to special-case "my word is above this leaf's root".

Worth noting the prose already matches the implementation, so keeping it costs no doc churn: spec 1.5 says nothing about domain, and both `docs/ragged_layout.md` and the three reader docstrings say "disjoint from the axis", not "outside the root's subtree". If the ruling goes the other way, the docs flip with the clamp. Either way the `block_order` bound sentence needs the fix in the separate comment on `read_tensors`' docstring.

src/zagg/readers/tdigest_tensor.py
```diff
@@ -337,17 +348,54 @@ def _iter_populated_chunks(arr) -> Iterator[tuple[int, list]]:
...
+    The anchor is the first stored span's first read chunk of the ``morton``
+    coordinate — one small slice, no payload bytes (the dense coordinate
+    write covers every chunk of a stored span). A malformed ``subtree``
+    raises even on an empty store.
+    """
+    if subtree is None:
+        return None
+    spans = _stored_chunk_spans(arr)
+    if not spans:
+        normalize_subtree(subtree)  # malformed still raises; nothing to visit
```
espg, line 409: 🤖 *from Claude (review)*

[question] On Question 2 in the PR body (empty store: a too-deep word yields silently instead of raising) - there is a partial answer that needs no store read, and the question reads as if the derivation were impossible in general when it is only impossible on one geometry.

On the FULLSPHERE axis the cells order is recoverable from `arr.shape[0]` alone (`n == 12 * 4**c`), so the "deeper than the cells axis" check could still fire on an empty fullsphere store. On a single-root leaf axis (`n == 4**depth`) it genuinely cannot - `depth` is known but the root order is not, so `cell_order` stays underivable without an anchor.

Options: (1) leave as is - uniformly silent on an empty store, simplest, pinned today by `test_empty_store_yields_nothing_but_malformed_still_raises`; or (2) derive `cell_order` from the shape on the fullsphere branch only, buying the grammar check on the geometry where a hand-typed too-deep word is most likely, at the cost of the two geometries behaving differently on an empty store. My read is (1) - the malformed-word raise already catches the common typo, and geometry-dependent error behavior is worse than a uniformly silent empty read.

espg:
🤖 *from Claude (review)*

Adversarial review of f4b76aa (fresh context). **No blocking findings** — the span arithmetic is correct, including the cases most likely to be wrong. Seven inline comments: two `[should-fix]` (a doubled LIST per subtree read; an unverified placement invariant), three `[nit]` (a docstring bound that is wrong for ancestor words, the warning's `stacklevel` / "once per call" claim, one test-coverage gap), and two `[question]` answering the PR body's own open questions.

What was verified rather than read, so the fold does not have to redo it:

- **Span identity, fullsphere** — for random words at orders 0–3 across positive and negative base cells (`1`..`6`, `-1`..`-6`), `subtree_cell_span` equals the exact `[min, max+1)` of `mort2healpix(generate_morton_children(w, c))`, and those ids are contiguous and ascending. Order-0 words and negative-face words both land correctly, so the `len(decimal) - (2 if "-" else 1)` order derivation is right (mortie renders base cells 7–12 with the sign, one digit either way).
- **Span identity, single-root leaf** — 189 words on a 4,096-cell order-4-rooted axis (self, ancestors, all descendants at orders 5–10, disjoint siblings, foreign base cells) against an independent golden (`clip2order` over the axis' own child words). Zero mismatches; the warning fires on exactly the disjoint cases and never on an in-domain one.
- **Multi-object spans** (untested in the PR — see the test comment): an order-5 word over two sibling shards on the flat fixture returns both shards' cells, and `block_order=5` assembles ONE (128, 128, 128) block across two stored objects that is bit-identical to the same block from the unrestricted sweep.
- **Sub-chunk rank filter on the sharded geometry** — order-9, order-10 and order-12 words against a 256-cell inner chunk return exactly the golden cells at 2 data GETs each (index suffix + one ranged inner chunk), and an order-12 word on an empty rank returns nothing.
- **Chunk alignment** — the round-down / round-up clip in `_iter_populated_chunks` was checked against spans starting before, inside, and after a stored object, and against the `(0, 0)` sentinel (which cannot collide with a real span, since a real span always has `start < stop`).
- **The anchor's morton invariant** — `_subtree_span` reads the first read chunk of the first stored object, which need not be *populated*; that is safe because `_build_output` builds a cell-complete carrier for every inner chunk, so both sharded writers (`write_shard_to_zarr`, `write_leaf_to_zarr`) write the `morton` coordinate over the whole shard, not just over chunks with payloads. Worth knowing that the docstring's "the dense coordinate write covers every chunk of a stored span" rests on that, not on the weaker "every chunk with payloads has a word" the rest of the module needs.
- **`subtree=None`** is unchanged on all three readers — `_subtree_span` returns before listing anything, and `_iter_populated_chunks`' only other edit is a local rename.
- Locally: `pytest tests/test_readers.py tests/test_reader_layout.py` → 104 passed; `ruff check` / `ruff format --check` clean on the three touched files.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Fixed in `812424b`. `_subtree_span` now returns `(span, spans)` — the `_stored_chunk_spans` listing it resolved the span against — and `_iter_populated_chunks` takes an optional pre-listed `spans=`, listing itself only when it gets none. All three readers thread it through, so a `subtree=` read LISTs the array once:

```python
span, spans = _subtree_span(arr, morton, field, subtree)
...
for start, populated in _iter_populated_chunks(arr, span, spans):
```

`subtree=None` is byte-identical: `_subtree_span` returns `(None, None)` without listing, and `_iter_populated_chunks` takes its own listing exactly as before.

Pinned by a new accounting test — `_CountingStore` now records `list_prefix` calls alongside GETs (a LIST is not a GET, which is why the existing tests could not see this), and `test_subtree_read_lists_the_data_keys_once` asserts `store.lists.count("12/h_tdigest/c/") == 1` for both `read_tensors` and `read_raw_values`. It fails at 2 on the pre-fix code.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Fixed in `980e4ee`. `subtree_cell_span` now takes the anchor's absolute axis index and checks the identity before deriving anything:

```python
(ah,), _ao = mort2healpix(np.asarray([anchor], dtype=np.uint64))
if int(ah) - root_start != int(anchor_index):
    raise ValueError(
        f"{field!r}'s cells axis is not in canonical nested placement: cell "
        f"{int(anchor_index)} carries morton {morton_decimal(int(anchor))}, whose "
        f"nested id puts it at axis position {int(ah) - root_start}. A subtree "
        f"span is derived arithmetically (cell position == nested id minus the "
        f"axis root's start), so this axis would yield the wrong cells"
    )
```

`_subtree_span` threads the index through as `spans[0][0] + int(np.flatnonzero(words)[0])`, exactly as suggested.

Two deliberate choices worth flagging:

1. The check runs on **both** geometries, not just the single-root one. `root_start` is `0` on a fullsphere axis (the nested id IS the axis position, which the code already asserted in a comment) and the root's own span start on a single-root axis, so one line covers both — and it makes the fullsphere claim checked rather than commented. It holds on all four fixture geometries, matching what you measured.
2. The message names the misplacement rather than the geometry inference, since the same identity is what breaks in either case.

Test: `test_misplaced_morton_coordinate_raises`, parametrized over `fullsphere` / `leaf`. It rolls a stored span's `morton` coordinate by one cell — which leaves every other guard satisfied (the rolled cells still share their chunk ancestor, so `_chunk_word` is happy) and would otherwise hand back a plausible-but-wrong span — and pins `"not in canonical nested placement"`.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Fixed in `654c13a` — sentence only, no behavior change. The `block_order` parameter docs now read:

> With a `subtree` the floor is the order of the span actually VISITED — the subtree CLIPPED to this axis — so the composed bound is `max(subtree_order, axis_root_order) <= block_order <= chunk_order`: blocks tile the visited span, and a word above the axis root (which clips to the whole axis) takes the ROOT's order as its floor, not its own.

`docs/ragged_layout.md` carries the same correction.

Also added `test_ancestor_word_block_order_floor_is_the_axis_root` on `_sharded_leaf_store`, which pins your reproduction: an order-4 ancestor word with `block_order=5` raises `must be between 6 and the chunk order 8`, and `block_order=6` (the root's order) reads one block. The in-code error message already stated the bound correctly — it derives `min_order` from the clipped span — so only the prose moved.

src/zagg/readers/_layout.py
espg: 🤖 *from Claude*

Both taken in `ec38291` — the cheap fix for (1), the exact wording for (2), no re-delivery machinery.

**(1) stacklevel.** `subtree_cell_span` grew a keyword-only `stacklevel: int = 2` (correct for a direct call of the pure function), and `_subtree_span` passes `4`:

```python
    # The disjoint-word warning belongs to the READER's caller: from
    # subtree_cell_span that is here (2), the reader's own frame (3) —
    # a generator, so it runs on the consumer's first next() — and the
    # consumer (4). Anything less points inside zagg.
    stacklevel=4,
```

That depth is uniform across all three readers (each calls `_subtree_span` from its own body), and it survives `list(...)` as the consumer because C frames do not count toward `stacklevel`. Pinned rather than asserted: `test_out_of_domain_warning_is_attributed_to_the_caller`, parametrized over all three readers, asserts `w.filename == __file__` on the recorded warning.

**(2) filter caveat.** The three reader docstrings now say "warns **at most once per call** … and whether it is DELIVERED is the caller's `warnings` filter's call (the default action dedups a repeat of the same warning in the same process)". `subtree_cell_span`'s own docstring states it in full (dedup key `(message, category, module, lineno)`), and `docs/ragged_layout.md` carries the same sentence. The ratified requirement — never one warning per cell or per chunk — is unchanged and is what the docstrings now claim, exactly.

Left as-is deliberately: nothing forces re-delivery (no `simplefilter` call inside the library, no per-call registry reset). A library that overrides the caller's filters is worse than one whose docstring is honest about them.

tests/test_readers.py
espg: 🤖 *from Claude*

Added in `8d9d422` — `test_span_covering_two_stored_objects`, your recipe on the flat K==1 fixture. Three shards written: `1121121` and `1121122` (the two populated siblings under order-5 parent `112112`) plus `2431123` as a disjoint object that must never be fetched, so the clip loop skips one object and clips two covered ones in the same read.

Asserted, in order:

- `read_raw_values(..., subtree="112112")` equals the whole sweep filtered to descendants of the parent (the independent `cell_index` + `clip2order` filter), 4 cells across both objects;
- `read_tensors(..., subtree="112112")` yields both objects' chunks — `[o[3] for o in per_chunk] == [words[_KEY_A], words[sibling]]` — and equals the filtered per-chunk sweep;
- `block_order=5` yields exactly ONE `(128, 128, 128)` block, bit-identical (tensor, mask, `(offset, gain)`, morton id) to the same block from the unrestricted `block_order=5` sweep — the case where `_chunk_word` resolves a block spanning two stored objects and the z-window is shared across them.

tests/test_readers.py
espg: 🤖 *from Claude*

Recorded — no code change, and the ruling stays open for review (the PR body asks it as Question 1).

Nothing in this fold moves the clamp: the implementation still picks the acceptance criterion (an ancestor word clips to the whole axis), and `test_ancestor_of_the_axis_root_clips_to_the_whole_axis` still pins it. The one adjacent change is the `block_order` bound sentence you flagged separately, folded in `654c13a`, which now states the floor as `max(subtree_order, axis_root_order)` — that phrasing is a consequence of the clamp, so if the ruling goes the other way it flips with it, alongside the prose in `docs/ragged_layout.md` and the three docstrings.

src/zagg/readers/tdigest_tensor.py
espg: 🤖 *from Claude*

Left by design — option (1), your own read, for your stated reason: geometry-dependent error behavior on an empty store is worse than a uniformly silent empty read, and the malformed-word raise (which fires even on an empty store) already catches the common typo. `test_empty_store_yields_nothing_but_malformed_still_raises` keeps pinning both halves.

Your correction to the framing is fair and stands: the too-deep check is not universally underivable, only underivable on a single-root leaf axis (`n == 4**depth` gives `depth` but not the root order, so `cell_order` needs an anchor). It is derivable on a fullsphere axis from `arr.shape[0]` alone. I have not rewritten the PR body's Question 2 — the ruling is @-free but is still yours to make, and the record of the correction lives here.

The reason not to take (2) even though it is two lines: the check would fire on a fullsphere store and stay silent on a leaf store for the identical mistake, and a caller cannot tell which store they hold without reading the shape. That asymmetry costs more than the raise buys.

espg:
🤖 *from Claude*

Adversarial-review fold complete — 5 findings implemented (one commit each), 2 answered without code. All replies are on their inline threads.

| # | Finding | Disposition |
|---|---|---|
| 1 | Doubled LIST per subtree read | fixed — `812424b` |
| 2 | Placement invariant never checked | fixed — `980e4ee` |
| 3 | `block_order` bound sentence wrong for an ancestor word | fixed — `654c13a` |
| 4 | Warning `stacklevel` + filter-dedup caveat | fixed — `ec38291` |
| 5 | No test for a span across >1 stored object | fixed — `8d9d422` |
| 6 | Ancestor-clipping (PR Question 1) | recorded, no code — ruling stays open |
| 7 | Empty store / too-deep check (PR Question 2) | left by design — option (1), the reviewer's own read |

What changed, briefly:

- **`812424b`** — `_subtree_span` returns `(span, spans)` and `_iter_populated_chunks` accepts a pre-listed `spans=`, so a restricted read LISTs the array's data keys once instead of twice. `subtree=None` is byte-identical. `_CountingStore` now records `list_prefix` too (a LIST is not a GET, which is why the existing accounting tests could not see it).
- **`980e4ee`** — `subtree_cell_span` takes the anchor's axis index and raises when `nested(anchor) - root_start != anchor_index`, on both geometries (`root_start` is 0 on a fullsphere axis). Test rolls a stored span's `morton` coordinate by one cell — which leaves every other guard satisfied — and pins the message.
- **`654c13a`** — the bound is stated as `max(subtree_order, axis_root_order) <= block_order <= chunk_order` in the docstring and `docs/ragged_layout.md`, with a test pinning the reviewer's order-4-ancestor reproduction. Behavior unchanged; the in-code error already derived the floor from the clipped span.
- **`ec38291`** — `stacklevel=4` from `_subtree_span` (the reader is a generator, so the warning fires on the consumer's first `next()`), pinned by asserting the recorded warning's `filename` is the caller's file for all three readers; docstrings now say "at most once per call, subject to the caller's warning filters". No re-delivery machinery.
- **`8d9d422`** — order-5 parent over two populated sibling order-6 shards plus a disjoint third: per-cell equality, both objects' chunks yielded, and one `(128, 128, 128)` `block_order=5` block assembled across the two stored objects, bit-identical to the filtered whole-sweep assembly.

Gates: `ruff check` / `ruff format --check` clean on the touched files; `tests/test_readers.py` + `tests/test_reader_layout.py` **113 passed** (up from 104 — 9 new tests, several parametrized).

Pre-existing and unrelated, flagged not fixed: `tests/test_client_transport.py::TestStatusPoller::test_invoke_fault_burns_an_attempt_and_retries` fails intermittently (~1 run in 3) on this branch **and on its base commit** `f4b76aa` — a timing race in the retry poller, untouched by this PR.

PR #360, opened 2026-07-31
declare-pyramid retrofit for existing stores (issue #358)

### LINKED ISSUE #358: declare-pyramid retrofit: install/update the manifest pyramid block on an existing store
🤖 *from Claude* (filed from the 0.41.0 benchmark-prep discussion with espg, 2026-07-31)

## Gap

The manifest `pyramid` block is written only at template time (`build_manifest` → `build_pyramid_block`, issue #201 / PR #344). A store templated without it — every pre-#344 store, including the production-scale cycle22 / 88S o9 / SERC stores — can never get overviews: the sweep no-ops with "no pyramid overview declaration in the manifest." Declaration is birth-only, which espg flagged as wrong: pyramids are regenerable caches (D9), so declaring them late should be a supported operation, not a manifest hand-edit.

## Why the retrofit is clean

1. The block is a pure function of the pipeline config + the manifest's own `shard_order` (`build_pyramid_block`) — no write-time state.
2. The `pyramid` manifest key is already the one key the sweep legally RMWs (D11: excluded from the frozen resume keys for the `materialized` update), so a retrofit write rides an existing, sanctioned seam.

## Tool sketch

`declare_pyramid(store_root, config, *, store_kwargs)` + a CLI entry (natural home: the `python -m zagg.sweep` CLI, e.g. `--declare-pyramid <config.yaml>`):

- Derive the block from the supplied pipeline config (the authoritative source — same code path as template time). v1 requires the config; deriving from the store's D19 `aggregation.yaml` semantic core alone can be a follow-up.
- **Validate against the store before writing**: every declared field must exist as an array whose dtype / ragged element / inner_shape matches the declaration — refuse pointedly on drift, so a wrong config cannot install a fold recipe the store contradicts.
- Idempotent RMW: identical declaration → no PUT; changed declaration → rewrite, preserving any existing `materialized` actuals.
- Re-declaration is the same operation (schedule changes now; per-field selection once #352 lands). Overviews at now-undeclared orders are left in place as regenerable-cache debris for the (stubbed) debris family — recorded absence in the manifest means readers never open them (D24 option A posture).
- `output.pyramid: false` configs retrofit the declared-off block — recording absence is also a valid retrofit.

## Immediate consumer

The pyramid-sweep benchmarks (#344 phases E/F): retrofitting the declaration onto an existing production-scale store + one synchronous `mode="sweep"` invoke benchmarks the overview fold in isolation against real digests (zagg 0.41.0 fleet, issue #353 telemetry), without re-running ingest.

Refs #201, PR #344, #352, #353.

espg:
Closes #358.

## What

The declare-pyramid retrofit tool: installs (or updates) the manifest `pyramid` declaration on an **existing** hive store, so pre-#344 stores — every production-scale store templated before PR #344 — can gain overviews without a manifest hand-edit. Declaration stops being birth-only; per the issue, the block is a pure function of the pipeline config + the manifest's own `shard_order`, and the `pyramid` key is already the one manifest key legally RMW'd (D11).

## Approach

- `declare_pyramid(store_root, config, *, store_kwargs=None) -> dict` in `src/zagg/sweep_overview.py`, next to the existing pyramid-block machinery. It derives the block via the **same template-time path** (`build_pyramid_block(config, manifest["shard_order"])` — no retrofit-specific grammar), validates it against store truth **before** writing, and RMWs the manifest.
- **Validation source**: one committed leaf found through the store's run records (`zagg.sweep.discover_leaves` — the sweep's own LIST-free discovery; probe bounded to the first 8 refs), read at the manifest `cell_order`. Leaves are the layer the overview fold actually reads, so they are the truth that can contradict a fold recipe; the D19 `aggregation.yaml` core is deliberately **not** used (itself config-derived, written fail-open). Checks per declared field: presence, dense dtype (exact class), D18 ragged element declaration — element dtype + `inner_shape` (approximate class). Drift refuses with a `ValueError` naming every drifted field; nothing is written. A store with **no committed leaf** is still declarable — field checks are skipped loudly and the returned summary says so.
- **Idempotent RMW**: identical existing declaration → no PUT (`summary["updated"] is False`); changed → one PUT of the whole manifest, **preserving** any `materialized` actuals (overviews at now-undeclared orders remain regenerable-cache debris — D24 option A recorded-absence posture). `output.pyramid: false` installs the declared-off block (`{"overview": {"orders": []}}`): recording absence is a valid retrofit.
- Unlike the sweep's fail-open `_update_manifest_pyramid`, every failure **raises** — this is the user's explicit operation (missing manifest, drift, orders outside `[0, shard_order)`).
- Summary dict returns: `orders`, `fields` (`{name: class}`), `validated` (what store truth was checked), `previous` (`absent`/`identical`/`replaced`), `updated` (whether a PUT happened).

Module size: `sweep_overview.py` lands at 1,194 lines after both review folds (`sweep.py` at 1,115); both under espg's standing sub-1,200 ruling ([issue #358 thread](https://github.com/englacial/zagg/issues/358#issuecomment-5147204385)) — but `sweep_overview.py` has only 6 lines of headroom left, flagged under "Questions for review".

## Phases

- [x] Phase 1 — `declare_pyramid` + store-truth validation + idempotent RMW + tests (8b7a4bd)
- [x] Phase 2 — `--declare-pyramid <config.yaml>` flag on `python -m zagg.sweep` + CLI tests (ca15733)
- [x] Phase 3 — docs: "Retrofitting the pyramid declaration" subsection in `docs/hive_layout.md` (the manifest section, where the pyramid-block narrative lives); no `docs/specification.md` change — the retrofit changes no wire format (2d35069)

## How it was tested

`tests/test_sweep_overview.py::TestDeclarePyramid` (12 tests, local hive fixtures reusing the existing `_write_manifest`/`_make_leaf`/`_leaf_cfg` helpers plus a real `write_run_parquet` run record for discovery):

- fresh install on a pre-#344-style manifest (no `pyramid` key): installed block `==` `build_pyramid_block(cfg, shard_order)` byte-for-byte, frozen resume keys untouched (`_frozen_matches`)
- idempotent second call: `obstore.put` monkeypatch-proven **zero** PUTs
- changed schedule rewrite preserving `materialized` (real actuals produced by a real overview sweep)
- drift refusals: dense dtype, ragged element dtype, store-side ragged `inner_shape` (the declared inner shape is pinned to `[2]` by composability, so the inner-shape drift that can occur is store-side — emulated by rewriting the stored D18 attr), absent declared field; each proves the manifest is never half-written
- declared-off retrofit (`output.pyramid: false`) + the sweep then no-ops on the recorded absence
- missing-manifest raise; declared-orders-outside-the-store raise; no-committed-leaf store still declarable with the validation skip recorded
- end-to-end: pre-#344 store → sweep no-ops (`declared: false`, the #358 gap) → `declare_pyramid` → `discover_leaves` → overview sweep materializes real overviews and populates `materialized`

A second adversarial review covered the phase-2 CLI and phase-3 docs; its six findings are folded in `f64e4b9` (empty `--declare-pyramid` no longer falls through to a writing sweep pass — a real bug), `34b6c70` (the declaration-only assertion now discriminates, plus a positive control), `6723c0d` (CLI refusal paths pinned), `bef7250` (stale pre-#201 manifest example corrected), `b1a8fad` (validation guarantee scoped to what is actually checked, now split into the semantics and typing layers), `52f3cd6` (two closing over-claims dropped).

Gates: `ruff check` + `ruff format --check` clean on touched files; `pytest tests/test_sweep_overview.py tests/test_sweep.py -q` → 167 passed after both folds; 163 after the phase-1 review fold (8 fold commits: semantic-hash guard, fresh-manifest RMW, element-dtype fail-closed, uncapped probe, docstring corrections, three gap tests, malformed-manifest refusal, canonical-JSON normalize). Full suite locally: 3,084 passed / 35 skipped pre-fold, 3,094 passed / 29 skipped (`-m "not slow"`) post-fold; the only failure either way is `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, environment-only (needs network).

## Questions for review

- **Pre-existing, unrelated lint**: `ruff check src` trips `N818` on `src/zagg/registry.py::UnknownCapability` under ruff ≥ 0.15 (local envs resolve 0.15/0.16), but not under the pre-commit-pinned v0.14.10 that CI mirrors. Not touched here per the don't-fix-unrelated rule — flagging for a separate small-fix (rename vs. per-file ignore vs. pin bump decision).
- **Validation probe bound** (resolved in the fold): the cap is gone — the probe now walks all discovered refs until the first committed leaf (run records are already fully parsed by then; one GET per ref), and the skip message distinguishes "no run-record refs" from "N refs, none committed". Both remain declarable per the issue contract.
- **Ragged `spec`-version check** — *ruled and folded* (espg, 2026-07-31): `_field_drift` now gates the D18 attrs' `spec` against `zagg.grids.base.RAGGED_SPEC` before the dtype/`inner_shape` checks, mirroring `readers.tdigest_tensor._open_ragged` (same constant, same fail-loud posture — not a second policy). Rationale: a store this zagg cannot fold is the purest instance of the tool's contract; without the gate the retrofit installs a declaration promising overviews the sweep then fails to read at fold time, later and with a worse error. `zagg-ragged/2` is a live migration path (issue #210). Test: `test_unreadable_ragged_spec_revision_refuses` (`fb29ac2`).
- **Module headroom**: `sweep_overview.py` is at **1,194** of the standing 1,200-line ceiling after both fold passes — 6 lines of headroom. If a follow-up grows it, the natural seam is the Phase-A kernels + block derivation vs. the Phase-B writer/sweep — flagging now so the split (if wanted) can be its own issue rather than a squeeze.


src/zagg/sweep_overview.py
```diff
@@ -297,6 +297,163 @@ def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
...
+
+    store_kwargs = dict(store_kwargs or {})
+    manifest = read_manifest(store_root, **store_kwargs)
+    if manifest is None:
+        raise ValueError(
+            f"no {MANIFEST_NAME} at {store_root} — not a hive store root; "
+            f"declare_pyramid retrofits existing hive stores only"
+        )
+    shard_order = int(manifest["shard_order"])
+    block = build_pyramid_block(config, shard_order)
```
espg, line 364: 🤖 *from Claude (review)*

[correctness] The validation cannot catch a wrong **reducer** — the store's own `semantic_hash` is never consulted.

`build_pyramid_block` puts the fold law into the declaration:

```python
fields[name] = {
    "class": "exact",
    "method": EXACT_MERGE_LAWS[_fold_function_name(meta.get("function")) or ""],
    ...
```

but `_field_drift` only checks presence + dtype + ragged `inner_shape`. Nothing in a leaf records which reducer produced a field (grep for `attributes=` in `src/zagg/grids/base.py`: the only per-array attrs block written is `RAGGED_ELEMENT_ATTR`). So a retrofit config that declares `h_min: {function: max, dtype: float32}` over a store whose leaves hold minima passes every check and installs `method: "max"` — the overview sweep then folds minima with `max` and writes silently wrong data. That is precisely the failure mode issue #358 asks the tool to make impossible ("refuse pointedly on drift, so a wrong config cannot install a fold recipe the store contradicts").

The store *does* record the reducer, in the one place the repo already calls authoritative: `semantic_hash` — a FROZEN manifest key written at `src/zagg/hive.py:318` from the same config, over a core that includes the whole `aggregation` block minus `AGGREGATION_PACKAGING_KEYS = ("handoff",)` (`semantics.semantic_core`), and documented at `hive.py:511` as "the manifest's `semantic_hash` is the truth". A one-line guard here would close it:

```python
if manifest.get("semantic_hash") and semantic_hash(config) != manifest["semantic_hash"]:
    raise ValueError(...)
```

Two properties make this cheap rather than a false-refusal generator: (1) `output.*` is NOT in the semantic core, so a retrofit config that differs from the original only by adding `output.pyramid` hashes **identically**; (2) the `_frozen_matches` precedent (`hive.py:561-576`) already establishes the "compare only when both sides declare it" exemption for pre-#299 manifests, which is exactly what a pre-#344 retrofit target may be.

If you'd rather not hard-refuse (a retrofit operator may only hold a partial config), a loud `logger.warning` naming both fingerprints (`semantic_fingerprint`) plus the mismatch recorded in `summary["validated"]` would still beat installing an unverified fold law silently.

src/zagg/sweep_overview.py
```diff
@@ -297,6 +297,163 @@ def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
...
+        "orders": list(block["overview"].get("orders") or []),
+        "fields": {n: m.get("class") for n, m in (block["overview"].get("fields") or {}).items()},
+        "validated": validated,
+        "previous": "absent" if prior is None else "identical" if prior == block else "replaced",
+        "updated": prior != block,
+    }
+    if prior == block:
+        logger.info("declare_pyramid: the manifest already carries this declaration; no write")
+        return summary
+    manifest["pyramid"] = block
```

[correctness] The RMW writes back a manifest read **before** validation — a long window in which a concurrent sweep's update is clobbered.

The read happens at the top:

```python
manifest = read_manifest(store_root, **store_kwargs)
```

then `_validate_block_against_store` runs `discover_leaves` (one root LIST + a GET-and-parse of **every** `stats_*.parquet` in the store), up to 8 leaf `read_commit` opens, and a `zarr.open_group` — seconds to minutes on the production-scale stores this tool targets. Only then is that stale dict mutated and PUT whole:

```python
manifest["pyramid"] = block
obstore.put(..., json.dumps(manifest, indent=1).encode())
```

The neighbouring writer of the same key deliberately does the opposite — `_update_manifest_pyramid` re-reads immediately before its PUT:

```python
fresh = read_manifest(store_root, **store_kwargs)
if fresh is None:
    return False
block = fresh.setdefault("pyramid", {})...
```

with a comment that the RMW "can never brick appends". Here, a sweep that lands `materialized` (or anything else — the PUT rewrites the whole manifest, not just `pyramid`) during the validation window is silently reverted. Suggest re-reading at line 369 and re-deriving `prior`/`materialized` from the fresh copy (the `block` derivation only needs `shard_order`, which is frozen), and raising if the fresh manifest's frozen keys no longer match the one validation ran against.

src/zagg/sweep.py
```diff
@@ -1078,6 +1086,15 @@ def main(argv=None) -> int:
             credentials = normalize_output_credentials(json.load(f))
         store_kwargs["credentials"] = credentials
         store_kwargs["endpoint_url"] = credentials.get("endpointUrl")
+    if args.declare_pyramid:
```
espg: 🤖 *from Claude (review)*

[correctness] The declaration-only branch is guarded on truthiness, so an **empty** `--declare-pyramid` value silently falls through and runs a full sweep pass instead of failing:

```python
if args.declare_pyramid:
```

The failure mode is a shell script with an unset variable — `python -m zagg.sweep s3://bucket/store --declare-pyramid=$CFG` with `CFG` unset expands to `--declare-pyramid=""`, argparse accepts it, and the CLI proceeds to `discover_leaves` + `run_sweep`. That is the exact operation the flag advertises it will not perform ("no sweep pass runs in the same invocation"), and it writes.

Reproduced against a store built with the file's own helpers (`_write_manifest` + two `_put_leaf` + `_run_record`):

```
main([root, "--declare-pyramid", ""])  -> 0
rollups after: ["stats.rollup.json", "stats.rollup.json", "stats.rollup.json", "stats.rollup.json"]
```

`if args.declare_pyramid is not None:` makes the empty string reach `load_config`, which fails loudly on the empty path instead of quietly doing the expensive/write-y thing.

src/zagg/sweep_overview.py
```diff
@@ -297,6 +297,163 @@ def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
...
+        return f"field {name!r} is declared but absent from the leaf"
+    if meta["class"] == "approximate":
+        element = ((arr.attrs.get(RAGGED_ELEMENT_ATTR) or {}).get("element")) or {}
+        if not element:
+            return (
+                f"field {name!r}: declared approximate (ragged) but the stored "
+                f"array carries no ragged element declaration"
+            )
+        declared_dt = np.dtype(meta.get("dtype") or "float32")
+        if np.dtype(element.get("dtype")) != declared_dt:
```

[bug] `np.dtype(None)` is `float64`, so a ragged element block missing its `dtype` key silently **passes** a `float64` declaration.

```python
declared_dt = np.dtype(meta.get("dtype") or "float32")
if np.dtype(element.get("dtype")) != declared_dt:
```

Verified against the env this branch runs in:

```
$ python -c "import numpy as np; print(np.dtype(None))"
float64
```

The `if not element:` guard above only proves the `element` mapping is non-empty — an element block of `{"shape": [-1, 2]}` (or one whose key got renamed by a future D18/#210 migration) reaches this line, `np.dtype(None)` evaluates to `float64`, and a declaration of `dtype: float64` validates against a store that declares nothing. Cheap fix: fail on the missing key explicitly, e.g.

```python
stored_dt = element.get("dtype")
if not stored_dt:
    return f"field {name!r}: ragged element declaration carries no dtype"
```

(The same class of silent-default lives one line down in `element.get("shape") or [-1]`, though there the fallback `[]` at least *fails* the `[2]` comparison rather than passing it. Worth noting the ragged `spec` version in the attrs block is not checked either — `read_coverage` in `hive.py` is strict on its spec by review ruling from PR #208; this validator accepts any `RAGGED_SPEC`.)

src/zagg/sweep_overview.py
```diff
@@ -297,6 +297,163 @@ def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
...
+
+    from zagg.hive import read_commit, shard_leaf_path
+    from zagg.store import open_store
+    from zagg.sweep import discover_leaves
+
+    fields = block["overview"].get("fields") or {}
+    if not fields:
+        return "nothing to check (declared off)"
+    leaf = None
+    for key, window in discover_leaves(store_root, store_kwargs=store_kwargs)[:8]:
```

[correctness] The 8-ref probe conflates two very different stores, and the cap does not bound the expensive part. (This answers the PR's own "Validation probe bound" question — I'd say yes, but the fix isn't "scan deeper".)

```python
for key, window in discover_leaves(store_root, store_kwargs=store_kwargs)[:8]:
    path = shard_leaf_path(store_root, key, window=window)
    if read_commit(open_store(path, **store_kwargs)) is not None:
```

(1) **Silent downgrade.** `leaf is None` currently means either "this store has no run records at all" (a genuinely pre-commit store — declarable, per the contract) or "this store has thousands of refs and the first 8 in `sorted()` order happened to be torn/rolled-back/deleted". The second case is a store where validation was *skipped* while the operator believes the tool validated — and it is exactly the retrofit target class (the cycle22 / 88S o9 stores named in issue #358, where early shards are the ones most likely to be debris). The two should not return the same string. Suggest branching on whether `refs` is empty: empty → today's declarable skip; non-empty but no committed leaf in the probe → either widen the probe (the refs are already in memory; the marginal cost is one `read_commit` per extra ref, not a re-discovery) or raise and let the operator pass an explicit opt-out.

(2) **The cap bounds the cheap half.** `[:8]` truncates *after* `discover_leaves` has already done a root `list_with_delimiter` **plus a GET + fastparquet parse of every `stats_*.parquet` in the store** (`sweep.py:959-978`), and, on a windowed store with pre-`window`-column records, one extra delimiter LIST per fallback shard. On a store with hundreds of run records that is the dominant cost of `declare_pyramid`, and it is paid to find one leaf. If the goal is a bounded probe, consider stopping after the first record that yields a ref rather than reading them all.

(3) Nit within the same loop: `open_store(path, **store_kwargs)` is opened writable. `sweep.py:183` does the same, so this matches the surrounding code — but on S3 `read_only=True` also selects the shorter `_S3_READONLY_RETRY_CONFIG` (issue #186), which is what a read probe wants.

src/zagg/sweep_overview.py
```diff
@@ -297,6 +297,163 @@ def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
...
+    source — the same :func:`build_pyramid_block` path template time uses) at
+    the manifest's own ``shard_order``, validated against store truth, and
+    RMW'd into the manifest. Unlike the sweep's fail-open ``materialized``
+    update (:func:`_update_manifest_pyramid`), this is the user's explicit
+    operation: every failure raises — a missing manifest, a declaration the
+    store contradicts — and nothing is half-written (the one PUT is the
+    whole write).
+
+    Validation source: ONE committed leaf, found through the store's run
+    records (:func:`zagg.sweep.discover_leaves` — the sweep's own LIST-free
```

[nit] The docstring's validation-source justification is inaccurate on both halves.

(1) "`zagg.sweep.discover_leaves` — the sweep's own **LIST-free** discovery". It is not LIST-free. Its own docstring:

> One shallow delimiter LIST of the product root finds the `stats_*.parquet` run records ... discovery is from run records, **never a tree enumeration**.

and the code does `listing = obstore.list_with_delimiter(store)` (`sweep.py:959`), plus "one delimiter LIST of that shard's node" per fallback shard for pre-`window`-column records on a windowed store. The accurate claim is "tree-enumeration-free / no recursive LIST", which is what the rest of the module says (see this file's header: "never a recursive LIST").

(2) "Leaves are the layer the overview fold actually reads, so they are the store truth that can contradict a fold recipe; the D19 `aggregation.yaml` core is NOT used". The premise is only true for the *typing* half of the recipe. A leaf records dtype and the D18 ragged element block, and nothing else about the aggregation — no reducer, no `nan_policy`, no `fill_value` check is performed here either. So a leaf cannot contradict `method`/`nan_policy`, which is half of what the declaration asserts (see my separate note on `semantic_hash`). Worth rewording so the docstring states what the leaf probe *can* falsify (presence, dtype, ragged element shape) rather than implying it covers the fold recipe.

Minor third: when the config declares the pyramid ON but the aggregation block has no variables, `_validate_block_against_store` returns `"nothing to check (declared off)"` — a message that will read as a lie in the summary.

tests/test_sweep.py
```diff
@@ -905,6 +905,54 @@ def test_end_to_end_windowed_discovery_folds_both_windows(self, tmp_path):
...
+        assert main([str(tmp_path), "--declare-pyramid", str(config_path)]) == 0
+        summary = json.loads(capsys.readouterr().out)
+        assert summary["updated"] is True and summary["orders"] == [0]
+        # No committed leaf on this store: validation records the loud skip.
+        assert summary["validated"].startswith("manifest only")
+        block = read_manifest(str(tmp_path))["pyramid"]["overview"]
+        assert block["orders"] == [0]  # manifest shard_order 2, spacing 2
+        assert block["fields"]["count"]["class"] == "exact"
+        # Declaration-only semantics: no sweep pass ran in the invocation.
+        assert not list(tmp_path.rglob("*.rollup.json"))
```
espg, line 980: 🤖 *from Claude (review)*

[test-gap] This negative assertion cannot fail — it holds whether or not a sweep ran. `_write_manifest` leaves a store with **no leaves and no run record**, so the fallthrough path also writes zero rollups (`test_empty_store_is_a_noop` above pins exactly that: `"nothing to sweep"`). The declaration-only claim is therefore untested.

Give the store something to sweep and the assertion becomes load-bearing. Adding the file's existing three-liner from `test_end_to_end_discovers_and_folds`:

```python
_put_leaf(tmp_path, "-311")
_put_leaf(tmp_path, "-312")
_run_record(tmp_path, [_row("-311"), _row("-312")])
```

I ran that combination locally against this branch:

- `main([root, "--declare-pyramid", cfg])` -> rollups after: `[]`, summary unchanged (`validated: "manifest only ..."`, `orders: [0]`)
- the sweep path on the same store -> `["stats.rollup.json"] x 4`

So the leaves do not perturb any existing assertion (`_put_leaf` writes a sidecar only — no commit stamp — so validation still takes the `manifest only` branch), and the `rglob` now actually discriminates.

tests/test_sweep_overview.py
```diff
@@ -972,3 +973,172 @@ def test_role_never_inferred_from_position(self, tmp_path):
...
+            declare_pyramid(str(tmp_path), _leaf_cfg())
+
+    def test_declared_off_retrofit(self, tmp_path):
+        self._pre_declaration_store(tmp_path)
+        cfg = _leaf_cfg()
+        cfg.output["pyramid"] = False
+        summary = declare_pyramid(str(tmp_path), cfg)
+        assert summary["updated"] is True and summary["orders"] == []
+        assert summary["fields"] == {}
+        assert read_manifest(str(tmp_path))["pyramid"] == {
```
espg, line 1205: 🤖 *from Claude (review)*

[test-gap] Three uncovered paths in `TestDeclarePyramid`, one of which pins an unstated design decision:

(1) **declared-off + prior `materialized`.** This assertion fixes the declared-off block shape only for the never-swept store. On a store that HAS been swept, the preservation branch runs unconditionally:

```python
materialized = ((prior or {}).get("overview") or {}).get("materialized")
if materialized is not None:
    block["overview"]["materialized"] = materialized
```

so `output.pyramid: false` writes `{"orders": [], "materialized": {"orders": [...], "generated_at": ...}}` — a declared-OFF block that still inventories on-disk overviews. That is arguably right under D24 option A (the debris is real and the sweep no-ops correctly, since `sweep_overviews` gates on `decl.get("orders")`), but it is the kind of thing a reader will be surprised by and nothing pins it. Suggest a test that runs `test_redeclaration_preserves_materialized`'s sweep first, then declares off, and asserts the intended shape either way.

(2) **Drift refusal over an EXISTING declaration.** `test_field_dtype_drift_refuses_and_writes_nothing` proves the never-half-written property only from the `pyramid`-absent state (`assert "pyramid" not in read_manifest(...)`). The more dangerous case — a store that already carries a good declaration, and a bad config that must leave it byte-identical — is untested.

(3) **Windowed stores.** `shard_leaf_path(store_root, key, window=window)` in the probe is never exercised with a non-`None` window; `_make_leaf`/`_write_manifest` already support `windowed=True`, so this is a couple of lines (and the windowed path is where `discover_leaves` has its trickiest fallback).

src/zagg/sweep_overview.py
```diff
@@ -297,6 +297,163 @@ def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
...
+    from zagg.store import open_object_store
+
+    store_kwargs = dict(store_kwargs or {})
+    manifest = read_manifest(store_root, **store_kwargs)
+    if manifest is None:
+        raise ValueError(
+            f"no {MANIFEST_NAME} at {store_root} — not a hive store root; "
+            f"declare_pyramid retrofits existing hive stores only"
+        )
+    shard_order = int(manifest["shard_order"])
```
espg, line 363: 🤖 *from Claude (review)*

[nit] Error quality: the pointed refusal covers only the *absent* manifest, not the malformed one.

```python
shard_order = int(manifest["shard_order"])
```

The contract in issue #358 is "raise loudly (NOT fail-open)" with a message that tells the operator what is wrong. A `morton_hive.json` that parses but isn't a hive manifest (hand-edited, truncated-and-rewritten, or another product's JSON dropped at the root) gets `KeyError: 'shard_order'` — and `int(manifest["cell_order"])` in `_validate_block_against_store` the same, later, after the whole discovery cost has been paid. `read_manifest` returning a non-dict (a JSON list) gets `TypeError: list indices must be integers`.

Same shape one block down:

```python
materialized = ((prior or {}).get("overview") or {}).get("materialized")
```

`(prior or {})` rescues `None`/`{}`/`[]` (falsy) but a `"pyramid": "off"` string prior raises `AttributeError: 'str' object has no attribute 'get'`, and a non-dict prior otherwise flows through to `prior != block` → silently replaced. Suggest a small guard next to the existing `manifest is None` raise: require `isinstance(manifest, dict)` and both `shard_order`/`cell_order` present, folded into the same "not a hive store root" message; and treat a non-dict `pyramid` prior as `absent` explicitly (`prior if isinstance(prior, dict) else None`).

src/zagg/sweep_overview.py
```diff
@@ -297,6 +297,163 @@ def _update_manifest_pyramid(store_root, orders, store_kwargs) -> bool:
...
+    if materialized is not None:
+        block["overview"]["materialized"] = materialized
+    summary = {
+        "orders": list(block["overview"].get("orders") or []),
+        "fields": {n: m.get("class") for n, m in (block["overview"].get("fields") or {}).items()},
+        "validated": validated,
+        "previous": "absent" if prior is None else "identical" if prior == block else "replaced",
+        "updated": prior != block,
+    }
+    if prior == block:
```
espg, line 408: 🤖 *from Claude (review)*

[nit] The idempotency compare is in-memory-vs-JSON-round-tripped, which leaves one PUT-loop hole open.

`prior` came back through `json.loads` (every value is a JSON primitive/list/dict); `block` is freshly built in Python. Most of `build_pyramid_block`'s output is normalized against that asymmetry on purpose — `int(...)`, `bool(...)`, `[int(x) for x in inner]`, `_json_fill` — but `summarize` is not:

```python
if knob.get("summarize"):
    overview["summarize"] = {str(k): dict(v) for k, v in knob["summarize"].items()}
```

`dict(v)` is a shallow copy, so any tuple inside a summarize declaration survives into `block` and comes back from the manifest as a list — making `prior != block` on **every** call, i.e. a PUT on every invocation of a tool whose headline property is "identical declaration → no PUT". YAML configs can't produce tuples, so this is latent rather than live, but the whole class dies with one line here:

```python
block = json.loads(json.dumps(block))   # compare (and write) canonical JSON
```

which also guarantees the block is serializable *before* the store-truth probe runs, rather than raising `TypeError` at the PUT after all the validation work (a programmatic config carrying e.g. a `np.float32` fill_value does exactly that today — `_json_fill` passes non-NaN values through untouched).

tests/test_sweep.py
```diff
@@ -905,6 +905,54 @@ def test_end_to_end_windowed_discovery_folds_both_windows(self, tmp_path):
...
+        from zagg.hive import read_manifest
+        from zagg.sweep import main
+
+        _write_manifest(tmp_path)  # legacy pyramid block, no run records
+        config_path = self._config_yaml(tmp_path)
+        assert main([str(tmp_path), "--declare-pyramid", str(config_path)]) == 0
+        summary = json.loads(capsys.readouterr().out)
+        assert summary["updated"] is True and summary["orders"] == [0]
+        # No committed leaf on this store: validation records the loud skip.
+        assert summary["validated"].startswith("manifest only")
```
espg, line 947: 🤖 *from Claude (review)*

[test-gap] Both CLI tests take the happy path, and this assertion pins the *degenerate* validation branch — `manifest only (no committed leaf ...)`. Nothing at the CLI layer covers what the flag's help and the docs promise on failure ("drift refuses loudly"): the wrapper has no error handling at all, so a `ValueError` from `declare_pyramid` (or a `FileNotFoundError` / `validate_config` `ValueError` from `load_config`) escapes `main()` as a raw traceback and the process exits 1 via the interpreter, never through `return 0`/`sys.exit(main())`.

Verified on this branch:

```
main([root, "--declare-pyramid", "<missing>.yaml"])  -> FileNotFoundError [Errno 2] ...
main([root_without_manifest, "--declare-pyramid", cfg]) -> ValueError: no morton_hive.json at ... — not a hive store root
```

That escape-and-exit-1 behavior matches `python -m zagg` (which also lets exceptions surface), so it is probably the intended shape — but then pin it, one line each, so a future `try/except` that swallows to `return 0` is caught:

```python
with pytest.raises(ValueError, match="not a hive store root"):
    main([str(tmp_path), "--declare-pyramid", str(config_path)])  # no manifest written
```

docs/hive_layout.md
```diff
@@ -218,6 +218,37 @@ while the digit tree has any `{sign+base}` children (one delimiter-LIST):
...
+
+```
+python -m zagg.sweep s3://bucket/store --declare-pyramid config.yaml
+```
+
+(declaration-only: no sweep pass runs in the same invocation). The tool
+derives the block through the same code path template time uses, then
+**validates it against store truth before writing**: every declared field
+must exist in a committed leaf with the declared dtype (dense fields) or
+ragged element declaration (t-digests), so a wrong config cannot install a
```

[docs] "so a wrong config cannot install a fold recipe the store contradicts" claims more than the code delivers. `declare_pyramid` checks (1) that the derived orders sit in `0 <= k < shard_order`, and (2) per-declared-field presence/dtype against **one** committed leaf. It never cross-checks the config's identity against the manifest — not `dataset`, not the grid's `parent_order`/`child_order` against the manifest's `shard_order`/`cell_order`.

Demonstrated with this PR's own test fixture: the shipped `atl06` config (healpix `parent_order: 6`, `child_order: 12`, ATL06) retrofits cleanly onto a `{"short_name": "TEST"}` manifest at `shard_order: 2` / `cell_order: 4`:

```
{"orders": [0], "fields": {"count": "exact", ...}, "validated": "manifest only (...)", "updated": true}
```

Two ways to make the sentence true-as-written: scope it ("every *declared field* is checked against one committed leaf; the config's dataset and grid orders are not cross-checked"), or say the quiet part the paragraph already half-says — the guarantee only bites once the store has a committed leaf, and only over field names/dtypes.

docs/hive_layout.md
```diff
@@ -218,6 +218,37 @@ while the digit tree has any `{sign+base}` children (one delimiter-LIST):
 clear the store root, or pick a new one, before writing with a different
 configuration.
 
+### Retrofitting the pyramid declaration
```
espg, line 230: 🤖 *from Claude (review)*

[docs] The new subsection lands directly under a manifest example and a sentence it now contradicts. Thirty lines up, the doc still shows the pre-#201 block shape:

```json
"pyramid": {"orders": [], "aggregation": {}},
```

followed by "`pyramid` is declared-only in round one: overview zarrs are generated by a later post-process sweep (D11), never at fan-out time."

Neither survives this PR intact:

(1) `build_pyramid_block` writes `{"spec": "zagg-pyramid/1", "overview": {...}}` — there is no top-level `orders`, and no `aggregation` key at all. Actual output of the retrofit on a `shard_order: 2` store (declared off, so the shortest form): `{"spec": "zagg-pyramid/1", "overview": {"orders": []}}`; the on-by-default form carries `spacing`/`orders`/`all_time`/`fields`.

(2) The example's `"orders": []` now specifically *means* declared-off (`get_pyramid` -> `None` -> empty schedule), so a reader lands on an example that reads "the shipped default declares no pyramid" and then a subsection describing a default every-2-orders schedule.

(3) "declared-only in round one" sits oddly next to "declaration is not birth-only" — the materialization half is still accurate, the birth-only half is what this PR retires.

A one-line fix to the example block plus a pointer from that paragraph to this subsection keeps the doc internally consistent. Flagging rather than folding silently, since the example predates this PR.

docs/hive_layout.md
```diff
@@ -218,6 +218,37 @@ while the digit tree has any `{sign+base}` children (one delimiter-LIST):
...
+ragged element declaration (t-digests), so a wrong config cannot install a
+fold recipe the store contradicts — drift refuses loudly, naming the field
+and the mismatch, and nothing is half-written. A store with no committed
+leaf yet is still declarable; the field checks are then skipped, loudly.
+The rewrite is idempotent (an identical declaration is not re-PUT) and
+preserves any `materialized` actuals the sweep has recorded.
+`output.pyramid: false` installs the declared-off block — recording absence
+is a valid retrofit (readers then never probe for overviews), and overviews
+at now-undeclared orders are left in place as regenerable-cache debris
+(D24). After a retrofit, the next overview sweep materializes the declared
```

[nit][docs] Two small over-claims in the closing sentences.

(1) "(readers then never probe for overviews)" is stated as present-tense fact, but nothing in this tree reads the declaration on the read path — `grep -rn pyramid src/zagg/` hits only `config.py`, `hive.py`, `sweep.py`, `sweep_overview.py`, a `coverage.py` comment, and an unrelated `viz/crs.py` word. The zero-open filtering is the D24 option-A *posture* the block enables for readers (moczarr and friends), not behavior this repo implements. "so a reader can skip probing for overviews entirely" reads honestly and costs nothing.

(2) "the next overview sweep materializes the declared orders (issue #358)" — #358 is the retrofit; the sweep that materializes is #201 / PR #344. Citing #358 here points a future reader at the wrong thread for the fold itself. (The claim is otherwise correct: `overview` is in `DEFAULT_FAMILIES`, so a plain `python -m zagg.sweep <root>` after the retrofit does pick it up — worth one clause, since the code block above only shows the retrofit invocation and not the follow-up.)

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Adopted — fixed in 37f4cc9.

New `_semantic_guard(manifest, config)` runs before the (slow) store probe: it compares `zagg.semantics.semantic_hash(config)` against the manifest's frozen `semantic_hash` (`hive.py:318`), and refuses pointedly on mismatch, naming both `semantic_fingerprint`s. Hard refusal rather than the warning fallback — installing an unverified fold law is the exact failure issue #358 exists to make impossible, and the operator always has the escape hatch of the original config.

Both properties you named are now pinned by tests:

- `test_pyramid_only_edit_hashes_identically` — the original config plus `output.pyramid` passes (verified directly: `output.*` is not in `semantic_core`, only `output.grid`, so the retrofit edit hashes identically).
- `test_wrong_reducer_refuses_via_semantic_hash` — `h_min` declared `max` over a store of minima now refuses, and the manifest is left with no `pyramid` key.
- `test_hashless_manifest_skips_the_semantic_check` — a pre-#299 manifest skips (the `_frozen_matches` both-present exemption), loudly: a `logger.warning` saying the semantics "could NOT be verified", plus `semantic_hash absent (pre-#299 store — fold methods unverified)` appended to `summary["'validated"']`. The hash-present case records `semantic_hash <fingerprint>` there instead.

Not folded here: the ragged `spec`-version check you raise under the separate `np.dtype(None)` note — replied on that thread.

src/zagg/sweep_overview.py

Adopted — fixed in 1617651.

The RMW now re-reads immediately before the PUT, mirroring `_update_manifest_pyramid`:

```python
fresh = read_manifest(store_root, **store_kwargs)
if fresh is None or not _frozen_matches(fresh, manifest):
    raise ValueError(... "changed under declare_pyramid's validation window" ...)
prior = fresh.get("pyramid")
```

`prior`/`materialized` are derived from `fresh`, and `fresh` (not the pre-validation copy) is what gets `block` installed and PUT — so a concurrent sweep's `materialized` landing in the validation window survives, and so does any other key it touched. The early read is kept for `shard_order` and the block derivation, as you suggested.

Your raise-on-frozen-drift is in too: `_frozen_matches(fresh, manifest)` (which also handles the vanished-manifest case, since it returns `False` for `None`), so the block is never installed on a manifest validation never saw.

Two tests: `test_concurrent_materialized_survives_the_validation_window` (monkeypatches `_validate_block_against_store` to land a real `_update_manifest_pyramid` mid-call, then asserts both the new `orders` and the concurrent `materialized`) and `test_manifest_changed_under_validation_refuses`.

src/zagg/sweep_overview.py

Adopted — fixed in e774877. Confirmed `np.dtype(None)` is `float64` in this venv before fixing.

```python
stored_dt = element.get("dtype")
if stored_dt is None:
    return f"field {name!r}: the stored ragged element declaration carries no dtype"
if np.dtype(stored_dt) != declared_dt:
    ...
```

`test_ragged_element_without_dtype_refuses` writes an element block of `{"shape": [-1, 2]}` and declares `float64` — the exact silent pass. It fails on the pre-fix code (dtype matched, `inner_shape` matched, no drift returned) and refuses now.

Two things left standing, deliberately:

1. `element.get("shape") or [-1]` — as you note, the fallback `[]` *fails* the `[2]` comparison, so a missing `shape` already refuses rather than passing. Fail-closed, so no change; the missing-key message is less pointed than it could be, but it is a refusal either way.
2. The ragged `spec`-version check (the PR #208 `read_coverage` precedent). Real, but it is a new validation axis rather than a fix to this one, and it wants the same "what is the compatible spec set" ruling `read_coverage` got — out of scope for this fold. Leaving it standing for espg to rule on.

src/zagg/sweep_overview.py

Adopted on all three — fixed in f0a90e6.

(1) The two cases no longer report the same thing. Neither refuses: a store with no committed leaf MUST stay declarable (mid-backfill retrofit targets are the point of the tool), so both are loud skips that name which one happened:

```python
detail = (
    "no run records at the product root"
    if not refs
    else f"none of the {len(refs)} run-record refs is a committed leaf"
)
```

That string goes into both the `logger.warning` and `summary["'validated"']`. `test_uncommitted_refs_are_not_reported_as_no_records` pins the debris case against `test_no_committed_leaf_still_declarable`'s never-ran case.

(2) The `[:8]` cap is gone. You are right that it bounded the wrong half — `discover_leaves` has already paid the root LIST plus a GET + parse of every run record by the time the slice applies, so truncating only risked giving up while committed leaves remained. Each extra ref is one `read_commit` GET and the loop breaks at the first hit. I did *not* take the "stop after the first record that yields a ref" variant: that is a change inside `discover_leaves`, which the sweep engine shares, so it belongs in its own issue rather than being smuggled in through the retrofit tool.

(3) Taken: the probe now opens `read_only=True`, which on S3 also selects the shorter `_S3_READONLY_RETRY_CONFIG` (issue #186) — right for a read probe.

src/zagg/sweep_overview.py

Adopted on all three — fixed in 20b52ea.

(1) "LIST-free" is gone; the docstring now says "tree-enumeration-free ... one shallow root LIST of the run records, never a recursive LIST", matching the module header and `discover_leaves`' own wording.

(2) Rewritten as two explicitly separate validation sources, since the leaf half really does cover typing only:

- **Typing** — one committed leaf, which can falsify "the declared **presence, dtype, and D18 ragged element shape** of each field. They cannot falsify more than that: no leaf records which reducer produced it."
- **Semantics** — the manifest's frozen `semantic_hash` via the new `_semantic_guard`, "which is what covers the `method`/`nan_policy` half the leaf is silent on" (your separate `semantic_hash` finding, folded in 37f4cc9).

(3) The declared-ON-but-no-variables case no longer claims "declared off". The declared-off block carries no `fields` key at all, an ON declaration over an empty aggregation carries an empty one, so the two are distinguishable and now read `nothing to check (declared off)` vs `nothing to check (no aggregation variables to declare)`.

tests/test_sweep_overview.py
espg: 🤖 *from Claude*

All three added — fixed in 0b61963.

(1) `test_declared_off_preserves_prior_materialized` — sweeps first, then declares off, and asserts the block you predicted, explicitly as the intended design:

```python
assert block == {"orders": [], "materialized": materialized}
```

with the D24 option A rationale in a comment (the overviews on disk are real regenerable-cache debris, so the inventory stays honest), plus a follow-up `run_sweep` asserting `declared is False` — the no-op that makes the shape safe.

(2) `test_drift_leaves_an_existing_declaration_untouched` — installs a good declaration, snapshots the whole manifest as sorted JSON, drives a `float64` drift refusal, and asserts the manifest is unchanged. That is the byte-identical property from the dangerous starting state, not just from `pyramid`-absent.

(3) `test_windowed_store_probe_resolves_the_window` — windowed manifest, a leaf under `2411.zarr`, a run record carrying `window="2411"` (`_run_record` grew a `window=` kwarg), and asserts the resolved path in `summary["'validated"']` contains `2411.zarr`. A dropped window would degrade to the no-committed-leaf skip, which the assertion catches.

I exercised the run-record `window`-column path rather than the pre-column fallback LIST: the fallback lives in `discover_leaves` and is the sweep engine's shared code, already covered there, whereas `window=window` in `shard_leaf_path` is the line this PR added.

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Adopted — fixed in 6745ec6.

A guard next to the existing `manifest is None` raise, covering the non-dict case and BOTH order keys:

```python
if not isinstance(manifest, dict) or any(
    manifest.get(k) is None for k in ("shard_order", "cell_order")
):
    raise ValueError(
        f"the {MANIFEST_NAME} at {store_root} declares no shard_order/cell_order — "
        f"not a hive store manifest; declare_pyramid retrofits existing hive stores only"
    )
```

`cell_order` is checked up front deliberately, per your point that it otherwise `KeyError`s inside `_validate_block_against_store` after the whole discovery cost. The `isinstance` check short-circuits before any `.get`, so a JSON list gets the pointed message instead of `TypeError`.

The non-dict `pyramid` prior no longer `AttributeError`s:

```python
prior_overview = prior.get("overview") if isinstance(prior, dict) else None
materialized = prior_overview.get("materialized") if isinstance(prior_overview, dict) else None
```

One deviation from your suggestion: a non-dict prior reports `previous: "'replaced"'"` rather than `"'absent"'`. Something *was* there and the call overwrote it — telling the operator "absent" would understate what happened. It carries no actuals to preserve either way.

Tests: `test_malformed_manifest_raises_pointedly` (parametrized over dropping `shard_order` / `cell_order`) and `test_non_dict_prior_pyramid_is_replaced` (a hand-edited `"'pyramid"': "'off"'`).

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Adopted — fixed in 1589511. Your one-liner, immediately after the derivation:

```python
block = json.loads(json.dumps(block))
```

placed before the orders check so the second property you name holds too: an unserializable block raises here rather than at the PUT, after the whole store-truth probe has been paid for.

I did add the regression test, since it was natural: `test_non_json_primitive_declaration_stays_idempotent` puts a tuple in a `summarize` declaration (`{"'quantiles"': (0.5,)}` — the exact `dict(v)` shallow-copy hole), asserts it lands in the manifest as `[0.5]`, then re-declares with the same config and asserts `previous == "'identical"'` with zero `obstore.put` calls. On the pre-fix code the second call re-PUT every time.

src/zagg/sweep.py
espg: 🤖 *from Claude*

Fixed in `f64e4b9` — the guard is now `if args.declare_pyramid is not None:`, with a comment naming the failure mode (an unset shell variable expanding to `--declare-pyramid=""`) so it does not regress to truthiness. An empty value reaches `load_config` and fails there instead of silently running the writing sweep pass.

Pinned by a new test in `34b6c70` (`test_empty_declare_pyramid_value_never_sweeps`): a store with a real run record and leaves, `main([root, "--declare-pyramid", ""])` raises, and `rglob("*.rollup.json")` is empty afterwards — so the fallthrough is caught, not just the exception type.

tests/test_sweep.py
espg: 🤖 *from Claude*

Fixed in `34b6c70` — took the suggested three-liner verbatim (`_put_leaf` x2 + `_run_record`), so the store now has real sweepable work and the `rglob` assertion discriminates. Confirmed your reading: the sidecars carry no commit stamp, so validation still lands on the no-committed-leaf branch and `summary["validated"]` is unchanged.

Added a positive control alongside it (`test_declare_pyramid_flag_does_not_sweep_a_sweepable_store`): the same fixture swept WITHOUT the flag writes four `stats.rollup.json` and the assertion inverts — so the absence in the flag test is demonstrably the flag's doing rather than an inert store.

tests/test_sweep.py
espg: 🤖 *from Claude*

Fixed in `6723c0d`, adopting your reading: the escape-and-exit-1 shape IS intended (it matches `python -m zagg`), so the fold pins it rather than adding error handling. `test_declare_pyramid_failures_surface_not_swallowed` covers both escapes you verified:

- `ValueError` / `not a hive store root` on a directory with no manifest
- `FileNotFoundError` on a missing config path

Both use `pytest.raises`, so a future `try/except` that swallows a refusal into `return 0` fails the test — which is the property worth protecting, since the flag's whole promise is that a wrong config does not get installed.

docs/hive_layout.md
espg: 🤖 *from Claude*

Fixed in `bef7250` — you were right to flag rather than fold silently, but since the subsection makes the contradiction load-bearing, the example is now corrected in this PR:

(1) the block shows the real shape (`{"spec": "zagg-pyramid/1", "overview": {"spacing": 2, "orders": [7, 5, 3, 1], "all_time": false, "fields": {…}}}`) — no top-level `orders`, no `aggregation` key. It also picks up `semantic_hash` and `path_grouping`, which `build_manifest` writes and the example was missing.
(2) the empty-`orders` reading is disambiguated in prose: an `overview` with an empty `orders` list is the declared-*off* form.
(3) "declared-only in round one" is now "It is **declaration** only — overview zarrs are generated by a later post-process sweep (D11), never at fan-out time", i.e. the materialization half kept, the birth-only half retired, plus a pointer to the new subsection.

docs/hive_layout.md

Fixed in `b1a8fad` — took your second option (say the quiet part), and the claim has also genuinely strengthened since you reviewed: the phase-1 fold added a `semantic_hash` guard (`37f4cc9`), so the sentence you flagged is no longer the whole story. The subsection now splits validation into two named layers:

- **Semantics** — the config's D19 `semantic_hash` vs the manifest's frozen one, explicitly called out as the layer that covers *reducers* (no leaf records which function produced a field); `output.*` is outside the semantic core so adding `output.pyramid` hashes identically; pre-#299 manifests skip it, loudly.
- **Typing** — per-declared-field presence/dtype/ragged element against one committed leaf.

And the non-guarantees are now stated outright, in your words: `dataset` identity and the grid `parent_order`/`child_order` are NOT cross-checked; the declared orders are only checked for being ancestor orders of the store's own `shard_order`.

Note your specific repro (shipped `atl06` onto a `{"short_name": "TEST"}` / `shard_order: 2` manifest) would now be refused by the semantic-hash layer on any post-#299 store — but it stands exactly as written for a pre-#299 manifest, which is the common retrofit target, so the doc keeps the caveat.

docs/hive_layout.md

Both fixed in `52f3cd6`.

(1) Took your wording plus the reason: "so a reader can skip probing for overviews entirely (the D24 option-A posture the block enables; no reader in this repo consumes the declaration yet)". Confirmed your grep — nothing on the read path reads the block today.

(2) Re-pointed the citation: "the fold itself is issue #201 / PR #344; the retrofit is issue #358", and added the clause you suggested — the overview family is in `DEFAULT_FAMILIES`, so a plain `python -m zagg.sweep <root>` after the retrofit picks it up, which the code block above (retrofit invocation only) did not show.

src/zagg/sweep_overview.py

Follow-up: the ragged `spec`-version sub-point left standing here has been **ruled and folded** — espg approved the refusal (2026-07-31). Implemented in `fb29ac2`.

`_field_drift` now gates the D18 attrs' `spec` against `zagg.grids.base.RAGGED_SPEC` before the dtype/`inner_shape` checks, mirroring `readers.tdigest_tensor._open_ragged` rather than inventing a second policy — same constant, same fail-loud-never-half-parse posture. The refusal names the field and the offending revision.

Rationale recorded in the docstring: a store whose ragged layout this zagg cannot read is the purest instance of the tool's contract ("a wrong config cannot install a fold recipe the store contradicts") — without the gate the retrofit installs a declaration promising overviews the sweep then fails to read at FOLD time, later and with a worse error. `zagg-ragged/2` is a live migration path (issue #210 moves the element declaration into the zarr data type), so it is not hypothetical.

Test: `test_unreadable_ragged_spec_revision_refuses` asserts the fixture is a current-spec store, rewrites the attr to `zagg-ragged/2`, and pins both the refusal (matching field + revision) and that the manifest keeps no `pyramid` key — nothing half-written.

PR #363, opened 2026-07-31
Benchmark object model: telemetry bucket + per-commit store scoping (issue #362)

### LINKED ISSUE #362: Benchmark object-count model: unbounded root-telemetry bucket + per-commit store scoping
🤖 *from Claude* (filed at espg's direction, 2026-07-31 — diagnosis of the red main benchmark)

## Symptom

`Lambda Benchmark` on **main** @ `5e35e2c` fails:

```
[tdigest_healpix_o9_hive] store object-count mismatch: metadata objects 7 outside [1, 5]; total objects 18 outside [12, 16]
```

Identical code passed on `claude/351-subtree-reads` @ `8d9d422` 36 minutes earlier, and #351 is a read-only feature that writes no objects — so this is not a write-path regression. The excess is **entirely root metadata** (+2 on both metadata and total; per-shard data counts are exact), which is the signature of accumulated root telemetry.

## Cause

The benchmark store path is `{BENCHMARK_STORE_PREFIX}/{variant}.zarr` (`.github/scripts/run_benchmark.py:528`) — keyed only by variant name, so **every run of every branch writes into the same store**, and there is no cleanup step in `.github/actions/run-benchmark/action.yml` or the runner scripts. Root telemetry objects are per-run unique — `stats_{ts}_{run_id}.parquet` (issue #297) and, since issue #353, `sweep_stats_{ts}.json` — so they accumulate run over run, while `expected_object_counts` models at most one of each. Twelve benchmark runs hit that shared store on 2026-07-31; the pile crossed the ceiling. Issue #353 is an accelerant (a second accumulating object per run), not the root cause.

## Fixes (both espg-approved, 2026-07-31)

**(b) Model root telemetry as its own unbounded bucket.** Give `stats_*.parquet` + `sweep_stats_*.json` an `objects_telemetry` bucket alongside the existing `objects_rollups` / `objects_overviews` D9 buckets, excluded from the audited write-path total — mirroring the precedent those two set. `metadata` then returns to the tight fixed window (manifest + optional root MOC + optional `aggregation.yaml`), so a real regression there still trips. Also close the PR #356 classifier gap in the same pass: node-level `{stem}.stats.json` overview sidecars (D23 grammar) currently fall to `other`; bucket them with the overviews when they sit at a non-leaf node, while leaf-node sidecars keep attributing to their shard.

**(c) Scope the benchmark store per run.** Include the commit under test in the store path (`{prefix}/{commit}/{variant}.zarr` — `--commit` is already plumbed into the runner and its `context`), so each run measures a store only it wrote and counts are deterministic again. Applies to the aggregation, full-AOI, and raster runners alike.

(b) restores green and is honest about a reused store; (c) removes the accumulation at the source. Note the two are complementary — (c) alone would still mis-bucket the #356 sidecars, and (b) alone leaves the shared store growing without bound.

Refs #353, #297, PR #356, issue #240 (object-count model lineage).

espg:
Closes #362

Restores the red `Lambda Benchmark` on `main` (`[tdigest_healpix_o9_hive] store object-count mismatch: metadata objects 7 outside [1, 5]; total objects 18 outside [12, 16]`) by implementing both espg-approved fixes from the issue: (b) model per-run root telemetry as its own unbounded bucket, and (c) scope the benchmark store per commit so a run measures only what it wrote.

## What / approach

**(b) `objects_telemetry` bucket — `.github/scripts/bench_objects.py`.** `stats_{ts}_{run_id}.parquet` (issue #297) and `sweep_stats_{ts}.json` (issue #353) are per-run *unique* names, so a shared store accumulates one of each per run; a fixed metadata ceiling can never be right for them. They now ride an unbounded `objects_telemetry` bucket (new `_is_root_telemetry`, wrapping the existing `_is_run_parquet` / `_is_sweep_record` grammars), counted and reported but netted out of the audited write-path total in `object_count_mismatch` — exactly the posture `objects_rollups` (issue #300) and `objects_overviews` (issue #201) already take:

```python
    total = (
        measured["objects_total"]
        - measured.get("objects_rollups", 0)
        - measured.get("objects_overviews", 0)
        - measured.get("objects_telemetry", 0)
    )
```

With telemetry out of the window, `expected_object_counts` returns to a tight fixed model: hive metadata is `[1, 3]` (manifest + root `coverage.moc` + optional `aggregation.yaml`, `metadata_min = 1` unchanged), flat metadata is exact again (`2 + len(members)`). Real root-metadata drift still trips.

**PR #356 classifier gap, same phase.** Node-level `{stem}.stats.json` (the D23 `morton-hive/3` sidecar grammar — `zagg.telemetry.sidecar_key`) fell to `other`. It now buckets with `overviews` when it sits at a node that actually holds an overview zarr (a one-pass `_overview_node` pre-collect over the listed keys), while a dispatched leaf node's sidecars keep attributing to their shard through the existing `stats_of` sibling path — the sibling grammar gained the D23 form alongside `stats.json` / `stats_{window}.json` for both the `stats` and `shardmap` stems. Everything sits inside the `for/else` of the leaf-prefix-membership check, so the issue #215 tripwire's "leaf-prefix membership FIRST" ordering is untouched: a sidecar-named object misplaced *inside* a leaf prefix still counts as that shard's data object, and a `.stats.json` anywhere else stays a loud `other` (both pinned by test).

**(c) per-commit store scoping.** One shared helper, `bench_metrics.store_path(prefix, name, commit=...)`, wired into all three runners (`run_benchmark.py`, `run_full_aoi_benchmark.py`, `run_raster_benchmark.py` — the last now imports `bench_metrics` like the other two):

```python
    prefix = store_prefix.rstrip("/")
    short = re.sub(r"[^0-9a-zA-Z]", "", commit)[:_COMMIT_COMPONENT_LEN].lower()
    return f"{prefix}/{short}/{name}.zarr" if short else f"{prefix}/{name}.zarr"
```

`{prefix}/{commit}/{name}.zarr` with a 12-character sanitized short sha; an empty `--commit` (local run, `--dry-run` wiring check) falls back to the historical unscoped path rather than emitting a `//` segment. `--commit` was already plumbed by every workflow and by `.github/actions/run-benchmark/action.yml`, so no workflow change was needed. Nothing else reconstructs the path: the full-AOI parity sibling is derived from the actual store (`store.rsplit('/', 1)[0]`), so it inherits the scope, and the harness `ListBucket` grant is conditioned on `s3:prefix` `zagg-bench/*`, which matches the deeper key unchanged.

**Consumers.** `measure_objects` returns `objects_telemetry`; `bench_metrics.build_record` and the full-AOI runner's hand-built record both carry it as a JSON-only key (the reindex to the series columns drops it, so no parquet-schema change); `run_benchmark.main`'s per-target line prints `objects=N (expected M, telemetry T)`. Visible in output, never a finding.

## Phases

- [x] Phase 1 — (b) telemetry bucket + tight metadata window + the PR #356 `{stem}.stats.json` classifier branch, with tests (`795d91e`).
- [x] Phase 2 — (c) per-commit store path in the three runners via `bench_metrics.store_path`, incl. the empty-`--commit` fallback (`7042a1d`).
- [x] Phase 3 — docs: the store layout + accumulation/lifecycle note in `docs/deployment/benchmark-cicd.md` §3, and what the audited total excludes in `docs/deployment/benchmark.md` (`91b960a`, link fix `52f2720`).

## Review round

Adversarial self-review posted 4 inline findings; 3 folded in full, 1 in part:

- `28fed7b` — the hive ceiling gated the root `coverage.moc` on `output.coverage_moc`, but `sweep.MocFamily.finish` → `hive.write_root_coverage` writes it gated only on `get_sweep` (it is in `DEFAULT_FAMILIES`). Two writers, one knob: the MOC term is now unconditional and the inert `coverage_moc=` parameter is gone from `expected_object_counts`/`measure_objects` and its call sites (zero numeric change — every caller passed `True`).
- `ac87a8c` — the new `.stats.json` branch had no locality gate, so any `.stats.json` anywhere classified as an overview. Now anchored to nodes that actually hold an overview zarr, via a `_overview_node` helper that also backs the overview-zarr branch (one predicate for "is an overview").
- `73e90bf` — `objects_telemetry` now rides the full-AOI record too (JSON-only, `None` on the dry-run null block).
- `fa664da` — pinned the D23 `{window}.shardmap.json` arm of the sibling grammar (`zagg.sweep.submap_key`).

The rendering half of the third finding (whether `_objects_cell` / `plot_summary` should display the netted write-path total instead of the raw one) was declined as a schema/semantics call — it is the third bullet under "Questions for review" below.

## How it was tested

- Full suite `uv run pytest -q` after the `main` merge: **3193 passed**, 37 skipped, 2 failed — `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` (needs network/docker) and `tests/test_client_transport.py::TestStatusPoller::test_invoke_fault_burns_an_attempt_and_retries` (timing-flaky poller test). Both are environment-only and untouched by this PR.
- `uv run ruff check --select=E,F,W,I --ignore=E501 src tests .github/scripts` clean; `ruff format --check` clean on every touched file; `uv run mkdocs build --strict` clean. One pre-existing repo-wide ruff finding unrelated to this PR: `N818` on `src/zagg/registry.py::UnknownCapability` (not selected by the CI lint bot).
- New coverage in `tests/test_benchmark_objects.py`:
  - `test_hive_root_telemetry_accumulates_without_a_finding` — twelve runs' worth of accumulated parquets + sweep records (24 objects) in one store: `objects_telemetry == 24`, metadata stays `1`, `objects_other == 0`, `object_count_mismatch(...) is None`. This is the exact shape that reddened main.
  - `test_hive_node_stats_sidecars_classified` — an overview zarr plus its `all.stats.json` / `2019.stats.json` sidecars at an ancestor node bucket as overviews; a leaf node's `stats.json`, `shardmap.json`, `2019.stats.json` and `2019.shardmap.json` attribute to their shard; an in-leaf `2019.stats.json` counts as that shard's data object; `objects_other == 0` for the whole mixed store.
  - `test_hive_stray_stats_json_stays_a_loud_other` — root-level `all.stats.json`, `random/deep/path/x.stats.json` and an undispatched leaf's node sidecar all stay `other`.
  - `test_hive_metadata_ceiling_covers_sweep_written_root_moc` — the sweep-written root MOC fits the ceiling.
- Path construction: `test_store_path_scopes_by_commit` / `test_store_path_empty_commit_falls_back_to_unscoped` (sanitization, truncation, trailing-slash prefix, empty and punctuation-only commit) plus one per runner asserting the store `main()` actually dispatches — `test_main_store_path_is_commit_scoped` (`tests/test_benchmark.py`), `test_store_path_is_commit_scoped` in `tests/test_full_aoi_benchmark.py` and `tests/test_raster_benchmark.py`, each covering both the scoped path and the empty-commit fallback.
- Updated pins: `tests/test_benchmark.py::test_hive_config_expected_counts_root_moc_optional` (now `metadata 3` / `total [12, 14]`), `test_build_record_cost_per_100km2` (record carries `objects_telemetry`), and the flat-model pins in `tests/test_benchmark_objects.py` (metadata exact again).

## PR #356 reconciliation (done, `05c2525`)

PR #356 merged to `main` while this branch was in review, so `main` is merged in here and its `tests/test_benchmark_objects.py` pins are reconciled: it had pinned the new overview sidecars as the **only** unclassified keys (`objects_other == len(overview_sidecars) > 0`) and hand-folded them out of the write-path total. This PR classifies them, so both end-to-end hive tests are back to the plain `objects_other == 0` / `object_count_mismatch(measured, expected) is None` form, with the `{window}.stats.json` sidecars netted out via `objects_overviews` like the zarrs they describe. `bench_objects.py` itself did not conflict — #356 never touched it.

## Questions for review
- **D23 leaf sidecars attribute to their shard.** The issue says "bucket node-level `{stem}.stats.json` with the overviews when it sits at a non-leaf node; leaf-node sidecars keep attributing to their shard". Implemented literally, and additionally taught the sibling grammar the D23 form — so when the `morton-hive/3` writer flips (issue #299), a leaf's `{window}.stats.json` attributes to its shard instead of reading as an overview. Say so if you'd rather the sibling grammar stayed frozen to today's names.
- **`objects_total` stays the raw LIST count** (telemetry + sweep caches included), matching how rollups/overviews already behave — only the audited total nets the buckets out. So `_objects_cell` and `plot_summary` can still render a total above `objects_expected` on a fully green run. Changing that means redefining a column already retained in `series.parquet`, which is your call: leave as-is, or switch the displayed/recorded total to the netted write-path count?
- **Store lifecycle.** (c) means the bench prefix now accumulates one directory per benchmarked commit instead of being overwritten. The docs note recommends an S3 lifecycle expiration on the prefix; setting it is an infra action outside this PR.


.github/scripts/bench_objects.py
```diff
@@ -151,17 +151,23 @@ def expected_object_counts(
...
+        # only while every run got a fresh store, and went red on main the day
+        # twelve runs shared one). They ride the unbounded ``objects_telemetry``
+        # bucket instead (:func:`_is_root_telemetry`), counted and reported but
+        # excluded from the audited write-path total — the same posture the D9
+        # rollup/overview buckets already take. This window therefore stays
+        # TIGHT (manifest + optional root MOC + optional aggregation.yaml), so
+        # real root-metadata drift still trips it.
         metadata_min = 1
-        metadata_max = 1 + (1 if coverage_moc else 0) + 1 + 1 + 1
+        metadata_max = 1 + (1 if coverage_moc else 0) + 1
```

**The tightened ceiling is now tighter than what the writers can emit when `output.coverage_moc: false`.**

The MOC term is gated on the config flag, but the store-root `coverage.moc` is written by two independent paths and only one of them honours that flag:

- `src/zagg/runner.py:1088` — the end-of-run write, gated on `get_coverage_moc(config)`;
- `src/zagg/sweep.py:237` (`MocFamily.finish` → `write_root_coverage`), reached through `sweep_after_run(store_path, leaves, store_kwargs=...)` at `runner.py:1118` with `families=None` → `DEFAULT_FAMILIES = ("stats", "moc", "submap", "overview")` (`sweep.py:61`). That path is gated on `get_sweep(config)` only — **never on `coverage_moc`**.

So a hive target with `output.coverage_moc: false` and the default sweep lands `morton_hive.json` + `aggregation.yaml` + a sweep-written `coverage.moc` = 3 root metadata objects against a ceiling of `1 + 0 + 1 = 2`, and `run_benchmark` hard-fails a store that is exactly right. Before this commit the (now removed) telemetry slack absorbed it, so the tightening is what exposes it.

Not live in CI today — all four hive benchmark configs (`atl03_tdigest_healpix_o9_hive{,_located,_strata}.yaml`, `s2_neon_o9.yaml`) leave `coverage_moc` at its hive default of `true` — but the module docstring's claim is that the model is derived from what the writers actually emit, and here it is derived from a knob one writer ignores.

Suggest making the MOC term unconditional in the ceiling (`metadata_max = 1 + 1 + 1`; `metadata_min` stays `1`, so the only cost is one object of tightness) or gating it on `coverage_moc or get_sweep(config)`, with a test pinning `coverage_moc: false` + sweep-on.

.github/scripts/bench_objects.py
```diff
@@ -365,13 +385,31 @@ def store_object_counts(
...
+                # construction ("stats.json" cannot end in ".stats.json").
                 is_sibling = any(
                     name == f"{stem}.json"
                     or (name.startswith(f"{stem}_") and name.endswith(".json"))
+                    or name.endswith(f".{stem}.json")
                     for stem in ("stats", "shardmap")
                 )
                 if is_sibling and node + "/" in stats_of:
                     per_shard[stats_of[node + "/"]] = per_shard.get(stats_of[node + "/"], 0) + 1
+                elif name.endswith(".stats.json"):
```

**This branch has no locality gate: it swallows any `*.stats.json` anywhere outside a leaf prefix, including an undispatched leaf's own sidecar.**

Every other bucket in this walk is anchored to something — the rollup branch to a closed family-name grammar, the overview-zarr branch to `_MORTON_ID_RE` on a `.zarr` basename, the sibling branch to `stats_of` membership. This one matches on the suffix alone. Probe run against this commit (dispatched leaf `1/1/2/1/1/2/1/1121121.zarr`, `shard_keys` holding only that word):

```
<undispatched node>/2019.stats.json     -> overviews   (silent)
<undispatched node>/stats.json          -> other       (loud)
all.stats.json                          -> overviews   (store ROOT)
random/deep/path/x.stats.json           -> overviews
<undispatched node>/2019.shardmap.json  -> other       (loud)
=> {'objects_overviews': 3, 'objects_other': 2, ...}
```

Two things fall out: (1) the *same* artifact class — a leaf stats sidecar at a node that is not a dispatched leaf's — is a loud `other` under the legacy name and silent under the D23 name, which inverts once the issue #299 writer flip makes D23 the leaf grammar (exactly the flip the PR body cites as the reason for teaching the sibling grammar the D23 form); (2) a `.stats.json` at the store root, or at any non-node path, now classifies as an overview.

The stray leaf is still caught through its *data* objects (a morton-id `.zarr` basename outside `leaf_of` falls to `other`), so this is a narrowing rather than a hole — but it narrows precisely the "the model claims to know every object the run writes" contract that `objects_other` exists to enforce, and the inline comment's "nor steal a real leaf sidecar's attribution" holds only for *dispatched* leaves.

An anchor is available from the input already in hand: do one pass over `keys` collecting the nodes that hold an overview `.zarr` (the branch above already identifies those keys) and require `node` to be one of them; failing that, require `node` to be a non-empty digit-node path, which at least keeps the store root and arbitrary prefixes loud.

.github/scripts/bench_metrics.py
```diff
@@ -292,6 +293,9 @@ def build_record(
     record["objects_total"] = o.get("objects_total")
     record["objects_expected"] = o.get("objects_expected")
     record["objects_per_shard"] = o.get("objects_per_shard")
+    # Per-run root telemetry (issue #362): JSON-only like per_shard/mismatch —
+    # reported so an accumulating shared store is legible, never audited.
+    record["objects_telemetry"] = o.get("objects_telemetry")
```
espg, line 342: 🤖 *from Claude (review)*

**The new key stops here; the two other object-count surfaces still show the pre-fix number.**

- `run_full_aoi_benchmark.py` builds its record by hand (`objects_total=` / `objects_expected=` / `objects_mismatch=`, lines ~658-660; documented schema at line 51) instead of through `build_record`, so on the release leg `objects_telemetry` comes back from `measure_objects` and is then dropped — that row cannot be netted after the fact, unlike the per-merge row this line fixes.
- `bench_metrics._objects_cell` (line ~386) and `plot_summary.py:517` render raw `objects_total` against `objects_expected`, so the PR-comment table cell keeps reading like `18/14` — the issue's own symptom string — on a run this PR has made green.

`objects_mismatch` is right in all three places (the netting lives in `object_count_mismatch`), so this is legibility rather than correctness, and phase 2's per-commit scoping caps the accumulation at one of each. But the rollup/overview terms already make that cell disagree with the expectation on every hive row, and this commit adds a third term to the gap while only teaching one of the three surfaces about it. Either carry the key into the full-AOI record too, or net/annotate in `_objects_cell` so one rule covers all of them.

tests/test_benchmark_objects.py
```diff
@@ -388,6 +393,84 @@ def test_hive_overview_zarrs_count_into_their_own_bucket(monkeypatch):
...
+    label = grid.shard_label(word)
+    leaf = hive.shard_leaf_path("", word).lstrip("/")
+    node = leaf.rsplit("/", 1)[0]  # the dispatched leaf's node dir
+    base = node.split("/", 1)[0]  # an ancestor node (overviews fold here)
+    keys = [
+        hive.MANIFEST_NAME,
+        f"{leaf}/count/c/0",  # in-leaf data -> this shard
+        f"{node}/stats.json",  # legacy leaf sidecar -> this shard
+        f"{node}/shardmap.json",  # leaf sub-map (issue #300) -> this shard
+        f"{node}/2019.stats.json",  # D23-named leaf sidecar -> this shard
```
espg, line 489: 🤖 *from Claude (review)*

The new sibling clause is `name.endswith(f".{stem}.json") for stem in ("stats", "shardmap")`, but only the `stats` arm is exercised — this store has a legacy `shardmap.json` at the leaf node and a D23 `2019.stats.json`, never a D23 `2019.shardmap.json`.

That untested arm is the one with no fallback: `sweep.submap_key` emits `{window}.shardmap.json` at `morton-hive/3` (`src/zagg/sweep.py:273-287`), and a D23 submap at a node outside `stats_of` lands in `other` while its stats twin lands in `overviews` — so the two halves of the same naming seam are classified asymmetrically and only one half is pinned. Adding `f"{node}/2019.shardmap.json"` to this key list (per-shard 6 instead of 5) pins the clause the PR body offers to drop.

.github/scripts/bench_objects.py

Folded in `28fed7b`. Verified the claim in the source before changing anything: `runner.agg` gates its root-MOC write on `get_coverage_moc(config)`, while `sweep.MocFamily.finish` → `hive.write_root_coverage` is reached from `sweep_after_run(store_path, leaves, ...)` with `families=None` → `DEFAULT_FAMILIES = ("stats", "moc", "submap", "overview")`, gated only on `get_sweep(config)` — two writers, one knob.

The ceiling now counts the root MOC unconditionally (floor stays the manifest alone, so the only cost is one object of tightness):

```python
metadata_min = 1
metadata_max = 1 + 1 + 1
```

with the comment above it naming both writers explicitly. Since the knob then determines nothing in the window, the now-inert `coverage_moc=` parameter is gone from `expected_object_counts` / `measure_objects` and from the three call sites (`run_benchmark._measure_objects`, `run_full_aoi_benchmark._measure_objects_recorded`, the two tests that threaded `get_coverage_moc`) rather than being left as dead API — no numeric change to any existing expectation, since every one of them passed `True`.

Pinned by `test_hive_metadata_ceiling_covers_sweep_written_root_moc` in `tests/test_benchmark_objects.py`: a hive config with `output.coverage_moc: false` asserted OFF, `get_sweep` asserted ON and `"moc" in DEFAULT_FAMILIES`, then `metadata == 3` / `metadata_min == 1` and a 3-root-object store producing no finding.

.github/scripts/bench_objects.py

Folded in `ac87a8c`, using the anchor you suggested. New `_overview_node(key)` helper returns the node dir holding the key's overview zarr (or `None`), taking the FIRST `.zarr` segment so anything nested under a real leaf's zarr can never register; a one-pass pre-collect over `keys` builds `overview_nodes`, and the branch becomes:

```python
elif name.endswith(".stats.json") and node in overview_nodes:
```

The same helper now backs the overview-zarr branch too, so one predicate defines "is an overview" in both places instead of the inline `seg = next(...)`.

Load-bearing ordering is unchanged: leaf-prefix membership still runs first (a misplaced in-leaf sidecar remains that shard's data object and still trips the issue #215 guard), and a dispatched leaf's sidecars still attribute to their shard via `stats_of`. Your three probe rows are now covered by `test_hive_stray_stats_json_stays_a_loud_other`: root-level `all.stats.json`, `random/deep/path/x.stats.json`, and an UNDISPATCHED leaf's `{node}/2019.stats.json` all land in `other` (asserted by exact `other_keys`), while `{base}/all.stats.json` beside a real `{base}/all.zarr/` still classifies as an overview.

.github/scripts/bench_metrics.py
espg: 🤖 *from Claude*

Partially folded — the full-AOI half in `73e90bf`, the rendering half declined and left standing.

**Folded:** `run_full_aoi_benchmark` now carries `objects_telemetry` on its hand-built record the same JSON-only way `bench_metrics.build_record` does — `objects_telemetry=objects.get("objects_telemetry")` in the `run.update(...)` beside the other three keys, `objects_telemetry=None` in the dry-run null block, and the key added to the module docstring's output schema. So the release row is netted the same way as the per-merge row instead of measuring the value and dropping it. Covered by the existing dry-run record test (`test_neon_catalog_builds_full_aoi_shardmap`), which now also asserts `run["objects_telemetry"] is None`.

**Declined:** the `_objects_cell` / `plot_summary.py` rendering half. Whether the displayed cell and the retained `objects_total` series should show the netted write-path count rather than the raw LIST total is a schema/semantics change to a retained parquet column (it would silently redefine an existing series against historical rows), and it is already the third bullet under "Questions for review" in the PR body. That is an espg call, not one to settle inside a fold, so it stays open there — I have not relitigated it here. As you note, `objects_mismatch` is already correct in all three places, so nothing is wrong today; only the cell's legibility is at stake.

tests/test_benchmark_objects.py
espg: 🤖 *from Claude*

Folded in `fa664da`. Added `f"{node}/2019.shardmap.json"` to this key list exactly as suggested, so the per-shard expectation goes 5 → 6, and updated the test's leading comment to say what it now pins: a dispatched leaf's sidecars attribute to their shard across BOTH stems of BOTH grammars — legacy `stats.json` / `shardmap.json` and D23 `{window}.stats.json` / `{window}.shardmap.json` (`zagg.sweep.submap_key`). The `shardmap` arm of the sibling clause is no longer unexercised.

espg:
🤖 *from Claude*

Status: all three phases are landed, the adversarial-review round is folded (4 findings — 3 in full, 1 in part, replies on every thread), and `main` is merged in with the PR #356 test pins reconciled. CI on head `05c2525` is green across the board — `test (3.12)` success, `test (3.13)` success, `ruff` success, `build` success — and the PR reads `MERGEABLE` / `CLEAN`.

Left in **draft** deliberately: the four bullets under "Questions for review" (notably whether the displayed/retained `objects_total` should become the netted write-path count) are decisions for the repo owner, not something to decide by flipping the PR ready.

espg:
🤖 *from Claude*

**Question (3) is answered and folded: no infra action needed.** espg confirms `sliderule-public` already carries a **one-month auto-delete policy**, which is exactly the retention these per-commit stores want (they are regenerable). The doc no longer tells the reader to go set a lifecycle rule — it records the existing one, and notes that a *different* bucket would need its own: 9e488f2. `mkdocs build --strict` clean.

espg:
🤖 *from Claude*

**Question (1) implemented as espg directed: 906405c.**

`objects_write_path` joins the record and the series — the netted count (gross minus D9 rollups/overviews minus unbounded per-run telemetry), which is the only figure comparable to `objects_expected`. `objects_total` keeps its gross meaning, because a reused store's real footprint is a legitimate storage/cost question and because redefining a shipped column in place would make pre-#362 rows mean something different from post-#362 rows in the same column, across a series meant to be read over months.

Details worth noting:

- The netting is now single-sourced in `bench_objects.write_path_total()`, used by BOTH `object_count_mismatch` and `measure_objects` — the reported figure and the asserted figure cannot drift apart.
- The column appends **last** in `RECORD_COLUMNS` per the repo's stable-schema rule (the three positional ordering tests were updated rather than worked around); its comment sits with it there rather than beside the objects columns it belongs with.
- The rendered `objects` cell and the run log now show the netted figure against the expectation, so a green run stops printing `objects=48 (expected 16)`. The log keeps the gross count and the telemetry count alongside it, so nothing is hidden. Pre-#362 rows fall back to the gross total — that WAS the audited figure when they were written — and rows with neither still read `n/a`; all three vintages are pinned in `test_objects_cell_rendered_in_table`.
- New `test_measure_objects_reports_the_netted_count` pins the property that matters: a store carrying 5 rollups, 3 overviews and 24 accumulated telemetry objects reports 16 and stays green.

Full suite: 3195 passed, 37 skipped, 1 failure — `test_lambda_build` (needs network), unchanged and environment-only. `ruff check`/`format --check` clean apart from the pre-existing `N818` in `registry.py`.

PR #367, opened 2026-08-03
small fixes 2026-08-03: explicit-window point entries

### LINKED ISSUE #355: Explicit windows: accept {label, timestamp} point entries (desugar to a second-wide half-open range)
🤖 *from Claude* (filed at espg's direction, 2026-07-31)

## Motivation

`schedule: explicit` already supports non-contiguous and singleton windows, but each entry must be a `{label, start, end}` range — clunky when the product's time axis is effectively discrete (single acquisitions, scene timestamps, isolated campaign instants). A 2-member `{label, timestamp}` entry currently fails loudly at validation (`each explicit window must be a {label, start, end} mapping` — `config.py::_validate_windowing_windows`); there is no graceful point form.

## Design (espg-approved shape)

Accept `{label, timestamp}` as **sugar that desugars to a range at the config layer**:

- Each explicit entry carries exactly one of `timestamp` **or** `start`+`end`, discriminated on keys (entries are mappings, not tuples); mixing the two forms in one entry raises.
- `timestamp: t` desugars to the half-open, second-wide `[t, t+1s)` inside `get_windowing` normalization, so **everything downstream is untouched** — `windows_intersecting`, the catalog time filters, `windowed_cell_config`, leaf naming, the sweep, and the pyramid all see an ordinary explicit window. Graceful degradation by construction.
- Why one second: the window grammar's native resolution is already whole seconds (`timespec="seconds"` throughout; `windows.py` explicitly scopes second-scale exactness out), so a second-wide half-open window *is* the point window at grammar resolution — and it avoids introducing float equality on observation timestamps (a footgun for point-cloud data; discrete scene timestamps fall inside their second).
- Existing validation composes unchanged: unique labels, the reserved `all` token, and pairwise disjointness — a desugared point inside another window's range is a genuine overlap and correctly rejects.

## Edge (documented, not solved)

Two acquisitions within the same wall-clock second share a window. If that ever bites, the fix is an explicit `width` key on the point form — not a guessed default now.

## Files likely touched

`src/zagg/config.py` (`_validate_windowing_windows` + `get_windowing` desugar), tests in `tests/test_config.py` (both forms, mixing raises, point-in-range overlap rejects, desugared window round-trips through `window_time_filters`), and the windowing section of the docs.

Refs #246 (explicit windowing), #201/PR #344 (per-window overviews are downstream consumers and need no change).

espg:
Bundled `small-fix` PR for 2026-08-03. Only one open `small-fix` issue exists today, so the bundle carries a single entry; the shape leaves room to append another.

## Bundle

- [x] Closes #355 — explicit windows accept `{label, timestamp}` point entries (desugar to a second-wide half-open range)
- [x] Range form rejects unknown keys (was "Questions for review" (5); fixed at espg's direction, 2026-08-05 — **behavior change**, described below)
- [x] Degenerate sub-second range entries refuse loudly (was "Questions for review" (6); fixed at espg's direction, 2026-08-05)

## What this changes (#355)

`schedule: explicit` windows previously had to be `{label, start, end}`; a 2-member `{label, timestamp}` entry failed with `each explicit window must be a {label, start, end} mapping`. This adds the point form as **sugar desugared at the config layer**:

- Each entry declares exactly one of `timestamp` **or** `start`+`end`, discriminated on keys; mixing the two in one entry raises, and the error names *every* offending range key so one edit fixes the entry.
- `timestamp: t` becomes the half-open `[t, t + 1s)`. One second is the window grammar's native resolution (boundaries render at `timespec="seconds"` throughout, and `windows.py` explicitly scopes second-scale exactness out), so a second-wide half-open window *is* the point window at grammar resolution — and it avoids float equality on observation timestamps.
- Any key on a point entry other than `label`/`timestamp` is **rejected**, so the documented-but-unimplemented `width` knob fails loud instead of being silently ignored.

Approach: one shared helper, `config._explicit_window_bounds(entry) -> (start, end)`, used by **both** `_validate_windowing_windows` and `get_windowing`, so validation and normalization can never disagree about what an entry means:

```python
    ranged = sorted({"start", "end"} & set(entry))
    if "timestamp" in entry:
        if ranged:
            raise ValueError(...)          # mixing the two forms
        extra = sorted(set(entry) - {"label", "timestamp"})
        if extra:
            raise ValueError(...)          # e.g. the not-yet-real `width`
        point = _windows.parse_utc(entry["timestamp"])
        return point, point + _POINT_WINDOW_WIDTH
    return _windows.parse_utc(entry["start"]), _windows.parse_utc(entry["end"])
```

`get_windowing`'s normalized `windows` list stays uniformly `{label, start, end}` — the desugar happens before it is built — so **everything downstream is untouched**: `windows_intersecting`, `window_time_filters`, `windowed_cell_config`, leaf naming, the sweep, and the pyramid all see an ordinary explicit window, and `hive.build_manifest`'s `temporal.windows` block (`src/zagg/hive.py`, which copies `windowing["windows"]` verbatim) never sees a `timestamp` key. No wire format, attrs grammar, or `spec` marker changes, so `docs/specification.md` and the conformance fixtures need no update.

Existing validation composes unchanged because it runs on the desugared bounds: frozen label grammar, the reserved `all` token, `start < end`, unique labels, and pairwise disjointness — a point that lands inside another window's range is a genuine overlap and is rejected.

**Documented edges (not solved, per the issue):**

- *Collision.* Two acquisitions within the same wall-clock second share a window. The stated fix if it ever bites is an explicit `width` key on the point form, not a guessed default now — and until that key exists it is rejected rather than ignored.
- *Miss, and it is silent.* A point window covers only the whole second containing `t`, so the declared `timestamp` must be the acquisition's own instant copied from the item, not a rounded-off approximation. STAC datetimes are rarely round seconds, so this is the likely first-contact mistake. On the point pipeline the window dispatches and the worker's `ge`/`lt` filter matches nothing (empty leaf); on raster, `runner._raster_windowed_units` drops the group at dispatch and no unit, leaf, or warning is produced at all. Neither path errors. Documented in `docs/hive_layout.md` and pinned by a raster test.

**Sub-second timestamps**, for the record: bounds are validated at declared precision but rendered through `windows.iso_utc` (`timespec="seconds"`), so `timestamp: "…T12:00:00.75Z"` normalizes to the whole second containing it, `[12:00:00, 12:00:01)`. Truncation is monotone, so it can neither move `t` out of its own window nor manufacture an overlap. Documented in the helper's docstring and pinned by a test.

## Review follow-ups fixed at espg's direction (2026-08-05)

Both were originally flagged-not-fixed under "Questions for review" ((5) and (6) below); espg ruled they be fixed inside this PR rather than filed as new issues.

**Range form rejects unknown keys** (`0de3d0a`). The `{label, start, end}` form now refuses unsupported keys exactly like the point form — same error style, naming every offending key and the allowed set (`explicit range window entry carries unsupported key(s) …; the range form is {label, start, end}`).

- **This is a deliberate behavior change**: a range entry carrying extra keys (say a typo'd `stop`, or a stale `width`) previously validated clean with the extras silently ignored; it now fails `validate_config`. Unknown keys have no semantic effect today, so the only configs broken are ones carrying latent typos or dead keys — which is the point of the change.

**Degenerate sub-second ranges refuse loudly** (`2e54462`). `{start: "…T12:00:01.0Z", end: "…T12:00:01.4Z"}` previously passed validation (`start < end` holds on the untruncated values) but both bounds render through `iso_utc` (`timespec="seconds"`) to the same instant, dispatching a `ge x` / `lt x` pair that silently matches nothing. `_explicit_window_bounds` now validates the *rendered* bounds: when the declared bounds satisfy `start < end` but truncate to the same whole second, it refuses with `explicit window range is empty after truncation to whole-second granularity: … both render as …; use the point form {label, timestamp} for a one-second window`. The range is never widened to rescue it — silent widening is worse than refusal, and the point form is the spelling for exactly-one-second intent. A sub-second-wide range that straddles a second boundary (e.g. `12:00:01.9Z → 12:00:02.1Z`) still renders to a real one-second window and remains valid; the point form is untouched.

## Phases

- [x] Phase 1 (`0401b25`) — `_explicit_window_bounds` desugar helper + `_validate_windowing_windows` / `get_windowing` wiring (`src/zagg/config.py`)
- [x] Phase 2 (`d0c8406`) — tests (`tests/test_hive_windows.py`)
- [x] Phase 3 (`7bd922f`) — docs (windowing section of `docs/hive_layout.md`)
- [x] In-context self-review + fold (`1f2542b`, `5b36af3`, `c11daaa`)
- [x] Fresh-context adversarial review + fold, five findings, one commit each: raster-path membership test (`fb3fd9d`), reject unsupported keys on a point entry (`614c7bd`), document the silent-miss edge (`f8db8a0`), name every offending range key in the mixed-form error (`1653707`), cut the duplicated prose (`6cd5bdc`)
- [x] Review follow-ups per espg direction, 2026-08-05: range-form unknown-key rejection (`0de3d0a`), degenerate sub-second range refusal (`2e54462`) — code, tests, and `docs/hive_layout.md` in each commit
- [x] Fresh-context adversarial review of the two follow-up commits + fold, three findings, one commit each: document range-bound truncation and name the retiming (`3d0576a`), compare rendered bounds through `iso_utc` (`f06a1e7`), name non-string extra keys instead of raising `TypeError` (`a34b7cf`)

## How it was tested

New coverage in `tests/test_hive_windows.py` (`TestWindowingConfig` + `TestWindowedUnits`), alongside the existing explicit-window tests:

- both forms in one `windows` list normalize to `{label, start, end}`, the point one second wide;
- a sub-second timestamp normalizes to its containing whole second;
- mixing `timestamp` with `start`/`end` in one entry is rejected (parametrized over all three shapes, each pinning the keys the message names);
- an unsupported key on a point entry — `width` included — is rejected;
- an unsupported key on a **range** entry is rejected, the message naming every offender and the allowed set (parametrized: one extra key, and two so the multi-key naming is pinned);
- a sub-second range that collapses at whole-second rendering is refused, the message naming the truncation;
- a sub-second-wide range spanning a second boundary stays valid and normalizes to its real one-second window;
- an entry with neither form — or half a range — is rejected;
- an unparseable `timestamp` is rejected;
- the reserved `all` token and duplicate labels reject through the point form too;
- a point inside another window's range is rejected as an overlap;
- two points inside the same wall-clock second reject as overlapping, one second apart merely touch;
- round trip on the point pipeline: a point window dispatches through `runner._windowed_units` (subsetting granules by their shardmap spans), its dataset-unit bounds are 1.0 apart, and `windowed_cell_config` injects a `ge`/`lt` filter pair exactly 1.0 apart;
- round trip on the **raster** path: `runner._raster_windowed_units` places scenes at `t` and `t + 0.9s` in the window, and dispatches **nothing at all** for a scene at exactly `t + 1s` (the half-open boundary) or at a realistic `12:00:31.024Z` (the silent-miss edge).

Gates: `uv sync --extra test`; `uv run pytest -v` → full suite green (results for the follow-up commits in the PR thread); `uv run ruff check` and `uv run ruff format --check` clean on every file this PR touches, and clean across `src tests` apart from the two pre-existing failures noted below.

## Questions for review

1. **Tests landed in `tests/test_hive_windows.py`, not `tests/test_config.py`.** The issue named `tests/test_config.py`, but the whole `output.windowing` block is covered by `TestWindowingConfig` in `tests/test_hive_windows.py` (`test_explicit_normalized`, `test_explicit_overlap`, `test_explicit_touching_ranges_allowed`, …) and `tests/test_config.py` has no windowing coverage at all. Point tests went next to their siblings. Say the word if they should move.
2. **Two pre-existing local gate failures, untouched here** (both also fail on `main` at `083af16`, and neither is reachable from this diff):
   - `ruff check src tests` → `N818` on `src/zagg/registry.py:64` (`class UnknownCapability(KeyError)`). The PR lint bot runs `--select=E,F,W,I`, so it does not see `N`; `pyproject.toml` selects `N`, so the local gate does.
   - `ruff format --check src tests` → `tests/data/benchmark/README.md` would be reformatted (a python code block inside a fixture README).

   Flagged rather than fixed, per the "do not fix unrelated pre-existing failures" rule. Happy to take either as a follow-up `small-fix` if wanted.
3. **`_POINT_WINDOW_WIDTH` is a module constant** in `config.py` rather than living in `windows.py`. It is a config-layer sugar detail that never reaches the window primitives, so it stayed with the desugar — moving it to `windows.py` is a one-liner if the primitive module is the preferred home.
4. **`src/zagg/config.py` is 2841 lines**, well past CLAUDE.md §4's ~1,200 guidance. It was 2785 lines on `main` at `083af16`, so the overage is pre-existing and this diff adds ~56 lines to an existing function pair rather than a new module's worth. Splitting a 2.8k-line config module is not a `small-fix`; raising it so the decision is on the record. #349 already owns config splitting, so that looks like the right home for it — confirming rather than assuming.
5. ~~**Should the `{label, start, end}` range form reject unsupported keys too?**~~ **Resolved — fixed at espg's direction, 2026-08-05** (`0de3d0a`). See "Review follow-ups" above; the unknown-key behavior change is described there.
6. ~~**A pre-existing bug found while reviewing, deliberately not fixed here:** a sub-second *range* entry normalizes to a degenerate empty window.~~ **Resolved — fixed at espg's direction, 2026-08-05** (`2e54462`). Validation now runs on the rendered bounds and refuses the collapse by name; see "Review follow-ups" above.

### DIFF
--- a/docs/hive_layout.md
+++ b/docs/hive_layout.md
@@ -182,7 +182,11 @@ Validation: `output.windowing` requires the hive layout on a healpix grid;
 filter what it reads); explicit windows must be well-formed (frozen label
 grammar, `start < end`, unique labels, disjoint ranges — point entries are
 desugared first, so a point landing inside another window's range is a genuine
-overlap and is rejected). On the raster path
+overlap and is rejected). Because boundaries render at whole-second
+granularity, a range narrower than the second containing it (say
+`12:00:01.0Z` → `12:00:01.4Z`) would dispatch as an empty window; it is
+refused rather than widened — the point form is the spelling for one-second
+intent. On the raster path
 ([issue #247](https://github.com/englacial/zagg/issues/247)) membership is
 the acquisition's STAC `datetime`: `time_field` is optional (fixed to
 `datetime`) and the `epoch`/`scale`/`units` conversion knobs are rejected.
--- a/docs/hive_layout.md
+++ b/docs/hive_layout.md
@@ -112,6 +112,7 @@ output:
     windows:                        # explicit schedule only:
       - {label: melt-2019, start: "2019-06-01", end: "2019-09-01"}
       - {label: melt-2020, start: "2020-06-01", end: "2020-09-01"}
+      - {label: scene-a, timestamp: "2021-03-14T12:00:00Z"}  # point form
 ```
 
 - **Leaf naming is frozen**: `{full_id}_{window}.zarr`, underscore separator,
@@ -120,6 +121,19 @@ output:
   chronological order; explicit labels are opaque (`[0-9A-Za-z-]{1,32}`) and
   decode only through the declared list. `quarterly` is grammar-reserved but
   not implemented (validation rejects it).
+- **An explicit entry may be a point** ([issue #355](https://github.com/englacial/zagg/issues/355)):
+  `{label, timestamp}` is sugar for the second-wide half-open `[t, t + 1s)`,
+  for time axes that are effectively discrete (single acquisitions, scene
+  timestamps, isolated campaign instants). Each entry declares *exactly one* of
+  `timestamp` or `start`+`end` — mixing them in one entry is rejected. The
+  desugaring happens in `zagg.config.get_windowing`, so the manifest and every
+  downstream consumer only ever see an ordinary `{label, start, end}` window.
+  One second is the grammar's own resolution (boundaries are whole seconds
+  throughout), which also keeps membership off float equality on observation
+  timestamps. *Consequence*: two acquisitions within the same wall-clock second
+  share a window, and two point entries inside one second are rejected as
+  overlapping — if that ever bites, the fix is an explicit `width` key on the
+  point form, not a guessed default.
 - **Boundaries are UTC calendar terms, half-open `[start, end)`.** Window
   bounds are converted to dataset units once at dispatch, using the declared
   `epoch`/`scale`/`units` and a fixed scale offset (`GPS−UTC = 18 s`,
@@ -149,7 +163,9 @@ output:
 Validation: `output.windowing` requires the hive layout on a healpix grid;
 `time_field` must be a declared `data_source` column (the worker can only
 filter what it reads); explicit windows must be well-formed (frozen label
-grammar, `start < end`, unique labels, disjoint ranges). On the raster path
+grammar, `start < end`, unique labels, disjoint ranges — point entries are
+desugared first, so a point landing inside another window's range is a genuine
+overlap and is rejected). On the raster path
 ([issue #247](https://github.com/englacial/zagg/issues/247)) membership is
 the acquisition's STAC `datetime`: `time_field` is optional (fixed to
 `datetime`) and the `epoch`/`scale`/`units` conversion knobs are rejected.
--- a/docs/hive_layout.md
+++ b/docs/hive_layout.md
@@ -130,10 +130,13 @@ output:
   downstream consumer only ever see an ordinary `{label, start, end}` window.
   One second is the grammar's own resolution (boundaries are whole seconds
   throughout), which also keeps membership off float equality on observation
-  timestamps. *Consequence*: two acquisitions within the same wall-clock second
-  share a window, and two point entries inside one second are rejected as
-  overlapping — if that ever bites, the fix is an explicit `width` key on the
-  point form, not a guessed default.
+  timestamps (a sub-second `timestamp` normalizes to the whole second
+  containing it). *Consequence*: two acquisitions within the same wall-clock
+  second share a window, and two point entries inside one second are rejected
+  as overlapping — as is a point that lands inside an already-declared range
+  window, which is an ordinary overlap (see the validation paragraph below).
+  If that ever bites, the fix is an explicit `width` key on the point form, not
+  a guessed default.
 - **Boundaries are UTC calendar terms, half-open `[start, end)`.** Window
   bounds are converted to dataset units once at dispatch, using the declared
   `epoch`/`scale`/`units` and a fixed scale offset (`GPS−UTC = 18 s`,
--- a/docs/hive_layout.md
+++ b/docs/hive_layout.md
@@ -138,8 +138,9 @@ output:
   as overlapping — as is a point that lands inside an already-declared range
   window, which is an ordinary overlap (see the validation paragraph below).
   If that ever bites, the fix is an explicit `width` key on the point form, not
-  a guessed default; until it exists, any key on a point entry other than
-  `label`/`timestamp` (`width` included) is rejected rather than ignored.
+  a guessed default; until it exists, any key on an entry beyond its own form's
+  (`label`/`timestamp` or `label`/`start`/`end`; `width` included) is rejected
+  rather than ignored.
 - **A point window that matches nothing fails SILENTLY** — it covers only the
   whole second containing `t`, so the declared `timestamp` must be the
   acquisition's own instant, copied from the item, not a rounded-off
--- a/docs/hive_layout.md
+++ b/docs/hive_layout.md
@@ -112,7 +112,9 @@ output:
     windows:                        # explicit schedule only:
       - {label: melt-2019, start: "2019-06-01", end: "2019-09-01"}
       - {label: melt-2020, start: "2020-06-01", end: "2020-09-01"}
-      - {label: scene-a, timestamp: "2021-03-14T12:00:00Z"}  # point form
+      - {label: scene-a, timestamp: "2021-03-14T12:00:31.024Z"}  # point form:
+                                    #   the acquisition's OWN instant, copied
+                                    #   from the item — not a rounded-off one
 ```
 
 - **Leaf naming is frozen**: `{full_id}_{window}.zarr`, underscore separator,
@@ -136,7 +138,18 @@ output:
   as overlapping — as is a point that lands inside an already-declared range
   window, which is an ordinary overlap (see the validation paragraph below).
   If that ever bites, the fix is an explicit `width` key on the point form, not
-  a guessed default.
+  a guessed default; until it exists, any key on a point entry other than
+  `label`/`timestamp` (`width` included) is rejected rather than ignored.
+- **A point window that matches nothing fails SILENTLY** — it covers only the
+  whole second containing `t`, so the declared `timestamp` must be the
+  acquisition's own instant, copied from the item, not a rounded-off
+  approximation of it. STAC datetimes are rarely round seconds, so this is the
+  likely first-contact mistake: `timestamp: "2021-03-14T12:00:00Z"` matches
+  nothing for a scene whose `datetime` is `2021-03-14T12:00:31.024Z`. On the
+  point pipeline the window still dispatches and the worker's `ge`/`lt` filter
+  matches nothing, leaving an empty leaf; on raster the group is dropped at
+  dispatch (`runner._raster_windowed_units`) and no work unit, leaf, or warning
+  is produced at all. Neither path errors.
 - **Boundaries are UTC calendar terms, half-open `[start, end)`.** Window
   bounds are converted to dataset units once at dispatch, using the declared
   `epoch`/`scale`/`units` and a fixed scale offset (`GPS−UTC = 18 s`,
--- a/docs/hive_layout.md
+++ b/docs/hive_layout.md
@@ -182,11 +182,18 @@ Validation: `output.windowing` requires the hive layout on a healpix grid;
 filter what it reads); explicit windows must be well-formed (frozen label
 grammar, `start < end`, unique labels, disjoint ranges — point entries are
 desugared first, so a point landing inside another window's range is a genuine
-overlap and is rejected). Because boundaries render at whole-second
-granularity, a range narrower than the second containing it (say
-`12:00:01.0Z` → `12:00:01.4Z`) would dispatch as an empty window; it is
-refused rather than widened — the point form is the spelling for one-second
-intent. On the raster path
+overlap and is rejected). Range bounds render at whole-second granularity
+exactly as a point `timestamp` does — **each bound truncates to the second
+containing it** — so a fractional bound silently retimes that edge of the
+window: `[12:00:00.0Z, 12:00:01.5Z)` dispatches as
+`[12:00:00, 12:00:01)`, dropping the declared half second of coverage at the
+tail. A range is never widened to recover the truncated fraction. The one
+case refused outright is total collapse: when *both* bounds render to the
+same second (say `12:00:01.0Z` → `12:00:01.4Z`) the window dispatches as an
+empty `ge x`/`lt x` pair, so it is rejected — the point form is the spelling
+for one-second intent. A sub-second range that *straddles* a second boundary
+(`12:00:01.9Z` → `12:00:02.1Z`) is still valid; it renders to the one-second
+window its truncated bounds describe. On the raster path
 ([issue #247](https://github.com/englacial/zagg/issues/247)) membership is
 the acquisition's STAC `datetime`: `time_field` is optional (fixed to
 `datetime`) and the `epoch`/`scale`/`units` conversion knobs are rejected.
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1091,6 +1091,16 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
                 f"entry declares exactly one of {{timestamp}} or {{start, end}} "
                 f"(got {entry!r})"
             )
+        extra = sorted(set(entry) - {"label", "timestamp"})
+        if extra:
+            # `width` is named as the escape hatch for the same-second edge but
+            # does not exist yet — accepting it silently would hand back a
+            # one-second window and a clean validate_config (review finding).
+            raise ValueError(
+                f"explicit point window entry carries unsupported key(s) "
+                f"{', '.join(extra)}; the point form is {{label, timestamp}} "
+                f"(got {entry!r})"
+            )
         point = _windows.parse_utc(entry["timestamp"])
         return point, point + _POINT_WINDOW_WIDTH
     return _windows.parse_utc(entry["start"]), _windows.parse_utc(entry["end"])
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1091,21 +1091,26 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
         # `width` is documented as the future escape hatch but does not exist
         # yet: accepting it silently would hand back a 1 s window and a clean
         # validate_config.
-        extra = sorted(set(entry) - {"label", "timestamp"})
+        extra = sorted(set(entry) - {"label", "timestamp"}, key=repr)
         if extra:
             raise ValueError(
                 f"explicit point window entry carries unsupported key(s) "
-                f"{', '.join(extra)}; the point form is {{label, timestamp}} "
-                f"(got {entry!r})"
+                f"{', '.join(map(repr, extra))}; the point form is "
+                f"{{label, timestamp}} (got {entry!r})"
             )
         point = _windows.parse_utc(entry["timestamp"])
         return point, point + _POINT_WINDOW_WIDTH
-    extra = sorted(set(entry) - {"label", "start", "end"})
+    # repr, not str: YAML 1.1 resolves bare `on`/`no`/numerals to bool/int
+    # keys, which are neither sortable against strings nor joinable — the
+    # TypeError would escape _validate_windowing_windows' ValueError wrapper
+    # and lose the `output.windowing.windows:` prefix. repr also shows the
+    # reader that the key came back coerced.
+    extra = sorted(set(entry) - {"label", "start", "end"}, key=repr)
     if extra:
         raise ValueError(
             f"explicit range window entry carries unsupported key(s) "
-            f"{', '.join(extra)}; the range form is {{label, start, end}} "
-            f"(got {entry!r})"
+            f"{', '.join(map(repr, extra))}; the range form is "
+            f"{{label, start, end}} (got {entry!r})"
         )
     start = _windows.parse_utc(entry["start"])
     end = _windows.parse_utc(entry["end"])
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1087,7 +1087,7 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
     if "timestamp" in entry:
         if ranged:
             raise ValueError(
-                f"explicit window entry carries both timestamp and {ranged[0]}; each "
+                f"explicit window entry carries both timestamp and {', '.join(ranged)}; each "
                 f"entry declares exactly one of {{timestamp}} or {{start, end}} "
                 f"(got {entry!r})"
             )
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1073,6 +1073,13 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
     range. Two acquisitions inside the same wall-clock second therefore share a
     window; if that ever bites, the fix is an explicit ``width`` key on the
     point form, not a guessed default.
+
+    Bounds are returned at the precision declared, but :func:`get_windowing`
+    renders them through ``windows.iso_utc`` (``timespec="seconds"``), so a
+    SUB-SECOND ``timestamp`` normalizes to the whole second CONTAINING it —
+    both ends truncate by the same fraction, so ``t`` still falls inside its
+    own window, and disjointness is preserved (validated-disjoint points are
+    >= 1 s apart, hence so are their truncated bounds).
     """
     from zagg import windows as _windows
 
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1070,10 +1070,13 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
     renders them through ``windows.iso_utc`` (``timespec="seconds"``), so a
     sub-second ``timestamp`` normalizes to the whole second containing it —
     truncation is monotone, so it can neither move ``t`` out of its own window
-    nor manufacture an overlap. A *range* whose bounds collapse to the same
-    rendered second would dispatch as an empty window, so it is refused here,
-    on the rendered values. See ``docs/hive_layout.md`` for the same-second
-    collision and the silent-miss edge.
+    nor manufacture an overlap. *Range* bounds truncate identically, each to
+    the second containing it, so a fractional ``start``/``end`` silently
+    retimes that edge of the window; a range is never widened to recover the
+    truncated fraction. The one refusal is total collapse — bounds that render
+    to the same second would dispatch as an empty window, so it is rejected
+    here, on the rendered values. See ``docs/hive_layout.md`` for the
+    bound-truncation edge, the same-second collision and the silent-miss edge.
     """
     from zagg import windows as _windows
 
@@ -1107,10 +1110,13 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
     start = _windows.parse_utc(entry["start"])
     end = _windows.parse_utc(entry["end"])
     if start < end and start.replace(microsecond=0) == end.replace(microsecond=0):
-        # Bounds render at whole-second granularity (windows.iso_utc,
-        # timespec="seconds"), so a sub-second range collapses to a `ge x` /
-        # `lt x` pair that silently matches nothing. Refuse rather than widen
-        # — the point form is the spelling for one-second intent.
+        # Each bound truncates to the second containing it (windows.iso_utc,
+        # timespec="seconds"), so a range with both bounds inside one second
+        # renders to a `ge x` / `lt x` pair that silently matches nothing.
+        # Refuse rather than widen — the point form is the spelling for
+        # one-second intent. Only this total collapse is refused: a fractional
+        # bound in a *different* second merely retimes that edge (documented in
+        # docs/hive_layout.md).
         raise ValueError(
             f"explicit window range is empty after truncation to whole-second "
             f"granularity: start {entry['start']!r} and end {entry['end']!r} "
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -6,6 +6,7 @@ import logging
 import re
 from collections.abc import Callable
 from dataclasses import dataclass, field
+from datetime import datetime, timedelta
 from importlib import resources
 from typing import Any, NotRequired, TypedDict
 
@@ -1052,13 +1053,53 @@ def _validate_windowing(config: PipelineConfig) -> None:
     _validate_windowing_windows(block, schedule)
 
 
+#: A ``{label, timestamp}`` point entry desugars to a window this wide (issue
+#: #355). One second is the window grammar's native resolution — boundaries
+#: are whole seconds throughout (``iso_utc`` renders ``timespec="seconds"``)
+#: — so a second-wide half-open window *is* the point window at grammar
+#: resolution, without introducing float equality on observation timestamps.
+_POINT_WINDOW_WIDTH = timedelta(seconds=1)
+
+
+def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
+    """The half-open UTC ``[start, end)`` of one explicit ``windows`` entry.
+
+    Two forms, discriminated on keys and never mixed (issue #355): the range
+    form ``{label, start, end}``, and the point sugar ``{label, timestamp}``
+    which desugars to ``[t, t + 1s)`` (:data:`_POINT_WINDOW_WIDTH`). Desugaring
+    happens here, at the config layer, so everything downstream —
+    ``windows_intersecting``, the catalog time filters, ``windowed_cell_config``,
+    leaf naming, the sweep, the pyramid — only ever sees an ordinary explicit
+    range. Two acquisitions inside the same wall-clock second therefore share a
+    window; if that ever bites, the fix is an explicit ``width`` key on the
+    point form, not a guessed default.
+    """
+    from zagg import windows as _windows
+
+    ranged = sorted({"start", "end"} & set(entry))
+    if "timestamp" in entry:
+        if ranged:
+            raise ValueError(
+                f"explicit window entry carries both timestamp and {ranged[0]}; each "
+                f"entry declares exactly one of {{timestamp}} or {{start, end}} "
+                f"(got {entry!r})"
+            )
+        point = _windows.parse_utc(entry["timestamp"])
+        return point, point + _POINT_WINDOW_WIDTH
+    return _windows.parse_utc(entry["start"]), _windows.parse_utc(entry["end"])
+
+
 def _validate_windowing_windows(block: dict, schedule: str) -> None:
     """Validate the ``windows`` list half of ``output.windowing`` (issue #246).
 
     Shared by the point-pipeline and raster branches of
     :func:`_validate_windowing`: frozen label grammar, half-open
     ``start < end``, unique labels, non-overlapping ranges — required for
-    ``schedule: explicit``, rejected otherwise.
+    ``schedule: explicit``, rejected otherwise. Point entries
+    (``{label, timestamp}``, issue #355) are desugared by
+    :func:`_explicit_window_bounds` before these checks, so they compose
+    unchanged: a point landing inside another window's range is a genuine
+    overlap and rejects.
     """
     from zagg import windows as _windows
 
@@ -1073,17 +1114,20 @@ def _validate_windowing_windows(block: dict, schedule: str) -> None:
     if not isinstance(declared_windows, list) or not declared_windows:
         raise ValueError(
             "output.windowing.windows is required for schedule: explicit — a "
-            "non-empty list of {label, start, end} entries"
+            "non-empty list of {label, start, end} (or {label, timestamp}) entries"
         )
     seen: dict = {}
     for entry in declared_windows:
-        if not isinstance(entry, dict) or not {"label", "start", "end"} <= set(entry):
+        if not isinstance(entry, dict) or not (
+            "label" in entry and ({"start", "end"} <= set(entry) or "timestamp" in entry)
+        ):
             raise ValueError(
-                f"each explicit window must be a {{label, start, end}} mapping (got {entry!r})"
+                f"each explicit window must be a {{label, start, end}} or "
+                f"{{label, timestamp}} mapping (got {entry!r})"
             )
         try:
             label = _windows.validate_label(entry["label"], "explicit")
-            start, end = _windows.parse_utc(entry["start"]), _windows.parse_utc(entry["end"])
+            start, end = _explicit_window_bounds(entry)
         except ValueError as e:
             raise ValueError(f"output.windowing.windows: {e}") from e
         if label == _windows.SCHEDULE_NONE_TOKEN:
@@ -1098,6 +1142,7 @@ def _validate_windowing_windows(block: dict, schedule: str) -> None:
                 f"declare a different explicit label"
             )
         if not start < end:
+            # Range form only — a desugared point is half-open by construction.
             raise ValueError(
                 f"explicit window {label!r} is not half-open: start "
                 f"{entry['start']!r} must precede end {entry['end']!r}"
@@ -2500,7 +2545,10 @@ def get_windowing(config: PipelineConfig) -> dict | None:
         {"schedule", "time_field", "epoch", "scale", "units", "windows"}
 
     ``epoch`` and explicit-window boundaries are canonicalized to ISO-8601
-    UTC strings; ``windows`` is ``None`` except for ``schedule: explicit``.
+    UTC strings; ``windows`` is ``None`` except for ``schedule: explicit``,
+    where a ``{label, timestamp}`` point entry is desugared to its second-wide
+    range (issue #355) so the normalized list is uniformly ``{label, start,
+    end}``.
     On the raster branch (``reader: raster``) ``time_field`` is the fixed STAC
     ``datetime`` and ``epoch``/``scale``/``units`` are hardcoded to the
     Unix-epoch UTC-seconds encoding any ISO instant normalizes to, rather than
@@ -2516,14 +2564,19 @@ def get_windowing(config: PipelineConfig) -> dict | None:
         return None
     declared = None
     if block["schedule"] == "explicit":
-        declared = [
-            {
-                "label": w["label"],
-                "start": _windows.iso_utc(_windows.parse_utc(w["start"])),
-                "end": _windows.iso_utc(_windows.parse_utc(w["end"])),
-            }
-            for w in block["windows"]
-        ]
+        # Point entries desugar HERE (issue #355): the normalized list is
+        # always the {label, start, end} range form, so no consumer — manifest
+        # temporal block included — ever sees a ``timestamp`` key.
+        declared = []
+        for w in block["windows"]:
+            start, end = _explicit_window_bounds(w)
+            declared.append(
+                {
+                    "label": w["label"],
+                    "start": _windows.iso_utc(start),
+                    "end": _windows.iso_utc(end),
+                }
+            )
     if (config.data_source or {}).get("reader") == "raster":
         # Raster membership is the acquisition's STAC ``datetime`` (issue
         # #247, ratified): the manifest records the resolved field plus the
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1070,7 +1070,9 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
     renders them through ``windows.iso_utc`` (``timespec="seconds"``), so a
     sub-second ``timestamp`` normalizes to the whole second containing it —
     truncation is monotone, so it can neither move ``t`` out of its own window
-    nor manufacture an overlap. See ``docs/hive_layout.md`` for the same-second
+    nor manufacture an overlap. A *range* whose bounds collapse to the same
+    rendered second would dispatch as an empty window, so it is refused here,
+    on the rendered values. See ``docs/hive_layout.md`` for the same-second
     collision and the silent-miss edge.
     """
     from zagg import windows as _windows
@@ -1102,7 +1104,20 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
             f"{', '.join(extra)}; the range form is {{label, start, end}} "
             f"(got {entry!r})"
         )
-    return _windows.parse_utc(entry["start"]), _windows.parse_utc(entry["end"])
+    start = _windows.parse_utc(entry["start"])
+    end = _windows.parse_utc(entry["end"])
+    if start < end and start.replace(microsecond=0) == end.replace(microsecond=0):
+        # Bounds render at whole-second granularity (windows.iso_utc,
+        # timespec="seconds"), so a sub-second range collapses to a `ge x` /
+        # `lt x` pair that silently matches nothing. Refuse rather than widen
+        # — the point form is the spelling for one-second intent.
+        raise ValueError(
+            f"explicit window range is empty after truncation to whole-second "
+            f"granularity: start {entry['start']!r} and end {entry['end']!r} "
+            f"both render as {_windows.iso_utc(start)!r}; use the point form "
+            f"{{label, timestamp}} for a one-second window"
+        )
+    return start, end
 
 
 def _validate_windowing_windows(block: dict, schedule: str) -> None:
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -1053,11 +1053,9 @@ def _validate_windowing(config: PipelineConfig) -> None:
     _validate_windowing_windows(block, schedule)
 
 
-#: A ``{label, timestamp}`` point entry desugars to a window this wide (issue
-#: #355). One second is the window grammar's native resolution — boundaries
-#: are whole seconds throughout (``iso_utc`` renders ``timespec="seconds"``)
-#: — so a second-wide half-open window *is* the point window at grammar
-#: resolution, without introducing float equality on observation timestamps.
+#: Width of the window a ``{label, timestamp}`` point entry desugars to (#355).
+#: One second is the grammar's native resolution — boundaries render at
+#: ``timespec="seconds"`` throughout — and keeps membership off float equality.
 _POINT_WINDOW_WIDTH = timedelta(seconds=1)
 
 
@@ -1065,21 +1063,13 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
     """The half-open UTC ``[start, end)`` of one explicit ``windows`` entry.
 
     Two forms, discriminated on keys and never mixed (issue #355): the range
-    form ``{label, start, end}``, and the point sugar ``{label, timestamp}``
-    which desugars to ``[t, t + 1s)`` (:data:`_POINT_WINDOW_WIDTH`). Desugaring
-    happens here, at the config layer, so everything downstream —
-    ``windows_intersecting``, the catalog time filters, ``windowed_cell_config``,
-    leaf naming, the sweep, the pyramid — only ever sees an ordinary explicit
-    range. Two acquisitions inside the same wall-clock second therefore share a
-    window; if that ever bites, the fix is an explicit ``width`` key on the
-    point form, not a guessed default.
-
-    Bounds are returned at the precision declared, but :func:`get_windowing`
+    ``{label, start, end}``, and the point sugar ``{label, timestamp}`` ->
+    ``[t, t + 1s)``. Bounds come back at declared precision; :func:`get_windowing`
     renders them through ``windows.iso_utc`` (``timespec="seconds"``), so a
-    SUB-SECOND ``timestamp`` normalizes to the whole second CONTAINING it —
-    both ends truncate by the same fraction, so ``t`` still falls inside its
-    own window, and disjointness is preserved (validated-disjoint points are
-    >= 1 s apart, hence so are their truncated bounds).
+    sub-second ``timestamp`` normalizes to the whole second containing it —
+    truncation is monotone, so it can neither move ``t`` out of its own window
+    nor manufacture an overlap. See ``docs/hive_layout.md`` for the same-second
+    collision and the silent-miss edge.
     """
     from zagg import windows as _windows
 
@@ -1091,11 +1081,11 @@ def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
                 f"entry declares exactly one of {{timestamp}} or {{start, end}} "
                 f"(got {entry!r})"
             )
+        # `width` is documented as the future escape hatch but does not exist
+        # yet: accepting it silently would hand back a 1 s window and a clean
+        # validate_config.
         extra = sorted(set(entry) - {"label", "timestamp"})
         if extra:
-            # `width` is named as the escape hatch for the same-second edge but
-            # does not exist yet — accepting it silently would hand back a
-            # one-second window and a clean validate_config (review finding).
             raise ValueError(
                 f"explicit point window entry carries unsupported key(s) "
                 f"{', '.join(extra)}; the point form is {{label, timestamp}} "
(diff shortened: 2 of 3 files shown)


src/zagg/config.py
```diff
@@ -1052,13 +1053,53 @@ def _validate_windowing(config: PipelineConfig) -> None:
...
+    ranged = sorted({"start", "end"} & set(entry))
+    if "timestamp" in entry:
+        if ranged:
+            raise ValueError(
+                f"explicit window entry carries both timestamp and {ranged[0]}; each "
+                f"entry declares exactly one of {{timestamp}} or {{start, end}} "
+                f"(got {entry!r})"
+            )
+        point = _windows.parse_utc(entry["timestamp"])
+        return point, point + _POINT_WINDOW_WIDTH
```
espg, line 1102: 🤖 *from Claude (review)*

**Sub-second timestamps are validated at full precision but normalized truncated — untested and undocumented.**

`parse_utc` keeps fractional seconds, so `timestamp: "2019-06-01T12:00:00.5Z"` is validated as `[12:00:00.5, 12:00:01.5)` — but `get_windowing` renders both ends through `windows.iso_utc`, which is `timespec="seconds"`, so what actually reaches the manifest and dispatch is `[12:00:00, 12:00:01)`.

The behavior is benign — both endpoints truncate by the same fraction, so `t` always lands inside the normalized window, and two points whose *validated* windows are disjoint have starts ≥ 1 s apart, hence floors ≥ 1 s apart, so normalized disjointness follows. But nothing in the diff pins it, and the docstring's `[t, t + 1s)` reads as exact. Add a test asserting a fractional timestamp normalizes to its containing whole second, and say so in the docstring — a reader should not have to derive the truncation from `iso_utc`.

---

src/zagg/config.py
```diff
@@ -2516,14 +2564,19 @@ def get_windowing(config: PipelineConfig) -> dict | None:
...
-                "end": _windows.iso_utc(_windows.parse_utc(w["end"])),
-            }
-            for w in block["windows"]
-        ]
+        # Point entries desugar HERE (issue #355): the normalized list is
+        # always the {label, start, end} range form, so no consumer — manifest
+        # temporal block included — ever sees a ``timestamp`` key.
+        declared = []
+        for w in block["windows"]:
+            start, end = _explicit_window_bounds(w)
```
espg, line 2612: 🤖 *from Claude (review)*

`get_windowing` calls `_explicit_window_bounds` bare, so on an **unvalidated** config a mixed `{timestamp, start}` entry surfaces as `explicit window entry carries both timestamp and start; …` without the `output.windowing.windows:` prefix that `_validate_windowing_windows` attaches — two spellings of the same error depending on the entry point.

The bar here is low: the pre-#355 code raised a bare `KeyError` on this same path, and every production caller runs after `validate_config`. Flagging rather than asserting it must change.

---

tests/test_hive_windows.py
```diff
@@ -318,6 +318,117 @@ def test_explicit_touching_ranges_allowed(self, cfg):
...
+                "end": "2019-06-01T12:00:01+00:00",
+            },
+            {
+                "label": "melt-2020",
+                "start": "2020-06-01T00:00:00+00:00",
+                "end": "2020-09-01T00:00:00+00:00",
+            },
+        ]
+
+    def test_explicit_entry_mixing_both_forms_rejected(self, cfg):
```
espg: 🤖 *from Claude (review)*

These two tests loop over cases inside the test body (`for over in (...)`), so a regression on the second case reports as a failure of the whole test with no indication of which entry shape broke. `tests/test_hive_windows.py` already uses `@pytest.mark.parametrize` for exactly this shape (`TestWindowedNodeInvariant.test_accepts` / `test_rejects`) — the surrounding-code convention.

Low severity; the raised entry appears in the assertion output either way.

---

docs/hive_layout.md
```diff
@@ -120,6 +121,19 @@ output:
   chronological order; explicit labels are opaque (`[0-9A-Za-z-]{1,32}`) and
   decode only through the declared list. `quarterly` is grammar-reserved but
   not implemented (validation rejects it).
+- **An explicit entry may be a point** ([issue #355](https://github.com/englacial/zagg/issues/355)):
+  `{label, timestamp}` is sugar for the second-wide half-open `[t, t + 1s)`,
```
espg, line 127: 🤖 *from Claude (review)*

The bullet claims "two point entries inside one second are rejected as overlapping", which the tests do pin — but it is stated only for *point* entries. The same desugaring means a point declared inside an existing range window is rejected too, which is the more likely user surprise (a scene timestamp that happens to fall inside a declared melt season). That case is mentioned further down in the validation paragraph; a forward reference here, or a one-clause mention, would keep the reader from having to find it.

Also worth confirming: the YAML example adds `{label: scene-a, timestamp: "2021-03-14T12:00:00Z"}` alongside two melt seasons — that instant sits outside both, so the example as written validates. Good; a doc example that would fail validation is a real trap.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial self-review of all three phases (`0401b25`, `d0c8406`, `7bd922f`) against CLAUDE.md and issue #355's acceptance criteria.

**Process deviation, stated up front:** the `Agent` / `Task` tool is **not available** in this environment, so the mandated fresh-context review subagent (and the separate fold subagent) could not be spawned. This review is a same-context self-review — the weaker fallback. Treat its independence accordingly.

**What holds up.** The desugar is a single helper used by both `_validate_windowing_windows` and `get_windowing`, so validation and normalization cannot disagree about what an entry means; the normalized list stays uniformly `{label, start, end}`, so `hive.build_manifest`'s `temporal.windows` block, `windows_intersecting`, `window_time_filters` and the sweep are genuinely untouched — which is what "graceful degradation by construction" was supposed to buy. No new dependency. No wire-format or `spec` change, so `docs/specification.md` and the conformance fixtures correctly stay out of the diff. Composition with existing validation is tested where it matters: reserved `all` token, duplicate labels, point-inside-range overlap, and the same-second collision the issue asked to *document* rather than solve.

**Findings** (4 inline):
1. `src/zagg/config.py` — sub-second timestamps are validated at full precision but normalized truncated to the second; benign but unpinned and undocumented. *The substantive one.*
2. `src/zagg/config.py` — `get_windowing` raises the mixed-form error without the `output.windowing.windows:` prefix on an unvalidated config.
3. `tests/test_hive_windows.py` — two in-body case loops where the file's convention is `@pytest.mark.parametrize`.
4. `docs/hive_layout.md` — the point-inside-a-range rejection deserves a mention in the new bullet, not only in the validation paragraph below.

**Standing item for review, not a diff finding:** `src/zagg/config.py` is now 2838 lines, against CLAUDE.md §4's ~1,200 guidance. It was 2785 lines on `main` at `083af16`, so this is pre-existing and the diff adds ~50 lines to an existing function pair. Splitting a 2.8k-line config module is not a `small-fix`; raising it here so the decision is on the record rather than silently deferred.

---

src/zagg/config.py
espg: 🤖 *from Claude*

Folded in `1f2542b`.

- `_explicit_window_bounds`'s docstring now states the truncation explicitly: bounds come back at the precision declared, but `get_windowing` renders them through `windows.iso_utc` (`timespec="seconds"`), so a sub-second `timestamp` normalizes to the whole second containing it — plus the reasoning that `t` still falls inside its own window and disjointness survives.
- New test `test_explicit_point_subsecond_normalizes_to_its_whole_second` pins `timestamp: "2019-06-01T12:00:00.75Z"` → `["2019-06-01T12:00:00+00:00", "2019-06-01T12:00:01+00:00")`.

No behavior change — the normalized output was already the containing second; it is now documented and covered.

---

src/zagg/config.py
espg: 🤖 *from Claude*

Left by design, no commit.

Wrapping the call in `get_windowing` would mean either a `try/except ValueError` re-raise there (duplicating the prefix logic in the one function that is supposed to be a pure normalizer) or moving the prefix into `_explicit_window_bounds` (which would then hardcode the `output.windowing.windows:` config path into a helper that only knows about a single entry). Both trade a nicer message on an unreachable path for coupling.

The path is genuinely unreachable in production: `validate_config` runs before every `get_windowing` caller, and pre-#355 the same path raised a bare `KeyError` for a malformed entry — so this is strictly an improvement on what was there. Standing for review if the two spellings are judged worth the coupling.

---

tests/test_hive_windows.py
espg: 🤖 *from Claude*

Folded in `c11daaa`.

Both tests now use `@pytest.mark.parametrize("entry", [...])`, matching `TestWindowedNodeInvariant.test_accepts` / `test_rejects` in this file. Each shape reports as its own case. While converting, each also gained the `end`-only arm it was missing (`{label, timestamp, end}` for the mixing test, `{label, end}` for the incomplete-range test), so both sides of the range form are covered rather than just `start`. 104 tests pass in `tests/test_hive_windows.py`.

---

docs/hive_layout.md
espg: 🤖 *from Claude*

Folded in `5b36af3`.

The bullet now names the point-inside-a-range case directly — "as is a point that lands inside an already-declared range window, which is an ordinary overlap (see the validation paragraph below)" — rather than leaving it to be discovered further down. The same edit folds in the sub-second note from the other thread.

Confirming the second half: the example instant `2021-03-14T12:00:00Z` sits outside both `melt-2019` and `melt-2020`, so the YAML block as printed passes `validate_config`. That was deliberate — an earlier draft used a July 2019 timestamp, which would have landed inside `melt-2019` and made the documented example a validation failure.

---

espg:
🤖 *from Claude*

Run status — all phases landed, all self-review findings resolved, CI green.

**Phases** (one commit each, per the body checklist)
- `0401b25` phase 1 — `_explicit_window_bounds` desugar + `_validate_windowing_windows` / `get_windowing` wiring
- `d0c8406` phase 2 — tests in `tests/test_hive_windows.py`
- `7bd922f` phase 3 — docs, `docs/hive_layout.md` windowing section

**Adversarial self-review + fold** — 4 inline findings, every thread answered:
- `1f2542b` folded (1): sub-second `timestamp` is validated at declared precision but rendered at whole-second precision by `iso_utc`; docstring now states it, new test pins `…T12:00:00.75Z` → `[12:00:00, 12:00:01)`.
- `5b36af3` folded (2): the docs bullet now names the point-inside-a-range overlap instead of leaving it to the validation paragraph below.
- `c11daaa` folded (3): the two entry-shape rejection tests converted to `@pytest.mark.parametrize` (this file's convention), each gaining the `end`-only arm it was missing.
- Left by design (4): `get_windowing` raises the mixed-form error without the `output.windowing.windows:` prefix. Adding it would either duplicate prefix logic in a pure normalizer or hardcode a config path into a single-entry helper, for a message on a path `validate_config` makes unreachable (pre-#355 it raised a bare `KeyError`). Rationale is on the thread; standing for review.

**Tests / lint**
- Local: full suite `3205 passed, 37 skipped` (9m14s); `tests/test_hive_windows.py` `104 passed` at HEAD.
- CI at `c11daaa`: `ruff`, `test (3.12)`, `test (3.13)`, `build`, `build / build-x86_64`, `build / build-arm64` all green. No inline comments from the ruff bot to answer.
- Two pre-existing local gate failures flagged (not fixed) under "Questions for review" — an `N818` in `src/zagg/registry.py` and a `ruff format` diff in `tests/data/benchmark/README.md`; both reproduce on `main` at `083af16`.

**Two process notes, both in the PR body:**
1. The `Agent` tool is unavailable in this environment, so the mandated fresh-context review and fold **subagents** could not be spawned — review and fold were done in-context, the weaker fallback. Reduced independence; worth a second pair of eyes.
2. This PR is deliberately left in **draft** despite CLAUDE.md §2's draft→ready flip being earned (phases done, findings folded, CI green). The routine driving this run overrides that rule and requires draft; recording the deviation rather than acting on it.

**Label state:** `implement`, with neither `waiting` nor `blocked` — awaiting review, no question pending.


---

src/zagg/config.py
```diff
@@ -1052,13 +1053,60 @@ def _validate_windowing(config: PipelineConfig) -> None:
...
+    ranged = sorted({"start", "end"} & set(entry))
+    if "timestamp" in entry:
+        if ranged:
+            raise ValueError(
+                f"explicit window entry carries both timestamp and {ranged[0]}; each "
+                f"entry declares exactly one of {{timestamp}} or {{start, end}} "
+                f"(got {entry!r})"
+            )
+        point = _windows.parse_utc(entry["timestamp"])
+        return point, point + _POINT_WINDOW_WIDTH
```
espg, line 1102: 🤖 *from Claude (review)*

**A `width` key on a point entry is silently ignored — and `width` is the exact key this PR tells readers to reach for.**

```python
    ranged = sorted({"start", "end"} & set(entry))
    if "timestamp" in entry:
        ...
        point = _windows.parse_utc(entry["timestamp"])
        return point, point + _POINT_WINDOW_WIDTH
```

Only `start`/`end` are checked for conflict; every other key falls through. Verified against this branch:

```
>>> _explicit_window_bounds({"label": "a", "timestamp": "2021-03-14T12:00:00Z", "width": 3600})
2021-03-14T12:00:00+00:00  2021-03-14T12:00:01+00:00     # width ignored, no error
```

That is a footgun specifically created by this diff, because three places in it name `width` as the escape hatch — the docstring right above ("if that ever bites, the fix is an explicit `width` key on the point form"), `docs/hive_layout.md` line 138, and the PR body. A user who hits the same-second collision, reads the docs, and writes `{label: scene-a, timestamp: "...", width: 3600}` gets a one-second window and a clean `validate_config`, then discovers at dispatch that 59 minutes of observations were filtered out by the `lt` predicate.

Cheapest fix that stays in scope: reject an unrecognized key on the point form, so the documented future knob fails loud until it exists — e.g. fold the check into the same `ranged` guard by comparing `set(entry) - {"label", "timestamp"}` against empty. Note that tightening this for the *range* form too would be a behavior change on existing configs (`{label, start, end, ...}` is equally lax today), so I would keep the reject scoped to the point form here and leave the general question standing.

---

src/zagg/config.py
```diff
@@ -1052,13 +1053,60 @@ def _validate_windowing(config: PipelineConfig) -> None:
...
+    own window, and disjointness is preserved (validated-disjoint points are
+    >= 1 s apart, hence so are their truncated bounds).
+    """
+    from zagg import windows as _windows
+
+    ranged = sorted({"start", "end"} & set(entry))
+    if "timestamp" in entry:
+        if ranged:
+            raise ValueError(
+                f"explicit window entry carries both timestamp and {ranged[0]}; each "
```

**The mixed-form error names only one of the two offending keys, and picks it alphabetically.**

```python
    ranged = sorted({"start", "end"} & set(entry))
    ...
            f"explicit window entry carries both timestamp and {ranged[0]}; each "
```

`sorted({"start", "end"})` is `["end", "start"]`, so `ranged[0]` is `"end"` whenever both are present. Verified on this branch:

```
>>> _explicit_window_bounds({"label": "w", "timestamp": "2019-06-01T12:00:00Z",
...                          "start": "2019-06-01", "end": "2019-09-01"})
ValueError: explicit window entry carries both timestamp and end; each entry declares
exactly one of {timestamp} or {start, end} (got {...})
```

The entry carries `start` *and* `end`, but the message names only `end`. A user who removes just `end` — the one key the error pointed at — lands on the *other* error path (`each explicit window must be a {label, start, end} or {label, timestamp} mapping`, since `{label, timestamp, start}` still trips the mixing guard first, and stripping `timestamp` instead leaves a half-range that fails the shape gate). Two round trips to fix one entry.

`', '.join(ranged)` makes it one round trip and costs nothing. The parametrized test `test_explicit_entry_mixing_both_forms_rejected` matches on `"exactly one of"`, so it does not currently distinguish the three shapes' messages — worth asserting the named key(s) per case while you are in there.

---

tests/test_hive_windows.py
```diff
@@ -1032,6 +1171,31 @@ def test_explicit_schedule_uses_declared_windows(self, cfg):
...
+        units = _windowed_units([(11, [hit, miss])], w, None)
+        assert [(w_["label"], [r["id"] for r in recs]) for _k, recs, w_ in units] == [
+            ("scene-a", ["g1"])
+        ]
+        window = units[0][2]
+        assert window["end"] - window["start"] == 1.0
+        unit_cfg, _windowing_out = windowed_cell_config(cfg, window)
+        ge, lt = unit_cfg.data_source["filters"][-2:]
+        assert (ge["op"], lt["op"]) == ("ge", "lt")
+        assert lt["value"] - ge["value"] == 1.0
```
espg, line 1316: 🤖 *from Claude (review)*

**The point form has no test on the raster path — which is the motivating use case and the one dispatch path with different mechanics.**

This test covers `runner._windowed_units` + `windowed_cell_config`, i.e. the point pipeline, where membership is ultimately decided by the worker's `ge`/`lt` observation filter. The raster path is a *separate* function with *separate* semantics — `runner._raster_windowed_units`:

```python
            labels = windows_intersecting(earliest, earliest, schedule, declared)
            if labels:
                by_window.setdefault(labels[0], []).extend(items)
```

Membership is decided at dispatch from the group's earliest STAC `datetime`, there is no worker-side filter, and — per that function's own docstring — "Explicit schedules DROP groups outside every declared window". So on raster, a point window that misses by a second produces **zero units and no error**. Verified against this branch's normalized output for the docs' own example point (`docs/hive_layout.md` line 115, `timestamp: "2021-03-14T12:00:00Z"`):

```
2021-03-14T12:00:00Z      -> ['scene-a']
2021-03-14T12:00:00.9Z    -> ['scene-a']
2021-03-14T12:00:01Z      -> []          # boundary exclusivity, correct
2021-03-14T12:00:31.024Z  -> []          # a real Sentinel-2 scene datetime
```

The first three lines are the correct behavior and are worth pinning — in particular `t+1s -> []` is the half-open boundary requirement and nothing in the diff asserts it at the membership level. The fourth is the hazard: the issue motivates the point form with "single acquisitions, scene timestamps", raster items routinely carry sub-second, non-round datetimes, and the failure mode is a silent empty run.

Suggested: one test alongside `test_raster_windowing_validates_on_hive` (line 460 already builds the raster fixture) that runs a point window through `_raster_windowed_units` and asserts (1) a scene at `t` and at `t+0.9s` land in the window, (2) a scene at exactly `t+1s` does not.

---

docs/hive_layout.md
```diff
@@ -120,6 +121,22 @@ output:
...
+  desugaring happens in `zagg.config.get_windowing`, so the manifest and every
+  downstream consumer only ever see an ordinary `{label, start, end}` window.
+  One second is the grammar's own resolution (boundaries are whole seconds
+  throughout), which also keeps membership off float equality on observation
+  timestamps (a sub-second `timestamp` normalizes to the whole second
+  containing it). *Consequence*: two acquisitions within the same wall-clock
+  second share a window, and two point entries inside one second are rejected
+  as overlapping — as is a point that lands inside an already-declared range
+  window, which is an ordinary overlap (see the validation paragraph below).
+  If that ever bites, the fix is an explicit `width` key on the point form, not
```
espg, line 140: 🤖 *from Claude (review)*

**The bullet documents the collision edge but not the miss edge, which is the one that fails silently.**

The paragraph covers what happens when two acquisitions land *inside* one second (they share a window; two point entries there reject as overlapping). It does not cover a point window that lands where nothing is — and the two paths behave very differently:

- Point pipeline: the window dispatches, the worker's `ge`/`lt` filter matches nothing, an empty leaf results.
- Raster: `runner._raster_windowed_units` "DROP[s] groups outside every declared window" — the window produces no work unit at all, and the run completes clean with no leaf and no warning.

Given that the stated motivation is scene timestamps, and STAC datetimes are rarely round seconds, this is the likely first-contact failure. Concretely, the example on line 115 of this file (`timestamp: "2021-03-14T12:00:00Z"`) matches nothing for a scene whose `datetime` is `2021-03-14T12:00:31.024Z` — verified against the normalized window on this branch.

One clause would do it: note that the point form matches only the whole second containing `t`, so the declared `timestamp` must be the acquisition's own instant (copied from the item), not a rounded-off approximation of it — and that a miss is silent.

---

src/zagg/config.py
```diff
@@ -1052,13 +1053,60 @@ def _validate_windowing(config: PipelineConfig) -> None:
...
+def _explicit_window_bounds(entry: dict) -> tuple[datetime, datetime]:
+    """The half-open UTC ``[start, end)`` of one explicit ``windows`` entry.
+
+    Two forms, discriminated on keys and never mixed (issue #355): the range
+    form ``{label, start, end}``, and the point sugar ``{label, timestamp}``
+    which desugars to ``[t, t + 1s)`` (:data:`_POINT_WINDOW_WIDTH`). Desugaring
+    happens here, at the config layer, so everything downstream —
+    ``windows_intersecting``, the catalog time filters, ``windowed_cell_config``,
+    leaf naming, the sweep, the pyramid — only ever sees an ordinary explicit
+    range. Two acquisitions inside the same wall-clock second therefore share a
```

**~51 lines of prose for ~14 lines of code, and the same three facts are restated five times — in a module already 2845 lines.**

The rationale for the one-second width, the same-second edge with its `width` escape hatch, and the "everything downstream is untouched" claim each appear in:

1. the `_POINT_WINDOW_WIDTH` comment (lines 1056-1060) — "One second is the window grammar's native resolution … so a second-wide half-open window *is* the point window at grammar resolution";
2. this docstring (lines 1065-1083) — the same width rationale, the full downstream consumer list, the same-second edge, the `width` escape hatch, the sub-second truncation;
3. `_validate_windowing_windows`'s docstring (the five added lines) — "a point landing inside another window's range is a genuine overlap and rejects";
4. `get_windowing`'s docstring (three added lines) *and* the inline comment at line 2574 (three more) — both saying the desugar happens there and no consumer sees a `timestamp` key;
5. `docs/hive_layout.md` lines 124-139 — all of the above again, for the audience that should actually be reading it.

CLAUDE.md §4 asks for terse, reviewable code; the narrative belongs in the docs page and the PR description (§2: "the PR description is where the substance lives"), and §3 keeps it out of commits for the same reason. The helper is a 12-line pure function whose behavior is fully described by "one of `{timestamp}` or `{start, end}`; a `timestamp` becomes `[t, t+1s)`, truncated to whole seconds by `iso_utc`" — four or five lines. Collapsing (1) into (2), dropping the duplicated downstream-consumer list from (2) and (3), and dropping either the docstring sentence or the inline comment in (4) removes roughly 30 of the ~57 lines this diff adds to `config.py`.

This is also the honest answer to question (4) in the PR body: the raise-rather-than-split call is right for a `small-fix`, but "this diff adds ~60 lines" understates how much of that was optional. See the summary comment for the module-size point itself.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the issue #355 desugar, at `c11daaa`. Independent of the in-context self-review already on this PR — none of the five findings below restate its four threads.

**Verdict: the core desugar is correct.** I tried to break it and could not. What I verified rather than took on trust:

- **Composition.** Every consumer reaches windows through `get_windowing`, never the raw block: `client.py:435`, `client_transport.py:493`, `notebook.py:73`, `runner.py:809/2787/3221`, and `hive.build_manifest` which receives the already-normalized dict (`hive.py:348-349`). The only raw `block["windows"]` reads in the tree are `config.py:1113` (the validator) and `config.py:2578` (the desugar itself). No `timestamp` key can reach `temporal.windows` — and the new test's exact dict equality on `get_windowing(cfg)["windows"]` pins that, so it is covered, not merely argued.
- **Boundary exclusivity.** `windows_intersecting` on the normalized point window returns `['scene-a']` at `t` and `t+0.9s`, `[]` at exactly `t+1s`. Half-open holds end to end; `window_time_filters` emits `lt end`, so the observation filter agrees.
- **Truncation is safe, for a stronger reason than the docstring gives.** Validation runs disjointness on untruncated bounds while `get_windowing` emits `iso_utc`-truncated ones, so the two *do* operate on different values. It is sound because `floor` is monotone: `end_a <= start_b` implies `floor(end_a) <= floor(start_b)`, so truncation can never manufacture an overlap — for mixed point/range pairs too, not just for points ">= 1 s apart". And `t` always lies in `[floor(t), floor(t)+1)`, so a point always falls inside its own normalized window.
- **Timezones.** `parse_utc` converts aware inputs to UTC and treats naive as UTC *before* the `+1s`; `timedelta` arithmetic on a UTC-aware datetime is absolute. There is no DST path to get wrong.
- **Validation reachability.** The shape gate `"label" in entry and ({"start","end"} <= set(entry) or "timestamp" in entry)` deliberately admits the mixed form so `_explicit_window_bounds` can raise the pointed error, and rejects partial ranges. The reserved `all` token, unique labels, and pairwise disjointness all run on desugared bounds. `not start < end` is genuinely unreachable for the point form, so the `entry['start']` lookup in its message cannot `KeyError`.
- **Store contract (CLAUDE.md §4).** Checked rather than trusted: `docs/specification.md` does not enumerate the config entry grammar, and the manifest's `temporal.windows` is a verbatim copy of the normalized list. No wire format, attrs grammar, or `spec` marker moves, so `tools/generate_spec_fixtures.py` / `tests/data/spec/` correctly need no update.

### Findings

Diff-scoped, ordered by weight:

1. **No raster-path coverage of the point form** (`tests/test_hive_windows.py`) — the motivating use case, a different dispatch function, and the only path where a missed window fails silently. Also the natural place to pin `t+1s -> []`.
2. **A `width` key on a point entry is silently ignored** (`src/zagg/config.py:1095`) — verified — while three places in this diff name `width` as the escape hatch to reach for.
3. **Documentation gap on the miss edge** (`docs/hive_layout.md:138`) — the collision edge is documented, the silent-empty-run edge is not.
4. **Mixed-form error names only one of the two offending keys, alphabetically** (`src/zagg/config.py:1090`) — verified; `', '.join(ranged)` fixes it.
5. **Prose duplication** (`src/zagg/config.py:1073`) — the same rationale five times, ~51 lines for ~14 lines of code.

### Standing for review, not for folding

- **Module size, question (4) in the PR body.** `config.py` goes 2785 → 2845, against §4's ~1,200 hard stop. Raising rather than splitting is the right call for a `small-fix` — splitting a 2.8k-line config module inside a bundled small-fix PR would be the worse convention violation. Two caveats for the record: §4 asks for the raise as an *issue comment* before continuing, and this one landed in the PR body after the fact; and per finding (5) a meaningful share of the +60 was optional prose. A standing issue to split `config.py` looks overdue, but opening one is a side-effecting action and not mine to take.
- **Pre-existing: sub-second *range* entries normalize to an empty window.** Verified on this branch — `{start: "2019-06-01T12:00:01.0Z", end: "2019-06-01T12:00:01.4Z"}` passes validation (`start < end` on untruncated bounds) but normalizes to `["2019-06-01T12:00:01+00:00", "2019-06-01T12:00:01+00:00")`, dispatching a `ge x` / `lt x` filter pair that matches nothing, silently. This is unchanged from `main` (the old code also rendered through `iso_utc(parse_utc(...))`), and the point form *cannot* produce it since its width is exactly 1 s by construction — so the PR's own sub-second reasoning is sound and this is out of scope per §4's "flag, don't fix" rule. Flagging it because this diff is what makes sub-second bounds a thing anyone will think about.
- **Question (1) in the PR body**, tests in `test_hive_windows.py` rather than `test_config.py`: agreed with the PR's call. `TestWindowingConfig` is where the windowing siblings live and `test_config.py` has no windowing coverage; the issue's file list reads as indicative, not binding.
- **Question (3)**, `_POINT_WINDOW_WIDTH` in `config.py`: agreed, it is config-layer sugar and `windows.py` never sees it. Leave it.

---

tests/test_hive_windows.py
espg: 🤖 *from Claude*

Folded in `fb3fd9d`.

New `TestWindowedUnits.test_raster_point_window_membership_is_half_open` drives `runner._raster_windowed_units` directly with the docs' point window and pins all four of the verified lines:

- a scene at exactly `t` and one at `t + 0.9s` land in `scene-a` (same window, both groups);
- a scene at exactly `t + 1s` yields `[]` — the half-open boundary, now asserted at the membership level rather than only through the worker's `lt` filter;
- a realistic `2021-03-14T12:00:31.024Z` yields `[]` too, which is the silent-miss hazard: `_raster_windowed_units` drops non-matching groups, so it is zero units and no error.

The class docstring picked up a clause noting it now covers the raster analog as well, since membership there is decided at dispatch rather than in the worker.


---

src/zagg/config.py
espg: 🤖 *from Claude*

Folded in `614c7bd`.

`_explicit_window_bounds` now rejects any key on a point entry other than `label`/`timestamp`, exactly as suggested — scoped to the point form, so the range form's laxness is unchanged and no existing config can start failing:

```python
        extra = sorted(set(entry) - {"label", "timestamp"})
        if extra:
            raise ValueError(
                f"explicit point window entry carries unsupported key(s) "
                f"{', '.join(extra)}; the point form is {{label, timestamp}} "
                f"(got {entry!r})"
            )
```

New parametrized `test_explicit_point_unsupported_key_rejected` covers `width: 3600` (the documented future knob) and a second unknown key. `docs/hive_layout.md` now says so in the same breath as naming `width`: "until it exists, any key on a point entry other than `label`/`timestamp` (`width` included) is rejected rather than ignored."

The general question — whether the range form should be tightened the same way — stays standing; that would be a behavior change on existing configs and is out of this PR's scope.


---

docs/hive_layout.md
espg: 🤖 *from Claude*

Folded in `f8db8a0`.

Two edits to `docs/hive_layout.md`:

- The YAML example's point entry is now `timestamp: "2021-03-14T12:00:31.024Z"` with an inline note that it is the acquisition's own instant copied from the item, not a rounded-off one. It still sits outside `melt-2019`/`melt-2020`, so the block as printed continues to pass `validate_config`.
- A new bullet documents the miss edge, with both paths spelled out — the point pipeline dispatches and the `ge`/`lt` filter matches nothing, leaving an empty leaf; raster drops the group at dispatch in `runner._raster_windowed_units` and produces no unit, leaf, or warning. Neither errors. The concrete case from this thread (`12:00:00Z` vs `12:00:31.024Z`) is the worked example.

The behavior itself is now pinned by the raster test in `fb3fd9d`, so the doc claim and the suite agree.


---

src/zagg/config.py

Folded in `1653707`.

`', '.join(ranged)` as suggested, so the three-key entry now reports `carries both timestamp and end, start` and one edit fixes it instead of two round trips.

`test_explicit_entry_mixing_both_forms_rejected` gained a `named` parameter and matches on `f"both timestamp and {named}; .*exactly one of"` — `start`, `end`, and `end, start` respectively — so the three shapes' messages are now distinguished rather than all collapsing onto the shared `exactly one of` tail.


---

src/zagg/config.py

Folded in `6cd5bdc`. Net `-28/+14` in `config.py`; the module is back to 2841 lines and the suite is unchanged, which is the guard that behavior did not move.

What went, following the five duplications as listed:

- (1) the `_POINT_WINDOW_WIDTH` comment collapsed from five lines to three, keeping only "one second is the grammar's native resolution" and the float-equality point.
- (2) the helper docstring dropped the downstream-consumer list, the same-second edge, and the `width` escape hatch — all of which live in `docs/hive_layout.md` — and now ends with a pointer there. The truncation paragraph stayed but shrank to the monotonicity argument, which is the stronger and shorter version of it.
- (3) `_validate_windowing_windows` no longer repeats the point-inside-a-range example; it just says the desugar runs first so every check below runs on ordinary bounds.
- (4) the inline comment in `get_windowing` is gone; the docstring sentence was kept as the single statement of where the desugar happens.

The rejection comment added by `614c7bd` is the one piece of new prose that stayed — it explains a raise, not a design.


---

espg:
🤖 *from Claude*

Fold pass on the fresh-context adversarial review ([review](https://github.com/englacial/zagg/pull/367#pullrequestreview-4843262305)) is complete. All five diff-scoped findings folded, one commit each, staging only the files that finding touches; every inline thread has a reply.

**Folded**

| # | Finding | Commit |
|---|---|---|
| (1) | No raster-path test for the point form; `t + 1s` boundary unpinned at the membership level | [`fb3fd9d`](https://github.com/englacial/zagg/commit/fb3fd9d) |
| (2) | A `width` (or any unknown) key on a point entry was silently ignored | [`614c7bd`](https://github.com/englacial/zagg/commit/614c7bd) |
| (3) | Docs covered the collision edge but not the silent-miss edge; the page's own example would miss a realistic scene datetime | [`f8db8a0`](https://github.com/englacial/zagg/commit/f8db8a0) |
| (4) | Mixed-form error named only one range key, picked alphabetically | [`1653707`](https://github.com/englacial/zagg/commit/1653707) |
| (5) | ~51 lines of prose for ~14 lines of code, the same rationale five times | [`6cd5bdc`](https://github.com/englacial/zagg/commit/6cd5bdc) |

Net effect on `config.py`: the diff against `main` is now +70/−14 (2841 lines total, down from 2845) despite (2) adding a new guard — (5) more than paid for it. The PR body is updated to match: the new rejection and the miss edge are described, the phases checklist carries this fold round, and the "How it was tested" list gained the two new tests.

**Standing for review, not folded**

- **Module size.** `config.py` remains far past §4's ~1,200 guidance. The raise-rather-than-split call was right for a `small-fix`, but opening a split issue is a side-effecting directive under §6 and not mine to take — and #349 already owns config splitting, so it likely belongs there rather than in a new issue. Recorded as question (4) in the body.
- **Pre-existing sub-second *range* bug.** A range entry narrower than one second normalizes to a degenerate empty window and dispatches a `ge x` / `lt x` pair that matches nothing, silently. Unchanged from `main`, and the point form cannot produce it (its width is exactly 1 s by construction), so it is out of this PR's scope per §4's flag-don't-fix rule. Recorded as question (6) in the body.
- **Whether the range form should also reject unknown keys.** (2) was deliberately scoped to the point form, since tightening the range form would change behavior on existing configs. Recorded as question (5).

**Local gate** — `uv sync --extra test`, then:

- `uv run pytest -v` → **3213 passed, 37 skipped** (9m51s), full suite;
- `uv run ruff check src tests` → the single pre-existing `N818` on `src/zagg/registry.py:64`;
- `uv run ruff format --check src tests` → the single pre-existing diff on `tests/data/benchmark/README.md`.

Both failures also reproduce on `main` at `083af16` and are untouched here (question (2) in the body). Every file this PR touches is clean under both commands.

**Label state: `implement` + `waiting`.** All phases are complete, all foldable findings are folded, and the remaining items need a decision that is not mine — so the ball is on the review side.

**Deviation, stated explicitly:** CLAUDE.md §2/§5 pair the `waiting` state with a draft → ready-for-review flip so the state is unambiguous. This routine run is instructed to leave PRs in draft, so the PR **stays a draft** — the more conservative of the two, and reversible either way. Flagging it rather than silently picking one.


---

espg:
🤖 *from Claude*

Both flagged-not-fixed items from "Questions for review" ((5) and (6)) are now fixed in this PR, at espg's direction (2026-08-05):

- **`0de3d0a`** — `reject unknown keys in range window entries`. The `{label, start, end}` form now refuses unsupported keys exactly like the point form, naming every offender and the allowed set. **Behavior change**: a range entry carrying dead/typo keys previously validated clean; it now fails `validate_config` (described under its own heading in the PR body).
- **`2e54462`** — `refuse window ranges that collapse at 1s granularity`. `_explicit_window_bounds` now validates the *rendered* bounds: a range whose bounds satisfy `start < end` at declared precision but truncate to the same whole second (e.g. `12:00:01.0Z → 12:00:01.4Z`) is refused with `explicit window range is empty after truncation to whole-second granularity …` instead of dispatching a `ge x` / `lt x` pair that silently matches nothing. Never widened to rescue it — the point form is the spelling for one-second intent. A sub-second-wide range straddling a second boundary (`12:00:01.9Z → 12:00:02.1Z`) stays valid; the point form is untouched.

Each commit carries its code, tests (`tests/test_hive_windows.py::TestWindowingConfig`), and the matching `docs/hive_layout.md` lines.

**Gates** (local, this worktree):

- `uv run pytest -v` — **3215 passed, 37 skipped, 2 failed**, both pre-existing and environment-dependent, neither reachable from this diff: `test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` (known env-dependent), and `test_client_transport.py::TestStatusPoller::test_invoke_fault_burns_an_attempt_and_retries` (`assert 1 == 2` on the retry count; fails identically at the pre-push tip `6cd5bdc`, so it predates these commits — likely a boto/env retry-behavior difference in this sandbox).
- `uv run ruff check src tests` — clean except the pre-existing `N818` at `src/zagg/registry.py:64` (already flagged in the body).
- `uv run ruff format --check src tests` — clean except the pre-existing `tests/data/benchmark/README.md` diff (already flagged).

**Quick audit** — other validate-at-full-precision → render-truncated sites (flagged only, per scope):

- `src/zagg/config.py:1040` vs `src/zagg/config.py:2624` — `output.windowing.epoch` is validated with `parse_utc` at declared precision but canonicalized through `iso_utc` (`timespec="seconds"`), so a sub-second epoch silently drops its fraction and shifts the dataset-unit↔UTC conversion by up to 1 s.
- Render-only narrowing (no validation mismatch, noting for completeness): `src/zagg/windows.py:302` (`union_time_range`) and `src/zagg/windows.py:313` (`iso_time_range`) truncate the range *end* downward, under-reporting recorded coverage by <1 s; `src/zagg/processing/raster.py:503` follows the same convention deliberately.

PR body updated: (5) and (6) marked resolved with provenance, the behavior change described under its own heading, and both items added to the bundle checklist. Labels unchanged (`implement` + `waiting`); the PR remains ready for espg review.

docs/hive_layout.md
```diff
@@ -149,7 +180,13 @@ output:
 Validation: `output.windowing` requires the hive layout on a healpix grid;
 `time_field` must be a declared `data_source` column (the worker can only
 filter what it reads); explicit windows must be well-formed (frozen label
-grammar, `start < end`, unique labels, disjoint ranges). On the raster path
+grammar, `start < end`, unique labels, disjoint ranges — point entries are
+desugared first, so a point landing inside another window's range is a genuine
+overlap and is rejected). Because boundaries render at whole-second
+granularity, a range narrower than the second containing it (say
+`12:00:01.0Z` → `12:00:01.4Z`) would dispatch as an empty window; it is
+refused rather than widened — the point form is the spelling for one-second
```

**The refusal closes the loud half of the sub-second problem and leaves the quiet half open — and this paragraph reads as though it closed both.**

The check fires only when `trunc(start) == trunc(end)`. Any range whose `end` has a nonzero fractional second but lands in a *different* second than `start` renders to something other than what was declared, silently. Two cases I ran against the branch:

```
declared [12:00:00.0Z, 12:00:01.5Z)  ->  rendered [12:00:00+00:00, 12:00:01+00:00)
declared [12:00:01.9Z, 12:00:02.1Z)  ->  rendered [12:00:01+00:00, 12:00:02+00:00)
```

The first **drops declared coverage**: observations in `[12:00:01.0, 12:00:01.5)` that the config explicitly put *inside* the window get dispatched outside it — no error, no warning. The second gains 0.9 s of coverage the config never declared at the head and drops 0.1 s at the tail. That is the same "the window does not mean what it says" failure `2e54462` set out to close, in the form that is *harder* to notice than the one it did close — an empty leaf is visible in the output; a correctly-sized window holding the wrong second of data is not.

I am explicitly **not** asking to widen or reject these (the ruling was collapse-only, and monotonicity means these cases are still mutually disjoint and internally consistent). The finding is that the documentation now overstates the guarantee:

- This sentence — "a range narrower than the second containing it ... is refused rather than widened" — is literally scoped to a range contained in one second, but a reader hitting it while writing sub-second bounds takes it as "zagg refuses sub-second ranges that would misbehave," and the straddling case is neither refused nor documented.
- The point form's normalization *is* documented ("a sub-second `timestamp` normalizes to the whole second containing it", line 137). The range form's truncation is documented nowhere — not in this paragraph, and not in `_explicit_window_bounds`'s docstring, which mentions truncation only for `timestamp`.
- `tests/test_hive_windows.py:455` freezes the straddling case as `..._allowed` and asserts the retimed output, but its comment says only "renders to a real one-second window and stays valid" — it never names the fact that the asserted `start` is 0.9 s earlier than the declared one.

Suggested fix, no behavior change: one sentence here (and a clause in the docstring) saying range bounds truncate to whole seconds exactly as the point form's do, so a sub-second `end` silently narrows the window — plus a test comment that reads as documenting the retiming rather than blessing it as a no-op.

src/zagg/config.py
```diff
@@ -1052,13 +1053,83 @@ def _validate_windowing(config: PipelineConfig) -> None:
...
+    extra = sorted(set(entry) - {"label", "start", "end"})
+    if extra:
+        raise ValueError(
+            f"explicit range window entry carries unsupported key(s) "
+            f"{', '.join(extra)}; the range form is {{label, start, end}} "
+            f"(got {entry!r})"
+        )
+    start = _windows.parse_utc(entry["start"])
+    end = _windows.parse_utc(entry["end"])
+    if start < end and start.replace(microsecond=0) == end.replace(microsecond=0):
```

**The collapse check reimplements `iso_utc`'s truncation instead of using it, while the docstring claims it is done "on the rendered values".**

`start.replace(microsecond=0) == end.replace(microsecond=0)` is a second copy of the `timespec="seconds"` truncation `windows.iso_utc` performs — equivalent today (both bounds are already UTC-aware out of `parse_utc`, so the `astimezone` is a no-op and `isoformat(timespec="seconds")` truncates identically), so there is no live misbehavior. The defect is that the two are now coupled by convention only, in exactly the place the docstring above asserts they are the same thing ("refused here, on the rendered values").

If `iso_utc` ever changes what it emits — a `timespec` change, or any normalization added for a future encoding — this check keeps passing ranges that render empty, and nothing in the suite would catch it: the collapse test asserts on the raised message and the straddling test asserts on `get_windowing`, but no test exercises the predicate and the renderer against each other.

`_windows.iso_utc(start) == _windows.iso_utc(end)` is the same one-liner, cannot drift, and matches the docstring's claim. The error message two lines down already calls `iso_utc(start)`, so the function is in hand.

src/zagg/config.py
```diff
@@ -1052,13 +1053,83 @@ def _validate_windowing(config: PipelineConfig) -> None:
...
+        extra = sorted(set(entry) - {"label", "timestamp"})
+        if extra:
+            raise ValueError(
+                f"explicit point window entry carries unsupported key(s) "
+                f"{', '.join(extra)}; the point form is {{label, timestamp}} "
+                f"(got {entry!r})"
+            )
+        point = _windows.parse_utc(entry["timestamp"])
+        return point, point + _POINT_WINDOW_WIDTH
+    extra = sorted(set(entry) - {"label", "start", "end"})
```

**A non-string extra key raises `TypeError`, not the `ValueError` this branch is written to raise, and it escapes `validate_config` uncaught.**

Both `sorted()` and `', '.join()` here assume every key is a string. YAML 1.1 does not guarantee that: PyYAML resolves bare `on`/`off`/`yes`/`no`/`y`/`n` to booleans and bare numerals to ints, so a plausible typo in a windows entry yields a non-string key. Against this branch:

```python
>>> d = yaml.safe_load('{label: w, start: "2019-06-01", end: "2019-09-01", on: 1, 3: x}')
>>> list(d)
['label', 'start', 'end', True, 3]
>>> _explicit_window_bounds(d)
TypeError: sequence item 0: expected str instance, bool found
```

and with a mixed set it fails one line earlier instead:

```python
>>> sorted({'note', 3})
TypeError: '<' not supported between instances of 'int' and 'str'
```

`_validate_windowing_windows` wraps this call in `except ValueError`, so the `TypeError` sails straight out of `validate_config` — the user loses not just the "unsupported key(s)" guidance but even the `output.windowing.windows:` prefix that would tell them which block is at fault. The failure mode this commit exists to prevent (a dead key riding along) is traded for a different opaque one.

`sorted(extra, key=repr)` plus `', '.join(map(repr, extra))` fixes both halves. The same two calls appear in the point-form check a few lines up (pre-existing, outside this review's scope) — worth fixing together, since this commit copied the pattern.

espg (commented):
🤖 *from Claude (review)*

Adversarial review of the two newest commits only (`0de3d0a` unknown-key rejection on the range form, `2e54462` rendered-bounds collapse refusal). The earlier #355 point-form work is out of scope and not relitigated.

**Verdict: the mechanism is sound.** I checked the invariant the collapse rule has to preserve and it holds: `iso_utc` truncation is monotone, so for two declared-disjoint entries (`e1 <= s2`) the rendered pair stays disjoint, a rendered range can never equal another rendered range or a rendered point window, and `trunc(t + 1s) == trunc(t) + 1s` keeps the desugared point form exactly one second wide. So the new check is *complete* for emptiness — `trunc(start) == trunc(end)` is precisely the empty-render condition, and nothing about it can manufacture an overlap the disjointness pass would miss. Check ordering is right too: the collapse branch is guarded by `start < end`, so a reversed range still falls through to the `not half-open` message rather than getting the misleading truncation one, and a mixed `{start, end, timestamp}` entry still errors on the mixed-form branch first. Timezone handling is fine — `parse_utc` normalizes before the comparison, so `13:00:01.0+01:00` / `14:00:01.4+02:00` is correctly refused as the same rendered second. 111/111 in `tests/test_hive_windows.py` pass locally.

Three findings below, none of which contest the ruled semantics.

1. `docs/hive_layout.md:188` (with `src/zagg/config.py:1109`, `tests/test_hive_windows.py:455`) — sub-second `end` truncation silently *narrows* a range, and only the total-collapse subcase is refused; the doc states the hazard as fully closed, and neither doc nor docstring ever says range bounds truncate at all.
2. `src/zagg/config.py:1109` — the check reimplements `iso_utc`'s truncation instead of comparing the rendered values the docstring says it compares; drift-prone, and no test pins the two together.
3. `src/zagg/config.py:1100` — non-string extra keys raise `TypeError`, not the intended `ValueError`, and it escapes `validate_config` uncaught.

docs/hive_layout.md

Folded in `3d0576a` — docs + docstring + test comment, no behavior change.

`docs/hive_layout.md` now states the truncation rule directly instead of implying a guarantee it does not make:

> Range bounds render at whole-second granularity exactly as a point `timestamp` does — **each bound truncates to the second containing it** — so a fractional bound silently retimes that edge of the window: `[12:00:00.0Z, 12:00:01.5Z)` dispatches as `[12:00:00, 12:00:01)`, dropping the declared half second of coverage at the tail. A range is never widened to recover the truncated fraction. The one case refused outright is total collapse […] A sub-second range that *straddles* a second boundary (`12:00:01.9Z` → `12:00:02.1Z`) is still valid; it renders to the one-second window its truncated bounds describe.

`_explicit_window_bounds`'s docstring gained the matching clause ("*Range* bounds truncate identically, each to the second containing it, so a fractional `start`/`end` silently retimes that edge of the window; a range is never widened […] The one refusal is total collapse"), and the inline comment at the check now says explicitly that only the total collapse is refused.

`test_explicit_subsecond_range_spanning_a_second_boundary_allowed` names the retiming rather than blessing it:

```
# A sub-second-wide range that straddles a second boundary does NOT
# collapse — but it is RETIMED, not preserved: each bound truncates to
# the second containing it, so the declared [12:00:01.9, 12:00:02.1)
# renders as [12:00:01, 12:00:02) — 0.9 s earlier at the head, 0.1 s
# short at the tail.
```

Semantics unchanged per the ruling: collapse-only refusal, no widening, straddling stays valid.

src/zagg/config.py

Folded in `f06a1e7`. The predicate is now the renderer:

```python
if start < end and _windows.iso_utc(start) == _windows.iso_utc(end):
```

so there is nothing left to drift — a future `timespec` or normalization change in `iso_utc` moves the check with it. The comment below records why (`The predicate above calls the renderer rather than repeating its truncation, so the two cannot drift.`).

Test side, `test_explicit_subsecond_range_collapses_and_is_rejected` now asserts the rendered second the predicate compared, not just the phrase:

```python
match=r"empty after truncation to whole-second .*"
r"both render as '2019-06-01T12:00:01\+00:00'",
```

which ties the refusal to the exact string `get_windowing` would emit for those bounds. Behavior identical.

src/zagg/config.py

Folded in `a34b7cf`, both copies — the range-form check and the point-form check it was copied from (same PR, same defect, so in scope):

```python
extra = sorted(set(entry) - {"label", "start", "end"}, key=repr)
...
f"{', '.join(map(repr, extra))}; the range form is "
```

`key=repr` makes the sort total across mixed key types and deterministic; `map(repr, ...)` makes the join type-agnostic and additionally shows the reader that the key came back coerced (`True`, not `on`), which is the actual YAML 1.1 clue. A comment at the range check records the reason, including that the `TypeError` would otherwise escape `_validate_windowing_windows`' `except ValueError` and lose the `output.windowing.windows:` prefix.

Tests: a new param on `test_explicit_range_unsupported_key_rejected` with `{"note": "x", True: 1, 3: "y"}` expecting `'note', 3, True`, and a new `test_explicit_point_non_string_key_named` on the point form that asserts the full path including the prefix:

```python
match=r"output\.windowing\.windows: .*unsupported key\(s\) 3, True; the point form",
```

Note the message text for string keys changed with this (`width` → `'width'`); the existing range params were updated to match.

PR #368, opened 2026-08-03
Benchmark: record the measurement before asserting object counts

### LINKED ISSUE #365: Benchmark: record the measurement before asserting object counts (a tripwire should not discard data)
🤖 *from Claude* (filed at espg's direction, 2026-07-31)

## Problem

In `.github/workflows/lambda-benchmark.yml`'s `merge-publish` job, the object-count assertion runs **inside** the "Run benchmark" step (`.github/actions/run-benchmark`), and that step has no `continue-on-error`. So a count mismatch exits 1 and aborts the job **before** the three steps that follow ever run:

1. check out the `benchmarks` data branch,
2. `update_series.py --series _series/series.parquet --records metrics.json`,
3. `plot_series.py` + the commit/push.

The benchmark itself has already completed at that point — the Lambda dispatch happened and the measurement was printed:

```
[tdigest_healpix_o9_hive] obs=3329701 runtime_s=96.88 cost/shard=$0.0052 cost/100km2=$0.0032 max_memory_mb=541.8
```

…and is then discarded. The run happened; the data point is thrown away, leaving a hole in `series.parquet` for that commit (e.g. `5e35e2c`, the englacial/zagg#351 merge).

## Why this is worth fixing beyond the current cause

Issue #362 fixes the *cause* of the recent mismatches (accumulating root telemetry in a shared store). This issue is about the *ordering*, which is independent and outlives that fix:

- **The holes are sporadic**, because the mismatch is timing-dependent — so the series quietly thins out rather than obviously breaking. Silent data loss reads exactly like "no run happened."
- **The discarded points are the interesting ones.** A run whose store carried unexpected objects is precisely a run worth having in the series next to the anomaly; instead its metrics are the ones deleted.
- **Every future tripwire inherits the behavior.** The overview-sidecar classifier gap from PR #356 would have been the next one to cost data points.

An invariant tripwire and a data-collection step should not be ordered so the tripwire destroys the collection.

## Proposal

Record first, then assert: let the series append/render/push run regardless of the count verdict, and fail the job afterward on the mismatch. Mechanically that is either `continue-on-error` on the measuring step with an explicit failing step after the append, or hoisting the assertion out of the composite action into its own step placed after the push. Either way the alarm still fires (the job still goes red) — it just stops taking the data with it. Whichever shape is chosen should also make the anomaly visible *in* the recorded row (a flag/column) rather than only in the log, so a thinned series can be explained later.

**Scope note:** this touches `.github/workflows/lambda-benchmark.yml` (and possibly `.github/actions/run-benchmark/action.yml`), which CLAUDE.md §1 puts off-limits without the issue naming them — this issue names them, and espg directed the filing. Backfilling the already-lost points from job logs is possible but deliberately out of scope here.

Refs #362, #240 (object-count model lineage), PR #356.

espg:
Closes #365

## What this changes

The `merge-publish` job runs the object-count tripwire **inside** the "Run benchmark" step, so a mismatch exits 1 and aborts the job before the three steps that retain the point (checkout `benchmarks`, `update_series.py`, `plot_series.py` + push). The measurement exists — `metrics.json` is written *before* the tripwire is evaluated — and is thrown away.

The fix is ordering, not tripwire semantics: **record first, then let the red step stand.** Two halves:

**(1) The retention boundary decides what may be kept — landed.** Appending unconditionally would retain the one row the *other* tripwire exists to reject: a silent OOM (`obs=0` / null peak memory, issue #145) plots as a real dip to zero. So `update_series.main` now filters non-merge **and** empty-metrics records, reporting every drop. The predicate is one definition — `bench_metrics.has_empty_metrics` — shared with `run_benchmark`'s `--fail-on-empty`, so the appender rejects exactly the rows the tripwire names:

```python
retained, dropped = [], []
for r in records:
    if r.get("event") != "merge":
        dropped.append((r.get("target"), "non-merge run"))
    elif bench_metrics.has_empty_metrics(r):
        dropped.append((r.get("target"), "empty metrics (obs=0 / no peak memory)"))
    else:
        retained.append(r)
```

An object-count mismatch is deliberately **not** in that list — that run measured fine and is precisely the point worth keeping next to the anomaly. And the anomaly rides the row: `objects_mismatch` moves from JSON-only to a retained series column (appended last, per the stable-schema rule), so a mismatched point is explicable from `series.parquet` alone rather than only from a job log that ages out. Null on clean runs and on every row recorded before the column existed.

**(2) The retain steps stop being skipped — written, but NOT in this branch (see below).** Each of the three follow-on steps in `.github/workflows/lambda-benchmark.yml` gets `!cancelled() && hashFiles('metrics.json') != ''`. The job still goes red (the measuring step is untouched, no `continue-on-error`) — it just stops taking the data with it. Gating on the *artifact* rather than on the step's status is what makes this general: **any** tripwire that fires after `metrics.json` is written inherits the behavior, which is the property issue #365 asks for ("every future tripwire inherits the behavior" — the PR #356 overview-sidecar classifier gap would have been next). A run that died *before* writing metrics (login/dispatch failure) matches nothing and skips cleanly, mirroring the `comment.md` existence check in `lambda-benchmark-command.yml`.

> **This half is blocked on a push permission, not on a decision about the design.** The push of the workflow file was rejected: `refusing to allow an OAuth App to create or update workflow '.github/workflows/lambda-benchmark.yml' without 'workflow' scope`. The diff and its tests are written and green locally; they are posted in full on the thread below rather than pushed through another credential path. Issue #365 names the file, so the *policy* permits the edit — the token does not. See the thread for the options.

Not touched, deliberately: the composite action `.github/actions/run-benchmark/action.yml` (the tripwire keeps its current semantics for the PR/on-demand leg), the release/full-AOI leg (record-only there — it never fails on a mismatch, so it has no ordering problem), and backfilling the already-lost points (out of scope per the issue).

## Phases

- [x] Phase 1 — retention boundary + the recorded verdict: `has_empty_metrics` as the shared predicate, `update_series` filter, `objects_mismatch` as a series column, tests (`c4830e3`). Adversarial-review findings folded in `1a1cc96`, `80f6d4e`, `58e0d96`, `220498e`.
- [ ] Phase 2 — workflow ordering: `!cancelled() && hashFiles('metrics.json') != ''` on the three retain steps in `lambda-benchmark.yml`, plus structural tests over the parsed workflow. **Written and green locally; blocked on the `workflow` push scope** — full patch on the thread.
- [ ] Phase 3 — docs: `docs/deployment/benchmark.md` on what a red merge benchmark means for the series. Held until phase 2 lands, so the page never describes a half-applied behavior.

## How it was tested

- `uv run pytest -v` on the pushed branch: **3207 passed, 37 skipped** (full suite, ~9m50s).
- New `tests/test_benchmark_recording.py`: a mismatched record survives `update_series.main` **with** its verdict in the parquet; `obs=0` / null-memory / non-merge records are dropped and reported; a dropped record cannot evict a retained one through the `(commit, target)` dedup; a mixed batch keeps the good rows.
- Added while folding the adversarial review: `has_empty_metrics` treats `NaN` as missing on either field, including end-to-end through a `metrics.json` round-trip (`json.dumps` emits the bare `NaN` token and `json.loads` reads it back); appending onto a series that **predates** the `objects_mismatch` column — the real shape of `series.parquet` on the `benchmarks` branch — keeps column order, row order and the integer dtypes, with the pre-existing row reading null; and a *partially* dropped batch, which takes the `pd.concat` path rather than the empty-batch early return, likewise leaves the retained history's integer dtypes alone.
- The phase-2 half is CI-shaped and cannot be tested by execution — the job it lives in costs a live Lambda dispatch. Its tests assert it *structurally* over the parsed YAML: the retain steps come after the measuring step and in order, each carries both `!cancelled()` and the `hashFiles('metrics.json')` gate, and the measuring step is **not** `continue-on-error` (i.e. the alarm still fires). That is a regression pin on the ordering, not a claim that the job was executed. Precedent for parsing this workflow in tests: `tests/test_deploy_lambda.py`'s `--variants` drift guard.
- New tests live in their own module rather than in `tests/test_benchmark.py` (already 2,468 lines) — see the questions below.

Pre-existing, unrelated, **not** fixed here (both on files this PR does not touch, and both from local ruff 0.16.1 vs the `.pre-commit-config.yaml` pin at v0.14.10):

- `ruff check src tests` → `N818` on `src/zagg/registry.py:64` (`UnknownCapability`).
- `ruff format --check src tests` → `tests/data/benchmark/README.md` (0.16 formats python blocks inside markdown).

`ruff check` / `ruff format --check` are clean on every file this PR touches.

## Questions for review

1. **How should the workflow half land?** It is blocked on the `workflow` push scope, not on anything about the change. Options are on the thread — deliberately no attempt was made to route the same edit through a different credential.
2. **Gate on `hashFiles('metrics.json')`, or on an explicit run-benchmark output?** The artifact gate is why this generalizes to future tripwires, and it needs no change to the composite action — but it is an implicit contract ("metrics.json existing means the records are worth appending"). The explicit alternative is a step output from the composite action (or a distinct exit code for "tripwire fired after the data was written"), which is louder but special-cases each tripwire and leans on composite-action outputs surviving a failed step.
3. **Should the release/full-AOI leg get the `objects_mismatch` column too?** It is record-only there (never fails), so it has no ordering problem — but the column would make the two series read the same way. Left out as scope.
4. **New test module vs. growing `tests/test_benchmark.py`.** Added `tests/test_benchmark_recording.py` rather than pushing a 2,468-line module further past the ~1,200-line guidance.

### DIFF
--- a/.github/scripts/bench_metrics.py
+++ b/.github/scripts/bench_metrics.py
@@ -279,11 +279,12 @@ def has_empty_metrics(record: dict) -> bool:
     is ``False``, so an unguarded predicate would retain that row as a real point.
 
     ONE definition, shared by ``run_benchmark``'s ``--fail-on-empty`` tripwire and
-    ``update_series``'s retention filter (issue #365). The two used to be the same
-    predicate by construction: an empty run aborted the job, so the append step
-    never ran. Now that the append runs regardless of the tripwire verdict, the
-    appender has to recognise exactly the junk row the tripwire names -- otherwise
-    record-before-assert would retain the obs=0 point it was meant to reject.
+    ``update_series``'s retention filter (issue #365). The two are still the same
+    predicate by construction today: an empty run aborts the job, so the append
+    step never runs. Once the append runs regardless of the tripwire verdict (the
+    workflow gate, a later phase), the appender has to recognise exactly the junk
+    row the tripwire names -- otherwise record-before-assert would retain the
+    obs=0 point it was meant to reject.
     """
     max_memory_mb = record.get("max_memory_mb")
     total_obs = record.get("total_obs")
--- a/.github/scripts/bench_metrics.py
+++ b/.github/scripts/bench_metrics.py
@@ -240,6 +240,16 @@ def _runtime_s(summary: dict) -> float | None:
     return None
 
 
+def _is_nan(value) -> bool:
+    """Is this the float NaN a missing number degrades to off parquet or JSON?
+
+    A legacy parquet row reads a missing number back as NaN rather than None
+    after the reindex, and ``json.dumps``/``json.loads`` round-trip the bare
+    ``NaN`` token, so a metrics JSON can carry one too.
+    """
+    return isinstance(value, float) and math.isnan(value)
+
+
 def memory_pct_of_cap(max_memory_mb, memory_gb) -> float | None:
     """Fraction of the Lambda memory cap a shard peaked at (issue #120).
 
@@ -251,8 +261,7 @@ def memory_pct_of_cap(max_memory_mb, memory_gb) -> float | None:
     """
     if max_memory_mb is None or memory_gb is None:
         return None
-    # A legacy parquet row reads back as NaN, not None, after the reindex.
-    if isinstance(max_memory_mb, float) and math.isnan(max_memory_mb):
+    if _is_nan(max_memory_mb):
         return None
     cap_mb = memory_gb * 1024.0
     if cap_mb <= 0:
@@ -265,6 +274,9 @@ def has_empty_metrics(record: dict) -> bool:
 
     ``total_obs`` 0/missing or a null ``max_memory_mb`` is the signature of a
     silent OOM -- there is no measurement in the row, only the shape of one.
+    NaN counts as missing on either field, same as in :func:`memory_pct_of_cap`:
+    a metrics JSON round-trips the bare ``NaN`` token, and ``not float("nan")``
+    is ``False``, so an unguarded predicate would retain that row as a real point.
 
     ONE definition, shared by ``run_benchmark``'s ``--fail-on-empty`` tripwire and
     ``update_series``'s retention filter (issue #365). The two used to be the same
@@ -273,7 +285,9 @@ def has_empty_metrics(record: dict) -> bool:
     appender has to recognise exactly the junk row the tripwire names -- otherwise
     record-before-assert would retain the obs=0 point it was meant to reject.
     """
-    return not record.get("total_obs") or record.get("max_memory_mb") is None
+    max_memory_mb = record.get("max_memory_mb")
+    total_obs = record.get("total_obs")
+    return not total_obs or _is_nan(total_obs) or max_memory_mb is None or _is_nan(max_memory_mb)
 
 
 def build_record(
--- a/.github/scripts/bench_metrics.py
+++ b/.github/scripts/bench_metrics.py
@@ -116,9 +116,8 @@ RECORD_COLUMNS = [
     # count is data-dependent, i.e. not exact) -- the sharded-write-bypass
     # tripwire (issue #215). Null on rows recorded before the metric existed.
     # The per-run record additionally carries the JSON-only
-    # ``objects_per_shard`` / ``objects_telemetry`` / ``objects_mismatch``
-    # keys, which the reindex to these columns deliberately drops from the
-    # parquet series.
+    # ``objects_per_shard`` / ``objects_telemetry`` keys, which the reindex to
+    # these columns deliberately drops from the parquet series.
     "objects_total",
     "objects_expected",
     # Store-layout axis (issue #240 phase 4): "flat" | "hive". Null on rows
@@ -161,6 +160,14 @@ RECORD_COLUMNS = [
     # earlier rows. Appended last per the stable-schema rule, so it sits away
     # from the objects columns it belongs with.
     "objects_write_path",
+    # The object-count tripwire's verdict (issue #365): the mismatch description
+    # when the audited count deviated from the expectation, null when it agreed.
+    # Retained rather than left JSON-only because the run that trips the tripwire
+    # is now RECORDED before the job goes red -- so the row has to say why it is
+    # anomalous, or a reader hits an unexplained step in the object counts with
+    # nothing in the series to attribute it to. Null on rows recorded before the
+    # column existed (and on every clean run since).
+    "objects_mismatch",
 ]
 
 # summary["worker_phase_max"] key -> series column (issues #250/#256). A phase
@@ -253,6 +260,22 @@ def memory_pct_of_cap(max_memory_mb, memory_gb) -> float | None:
     return max_memory_mb / cap_mb
 
 
+def has_empty_metrics(record: dict) -> bool:
+    """Did this target come back without a usable measurement (issue #145)?
+
+    ``total_obs`` 0/missing or a null ``max_memory_mb`` is the signature of a
+    silent OOM -- there is no measurement in the row, only the shape of one.
+
+    ONE definition, shared by ``run_benchmark``'s ``--fail-on-empty`` tripwire and
+    ``update_series``'s retention filter (issue #365). The two used to be the same
+    predicate by construction: an empty run aborted the job, so the append step
+    never ran. Now that the append runs regardless of the tripwire verdict, the
+    appender has to recognise exactly the junk row the tripwire names -- otherwise
+    record-before-assert would retain the obs=0 point it was meant to reject.
+    """
+    return not record.get("total_obs") or record.get("max_memory_mb") is None
+
+
 def build_record(
     summary: dict,
     *,
@@ -329,15 +352,15 @@ def build_record(
     wpm = summary.get("worker_phase_max") or {}
     for src, col in PHASE_MAP.items():
         record[col] = wpm.get(src)
-    # Store object counts (issue #240). The two scalar columns are retained in
-    # the parquet series; per_shard/mismatch ride the metrics.json record only
-    # (update_series's reindex drops them).
+    # Store object counts (issue #240). The scalar columns and the mismatch
+    # verdict (issue #365) are retained in the parquet series; per_shard rides
+    # the metrics.json record only (update_series's reindex drops it).
     o = objects or {}
     record["objects_total"] = o.get("objects_total")
     record["objects_write_path"] = o.get("objects_write_path")
     record["objects_expected"] = o.get("objects_expected")
     record["objects_per_shard"] = o.get("objects_per_shard")
-    # Per-run root telemetry (issue #362): JSON-only like per_shard/mismatch —
+    # Per-run root telemetry (issue #362): JSON-only like per_shard --
     # reported so an accumulating shared store is legible, never audited.
     record["objects_telemetry"] = o.get("objects_telemetry")
     record["objects_mismatch"] = o.get("objects_mismatch")
--- a/.github/scripts/full_aoi_series.py
+++ b/.github/scripts/full_aoi_series.py
@@ -80,7 +80,9 @@ FULL_AOI_COLUMNS = [
     # measured store object total and the config-derived expectation (null
     # when the layout's count is data-dependent). The run record's
     # ``objects_mismatch`` description stays JSON-only (dropped by the
-    # reindex), like the per-merge series.
+    # reindex) on this leg: the release harness never fails on a mismatch, so
+    # its point is retained either way -- unlike the per-merge series, where a
+    # mismatch turns the job red and the row has to carry the reason (#365).
     "objects_total",
     "objects_expected",
     # Store-layout axis + flat<->hive parity verdict (issue #240 phase 4).
--- a/.github/scripts/full_aoi_series.py
+++ b/.github/scripts/full_aoi_series.py
@@ -81,8 +81,10 @@ FULL_AOI_COLUMNS = [
     # when the layout's count is data-dependent). The run record's
     # ``objects_mismatch`` description stays JSON-only (dropped by the
     # reindex) on this leg: the release harness never fails on a mismatch, so
-    # its point is retained either way -- unlike the per-merge series, where a
-    # mismatch turns the job red and the row has to carry the reason (#365).
+    # its point is retained either way. The per-merge series takes the other
+    # route: a mismatch there turns the job red and today the row is discarded
+    # with it, and once the workflow gate of issue #365 lets that run reach the
+    # append, the row will carry the reason in its ``objects_mismatch`` column.
     "objects_total",
     "objects_expected",
     # Store-layout axis + flat<->hive parity verdict (issue #240 phase 4).
--- a/.github/scripts/update_series.py
+++ b/.github/scripts/update_series.py
@@ -6,6 +6,10 @@ pre-merge PR runs are reported as an ephemeral comment and dropped (too noisy
 while chasing a regression). This module is the read/append/write core plus a
 thin CLI; ``plot_series.py`` renders the GitHub Pages charts from the same file.
 
+This is also where a run's records are judged RETAINABLE (issue #365): the merge
+job appends whatever the benchmark's tripwires concluded, so rejecting a junk
+measurement is this module's job, not the job ordering's.
+
 Re-running a merge (a re-dispatch of the same commit) replaces that commit's rows
 rather than double-counting, so the series stays one row per (commit, target).
 """
@@ -44,7 +48,16 @@ def append_records(existing: pd.DataFrame, records: list[dict]) -> pd.DataFrame:
 
     Keeping the last write makes a merge re-run idempotent instead of duplicating
     a point in the plotted history.
+
+    Appending NOTHING leaves the series byte-identical rather than rewriting it:
+    concat against an all-null frame widens every integer column to float (issue
+    #365). That path is no longer rare -- the retention filter in :func:`main`
+    drops an unusable run's records, so a silently OOM'd merge now arrives here
+    with an empty list, and a run that measured nothing must not be able to
+    change the dtypes of every point that came before it.
     """
+    if not records:
+        return existing.reset_index(drop=True)
     new = records_to_frame(records)
     # Avoid concat with an all-empty frame (pandas FutureWarning on dtype union).
     combined = new if existing.empty else pd.concat([existing, new], ignore_index=True)
--- a/.github/scripts/update_series.py
+++ b/.github/scripts/update_series.py
@@ -6,9 +6,12 @@ pre-merge PR runs are reported as an ephemeral comment and dropped (too noisy
 while chasing a regression). This module is the read/append/write core plus a
 thin CLI; ``plot_series.py`` renders the GitHub Pages charts from the same file.
 
-This is also where a run's records are judged RETAINABLE (issue #365): the merge
-job appends whatever the benchmark's tripwires concluded, so rejecting a junk
-measurement is this module's job, not the job ordering's.
+This is also where a run's records are judged RETAINABLE (issue #365): once the
+merge job appends whatever the benchmark's tripwires concluded, rejecting a junk
+measurement becomes this module's job rather than the job ordering's. That
+workflow gate is a later phase and is NOT on this branch -- today the append
+still only runs when the tripwires passed, so the filter in :func:`main` is the
+precondition for that change, inert until the gate lands.
 
 Re-running a merge (a re-dispatch of the same commit) replaces that commit's rows
 rather than double-counting, so the series stays one row per (commit, target).
@@ -49,12 +52,14 @@ def append_records(existing: pd.DataFrame, records: list[dict]) -> pd.DataFrame:
     Keeping the last write makes a merge re-run idempotent instead of duplicating
     a point in the plotted history.
 
-    Appending NOTHING leaves the series byte-identical rather than rewriting it:
-    concat against an all-null frame widens every integer column to float (issue
-    #365). That path is no longer rare -- the retention filter in :func:`main`
-    drops an unusable run's records, so a silently OOM'd merge now arrives here
-    with an empty list, and a run that measured nothing must not be able to
-    change the dtypes of every point that came before it.
+    Appending NOTHING returns the series unchanged, so the unconditional
+    :func:`save_series` that follows rewrites it byte-identically and the data
+    branch sees no commit: concat against an all-null frame would instead widen
+    every integer column to float (issue #365). That path stops being rare once
+    the workflow gate lands -- the retention filter in :func:`main` drops an
+    unusable run's records, so a silently OOM'd merge will arrive here with an
+    empty list, and a run that measured nothing must not be able to change the
+    dtypes of every point that came before it.
     """
     if not records:
         return existing.reset_index(drop=True)
@@ -84,9 +89,11 @@ def main(argv: list[str] | None = None) -> int:
         raise SystemExit("records JSON must be a list of record objects")
 
     # What may be retained is decided HERE, at the retention boundary, rather than
-    # by whether the benchmark job got this far (issue #365): the merge workflow now
-    # appends even when a tripwire fired, so a run whose metrics are junk must be
-    # rejected by this filter instead of by the job aborting before the append.
+    # by whether the benchmark job got this far (issue #365): once the merge workflow
+    # appends even when a tripwire fired, a run whose metrics are junk has to be
+    # rejected by this filter instead of by the job aborting before the append. The
+    # workflow gate is a later phase, so until it lands this filter only ever sees
+    # runs that already passed their tripwires.
     #   - non-merge (the locked design): a stray PR record must never evict a
     #     retained merge point via the (commit, target) dedup.
     #   - empty metrics (issue #145): obs=0 / null peak memory is a silent OOM, a
--- a/.github/scripts/update_series.py
+++ b/.github/scripts/update_series.py
@@ -70,13 +70,28 @@ def main(argv: list[str] | None = None) -> int:
     if not isinstance(records, list):
         raise SystemExit("records JSON must be a list of record objects")
 
-    # Only merge runs are retained (the locked design). Enforce it at the boundary
-    # so a stray non-merge record can never evict a retained merge point via the
-    # (commit, target) dedup. Report drops -- a silent skip would read as "stored".
-    retained = [r for r in records if r.get("event") == "merge"]
-    dropped = len(records) - len(retained)
-    if dropped:
-        print(f"skipping {dropped} non-merge record(s); only merge runs are retained")
+    # What may be retained is decided HERE, at the retention boundary, rather than
+    # by whether the benchmark job got this far (issue #365): the merge workflow now
+    # appends even when a tripwire fired, so a run whose metrics are junk must be
+    # rejected by this filter instead of by the job aborting before the append.
+    #   - non-merge (the locked design): a stray PR record must never evict a
+    #     retained merge point via the (commit, target) dedup.
+    #   - empty metrics (issue #145): obs=0 / null peak memory is a silent OOM, a
+    #     junk point that would plot as a real dip to zero.
+    # An object-count mismatch (issue #240) is NOT in this list -- that run measured
+    # fine and is exactly the point worth keeping; its verdict rides the row's
+    # ``objects_mismatch`` column. Every drop is reported: a silent skip reads as
+    # "stored".
+    retained, dropped = [], []
+    for r in records:
+        if r.get("event") != "merge":
+            dropped.append((r.get("target"), "non-merge run"))
+        elif bench_metrics.has_empty_metrics(r):
+            dropped.append((r.get("target"), "empty metrics (obs=0 / no peak memory)"))
+        else:
+            retained.append(r)
+    for target, why in dropped:
+        print(f"skipping record for '{target}': {why}")
 
     existing = load_series(args.series)
     updated = append_records(existing, retained)
--- a/tests/test_benchmark_recording.py
+++ b/tests/test_benchmark_recording.py
@@ -152,6 +152,48 @@ def test_an_all_dropped_batch_leaves_the_series_untouched(tmp_path):
     assert after.equals(before)
 
 
+def test_appending_to_a_series_that_predates_the_column(tmp_path):
+    # The riskiest property here: series.parquet is a long-lived file on the
+    # benchmarks data branch, written before ``objects_mismatch`` existed, and
+    # append_records never reindexes the COMBINED frame -- the column arrives
+    # purely via pd.concat's column union. Every other test builds its "existing"
+    # frame from records_to_frame, which already has the column, so this is the
+    # one path they don't cover.
+    series = tmp_path / "series.parquet"
+    legacy = update_series.records_to_frame([_rec(commit="c0")]).drop(columns=["objects_mismatch"])
+    update_series.save_series(legacy, series)
+    assert "objects_mismatch" not in update_series.load_series(series).columns
+
+    update_series.main(
+        [
+            "--series",
+            str(series),
+            "--records",
+            _write(tmp_path, [_rec(commit="c1", mismatch="expected 10, measured 12")]),
+        ]
+    )
+    out = update_series.load_series(series)
+    # The column is added, the pre-existing row reads as "no anomaly" rather than
+    # dropping out, and the history keeps its order and its integer dtypes.
+    assert list(out.columns) == list(bench_metrics.RECORD_COLUMNS)
+    assert out["commit"].tolist() == ["c0", "c1"]
+    assert out["objects_mismatch"].tolist() == [None, "expected 10, measured 12"]
+    assert out["total_obs"].dtype == legacy["total_obs"].dtype
+
+    # ...and a later batch onto that now-present column still lands as a string.
+    update_series.main(
+        [
+            "--series",
+            str(series),
+            "--records",
+            _write(tmp_path, [_rec(commit="c2", mismatch="expected 10, measured 11")]),
+        ]
+    )
+    out = update_series.load_series(series)
+    assert out["objects_mismatch"].tolist()[-1] == "expected 10, measured 11"
+    assert out["total_obs"].dtype == legacy["total_obs"].dtype
+
+
 def test_mixed_batch_keeps_the_good_records(tmp_path):
     series = tmp_path / "series.parquet"
     update_series.main(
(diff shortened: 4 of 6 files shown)


espg:
🤖 *from Claude*

**Blocked on a push permission, with a decision needed.** Phase 1 is pushed and green. Phase 2 — the actual ordering fix in `.github/workflows/lambda-benchmark.yml` — is written and green locally but **could not be pushed**:

```
 ! [remote rejected] claude/365-record-before-assert -> claude/365-record-before-assert
   (refusing to allow an OAuth App to create or update workflow
    `.github/workflows/lambda-benchmark.yml` without `workflow` scope)
```

Issue #365 names that file, so CLAUDE.md §1's carve-out permits the edit — the routine's git credential does not carry the scope to make it. I did **not** re-route the same edit through another credential path (the GitHub API is available to me and might have accepted it): defeating a scope restriction on CI config is exactly the kind of thing that should be asked about rather than probed. The commit was reset out of the branch and is reproduced below in full, so nothing is lost.

**There is no equivalent that avoids `.github/workflows/`.** I checked: `.github/actions/run-benchmark/action.yml` *is* pushable (the restriction is path-scoped to `workflows/`), but the composite action runs **before** the retain steps, so nothing it can do makes the job go red *after* the append. The issue requires both — the point recorded *and* the alarm firing — and only a workflow-level `if:` (or a post-push failing step) delivers that pair.

### Options (any one answer unblocks the next run)

**(A)** Grant the routine's push credential the `workflow` scope, and the next run pushes the commit below as-is. Narrowest change to the situation; leaves the routine able to edit any workflow file in future, which is the reason the scope is withheld.

**(B)** Apply the patch below by hand onto `claude/365-record-before-assert` (`git am` on the attached commit, or paste the three `if:` blocks). The next run continues from there with phase 3 and the review pass. No standing permission change.

**(C)** Reshape the fix so the workflow file is not touched at all — accepting that the merge job no longer goes **red** on an object-count mismatch, and the alarm becomes "the recorded row carries `objects_mismatch`, and the chart/table surfaces it". This contradicts the issue's "the alarm still fires (the job still goes red)", so I would not take it without an explicit call.

**(D)** Land phase 1 only (retention boundary + the recorded verdict) and split the workflow ordering into its own issue for a human to carry.

Preference: **(B)** — it unblocks this PR without widening what the routine can write, and the patch is already reviewed against the acceptance criteria.

<details>
<summary>Phase 2 patch — <code>git am</code>-able (workflow + its structural tests)</summary>

```diff
diff --git a/.github/workflows/lambda-benchmark.yml b/.github/workflows/lambda-benchmark.yml
--- a/.github/workflows/lambda-benchmark.yml
+++ b/.github/workflows/lambda-benchmark.yml
@@ -200,7 +200,24 @@ jobs:
           commit: ${{ github.sha }}
           ref-label: ${{ github.ref_name }}
 
+      # Record the measurement, THEN let the verdict stand (issue #365). The
+      # tripwires inside the step above -- the object-count mismatch (issue #240)
+      # and the empty-metrics guard (issue #145) -- are evaluated AFTER
+      # metrics.json is written, so a red step still leaves a real, completed
+      # measurement on disk. These three steps used to be skipped by it, which
+      # silently deleted the run's data point (a hole in series.parquet reads
+      # exactly like "no run happened"), and every future tripwire would have
+      # inherited that. So they opt out of the skip and gate on the ARTIFACT
+      # instead of the step's status: any tripwire that fires after the records
+      # are written now lands its point, while a run that died BEFORE writing
+      # them (login/dispatch failure) matches nothing and skips cleanly. The job
+      # still goes red -- the measuring step is not continue-on-error, and the
+      # anomaly also rides the recorded row's ``objects_mismatch`` column.
+      # Retaining a point is NOT the same as retaining junk: an unusable
+      # measurement is rejected by update_series.py at the retention boundary.
       - name: Check out benchmarks data branch
+        if: >-
+          !cancelled() && hashFiles('metrics.json') != ''
         uses: actions/checkout@v5
         with:
           ref: benchmarks
@@ -208,6 +225,8 @@ jobs:
           fetch-depth: 0
 
       - name: Append series + render charts
+        if: >-
+          !cancelled() && hashFiles('metrics.json') != ''
         run: |
           set -euo pipefail
           uv sync --extra benchmark
@@ -217,6 +236,8 @@ jobs:
             --series _series/series.parquet --out _series/site
 
       - name: Commit + push to benchmarks branch
+        if: >-
+          !cancelled() && hashFiles('metrics.json') != ''
         working-directory: _series
         run: |
           set -euo pipefail
```

The companion test block appended to `tests/test_benchmark_recording.py` (plus `import yaml` at the top, and the module docstring gaining the workflow half):

```python
WORKFLOW = REPO / ".github" / "workflows" / "lambda-benchmark.yml"
#: The steps that turn a completed benchmark into a retained series point --
#: the ones a failed measuring step used to skip.
RETAIN_STEPS = [
    "Check out benchmarks data branch",
    "Append series + render charts",
    "Commit + push to benchmarks branch",
]


def _merge_publish_steps():
    return yaml.safe_load(WORKFLOW.read_text())["jobs"]["merge-publish"]["steps"]


def test_the_retain_steps_run_after_the_measuring_step_in_order():
    # Ordering is the whole subject: the measurement is taken first, then the
    # steps that record it -- checkout, append/render, push, in that order.
    names = [s.get("name") for s in _merge_publish_steps()]
    assert "Run benchmark" in names
    at = [names.index(n) for n in RETAIN_STEPS]
    assert at == sorted(at), f"the retain steps in {WORKFLOW} are out of order"
    assert at[0] > names.index("Run benchmark")


@pytest.mark.parametrize("name", RETAIN_STEPS)
def test_a_retain_step_survives_a_failed_measuring_step(name):
    # Without a status-function condition a step is skipped as soon as an earlier
    # one fails -- which is how a tripwire came to discard a real data point. Each
    # retain step opts out of that, and gates on the ARTIFACT so a run that died
    # BEFORE writing metrics.json skips cleanly instead of erroring (issue #365).
    step = next(s for s in _merge_publish_steps() if s.get("name") == name)
    condition = " ".join(str(step.get("if", "")).split())
    assert "!cancelled()" in condition, (
        f"step '{name}' in {WORKFLOW} is skipped when the benchmark step fails, so "
        "a tripwire would again discard the measurement (issue #365)"
    )
    assert "hashFiles('metrics.json') != ''" in condition, (
        f"step '{name}' in {WORKFLOW} must gate on metrics.json existing, so a run "
        "that failed before writing records skips instead of erroring"
    )


def test_the_measuring_step_still_fails_the_job():
    # Record-first must not become record-and-shrug: the alarm still fires. The
    # tripwires live in the composite action, so the step invoking it must NOT be
    # continue-on-error -- that would leave the job green on a mismatch.
    step = next(s for s in _merge_publish_steps() if s.get("name") == "Run benchmark")
    assert step["uses"] == "./.github/actions/run-benchmark"
    assert not step.get("continue-on-error"), (
        f"'Run benchmark' in {WORKFLOW} is continue-on-error, so an object-count "
        "mismatch would no longer turn the job red (issue #365)"
    )
```

With both applied, `uv run pytest tests/test_benchmark_recording.py -q` → **13 passed**.

</details>


---

espg:
🤖 *from Claude*

**⚠️ The per-phase adversarial self-review ran in the WEAKER fallback mode.** CLAUDE.md §2 calls for a *separate* review subagent on an Opus-class model with fresh context, and a *separate* fold subagent. **No agent-spawning tool was available to this run** (the tool surface offered no `Agent`/`Task`), so the review below was done in the same context as the implementation — the reviewer is not independent of the author, and that is exactly the weakness the two-subagent design exists to remove. Treat these findings as lower-assurance than a normal review pass, and the absence of findings on the phase-2 patch as weaker still.

### Findings

**(1) Landed — an all-dropped batch silently rewrote the retained history.** `7930d46`. The new retention filter means a silently OOM'd merge now reaches `append_records` with an **empty** list. That path went straight into `pd.concat([existing, all_null_frame])`, which widens every integer column to float:

```
before: int64 int64     # total_obs, shard_key
after : float64 float64
```

So a run that measured *nothing* could change the dtypes of every point that came before it — the same "a bad run damages good data" shape this issue is about, reintroduced one layer down. `append_records` now returns the existing frame untouched when there is nothing to append, mirroring the guard already on the other side of that concat. Test: `test_an_all_dropped_batch_leaves_the_series_untouched` asserts both `dtype` and `DataFrame.equals` are unchanged.

**(2) Standing, not folded — `objects_mismatch` on the release/full-AOI leg.** The column was added to the per-merge series only. Adding it to `full_aoi_series.py` would change what the *release* leg records, which is a scope change beyond what issue #365 asks for, so it stays a question for review rather than a fold (question 3 in the PR body). The stale cross-reference in that module's comment was corrected in phase 1.

**(3) Declined, by design — `has_empty_metrics` is not NaN-safe.** A record read back from parquet carries `NaN`, not `None`, for a missing `max_memory_mb`, and the predicate would read that as "usable". It is only ever applied to fresh records out of `metrics.json` (JSON `null` → `None`), never to series rows, so NaN handling would be speculative code for a caller that does not exist. Noted here rather than silently.

**(4) Declined — no test pins the two callers of the predicate together.** `run_benchmark`'s `--fail-on-empty` and the retention filter now share `bench_metrics.has_empty_metrics`; drift would require deleting a call, and both sides already have behavioural tests. A test asserting "these two agree" would only restate the shared function.

Local gate after the fold: `uv run pytest -q` → 3204 passed / 37 skipped on the phase-1 tree, and 193 passed / 27 skipped across the benchmark + deploy modules on the folded tree. `ruff check` / `ruff format --check` clean on every touched file (the two pre-existing failures on untouched files are listed in the PR body).


---

espg:
🤖 *from Claude*

**Run status — stopping here on `waiting`; the ordering half needs a call before it can land.**

**Landed (2 commits, both pushed):**
- `c4830e3` phase 1 — `bench_metrics.has_empty_metrics` as the one definition of "no usable measurement", the retention filter in `update_series.main`, `objects_mismatch` promoted to a retained series column, and `tests/test_benchmark_recording.py`.
- `7930d46` fold review — an all-dropped batch no longer rewrites the retained series (int columns were silently widening to float).

**Not landed:**
- Phase 2 (the `.github/workflows/lambda-benchmark.yml` ordering) — **push rejected for lack of `workflow` scope**; written, green locally, and posted in full [above](https://github.com/englacial/zagg/pull/368#issuecomment-5165249856) with options (A)–(D). Deliberately not re-routed through another credential.
- Phase 3 (docs) — held, so `docs/deployment/benchmark.md` never describes a half-applied behavior.

**Review/fold:** ran in the **weaker same-context fallback** — no agent-spawning tool was available this run, so the reviewer was not independent of the author. Details and the four findings (one folded, three standing with rationale) are in the [self-review comment](https://github.com/englacial/zagg/pull/368#issuecomment-5165278239).

**CI on `7930d46`:** `test (3.12)` ✅, `test (3.13)` ✅, `ruff` ✅, `build` ✅. Locally: 3204 passed / 37 skipped, `ruff check` and `ruff format --check` clean on every touched file (two pre-existing failures on untouched files are itemised in the PR description). No linter-bot inline comments were posted on this PR — nothing left unanswered.

**Labels:** `implement` + `waiting`. **The PR stays a draft** — not flipped to ready-for-review even though CI is green, because the phase checklist is incomplete; the routine's instruction for this run also overrode the CLAUDE.md §2 draft→ready flip, and that deviation is recorded here rather than left silent.


---

.github/scripts/bench_metrics.py
```diff
@@ -253,6 +260,22 @@ def memory_pct_of_cap(max_memory_mb, memory_gb) -> float | None:
...
+    silent OOM -- there is no measurement in the row, only the shape of one.
+
+    ONE definition, shared by ``run_benchmark``'s ``--fail-on-empty`` tripwire and
+    ``update_series``'s retention filter (issue #365). The two used to be the same
+    predicate by construction: an empty run aborted the job, so the append step
+    never ran. Now that the append runs regardless of the tripwire verdict, the
+    appender has to recognise exactly the junk row the tripwire names -- otherwise
+    record-before-assert would retain the obs=0 point it was meant to reject.
+    """
+    return not record.get("total_obs") or record.get("max_memory_mb") is None
```
espg: 🤖 *from Claude (review)*

**[diff-scoped]** The predicate is not NaN-safe, and the argument that NaN is "speculative" is contradicted 12 lines above it in this same file.

```python
return not record.get("total_obs") or record.get("max_memory_mb") is None
```

`memory_pct_of_cap`, immediately above, already treats NaN as a live shape of exactly this field:

```python
if max_memory_mb is None or memory_gb is None:
    return None
# A legacy parquet row reads back as NaN, not None, after the reindex.
if isinstance(max_memory_mb, float) and math.isnan(max_memory_mb):
    return None
```

Concrete path (verified against this branch, pandas 2.2.3 / pyarrow 25.0.0): `json.dumps` emits the bare token `NaN` for a float nan and `json.loads` reads it straight back, so a `metrics.json` carrying `"max_memory_mb": NaN` survives the round-trip into `update_series.main`. `has_empty_metrics` then returns `False`, the row is **retained**, and it lands in `series.parquet` as a real point whose memory cell renders `n/a` — the precise silent-OOM shape issue #145 exists to reject. Same for `"total_obs": NaN` (`not float("nan")` is `False`).

I could not identify a producer that emits `NaN` today — the worker summary would have to serialise one — so this is hardening, not a live bug. But two things make it worth doing here rather than deferring: (1) this PR promotes the expression into *the* single "no usable measurement" definition and documents it as recognising "exactly the junk row the tripwire names", which raises the bar on it; (2) the guard is already written in this module, so the fold is a one-liner reusing `math.isnan` rather than new machinery.

Separately on this docstring — "**Now that** the append runs regardless of the tripwire verdict" is present-tense about behavior that does not exist on `main` yet (see the note on `update_series.py`). Worth tensing to the phase-2 dependency.

---

.github/scripts/update_series.py
```diff
@@ -6,6 +6,10 @@
 while chasing a regression). This module is the read/append/write core plus a
 thin CLI; ``plot_series.py`` renders the GitHub Pages charts from the same file.
 
+This is also where a run's records are judged RETAINABLE (issue #365): the merge
+job appends whatever the benchmark's tripwires concluded, so rejecting a junk
```
espg: 🤖 *from Claude (review)*

**[diff-scoped]** This docstring states phase-2 behavior in the present tense, and it will be false the moment this branch merges.

```
This is also where a run's records are judged RETAINABLE (issue #365): the merge
job appends whatever the benchmark's tripwires concluded, ...
```

The merge job does not do that on this branch. `.github/workflows/lambda-benchmark.yml`'s `merge-publish` is unchanged here — "Run benchmark" has no `continue-on-error`, and the three retain steps carry no gate:

```yaml
      - name: Run benchmark
        uses: ./.github/actions/run-benchmark
      - name: Check out benchmarks data branch
      - name: Append series + render charts
      - name: Commit + push to benchmarks branch
```

So `run_benchmark.py` still `return 1`s at the object-count tripwire and the append never runs. Two consequences for the code this PR *did* land, which I think are worth saying out loud in the module rather than only in the PR body:

1. **`objects_mismatch` is structurally unpopulatable.** The composite action passes neither `--no-fail-on-object-mismatch` nor `--no-fail-on-empty`, both default `True`. The only record that would carry a non-null verdict is exactly the record whose run aborts before `update_series.py` is invoked. Until phase 2 lands, the column is null on every row on both legs (the release leg deliberately doesn't get it — question (3)).
2. **The empty-metrics arm of the filter is unreachable.** `--fail-on-empty` returns 1 before the append step, so an `obs=0` record never arrives here.

None of that is a reason to hold the PR — the schema addition and the filter are the correct *preconditions* for phase 2, and landing them first is a sensible split. The ask is only that the comments say so, so `main` never documents a behavior it doesn't have. Same wording issue in `bench_metrics.has_empty_metrics` ("Now that the append runs...") and in `full_aoi_series.py` (separate comment).

Minor, same file: `append_records`'s new docstring says the empty-batch path "leaves the series byte-identical **rather than rewriting it**" — but `main` calls `save_series` unconditionally, so the file *is* rewritten. I verified the bytes do come out identical (so `git diff --cached --quiet` still suppresses the commit, and the guarantee that matters holds), but "rather than rewriting it" describes an early return that isn't there.

---

.github/scripts/full_aoi_series.py
```diff
@@ -80,7 +80,9 @@
     # measured store object total and the config-derived expectation (null
     # when the layout's count is data-dependent). The run record's
     # ``objects_mismatch`` description stays JSON-only (dropped by the
-    # reindex), like the per-merge series.
+    # reindex) on this leg: the release harness never fails on a mismatch, so
+    # its point is retained either way -- unlike the per-merge series, where a
+    # mismatch turns the job red and the row has to carry the reason (#365).
```
espg: 🤖 *from Claude (review)*

**[diff-scoped]** Same present-tense problem as the `update_series.py` docstring, and this one is the most misleading of the three because it is a *contrast* — it tells the reader what the other leg does.

```python
    # its point is retained either way -- unlike the per-merge series, where a
    # mismatch turns the job red and the row has to carry the reason (#365).
```

On this branch a mismatch turns the per-merge job red **and the row is still discarded** — the very behavior issue #365 was filed against. So a reader of `main` after this merges learns that the per-merge series retains mismatched rows with a reason column, and it does not. The contrast this comment draws is exactly backwards until phase 2 lands.

Suggest tensing it ("will carry" / "once the workflow gate lands"), or dropping the contrast clause and keeping only the statement about this leg, which is true today.

---

tests/test_benchmark_recording.py
```diff
@@ -0,0 +1,157 @@
...
+                [
+                    _rec(target="t1", mismatch="expected 10, measured 12"),
+                    _rec(target="t2", obs=0, mem=None),
+                    _rec(target="t3"),
+                ],
+            ),
+        ]
+    )
+    out = update_series.load_series(series)
+    assert sorted(out["target"]) == ["t1", "t3"]
```
espg: 🤖 *from Claude (review)*

**[diff-scoped, missing tests]** Two gaps. I ran both against this branch first — **neither is a bug**, so these are regression pins, not fixes.

**(1) Nothing appends to a series that predates the `objects_mismatch` column.** This is the riskiest property in the PR: `series.parquet` is a long-lived file on the `benchmarks` data branch with no such column, and `append_records` never reindexes the combined frame — the new column arrives purely via `pd.concat`'s column union. Every test here builds its "existing" series from `records_to_frame`, i.e. from a frame that *already has* the column, so the schema-evolution path is the one case the module doesn't exercise. Built the real one (a parquet written with `RECORD_COLUMNS` minus `objects_mismatch`) and appended:

```
old has objects_mismatch? False
cols tail: ['variant_guard', 'objects_write_path', 'objects_mismatch']
objects_mismatch dtype: object [None, None]
total_obs dtype: int64
col order == RECORD_COLUMNS? True
```

and a later batch carrying a string onto that all-null column: `object ['expected 10, measured 12']`, no dtype clash. (The intermediate parquet holds the column as arrow `null` type, which round-trips fine.) All good — but this is the assertion a future edit to `append_records` most needs to trip over, and `test_objects_mismatch_is_a_retained_column` only checks `RECORD_COLUMNS` membership plus a fresh frame, which is closer to restating the implementation than pinning the risk.

**(2) The concat/dtype fix from `7930d46` is pinned only for the all-dropped case.** `test_an_all_dropped_batch_leaves_the_series_untouched` covers `records == []`, which takes the new `if not records:` early return. The **partially**-dropped batch takes the *other* branch — a real `pd.concat` — and is covered only by `test_mixed_batch_keeps_the_good_records` right here, which asserts targets and says nothing about dtype:

```python
    assert sorted(out["target"]) == ["t1", "t3"]
```

Verified the partial case is currently safe (existing `total_obs`/`objects_total` stay `int64` across a 2-record batch with one dropped), so this is one added assertion on the existing test rather than a new one.

Nit, same file: `_rec()`'s defaults are `objects_total=12` / `objects_expected=10` with `mismatch=None` — a "clean" record whose counts disagree. Harmless, but it's the helper every test reads to understand the fixtures; `12`/`12` would say what's meant.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of the landed (Python) half. **The code is sound** — I could not find a correctness bug in it. I ran the paths the tests don't cover (pandas 2.2.3 / pyarrow 25.0.0, scripts exported from `7930d46`) and they behave: schema evolution onto a column-less `series.parquet` works and preserves column order + `int64`; the partial-drop concat does not widen dtypes; the all-dropped early return leaves the file byte-identical; a first-ever write types correctly. Four findings, all diff-scoped, none blocking.

1. **`has_empty_metrics` is not NaN-safe** — and the "speculative" rationale is contradicted by `memory_pct_of_cap` 12 lines above it, which explicitly guards `math.isnan` on the same field. `json.dumps`/`json.loads` round-trip a bare `NaN` token, so a `metrics.json` with `"max_memory_mb": NaN` is *retained* as a real point. No producer identified today, so it's hardening — but this PR promotes the expression to *the* single definition, and the guard is already in the file. **[fold]**
2. **Three comments state phase-2 behavior in the present tense** (`update_series.py` module docstring, `has_empty_metrics`, `full_aoi_series.py`). The `full_aoi_series.py` one is a *contrast* and is currently backwards. **[fold]**
3. **No test appends to a series predating `objects_mismatch`** — the riskiest property here, and the only path the module doesn't exercise. Verified working; wants a pin. **[fold]**
4. **The `7930d46` dtype fix is pinned only for the all-dropped case**; the partially-dropped batch takes the other branch and is asserted only on targets. Verified safe; one added assertion. **[fold]**

Minor: `append_records`'s docstring says the empty batch is not "rewritten", but `main` always calls `save_series` (bytes are identical, so the guarantee that matters holds); `_rec()`'s `objects_total=12`/`objects_expected=10` defaults describe a "clean" run whose counts disagree.

**Conventions:** clean. No new dependency; no weakened, skipped or deleted tests (the `test_benchmark.py` edits move `objects_mismatch` from the JSON-only set to the retained set and re-anchor the tail-of-schema assertions — they get *stronger*, `cols[-1]` becomes `cols[-2:]`); no blanket `# noqa`/`# type: ignore` (the `E402` markers match the existing `sys.path`-insert idiom in the neighbouring scripts); commit subjects are title-only per §3; every touched module is far under the ~1,200-line limit (`bench_metrics.py` 535, `update_series.py` 119, new test module 157), and splitting the new tests out rather than growing the 2,468-line `test_benchmark.py` is the right call.

## On merging the landed half without the workflow change

**Safe, but inert — not half-applied.** It changes nothing observable on `main` today, which is both the reassurance and the caveat:

- `merge-publish` is untouched, so `run_benchmark.py` still `return 1`s at the object-count tripwire before the three retain steps. Ordering — the actual subject of issue #365 — is **not** achieved by any part of this branch.
- The composite action passes neither `--no-fail-on-object-mismatch` nor `--no-fail-on-empty` (both default `True`), so the only record that could carry a non-null `objects_mismatch` is exactly the record whose job aborts before the append. The new column is **structurally unpopulatable** until phase 2, and the empty-metrics arm of the retention filter is **unreachable** for the same reason.

So the landed half is precondition-only: a schema addition plus a filter that is correct and currently dead. That is a reasonable split and the sequencing is right (the retention boundary genuinely must exist before the append becomes unconditional). The one thing that makes it read as half-applied is documentation, not code — finding (2). Fixing the three tenses makes the merge unambiguously coherent standing alone: `main` would then carry a correct filter, a correct schema, and comments that name the phase-2 dependency instead of asserting behavior the repo doesn't have.

The one visible side effect of merging alone: the next merge benchmark rewrites `series.parquet` with an all-null `objects_mismatch` column (arrow `null` type). Verified harmless on read/append/re-append, and consistent with how prior columns (`objects_write_path`, `variant_guard`) were introduced.

No opinion offered on question (1) — how the workflow half should land is an @espg call, and correctly left as one.

---

.github/scripts/bench_metrics.py
espg: 🤖 *from Claude*

Both folded.

**NaN guard** — `1a1cc96`. Lifted the inline guard from `memory_pct_of_cap` into a module-private `_is_nan` and used it on both fields, so the two callers share one definition of "missing number" instead of one guarding and one not:

```python
    max_memory_mb = record.get("max_memory_mb")
    total_obs = record.get("total_obs")
    return not total_obs or _is_nan(total_obs) or max_memory_mb is None or _is_nan(max_memory_mb)
```

`memory_pct_of_cap` now calls `_is_nan(max_memory_mb)` in place of its own `isinstance`/`math.isnan` pair — same behavior, no second copy to drift. Test in `test_has_empty_metrics_treats_nan_as_missing`: the predicate directly on either field, plus the end-to-end path you described — the record goes through `json.dumps`/`json.loads`, the assertion `records[0]["max_memory_mb"] != records[0]["max_memory_mb"]` confirms the bare `NaN` token really did survive the round-trip, and `update_series.main` then drops the row and leaves the series empty.

**Tense** — `80f6d4e`. The docstring now reads "The two are still the same predicate by construction today: an empty run aborts the job, so the append step never runs. Once the append runs regardless of the tripwire verdict (the workflow gate, a later phase), the appender has to recognise exactly the junk row the tripwire names."

---

.github/scripts/update_series.py
espg: 🤖 *from Claude*

Folded in `80f6d4e` — both the docstring and the minor.

**Module docstring** now says the gate is not on this branch, and names the filter as the precondition rather than as something already load-bearing:

> This is also where a run's records are judged RETAINABLE (issue #365): once the merge job appends whatever the benchmark's tripwires concluded, rejecting a junk measurement becomes this module's job rather than the job ordering's. That workflow gate is a later phase and is NOT on this branch -- today the append still only runs when the tripwires passed, so the filter in `main` is the precondition for that change, inert until the gate lands.

**`main`'s retention comment** had the same present-tense claim ("the merge workflow now appends even when a tripwire fired") one screen down, so it got the same treatment — "once the merge workflow appends even when a tripwire fired ... The workflow gate is a later phase, so until it lands this filter only ever sees runs that already passed their tripwires." That covers your two consequences: with the gate absent, the `objects_mismatch` column is unpopulatable and the empty-metrics arm unreachable, and the comment no longer implies otherwise.

**`append_records`'s "rather than rewriting it"** — you're right that there is no early return at the file level; `save_series` is unconditional. Reworded to describe what actually happens, keeping the guarantee that matters (the empty data-branch commit):

> Appending NOTHING returns the series unchanged, so the unconditional `save_series` that follows rewrites it byte-identically and the data branch sees no commit: concat against an all-null frame would instead widen every integer column to float (issue #365).

---

.github/scripts/full_aoi_series.py
espg: 🤖 *from Claude*

Folded in `80f6d4e`. Kept the contrast but made it directional, so the reader gets both what the per-merge leg does today and what it will do, rather than a claim that is backwards until the gate lands:

```python
    # its point is retained either way. The per-merge series takes the other
    # route: a mismatch there turns the job red and today the row is discarded
    # with it, and once the workflow gate of issue #365 lets that run reach the
    # append, the row will carry the reason in its ``objects_mismatch`` column.
```

"today the row is discarded with it" is the behavior issue #365 was filed against, so the comment now names the problem rather than describing the fix as already in place.

---

tests/test_benchmark_recording.py
espg: 🤖 *from Claude*

All three added.

**(1) Schema evolution** — `58e0d96`, `test_appending_to_a_series_that_predates_the_column`. Builds the "existing" series the way the data branch actually has it — `records_to_frame(...).drop(columns=["objects_mismatch"])`, written to parquet, with `assert "objects_mismatch" not in load_series(series).columns` so the fixture can't silently stop being a legacy one. Then appends through `main` and pins the four properties you measured: `list(out.columns) == list(RECORD_COLUMNS)`, the pre-existing row reads `None` rather than dropping out (`["c0", "c1"]` / `[None, "expected 10, measured 12"]`), and `total_obs` keeps its dtype. It also runs your second step — a later batch onto the now-present column — to cover the arrow `null`-type intermediate round-tripping back and taking a string.

**(2) Partial-batch dtypes** — `220498e`, on `test_mixed_batch_keeps_the_good_records`. One note while implementing: the test as written started from a *nonexistent* series, so `existing.empty` was true and it took the `combined = new` path, not a real `pd.concat`. Seeded it with a prior-commit row (`c0`/`t0`) so it exercises the branch you described, then asserted dtypes against the pre-append frame:

```python
    assert sorted(out["target"]) == ["t0", "t1", "t3"]
    assert out["total_obs"].dtype == before["total_obs"].dtype
    assert out["objects_total"].dtype == before["objects_total"].dtype
```

**(3) Nit** — same commit, `_rec()`'s defaults are now `objects_total=12` / `objects_expected=12`, so the default record is coherently clean. Left the `mismatch=` strings the callers pass as they are: they're the harness's free-text verdict, not derived from the fixture's columns, and rewriting them to match would touch assertions in three tests for no gain in what's pinned.

---

espg:
🤖 *from Claude*

Adversarial-review fold complete — all four inline findings folded, each with a reply on its thread. Nothing from the review is left standing for decision.

| Finding | Commit |
| --- | --- |
| `has_empty_metrics` was not NaN-safe — guard lifted into a shared `_is_nan` helper and applied to both `total_obs` and `max_memory_mb`, with a test covering the `metrics.json` round-trip | `1a1cc96` |
| Four comments (three flagged, plus the same claim in `update_series.main`) asserted the phase-2 workflow behavior in the present tense; the `full_aoi_series.py` contrast was backwards. All now describe what this branch does, and name the gate as a later phase | `80f6d4e` |
| No test appended onto a series predating the `objects_mismatch` column — the schema-evolution path that only `pd.concat`'s column union covers. Pinned: column order, row order, the null on the pre-existing row, and the `int64` dtypes | `58e0d96` |
| The `7930d46` dtype fix was pinned only for the all-dropped batch; the partially-dropped batch takes the `pd.concat` branch. Dtype assertions added there (and the test seeded so it actually reaches that branch), plus the `_rec()` defaults nit | `220498e` |

Local gate on `220498e`: `uv run pytest -v` → **3207 passed, 37 skipped**; `ruff check` and `ruff format --check` clean over `src`, `tests` and `.github/scripts` apart from the two pre-existing failures already recorded in the PR body (`N818` in `src/zagg/registry.py`, and the `ruff format` diff on `tests/data/benchmark/README.md` — both on files this PR does not touch, neither fixed here).

Phase 1 is complete and its review findings are folded. The PR stays **`waiting`** and stays a draft: phase 2 is the `.github/workflows/lambda-benchmark.yml` half, which cannot be pushed with the current credential (no `workflow` OAuth scope), and the lettered options for how that lands are on the thread above awaiting a decision. No attempt was made to work around the missing scope through the API or another credential path. Phase 3 (docs) is deliberately held until phase 2 lands so the page never documents a half-applied behavior.

---

PR #369, opened 2026-08-03
output.pyramid: per-field overview selection (issue #352)

### LINKED ISSUE #352: output.pyramid: per-field overview selection — cost lever for tdigest vs exact folds

## Motivation

Which fields get overviews is currently class-driven only: `build_pyramid_block` includes every `exact` and `approximate` agg field at every declared order, and auto-excludes `none`-class fields (D24). There is no per-field lever — you cannot declare "`count` at every order but `h_tdigest` only every 4," or "dense fields yes, t-digests no." The cost asymmetry is real: exact folds (count/sum/min/max) are near-free dense merges, while the t-digest fold is a full ragged read + k-way merge per ancestor node.

The only current workarounds are all-or-nothing (`output.pyramid: false`) or coarsening the whole schedule (`spacing`/`orders`), which penalizes the cheap fields too.

## Proposal

Additive grammar inside the pyramid block (unknown keys already error, so this is non-breaking for existing configs):

```yaml
output:
  pyramid:
    fields:                 # optional; absent = today's behavior
      h_tdigest: {orders: [7, 5]}   # per-field schedule override
      some_field: false             # exclude from every overview order
```

(or an `exclude:` list, if per-field schedules prove overkill — decide at plan time). Defaults unchanged: every composable field at every declared order, so the pinned default sweep still hits everything. The manifest pyramid block records the per-field schedule, so readers get the recorded absence exactly as D24 option A intends (zero-open filtering, never a probe).

## Caveat

May be moot if the #344 Phase E/F benchmarks show the t-digest fold cost is negligible at production scale — espg's call was to have the option regardless, since the grammar has room.

Refs #201, PR #344, #346 review thread.

espg:
Closes #352

## What this does

Adds the additive `output.pyramid.fields` grammar — a per-field overview schedule, so which fields ride the pyramid is no longer class-driven only. The cost asymmetry issue #352 names is real: `fold_dense` is a `reshape(-1, factor)` reduce, while the approximate fold is a full ragged decode plus `merge_tdigests_kway` per ancestor node (`fold_digests`). Before this change the only levers were all-or-nothing (`output.pyramid: false`) or coarsening the whole schedule, which penalizes the near-free dense folds too.

```yaml
output:
  pyramid:
    orders: [7, 5, 3, 1]
    fields:                       # optional; absent = exactly today's behavior
      h_tdigest: {orders: [7, 5]} # the costly fold, only near the top
      photon_ids: false           # never in an overview
```

### Approach

**The `fields:` mapping, not an `exclude:` list.** The issue left this open ("or an `exclude:` list, if per-field schedules prove overkill — decide at plan time"). The code makes the answer concrete:

1. The issue's own motivating example — "`count` at every order but `h_tdigest` only every 4" — is **inexpressible** with an exclude list. An exclude list only covers the degenerate all-orders case.
2. The manifest already expresses exclusion *field-wise*: the D24 `none`-class auto-exclusion writes `{"class": "none"}` into `pyramid.overview.fields`, and the spec (§4.5) calls that the recorded absence. A per-field `orders` key generalizes that same shape uniformly rather than adding a second, differently-shaped absence record next to it.
3. `exclude: [f]` is a strict subset of `fields: {f: false}`, and the implementation cost of the difference is one intersection helper (`_field_orders`) plus one predicate (`_scheduled`) — the sweep-side filter is the same one line either way.

See lettered question (a) below anyway, since the issue flagged the fork explicitly.

**Where the logic lives.**

- `zagg.sweep_overview.pyramid_orders(knob, shard_order)` — extracted the block-schedule derivation (explicit `orders` win over `spacing`, default `DEFAULT_SPACING`) so config validation checks a per-field schedule against *the very list the sweep walks*, not a second derivation of it. Pure refactor of code that was inline in `build_pyramid_block`.
- `zagg.config._validate_pyramid_fields` — grammar + refusals. A per-field order outside the block schedule **refuses** rather than widening: the sweep only walks the block's orders, so a widening entry would silently do nothing. Scheduling a `none`-class field refuses for the same reason, while `false`/`[]` over one stays legal. `orders` is shape-checked (a `list` of `int`), so a string never sneaks through as a digit-wise schedule.
- `build_pyramid_block` — records the narrowed schedule as the field entry's own `orders` (`[]` for an outright exclusion). Fields the knob does not name carry **no** `orders` key, so a config without `fields:` declares byte-identically to pre-#352 zagg. A `none` entry stays class-only per spec §4.5 — it is absent everywhere already. An empty per-field schedule warns at template time on the same footing as the D24 `none` exclusion, and excluding *every* composable field says so loudly: that declares a pyramid the sweep would never write, which `output.pyramid: false` is the spelling for.
- `sweep_overviews` — the field set is now computed **per order** inside the `for k in orders` loop and passed down to `_roll_node`/`_write_overview`, so a narrowed-out field is absent from that order's zarr *and* from its `zagg_overview.fields` attrs map (spec §4.3: that map is the materialized subset). An order every field narrows out is skipped whole, and does not enter `materialized`.

`_content_hash` already digests `sorted(slabs)`, so changing a field's schedule changes the hash and the affected overviews regenerate rather than reading as `current` — covered by a test.

**The unvalidated caller.** `declare_pyramid` (the issue #358 retrofit) re-derives the block at the *manifest's* `shard_order`, which need not equal the config's `parent_order`, and never runs `validate_config`. Both helpers therefore refuse to be silent there: `_field_orders` warns — naming the field and the orders it dropped, loudest when nothing remains and the field leaves the pyramid entirely — and `_order_set` gates a declaration read back off a manifest on shape before iterating it, falling back to the full block schedule rather than narrowing to a schedule nobody declared.

### Phases

- [x] **Phase 1** — config grammar + validation (`output.pyramid.fields`), plus the shared `pyramid_orders` schedule helper (`10b5593`)
- [x] **Phase 2** — `build_pyramid_block` records the per-field schedule in the manifest block (`849ac1e`)
- [x] **Phase 3** — `sweep_overviews` per-order field filtering (`e50ae13`)
- [x] **Phase 4** — normative `docs/specification.md` §4.3/§4.4/§4.5 + `docs/hive_layout.md` narrative (`673ce1c`)
- [x] **Self-review round 1, folded** — `c4aa32b` (refuse a schedule on a `none`-class field), `e066a9a` (spec: the declaration is authoritative over stale overview objects), `c65b775` (docstring named the wrong validator), `aa6fc77` (retrofit-path test coverage)
- [x] **Self-review round 2, folded** — `59aafc2` (warn when a per-field schedule empties on the retrofit path), `f4e534b` (gate `orders` on shape — a `str` is iterable and its digits parse), `99b8b6d` (warn when a per-field schedule excludes the whole pyramid), `81a8540` (a pyramid-declaring §7 conformance fixture)

### Spec impact (CLAUDE.md §4)

This **is** an attrs/manifest grammar change, so `docs/specification.md` and the §7 conformance fixtures are updated in the same PR:

- §4.3 — a narrowed-out field must also be absent from `zagg_overview.fields`, which is why that map may differ between two overviews of the same product;
- §4.4 — inclusion is gated by class **and** by the field's own schedule;
- §4.5 — the new optional `fields[….].orders` key, its absent/present semantics, the `[]`-vs-`"class": "none"` distinction (cannot fold vs. not scheduled to), the reader rule that this is recorded absence and never an incomplete sweep, and that the declaration is authoritative over overview objects left behind by a wider previous schedule;
- §7 — a **new `pyramid/` fixture** (`tools/generate_spec_fixtures.py` → `tests/data/spec/pyramid/`), so the new grammar ships with something to decode against rather than prose alone. It is manifest-only by design — the pyramid block is a template-time artifact, decodable from `morton_hive.json` before any leaf exists — and one committed block carries all four readings a decoder must tell apart: a field that inherits the block schedule (`count`, no `orders` key), one narrowed to a subset (`h_min`, `[3, 1]` of `[3, 2, 1]`), one excluded outright but still declaring its fold law (`h_tdigest`, `orders: []`), and a non-composable `"class": "none"` one (`h_mean`).

The `spec` marker stays `zagg-pyramid/1`: the key is optional and its absence is the pre-existing behavior, so no existing reader is invalidated. See question (c).

### How it was tested

`uv sync --extra test`, then `uv run pytest -v` — **3219 passed, 37 skipped**. `uv run ruff check src tests` and `uv run ruff format --check src tests` are clean apart from the two pre-existing failures in question (d).

New tests in `tests/test_sweep_overview.py`:

- `TestPyramidBlock` — `test_validate_field_schedules` (narrow / exclude / explicit-block-schedule + the refusals, including a string `orders`), `test_validate_refuses_a_schedule_on_a_none_class_field`, `test_field_schedule_recorded_in_block`, `test_absent_field_knob_declares_identically` (the non-breaking guarantee), `test_field_schedule_never_widens_the_block` (and that the drop is warned, naming the field), `test_excluded_fields_warn_like_the_none_class_ones`, `test_excluding_every_composable_field_warns`, `test_unusable_field_schedule_inherits_the_block`, `test_pyramid_orders_derivation`.
- `TestOverviewWriter` — `test_per_field_schedule_drops_a_field_at_one_order` (real leaves swept at two orders: the t-digest lands at order 1, is absent from the order-0 zarr and its attrs, and the exact folds still land), `test_field_excluded_everywhere`, `test_order_with_no_scheduled_field_is_skipped` (nothing written, nothing recorded `materialized`), `test_unparseable_field_schedule_folds_everywhere`, `test_string_field_schedule_folds_everywhere`, `test_narrowed_schedule_regenerates_a_stale_overview`.
- `TestDeclarePyramid` — `test_retrofit_records_a_per_field_schedule` (the knob through the issue #358 retrofit, JSON round-trip and no-re-PUT idempotency included), `test_retrofit_warns_when_a_field_schedule_empties` (a schedule valid at the config's `parent_order` that intersects to nothing at the store's `shard_order`).

And in `tests/test_spec_conformance.py`, a new `TestPyramidDeclaration` that decodes the `pyramid/` fixture's committed manifest by §4.5 text alone: block shape and `zagg-pyramid/1` marker, the four per-field readings against `pyramid.expected.json`, subset-and-descending-order, an excluded field still carrying its fold law, and a `none` entry staying class-only.

## Questions for review

**(a) The grammar fork the issue left open.** Implemented as `fields: {name: {orders: [...]} | false}` for the three reasons above. The alternative shape is `exclude: [name, ...]`, which is smaller but cannot express the issue's own motivating case. Pick one:

- **(a1)** keep `fields:` as implemented (recommended);
- **(a2)** ship `exclude:` only — I drop the `orders` half and the manifest records `"orders": []` for excluded fields only;
- **(a3)** ship both, with `exclude:` as sugar over `fields: {name: false}`.

**(b) `src/zagg/sweep_overview.py` is over the ~1,200-line limit.** It was **1,227 lines on `main` before this PR** and is **1,361** after (+134: the schedule helpers, the shape gate, the per-order filter, the template-time warnings, and their docstrings). CLAUDE.md §4 pre-approves overages *below* 1,200, so the file was already past the bar independently of this change. Per the routine's instruction I am **raising this rather than splitting** — splitting a module is a scope change and not mine to take.

**Correction.** An earlier version of this question said a split of this file "is already proposed under PR #349". That was **factually wrong**, and the option built on it was too. PR #349 splits five *other* modules — its own before/after table lists `hive.py`, `processing/raster.py`, `sweep.py`, `config.py`, `runner.py` — and its phase-3 rationale cites this file as the convention it is imitating, not as a target: "Sibling modules also match the tree's existing convention: `zagg/sweep_overview.py` is already the split-out overview family." So there is **no pending split to defer to**: landing as-is parks nothing behind another PR, it leaves `sweep_overview.py` at 1,361 lines with nothing queued to bring it down.

Two further facts for the call:

- the additions (`pyramid_orders`, `_field_orders`, `_order_set`, `_scheduled`) are pure predicates over a declaration dict — no store, no zarr, no manifest I/O — so they would move out cleanly rather than splitting the file arbitrarily;
- this PR introduces the **first runtime import from `zagg.config` into `zagg.sweep_overview`** (`_validate_pyramid_fields` calls `pyramid_orders` so validation and the sweep share one derivation; on `main`, `config.py` names `zagg.sweep_overview` only in a docstring). A neutral `zagg/pyramid.py` owning that helper would give both a common dependency instead of an inverted one. Note PR #349 phase 4 moves `_validate_pyramid` into `config/validate_output.py`, so this import lands there on merge — relevant to ordering the two.

Options: **(b1)** land as-is and raise the file size separately; **(b2)** I move `pyramid_orders`/`_field_orders`/`_order_set`/`_scheduled` (and possibly `build_pyramid_block`) out to a new `src/zagg/pyramid.py` in this PR; **(b3)** something else.

**(c) Should the per-field `orders` key bump `PYRAMID_SPEC` to `zagg-pyramid/2`?** I left it at `/1`: the key is optional, absent by default, and a `/1` reader that ignores it still gets a correct *superset* answer for "which fields might be here" — but it would over-promise on a narrowed field, and §4.5's conformance rule is strict-check-and-fail-loudly on an unknown revision. **(c1)** stay `/1` (recommended — bumping would make every existing reader refuse every manifest, including those with no per-field schedules); **(c2)** bump to `/2`.

**(d) Pre-existing failures I did not touch** (CLAUDE.md §4: flag, don't fix):

- `uv run ruff check src tests` fails on `main` with `N818 Exception name UnknownCapability should be named with an Error suffix` (`src/zagg/registry.py:64`). The PR lint bot's ruleset is `--select=E,F,W,I` so CI does not see it, but the pyproject `select` includes `N`, so the plain local command is red on `main`.
- `uv run ruff format --check src tests` flags `tests/data/benchmark/README.md` on `main` (a python code fence).
- Relatedly: the two **existing** spec fixtures (`tests/data/spec/{minimal,kitchen_sink}/morton_hive.json`) carry a **pre-#201** pyramid block, `{"orders": [], "aggregation": {}}`, not the `zagg-pyramid/1` block `build_manifest` writes today. That staleness is independent of this PR — this PR does not change what the generator emits for them, since neither fixture config declares `output.pyramid` — and the new `pyramid/` fixture above is what closes the §4 obligation for the new grammar. Refreshing the two stale ones is left as its own change; say the word if you want it here instead.

---
_Generated by [Claude Code](https://claude.ai/code/session_01E5GA2bvsvSXYMEx4yjiC5t)_

### DIFF
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -893,7 +893,8 @@ gates (espg/moczarr#19/#20).
 
 Two tiny single-shard hive stores, both on the same deliberately small
 geometry — shard order 4, inner-chunk order 5, cell order 6 (16 cells,
-K = 4 inner chunks of 4 cells), sharded (the hive default):
+K = 4 inner chunks of 4 cells), sharded (the hive default) — plus one
+manifest-only fixture for the §4.5 declaration:
 
 - **`minimal/`** — one *unlocated* digest field (`h_tdigest`) plus `count`.
   The smallest thing that is a conforming store.
@@ -903,8 +904,20 @@ K = 4 inner chunks of 4 cells), sharded (the hive default):
   (including a single-photon cell packing the §3.1 golden word
   `0xFF000000FF0000FF` and a noise-only cell whose signal payload is the
   empty `(0, 2)` array), and `count`.
+- **`pyramid/`** — the §4.5 overview **declaration**, and nothing else: a
+  committed `morton_hive.json` carrying all four readings of a `fields` entry
+  in one block — a field that inherits the block schedule (`count`, no
+  `orders` key), one narrowed to a subset (`h_min`, `[3, 1]` of `[3, 2, 1]`),
+  one excluded outright but still declaring its fold law (`h_tdigest`,
+  `orders: []`), and a non-composable `"class": "none"` one (`h_mean`). The
+  block is a **template-time** artifact — it exists before the first leaf is
+  written and is decodable from the manifest alone — so this fixture carries
+  no store beneath it; a third leaf would pin nothing the other two do not.
+  Its sibling `pyramid.expected.json` records the declared knob and the
+  schedule each field resolves to under §4.5, so a reader's decoding of the
+  key is checkable without running zagg.
 
-Both stores pin the layout edge cases a reader must handle: inner chunk
+Both leaf stores pin the layout edge cases a reader must handle: inner chunk
 ordinal 2 is **empty** (absent from the shard index — the §1.5 sentinel, and
 that sparsity reaches the dense arrays too: the `morton` coordinate and
 `count` hold their fill across that chunk, so a reader MUST NOT assume the
@@ -912,7 +925,7 @@ coordinate is dense across a shard), populated chunks contain empty cells
 (the `b""` fill), and one cell's digest carries merged centroids (weight > 1)
 whose location words are common ancestors (§2.2).
 
-Both fixtures are **sharded**, so §7's conformance claim is scoped to the
+Both leaf fixtures are **sharded**, so §7's conformance claim is scoped to the
 §1.5 sharded geometry. The per-inner-chunk geometry — identical §1.4 framing,
 one object per inner chunk, no shard index — has no committed golden here; it
 is pinned zagg-side by
@@ -921,8 +934,8 @@ stored span from the array's own metadata, as §1.5 requires, reads both from
 one code path; a reader that hard-codes the shard-index suffix passes §7 and
 still fails on an unsharded store.
 
-Each fixture ships a sibling **`{name}.expected.json`** recording the shard
-key, leaf path, geometry, every populated cell's decoded values (digest
+Each leaf fixture ships a sibling **`{name}.expected.json`** recording the
+shard key, leaf path, geometry, every populated cell's decoded values (digest
 centroids, location words and composition words as decimal strings — JSON
 numbers cannot carry `uint64` faithfully), the per-stratum exact counts,
 and the §5 O11 `content_hashes` (per-array + combined). The expected
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -629,6 +629,16 @@ overview family under the versioned `pyramid` block:
   here?" question with zero opens and never a probe. Values outside the
   block's `orders` are not legal; a reader MAY ignore them (they name no
   overview the family walks).
+
+  **The declaration is authoritative over the objects.** Narrowing a schedule
+  does not delete the overviews written under the previous, wider one — like
+  shrinking `orders` itself, it leaves them as regenerable-cache debris (D9) —
+  so a stamped overview whose own `zagg_overview.fields` (§4.3) names a field
+  the manifest no longer schedules at that order MAY exist. A reader
+  encountering that contradiction MUST take the manifest's schedule as truth
+  and treat the extra array as stale; it is not evidence of a torn write, and
+  the array's *contents* remain a valid fold of whatever leaves existed when
+  it was written.
 - **`all_time`** — whether the `all.zarr` all-time fold is materialized at
   the declared orders (windowed stores only; a `schedule: none` store's
   single fold is already all-time).
--- a/docs/specification.md
+++ b/docs/specification.md
@@ -503,9 +503,11 @@ regionally heterogeneous resolution).
   **`fields` enumerates the materialized fields only** — exactly the fields
   present as arrays in *this* overview, each recording the fold that was
   actually applied. A `none`-class field is absent from the zarr (§4.4) and so
-  MUST be absent from this map: its recorded absence lives in the manifest's
-  `pyramid.overview.fields` (§4.5), which is the map that enumerates **every**
-  declared field. Consequently a reader MAY treat this map as the overview's
+  MUST be absent from this map — as is a field whose own schedule
+  (§4.5 `fields[…].orders`) does not list this overview's `order`, which is
+  why this map may differ between two overviews of the same product. Recorded
+  absence lives in the manifest's `pyramid.overview.fields` (§4.5), which is
+  the map that enumerates **every** declared field. Consequently a reader MAY treat this map as the overview's
   variable list and MUST be able to open every array it names; cross-checking
   it against the arrays present is a valid integrity check. Each entry carries
   at least `class` and `method`, and MAY carry further fold provenance — an
@@ -535,12 +537,15 @@ is the store's resolution axis, partially materialized). Concretely:
   `2^24` observations, and there `sum(weights)` is the nearest float32 to the
   true count rather than the count itself;
 - field inclusion is gated by the field's **composability class** (§4.5):
-  `exact` and `approximate` fields appear, `none` fields are **absent**.
+  `exact` and `approximate` fields appear, `none` fields are **absent**;
+- inclusion is additionally gated by the field's own **schedule** when it
+  declares one (§4.5 `fields[…].orders`): a field may carry overviews at a
+  subset of the block's orders, or at none of them.
 
-An overview's variable set may therefore be a *subset* of the leaf's —
-heterogeneous variable sets across level nodes are in contract, and a reader
-MUST NOT assume every leaf field exists at every overview order (the
-manifest declaration below is the zero-open way to know).
+An overview's variable set may therefore be a *subset* of the leaf's, and may
+differ **between orders** — heterogeneous variable sets across level nodes are
+in contract, and a reader MUST NOT assume every leaf field exists at every
+overview order (the manifest declaration below is the zero-open way to know).
 
 ### 4.5 The manifest `pyramid` block
 
@@ -560,7 +565,8 @@ overview family under the versioned `pyramid` block:
       "count":     {"class": "exact", "method": "sum", "nan_policy": "skip",
                     "dtype": "int32", "fill_value": 0},
       "h_tdigest": {"class": "approximate", "method": "tdigest_kway",
-                    "dtype": "float32", "inner_shape": [2], "delta": 512},
+                    "dtype": "float32", "inner_shape": [2], "delta": 512,
+                    "orders": [3]},
       "photon_ids": {"class": "none"}
     }
   }
@@ -599,6 +605,30 @@ overview family under the versioned `pyramid` block:
   metadata to know the overview array's form up front. This map is the
   **all-fields** view; the per-overview `zagg_overview.fields` attrs map
   (§4.3) is the materialized subset.
+- **`fields[…].orders`** (optional, `exact`/`approximate` entries only) — the
+  field's **own** overview schedule, a subset of the block's `orders` in the
+  same descending order. It exists because the folds are not equally cheap:
+  the exact folds are dense merges, while an `approximate` fold is a ragged
+  read plus a k-way merge per ancestor node, so a product may carry `count` at
+  every order and `h_tdigest` at every other one. Semantics:
+
+  - **absent** — the field carries an overview at **every** order in the
+    block's `orders`. This is the default and the pre-existing behavior; a
+    declaration with no per-field schedules is byte-identical to one written
+    before this key existed;
+  - **present** — the field carries an overview at exactly the listed orders
+    and at no others. `[]` therefore excludes it from the whole pyramid while
+    still recording its class and fold law (which is what distinguishes it
+    from a `"class": "none"` field: that one *cannot* fold, this one is not
+    *scheduled* to).
+
+  A `none` entry never carries `orders` — it is absent everywhere already, so
+  the key would be redundant. A reader MUST NOT read an order outside a
+  field's declared schedule as an incomplete sweep: like `"class": "none"`,
+  this is **recorded absence** (D24 option A) and answers the "is this field
+  here?" question with zero opens and never a probe. Values outside the
+  block's `orders` are not legal; a reader MAY ignore them (they name no
+  overview the family walks).
 - **`all_time`** — whether the `all.zarr` all-time fold is materialized at
   the declared orders (windowed stores only; a `schedule: none` store's
   single fold is already all-time).
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -2456,8 +2456,12 @@ def _validate_pyramid_fields(knob: dict, declared: dict, parent_order: int) -> N
     are near-free dense merges, so ``h_tdigest`` may run every 4 orders while
     ``count`` stays at every one. A per-field order outside the block schedule
     refuses rather than widening it: the sweep only ever walks the block's own
-    orders, so a widening entry would silently do nothing.
+    orders, so a widening entry would silently do nothing. Scheduling a
+    ``none``-class field refuses for the same reason — D24 excludes it from
+    every order regardless — while ``false``/``[]`` over one stays legal, since
+    that only states the exclusion the block already records.
     """
+    from zagg.semantics import field_composability
     from zagg.sweep_overview import pyramid_orders
 
     fields = knob["fields"]
@@ -2489,6 +2493,18 @@ def _validate_pyramid_fields(knob: dict, declared: dict, parent_order: int) -> N
             raise ValueError(
                 f"output.pyramid.fields.{name}.orders {extra} are outside the pyramid "
                 f"schedule {schedule} — a per-field schedule may only narrow it"
+            )
+        if not orders:
+            continue
+        try:
+            composable = field_composability(declared[name]) != "none"
+        except Exception:
+            continue  # malformed meta: the kind/shape validators own that error
+        if not composable:
+            raise ValueError(
+                f"output.pyramid.fields.{name}.orders schedules overviews for a "
+                f"non-composable field (D24 class 'none'), which is excluded from every "
+                f"order regardless — drop the entry, or write `false` to say so explicitly"
             )
 
 
--- a/src/zagg/config.py
+++ b/src/zagg/config.py
@@ -2370,7 +2370,10 @@ def get_pyramid(config: PipelineConfig) -> dict | None:
     store. An explicit mapping may carry ``spacing`` (int >= 1), ``orders``
     (explicit ancestor orders, winning over ``spacing``), ``all_time``
     (bool: also maintain the ``all.zarr`` cross-window fold on windowed
-    stores), and ``summarize`` (the D24 opt-in derived-summary declarations,
+    stores), ``fields`` (the issue #352 per-field overview schedules,
+    ``{field: {"orders": [...]}}`` or ``{field: false}`` — absent means every
+    composable field at every declared order, exactly the pre-#352 behavior),
+    and ``summarize`` (the D24 opt-in derived-summary declarations,
     ``{source_field: {"as": name, ...}}`` — recorded in the pyramid block,
     never the semantic core).
     """
@@ -2400,7 +2403,7 @@ def _validate_pyramid(config: PipelineConfig) -> None:
             "output.pyramid requires output.store_layout: hive (overviews live at "
             "hive-tree ancestor nodes; flat stores have no digit tree)"
         )
-    unknown = set(knob) - {"spacing", "orders", "all_time", "summarize"}
+    unknown = set(knob) - {"spacing", "orders", "all_time", "fields", "summarize"}
     if unknown:
         raise ValueError(f"output.pyramid has unknown keys {sorted(unknown)}")
     spacing = knob.get("spacing")
@@ -2418,6 +2421,8 @@ def _validate_pyramid(config: PipelineConfig) -> None:
     all_time = knob.get("all_time")
     if all_time is not None and not isinstance(all_time, bool):
         raise ValueError(f"output.pyramid.all_time must be a boolean (got {all_time!r})")
+    if knob.get("fields") is not None:
+        _validate_pyramid_fields(knob, get_agg_fields(config), parent_order)
     summarize = knob.get("summarize")
     if summarize is None:
         return
@@ -2438,6 +2443,52 @@ def _validate_pyramid(config: PipelineConfig) -> None:
                 f"output.pyramid.summarize.{source}.as = {target!r} collides with a "
                 f"declared field: derived summaries must use a DIFFERENT name so overview "
                 f"schema never silently differs from source (D24)"
+            )
+
+
+def _validate_pyramid_fields(knob: dict, declared: dict, parent_order: int) -> None:
+    """Validate ``output.pyramid.fields`` — the per-field schedules (issue #352).
+
+    Each entry is either ``false`` (exclude the field from every overview
+    order) or a mapping carrying ``orders``: a SUBSET of the block-wide
+    schedule, which is the cost lever the issue asks for — the t-digest fold
+    is a ragged read + k-way merge per ancestor node, while count/sum/min/max
+    are near-free dense merges, so ``h_tdigest`` may run every 4 orders while
+    ``count`` stays at every one. A per-field order outside the block schedule
+    refuses rather than widening it: the sweep only ever walks the block's own
+    orders, so a widening entry would silently do nothing.
+    """
+    from zagg.sweep_overview import pyramid_orders
+
+    fields = knob["fields"]
+    if not isinstance(fields, dict):
+        raise ValueError(f"output.pyramid.fields must be a mapping (got {fields!r})")
+    schedule = pyramid_orders(knob, parent_order)
+    for name, decl in fields.items():
+        if name not in declared:
+            raise ValueError(
+                f"output.pyramid.fields names unknown field {name!r} (declared: {sorted(declared)})"
+            )
+        if decl is False:
+            continue
+        if not isinstance(decl, dict):
+            raise ValueError(
+                f"output.pyramid.fields.{name} must be false or a mapping with 'orders' "
+                f"(got {decl!r})"
+            )
+        unknown = set(decl) - {"orders"}
+        if unknown:
+            raise ValueError(f"output.pyramid.fields.{name} has unknown keys {sorted(unknown)}")
+        orders = decl.get("orders")
+        if not isinstance(orders, list) or not all(isinstance(k, int) for k in orders):
+            raise ValueError(
+                f"output.pyramid.fields.{name}.orders must be a list of ints (got {orders!r})"
+            )
+        extra = sorted(set(orders) - set(schedule))
+        if extra:
+            raise ValueError(
+                f"output.pyramid.fields.{name}.orders {extra} are outside the pyramid "
+                f"schedule {schedule} — a per-field schedule may only narrow it"
             )
 
 
--- a/src/zagg/sweep_overview.py
+++ b/src/zagg/sweep_overview.py
@@ -192,6 +192,22 @@ def fold_digests(cell_digests: list, *, delta: int, dtype="float32") -> bytes:
 # ---------------------------------------------------------------------------
 
 
+def pyramid_orders(knob: dict, shard_order: int) -> list:
+    """The block-wide overview schedule an ``output.pyramid`` knob declares.
+
+    Explicit ``orders`` win over ``spacing``; absent both, the default is
+    every :data:`DEFAULT_SPACING` orders below the shard order (the
+    espg-ratified display schedule). Shared with
+    :func:`zagg.config._validate_pyramid` so a per-field schedule (issue #352)
+    is checked against the very list the sweep walks, never a second
+    derivation of it.
+    """
+    if knob.get("orders") is not None:
+        return sorted({int(k) for k in knob["orders"]}, reverse=True)
+    spacing = int(knob.get("spacing") or DEFAULT_SPACING)
+    return list(range(int(shard_order) - spacing, -1, -spacing))
+
+
 def build_pyramid_block(config, shard_order: int) -> dict:
     """The manifest ``pyramid`` block: the template-time declaration (D22).
 
@@ -210,10 +226,7 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     if knob is None:  # output.pyramid: false — declared off
         return {"spec": PYRAMID_SPEC, "overview": {"orders": []}}
     spacing = int(knob.get("spacing") or DEFAULT_SPACING)
-    if knob.get("orders") is not None:
-        orders = sorted({int(k) for k in knob["orders"]}, reverse=True)
-    else:
-        orders = list(range(int(shard_order) - spacing, -1, -spacing))
+    orders = pyramid_orders(knob, shard_order)
     agg = get_agg_fields(config)
     fields: dict = {}
     excluded = []
--- a/src/zagg/sweep_overview.py
+++ b/src/zagg/sweep_overview.py
@@ -224,7 +224,11 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     exclusion — so the per-(field, order) absence is recorded exactly as the
     block-wide one is: zero-open filtering, never a probe. Fields not named
     there carry no ``orders`` key and inherit the block schedule, so a config
-    without the knob declares byte-identically to pre-#352 zagg.
+    without the knob declares byte-identically to pre-#352 zagg. An empty
+    per-field schedule is warned about on the same footing as the D24 ``none``
+    exclusion below it — structurally the same outcome (a declared field that
+    is absent everywhere) — and excluding EVERY composable field says so
+    loudly: that declares a pyramid the sweep would never write.
     """
     from zagg.config import get_agg_fields, get_pyramid
     from zagg.semantics import EXACT_MERGE_LAWS, _fold_function_name, composability_classes
@@ -237,7 +241,8 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     agg = get_agg_fields(config)
     selection = knob.get("fields") or {}
     fields: dict = {}
-    excluded = []
+    excluded: list = []
+    deselected: list = []
     for name, cls in composability_classes(config).items():
         meta = agg[name]
         if cls == "exact":
@@ -265,7 +270,22 @@ def build_pyramid_block(config, shard_order: int) -> dict:
             continue
         if name in selection:
             entry["orders"] = _field_orders(name, selection[name], orders)
+            if not entry["orders"]:
+                deselected.append(name)
         fields[name] = entry
+    if deselected and orders:
+        every = len(deselected) == len([e for e in fields.values() if e.get("class") != "none"])
+        logger.warning(
+            f"pyramid: fields {deselected} are excluded from every declared order by "
+            f"output.pyramid.fields — recorded absence (issue #352), not a missing sweep"
+            + (
+                f"; that is EVERY composable field, so the declared orders {orders} would "
+                f"materialize nothing — `output.pyramid: false` is how a store declares no "
+                f"pyramid at all"
+                if every
+                else ""
+            )
+        )
     if excluded and orders:
         logger.warning(
             f"pyramid: fields {excluded} are non-composable (D24 class 'none') and will "
--- a/src/zagg/sweep_overview.py
+++ b/src/zagg/sweep_overview.py
@@ -264,7 +264,7 @@ def build_pyramid_block(config, shard_order: int) -> dict:
             excluded.append(name)
             continue
         if name in selection:
-            entry["orders"] = _field_orders(selection[name], orders)
+            entry["orders"] = _field_orders(name, selection[name], orders)
         fields[name] = entry
     if excluded and orders:
         logger.warning(
@@ -283,7 +283,7 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     return {"spec": PYRAMID_SPEC, "overview": overview}
 
 
-def _field_orders(decl, orders: list) -> list:
+def _field_orders(name: str, decl, orders: list) -> list:
     """One field's declared overview schedule (issue #352), as a subset of ``orders``.
 
     ``false`` is the outright exclusion (the empty schedule); a mapping
@@ -293,11 +293,29 @@ def _field_orders(decl, orders: list) -> list:
     makes the invariant unfalsifiable for the unvalidated caller
     (``declare_pyramid`` against a manifest whose ``shard_order`` differs from
     the config's).
+
+    That caller is exactly why the drop is WARNED and names the field: a
+    validated ``[6, 4]`` retrofitted onto a store whose schedule is
+    ``[7, 5, 3, 1]`` intersects to nothing, and the summary reports classes
+    only — so a field the user asked for at two orders would otherwise be
+    dropped from the whole pyramid and reported as success.
     """
     if decl is False:
         return []
     want = {int(k) for k in ((decl or {}).get("orders") or [])}
-    return [k for k in orders if k in want]
+    kept = [k for k in orders if k in want]
+    dropped = sorted(want - set(orders), reverse=True)
+    if dropped:
+        logger.warning(
+            f"pyramid: field {name!r} schedules orders {dropped} outside the block schedule "
+            f"{orders}; they are dropped"
+            + (
+                " — NOTHING remains, so the field is excluded from the entire pyramid"
+                if not kept
+                else f", leaving {kept}"
+            )
+        )
+    return kept
 
 
 def _json_fill(fill_value):
--- a/src/zagg/sweep_overview.py
+++ b/src/zagg/sweep_overview.py
@@ -218,6 +218,13 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     that gives readers zero-open filtering (option A, the ruled default) —
     and warned about loudly at template time, per the D24 ruling. The sweep
     later adds ``materialized`` actuals; it never rewrites the declaration.
+
+    A composable field named in ``output.pyramid.fields`` (issue #352) also
+    records its own ``orders`` — the narrowed schedule, ``[]`` for an outright
+    exclusion — so the per-(field, order) absence is recorded exactly as the
+    block-wide one is: zero-open filtering, never a probe. Fields not named
+    there carry no ``orders`` key and inherit the block schedule, so a config
+    without the knob declares byte-identically to pre-#352 zagg.
     """
     from zagg.config import get_agg_fields, get_pyramid
     from zagg.semantics import EXACT_MERGE_LAWS, _fold_function_name, composability_classes
@@ -228,12 +235,13 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     spacing = int(knob.get("spacing") or DEFAULT_SPACING)
     orders = pyramid_orders(knob, shard_order)
     agg = get_agg_fields(config)
+    selection = knob.get("fields") or {}
     fields: dict = {}
     excluded = []
     for name, cls in composability_classes(config).items():
         meta = agg[name]
         if cls == "exact":
-            fields[name] = {
+            entry = {
                 "class": "exact",
                 "method": EXACT_MERGE_LAWS[_fold_function_name(meta.get("function")) or ""],
                 "nan_policy": EXACT_NAN_POLICY,
@@ -242,7 +250,7 @@ def build_pyramid_block(config, shard_order: int) -> dict:
             }
         elif cls == "approximate":
             inner = meta.get("inner_shape") or (2,)
-            fields[name] = {
+            entry = {
                 "class": "approximate",
                 "method": TDIGEST_LAW,
                 "dtype": meta.get("dtype", "float32"),
@@ -250,8 +258,14 @@ def build_pyramid_block(config, shard_order: int) -> dict:
                 "delta": int((meta.get("params") or {}).get("delta", 512)),
             }
         else:
+            # A ``none`` entry carries ``class`` only (spec §4.5): it is already
+            # absent at every order, so a per-field schedule over it is moot.
             fields[name] = {"class": "none"}
             excluded.append(name)
+            continue
+        if name in selection:
+            entry["orders"] = _field_orders(selection[name], orders)
+        fields[name] = entry
     if excluded and orders:
         logger.warning(
             f"pyramid: fields {excluded} are non-composable (D24 class 'none') and will "
@@ -267,6 +281,22 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     if knob.get("summarize"):
         overview["summarize"] = {str(k): dict(v) for k, v in knob["summarize"].items()}
     return {"spec": PYRAMID_SPEC, "overview": overview}
+
+
+def _field_orders(decl, orders: list) -> list:
+    """One field's declared overview schedule (issue #352), as a subset of ``orders``.
+
+    ``false`` is the outright exclusion (the empty schedule); a mapping
+    narrows to its ``orders``. Always intersected with the block schedule and
+    kept in the block's descending order: :func:`zagg.config._validate_pyramid`
+    refuses a widening entry up front, so the intersection here only makes the
+    invariant unfalsifiable for the unvalidated caller (``declare_pyramid``
+    against a manifest whose ``shard_order`` differs from the config's).
+    """
+    if decl is False:
+        return []
+    want = {int(k) for k in ((decl or {}).get("orders") or [])}
+    return [k for k in orders if k in want]
 
 
 def _json_fill(fill_value):
--- a/src/zagg/sweep_overview.py
+++ b/src/zagg/sweep_overview.py
@@ -302,7 +302,9 @@ def _field_orders(name: str, decl, orders: list) -> list:
     """
     if decl is False:
         return []
-    want = {int(k) for k in ((decl or {}).get("orders") or [])}
+    want = _order_set(name, (decl or {}).get("orders") or [])
+    if want is None:  # unusable shape: inherit the block schedule, never narrow
+        return list(orders)
     kept = [k for k in orders if k in want]
     dropped = sorted(want - set(orders), reverse=True)
     if dropped:
@@ -316,6 +318,28 @@ def _field_orders(name: str, decl, orders: list) -> list:
             )
         )
     return kept
+
+
+def _order_set(name: str, sched) -> set | None:
+    """A declared per-field ``orders`` as a set of ints — ``None`` if unusable.
+
+    The shape is gated BEFORE iterating, because a ``str`` is iterable and its
+    digits parse: ``"13"`` would otherwise narrow the field to ``{1, 3}``, a
+    schedule nobody declared, instead of taking the fallback its callers
+    document. :func:`zagg.config._validate_pyramid_fields` refuses the same
+    shapes up front; this covers the callers that read a declaration back off
+    a manifest, which no validator has seen.
+    """
+    if isinstance(sched, list | tuple):
+        try:
+            return {int(k) for k in sched}
+        except (TypeError, ValueError):
+            pass
+    logger.warning(
+        f"pyramid: unusable per-field orders {sched!r} declared for field {name!r}; "
+        f"folding it at every declared order instead"
+    )
+    return None
 
 
 def _json_fill(fill_value):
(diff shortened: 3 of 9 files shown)


src/zagg/config.py
```diff
@@ -2441,6 +2446,52 @@ def _validate_pyramid(config: PipelineConfig) -> None:
...
+        if not isinstance(orders, list) or not all(isinstance(k, int) for k in orders):
+            raise ValueError(
+                f"output.pyramid.fields.{name}.orders must be a list of ints (got {orders!r})"
+            )
+        extra = sorted(set(orders) - set(schedule))
+        if extra:
+            raise ValueError(
+                f"output.pyramid.fields.{name}.orders {extra} are outside the pyramid "
+                f"schedule {schedule} — a per-field schedule may only narrow it"
+            )
```
espg, line 2496: 🤖 *from Claude (review)*

**A per-field schedule over a `none`-class field is silently dropped — the exact case the widening rule refuses.**

This validator accepts any name in `get_agg_fields(config)`, so `fields: {h_mean: {orders: [4]}}` passes. But `build_pyramid_block` short-circuits the `none` branch before the selection is consulted:

```python
        else:
            # A ``none`` entry carries ``class`` only (spec §4.5): it is already
            # absent at every order, so a per-field schedule over it is moot.
            fields[name] = {"class": "none"}
            excluded.append(name)
            continue
```

so the request evaporates with no error and no warning. That is inconsistent with the refusal five lines below this one, whose whole justification is "a widening entry would silently do nothing" — a *narrowing* entry on a non-composable field does nothing just as silently, and is a likelier user mistake (someone reads the issue's cost-lever framing and reaches for the field whose fold they think is expensive, e.g. `h_mean` on the shipped atl06 config, which is `none`).

Suggest refusing a **non-empty** `orders` on a `none`-class field, and continuing to allow `false`/`[]` there (redundant, but self-documenting and consistent with what the block already records for it). `zagg.semantics.field_composability` is already imported elsewhere in this module (`_warn_nan_ambiguous_reductions`), so the check is cheap.

---

docs/specification.md
```diff
@@ -599,6 +605,30 @@ overview family under the versioned `pyramid` block:
...
+    block's `orders`. This is the default and the pre-existing behavior; a
+    declaration with no per-field schedules is byte-identical to one written
+    before this key existed;
+  - **present** — the field carries an overview at exactly the listed orders
+    and at no others. `[]` therefore excludes it from the whole pyramid while
+    still recording its class and fold law (which is what distinguishes it
+    from a `"class": "none"` field: that one *cannot* fold, this one is not
+    *scheduled* to).
+
+  A `none` entry never carries `orders` — it is absent everywhere already, so
```
espg, line 625: 🤖 *from Claude (review)*

**The new reader rule needs the stale-artifact caveat the rest of the block already carries.**

"A reader MUST NOT read an order outside a field's declared schedule as an incomplete sweep" is right, but it is stated without saying what happens to overviews written under a *previous*, wider schedule. Narrowing `h_tdigest` from `[7, 5, 3]` to `[7]` does not delete the order-5/order-3 zarrs: `sweep_overviews` rewrites an overview only when it walks that node, and an order every field narrows out is now `continue`d past entirely — so a stamped overview whose `zagg_overview.fields` still names `h_tdigest` can outlive its declaration.

That is the correct D9 posture (overviews are regenerable caches; `declare_pyramid`'s own docstring already says "overviews at now-undeclared orders stay as regenerable-cache debris, D24 option A"), but a reader implementing §4.5 from this page alone will hit the contradiction — declaration says absent, the object on disk says present — and has no text telling it which wins. One sentence saying the declaration is authoritative and a wider-than-declared overview is debris would close it, and matches how the `orders` key already behaves when it shrinks.

---

src/zagg/sweep_overview.py
```diff
@@ -256,6 +283,22 @@ def build_pyramid_block(config, shard_order: int) -> dict:
...
 
+def _field_orders(decl, orders: list) -> list:
+    """One field's declared overview schedule (issue #352), as a subset of ``orders``.
+
+    ``false`` is the outright exclusion (the empty schedule); a mapping
+    narrows to its ``orders``. Always intersected with the block schedule and
+    kept in the block's descending order: :func:`zagg.config._validate_pyramid`
+    refuses a widening entry up front, so the intersection here only makes the
+    invariant unfalsifiable for the unvalidated caller (``declare_pyramid``
+    against a manifest whose ``shard_order`` differs from the config's).
```

**Docstring names a function that does not do the refusing.**

`:func:`zagg.config._validate_pyramid`` is the caller; the widening refusal this sentence relies on lives in `_validate_pyramid_fields`. Since the whole point of the sentence is to tell a future reader *where* the invariant is enforced (so they know why the intersection here is belt-and-braces rather than the primary guard), the wrong name is worse than no name.

---

tests/test_sweep_overview.py
```diff
@@ -468,6 +468,96 @@ def test_default_pyramid_validates_clean(self):
...
+
+    def test_field_schedule_never_widens_the_block(self):
+        """The unvalidated path (declare_pyramid) intersects, never widens."""
+        from zagg.sweep_overview import build_pyramid_block
+
+        # shard_order 4 => schedule [2, 0]; the knob asks for 5 as well.
+        cfg = self._cfg(pyramid={"fields": {"h_min": {"orders": [5, 2]}}})
+        overview = build_pyramid_block(cfg, shard_order=4)["overview"]
+        assert overview["orders"] == [2, 0]
+        assert overview["fields"]["h_min"]["orders"] == [2]
```
espg, line 570: 🤖 *from Claude (review)*

**Test gap: the `declare_pyramid` retrofit path never sees a per-field schedule.**

`build_pyramid_block` is covered directly and the sweep is covered end-to-end, but the third caller — `declare_pyramid`, the issue #358 retrofit (`python -m zagg.sweep <root> --declare-pyramid config.yaml`) — is not exercised with `output.pyramid.fields` at all. That path is not a thin wrapper over `build_pyramid_block`: it round-trips the block through `json.loads(json.dumps(block))` for its idempotency compare, validates every declared field against a real committed leaf in `_validate_block_against_store`, and preserves `materialized`. Two things worth pinning:

1. an excluded field (`orders: []`) is still `_field_drift`-checked against the leaf — it exists at native resolution, so a declaration that skipped it would be wrong, and a declaration that *refused* on it would be a regression;
2. `previous == "identical"` on a second call — a `[]` or a narrowed list must survive the JSON round-trip unchanged, or the headline idempotency property ("an identical existing declaration is not re-PUT") quietly breaks for every config using this knob.

`TestDeclarePyramid` already has the fixtures for this.

---

espg (commented):
🤖 *from Claude (review)*

Adversarial review of phases 1–4 against CLAUDE.md and issue #352's acceptance criteria.

**Note on provenance (important):** the Agent/Task tool is **not available** in this environment, so this review could not be run as the separate fresh-context Opus subagent CLAUDE.md §2 mandates. It is the weaker same-context fallback. Weight it accordingly — a same-context reviewer is blind to exactly the assumptions the author already made.

**What holds up.** The non-breaking guarantee is real and directly tested (`test_absent_field_knob_declares_identically`): fields the knob does not name carry no `orders` key, so an existing config's manifest is byte-identical. The per-order filter is in the right place — computed inside `for k in orders` and threaded through `_roll_node` → `_write_overview`, so a narrowed-out field drops out of the zarr *and* the `zagg_overview.fields` attrs map, which is what spec §4.3 requires of the materialized subset. `_content_hash` already digests `sorted(slabs)`, so a schedule change invalidates skip-if-current for free rather than leaving a half-stale overview reading as `current` — good that this is pinned by a test rather than left as an argument. The `none`-stays-class-only invariant (spec §4.5) survives. No new dependency; ruff/format clean on every touched file.

**Findings** (4 inline, none blocking the phase checklist):

1. `src/zagg/config.py` — a per-field schedule over a `none`-class field is accepted and then silently dropped, which is the same "silently does nothing" failure the widening refusal exists to prevent.
2. `docs/specification.md` — the new reader rule needs the stale-artifact caveat (a narrowed schedule does not delete the previously-written wider overview).
3. `src/zagg/sweep_overview.py` — `_field_orders` docstring points at the wrong validator.
4. `tests/test_sweep_overview.py` — the `declare_pyramid` retrofit path is never exercised with a per-field schedule, including the JSON round-trip its idempotency compare depends on.

**Explicitly *not* raised as findings** (checked, judged fine): the repeated `_scheduled` warning on a malformed declaration is noisy but correct and self-limiting; `spacing` still being recorded as the default when explicit `orders` are given is pre-existing behavior, untouched by this diff; the `int(x) for x in sched` leniency that makes `"orders": "1"` parse as `[1]` is harmless.

**Out of scope for the fold, standing for @espg:** the `src/zagg/sweep_overview.py` line count (1,227 on `main` → 1,303 here, both over CLAUDE.md §4's ~1,200) — raised as question (b) in the PR body, and a split is PR #349's to make, not this PR's.

---

src/zagg/config.py
espg: 🤖 *from Claude*

Folded in `c4aa32b`.

`_validate_pyramid_fields` now refuses a **non-empty** `orders` on a `none`-class field, and leaves `false`/`[]` legal there:

```python
        if not orders:
            continue
        try:
            composable = field_composability(declared[name]) != "none"
        except Exception:
            continue  # malformed meta: the kind/shape validators own that error
        if not composable:
            raise ValueError(
                f"output.pyramid.fields.{name}.orders schedules overviews for a "
                f"non-composable field (D24 class 'none'), which is excluded from every "
                f"order regardless — drop the entry, or write `false` to say so explicitly"
            )
```

The malformed-meta `continue` mirrors `_warn_nan_ambiguous_reductions`, which already declines to own that error class. Covered by `test_validate_refuses_a_schedule_on_a_none_class_field`, which pins both halves: `h_mean: {orders: [4]}` refuses on the shipped atl06 config, while `h_mean: false` and `h_mean: {orders: []}` still validate. Docstring updated to state the rule.

---

docs/specification.md
espg: 🤖 *from Claude*

Folded in `e066a9a`. Added the caveat as its own paragraph under `fields[….].orders` in §4.5:

> **The declaration is authoritative over the objects.** Narrowing a schedule does not delete the overviews written under the previous, wider one — like shrinking `orders` itself, it leaves them as regenerable-cache debris (D9) — so a stamped overview whose own `zagg_overview.fields` (§4.3) names a field the manifest no longer schedules at that order MAY exist. A reader encountering that contradiction MUST take the manifest's schedule as truth and treat the extra array as stale; it is not evidence of a torn write, and the array's *contents* remain a valid fold of whatever leaves existed when it was written.

The last clause is the part that was least obvious and most worth writing down: the stale array is not *corrupt*, so a reader has no cause to refuse the whole overview over it — it is simply outside what the declaration promises.

---

src/zagg/sweep_overview.py

Folded in `c65b775` — the docstring now names `zagg.config._validate_pyramid_fields`, which is where the widening refusal actually is.

---

tests/test_sweep_overview.py
espg: 🤖 *from Claude*

Folded in `aa6fc77` — `TestDeclarePyramid::test_retrofit_records_a_per_field_schedule`, pinning both points you named:

```python
        cfg.output["pyramid"] = {
            "orders": [1, 0],
            "fields": {"h_tdigest": {"orders": [1]}, "h_min": False},
        }
        summary = declare_pyramid(str(tmp_path), cfg)
        assert summary["updated"] is True and summary["orders"] == [1, 0]
        assert summary["validated"].startswith("leaf ")   # (1) excluded field still probed
        ...
        again = declare_pyramid(str(tmp_path), cfg)
        assert again["previous"] == "identical" and again["updated"] is False
        assert puts == []                                  # (2) no re-PUT
```

It also asserts `"orders" not in block["fields"]["count"]`, so the inherit-the-block-schedule case is pinned on the retrofit path too, not only at template time. Point (2) turned out to hold for a non-obvious reason worth recording: `_field_orders` returns a plain `list[int]` built from the block's own `orders`, so `json.loads(json.dumps(block))` is a fixed point — the same property `test_non_json_primitive_declaration_stays_idempotent` protects for `summarize`, now covered for this key as well. It passes as written (no code change needed).

---

espg:
🤖 *from Claude*

Run status — all four phases landed, review folded, CI green. Labelled `waiting`: three lettered questions in the PR body need an @espg call.

**Phases** (one commit each, per CLAUDE.md §3)

| | commit |
|---|---|
| 1 — config grammar + validation + the shared `pyramid_orders` helper | `10b5593` |
| 2 — `build_pyramid_block` records the per-field schedule | `849ac1e` |
| 3 — `sweep_overviews` per-order field filtering | `e50ae13` |
| 4 — `docs/specification.md` §4.3/§4.4/§4.5 + `docs/hive_layout.md` | `673ce1c` |

**Adversarial self-review — provenance caveat, please read.** CLAUDE.md §2 requires the review and the fold to run as two *separate fresh-context Opus subagents*. **The Agent/Task tool is not available in this environment**, so neither could be spawned. The review was run as the weaker same-context fallback: a same-context reviewer shares the author's blind spots by construction, so treat its clean bill on anything it did *not* flag as much weaker evidence than the usual two-subagent pass. The review/fold *discipline* was kept intact — the review was posted first as a `COMMENT` PR review with four inline findings and no edits, and only then folded, one commit per finding, staging only that finding's files, with a reply on every thread:

| finding | fold |
|---|---|
| a per-field schedule over a `none`-class field was silently dropped | `c4aa32b` |
| §4.5 needed the stale-artifact caveat (narrowing does not delete prior overviews) | `e066a9a` |
| `_field_orders` docstring named the wrong validator | `c65b775` |
| `declare_pyramid` retrofit path untested with a per-field schedule | `aa6fc77` |

All four inline threads are answered; none were declined, so nothing is left standing from the review itself. The `sweep_overview.py` line count was raised *by* the review as out of scope for the fold, and stands for @espg as question (b).

**Tests / lint.** Local `uv run pytest -q` on `aa6fc77`: **3208 passed, 37 skipped** (+12 new, all in `tests/test_sweep_overview.py`). CI on `aa6fc77` is green across the board — `ruff`, `test (3.12)`, `test (3.13)`, `build`, `build / build-x86_64`, `build / build-arm64`. `ruff format --check` is clean on every file this PR touches. No ruff-bot inline comments were posted on this PR (the ruff check ran as a workflow job and passed), so there are none to fix-or-reply to. Two pre-existing `main` failures are flagged, not fixed, under question (d) in the PR body.

**Deviation from CLAUDE.md §2, stated explicitly.** §2/§5 say a PR whose phases are all complete and which is awaiting review should be flipped draft → ready-for-review alongside the `waiting` label. **This routine's instructions override that and require the PR stay a draft**, so it stays a draft despite meeting the §2 conditions. Flagging it because the resulting state is otherwise ambiguous: draft here does **not** mean "still being worked" — it means the routine was told not to flip it. `gh pr ready` when you want it flipped.

**Resulting label state:** `implement` + `waiting`. No follow-on work is queued for the next run; the next action is an @espg decision on (a), (b), (c).

---

docs/specification.md
```diff
@@ -599,6 +605,40 @@ overview family under the versioned `pyramid` block:
   metadata to know the overview array's form up front. This map is the
   **all-fields** view; the per-overview `zagg_overview.fields` attrs map
   (§4.3) is the materialized subset.
+- **`fields[…].orders`** (optional, `exact`/`approximate` entries only) — the
```
espg, line 608: 🤖 *from Claude (review)*

**[for espg] — the fixture rationale is half right, and the half that is wrong is the half §4 exists for.**

I checked the claim in question (d) rather than taking it. Both parts of it verify:

```
$ python -c "import json;print(json.load(open('tests/data/spec/minimal/morton_hive.json'))['pyramid'])"
{'orders': [], 'aggregation': {}}
```

That is the pre-#201 shape; `hive.build_manifest` -> `build_pyramid_block` emits `{"spec": "zagg-pyramid/1", "overview": {...}}` today. So **(a) the fixtures are genuinely already stale**, independently of this PR. And **(b) this PR does not change what the generator would emit for them** — I ran `build_pyramid_block` on `origin/main` and on this branch against a no-`fields:` config and the JSON is byte-identical, which is exactly the fixture configs' situation (`tools/generate_spec_fixtures.py::_config` declares no `output.pyramid`).

Two things follow that the same-context review did not reach.

**(1) The staleness is invisible because nothing asserts it.** `tests/test_spec_conformance.py` contains no occurrence of the string `pyramid` — the fixtures' manifest pyramid block is never decoded, never compared, never spec-checked. That is *why* a block shape from before issue #201 survived this long. This is its own defect and worth its own issue: the conformance suite's premise is that spec, fixtures and reader cannot drift silently, and here they did.

**(2) The §4 obligation is still unmet, for a different reason than the PR body gives.** The PR body treats regeneration as the open question. It is not the interesting one — regenerating produces the same absence, because neither fixture config declares `output.pyramid` at all. The consequence is that after this PR, **`fields[…].orders` has zero fixture coverage**: an external reader that must "decode from the spec + fixtures alone" gets prose here and nothing to decode against. That is the substantive gap, and regenerating the two existing fixtures does not close it.

Plainly, on the adjudication: **shipping this spec-page change with unregenerated fixtures violates §4 as written** ("must update that page *and* the conformance fixtures ... in the same PR"), but the fix §4 actually wants is not `uv run python tools/generate_spec_fixtures.py` — it is a fixture whose config declares a pyramid with a narrowed schedule and an `orders: []` exclusion, so the new key appears in a committed `morton_hive.json` and in the `*.expected.json`.

Smallest concrete version: give `kitchen_sink`'s config an `output.pyramid` with `fields: {h_tdigest_signal: {orders: [...]}, h_tdigest_noise: false}`, and add the decoded pyramid block to the expected JSON so `test_spec_conformance` asserts it.

Tagging **for espg** rather than fold: it changes fixture scope, touches `tools/generate_spec_fixtures.py`, and entangles with the pre-existing staleness above, which wants a decision on whether the refresh lands here or separately.

---

src/zagg/sweep_overview.py
```diff
@@ -192,6 +192,22 @@ def fold_digests(cell_digests: list, *, delta: int, dtype="float32") -> bytes:
 # ---------------------------------------------------------------------------
 
 
+def pyramid_orders(knob: dict, shard_order: int) -> list:
```
espg, line 195: 🤖 *from Claude (review)*

**[for espg] — question (b)'s option (b1) rests on a false premise: PR #349 does not touch this file.**

The PR body offers "**(b1)** land as-is and let PR #349's split absorb it." PR #349 splits exactly five modules, and `sweep_overview.py` is not among them — its own before/after table lists `hive.py`, `processing/raster.py`, `sweep.py`, `config.py`, `runner.py`. More pointedly, PR #349's phase-3 rationale cites this file as an example of the pattern it is imitating, not as a target:

> Sibling modules also match the tree's existing convention: `zagg/sweep_overview.py` is already the split-out overview family.

So there is no pending split to defer to. Choosing (b1) does not park the overage behind another PR — it leaves `sweep_overview.py` at 1,304 lines (I measured `git show <branch>:src/zagg/sweep_overview.py | wc -l`) with nothing queued to bring it back down. That materially changes the question being asked, so it should be restated before it is answered.

**On whether the +76 could have gone elsewhere without contortion: yes.** The three additions are `pyramid_orders` (this hunk), `_field_orders` and `_scheduled` — all pure predicates over a declaration dict, with zero dependency on the sweep machinery (no store, no zarr, no manifest I/O, no `logger` except one warning). They are the natural seed of the `src/zagg/pyramid.py` option (b2) names, and extracting them would take the *declaration* helpers with them rather than splitting the file arbitrarily.

There is an independent argument for the same move that the PR does not mention: this PR introduces the **first runtime import from `zagg.config` into `zagg.sweep_overview`**:

```python
    from zagg.semantics import field_composability
    from zagg.sweep_overview import pyramid_orders
```

On `main`, `config.py` names `zagg.sweep_overview` only in a docstring — there is no import at all. Config validation now reaches into the sweep engine, and the reason the helper had to be shared is precisely that the schedule must be derived once. A neutral `zagg/pyramid.py` owning `pyramid_orders` would give both `config` and `sweep_overview` a common dependency instead of an inverted one. Note also that PR #349 phase 4 moves `_validate_pyramid` into `config/validate_output.py`, so this import lands there on merge — worth knowing when ordering the two.

Not folding this myself: splitting a module is a §6 side-effecting scope change and §4 says raise it. But the raise should be against the corrected premise.

---

src/zagg/sweep_overview.py
```diff
@@ -256,6 +283,23 @@ def build_pyramid_block(config, shard_order: int) -> dict:
...
+    kept in the block's descending order: the widening refusal lives in
+    :func:`zagg.config._validate_pyramid_fields`, so the intersection here only
+    makes the invariant unfalsifiable for the unvalidated caller
+    (``declare_pyramid`` against a manifest whose ``shard_order`` differs from
+    the config's).
+    """
+    if decl is False:
+        return []
+    want = {int(k) for k in ((decl or {}).get("orders") or [])}
+    return [k for k in orders if k in want]
```

**[fold] — the silent intersection can empty a field's whole schedule on exactly the path this docstring names, with no warning and no trace in the summary.**

```python
    want = {int(k) for k in ((decl or {}).get("orders") or [])}
    return [k for k in orders if k in want]
```

The docstring says the intersection "only makes the invariant unfalsifiable for the unvalidated caller (`declare_pyramid` against a manifest whose `shard_order` differs from the config's)". That caller exists, and on it the intersection is not a safety net — it is silent data loss. Concrete path:

- Config: `output.grid.parent_order: 8`, `output.pyramid: {spacing: 2, fields: {h_tdigest: {orders: [6, 4]}}}`.
- `_validate_pyramid` derives `pyramid_orders(knob, parent_order=8)` = `[6, 4, 2, 0]`; `[6, 4]` is a subset, so **validation accepts** (I ran it — accepted, block records `h_tdigest.orders == [6, 4]`).
- Retrofit that config onto a store whose manifest says `shard_order: 9`. `declare_pyramid` calls `build_pyramid_block(config, 9)` → `pyramid_orders(knob, 9)` = `[7, 5, 3, 1]`.
- `_field_orders({"orders": [6, 4]}, [7, 5, 3, 1])` → `[]`. **The field is now excluded from the entire pyramid.**

Nothing catches it on the way out:

- the range guard cannot fire, since every derived order is in `[0, shard_order)`:
  ```python
  bad = [k for k in block["overview"].get("orders") or [] if not 0 <= int(k) < shard_order]
  ```
- `_semantic_guard` is scoped to the aggregation core and its own docstring says so — "`output.*` is not in the semantic core, so the intended retrofit config ... hashes identically";
- the returned summary reports `{"fields": {n: m.get("class") ...}}` — classes only, so `h_tdigest` still reads as `"approximate"` and the emptied schedule is invisible.

The user asked for a t-digest at two orders and gets one at none, reported as success. That is the opposite posture from `declare_pyramid`'s documented contract ("this is the user's explicit operation: every failure raises") and from `_validate_pyramid_fields`, which refuses a *widening* entry for the much milder reason that it "would silently do nothing".

Suggested fold: in `build_pyramid_block` (or `_field_orders`, passing the field name), compare the requested set against the intersection and `logger.warning` — or raise from `declare_pyramid` — when a non-empty request narrows to fewer orders than were named. A test at `TestDeclarePyramid` with a config whose `parent_order` differs from the store's `shard_order` would pin it; `test_retrofit_records_a_per_field_schedule` currently only exercises the matching case, and `test_field_schedule_never_widens_the_block` asserts the silent narrowing as intended rather than probing whether it should be silent.

---

src/zagg/sweep_overview.py
```diff
@@ -709,6 +765,27 @@ def sweep_overviews(store_root: str, manifest: dict, by_shard: dict, *, store_kw
...
+    No ``orders`` key is the pre-#352 default — the field runs at every order
+    the block declares; a present one is its own narrowed schedule (``[]``
+    excludes it everywhere). Unparseable entries fall back to the default
+    rather than silently dropping a field the declaration promised.
+    """
+    sched = meta.get("orders")
+    if sched is None:
+        return True
+    try:
+        return int(k) in {int(x) for x in sched}
```

**[fold] — a string `orders` silently narrows instead of taking the documented fallback.**

```python
    sched = meta.get("orders")
    if sched is None:
        return True
    try:
        return int(k) in {int(x) for x in sched}
    except (TypeError, ValueError):
```

The docstring promises "Unparseable entries fall back to the default rather than silently dropping a field the declaration promised." A `str` defeats that promise, because it is iterable and its digits parse:

- manifest carries `"orders": "13"` (hand-edited, or a foreign writer that took the spec's list loosely);
- `{int(x) for x in "13"}` → `{1, 3}`, no exception;
- `_scheduled` therefore returns `True` at orders 1 and 3 and **`False` everywhere else**.

The field is silently narrowed to a schedule nobody declared — the exact failure mode the `except` branch exists to prevent, and the one the field the sweep is about to drop cannot recover from. `test_unparseable_field_schedule_folds_everywhere` uses a scalar `1`, which does raise `TypeError`, so this variant is uncovered.

Fold: gate on shape before iterating, e.g.

```python
    if not isinstance(sched, (list, tuple)):
        ...warn, return True
```

Two smaller things in the same block while it is being touched:

- the warning does not name the field — `f"unusable per-field orders {sched!r} in the pyramid declaration"` leaves the operator to guess which of N declared fields is misdeclared, and the call site (`{n: m for n, m in fields.items() if _scheduled(m, k)}`) has `n` in hand;
- the same shape hole exists in `_field_orders` (`{int(k) for k in (... .get("orders") or [])}`), reachable on the `declare_pyramid` path, which does not run `validate_config`. Worth the same guard.

---

src/zagg/sweep_overview.py
```diff
@@ -210,17 +233,15 @@ def build_pyramid_block(config, shard_order: int) -> dict:
     if knob is None:  # output.pyramid: false — declared off
         return {"spec": PYRAMID_SPEC, "overview": {"orders": []}}
     spacing = int(knob.get("spacing") or DEFAULT_SPACING)
-    if knob.get("orders") is not None:
-        orders = sorted({int(k) for k in knob["orders"]}, reverse=True)
-    else:
-        orders = list(range(int(shard_order) - spacing, -1, -spacing))
+    orders = pyramid_orders(knob, shard_order)
     agg = get_agg_fields(config)
+    selection = knob.get("fields") or {}
```
espg, line 242: 🤖 *from Claude (review)*

**[fold] — a config that excludes every field declares a pyramid and builds nothing, with no warning, while the analogous D24 case eight lines below warns loudly.**

I ran this against the branch:

```yaml
output:
  pyramid:
    orders: [7, 5, 3, 1]
    fields: {count: false, h_tdigest: false}
```

`validate_config` **accepts** it. The block it produces declares `"orders": [7, 5, 3, 1]` with `"orders": []` on every field. The sweep then hits `if not order_fields: continue` at all four orders, returns `{"written": 0, ..., "declared": True}`, never reaches `_update_manifest_pyramid`, and the only signal anywhere is four `logger.info` lines.

The inconsistency is with the code immediately below this hunk, which handles the *structurally identical* situation — a declared pyramid whose fields will all be absent — at warning level:

```python
    if excluded and orders:
        logger.warning(
            f"pyramid: fields {excluded} are non-composable (D24 class 'none') and will "
            ...
```

`excluded` is populated only in the `none`-class branch, so the issue #352 spelling of the same outcome gets nothing. A user who meant `output.pyramid: false` and instead excluded every field by hand gets a store that claims a four-order pyramid and materializes none of it, silently.

Fold: either extend that warning to cover fields whose `_field_orders` result is empty at template time (cheap — `selection` is right here), or refuse in `_validate_pyramid_fields` when every composable field is excluded, pointing at `output.pyramid: false` as the intended spelling. A `TestPyramidBlock` case asserting the warning (or the refusal) would pin whichever is chosen; neither is covered today.

For the record, the neighbouring grammar checks are otherwise consistent and I could not break them: unknown keys still error, an unknown field name refuses, `true` refuses, `orders: 4` refuses, `{}` refuses on the missing key, a `none`-class field with a non-empty schedule refuses while `false` / `{orders: []}` over one stays legal, and `false` vs `{orders: []}` genuinely are two spellings of one thing (both emit `"orders": []`) — which §4.5 documents, so that is a deliberate synonym rather than a fork.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of PR #369 against issue #352 and CLAUDE.md. Five inline findings: two **[for espg]**, three **[fold]**. The core mechanism is sound — I tried to break the grammar and the per-order filter and mostly could not — so most of what follows is about the edges and about two claims in the PR body that do not survive checking.

**The three adjudications asked for.**

**(1) Spec fixtures.** The "already stale" rationale is **true** — I verified `tests/data/spec/*/morton_hive.json` carries the pre-#201 `{"orders": [], "aggregation": {}}` while `build_manifest` emits `zagg-pyramid/1` — and it is **true** that this PR would not change generator output for those two fixtures (I ran `build_pyramid_block` on both refs against a no-`fields:` config: byte-identical). But the conclusion drawn from it is wrong. §4 is violated, just not in the way the PR body considers: after this change **`fields[…].orders` has no fixture coverage at all**, because neither fixture config declares `output.pyramid` — so regenerating would close nothing, and an external reader decoding "from the spec + fixtures alone" gets prose and no artifact. The real ask is a pyramid-declaring fixture. Separately, the staleness went unnoticed because `tests/test_spec_conformance.py` never mentions `pyramid` — that is its own defect. Inline on `docs/specification.md`.

**(2) Module size.** Continuing was **not** defensible on the reasoning given, because the reasoning is factually wrong: option (b1) says PR #349's split will absorb this file, and PR #349 does not include `sweep_overview.py` — it lists five other modules and cites this one as the *precedent* it is imitating. And **yes, the ~76 lines had a natural home**: `pyramid_orders` / `_field_orders` / `_scheduled` are pure predicates over a declaration dict with no sweep dependency, and this PR introduces the first-ever runtime import from `zagg.config` into `zagg.sweep_overview` to share one of them — which is itself the argument for a neutral module. Inline on `sweep_overview.py`.

**(3) Manifest back-compat.** The byte-identity claim **holds** — I diffed `build_pyramid_block` output across `origin/main` and this branch for a no-`fields:` config and it is identical, key order included. Reader side: `orders: []` does give correct zero-open filtering, and an old `/1` reader degrades **safely, not silently** — §4.3 already tells readers the overview's own `zagg_overview.fields` attrs map is the materialized subset (a `"class": "none"` field forced that rule long before this PR), so a reader keyed there sees the truth, and one keyed on the manifest's all-fields map gets a failed open rather than a wrong answer. On question (c): **staying at `zagg-pyramid/1` is right** — bumping would make every existing reader refuse every manifest, including ones with no per-field schedules, to fix an over-promise that already resolves to a loud failure.

**Elsewhere.** Per-order narrowing itself checks out: the field set is recomputed inside the `for k in orders` loop and threaded through `_roll_node` -> `_fold_node` / `_write_overview`, so a narrowed-out field is absent from the zarr *and* from the attrs map; `_field_provenance` rebuilds entries and does not leak `orders` into the overview attrs; `_content_hash` feeds each field *name* into the digest before its bytes, so dropping a field really does change the hash and force regeneration (the PR's claim, and `test_narrowed_schedule_regenerates_a_stale_overview` pins it); `emit_shard_template(overwrite=True)` means a rewritten order does not keep the dropped array as debris. Skipping an order whole is safe because §4.5's new "declaration is authoritative over the objects" paragraph covers the residue, and `materialized` is a union that is explicitly a convenience actual. Unsorted and duplicate per-field orders normalise correctly through the intersection. Validation and the sweep do derive one schedule via `pyramid_orders` — with the one caveat that the *inputs* can differ (`config.output.grid.parent_order` vs the manifest's `shard_order`), which is finding (3) inline. Tests are behavioral, not restatement; the gaps I found are named in the inline threads. No new dependency, no `# noqa`, no `# type: ignore`, no weakened tests, docstrings match the neighbours' voice.

**Verdict.** The implementation is good and the grammar choice in question (a) is well argued — the issue's own "`count` at every order but `h_tdigest` only every 4" really is inexpressible as an `exclude:` list, so (a1) is the right call on the merits. Three foldable defects and two premises that need restating before espg answers (b) and (d). Nothing here is a reason to unwind the approach.

---

src/zagg/sweep_overview.py

Folded in `59aafc2`. `_field_orders` now takes the field name and warns when a request falls outside the block schedule, loudest when the intersection empties:

```python
    want = _order_set(name, (decl or {}).get("orders") or [])
    ...
    kept = [k for k in orders if k in want]
    dropped = sorted(want - set(orders), reverse=True)
    if dropped:
        logger.warning(
            f"pyramid: field {name!r} schedules orders {dropped} outside the block schedule "
            f"{orders}; they are dropped"
            + (
                " — NOTHING remains, so the field is excluded from the entire pyramid"
                if not kept
                else f", leaving {kept}"
            )
        )
```

Warning rather than raising, deliberately: `build_pyramid_block` is also the template-time path, and it is reached with configs that never went through `validate_config` (the fixture generator and several tests build one directly), so a raise there would refuse declarations that are legal today for a reason unrelated to this key. The warning names the field and the exact orders, which is what the summary could not tell you.

Two tests pin it, one per branch of the message:

- `TestDeclarePyramid::test_retrofit_warns_when_a_field_schedule_empties` — your path exactly. `{"spacing": 2, "fields": {"h_tdigest": {"orders": [1]}}}` is what a `parent_order: 3` config validates as (schedule `[1]`); retrofitted onto the fixture store's `shard_order: 2` the block schedule is `[0]`, the recorded `orders` is `[]`, and the summary still reports `h_tdigest: "approximate"` — so the assertion is that the warning is present, since it is the only signal;
- `TestPyramidBlock::test_field_schedule_never_widens_the_block` — extended to assert the partial-drop case names `'h_min'` and `[5]`, so the previously-silent intersection that test asserted is now asserted to be audible.


---

src/zagg/sweep_overview.py

Folded in `f4e534b`. The shape gate is now a shared helper both call sites use, so the hole cannot exist in one and not the other:

```python
def _order_set(name: str, sched) -> set | None:
    if isinstance(sched, list | tuple):
        try:
            return {int(k) for k in sched}
        except (TypeError, ValueError):
            pass
    logger.warning(
        f"pyramid: unusable per-field orders {sched!r} declared for field {name!r}; "
        f"folding it at every declared order instead"
    )
    return None
```

`_scheduled(name, meta, k)` and `_field_orders(name, decl, orders)` both take the field name now and return the documented fallback when it returns `None` — `True` at every order, and the full block schedule respectively (never a narrowed one).

On the validation half: a string was already refused there — `_validate_pyramid_fields` gates on `isinstance(orders, list)` before anything else, so `{"orders": "40"}` raises `output.pyramid.fields.h_min.orders must be a list of ints`. It was untested, which is why it read as open; `test_validate_field_schedules` now pins it explicitly with a comment saying why a string in particular needs a shape check.

Three tests:

- `TestPyramidBlock::test_unusable_field_schedule_inherits_the_block` — `{"orders": "40"}` at `shard_order=6` records `[4, 2, 0]`, the whole block schedule, and explicitly *not* `[4, 0]` from the string's digits;
- `TestOverviewWriter::test_string_field_schedule_folds_everywhere` — the sweep-side twin of your example: a manifest carrying `"orders": "3"` while order 1 is what is swept. Digit-wise that reads as `{3}` and drops `h_min` from the only order there is; the field lands;
- `test_unparseable_field_schedule_folds_everywhere` — extended to assert the warning names `'h_min'`, which it did not before.


---

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Folded in `99b8b6d`, taking the warn option rather than the refusal — an all-excluded pyramid is a legal declaration (it still records every field's class and fold law, which is a useful artifact), so the defect was the silence, not the state. It now sits immediately above the `none`-class warning it was inconsistent with:

```python
    if deselected and orders:
        every = len(deselected) == len([e for e in fields.values() if e.get("class") != "none"])
        logger.warning(
            f"pyramid: fields {deselected} are excluded from every declared order by "
            f"output.pyramid.fields — recorded absence (issue #352), not a missing sweep"
            + (
                f"; that is EVERY composable field, so the declared orders {orders} would "
                f"materialize nothing — `output.pyramid: false` is how a store declares no "
                f"pyramid at all"
                if every
                else ""
            )
        )
```

`deselected` collects any field whose `_field_orders` came back empty, so a single `orders: []` warns too (symmetric with the D24 case, which also warns for one field); the tail fires only when nothing composable is left, and names `output.pyramid: false` as the spelling the user probably meant.

Two `TestPyramidBlock` tests: `test_excluded_fields_warn_like_the_none_class_ones` (one field — warns, and asserts the whole-pyramid tail is *not* present) and `test_excluding_every_composable_field_warns` (your `fields: {count: false, h_min: false, h_max: false}` on the atl06 config — the block still declares `[4, 2, 0]`, and the warning carries both `EVERY composable field` and `output.pyramid: false`).

Thanks for the negative results in the last paragraph — I did not re-probe those, on the reviewer/folder split.


---

docs/specification.md
espg: 🤖 *from Claude*

Split in two: the new-grammar coverage is folded in `81a8540`; the pre-existing staleness is not touched and stands for espg.

**Folded — `tests/data/spec/pyramid/`, a third fixture.** You are right that regenerating the two existing ones closes nothing, because neither config declares `output.pyramid`. So the fixture is a new one whose config does. It is **manifest-only**: the pyramid block is a template-time artifact — written by `hive.build_manifest` → `build_pyramid_block` before a single leaf exists, and decodable from `morton_hive.json` alone — so a third leaf store would have pinned nothing `minimal/` and `kitchen_sink/` do not already pin. The committed block carries all four readings a decoder has to tell apart at once:

```json
"count":     {"class": "exact", "method": "sum", ...},                       // no orders key: inherits [3, 2, 1]
"h_min":     {"class": "exact", "method": "min", ..., "orders": [3, 1]},     // narrowed
"h_tdigest": {"class": "approximate", "method": "tdigest_kway", ..., "orders": []},  // excluded, fold law intact
"h_mean":    {"class": "none"}                                              // recorded absence, class only
```

I went with a fresh config rather than your `kitchen_sink` suggestion for a reason worth recording: on that config `h_tdigest_signal`/`h_tdigest_noise` are **located** ragged fields, which `field_composability` classes as `none` — so `fields: {h_tdigest_signal: {orders: [...]}}` is exactly the declaration `c4aa32b` now refuses, and the block would have recorded `{"class": "none"}` with no schedule at all. The `pyramid/` config uses unlocated `h_tdigest` (approximate) plus `nanmin`/`mean` to get an `exact` and a `none` alongside it.

`pyramid.expected.json` records the declared knob and the schedule each field resolves to, derived in the generator from the knob by the §4.5 rules — never read back out of the written block, same discipline as the cell values. `tests/test_spec_conformance.py` gains `TestPyramidDeclaration` (5 tests) with a `_declared_orders()` decoder written from §4.5 text alone: none-class → absent everywhere, absent key → inherit, present → exact. §7 of the spec page documents the fixture.

**Not folded — the pre-existing staleness of `minimal/` and `kitchen_sink/`.** Left exactly as it was, per your own framing: it predates this PR, this PR does not change what the generator emits for those two, and refreshing them would put an unrelated fixture-regeneration diff (including their `semantic_hash`/`generated_at` churn) into this one. Question (d) in the PR body is updated to say so precisely rather than treating regeneration as the open question. Your point that the conformance suite never decodes the manifest pyramid block at all — which is *why* it drifted — is now partly answered for the new key by `TestPyramidDeclaration`, but the two stale fixtures still have no assertion over their own block; whether that refresh plus a guard lands as its own change is an espg call.


---

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Not folded — **stands for espg.** Relocating the helpers or splitting the module is a scope change under §6, so it is not mine to take; the fold's job here is only to make sure the question is asked on true facts.

**The factual error is corrected in the PR body.** Question (b) no longer claims a split "is already proposed under PR #349". It now states that PR #349 splits five other modules (`hive.py`, `processing/raster.py`, `sweep.py`, `config.py`, `runner.py`) and cites this file as the convention it imitates — quoting its phase-3 line — so option (b1) parks nothing behind another PR. Both of your supporting facts are carried into the question as well: the three helpers are pure declaration predicates with no sweep-machinery dependency, and this PR introduces the first runtime `zagg.config` → `zagg.sweep_overview` import, which a neutral `zagg/pyramid.py` would invert back. The (b2) option now names `_order_set` too.

One number changed under you: today's four folds land in this same file, so it is **1,361** lines now, not 1,303 (+134 over `main`'s 1,227). The body is updated. I kept the fixes here rather than seeding `zagg/pyramid.py` with them precisely because creating that module *is* option (b2) — doing it as a side effect of a fold would have pre-empted the decision this thread exists to get. The three folds are 12–25 lines each and all sit inside the functions your finding is about, so whichever way (b) goes, they move with them.


---

espg:
🤖 *from Claude*

Second adversarial review folded. Four commits, one per finding, each staging only the files that finding touches:

| Finding | Commit |
| --- | --- |
| `_field_orders` silently emptied a field's whole schedule on the `declare_pyramid` retrofit path — now warns, naming the field and the dropped orders ([thread](https://github.com/englacial/zagg/pull/369#discussion_r3703308555)) | `59aafc2` |
| `orders` was iterated before its shape was checked, so a `str` narrowed digit-wise (`"13"` → `{1, 3}`) instead of taking the fallback — one shared `_order_set` gate now covers both `_scheduled` and `_field_orders`, and the warning names the field ([thread](https://github.com/englacial/zagg/pull/369#discussion_r3703310690)) | `f4e534b` |
| A pyramid with every field excluded warned nothing, while the structurally identical D24 `none` case warned — both warn now, and the all-excluded case names `output.pyramid: false` as the intended spelling ([thread](https://github.com/englacial/zagg/pull/369#discussion_r3703312501)) | `99b8b6d` |
| §4 fixture coverage for the new grammar: a new manifest-only `tests/data/spec/pyramid/` fixture + `TestPyramidDeclaration` in `tests/test_spec_conformance.py`, which previously never mentioned `pyramid` at all ([thread](https://github.com/englacial/zagg/pull/369#discussion_r3703304270)) | `81a8540` |

Each fold carries its own tests — 8 new ones across `TestPyramidBlock`, `TestOverviewWriter`, `TestDeclarePyramid`, plus the 5 conformance tests.

**Standing for review, not folded:**

- **Module size** ([thread](https://github.com/englacial/zagg/pull/369#discussion_r3703306695)) — relocating the helpers or splitting `sweep_overview.py` is a scope change (§6), so it stays question (b). The review established that the old framing was wrong on its facts, so the PR body's question (b) is **corrected**: PR #349 does not split this file (it splits five others and cites this one as precedent), so option (b1) defers to nothing. The line count is also updated — today's folds put the file at 1,361, not 1,303.
- **The two existing spec fixtures are independently stale** (`{"orders": [], "aggregation": {}}`, pre-#201). Untouched on purpose: it predates this PR, this PR does not change what the generator emits for them, and the new `pyramid/` fixture is what closes the §4 obligation for the new key. Question (d) says so precisely now. Whether that refresh lands as its own change is an espg call.

**Local gate:** `uv sync --extra test`, then `uv run pytest -v` → **3219 passed, 37 skipped**; `uv run ruff check src tests` and `uv run ruff format --check src tests` clean apart from the two pre-existing failures already flagged in question (d) (`N818` on `src/zagg/registry.py:64`, and the `README.md` code fence), which are untouched.

**Labels:** staying `implement` + `waiting` — all phases and all review findings are done, but questions (a) the grammar fork, (b) module size, and (c) the `zagg-pyramid` spec-marker bump each need a call. Left in **draft** for the same reason.


---

espg:
🤖 *from Claude* (closing at espg's direction, 2026-08-05)

Closed **unmerged** — superseded by the pyramid v2 design (#381), not by any defect in this PR.

The ruling that retires it is #381 point (8): **no per-field opt-out** — if a field ships and is mergeable, it is merged; the only exclusion stays class-driven (D24). The cost asymmetry that motivated #352's `fields:` lever is dissolved structurally by the leaf-push architecture (#381 points (1)–(3)): the expensive t-digest folds happen once, resident at the leaf worker, so there is no per-field cost to steer. The `orders`-schedule surface this PR extends is itself being replaced by the `(node, cells)` level grammar (#382, `zagg-pyramid/2`).

Salvage: two pieces of this branch remain useful to the #382 work and are flagged to it — the `pyramid_orders` schedule-derivation extraction (one shared derivation for validator and sweep) and the `pyramid/` conformance-fixture scaffolding (`tools/generate_spec_fixtures.py` → `tests/data/spec/pyramid/` + `TestPyramidDeclaration`). The branch `claude/352-pyramid-field-selection` is left in place as the reference.

The lettered questions in the body are moot with the close, for the record: (a) neither grammar ships; (b) the `sweep_overview.py` line count concern transfers to the #382 plan (new code goes to a new `src/zagg/pyramid.py`); (c) no `/1` bump — `/2` arrives with #382; (d) the pre-existing failures stay flagged on `main`.

PR #371, opened 2026-08-03
spill: cross-block merge for located / strata / composition fields (issue #370)

### LINKED ISSUE #370: spill: cross-block merge for located / strata / composition fields (lift the single-block restriction)
🤖 *from Claude*

## Problem

`aggregation.streaming: {mode: spill}` is exact for **every** reducer in the single-block regime, but a shard whose spilled bytes cross the block threshold raises `SpillOverflowError` when the config carries located ragged fields, `build_tdigest_where` strata, or `composition` — `SpillAggregator` classifies them non-mergeable (`src/zagg/processing/spill.py:432`) because `validate_streaming` (`src/zagg/processing/streaming.py:92`) has no cross-block law wired for them. The densest production shards (88S crossover belt; the fat-tail Sierra shards on a California o8 strata run) are exactly the ones that both want strata+location *and* risk a block close.

The blocking is plumbing, not algebra:

- `merge_tdigests` **already implements the located pairwise merge** (`locations1=`/`locations2=` overloads, `src/zagg/stats/tdigest.py:328`) — centroids that combine fold their morton words via `common_ancestor`.
- Stratum membership (`build_tdigest_where`) is per-photon row selection *before* the build — independent of block boundaries — so per-block strata digests merge like any digest.
- `merge_composition` exists with documented error bounds (lanes quantized to 1/510; folds are not byte-stable across orders — `src/zagg/stats/composition.py`).

## Scope

1. **Spill schema**: when any field declares `location:`, append the per-photon order-29 point-word column (uint64, +8 B/photon of `/tmp`) to the spill record — the worker already computes these for cell assignment.
2. **Block close** (`_fold_block`): located fields build `(digest, locations)` partials per cell via the same `build_tdigest(values, locations=...)` call the pooled path uses; strata fields evaluate their `where` per block over the already-spilled confidence columns.
3. **Cross-block fold**: wire the existing located `merge_tdigests`; k-way vs pairwise follows the field's function exactly as the plain-digest fold does today (#279).
4. **Composition fold**: use `merge_composition` — **needs a ruling** (see below).
5. **Validation split**: spill's mergeability probe widens; `mode: merge` keeps rejecting located (it folds every `buffer_granules` flush, degrading locations continuously; spill folds only on the rare block close).
6. **Spec**: one sentence in `docs/specification.md` (digest payload / located channel) naming block closes as an additional merge source. No wire-format change — cross-block-folded payloads are already conformant `zagg-ragged/1` + locations, so no fixture change.
7. **Semantics**: within the existing contract — a fold pushes affected cells past the compression knee, coarsening merged centroids' locations to common ancestors, which is the located channel's defined behavior above weight 1 (the 88S regime). Overflow shards trade location resolution instead of failing.

## Open decision (espg)

**Composition multi-block fold** — the question #321 deferred: `merge_composition`'s lane quantization means an overflow shard's composition word is *not byte-stable* against the single-block result (error O(n/510) per fold). Options: (a) accept the documented error on overflow shards (composition stays useful everywhere), or (b) leave composition non-mergeable so a strata-with-composition config still fails loudly on a block close. (a) is the working assumption; flagged in the PR's "Questions for review".

## Refs

#217 (spill), #87 (located channel), #279 (merge order), #321/#324 (strata, located ladder), #340 (spec discipline). Surfaced scoping this week's demo (California o8 strata+located on the 4096-disk tier is plausibly within ~2× of the block threshold on its densest shards).

espg:
Closes #370

## What

Lifts the single-block restriction for located ragged fields, `build_tdigest_where` strata, and the packed composition word on the spill path (`aggregation.streaming.mode: spill`). Today `SpillAggregator` classifies these non-mergeable (`_mergeable` via bare `validate_streaming`) and raises `SpillOverflowError` the moment a second block opens — exactly on the densest shards (88S crossover belt, fat-tail Sierra shards on a California o8 strata run) that both want strata+location and risk a block close. Per the issue, the blocking is plumbing, not algebra: `merge_tdigests` already implements the located pairwise merge (`locations1=`/`locations2=`, `src/zagg/stats/tdigest.py`), `merge_tdigests_kway` the located k-way merge, stratum membership is per-photon row selection before the build (independent of block boundaries), and `merge_composition` exists with documented error bounds.

## Approach

- **No spill-record schema change turned out to be needed.** The plan's phase 1 anticipated appending the order-29 point-word column to the spill record gated on `location:`; on inspection the flush already spills **every** read-carrier column wholesale (`SpillAggregator.flush` → `_concat_and_group` → `SpillBlock.append`), and `leaf_id` is a read-carrier column (`read.py` materializes it for cell assignment). The single-block exact replay already depends on it — located fields work single-block today. So phase 1 reduces to the block-close build, the flush/append path is untouched, and unlocated spill records are byte-identical by construction (zero diff on the write path).
- **Block close** (`_fold_block`): located fields build `(digest, locations)` partials per cell via the same `build_tdigest(values, locations=...)` the pooled path uses; strata fields resolve their `where` param per cell over the spilled columns (same bare-column/expression resolution as `calculate_cell_statistics`, sharing `_compile_param_expr`) and delegate to `build_tdigest_where`.
- **Cross-block fold**: located pairwise via `merge_tdigests(..., locations1=, locations2=)`, located k-way via `merge_tdigests_kway(..., locations=)` at finalize — k-way vs pairwise follows the field's function exactly as the plain-digest fold does today (issue #279). Composition folds under option (a) — per-block `(word, n_signal)` parts collapse in ONE k-way weighted lane mean at finalize (`merge_composition_kway`, added on review to make the fold order-independent and single-quantization; `merge_composition` itself is untouched as the spec §3.4 binary law). See Questions for review.
- **Validation split**: spill's mergeability probe is the new `validate_spill_fold` (`streaming.py`) — wider than `validate_streaming`, which stays exactly as strict: `mode: merge` folds every `buffer_granules` flush, degrading locations continuously, so it keeps rejecting located/strata/composition; phase 3 only updates its messages to name the fold behavior and the `mode: spill` remedy.
- **Spec**: one descriptive sentence in `docs/specification.md` §2.2 naming spill-block closes as an additional merge source for the located channel. **No wire-format change and no fixture change** — cross-block-folded payloads are already conformant `zagg-ragged/1` + locations (a fold coarsens merged centroids' locations to common ancestors, the located channel's defined behavior above weight 1).

## Phases

- [x] Phase 1 — located + strata cross-block folds: `validate_spill_fold` probe (located + `build_tdigest_where`; composition still excluded), `_DigestField` classification, `_fold_block` builds located `(digest, locations)` / stratum partials, pairwise + k-way located folds, merged emission carries the located channel. (The probe move was planned for phase 2 but the pairwise fold happens inside `_fold_block`, so landing build-without-fold would have been incoherent; phase 2 is composition.)
- [x] Phase 2 — composition cross-block fold: per-block `(word, n_signal)` build, composition fold (k-way collapse at finalize after the round-2 review fold), probe accepts `pack_composition`, merged emission of the packed word.
- [x] Phase 3 — validation split + guardrails: `mode: merge` messages name the fold behavior and the spill remedy (strictness unchanged); `SpillOverflowError` message names what still has no fold law.
- [x] Phase 4 — spec sentence + acceptance battery: located+strata+composition config through a forced multi-block run end-to-end; unlocated spill unchanged; composition within the documented O(n/510)-per-fold bound.

## How tested

New battery in `tests/test_spill_crossblock.py` (phase 1: probe classification; located multi-block digests byte-identical to the unlocated fold with locations bounded — every merged location is an ancestor-or-equal of a contributor word and a descendant of the contributors' common ancestor; below-compression-knee locations are the exact order-29 point words; stratum weights exact vs pooled and invariant to block placement; quantiles within t-digest tolerance vs pooled). Plus `uv run pytest tests/test_spill.py tests/test_streaming.py` (all pre-existing spill/streaming tests green, unchanged), `ruff check` / `ruff format --check`, and the full suite before ready-for-review.

Full-suite status: `uv run pytest` (minus `tests/test_lambda_build.py`, environmental) is 3200 passed / 37 skipped locally. Known pre-existing local issues, not touched here: `ruff check src` flags `N818` on `src/zagg/registry.py` (`UnknownCapability`) on a clean main checkout, and `ruff format --check` flags `tests/data/benchmark/README.md` (both reproduce with this PR's changes stashed); `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` fails locally for environmental reasons.

## Questions for review

- **Composition multi-block fold — needs a ruling (the question #321 deferred).** Phase 2 implements **option (a)**: fold composition words across blocks via `merge_composition`, accepting the documented lane-quantization error (O(n/510) per fold; presence exact via the floor) — an overflow shard's composition word is *not byte-stable* against the single-block result. Option (b) would leave composition non-mergeable so a strata-with-composition config still fails loudly on a block close. Flipping to (b) is a one-line change: remove the `pack_composition` acceptance branch in `validate_spill_fold` (`src/zagg/processing/streaming.py`). Working assumption is (a) per the issue; awaiting ratification.

- **Follow-up flagged during review (pre-existing, outside this diff):** `validate_streaming` (mode: merge) accepts a `build_tdigest` field carrying a stray `where` param and would silently ignore the mask in its fold — the spill probe now rejects that mismatch, but the merge-mode validator has the same hole on main. Standing on the review thread; happy to file it as a small-fix issue if wanted.

### DIFF
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -74,7 +74,8 @@ class SpillOverflowError(RuntimeError):
     """A spill block hit its threshold under a config with no merge law.
 
     Raised the moment a second block would open (never on single-block
-    shards, where every reducer is exact); the message names the remedies.
+    shards, where every reducer is exact); the message carries the probe's
+    per-field verdict (``validate_spill_fold``) and the remedies.
     Deliberately a distinct type so the worker's tolerated per-granule
     ``except`` can re-raise it instead of warn-and-continue.
     """
@@ -496,11 +497,17 @@ class SpillAggregator:
         # block close (SpillOverflowError).
         from zagg.processing.streaming import validate_spill_fold
 
+        # The probe names every offending field; keep its text so the overflow
+        # raised at the first block close can say WHICH reducer has no fold law
+        # (this is its only call site — dropping the message left an operator
+        # re-deriving it from a 900 s Lambda they cannot reproduce locally).
+        self._fold_problems: str | None = None
         try:
             validate_spill_fold(config)
             self._mergeable = True
-        except ValueError:
+        except ValueError as e:
             self._mergeable = False
+            self._fold_problems = str(e)
         self._count_fields: list[str] = []
         self._digest_fields: dict[str, _DigestField] = {}
         # name -> (source, params): scalar pack_composition fields (issue #370
@@ -688,6 +695,7 @@ class SpillAggregator:
                 f"results cannot combine (the fold covers 'len'/'count', tdigest "
                 f"fields — located, where-strata, pairwise — and the packed "
                 f"composition word; single-block spill is exact for every reducer). "
+                f"{self._fold_problems}. "
                 f"Remedies: a bigger memory tier, a '-disk' function variant with "
                 f"more ephemeral storage, or a finer parent_order (smaller shards)."
             )
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -43,6 +43,7 @@ naming the ``-disk`` function-variant fix.
 from __future__ import annotations
 
 import ast
+import logging
 import os
 import tempfile
 import threading
@@ -65,6 +66,8 @@ from zagg.stats.tdigest import (
     merge_tdigests,
     merge_tdigests_kway,
 )
+
+logger = logging.getLogger(__name__)
 
 #: Floor for the spill-enable /tmp check: below this, even a degraded
 #: many-block run is pointless — fail at config time instead of thrashing.
@@ -706,6 +709,16 @@ class SpillAggregator:
         """True when no observation ever survived filtering."""
         return self.n_obs_total == 0 and not self._buffer
 
+    @property
+    def closed_blocks(self) -> int:
+        """Blocks closed at the threshold; 0 = exact single-block regime.
+
+        Ridden into the worker's ``phase_timings`` as ``spill_blocks_closed``
+        (issue #370) so an overflow-folded leaf is queryable in the run
+        telemetry, the same route ``spill_bytes`` takes.
+        """
+        return self._closed_blocks
+
     def occupied_cells(self) -> np.ndarray:
         """Distinct populated cell words (issue #200 coverage sink), sorted."""
         if not self._occupied:
@@ -736,6 +749,17 @@ class SpillAggregator:
                 f"{self._fold_problems}. "
                 f"Remedies: a bigger memory tier, a '-disk' function variant with "
                 f"more ephemeral storage, or a finer parent_order (smaller shards)."
+            )
+        if self._closed_blocks == 0:
+            # Once per shard, at the moment the exact regime is left (issue
+            # #370): from here on outputs FOLD across blocks — an overflow
+            # leaf is not byte-identical to its single-block result.
+            logger.warning(
+                f"spill block threshold ({self.block_bytes:,} bytes) crossed: this "
+                f"shard leaves the exact single-block regime and its outputs now "
+                f"fold across blocks — digests merge under t-digest semantics, "
+                f"located centroids coarsen to common ancestors, composition takes "
+                f"one k-way re-quantization; counts stay exact."
             )
         block = self._block
         self._block = SpillBlock(self.tmp_dir)
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -56,6 +56,7 @@ from zagg.config import (
     get_data_vars,
     get_output_signature,
 )
+from zagg.stats.composition import merge_composition, pack_composition_n
 from zagg.stats.tdigest import (
     _DEFAULT_DELTA,
     build_tdigest,
@@ -339,15 +340,16 @@ class _DigestField(NamedTuple):
     where: object | None
 
 
-def _resolve_where(param, col_arrays: dict[str, np.ndarray], start: int, end: int):
-    """Resolve one cell's ``where`` param over its block rows, the pooled way.
+def _resolve_param(param, col_arrays: dict[str, np.ndarray], start: int, end: int):
+    """Resolve one cell's param over its block rows, the pooled path's way.
 
     Mirrors ``calculate_cell_statistics``'s params resolution exactly (bare
     column name -> that column's rows; a string naming columns -> eval'd over
     the cell's rows with the same cached code object; anything else passes
-    through), so a stratum's per-block membership is computed by the identical
-    rule the pooled/single-block replay applies per cell — block boundaries
-    cannot shift a photon between strata.
+    through), so a stratum's per-block ``where`` membership and a composition
+    field's conf columns are computed by the identical rule the pooled/
+    single-block replay applies per cell — block boundaries cannot shift a
+    photon between strata or lanes.
     """
     from zagg.processing.aggregate import _compile_param_expr
 
@@ -440,7 +442,8 @@ class SpillAggregator:
       partition-by-partition into running mergeable state (counts by
       summation, tdigests via ``merge_tdigests``/``merge_tdigests_kway`` —
       including located fields and ``build_tdigest_where`` strata, whose
-      per-block builds fold under the located overloads, issue #370),
+      per-block builds fold under the located overloads — and composition
+      words via ``merge_composition``, issue #370),
       collapsing merge rounds from N/buffer to ~spill/threshold. A
       config with any reducer outside the ``validate_spill_fold`` surface
       raises :class:`SpillOverflowError` at the first crossing instead of
@@ -483,8 +486,12 @@ class SpillAggregator:
             self._mergeable = False
         self._count_fields: list[str] = []
         self._digest_fields: dict[str, _DigestField] = {}
+        # name -> (source, params): scalar pack_composition fields (issue #370
+        # option (a)); params carry the conf column names + threshold, resolved
+        # per cell at fold time exactly like the pooled path.
+        self._composition_fields: dict[str, tuple[str, dict]] = {}
         if self._mergeable:
-            from zagg.processing.streaming import _TDIGEST_PAIRWISE_FUNCTION
+            from zagg.processing.streaming import _COMPOSITION_FUNCTION, _TDIGEST_PAIRWISE_FUNCTION
 
             for name, meta in agg_fields.items():
                 sig = get_output_signature(meta)
@@ -497,6 +504,11 @@ class SpillAggregator:
                         pairwise=meta.get("function") == _TDIGEST_PAIRWISE_FUNCTION,
                         location=sig["location"],
                         where=params.get("where"),
+                    )
+                elif meta.get("function") == _COMPOSITION_FUNCTION:
+                    self._composition_fields[name] = (
+                        meta.get("source") or "h_li",
+                        dict(meta.get("params") or {}),
                     )
                 else:
                     self._count_fields.append(name)
@@ -531,6 +543,14 @@ class SpillAggregator:
         # merge_tdigests / merge_tdigests_kway overloads.
         self._digest_locs: dict[str, dict[int, np.ndarray]] = {
             n: {} for n, f in self._digest_fields.items() if f.location
+        }
+        # Composition running state (issue #370): per-cell (word, n_signal),
+        # folded pairwise in block-close order via merge_composition — presence
+        # exact (the floor survives re-quantization), counts within the
+        # documented O(n/510)-per-fold bound, so an overflow shard's word is
+        # NOT byte-stable against the single-block result (option (a)).
+        self._compositions: dict[str, dict[int, tuple[int, int]]] = {
+            n: {} for n in self._composition_fields
         }
         # k-way fold accumulator: for order-independent fields, each block's
         # per-cell digest is stashed here and collapsed once at finalize
@@ -691,13 +711,16 @@ class SpillAggregator:
         ``leaf_id`` column is a spilled read column, so the block holds the
         exact order-29 point words), a ``where`` stratum field (issue #321)
         via ``build_tdigest_where`` with its mask resolved per cell over the
-        block's spilled columns (:func:`_resolve_where` — row selection
+        block's spilled columns (:func:`_resolve_param` — row selection
         precedes the build, so block boundaries cannot shift a photon between
         strata, issue #370) — and then, per the field's fold law: **k-way**
         fields stash the block digest (+ locations) for one order-independent
         collapse at finalize (:meth:`_finalize_kway`); **pairwise** fields
         merge it into the running digest here, the located overload carrying
-        the location channel. Either way it is one build round per block
+        the location channel. Composition fields pack a per-block ``(word,
+        n_signal)`` pair (``pack_composition_n`` — the same predicate and
+        bytes as the pooled reducer) and fold it via ``merge_composition`` in
+        block-close order. Either way it is one build round per block
         instead of per buffer — the ~6x merge-CPU collapse the design targets
         (issue #279).
         """
@@ -715,7 +738,7 @@ class SpillAggregator:
                     values = col_arrays[f.source][start:end]
                     locs = col_arrays[f.location][start:end] if f.location else None
                     if f.where is not None:
-                        where = _resolve_where(f.where, col_arrays, start, end)
+                        where = _resolve_param(f.where, col_arrays, start, end)
                         built = build_tdigest_where(
                             values, delta=f.delta, where=where, locations=locs
                         )
@@ -744,6 +767,21 @@ class SpillAggregator:
                         self._digest_locs[name][cell] = merged_locs
                     else:
                         self._digests[name][cell] = merge_tdigests(held, fresh, delta=f.delta)
+                for name, (source, params) in self._composition_fields.items():
+                    values = col_arrays[source][start:end]
+                    kwargs = {
+                        k: _resolve_param(v, col_arrays, start, end) for k, v in params.items()
+                    }
+                    word, n = pack_composition_n(values, **kwargs)
+                    held_comp = self._compositions[name].get(cell)
+                    if held_comp is None:
+                        self._compositions[name][cell] = (word, n)
+                    else:
+                        held_word, held_n = held_comp
+                        self._compositions[name][cell] = (
+                            merge_composition(held_word, held_n, word, n),
+                            held_n + n,
+                        )
 
     def _finalize_kway(self) -> None:
         """Collapse each k-way field's per-block parts into one digest per cell.
@@ -859,6 +897,12 @@ class SpillAggregator:
                 stats_arrays[name] = np.full(n_cells, np.nan, dtype=dtype)
             else:
                 stats_arrays[name] = np.zeros(n_cells, dtype=dtype)
+        for name in self._composition_fields:
+            # The packed word (issue #321): integer dtype, numeric fill — the
+            # spec mandates fill_value 0 so unwritten cells read as "no lanes".
+            meta = agg_fields[name]
+            dtype = np.dtype(meta.get("dtype", "uint64"))
+            stats_arrays[name] = np.full(n_cells, meta.get("fill_value", 0), dtype=dtype)
         ragged_payloads: dict[str, list] = {n: [] for n in self._digest_fields}
         ragged_cell_indices: dict[str, list[int]] = {n: [] for n in self._digest_fields}
         # Located fields only (issue #87): keyed presence tells the worker to
@@ -875,6 +919,10 @@ class SpillAggregator:
             cells_with_data += 1
             for name in self._count_fields:
                 stats_arrays[name][i] = count
+            for name in self._composition_fields:
+                # Every occupied cell has an entry (the fold writes one per
+                # cell per block); the word may legitimately be 0 (no signal).
+                stats_arrays[name][i] = self._compositions[name][cell][0]
             for name in self._digest_fields:
                 digest = self._digests[name].get(cell)
                 if digest is not None and digest.size > 0:
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -954,6 +954,8 @@ class SpillAggregator:
 
     def _chunk_outputs_merged(self, children, agg_fields: dict):
         """Multi-block regime: emit from the cross-block mergeable state."""
+        from zagg.processing.aggregate import _field_sentinel, _integer_fill
+
         if not self._finalized:
             # Join the in-flight overlap reduce (its failure re-raises here),
             # then fold the final (still-open, never threshold-closed) block
@@ -977,11 +979,21 @@ class SpillAggregator:
             else:
                 stats_arrays[name] = np.zeros(n_cells, dtype=dtype)
         for name in self._composition_fields:
-            # The packed word (issue #321): integer dtype, numeric fill — the
-            # spec mandates fill_value 0 so unwritten cells read as "no lanes".
+            # The packed word (issue #321). Same dtype default as every other
+            # emission path (`_aggregate_chunk_cells`, the count loop above) and
+            # the same fill derivation the pooled path uses for an empty cell
+            # (`_integer_fill` for integer dtypes — it names a non-numeric
+            # sentinel instead of letting it reach numpy — else
+            # `_field_sentinel`): a dtype-omitted composition field must not
+            # store float32 single-block and uint64 multi-block.
             meta = agg_fields[name]
-            dtype = np.dtype(meta.get("dtype", "uint64"))
-            stats_arrays[name] = np.full(n_cells, meta.get("fill_value", 0), dtype=dtype)
+            dtype = np.dtype(meta.get("dtype", "float32"))
+            fill = (
+                _integer_fill(meta, dtype)
+                if np.issubdtype(dtype, np.integer)
+                else _field_sentinel(meta)
+            )
+            stats_arrays[name] = np.full(n_cells, fill, dtype=dtype)
         ragged_payloads: dict[str, list] = {n: [] for n in self._digest_fields}
         ragged_cell_indices: dict[str, list[int]] = {n: [] for n in self._digest_fields}
         # Located fields only (issue #87): keyed presence tells the worker to
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -747,6 +747,7 @@ class SpillAggregator:
         """
         from zagg.processing.aggregate import _group_columns
 
+        self._check_fold_columns(block.schema)
         for key in block.partition_keys():
             t0 = time.perf_counter()
             cells, cols = block.read_partition(key, close=True)
@@ -803,6 +804,42 @@ class SpillAggregator:
                             merge_composition(held_word, held_n, word, n),
                             held_n + n,
                         )
+
+    def _check_fold_columns(self, schema) -> None:
+        """Name a missing source/location column, as the pooled path does.
+
+        The fold indexes ``col_arrays`` by each field's declared ``source`` and
+        ``location``; a column the block never carried would surface as a bare
+        ``KeyError`` raised on the overlap reducer thread, which
+        :meth:`_join_reducer` re-wraps as :class:`SpillReduceError` with the real
+        cause reachable only via ``__cause__`` — a materially worse diagnostic
+        than ``calculate_cell_statistics``'s named ``ValueError`` for the same
+        config error, and one that only appears on shards big enough to close a
+        block. Checked once per block against the block schema, off the per-cell
+        loop. An empty block (nothing appended, so no schema) has nothing to
+        fold and nothing to check.
+        """
+        available = sorted(name for name, _ in schema or [])
+        if not available:
+            return
+        for name, f in self._digest_fields.items():
+            if f.source not in available:
+                raise ValueError(
+                    f"ragged field {name!r} declares source: {f.source!r} but that "
+                    f"column is not in the spilled block (available: {available})"
+                )
+            if f.location is not None and f.location not in available:
+                raise ValueError(
+                    f"ragged field {name!r} declares location: {f.location!r} but that "
+                    f"column is not in the spilled block (available: {available}); "
+                    f"per-observation mortons require a HEALPix grid"
+                )
+        for name, (source, _) in self._composition_fields.items():
+            if source not in available:
+                raise ValueError(
+                    f"field {name!r} declares source: {source!r} but that column is "
+                    f"not in the spilled block (available: {available})"
+                )
 
     def _finalize_kway(self) -> None:
         """Collapse each k-way field's per-block parts into one digest per cell.
--- a/src/zagg/processing/spill.py
+++ b/src/zagg/processing/spill.py
@@ -46,6 +46,7 @@ import os
 import tempfile
 import threading
 import time
+from typing import NamedTuple
 
 import numpy as np
 
@@ -58,6 +59,7 @@ from zagg.config import (
 from zagg.stats.tdigest import (
     _DEFAULT_DELTA,
     build_tdigest,
+    build_tdigest_where,
     merge_tdigests,
     merge_tdigests_kway,
 )
@@ -319,6 +321,45 @@ class SpillBlock:
         self._partitions.clear()
 
 
+class _DigestField(NamedTuple):
+    """One ragged tdigest field's fold-relevant declaration (issues #279/#370).
+
+    ``pairwise`` selects the cross-block fold law: ``False`` — order-independent
+    k-way collapse at finalize (``build_tdigest``/``build_tdigest_where``, the
+    default); ``True`` — pairwise left-fold per block (``build_tdigest_pairwise``).
+    ``location`` is the per-observation morton column (issue #87) or ``None``;
+    ``where`` the field's raw ``where`` param (a column name or expression the
+    fold resolves per cell, exactly like the pooled path) or ``None``.
+    """
+
+    source: str
+    delta: int
+    pairwise: bool
+    location: str | None
+    where: object | None
+
+
+def _resolve_where(param, col_arrays: dict[str, np.ndarray], start: int, end: int):
+    """Resolve one cell's ``where`` param over its block rows, the pooled way.
+
+    Mirrors ``calculate_cell_statistics``'s params resolution exactly (bare
+    column name -> that column's rows; a string naming columns -> eval'd over
+    the cell's rows with the same cached code object; anything else passes
+    through), so a stratum's per-block membership is computed by the identical
+    rule the pooled/single-block replay applies per cell — block boundaries
+    cannot shift a photon between strata.
+    """
+    from zagg.processing.aggregate import _compile_param_expr
+
+    if isinstance(param, str) and param in col_arrays:
+        return col_arrays[param][start:end]
+    if isinstance(param, str) and any(c in param for c in col_arrays):
+        cell_data = {col: arr[start:end] for col, arr in col_arrays.items()}
+        ns = {"__builtins__": {}, "np": np, "numpy": np, **cell_data}
+        return eval(_compile_param_expr(param), ns)  # noqa: S307
+    return param
+
+
 def _memory_budget_bytes() -> int:
     """Worker memory budget: Lambda env, else cgroup v2 limit, else RAM."""
     mb = os.environ.get("AWS_LAMBDA_FUNCTION_MEMORY_SIZE")
@@ -397,11 +438,13 @@ class SpillAggregator:
     - **Multi block** (bytes hit the threshold — see
       :func:`_default_block_bytes`): each closing block is reduced
       partition-by-partition into running mergeable state (counts by
-      summation, tdigests via ``merge_tdigests`` — the StreamingAggregator
-      laws), collapsing merge rounds from N/buffer to ~spill/threshold. A
-      config with any non-mergeable reducer raises
-      :class:`SpillOverflowError` at the first crossing instead of silently
-      approximating.
+      summation, tdigests via ``merge_tdigests``/``merge_tdigests_kway`` —
+      including located fields and ``build_tdigest_where`` strata, whose
+      per-block builds fold under the located overloads, issue #370),
+      collapsing merge rounds from N/buffer to ~spill/threshold. A
+      config with any reducer outside the ``validate_spill_fold`` surface
+      raises :class:`SpillOverflowError` at the first crossing instead of
+      silently approximating.
 
     ``chunk_outputs`` returns the 5-tuple ``_aggregate_chunk_cells`` contract
     (``stats_arrays, ragged_payloads, ragged_cell_indices, ragged_locations,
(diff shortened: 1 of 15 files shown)


src/zagg/processing/spill.py
```diff
@@ -641,29 +711,64 @@ def _fold_block(self, block: SpillBlock) -> None:
...
-                for name, (source, delta, pairwise) in self._digest_fields.items():
-                    fresh = build_tdigest(col_arrays[source][start:end], delta=delta)
-                    if pairwise:
-                        held = self._digests[name].get(cell)
-                        self._digests[name][cell] = (
-                            fresh if held is None else merge_tdigests(held, fresh, delta=delta)
+                for name, f in self._digest_fields.items():
+                    values = col_arrays[f.source][start:end]
+                    locs = col_arrays[f.location][start:end] if f.location else None
+                    if f.where is not None:
```

**blocking** — the fold picks its reducer from `f.where is not None`, not from the declared `function`, and `validate_spill_fold` never ties the two together. Both directions are accepted by the probe and then silently disagree with the pooled/single-block replay:

(1) `function: build_tdigest_where` with **no** `params.where` — probe accepts, `_DigestField.where` is `None`, so `_fold_block` takes the `else` branch and folds the **whole population** into the stratum field. The pooled path raises instead.

(2) `function: build_tdigest` (or `build_tdigest_pairwise`) **carrying** a `where` param — probe accepts, `_DigestField.where` is set, so `_fold_block` calls `build_tdigest_where` and silently **masks**. The pooled path raises instead.

Repro against this commit (`git archive d33f93f` + `PYTHONPATH=src`):

```
(1) probe ACCEPTS build_tdigest_where with no `where`
    _mergeable: True  field.where: None
(2) probe ACCEPTS build_tdigest carrying a `where` param
    _mergeable: True  field.where: h_ph > 0
(1) pooled -> raises TypeError build_tdigest_where() missing 1 required keyword-only argument: 'where'
(2) pooled -> raises TypeError build_tdigest() got an unexpected keyword argument 'where'
```

So for the *same* config the shard either crashes or emits a plausible-looking-but-wrong digest depending purely on whether the block threshold was crossed — which is exactly the single-block/multi-block equivalence the PR is claiming. Case (1) is the dangerous one: a config that drops `where` by copy-paste writes the unstratified digest into `h_sig`/`h_noise` and nothing downstream can tell.

Concrete fix, two parts:
- in `validate_spill_fold`, reject the mismatch: `("where" in (meta.get("params") or {})) != (meta.get("function") == _TDIGEST_WHERE_FUNCTION)` is a config error in both directions, and the message can name it (`build_tdigest_where requires a 'where' param` / `'where' is only meaningful for build_tdigest_where`).
- in `_fold_block`, branch on the declared function (carry it on `_DigestField`, or resolve the callable once at construction via `resolve_function`) rather than on `where` presence, so the fold can never substitute a different reducer than the config declares.

src/zagg/processing/spill.py
```diff
@@ -641,29 +711,64 @@ def _fold_block(self, block: SpillBlock) -> None:
...
-            delta = self._digest_fields[name][1]
+            f = self._digest_fields[name]
             dest = self._digests[name]
-            for cell, parts in cell_parts.items():
-                dest[cell] = merge_tdigests_kway(parts, delta=delta)
+            if f.location:
+                loc_parts = self._digest_loc_parts[name]
+                dest_locs = self._digest_locs[name]
+                for cell, parts in cell_parts.items():
+                    dest[cell], dest_locs[cell] = merge_tdigests_kway(
```
espg, line 964: 🤖 *from Claude (review)*

**non-blocking** — the docstring extends the order-independence guarantee to the location channel ("A located field's location parts collapse in the same pass ... keeping both channels row-aligned", under "the result does not depend on block reduce order — the property #280 relies on to parallelize the reducer"). That holds for the digest, but **not** for the locations.

`merge_tdigests_kway` lexsorts on `(mean, weight)` and notes that sub-centroids with identical `(mean, weight)` "are interchangeable, so their residual order does not affect the result" — true for the centroids, but `_centroid_ancestors` then groups the *locations* by the `starts` of that same order. Tied sub-centroids carry different location words, so when a tie straddles a compression boundary the per-centroid ancestors differ with part order:

```python
vals = np.array([1., 1., 2., 2., 3., 3.])          # ties across two blocks
d1, l1 = build_tdigest(vals, delta=4, locations=la)  # block A words
d2, l2 = build_tdigest(vals, delta=4, locations=lb)  # block B words
o1, ol1 = merge_tdigests_kway([d1, d2], delta=4, locations=[l1, l2])
o2, ol2 = merge_tdigests_kway([d2, d1], delta=4, locations=[l2, l1])
# digest equal: True
# locs equal  : False
```

Today the part order is deterministic (block-close order), so this is latent rather than live — but issue #280 parallelizing the reducer is precisely the case the docstring invokes, and quantized/discrete sources (the tie case `merge_tdigests_kway`'s own comment calls out) are common in ATL03.

Suggest one of: (a) add the location word as a final lexsort key in the located branch of `merge_tdigests_kway` so ties break intrinsically, or (b) narrow this docstring (and `merge_tdigests_kway`'s) to say the *digest* is order-independent while the location channel is order-independent only up to `(mean, weight)` ties — and note it on issue #280 so the parallel reducer doesn't inherit an unstated assumption.

tests/test_spill_crossblock.py
```diff
@@ -0,0 +1,385 @@
...
+        grid = _grid(cfg)
+        dfs = _granule_dfs(grid, key, _CELL_LISTS, obs_per_cell=20, seed=8)
+        _, ragged, _ = _run(monkeypatch, cfg, grid, key, list(dfs))
+        vals, idx, locs = ragged["h_tdigest"]
+        children = np.asarray(grid.children(key), dtype=np.uint64)
+        assert len(vals) > 0
+        for cell_i, digest, cell_locs in zip(idx, vals, locs, strict=True):
+            assert (digest[:, 1] == 1.0).all()
+            contributors = _contributors(dfs, grid, int(children[cell_i]))
+            np.testing.assert_array_equal(np.sort(cell_locs), np.sort(contributors))
```
espg, line 454: 🤖 *from Claude (review)*

**non-blocking (test sufficiency)** — nothing in the new file pins the digest-row ↔ location-row **alignment**, which is the property most likely to regress silently in the located fold (e.g. dropping the `[order]` reindex of `combined_locs`, or the `_digest_parts`/`_digest_loc_parts` index alignment in `_finalize_kway`). Every location assertion here is permutation-invariant:

- `_assert_ancestor_or_equal` bound (a) only asks whether `loc` is an ancestor of *some* contributor. Every location in a cell's vector is derived from that cell's contributors, so any permutation of the vector still satisfies it.
- bound (b) clips to `common_ancestor(all contributors)`, which for a single child cell is essentially "the location is inside the cell" — also permutation-invariant.
- `test_below_knee_locations_are_exact_point_words` compares `np.sort(cell_locs)` against `np.sort(contributors)` — a multiset check, again permutation-invariant.

Verified against this commit: shuffling every emitted cell's location vector and re-running the file's own helper passes on all 8 populated cells (17–19 centroids each), and the below-knee `np.sort` comparison is invariant by construction.

Concrete strengthening, cheapest in the below-knee test (which is already the loss-free regime): every centroid is weight 1 and `digest[i, 0]` **is** an observation value, so alignment is directly checkable. Build `{h_ph value -> leaf_id}` from the contributing rows (assert the values are unique first — they are drawn from a continuous normal, so ties are measure-zero) and assert `cell_locs[i] == value_to_leaf[digest[i, 0]]` for every `i`. That turns the below-knee case into a full per-row alignment pin, and a shuffled location channel fails immediately.

src/zagg/processing/spill.py
```diff
@@ -641,29 +711,64 @@ def _fold_block(self, block: SpillBlock) -> None:
...
                 self._counts[cell] = self._counts.get(cell, 0) + (end - start)
-                for name, (source, delta, pairwise) in self._digest_fields.items():
-                    fresh = build_tdigest(col_arrays[source][start:end], delta=delta)
-                    if pairwise:
-                        held = self._digests[name].get(cell)
-                        self._digests[name][cell] = (
-                            fresh if held is None else merge_tdigests(held, fresh, delta=delta)
+                for name, f in self._digest_fields.items():
+                    values = col_arrays[f.source][start:end]
+                    locs = col_arrays[f.location][start:end] if f.location else None
```

**non-blocking (error quality)** — `col_arrays[f.location]` (and `col_arrays[f.source]` on the line above) go straight to a bare `KeyError` when the column isn't in the block schema. The pooled path deliberately does not: `calculate_cell_statistics` raises a named `ValueError` for exactly this case —

```python
raise ValueError(
    f"ragged field {name!r} declares location: {loc_col!r} but that "
    f"column is not in the cell data (available: {sorted(cell_data)}); "
    f"per-observation mortons require a HEALPix grid"
)
```

`validate_spill_fold` accepts a located ragged field on any grid, so a `location: leaf_id` config on a grid/reader that doesn't emit `leaf_id` now gets the good pooled message in the single-block regime and, once a block closes, a bare `KeyError: 'leaf_id'` raised on the overlap reducer thread — which `_join_reducer` re-wraps as `SpillReduceError("spill block reduce failed on the overlap thread; the merged running state is incomplete")` with the real cause only reachable via `__cause__`. That is a materially worse diagnostic for the same config error, and it only appears on large shards.

Suggest validating the location/source columns once against `block.schema` at the top of `_fold_block` (or against the first flush's `col_arrays` keys) and raising the same named `ValueError` the pooled path does.

src/zagg/processing/spill.py
```diff
@@ -319,6 +321,45 @@ def close(self) -> None:
...
+    through), so a stratum's per-block membership is computed by the identical
+    rule the pooled/single-block replay applies per cell — block boundaries
+    cannot shift a photon between strata.
+    """
+    from zagg.processing.aggregate import _compile_param_expr
+
+    if isinstance(param, str) and param in col_arrays:
+        return col_arrays[param][start:end]
+    if isinstance(param, str) and any(c in param for c in col_arrays):
+        cell_data = {col: arr[start:end] for col, arr in col_arrays.items()}
```

**nit (per-cell cost)** — this rebuilds the whole per-cell namespace *per field*, inside `_fold_block`'s per-cell loop. The issue #321 strata config declares two complementary `where` fields over the same source, so each cell pays the slice-dict construction and the `any(c in param for c in col_arrays)` scan twice per block. The pooled path it mirrors builds `cell_data` **once** per cell (`aggregate.py`, `_aggregate_chunk_cells` → `calculate_cell_statistics`) and resolves each field's params against it.

The views themselves are cheap, but at o11 geometry (65,536 cells) × B blocks × 2 strata fields the dict/scan overhead is paid ~130k times per block instead of ~65k. `_fold_block` is already slicing per cell (`values`, `locs`), so hoisting is a small change: build `cell_data = {col: arr[start:end] for col, arr in col_arrays.items()}` once at the top of the per-cell body and have `_resolve_where` take the already-sliced namespace instead of `(col_arrays, start, end)`. That also makes the parity with `calculate_cell_statistics` structurally obvious rather than a comment.

Parity itself checks out otherwise: the namespace `{"__builtins__": {}, "np": np, "numpy": np, **cell_data}` and the bare-column / substring-match branch order match `aggregate.py:499-514` exactly, and the pooled namespace's one extra ingredient — `chunk_scalars` — cannot be present here because `validate_spill_fold` rejects `chunk_precompute` outright. Worth keeping that dependency noted in the docstring, since it is the only thing making the two namespaces identical.

src/zagg/processing/streaming.py
```diff
@@ -145,6 +153,62 @@ def validate_streaming(config: PipelineConfig) -> None:
         )
 
 
+def validate_spill_fold(config: PipelineConfig) -> None:
```
espg, line 193: 🤖 *from Claude (review)*

**nit** — `validate_spill_fold` is a ~50-line near-verbatim clone of `validate_streaming` directly above it. The two differ in exactly three places: the ragged branch drops the located rejection, the accepted function tuple is `_TDIGEST_SPILL_FUNCTIONS` instead of `_TDIGEST_FUNCTIONS`, and every message says "has no cross-block fold" instead of "cannot stream". Everything else — the `chunk_precompute` guard, the `expression` / `resolution` / `inner_shape` / scalar-`len` branches, the accumulate-then-join shape — is duplicated character for character.

That is a divergence trap: a future tightening of one branch (say the `inner_shape` check, or adding a `dtype` guard) lands in one validator and silently misses the other, and the miss is invisible because the two now govern different code paths with different failure modes (config error vs. `SpillOverflowError` at the first block close).

Suggest collapsing to one parametrized core, e.g. `_validate_fold(config, *, ragged_functions, allow_located, what, prefix)` with the two public functions as thin wrappers — the surface stays exactly as it is now (§4: terse, reviewable, no duplication), and future tightenings apply to both by construction.

Also untested in either validator's new copy: the `chunk_precompute` branch and the `inner_shape != (2,)` branch. `TestSpillFoldProbe` covers only the `expression` and non-tdigest-ragged rejections plus the three acceptances — worth two more one-liners so the widened probe's *rejections* are pinned as tightly as its acceptances.

src/zagg/processing/spill.py

Fixed in c047113 — both directions now fail at the probe, and the fold no longer infers its reducer from the param.

`_DigestField` gained a `stratum` flag set from the **declared** function, and `_fold_block` branches on it:

```python
pairwise=meta.get("function") == _TDIGEST_PAIRWISE_FUNCTION,
location=sig["location"],
where=params.get("where"),
stratum=meta.get("function") == _TDIGEST_WHERE_FUNCTION,
```
```python
if f.stratum:
    where = _resolve_param(f.where, cell_data)
    built = build_tdigest_where(values, delta=f.delta, where=where, locations=locs)
else:
    built = build_tdigest(values, delta=f.delta, locations=locs)
```

`validate_spill_fold` rejects the mismatch in both directions, exactly as suggested, so `stratum` implies `where is not None` and vice versa:

```python
elif ("where" in (meta.get("params") or {})) != (
    meta.get("function") == _TDIGEST_WHERE_FUNCTION
):
```
→ `field 'h_sig': zagg.stats.tdigest.build_tdigest_where requires a 'where' param`
→ `field 'h_tdigest': 'where' is only meaningful for zagg.stats.tdigest.build_tdigest_where (declared 'zagg.stats.tdigest.build_tdigest')`

Three tests in `TestSpillFoldProbe`: `test_where_function_without_where_param_rejected` and `test_where_param_on_plain_tdigest_rejected` each pin the probe message **and** assert the pooled replay raises the matching `TypeError` for the same field (via a new `_pooled_build` helper that resolves params and calls `resolve_function(meta["function"])` the way `calculate_cell_statistics` does), so the two paths are pinned as agreeing rather than merely both-failing. `test_mis_declared_where_config_is_not_mergeable` asserts `SpillAggregator` classifies neither field (`_mergeable` false, `_digest_fields == {}`), so `_fold_block` can never see one.

One related hole left standing, out of this diff: `validate_streaming` accepts `build_tdigest` carrying a `where` param too, and merge mode's `_digest_fields` is `(source, delta)` — it would silently ignore the mask. That is pre-existing on `main` (merge mode already rejects `build_tdigest_where` outright, so only direction (2) applies there) and is a change to a validator this PR does not otherwise touch, so it is flagged rather than folded.

src/zagg/processing/spill.py
espg: 🤖 *from Claude*

Took option (a) — the real fix — in bad4853: the located branch of `merge_tdigests_kway` adds the location word as a tertiary (least-significant) lexsort key, so ties break intrinsically and the channel is permutation-independent.

```python
if located:
    combined_locs = np.concatenate(locs)
    order = np.lexsort((combined_locs, combined[:, 1], combined[:, 0]))
    combined_locs = combined_locs[order]
else:
    order = np.lexsort((combined[:, 1], combined[:, 0]))
```

It changes no digest byte by construction — the tied rows carry identical means and weights, so permuting them leaves `combined[:, 0]` / `combined[:, 1]` untouched — and the regression test asserts exactly that alongside the permutation invariance.

Your repro confirmed the fault and the fix. Against `aadd262`:

```
PRE-FIX digest equal: True locs equal: False
```

and after: both `True`.

Regression test `TestMergeTDigestsKway::test_located_order_independent_with_tied_means` (`tests/test_tdigest.py`) builds three digests over `[1,1,2,2,3,3]` with different location seeds, then asserts across three permutations that the digest, the location vector, **and** the unlocated fold of the same digests all agree:

```python
ref, ref_locs = merge_tdigests_kway(ds, delta=4, locations=ls)
assert np.array_equal(ref, merge_tdigests_kway(ds, delta=4))
```

Docstrings updated to state the guarantee rather than narrow it: `merge_tdigests_kway` now says order-independence covers **both** channels, and `_finalize_kway` notes that a reducer parallelized under #280 inherits both. Did not comment on #280 itself — that is a side-effecting action on another issue, so it stays for a human.

tests/test_spill_crossblock.py
espg: 🤖 *from Claude*

Fixed in 6c87d0b, exactly the strengthening you proposed: `test_below_knee_locations_are_exact_point_words` now pins per-row alignment on top of the multiset check.

New helper `_contributor_rows` returns paired `(h_ph value, leaf word)` rows for a cell (`_contributors` is now a thin wrapper over it, so the other call sites are unchanged), and the test builds the value→leaf map:

```python
keys = values.astype(np.float32)
assert len(np.unique(keys)) == len(keys)
value_to_leaf = dict(zip(keys.tolist(), contributors.tolist(), strict=True))
for centroid, loc in zip(digest, cell_locs, strict=True):
    assert value_to_leaf[float(centroid[0])] == int(loc)
```

Uniqueness is asserted rather than assumed, per your note.

Verified it actually bites: injecting a shuffle of `dest_locs[cell]` into `_finalize_kway` fails the test immediately —

```
assert value_to_leaf[float(centroid[0])] == int(loc)
E   assert 12711976641192705660 == 12711983251147378621
```

whereas the pre-existing `np.sort` multiset comparison still passed. The shuffle was reverted; only the test file is in this commit.

src/zagg/processing/spill.py

Fixed in 213df5b — a new `_check_fold_columns` runs once per block at the top of `_fold_block`, against `block.schema`, off the per-cell loop:

```python
self._check_fold_columns(block.schema)
```

It names the field and the column for a missing ragged `source`, a missing ragged `location` (carrying the pooled path's `per-observation mortons require a HEALPix grid` tail), and a missing composition `source`:

```
ragged field 'h_tdigest' declares location: 'leaf_id' but that column is not in
the spilled block (available: ['h_ph']); per-observation mortons require a HEALPix grid
```

"spilled block" rather than "cell data" because that is what is actually being reported on; the rest mirrors `calculate_cell_statistics`. An empty block (no schema — the still-open final block of a shard whose last flush closed one) returns early: nothing to fold, nothing to check.

Five tests in a new `TestFoldColumnDiagnostics`: missing location, missing source, missing composition source, the empty-block no-op, and — for the diagnostic that motivated this — the overlap path, asserting the named `ValueError` survives as the `SpillReduceError`'s `__cause__` rather than a bare `KeyError`:

```python
agg._reduce_one(agg._block)
with pytest.raises(SpillReduceError) as exc:
    agg._join_reducer()
assert isinstance(exc.value.__cause__, ValueError)
assert "location: 'leaf_id'" in str(exc.value.__cause__)
```

src/zagg/processing/spill.py

Hoisted in 103c81e. `_resolve_param` now takes the already-sliced namespace, and `_fold_block` builds it once per cell:

```python
cell_data = {col: arr[start:end] for col, arr in col_arrays.items()}
for name, f in self._digest_fields.items():
    values = cell_data[f.source]
    locs = cell_data[f.location] if f.location else None
```

The composition kwargs loop shares the same namespace rather than rebuilding it per param:

```python
kwargs = {k: _resolve_param(v, cell_data) for k, v in params.items()}
```

so the per-cell dict construction and the `any(c in param for c in ...)` scan are paid once instead of once per field (and, for composition, once per param). The per-field `col_arrays[...][start:end]` slices are gone too — they now read out of `cell_data`. That makes the parity with `calculate_cell_statistics` structural, as you noted: one `cell_data` per cell, every field's params resolved against it.

Also took the docstring suggestion — `_resolve_param` now records the dependency explicitly:

> The two namespaces are identical only because `validate_spill_fold` rejects `chunk_precompute` outright: the pooled namespace's one extra ingredient is its `chunk_scalars`, which no config reaching this fold can carry.

and that branch is now pinned by a test (see the reply on the `validate_spill_fold` thread).

src/zagg/processing/streaming.py
espg: 🤖 *from Claude*

Split verdict: **tests added** (b20c084), **dedup declined**.

Added the two missing rejection tests to `TestSpillFoldProbe`, so the widened probe's rejections are pinned as tightly as its acceptances:

- `test_chunk_precompute_still_has_no_fold` — matches `chunk_precompute.*no cross-block fold`
- `test_mis_declared_inner_shape_still_has_no_fold` — `inner_shape: [3]`, matches `h_tdigest.*inner_shape.*no cross-block fold`

The `chunk_precompute` test carries the reason it is load-bearing beyond validation: rejecting it is what makes `_resolve_param`'s namespace identical to the pooled one (no `chunk_scalars` ingredient), which is the parity argument the fold rests on.

Declining the shared `_validate_fold(...)` core, for two reasons:

1. The two validators have **diverged further** in this fold, not converged. The ragged branch now carries a `where`/`function` agreement check that only `validate_spill_fold` can make — merge mode rejects `build_tdigest_where` outright, so the same predicate would be dead there. Parametrizing that as `allow_located` + `ragged_functions` + `what` + `prefix` no longer covers it; it would need a fifth knob whose only value is "spill", which is the point at which the wrapper costs more clarity than the duplication.
2. The messages route differently by design (`use mode: spill, issue #370` vs. naming the remedies), and that routing is per-branch, not a suffix — `prefix` alone will not reconstruct it.

The divergence trap you name is real and I would rather not paper over it, so the concrete counter-measure is the tests: the spill probe's `chunk_precompute`, `inner_shape`, `expression`, non-tdigest-ragged, `where` mismatch (both directions) and scalar branches now all have pinning tests, and `test_mode_merge_validation_unchanged` pins that the widening did not leak into merge mode. A future tightening that lands in one validator and misses the other shows up as an unpinned branch rather than as silence.

Leaving the dedup standing as a judgment call rather than a settled one — if you would rather have the single parametrized core with the fifth knob, say so and I will fold it.

src/zagg/processing/spill.py
```diff
@@ -641,29 +763,121 @@ def _fold_block(self, block: SpillBlock) -> None:
...
+                    values = cell_data[source]
+                    kwargs = {k: _resolve_param(v, cell_data) for k, v in params.items()}
+                    word, n = pack_composition_n(values, **kwargs)
+                    held_comp = self._compositions[name].get(cell)
+                    if held_comp is None:
+                        self._compositions[name][cell] = (word, n)
+                    else:
+                        held_word, held_n = held_comp
+                        self._compositions[name][cell] = (
+                            merge_composition(held_word, held_n, word, n),
```

**non-blocking** — the composition cross-block fold is a pairwise left-fold, which gives up both determinism-under-#280 and accuracy that a one-pass k-way fold would provide for free.

`merge_composition` re-quantizes on every block close, so error compounds with block count *and* the result depends on close order. Both are avoidable: the per-block parts are `(word, n)` pairs and the merge law is a weighted **sum**, so folding all B parts in one pass — `lane = round(Σ nᵢ·laneᵢ / Σ nᵢ)` with the same presence floor — is exactly order-independent and re-quantizes once instead of B−1 times. Measured here on 12 blocks × 40 rows (n_signal = 467), recovered counts vs the pooled word:

```
pooled          [256 249 244 242 260  42 125 300]
pairwise fold   [255 247 242 244 258  42 125 299]   <- this PR
k-way one-pass  [256 249 244 242 260  42 125 300]   <- exact
```

It also closes a hazard this PR itself opens: `_finalize_kway`'s docstring (spill.py:854-865, updated in bad4853) now advertises that "a reducer parallelized under #280 inherits both" channels' order-independence — but `self._compositions` is running state that would silently change stored bytes if #280 reordered block reduces, and unlike `build_tdigest_pairwise` it is not an order-dependence the config opted into.

Suggestion: stash parts (`self._composition_parts[name].setdefault(cell, []).append((word, n))`) and collapse them in `_finalize_kway` via a `merge_composition_kway(parts)` in `zagg/stats/composition.py`, mirroring `merge_tdigests_kway`. Spec §3.4's binary law stays the reader-facing contract — a writer folding k-way is inside its documented "associative up to the bounded re-quantization error" tolerance, and strictly tighter than the binary chain. If the pairwise fold is preferred instead, `_finalize_kway`'s docstring should say the composition channel is explicitly **excluded** from the #280 property, so the parallel-reduce work doesn't inherit a false guarantee.

src/zagg/processing/spill.py
```diff
@@ -641,29 +763,121 @@ def _fold_block(self, block: SpillBlock) -> None:
...
+                    f"ragged field {name!r} declares source: {f.source!r} but that "
+                    f"column is not in the spilled block (available: {available})"
+                )
+            if f.location is not None and f.location not in available:
+                raise ValueError(
+                    f"ragged field {name!r} declares location: {f.location!r} but that "
+                    f"column is not in the spilled block (available: {available}); "
+                    f"per-observation mortons require a HEALPix grid"
+                )
+        for name, (source, _) in self._composition_fields.items():
```

**non-blocking** — `_check_fold_columns` names a missing `source`/`location`, but not a composition field's **param** columns, which are the ones most likely to be mistyped (five per field, all hand-written).

A conf column absent from the spilled block falls straight through `_resolve_param` (spill.py:374-379 — not a key, and the `any(c in param ...)` substring branch is false), so the raw string reaches `pack_composition_n` and dies inside `np.column_stack`. Reproduced:

```
ValueError: all the input array dimensions except for the concatenation axis must
match exactly, but along dimension 0, the array at index 0 has size 4 and the
array at index 4 has size 1
```

No field name, no column name — and on the overlap path it arrives only as `SpillReduceError.__cause__`, which is precisely the diagnostic this method's docstring says it exists to prevent, one layer over. Read columns come from `data_source.variables`, not from the agg params, so a param naming an unread column is reachable config, not a hypothetical.

Suggestion: `self._composition_fields` already carries `params` (it's discarded here via `(source, _)`), so this is a short extension mirroring `_resolve_param`'s own rule:

```python
for name, (source, params) in self._composition_fields.items():
    ...
    for pname, pval in params.items():
        if isinstance(pval, str) and pval not in available and not any(
            c in pval for c in available
        ):
            raise ValueError(
                f"field {name!r} param {pname}: {pval!r} names neither a spilled "
                f"column nor an expression over one (available: {available})"
            )
```

The same check applied to `_DigestField.where` would upgrade the stratum case too: a `where` expression over an unread column currently reaches `build_tdigest_where` and raises `where shape () does not match values shape (n,)`, which names neither the field nor the column.

src/zagg/processing/spill.py
```diff
@@ -754,8 +968,17 @@ def _chunk_outputs_merged(self, children, agg_fields: dict):
                 stats_arrays[name] = np.full(n_cells, np.nan, dtype=dtype)
             else:
                 stats_arrays[name] = np.zeros(n_cells, dtype=dtype)
+        for name in self._composition_fields:
+            # The packed word (issue #321): integer dtype, numeric fill — the
+            # spec mandates fill_value 0 so unwritten cells read as "no lanes".
+            meta = agg_fields[name]
+            dtype = np.dtype(meta.get("dtype", "uint64"))
+            stats_arrays[name] = np.full(n_cells, meta.get("fill_value", 0), dtype=dtype)
```

**non-blocking** — the composition prealloc diverges from every other emission path in two ways, and the dtype one makes single-block and multi-block spill disagree on the stored word.

**dtype.** `meta.get("dtype", "uint64")` — everywhere else the default is `float32` (`_aggregate_chunk_cells`, aggregate.py:792, and the `_count_fields` loop five lines above). Nothing forces a composition field to declare `dtype`: `_validate_composition_attrs` (config.py:642) pins `fill_value: 0` but says nothing about dtype, and I confirmed the probe accepts a `pack_composition` field with `dtype` omitted:

```
mergeable: True    sig["dtype"] (what pooled uses): None    spill merged dtype: uint64
```

So that config takes `_coerce_scalar_value`'s `float(word)` into a **float32** array on the single-block path (lossy above 2^24) and an exact **uint64** word on the multi-block path — a regime-dependent result in the one PR whose thesis is regime parity. Whichever default is chosen, the two spill regimes should pick the same one.

**fill.** `meta.get("fill_value", 0)` bypasses `_integer_fill` (aggregate.py:82), the helper that exists so the integer paths "agree with each other and with `_field_sentinel` on where the value comes from". Its named rejection of a non-numeric sentinel is lost here — a string sentinel reaches numpy instead:

```
ValueError: invalid literal for int() with base 10: np.str_('NaN')
```

Suggestion: `dtype = np.dtype(meta.get("dtype", "float32"))` and route the fill through `_integer_fill(meta, dtype)` when the dtype is integral, so both regimes share the pooled helpers rather than re-deriving the rule. If uint64 is meant to be non-negotiable (spec §3 says the word *is* uint64), the place to enforce it is config validation next to the `fill_value: 0` check, not a local default here.

src/zagg/processing/spill.py
```diff
@@ -580,8 +684,10 @@ def _close_block(self) -> None:
         if not self._mergeable:
             raise SpillOverflowError(
                 f"spill block hit the {self.block_bytes:,}-byte threshold but the "
-                f"config carries reducers with no merge law, so per-block results "
-                f"cannot combine (single-block spill is exact for every reducer). "
+                f"config carries reducers with no cross-block fold law, so per-block "
```
espg, line 745: 🤖 *from Claude (review)*

**non-blocking** — phase 3 made this message enumerate what the fold *does* cover, but it still cannot say which field failed, because the probe's per-field message is thrown away at the only call site.

spill.py:499-503:

```python
try:
    validate_spill_fold(config)
    self._mergeable = True
except ValueError:
    self._mergeable = False
```

`validate_spill_fold` is written to name every offender ("Raises `ValueError` naming every offending field", streaming.py:215) and this is its sole caller anywhere in `src/` — so the one place that text could reach an operator discards it. The result is that a 900 s Lambda dying at the first block close hands back a generic catalogue of supported reducers and leaves the author to re-derive which of their fields is outside it, on a shard large enough that they cannot reproduce it locally.

Suggestion: keep the reason and splice it in —

```python
except ValueError as e:
    self._mergeable = False
    self._fold_reason = str(e)
```

then `... single-block spill is exact for every reducer). {self._fold_reason}. Remedies: ...`. That also lets `test_overflow_message_names_the_fold_surface` (tests/test_spill_crossblock.py) assert `h_spread` appears by name, which is the property the test is actually after — the current `match="cross-block fold law.*memory tier"` passes for any rejected config whatsoever.

src/zagg/processing/spill.py

Accepted — the composition channel now folds one-pass k-way, in 08f1b7b.

- New `merge_composition_kway(parts)` in `src/zagg/stats/composition.py`: one weighted lane mean over every part, quantized **once**, same presence floor, `(0, 0)` identity (parts with `n <= 0` skipped; empty/all-empty returns `0`).

```python
    for word, n in parts:
        if n <= 0:
            continue
        la = unpack_composition(word).astype(np.float64)
        lanes += n * la
        present |= la > 0
        total += int(n)
    k = np.rint(lanes / total)
    k = np.where(present & (k == 0), 1.0, k)
```

- `merge_composition` itself is **untouched** — it stays the documented pairwise law and the reader-facing contract of spec §3.4; the k-way collapse is a writer-side refinement inside its documented "associative up to the bounded re-quantization error" tolerance.
- `SpillAggregator` stashes per-block `(word, n_signal)` parts in `_composition_parts` (same shape as `_digest_parts`) and collapses them in `_finalize_kway`, whose docstring now covers all three channels — so the #280 order-independence claim is true for composition too instead of being advertised over running state. Two ints per cell per block: negligible beside the digest parts.
- Error bound tightens accordingly: the per-block quantization contributes ≤ 0.5 lane to the weighted mean and the single final `rint` ≤ 0.5, so lane error ≤ 1 step **independent of block count**. `test_presence_exact_counts_within_fold_bound`'s tolerance drops from `1 + len(dfs) * n / 510` to `1 + n / 255`.
- New `TestMergeCompositionKway` in `tests/test_composition.py` pins identity/empty, order-independence under shuffled parts, the presence floor across 21 parts, `n`-weighting (a 1-row part against a 200-row one → lanes `254` / floored `1`), and that the k-way word is no further from the pooled word than the 12-part pairwise chain is.

Docstrings/comments describing the fold updated in `spill.py`, `streaming.py` (`_COMPOSITION_FUNCTION`, `validate_spill_fold`) and `configs/atl03_tdigest_strata_healpix.yaml`.

The option (a)/(b) ruling for review is **unchanged**: this is still option (a) — the stored word is not byte-stable against the single-block result, presence exact via the floor — and the flip point is still the one-line removal of `_COMPOSITION_FUNCTION` from the `validate_spill_fold` scalar branch.

src/zagg/processing/spill.py

Accepted, and extended to expressions as well — 01d62b0.

`_check_fold_columns` now also covers a composition field's `conf_*` params and a stratum's `where`, via one helper that mirrors `_resolve_param`'s namespace rather than its substring rule:

```python
def _unresolvable_names(param, available) -> tuple[str, ...]:
    ...
    tree = ast.parse(param, mode="eval")   # SyntaxError -> () : a genuine literal
    allowed = set(available) | {"np", "numpy"}
    return tuple(sorted({n.id for n in ast.walk(tree) if isinstance(n, ast.Name)} - allowed))
```

Why identifiers rather than the bare-name check you sketched: the eval namespace is exactly `{"__builtins__": {}, "np": np, "numpy": np, **cell_data}`, so a free `ast.Name` outside `available ∪ {np, numpy}` is *provably* unresolvable — and a bare column name parses as a single `Name`, so **one** rule covers both the bare-string and expression forms. That closes the `where` case you flagged in both directions: the fall-through-as-literal one (`where shape () does not match values shape (n,)`) and the `NameError` one, which named the identifier but never the field. Non-strings and unparseable strings return `()`, so genuine literals are untouched.

On scope, per the "don't reject strings that are genuine literals for other reducers" caveat: the check is confined to `pname.startswith("conf_")` for composition fields and to `_DigestField.where`. Those are the params that *must* be columns for `np.column_stack` / the mask to work; every other param keeps `_resolve_param`'s pass-through semantics unchanged. `merge_composition`/`_resolve_param` themselves are untouched.

Tests in `TestFoldColumnDiagnostics`: `test_missing_composition_param_column_named` (typo'd `conf_ocean: signal_conf_ocaen` → names field, param and column instead of "the array at index 4 has size 1"), `test_missing_where_column_named` (`where: "flag > 0"` → names `h_sig` and `['flag']`), and `test_expression_over_spilled_columns_passes` (`np.isfinite(h_ph) & (h_ph > 0)` must not trip it).

src/zagg/processing/spill.py

Accepted — both halves, in 8de7153. The merged prealloc now derives dtype and fill the way every other emission path does:

```python
            dtype = np.dtype(meta.get("dtype", "float32"))
            fill = (
                _integer_fill(meta, dtype)
                if np.issubdtype(dtype, np.integer)
                else _field_sentinel(meta)
            )
            stats_arrays[name] = np.full(n_cells, fill, dtype=dtype)
```

`_field_sentinel` / `_integer_fill` are imported at function level, like the module's other `zagg.processing.aggregate` imports. `np.full` (not `np.zeros`) is deliberate here: unlike the count fields, an unoccupied cell's composition slot is never rewritten in the loop below, so the prealloc value *is* the pooled path's `_empty_field_value`.

Enforcing uint64 at config validation instead was the alternative you named; I left that standing rather than widen the PR's config surface — this commit is the parity fix, and the two regimes now agree whichever default a config takes.

Tests, both in `TestCompositionMultiBlock`:
- `test_prealloc_dtype_and_fill_match_single_block[None|uint64]` — runs the same config through the single-block and forced-multi-block regimes and asserts identical stored dtype and identical empty-cell fill. With `dtype`/`fill_value` omitted that is `float32` + `NaN` in both (it was `uint64` + `0` multi-block before); with `uint64` declared it is `uint64` + `0` in both.
- `test_non_numeric_fill_on_the_word_is_named` — a `uint64` field with `fill_value: "NaN"` now fails with `_integer_fill`'s named message ("cannot hold a non-numeric sentinel") rather than `int('NaN')` out of numpy.

src/zagg/processing/spill.py
espg: 🤖 *from Claude*

Accepted — c415946. The probe's verdict is kept and spliced into the overflow message:

```python
        self._fold_problems: str | None = None
        try:
            validate_spill_fold(config)
            self._mergeable = True
        except ValueError as e:
            self._mergeable = False
            self._fold_problems = str(e)
```

`SpillOverflowError` now reads `... single-block spill is exact for every reducer). {self._fold_problems}. Remedies: a bigger memory tier, ...`, so the operator gets the per-field reason — e.g. `spill blocks cannot cross-block fold this config: field 'h_spread': expression fields have no cross-block fold` — beside the fold surface and the remedies. The "cross-block fold law" and "memory tier" phrases are unchanged.

`test_overflow_message_names_the_fold_surface` now matches `"cross-block fold law.*'h_spread': expression fields.*memory tier"`, so it fails if the offending field stops being named rather than passing for any rejected config. The class docstring says the message carries the probe's per-field verdict too.

espg:
🤖 *from Claude*

Folded a review finding on fold-regime visibility: entering the multi-block fold was silent and untracked, so an overflow-folded leaf was indistinguishable from an exact one in the console and the run telemetry. a11d713 adds both halves:

- **One warning per shard** at the first block close (`SpillAggregator._close_block`), naming the consequences: digests fold under t-digest merge semantics, located centroids coarsen to common ancestors, composition takes one k-way re-quantization; counts stay exact. Fires exactly once per aggregator lifetime, never on single-block shards.
- **`spill_blocks_closed` in the stats record** (0/absent = exact regime): ridden through `phase_timings` like `spill_bytes`, split out of the seconds-only block by `telemetry.build_record` into its own field, summed on rollup (`_SUM_OR_NONE_KEYS`), and flattened to a real run-parquet column (`_ROW_SCALARS`) so folded leaves are queryable after the fact.

Tests: warning-once on a forced multi-block run + silence and 0 on single-block (`TestFoldRegimeVisibility`, tests/test_spill_crossblock.py); record split/merge-sum/flatten coverage in tests/test_telemetry.py; the spill instrumentation set test updated. Full suite green locally (3,226 passed / 37 skipped, lambda-build excluded as environmental).

PR #378, opened 2026-08-04
small fixes: event-transport pacing window, point-path observations-read counter

### LINKED ISSUE #372: conus shardmap builder mislabels the considered count as intersecting
🤖 *from Claude*

`data/conus/build_conus_shardmap.py:254` writes `"granules_intersecting_conus": sm.metadata["total_granules"]` — but `total_granules` is the catalog records **considered** (the bbox-prefilter input), not the intersecting set. Line 267 of the same stats JSON writes the correct distinct count under `summary.total_granules_intersecting`, so the emitted file contradicts itself today.

One-line fix now that `ShardMap.build` records the distinct assigned count (landed with the demo-week `granules_assigned` metadata addition): swap line 254 to `sm.metadata["granules_assigned"]`. While there, `data/build_aoi_shardmap.py` hand-computes the same distinct set (`{g["id"] for gl in sm.granules for g in gl}`) and can read the field instead — equivalent, just deduplicated.

Surfaced by the adversarial review of the `granules_assigned` change (2026-08-04); flagged out-of-scope there.

### LINKED ISSUE #374: point path: record observations READ (pre-filter) alongside n_obs in the stats record
🤖 *from Claude*

The point path's telemetry records **kept** observations only: `n_obs` is the worker's `total_obs` — photons that survived the TEP filter, fell in the shard's cells, and were aggregated. Photons fetched-but-discarded (read_plan `pad: 1` segment padding, segments straddling the shard boundary, filter rejects) are invisible, so the read-vs-keep ratio — the direct measure of the #43 segment→photon indexed-IO efficiency — cannot be computed from the run parquet.

The **raster** path already has exactly this class of counter (issue #297: `raster_bytes_read` / `px_decoded` / `px_sampled`, mergeable-by-construction, ratio derived at read time never stored). This is the point-path counterpart:

- Worker-side: count rows decoded per granule **before** filtering/assignment, summed per shard as `n_obs_read` (int, ≥ `n_obs`), stamped into the worker result metadata alongside `total_obs`.
- `telemetry.build_record` + `flatten_record`/`_ROW_SCALARS`: carry it as a nullable column (absent on records from workers predating the field — absence reads as unmeasured, per the #297 convention); `telemetry.merge` sums it.
- Tests: a shard whose read plan pads across the boundary pins `n_obs_read > n_obs`; merge law; back-compat row without the field.

Motivating context (2026-08-04 California o9 demo run): 31.9B kept observations across 2,721 shards with 190,625 shard-granule read events over 2,356 distinct granules (~81 opens/granule; byte duplication bounded by the segment index) — "read X B, kept 31.9 B" is the efficiency claim the telemetry can't currently support.

### LINKED ISSUE #375: event-transport dispatch pacing inherits the v1 fd ceiling
🤖 *from Claude*

`Run.dispatch(transport="event")` sizes its pacing window via the same `_probe_workers` → `compute_available_workers` path as the sync transport: `min(n, fd_ceiling, account_headroom)`. The fd term exists because v1 holds one socket per in-flight shard — but the event transport holds **no** connections; its window is purely dispatched-but-unresolved pacing, bounded by account headroom (the 60 s `MaximumEventAgeInSeconds` drop window), not client fds.

Observed (2026-08-04, California o9 demo, 2,721 shards): account headroom ~1,898, but the kernel's `RLIMIT_NOFILE` clamped the window to ~1,000 — CloudWatch `ConcurrentExecutions` plateaued at 1,005 while a thousand workers of paid-for headroom idled; the wall took ~3 waves instead of 2, and (with biggest-first dispatch ordering) first-completion latency rose to 5.4 min because the small-shard buckets waited for wave two. `max_workers=1800` explicit was the workaround.

Fix: in `Run.dispatch`, when `transport == "event"` and `max_workers` is not explicit, compute the window from the account-concurrency probe **without** the `fd_safe_max_workers` term (keep it for sync, where it is load-bearing). Test: stub the probe, assert the event path's window equals account headroom while the sync path stays fd-clamped.

espg:
Bundled `small-fix` PR for the 2026-08-04 sweep.

Refs #372
Closes #374
Closes #375

## Phases

- [x] **Phase 1 — #375**: event-transport dispatch pacing no longer inherits the v1 fd ceiling.
- [x] **Phase 2 — #374**: point path records observations READ (pre-filter) alongside `n_obs`.
- [x] **Fold round (adversarial self-review of `e67edd8`)**: all six findings folded, one commit each — `8fc63b0`, `3473fe3`, `893f062`, `a977723`, `4177cff`, `f72e706`. Detail in the fold status comment on this thread.
- [ ] **#372** — conus shardmap builder count mislabel. **Cannot be implemented as written**; see "Questions for review" (1). No commit here touches it.
- [x] **Question (5) ruling**: shared-client `max_pool_connections` capped at the fd-safe bound regardless of transport — `058b9cb`.
- [x] **Fold round 2 (adversarial self-review of `058b9cb`)**: all six findings addressed — five commits (`8d47425`, `89c8009`, `4b609d2`, `28f9b31`, `1f297b5`) plus the question (5) scoping below. Detail in the inline replies.

## Phase 1 — #375: event-transport pacing window

`Run.dispatch(transport="event")` sized its pacing window through the same `_probe_workers` → `compute_available_workers` path as the sync transport, so the window was `min(n, fd_ceiling, account_headroom)`. The fd term exists because the v1 sync transport holds one socket per in-flight shard; the event transport holds **no** connections, so `RLIMIT_NOFILE` was clamping pure dispatch pacing.

**Approach.** The clamp composition stays in one place — `concurrency.compute_available_workers` grows a `fd_bound: bool = True` keyword rather than the caller re-deriving the window:

```python
if report.available is not None:
    workers = min(requested, report.available)
    if fd_bound:
        workers = min(workers, fd_safe_max_workers())
else:
    workers = min(requested, fd_safe_max_workers())
```

`Run.dispatch` threads the transport's connection model through `_probe_workers(session, n, fd_bound=transport == "sync")`. An explicit `max_workers` still wins verbatim and skips the probe, and the sync path is byte-identical.

**One judgement call, deliberately conservative:** when account headroom is **unreadable** (denied `GetAccountSettings`), the fd ceiling still applies even under `fd_bound=False`. It is then the only bound left, and an unbounded window would fire the whole shard set at once — exactly what the fleet's 60 s `MaximumEventAgeInSeconds` drops. Pinned by `test_unreadable_account_falls_back_to_fd_even_unbound`. Raised as question (2).

**Second-order effect, now documented** (fold `f72e706`): the event window is fired **serially** by the poller thread (`StatusPoller._dispatch_up_to_window`, `client_transport.py:901-903`), so a wider window lengthens the opening dispatch ramp proportionally — order 10 ms of invoke round-trip per shard, ~10 s → ~19 s in this issue's scenario, with no LIST running during it. That is the intended trade (the ramp grows by seconds; the clamp cost whole 900 s waves), and `dispatch`'s `max_workers` docstring now says so.

**Tests.** `tests/test_concurrency.py::TestFdBound` — the event window clears the fd ceiling (1900 vs a 992 ceiling) while the sync path on *identical* inputs stays clamped, so the divergence is attributable to `fd_bound` alone; account headroom still clamps without the fd term; `requested` still wins below headroom; the unreadable-account fallback above. `tests/test_client.py` pins the wiring itself, in two links that meet at the seam: `_probe_fd_bound` pins what `dispatch(transport=…)` hands `_probe_workers` (`False` for event, `True` for sync), and `test_probe_body_forwards_fd_bound` drives the **real** `_probe_workers` body and asserts the value it forwards to `compute_available_workers` is `[False, True]`, not the callee's default. The `_dispatch` helper's probe stub records the kwarg too, so `test_probe_sizes_the_pool` covers the sync direction end-to-end.

## Phase 2 — #374: observations READ on the point path

`n_obs` is the worker's `total_obs` — photons that survived the filters and landed in the shard. Photons fetched-but-discarded (read-plan `pad` segment padding, segments straddling the shard boundary, filter rejects) left no trace, so the read-vs-keep ratio — the direct measure of the #43 segment→photon indexed-IO efficiency — was not derivable from the run parquet.

**Approach — mirror the raster counters (issue #297), including their thread-safety argument.** The raster path threads a mutable `io_stats` dict down the read and accumulates in place. The point path does the same, with one adaptation: the granule fan-out is a `ThreadPoolExecutor` (issue #180), not an event loop, so a single shared dict would race. Instead **each granule gets its own sink**, allocated inside `_read_granule` and returned alongside its reads; the main thread sums them as it folds granules in order. Same no-lock argument the existing `group_errors` counter already relies on — written by the owning thread, read after the result is handed back.

Counting happens at the two places that actually decode base-rate rows, before any mask is applied:

| route | counted at | value |
|---|---|---|
| planned (`read_plan.spatial_index`) | `_execute_plan_group` | rows the plan decoded, including `pad` |
| full / flat, and the selectivity fallback | `_read_group_full` | the whole group |
| a-priori (`chunk_boundaries`) | whichever of the two it delegates to | as above |

Exactly one arm counts per group — `test_full_read_fallback_counts_once_not_twice` pins that the selectivity fallback does not double-count when it re-enters `_read_group_full` from inside the planned path.

**Absence means unmeasured, not zero** (the #297 nullable convention). `_record_obs_read` makes the *presence of the key* the "measured" signal, so a stubbed read seam or a worker predating the field reports `None` rather than a fabricated `0`, and `telemetry.merge`'s `_SUM_OR_NONE` disposition skips it instead of poisoning a mixed rollup.

That contract only held on the full-read path in the first cut. The planned and a-priori routes have **five exits before any base-rate decode** — `read.py:432`/`:442`/`:470` and `apriori.py:230`/`:241` — and none touched the sink, so a beam that short-circuits (the *common* per-beam outcome on the spatial-index route: a granule assigned for one beam misses on the other five) reported `None`, indistinguishable from an uninstrumented seam. Each now stamps `_record_obs_read(io_stats, 0)` (fold `893f062`); the numeric rollup is unchanged, since `_SUM_OR_NONE` skipping a `None` and summing a `0` agree.

Plumbing: `metadata["total_obs_read"]` (worker) → `n_obs_read` in `build_record`, `_SUM_OR_NONE_KEYS`, and `_ROW_SCALARS`. `deployment/aws/lambda_handler.py` already forwards worker metadata verbatim to `build_record`, so the field reaches the sidecar and run parquet with **no change to any deploy/infra file**.

**The stamp lands before the no-data early return** — a shard that decoded millions of photons and kept none is precisely the case the counter exists to expose, and stamping it only at the normal exit would have lost it. Caught in self-review (commit `e67edd8`), pinned by `test_stamped_on_the_no_data_shard`.

**`io_stats` is forwarded unconditionally by both index backends.** `HierarchicalIndex.read_group` briefly presence-gated it the way it gates `granule_url`; that gate protected nothing (the worker always passes a dict — `worker.py:422`/`:436`) and diverged from `InlineIndex`, which forwards it always. Dropped in fold `a977723`; `granule_url`'s gate is genuinely load-bearing and stays.

**One verb, one meaning in the shard log.** The new pre-filter count first shipped as a second `Read N observations` line, colliding with the pre-existing `worker.py:673`/`:676` lines that mean *kept* and are what CloudWatch is grepped for today. It now reads `Decoded {n} observations pre-filter, kept {m}` (fold `4177cff`).

**Tests.** `tests/test_processing.py::TestPlannedReadGroup` — padded decode counts 8 read against 4 kept on the existing 6-segment/12-photon fixture; per-granule accumulation across groups; the full-read route counting the whole group (12 decoded to keep 1); the fallback counting once; `None` sink counting nothing; an instrumented route that decodes-then-masks-everything reporting measured zero; and `test_planned_short_circuits_read_as_measured_zero` over all three pre-decode exits (empty AOI, empty coarse level, link counts summing to zero base rows). `tests/test_apriori.py::TestAprioriReadGroup::test_short_circuits_read_as_measured_zero` does the same for the a-priori pair. `tests/test_processing.py::TestObsReadMetadata` — the worker's half: sums across granules and exceeds kept, omits the key when the seam never measured, hands each granule a distinct sink, stamps the no-data shard, and emits exactly one `Read … observations` line per shard. `tests/test_index.py::TestHierarchicalDelegation::test_io_stats_forwarded_unconditionally` — the sink reaches `_read_group` on every call, `None` included. `tests/test_telemetry.py` — record field, measured-zero vs absent, merge sums, back-compat row folds without poisoning, parquet column.

## How it was tested

Full local gate on the final commit (`1f297b5`):

- `uv run pytest -q` → **3290 passed, 38 skipped**, 0 failures (`test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` deselected — it fails identically on the base commit `f72e706` in this environment, env-dependent Docker build, untouched by this PR).
- `uv run ruff check src tests` → clean except the **pre-existing** `N818` on `src/zagg/registry.py:64` (`UnknownCapability`), untouched by this PR.
- `uv run ruff format --check src tests` → clean except the **pre-existing** diff on `tests/data/benchmark/README.md`, untouched by this PR.

Both pre-existing failures reproduce on `main` at 66cb111 and are deliberately left alone per the "don't fix unrelated CI failures" rule.

Every fold commit that changed behavior was **mutation-verified**: the fix reverted locally, the new test observed to fail, the fix restored. Per-finding detail in the inline replies.

## Questions for review

1. **#372 cannot be implemented on `main` as written — it is the reason this PR is `waiting`.** The issue names `data/conus/build_conus_shardmap.py:254` and `data/build_aoi_shardmap.py`, but at `main` (66cb111) there is **no `data/` directory in the repository at all** — and it is not gitignored (`.gitignore` has no `data/` entry, so the files are genuinely absent, not merely untracked). The premise field is missing too: the issue says the fix is "one-line now that `ShardMap.build` records the distinct assigned count (landed with the demo-week `granules_assigned` metadata addition)", but `granules_assigned` appears nowhere in the tree — `ShardMap.build` writes `total_granules`/`total_shards`/`total_pairs`/`build_wall_s` only (`src/zagg/catalog/shardmap.py:534-542`), and a repo-wide search for `granules_assigned` and `granules_intersecting_conus` returns nothing. The closest live analogue to the hand-computed distinct set is `src/zagg/sweep.py:447` (`meta["total_granules"] = len({e["id"] for g in granules for e in g})`), which is not what the issue asks to change. Which do you want?
   - (1) The scripts live outside this repo (a demo/scratch tree) — close #372 here, or move it to that repo.
   - (2) They are meant to land here and haven't yet — #372 waits on the PR that adds them.
   - (3) You actually want `ShardMap.build` to record a `granules_assigned` distinct count first (the prerequisite the issue assumes) — a behavioral change to the shardmap builder, well beyond `small-fix`, so it wants its own issue/PR rather than being folded in here.

   Referenced with `Refs`, not `Closes`, so merging will not silently close it.

2. **#375, the unreadable-account fallback.** Keeping the fd clamp when the probe is denied is my call, not the issue's text (which says "compute the window from the account-concurrency probe without the `fd_safe_max_workers` term"). The literal reading gives a denied probe under `transport="event"` a window of `n` — dispatching every shard at once on exactly the runs where headroom is unknown. Self-review confirmed the code does what this says and let the call stand, with one caveat worth recording: on that branch the FD ceiling is no longer bounding anything physical (the event transport holds no sockets), so it is an **arbitrary-but-bounded pacing proxy** — the right shape of answer, but the number itself carries no meaning. Say the word if you want the literal reading instead, or a named pacing constant in place of the FD ceiling.

3. **#374 changes the `_read_group` call contract: `io_stats` is passed ALWAYS, not feature-gated** — and both index backends now agree on that. `granule_url` stays gated on the a-priori feature being configured; `tests/test_apriori.py::TestWorkerSeam::test_granule_url_absent_when_flag_off` pins that gate on the *received kwargs*, so it survives future always-on kwargs (mutation-verified in review: removing the gate still fails it). `io_stats` is always-on instrumentation matching the raster sink, so `HierarchicalIndex.read_group` now forwards it unconditionally like `InlineIndex` always did — the presence-gate it briefly carried protected nothing, since the worker always passes a dict (`worker.py:422`/`:436`), and `_record_obs_read` already no-ops on `None` for direct callers. Nothing to decide here unless you disagree that the sink should be always-on.

4. **`src/zagg/client.py` is over the §4 size limit — pre-existing, and I did not split it.** It was **1211** lines at 66cb111 (already past ~1,200); this PR adds 24, almost all docstring on the existing `_probe_workers`/`dispatch`, taking it to **1235**. §4 says raise it rather than split, so: flagging, not splitting. A split is its own PR.

5. **Resolved — ruled "fix it here, now" ([ruling](https://github.com/englacial/zagg/pull/378#issuecomment-5185073447), row (5)); landed in `058b9cb`.** The shared client's declared pool is now `max_pool_connections=min(workers, concurrency.fd_safe_max_workers())` regardless of transport, so it can never exceed `RLIMIT_NOFILE` even when the event pacing window legitimately does (`fd_bound=False`). What it declares is a **retention** bound, not an EMFILE guard: botocore passes only `maxsize` to urllib3 (`block=False`), so an oversized declaration allocates nothing by itself — it is the number of sockets the pool would keep once a concurrent dispatch loop opened them.

   **Scope of "no behavioral change" (review finding, fold round 2).** Inert on both **probe-driven** paths: the sync probe already applies the fd term inside `compute_available_workers`, so `min()` is a no-op, and today's serial event dispatch loop holds at most one connection. It is **not** literally inert on the **explicit `max_workers`** path, which skips the probe on either transport and so carries no fd term: `dispatch(max_workers=300)` under `ulimit -n 256` now declares a 224-connection pool against a 300-wide thread pool, i.e. urllib3 "Connection pool is full, discarding connection" churn where it previously retained one connection per thread. That is exactly what the ruling asks for ("regardless of transport") and is arguably the better behavior — recording it here because it is a change, not nothing.

   Pinned by `tests/test_client.py::TestConcurrencyPreflight::test_pool_cap_clamped_when_window_exceeds_fd_bound` and `test_pool_cap_unchanged_below_fd_bound`, which assert the declared pool **and** the fan-out width together (`(pool, width) == (2, 3)`): the clamp must stop at the pool and must not re-clamp the pacing window, which would undo #375. Mutation-verified both ways — reverting the cap to `max_pool_connections=workers` fails the clamp test, and moving the fd term up into `workers` fails the width assertion.

### DIFF
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -61,7 +61,6 @@ from concurrent.futures import wait as futures_wait
 from dataclasses import asdict
 from typing import Any, Iterator
 
-from zagg.concurrency import fd_safe_max_workers
 from zagg.config import (
     PipelineConfig,
     get_child_order,
@@ -733,7 +732,7 @@ class Run:
             writes status objects (issue #327) and read access to the status
             prefix for the poll.
         """
-        from zagg import runner
+        from zagg import concurrency, runner
 
         if transport not in ("sync", "event"):
             raise ValueError(f'transport must be "sync" or "event", got {transport!r}')
@@ -767,6 +766,7 @@ class Run:
             # connection pool tracks the fan-out width, but must never exceed
             # the fd ceiling regardless of pacing window — the event window
             # may clear RLIMIT_NOFILE (issue #375), a pool of sockets may not.
+            # Through the module seam, so the fd ceiling has one patch point.
             client = session.client(
                 "lambda",
                 region_name=self.region,
@@ -774,7 +774,7 @@ class Run:
                     read_timeout=960,
                     connect_timeout=10,
                     retries={"max_attempts": 0},
-                    max_pool_connections=min(workers, fd_safe_max_workers()),
+                    max_pool_connections=min(workers, concurrency.fd_safe_max_workers()),
                 ),
             )
             invoked_by = runner._resolve_invoked_by(session, self.region)
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -709,7 +709,12 @@ class Run:
             held socket per in-flight shard), while the event window — which
             holds no connections — is sized by account headroom alone, so a
             low ``ulimit -n`` no longer strands concurrency the account has
-            already granted.
+            already granted. The window is fired serially by the poller thread
+            (:meth:`~zagg.client_transport.StatusPoller._dispatch_up_to_window`),
+            so a wider window also lengthens the opening dispatch ramp
+            proportionally — roughly 10 ms of invoke round-trip per shard, and
+            no LIST runs during it. Deliberate: the ramp grows by seconds while
+            the clamp cost whole 900 s waves (issue #375).
         transport : str
             ``"sync"`` (default, the v1 transport): a thread pool over
             synchronous ``RequestResponse`` invokes — one held connection per
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -631,7 +631,7 @@ class Run:
 
     # -- dispatch -----------------------------------------------------------
 
-    def _probe_workers(self, session, n: int) -> int:
+    def _probe_workers(self, session, n: int, *, fd_bound: bool = True) -> int:
         """Account-concurrency preflight, degrading to the default on denial.
 
         Runs ``agg``'s own :func:`~zagg.concurrency.compute_available_workers`
@@ -639,6 +639,11 @@ class Run:
         size, so a notebook fan-out is sized by what the account can actually
         take rather than by a hardcoded guess (espg ruling on PR #333: whoever
         can invoke the workers normally can read the concurrency too).
+
+        ``fd_bound`` carries the transport's connection model through to the
+        probe: the sync transport holds one socket per in-flight shard, so its
+        window is FD-bounded; the event transport holds none and passes
+        ``False`` (issue #375).
 
         Fail-open by design: ANY probe failure — ``AccessDenied`` under a
         narrow-permission deployment (the cryocloud case, where the worker role
@@ -656,6 +661,7 @@ class Run:
                 session.client("lambda", region_name=self.region),
                 session.client("cloudwatch", region_name=self.region),
                 self.function_name,
+                fd_bound=fd_bound,
             )
             runner._log_concurrency_report(report, workers)
             return workers
@@ -698,6 +704,12 @@ class Run:
             shards may be dispatched-but-unresolved at once, load-bearing
             because the fleet's 60 s ``MaximumEventAgeInSeconds`` silently
             drops an Event invoke the account has no concurrency to start.
+            The two transports therefore probe differently (issue #375): the
+            sync pool is clamped by the local ``RLIMIT_NOFILE`` ceiling (one
+            held socket per in-flight shard), while the event window — which
+            holds no connections — is sized by account headroom alone, so a
+            low ``ulimit -n`` no longer strands concurrency the account has
+            already granted.
         transport : str
             ``"sync"`` (default, the v1 transport): a thread pool over
             synchronous ``RequestResponse`` invokes — one held connection per
@@ -735,7 +747,14 @@ class Run:
             from botocore.config import Config
 
             session = boto3.Session()
-            requested = max_workers if max_workers is not None else self._probe_workers(session, n)
+            # The FD ceiling bounds the SYNC pool (one held socket per in-flight
+            # shard); the event transport holds no connections, so its pacing
+            # window comes from account headroom alone (issue #375).
+            requested = (
+                max_workers
+                if max_workers is not None
+                else self._probe_workers(session, n, fd_bound=transport == "sync")
+            )
             workers = max(1, min(requested, n))
             # read_timeout must exceed the 900 s function ceiling (same
             # rationale as agg's preflight-built client, issue #148); the
--- a/src/zagg/client.py
+++ b/src/zagg/client.py
@@ -61,6 +61,7 @@ from concurrent.futures import wait as futures_wait
 from dataclasses import asdict
 from typing import Any, Iterator
 
+from zagg.concurrency import fd_safe_max_workers
 from zagg.config import (
     PipelineConfig,
     get_child_order,
@@ -763,7 +764,9 @@ class Run:
             workers = max(1, min(requested, n))
             # read_timeout must exceed the 900 s function ceiling (same
             # rationale as agg's preflight-built client, issue #148); the
-            # connection pool tracks the fan-out width.
+            # connection pool tracks the fan-out width, but must never exceed
+            # the fd ceiling regardless of pacing window — the event window
+            # may clear RLIMIT_NOFILE (issue #375), a pool of sockets may not.
             client = session.client(
                 "lambda",
                 region_name=self.region,
@@ -771,7 +774,7 @@ class Run:
                     read_timeout=960,
                     connect_timeout=10,
                     retries={"max_attempts": 0},
-                    max_pool_connections=workers,
+                    max_pool_connections=min(workers, fd_safe_max_workers()),
                 ),
             )
             invoked_by = runner._resolve_invoked_by(session, self.region)
--- a/src/zagg/index/hierarchical.py
+++ b/src/zagg/index/hierarchical.py
@@ -20,7 +20,17 @@ class HierarchicalIndex(VirtualIndex):
     # No backend-specific keys: today's path is configured by the existing
     # ``data_source.read_plan`` / ``levels`` surface, not the index block.
 
-    def read_group(self, h5obj, group, data_source, shard_key, grid, arrow=False, granule_url=None):
+    def read_group(
+        self,
+        h5obj,
+        group,
+        data_source,
+        shard_key,
+        grid,
+        arrow=False,
+        granule_url=None,
+        io_stats=None,
+    ):
         # Resolve through the package namespace at call time (not an import-time
         # binding) so tests that ``monkeypatch.setattr("zagg.processing._read_group",
         # ...)`` keep intercepting the worker's reads, exactly as before the seam.
@@ -29,7 +39,11 @@ class HierarchicalIndex(VirtualIndex):
         # ``granule_url`` is forwarded only when set (the a-priori arm), so
         # monkeypatched ``_read_group`` fakes keep their existing signature on
         # every other path -- mirroring the worker's presence-gated kwarg.
+        # ``io_stats`` (issue #374) is gated the same way, for the same reason:
+        # a fake that takes neither keeps working, and simply reports no count.
         kwargs = {"arrow": arrow}
         if granule_url is not None:
             kwargs["granule_url"] = granule_url
+        if io_stats is not None:
+            kwargs["io_stats"] = io_stats
         return _processing._read_group(h5obj, group, data_source, shard_key, grid, **kwargs)
--- a/src/zagg/index/hierarchical.py
+++ b/src/zagg/index/hierarchical.py
@@ -39,11 +39,13 @@ class HierarchicalIndex(VirtualIndex):
         # ``granule_url`` is forwarded only when set (the a-priori arm), so
         # monkeypatched ``_read_group`` fakes keep their existing signature on
         # every other path -- mirroring the worker's presence-gated kwarg.
-        # ``io_stats`` (issue #374) is gated the same way, for the same reason:
-        # a fake that takes neither keeps working, and simply reports no count.
-        kwargs = {"arrow": arrow}
+        # ``io_stats`` (issue #374) is NOT gated: it is always-on
+        # instrumentation, the worker always passes a dict (``worker.py:422``,
+        # ``worker.py:436``), and ``_record_obs_read`` no-ops on ``None`` for
+        # the direct callers that omit it -- so gating it would protect nothing
+        # while diverging from ``InlineIndex.read_group``, which forwards it
+        # unconditionally (review finding).
+        kwargs = {"arrow": arrow, "io_stats": io_stats}
         if granule_url is not None:
             kwargs["granule_url"] = granule_url
-        if io_stats is not None:
-            kwargs["io_stats"] = io_stats
         return _processing._read_group(h5obj, group, data_source, shard_key, grid, **kwargs)
--- a/src/zagg/processing/read.py
+++ b/src/zagg/processing/read.py
@@ -382,7 +382,9 @@ def _planned_read_group(
 
     ``io_stats`` (issue #374) is the optional read-counter sink, forwarded to
     whichever arm actually decodes base-rate rows — the full-read fallback or
-    :func:`_execute_plan_group` — so a group is counted exactly once.
+    :func:`_execute_plan_group` — so a group is counted exactly once. The
+    pre-decode short-circuits stamp a measured zero themselves, so this route
+    never reports "unmeasured" for a group it did look at.
     """
     levels = data_source["levels"]
     base_level_key = data_source["base_level"]
@@ -429,7 +431,14 @@ def _planned_read_group(
     ibeg_arr = coarse_data[ibeg_path]
     cnt_arr = coarse_data[cnt_path]
 
+    # The three short-circuits below return before any base-rate decode, so
+    # each stamps a measured zero (issue #374): without it the planned route --
+    # the production route for ATL03, where a granule assigned for one beam
+    # short-circuits on the other five -- reports ``n_obs_read = None``,
+    # indistinguishable from an uninstrumented read seam. That discrimination
+    # is the whole point of the nullable column (review finding).
     if len(coarse_lats) == 0:
+        _record_obs_read(io_stats, 0)
         return None
 
     # ``n_base`` under #43's contiguity assumption ("ranges do not overlap and
@@ -440,6 +449,7 @@ def _planned_read_group(
     # either form under- or over-estimates -- track via a follow-up to #43.
     n_base = int(np.asarray(cnt_arr).sum())
     if n_base <= 0:
+        _record_obs_read(io_stats, 0)
         return None
 
     # Match segments to this shard with the SAME mortie test the photon path
@@ -468,6 +478,7 @@ def _planned_read_group(
     )
 
     if not plan.parent_runs:
+        _record_obs_read(io_stats, 0)
         return None  # empty AOI -- no parent intersects, skip the group entirely
 
     if plan.full_read:
@@ -571,7 +582,9 @@ def _execute_plan_group(
     # straddling reads counted BEFORE the shard mask and filters below, so the
     # segment→photon indexed-IO efficiency (issue #43) is derivable from the
     # run parquet. Recorded ahead of the empty short-circuit so an instrumented
-    # route that decodes nothing still reads as measured-zero, not unmeasured.
+    # route that decodes nothing still reads as measured-zero, not unmeasured
+    # -- as do the pre-decode short-circuits in ``_planned_read_group`` and
+    # ``_apriori_read_group``, which never reach this line.
     _record_obs_read(io_stats, len(lats))
 
     if len(lats) == 0:
--- a/src/zagg/processing/read.py
+++ b/src/zagg/processing/read.py
@@ -321,6 +321,23 @@ def _level_coord_paths(level: dict, group: str) -> tuple[str, str]:
     return lat_path, lon_path
 
 
+def _record_obs_read(io_stats: dict | None, n: int) -> None:
+    """Accumulate base-rate rows DECODED (pre-filter) for the read counter.
+
+    The point-path counterpart of the raster ``io_stats`` counters (issue #374
+    / issue #297): rows fetched, before the shard mask, structured/expression
+    filters, and segment padding are applied — so ``n_obs_read / n_obs`` is the
+    read-vs-keep ratio, derived at read time and never stored.
+
+    PRESENCE of the key is the "measured" signal, not its value: a read seam
+    that never calls this (a stubbed ``_read_group``, or a worker predating the
+    field) leaves it absent, which the worker reports as ``None`` — unmeasured,
+    per the #297 nullable-column convention — rather than a real zero.
+    """
+    if io_stats is not None:
+        io_stats["obs_read"] = io_stats.get("obs_read", 0) + int(n)
+
+
 def _planned_read_group(
     h5obj,
     group: str,
@@ -329,6 +346,7 @@ def _planned_read_group(
     grid,
     arrow: bool = False,
     read_fn=None,
+    io_stats: dict | None = None,
 ):
     """Planned (AOI-bounded) read of one HDF5 group via the coarse spatial index.
 
@@ -361,6 +379,10 @@ def _planned_read_group(
     reads; ``None`` (default) uses the plain h5coro bridge — the hierarchical
     baseline. Selection semantics (the plan) and everything downstream of the
     returned columns are identical regardless.
+
+    ``io_stats`` (issue #374) is the optional read-counter sink, forwarded to
+    whichever arm actually decodes base-rate rows — the full-read fallback or
+    :func:`_execute_plan_group` — so a group is counted exactly once.
     """
     levels = data_source["levels"]
     base_level_key = data_source["base_level"]
@@ -454,11 +476,27 @@ def _planned_read_group(
         # ``read_fn`` rides along (issue #179): the fallback keeps the compiled
         # + pooled addressing instead of dropping to serial h5coro.
         return _read_group_full(
-            h5obj, group, data_source, shard_key, grid, arrow=arrow, read_fn=read_fn
+            h5obj,
+            group,
+            data_source,
+            shard_key,
+            grid,
+            arrow=arrow,
+            read_fn=read_fn,
+            io_stats=io_stats,
         )
 
     return _execute_plan_group(
-        h5obj, group, data_source, shard_key, grid, plan, n_base, arrow, read_fn=read_fn
+        h5obj,
+        group,
+        data_source,
+        shard_key,
+        grid,
+        plan,
+        n_base,
+        arrow,
+        read_fn=read_fn,
+        io_stats=io_stats,
     )
 
 
@@ -472,6 +510,7 @@ def _execute_plan_group(
     n_base,
     arrow=False,
     read_fn=None,
+    io_stats: dict | None = None,
 ):
     """Execute a computed :class:`~zagg.read_plan.ReadPlan` over one group.
 
@@ -527,6 +566,13 @@ def _execute_plan_group(
         workers,
     )
     lats, lons = coord_arrays[lat_path], coord_arrays[lon_path]
+
+    # Rows DECODED by this group's plan (issue #374): the padded, boundary-
+    # straddling reads counted BEFORE the shard mask and filters below, so the
+    # segment→photon indexed-IO efficiency (issue #43) is derivable from the
+    # run parquet. Recorded ahead of the empty short-circuit so an instrumented
+    # route that decodes nothing still reads as measured-zero, not unmeasured.
+    _record_obs_read(io_stats, len(lats))
 
     if len(lats) == 0:
         return None
@@ -671,6 +717,7 @@ def _read_group(
     grid,
     arrow: bool = False,
     granule_url: str | None = None,
+    io_stats: dict | None = None,
 ):
     """Read and spatially filter one HDF5 group.
 
@@ -706,6 +753,12 @@ def _read_group(
     Empty-AOI groups short-circuit to ``None``. Selectivity above the configured
     threshold falls back to the full-read path; the planned and full paths
     produce row-for-row identical output (#43 Phase C parity).
+
+    ``io_stats`` (issue #374) is an optional mutable dict the read routes
+    accumulate ``obs_read`` (rows decoded, pre-filter) into. The worker owns
+    one per granule, so it is written only by that granule's thread and read
+    by the dispatcher after the read completes — no lock (the point-path
+    analogue of the raster ``io_stats`` sink). ``None`` counts nothing.
     """
     rp = data_source.get("read_plan")
     # Presence check, not truthiness: an empty/misconfigured ``chunk_boundaries``
@@ -715,7 +768,14 @@ def _read_group(
         from zagg.processing.apriori import _apriori_read_group
 
         return _apriori_read_group(
-            h5obj, group, data_source, shard_key, grid, arrow=arrow, granule_url=granule_url
+            h5obj,
+            group,
+            data_source,
+            shard_key,
+            grid,
+            arrow=arrow,
+            granule_url=granule_url,
+            io_stats=io_stats,
         )
     # Truthy-checking ``levels``/``base_level`` would route an empty ``{}`` (a
     # config typo, easy to do) back to the full-read path silently. Reject
@@ -724,8 +784,12 @@ def _read_group(
     # multi-level structure to operate on.
     if isinstance(rp, dict) and rp.get("spatial_index"):
         _validate_planned_config(data_source)
-        return _planned_read_group(h5obj, group, data_source, shard_key, grid, arrow=arrow)
-    return _read_group_full(h5obj, group, data_source, shard_key, grid, arrow=arrow)
+        return _planned_read_group(
+            h5obj, group, data_source, shard_key, grid, arrow=arrow, io_stats=io_stats
+        )
+    return _read_group_full(
+        h5obj, group, data_source, shard_key, grid, arrow=arrow, io_stats=io_stats
+    )
 
 
 def _validate_planned_config(data_source: dict) -> None:
@@ -778,7 +842,14 @@ def _read_paths_pooled(entries, read_one, workers: int) -> dict:
 
 
 def _read_group_full(
-    h5obj, group: str, data_source: dict, shard_key: int, grid, arrow: bool = False, read_fn=None
+    h5obj,
+    group: str,
+    data_source: dict,
+    shard_key: int,
+    grid,
+    arrow: bool = False,
+    read_fn=None,
+    io_stats: dict | None = None,
 ):
     """Full-coord-read variant of :func:`_read_group` (the pre-#49-Phase-C path).
 
@@ -797,6 +868,10 @@ def _read_group_full(
     — so index backends (``inline``) can route this path's decode through the
     compiled reader too, giving read-plan-less (flat) data sources the fast
     decode. ``None`` keeps the batched h5coro reads byte-identical.
+
+    ``io_stats`` (issue #374) is the optional read-counter sink; this path
+    decodes the base-rate coordinates in FULL, so what it records is the whole
+    group's row count regardless of how few rows survive the shard mask.
     """
     coordinates = data_source["coordinates"]
     variables = data_source["variables"]
@@ -835,6 +910,10 @@ def _read_group_full(
     lon_path = coordinates["longitude"].format(group=group)
     lats = coord_data[lat_path]
     lons = coord_data[lon_path]
+
+    # Rows DECODED by this group (issue #374) — the full base-rate coordinate
+    # read, counted before the shard mask and filters below.
+    _record_obs_read(io_stats, len(lats))
 
     if len(lats) == 0:
         return None
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -235,8 +235,9 @@ def process_shard(
     -------
     (DataFrame, metadata)
         DataFrame in canonical chunk order; metadata dict with ``shard_key``,
-        ``cells_with_data``, ``total_obs``, ``granule_count``,
-        ``files_processed``, ``duration_s``, ``error``. Ragged fields are
+        ``cells_with_data``, ``total_obs``, ``total_obs_read`` (issue #374 —
+        rows decoded pre-filter, present only when a read route measured it),
+        ``granule_count``, ``files_processed``, ``duration_s``, ``error``. Ragged fields are
         delivered out-of-band via ``ragged_out`` (above), not in this tuple. At
         K>1 the per-chunk carriers + ragged are delivered via ``chunk_results``.
     """
@@ -403,17 +404,22 @@ def process_shard(
         → close, all in the calling thread (issue #180 — under the pool each
         granule gets its own ``H5Coro``, never shared across threads).
 
-        Returns ``(reads, group_errors)``: ``reads`` is the carriers of the
-        groups that returned data (group order, ``None`` — legitimately empty
-        — groups dropped); ``group_errors`` counts raised group reads, warned
-        here but folded into ``read_errors`` by the main thread (a shared
-        ``+= 1`` from worker threads could race). Raises when the granule
-        itself fails (e.g. the open) — the caller warns and skips it, shard
-        continues (issue #116 semantics).
+        Returns ``(reads, group_errors, io_stats)``: ``reads`` is the carriers
+        of the groups that returned data (group order, ``None`` — legitimately
+        empty — groups dropped); ``group_errors`` counts raised group reads,
+        warned here but folded into ``read_errors`` by the main thread (a
+        shared ``+= 1`` from worker threads could race); ``io_stats`` is this
+        granule's read counters (issue #374), allocated FRESH here so it is
+        written only by the thread that owns the granule and read by the main
+        thread after the result is handed back — the same no-lock argument as
+        ``group_errors``, rather than a shared dict the pool would race on.
+        Raises when the granule itself fails (e.g. the open) — the caller warns
+        and skips it, shard continues (issue #116 semantics).
         """
         h5obj = None
         reads: list = []
         group_errors = 0
+        io_stats: dict = {}
         try:
             resource_path = _rewrite_url(s3_url)
 
@@ -427,7 +433,7 @@ def process_shard(
 
             for g in data_source["groups"]:
                 try:
-                    read_kwargs = {"arrow": use_arrow}
+                    read_kwargs = {"arrow": use_arrow, "io_stats": io_stats}
                     if apriori:
                         read_kwargs["granule_url"] = s3_url
                     chunk = index_backend.read_group(
@@ -454,7 +460,7 @@ def process_shard(
                 # #175) on a path that never fails the read.
                 logger.warning(f"  index backend finish_granule failed for {s3_url}: {e}")
 
-            return reads, group_errors
+            return reads, group_errors, io_stats
         finally:
             # Release this granule's h5coro cache before the next one (issue #66):
             # without it each granule's unevicted cache stays resident for the whole
@@ -471,7 +477,7 @@ def process_shard(
                     logger.debug("h5coro cache release failed", exc_info=True)
 
     def _iter_granule_reads():
-        """Yield ``(s3_url, reads, group_errors)`` in original ``granule_urls`` order.
+        """Yield ``(s3_url, reads, group_errors, io_stats)`` in ``granule_urls`` order.
 
         ``granule_workers == 1`` reads each granule in this thread — the
         unchanged serial loop. Above 1, up to ``granule_workers`` granules are
@@ -502,7 +508,7 @@ def process_shard(
                 while in_flight:
                     s3_url, future = in_flight.popleft()
                     try:
-                        reads, group_errors = future.result()
+                        reads, group_errors, granule_io = future.result()
                     except Exception as e:
                         granule_errors += 1
                         _record_read_error(f"processing file {s3_url}", e)
@@ -515,11 +521,21 @@ def process_shard(
                     for u in islice(urls, 1):
                         in_flight.append((u, pool.submit(_read_granule, u)))
                     if reads is not None:
-                        yield s3_url, reads, group_errors
+                        yield s3_url, reads, group_errors, granule_io
+
+    # Observations READ (issue #374): base-rate rows decoded across every
+    # granule BEFORE the shard mask, filters, and read-plan segment padding
+    # cut them down — the numerator of the read-vs-keep ratio whose
+    # denominator is ``total_obs``. ``None`` until some read route reports a
+    # count, so a stubbed/older read seam reads as UNMEASURED rather than a
+    # real zero (the issue #297 nullable-counter convention).
+    obs_read_total: int | None = None
 
     # Read files and filter spatially, folding granules in original order.
-    for s3_url, reads, group_errors in _iter_granule_reads():
+    for s3_url, reads, group_errors, granule_io in _iter_granule_reads():
         read_errors += group_errors
+        if "obs_read" in granule_io:
+            obs_read_total = (obs_read_total or 0) + int(granule_io["obs_read"])
         try:
             for chunk in reads:
                 if buffered is not None:
@@ -804,6 +820,11 @@ def process_shard(
 
     metadata["cells_with_data"] = cells_with_data
     metadata["total_obs"] = n_obs_total
+    # Pre-filter read volume (issue #374), stamped only when a read route
+    # actually measured it — absence is "unmeasured", not zero.
+    if obs_read_total is not None:
+        metadata["total_obs_read"] = obs_read_total
+        logger.info(f"  Read {obs_read_total:,} observations, kept {n_obs_total:,}")
     metadata["duration_s"] = duration
 
     # K==1: deliver the lone chunk's carrier as the 2-tuple ``df_out`` and its
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -584,6 +584,14 @@ def process_shard(
         buffered.flush()
     phase_timings["read"] = time.time() - _read_t0
 
+    # Pre-filter read volume (issue #374), stamped as soon as the read loop is
+    # done so it survives the no-data early return below — a shard that decoded
+    # millions of photons and kept NONE is precisely the read-vs-keep case this
+    # counter exists to expose. Absent when no read route measured it, so
+    # absence stays "unmeasured" rather than a fabricated zero.
+    if obs_read_total is not None:
+        metadata["total_obs_read"] = obs_read_total
+
     if buffered.empty if buffered is not None else not all_reads:
         # Distinguish a genuinely-empty read from one where a group read raised
         # (issue #116): a raised read is a real error masquerading as "no data",
@@ -820,10 +828,7 @@ def process_shard(
 
     metadata["cells_with_data"] = cells_with_data
     metadata["total_obs"] = n_obs_total
-    # Pre-filter read volume (issue #374), stamped only when a read route
-    # actually measured it — absence is "unmeasured", not zero.
-    if obs_read_total is not None:
-        metadata["total_obs_read"] = obs_read_total
+    if obs_read_total is not None:  # stamped above, before the no-data return
         logger.info(f"  Read {obs_read_total:,} observations, kept {n_obs_total:,}")
     metadata["duration_s"] = duration
 
--- a/src/zagg/processing/worker.py
+++ b/src/zagg/processing/worker.py
@@ -828,9 +828,12 @@ def process_shard(
 
     metadata["cells_with_data"] = cells_with_data
     metadata["total_obs"] = n_obs_total
-    if obs_read_total is not None:  # stamped above, before the no-data return
-        logger.info(f"  Read {obs_read_total:,} observations, kept {n_obs_total:,}")
     metadata["duration_s"] = duration
+    # "Read" already means "kept" in the two lines above (worker.py:673/:676),
+    # which is what CloudWatch greps for today, so this one says "Decoded" —
+    # one verb, one meaning (review finding, issue #374).
+    if obs_read_total is not None:
+        logger.info(f"  Decoded {obs_read_total:,} observations pre-filter, kept {n_obs_total:,}")
 
     # K==1: deliver the lone chunk's carrier as the 2-tuple ``df_out`` and its
     # ragged via ``ragged_out`` (unchanged contract). K>1: the carriers + ragged
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -499,13 +499,16 @@ class TestConcurrencyPreflight:
         # lazily, but the declared cap is what turns into EMFILE if the loop
         # ever goes concurrent (PR #378, question (5) ruling).
         monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 2)
-        _, _, pool = self._dispatch(catalog, monkeypatch, max_workers=5)
-        assert pool == 2
+        width, _, pool = self._dispatch(catalog, monkeypatch, max_workers=5)
+        # The clamp stops at the DECLARED pool: the fan-out width is still the
+        # shard-count clamp of the requested 5, un-touched by the fd bound --
+        # re-clamping it there would undo issue #375.
+        assert (pool, width) == (2, 3)
 
     def test_pool_cap_unchanged_below_fd_bound(self, catalog, monkeypatch):
         monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 100)
-        _, _, pool = self._dispatch(catalog, monkeypatch, max_workers=2)
-        assert pool == 2
+        width, _, pool = self._dispatch(catalog, monkeypatch, max_workers=2)
+        assert (pool, width) == (2, 2)
 
     @staticmethod
     def _probe_fd_bound(catalog, monkeypatch, *, transport):
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -18,6 +18,7 @@ from concurrent.futures import ThreadPoolExecutor
 import pytest
 
 import zagg.client
+import zagg.concurrency
 from zagg.client import Run, RunHandle, ShardError
 from zagg.config import PipelineConfig, default_config
 
@@ -504,7 +505,7 @@ class TestConcurrencyPreflight:
         # `maxsize` to urllib3 (block=False), so an oversized declaration
         # allocates nothing on its own -- it is the number of sockets the
         # pool would KEEP once a concurrent dispatch loop opened them.
-        monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 2)
+        monkeypatch.setattr(zagg.concurrency, "fd_safe_max_workers", lambda: 2)
         width, _, pool = self._dispatch(catalog, monkeypatch, max_workers=5)
         # The clamp stops at the DECLARED pool: the fan-out width is still the
         # shard-count clamp of the requested 5, un-touched by the fd bound --
@@ -512,7 +513,7 @@ class TestConcurrencyPreflight:
         assert (pool, width) == (2, 3)
 
     def test_pool_cap_unchanged_below_fd_bound(self, catalog, monkeypatch):
-        monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 100)
+        monkeypatch.setattr(zagg.concurrency, "fd_safe_max_workers", lambda: 100)
         width, _, pool = self._dispatch(catalog, monkeypatch, max_workers=2)
         assert (pool, width) == (2, 2)
 
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -493,11 +493,16 @@ class TestConcurrencyPreflight:
         assert seen == [False, True]
 
     def test_pool_cap_clamped_when_window_exceeds_fd_bound(self, catalog, monkeypatch):
-        # The pacing window may clear RLIMIT_NOFILE (fd_bound=False, issue
-        # #375), but the shared client's DECLARED pool never may: today the
-        # serial dispatch loop holds <=1 connection and urllib3 sizes pools
-        # lazily, but the declared cap is what turns into EMFILE if the loop
-        # ever goes concurrent (PR #378, question (5) ruling).
+        # A fan-out width above RLIMIT_NOFILE must not reach the shared
+        # client's DECLARED pool (PR #378, question (5) ruling). Forced here
+        # by an explicit max_workers, which skips the probe so `workers`
+        # carries no fd term at all -- the same shape the event transport
+        # reaches through the probe when it passes fd_bound=False (issue
+        # #375; that wiring is pinned by test_event_window_is_not_fd_bound,
+        # which this test deliberately does not re-derive). Today the serial
+        # dispatch loop holds <=1 connection and urllib3 sizes pools lazily,
+        # but the declared cap is what turns into EMFILE if the loop ever
+        # goes concurrent.
         monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 2)
         width, _, pool = self._dispatch(catalog, monkeypatch, max_workers=5)
         # The clamp stops at the DECLARED pool: the fan-out width is still the
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -357,7 +357,7 @@ class TestConcurrencyPreflight:
         if probe is not None:
 
             def _probe(requested, lambda_client, cloudwatch_client, function_name, **k):
-                calls.append((requested, function_name))
+                calls.append((requested, function_name, k.get("fd_bound")))
                 return probe(requested)
 
             monkeypatch.setattr(runner, "compute_available_workers", _probe)
@@ -398,7 +398,7 @@ class TestConcurrencyPreflight:
         )
         assert width == 2  # the probe's clamp, not _DEFAULT_MAX_WORKERS
         # It asks for the work size and names the dispatch target.
-        assert calls == [(3, "process-shard-test")]
+        assert calls == [(3, "process-shard-test", True)]
 
     def test_probe_denial_falls_back_with_a_warning(self, catalog, monkeypatch):
         # Distinguish the fallback from the shard-count clamp by shrinking the
@@ -416,7 +416,7 @@ class TestConcurrencyPreflight:
         with pytest.warns(RuntimeWarning, match="AccessDenied"):
             width, calls = self._dispatch(catalog, monkeypatch, probe=_denied)
         assert width == 2  # degraded to the default, run still dispatched
-        assert calls == [(3, "process-shard-test")]
+        assert calls == [(3, "process-shard-test", True)]
 
     def test_explicit_max_workers_skips_the_probe(self, catalog, monkeypatch):
         def _must_not_run(requested):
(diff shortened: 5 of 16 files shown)


espg:
🤖 *from Claude*

**Run status — 2026-08-04 small-fix sweep.** Two of the three bundled issues are implemented and green; the third is a genuine block. Final state: `implement` + `waiting`, left as a **draft** (see the deviation note at the bottom).

### Phases landed

| phase | issue | commit | state |
|---|---|---|---|
| 1 | #375 — event-transport pacing window | `56c65c6` | done |
| 2 | #374 — observations-read counter | `c13a59f` | done |
| review fold | #374 — stamp on the no-data shard | `e67edd8` | done |
| — | #372 — conus shardmap count mislabel | *no commit* | **blocked, needs a decision** |

### Why #372 is not in this PR

It cannot be implemented against `main` as written, and this is what puts the PR in `waiting`. The issue names `data/conus/build_conus_shardmap.py:254` and `data/build_aoi_shardmap.py`; at 66cb111 there is no `data/` directory in the repository, and `.gitignore` has no `data/` entry, so those files are genuinely absent rather than untracked. The prerequisite the issue assumes is missing too — `granules_assigned` appears nowhere in the tree, and `ShardMap.build` writes only `total_granules`/`total_shards`/`total_pairs`/`build_wall_s` (`src/zagg/catalog/shardmap.py:534-542`). Three concrete options are laid out in the PR body under "Questions for review" (1); it is referenced with `Refs`, not `Closes`, so merging cannot silently close it.

### Review findings: folded vs standing

The per-phase adversarial self-review could **not** be run as specified. The routine requires two separate subagents on an Opus-class model, but no agent-spawning tool is exposed in this environment (the available toolset has no Agent/Task tool), so the review/fold subagent split was impossible. Rather than skip the step or misrepresent an ordinary self-check as the two-subagent loop, I ran a single-context review of my own diff and am labelling it honestly. **A fresh-context adversarial review of this PR has not happened and is still owed.**

That self-review found and folded one real defect:

- **Folded** (`e67edd8`): `total_obs_read` was stamped only at the normal exit, so the "No data after filtering" early return dropped it — losing the counter on exactly the shards where read-vs-keep matters most (decoded millions, kept none). Moved the stamp to immediately after the read loop; pinned by `test_stamped_on_the_no_data_shard`.

Three judgement calls are **standing** for review rather than folded, since each implies a design decision that is not mine to make — all three are written up in the PR body: the `fd_bound` fallback when the account probe is denied (2), the always-on rather than feature-gated `io_stats` kwarg and the `tests/test_apriori.py::TestWorkerSeam` test it required reshaping (3), and `src/zagg/client.py` sitting over the §4 size limit (4).

### Gate results

Local, on `e67edd8`: `uv run pytest -q` → 3284 passed, 38 skipped, 0 failures. `ruff check` and `ruff format --check` clean except two **pre-existing** failures that also reproduce on `main` at 66cb111 and were deliberately left alone — `N818` on `src/zagg/registry.py:64`, and a format diff on `tests/data/benchmark/README.md`.

CI on `e67edd8`: `ruff` success, `test (3.12)` success, `test (3.13)` success, `build` / `build-arm64` / `build-x86_64` success. No ruff-bot inline comments were posted, so there are none outstanding.

### Deviations and things skipped

- **Left as a draft despite the phases being complete.** CLAUDE.md §2 would have me flip a finished PR to ready-for-review, but this run was explicitly instructed to keep it a draft, and that instruction is recorded here rather than acted against. It is also the right end state independently: `waiting` on the #372 decision, and a fresh-context review still owed.
- **Nothing under `.github/workflows/`, `deployment/aws/`, or any infra file was touched.** #374 needed no handler change — `deployment/aws/lambda_handler.py` already forwards worker metadata verbatim to `build_record`, so the new field reaches the sidecar and run parquet on its own.
- **No dependency added**, no tag created, no deploy or publish script run.
- The two pre-existing lint failures were flagged, not fixed.


---

src/zagg/client.py
```diff
@@ -656,6 +661,7 @@ def _probe_workers(self, session, n: int) -> int:
                 session.client("lambda", region_name=self.region),
                 session.client("cloudwatch", region_name=self.region),
                 self.function_name,
+                fd_bound=fd_bound,
```
espg, line 664: 🤖 *from Claude (review)*

**The one line that delivers issue #375 is not covered by any test — deleting it leaves the whole suite green.**

I mutated the branch by removing exactly this line:

```python
             workers, report = runner.compute_available_workers(
                 n,
                 session.client("lambda", region_name=self.region),
                 session.client("cloudwatch", region_name=self.region),
                 self.function_name,
-                fd_bound=fd_bound,
             )
```

and ran the two suites that own this seam:

```
$ pytest tests/test_client.py tests/test_concurrency.py -q
85 passed, 4 warnings in 3.64s
```

With that line gone, `compute_available_workers` falls back to its `fd_bound=True` default, the event window is FD-clamped again, and issue #375 is fully regressed — with no failing test.

The reason is that the two new tests bracket the seam without crossing it:

- `tests/test_concurrency.py::TestFdBound` calls `compute_available_workers(..., fd_bound=False)` **directly**, so it never goes through `_probe_workers`.
- `tests/test_client.py::_probe_fd_bound` monkeypatches `Run._probe_workers` **wholesale** (`monkeypatch.setattr(Run, "_probe_workers", _probe)`), so the real body — this call — never runs.
- The one test that *does* execute the real `_probe_workers` body is `test_probe_sizes_the_pool` via `TestConcurrencyPreflight._dispatch`, whose stub signature is `def _probe(requested, lambda_client, cloudwatch_client, function_name, **k)` (`tests/test_client.py:358`) — the `**k` silently swallows `fd_bound`.

Concrete fix, ~3 lines in the existing helper: have `_dispatch`'s `_probe` record the kwarg alongside what it already records, and assert on it.

```python
def _probe(requested, lambda_client, cloudwatch_client, function_name, **k):
    calls.append((requested, function_name, k.get("fd_bound")))
    return probe(requested)
```

then `test_event_window_is_not_fd_bound` can assert the *forwarded* value (`calls == [(3, "process-shard-test", False)]`) rather than only the value `dispatch` computed. That closes the gap and makes the mutation above fail.

---

tests/test_client.py
```diff
@@ -439,6 +439,57 @@ def _must_not_run(*a, **k):
         handle.results()
         assert len(handle) == 3
 
+    def test_sync_probes_fd_bound(self, catalog, monkeypatch):
+        # The sync pool holds one socket per in-flight shard: FD-bounded.
+        width, calls = self._dispatch(
+            catalog, monkeypatch, probe=lambda requested: (2, self._report())
+        )
+        assert width == 2
```
espg: 🤖 *from Claude (review)*

**These four lines are a verbatim copy of `test_probe_sizes_the_pool` and assert nothing about `fd_bound`.** Compare with `tests/test_client.py:395-401`:

```python
    def test_probe_sizes_the_pool(self, catalog, monkeypatch):
        width, calls = self._dispatch(
            catalog, monkeypatch, probe=lambda requested: (2, self._report())
        )
        assert width == 2  # the probe's clamp, not _DEFAULT_MAX_WORKERS
```

Identical call, identical assertion. `_dispatch` cannot observe `fd_bound` at all — its stub is `def _probe(requested, lambda_client, cloudwatch_client, function_name, **k)` (line 358), so the kwarg vanishes into `**k`. The whole content of `test_sync_probes_fd_bound` is therefore line 448. `calls` is bound and never read.

Two things worth doing here, and they compose:

1. Drop lines 444-447 — the sibling test already pins pool width, and re-asserting it under a name about FD-bounding just misdirects the reader (§4: terse, reviewable, no dead code).
2. Better: keep the `_dispatch` call but make it *earn its place* by having it observe the forwarded kwarg — see my comment on `src/zagg/client.py:664`, where a mutation showing `fd_bound=fd_bound` can be deleted with the suite still green is the underlying gap. Threading `k.get("fd_bound")` into `calls` fixes both at once and makes these lines the real test rather than a duplicate.

Separately, running `_dispatch` and then `_probe_fd_bound` against the same `monkeypatch` in one test means the second call re-patches `boto3.Session` over the first's patch. It works, but it reads as two unrelated scenarios sharing a name.

---

src/zagg/processing/read.py
```diff
@@ -528,6 +567,13 @@ def _read_fn(path, hyperslice=None):
     )
     lats, lons = coord_arrays[lat_path], coord_arrays[lon_path]
 
+    # Rows DECODED by this group's plan (issue #374): the padded, boundary-
+    # straddling reads counted BEFORE the shard mask and filters below, so the
+    # segment→photon indexed-IO efficiency (issue #43) is derivable from the
+    # run parquet. Recorded ahead of the empty short-circuit so an instrumented
+    # route that decodes nothing still reads as measured-zero, not unmeasured.
+    _record_obs_read(io_stats, len(lats))
```
espg, line 588: 🤖 *from Claude (review)*

**"an instrumented route that decodes nothing still reads as measured-zero, not unmeasured" is false for the planned route — which is the production route for ATL03.** Guarding `_execute_plan_group`'s empty short-circuit is necessary but not sufficient: the planned path has three earlier exits, all of them *before* this line, and none of them touch the sink.

`_planned_read_group`:

```python
    if len(coarse_lats) == 0:
        return None                                       # read.py:432-433

    n_base = int(np.asarray(cnt_arr).sum())
    if n_base <= 0:
        return None                                       # read.py:442-443

    if not plan.parent_runs:
        return None  # empty AOI -- no parent intersects  # read.py:470-471
```

and `_apriori_read_group` has the same shape at `apriori.py:230-231` (`bdf.empty`) and `apriori.py:241-242` (`not plan.parent_runs`).

Empirically, on this branch, using the existing `test_processing.py` fixtures — same granule, same total spatial miss, only the route differs:

```
PLANNED empty-AOI io_stats -> {}
FULL   total-miss  io_stats -> {'obs_read': 12}
```

So `_record_obs_read`'s own docstring contract (`read.py:332-335`, "PRESENCE of the key is the 'measured' signal") and `schema.py`'s `NotRequired` rationale ("absence reads as unmeasured … rather than a genuine zero") hold only on the full-read path. `test_empty_decode_reads_as_measured_zero` pins it there and nowhere else.

Consequence: a shard whose granules were assigned by the shardmap but whose beams do not intersect at the exact-leaf level decodes zero base-rate rows and reports `n_obs_read = None` — indistinguishable from an uninstrumented read seam. That is precisely the discrimination the nullable-column design exists to provide, and the empty-AOI short-circuit is the *common* outcome per beam on the spatial-index route (a granule is assigned for one beam; the other five short-circuit). The numeric rollup is unaffected (`_SUM_OR_NONE` skipping a `None` and summing a `0` agree), so this is a semantics defect, not a wrong number.

Fix is five one-line calls — `_record_obs_read(io_stats, 0)` ahead of each of the five returns above — plus one test in `TestPlannedReadGroup` mirroring `test_empty_decode_reads_as_measured_zero` for the planned route, e.g.:

```python
    def test_planned_empty_aoi_reads_as_measured_zero(self):
        h5 = _planned_read_h5()
        ds = _planned_read_data_source()
        grid = _LatBboxGrid((-0.1, 100000.0, 0.1, 100001.0))  # no segment matches
        io_stats: dict = {}
        assert _read_group(h5, "gt1l", ds, 0, grid, io_stats=io_stats) is None
        assert io_stats["obs_read"] == 0
```

(That test fails on the current branch — it is the probe that produced the output above.)

If instead the intended reading is "unmeasured = no base-rate decode was even attempted", then the comment on these lines and the `_record_obs_read` docstring should say so, and `test_empty_decode_reads_as_measured_zero` should be renamed to make clear it pins the full path only — but the two routes should not silently disagree either way.

---

src/zagg/index/hierarchical.py
```diff
@@ -29,7 +39,11 @@ def read_group(self, h5obj, group, data_source, shard_key, grid, arrow=False, gr
         # ``granule_url`` is forwarded only when set (the a-priori arm), so
         # monkeypatched ``_read_group`` fakes keep their existing signature on
         # every other path -- mirroring the worker's presence-gated kwarg.
+        # ``io_stats`` (issue #374) is gated the same way, for the same reason:
+        # a fake that takes neither keeps working, and simply reports no count.
         kwargs = {"arrow": arrow}
         if granule_url is not None:
             kwargs["granule_url"] = granule_url
+        if io_stats is not None:
+            kwargs["io_stats"] = io_stats
```
espg: 🤖 *from Claude (review)*

**The stated reason for this gate is not true, and the PR body says so itself.** The comment claims:

> ``io_stats`` (issue #374) is gated the same way, for the same reason:
> a fake that takes neither keeps working, and simply reports no count.

But the only production caller always passes a dict, never `None` — `worker.py:422` and `worker.py:436`:

```python
        io_stats: dict = {}
        ...
                    read_kwargs = {"arrow": use_arrow, "io_stats": io_stats}
```

So `if io_stats is not None` is unconditionally true on every real read, a fake that omits `io_stats` does **not** keep working, and the branch never protects anything in production. The PR body's "Questions for review" (3) states the opposite of this comment — "`io_stats` is passed ALWAYS, not feature-gated" — and `tests/test_apriori.py::TestWorkerSeam`'s fakes had to grow the parameter for exactly that reason. The comment as written would lead a future reader to believe old-signature `_read_group` fakes still work.

It is also asymmetric with the sibling backend: `InlineIndex.read_group` forwards `io_stats=io_stats` unconditionally (`inline.py:651`, `inline.py:661`), relying on the `None`-safe `_record_obs_read`. Two backends, two conventions, one of them justified by a false premise.

Pick one and make it consistent — I'd drop the gate, matching inline and the always-on intent:

```python
        kwargs = {"arrow": arrow, "io_stats": io_stats}
        if granule_url is not None:
            kwargs["granule_url"] = granule_url
```

`_record_obs_read` already no-ops on `None`, and the direct-call tests in `tests/test_index.py` (e.g. `HierarchicalIndex().read_group(_open_fixture(), "gt1l", ds, 1, grid)` at line 607) pass `io_stats=None` through happily. If you prefer to keep the gate for those direct callers, then the two comment lines need to be replaced with the actual reason (it is for direct/test callers, not for the worker), not the current claim.

Note the `granule_url` gate on lines 45-46 is genuinely load-bearing and correctly described — I mutation-tested it and `test_granule_url_absent_when_flag_off` does still fail when it is removed. Only the `io_stats` half is the problem.

---

src/zagg/processing/worker.py
```diff
@@ -804,6 +828,8 @@ def _iter_granule_reads():
 
     metadata["cells_with_data"] = cells_with_data
     metadata["total_obs"] = n_obs_total
+    if obs_read_total is not None:  # stamped above, before the no-data return
+        logger.info(f"  Read {obs_read_total:,} observations, kept {n_obs_total:,}")
```

**Every shard now logs "Read N observations" twice, with two different numbers.** The pre-existing lines are `worker.py:673` and `worker.py:676`:

```python
        n_obs_total = buffered.n_obs_total
        logger.info(f"  Read {n_obs_total:,} observations ({buffered.flushes} buffer flushes)")
    else:
        col_arrays, cell_to_slice, n_obs_total = _concat_and_group(all_reads, grid, handoff)
        logger.info(f"  Read {n_obs_total:,} observations")
```

and this one adds:

```python
        logger.info(f"  Read {obs_read_total:,} observations, kept {n_obs_total:,}")
```

On the demo-run numbers from issue #374 that is `Read 31,900,000,000 observations` followed a few lines later by `Read 100,000,000,000 observations, kept 31,900,000,000`. Anyone grepping CloudWatch for `Read .* observations` — which is how a shard's obs tally is read today — now gets two contradictory hits per shard, and the older line is the one that does *not* mean "read". Since this counter exists specifically so "read X, kept Y" can be stated, the log that undermines the claim is worth a word change:

```python
        logger.info(f"  Decoded {obs_read_total:,} observations pre-filter, kept {n_obs_total:,}")
```

(or rewording line 676 to `Kept {n_obs_total:,} observations` — either direction, as long as one verb means one thing).

Minor second point: this is a pure log statement wedged between two `metadata[...]` assignments —

```python
    metadata["cells_with_data"] = cells_with_data
    metadata["total_obs"] = n_obs_total
    if obs_read_total is not None:  # stamped above, before the no-data return
        logger.info(...)
    metadata["duration_s"] = duration
```

— which reads at a glance as if it were stamping something, hence the defensive trailing comment. Moving it below `metadata["duration_s"]` keeps the assignment block contiguous and lets the comment go.

---

src/zagg/client.py
```diff
@@ -698,6 +704,12 @@ def dispatch(
             shards may be dispatched-but-unresolved at once, load-bearing
             because the fleet's 60 s ``MaximumEventAgeInSeconds`` silently
             drops an Event invoke the account has no concurrency to start.
+            The two transports therefore probe differently (issue #375): the
+            sync pool is clamped by the local ``RLIMIT_NOFILE`` ceiling (one
+            held socket per in-flight shard), while the event window — which
+            holds no connections — is sized by account headroom alone, so a
+            low ``ulimit -n`` no longer strands concurrency the account has
+            already granted.
```
espg: 🤖 *from Claude (review)*

**Widening the event window also lengthens the opening ramp, and neither this docstring nor the PR body mentions it.** The event transport does not fire its window in parallel — `client_transport.py:631-636` is explicit:

> `_dispatch_up_to_window` fires its whole batch SERIALLY on this thread, where v1 invoked from `workers` threads — so a `max_in_flight`-wide opening burst is one thread's sequential invoke round-trips (order 10 ms each, i.e. tens of seconds at a 1000-wide window) during which no LIST runs.

`_dispatch_up_to_window` confirms it (`client_transport.py:901-903`): `for entry in batch: self._fire(entry)` on the poller thread. So going from a ~992-wide FD-clamped window to the ~1,898 account headroom in issue #375's scenario roughly doubles that serial ramp, ~10 s → ~19 s, during which the poller runs no LIST. That is very likely the right trade (the issue reports the clamp cost ~3 waves instead of 2 and 5.4 min to first completion), but it is a real second-order effect of this change and the docstring is the place a future reader will look.

Worth one sentence here, e.g. after "already granted": *"The window is fired serially by the poller thread, so a wider window also lengthens the opening dispatch ramp proportionally (`StatusPoller._dispatch_up_to_window`)."*

Two things I checked here and found **sound**, recorded so the next reader does not have to redo them:

- **No FD regression from the wider window.** `max_pool_connections=workers` (line 769) is set on the shared client for both transports, so `fd_bound=False` can now hand botocore a pool cap of ~1,898 against a 1,024 `RLIMIT_NOFILE`. That is safe only because urllib3 sizes pools lazily *and* event dispatch is serial — one connection in flight. It would become a real EMFILE if `_dispatch_up_to_window` ever went concurrent. Not a change request; the coupling is just now load-bearing and undocumented.
- **The sync path is byte-identical.** `min(min(requested, fd), available)` (old) vs `min(min(requested, available), fd)` (new, `fd_bound=True`), and the `available is None` arm is `min(requested, fd)` in both — `min` is associative and commutative, so the two agree on every input.

---

espg (commented):
🤖 *from Claude (review)*

Fresh-context adversarial review of `e67edd8`, the pass the run note (https://github.com/englacial/zagg/pull/378#issuecomment-5178137901) records as still owed. Six inline findings, all diff-scoped and foldable. Nothing here blocks; the two implemented phases are correct in the ways that matter most, and the failure modes I went looking for are not present.

Per the routine's scope, the absence of a #372 implementation is **not** treated as a finding.

## Findings

| # | file:line | what | disposition |
|---|---|---|---|
| (1) | `src/zagg/client.py:664` | `fd_bound=fd_bound` is untested — deleting it leaves the suite green and fully regresses #375 | FOLD |
| (2) | `tests/test_client.py:444-447` | four lines duplicated verbatim from `test_probe_sizes_the_pool`, asserting nothing about `fd_bound` | FOLD |
| (3) | `src/zagg/processing/read.py:570-575` | "measured-zero, not unmeasured" is false on the planned/a-priori routes — five short-circuits leave the sink untouched | FOLD |
| (4) | `src/zagg/index/hierarchical.py:42-48` | the `io_stats` gate's stated reason is false (the worker always passes a dict) and contradicts the PR body | FOLD |
| (5) | `src/zagg/processing/worker.py:831-832` | two "Read N observations" log lines per shard with different numbers | FOLD |
| (6) | `src/zagg/client.py:707-712` | a wider event window also lengthens the serial opening dispatch ramp; undocumented | FOLD (one docstring line) |

(1) and (3) are the ones I would not merge without: (1) means the fix has no regression guard at all, and (3) means the nullable-column semantics the design is sold on hold only on the full-read path.

## Where the implementation is genuinely sound

Stated plainly so the fold does not over-correct:

- **The lock-free per-granule sink holds on every path I traced.** There are no retries in `process_shard`. `_read_granule`'s only uncaught raise sites (`_rewrite_url`, the `H5Coro` open, the `data_source["groups"]` lookup) all precede the group loop, so there is no counted-then-discarded granule; per-group failures are caught inside and keep their counts; `finish_granule` is wrapped and the `finally` swallows. The pooled arm gates its yield on `reads is not None`, and the fold-side `except` at `worker.py:556-564` runs *after* the accumulation at `worker.py:537-538`, so a fold failure keeps the read volume. The serial arm's `yield s3_url, *_read_granule(s3_url)` unpacks the new 3-tuple correctly.
- **Exactly one arm counts per group; no double count and no loss.** Traced all four routes — planned → `_execute_plan_group`; planned selectivity fallback → `_read_group_full` only; flat/full → `_read_group_full`; a-priori → whichever of the two it delegates to (`apriori.py:244` / `apriori.py:249`), never both. `test_full_read_fallback_counts_once_not_twice` pins the one case where re-entry could have double-counted.
- **What is counted really is pre-filter at both call sites.** `read.py:575` and `read.py:916` both record `len(lats)` before the shard mask, the structured/expression filters, and the segment broadcasts. `n_obs_read >= n_obs` therefore holds: kept rows are always a subset of the rows counted, and neither the K>1 chunk split nor the buffered/spill path adds rows.
- **`None` vs `0` plumbs correctly end to end** — `build_record`'s `_opt_int(metadata.get("total_obs_read"))`, `merge`'s `_SUM_OR_NONE_KEYS` (skips `None`, sums a real `0`, all-`None` → `None`), `flatten_record` via `_ROW_SCALARS`. Not bumping `SCHEMA_VERSION` is right, not an oversight: `telemetry.py:33-35` says pre-release key changes rev in place, and an old worker's record simply reads `None` and folds without poisoning — which `test_n_obs_read_back_compat_row_folds` pins.
- **The `TestWorkerSeam` reshape did not weaken the gate it was pinning.** I mutated `hierarchical.py` to pass `granule_url` unconditionally; `test_granule_url_absent_when_flag_off` fails with `assert 'granule_url' not in {'granule_url': None, 'io_stats': {}, 'seen': True}`. Swapping a fixed signature for `**kwargs` + an explicit membership assertion is a legitimate reshape here, since the invariant the old signature encoded ("no extra kwargs at all") is deliberately no longer true.
- **No FD regression on the event path**, despite `max_pool_connections=workers` now potentially being ~1,898 against a 1,024 `RLIMIT_NOFILE` — see (6) for why, and for the coupling that makes it true.
- **The sync path is byte-identical** — `min` is associative and commutative, and both `available` arms agree term-for-term.

## Standing items (for @espg, not for the fold)

- **The denied-probe fallback under `fd_bound=False` — question (2) in the PR body.** The code does what the author says: `concurrency.py:328-334` applies the FD clamp whenever `report.available is None`, independent of `fd_bound`. The reasoning also holds — the literal reading of the issue gives a denied probe a window of `requested`, i.e. `n`, and `_dispatch_up_to_window` would then fire the entire shard set with no pacing, straight into the 60 s `MaximumEventAgeInSeconds` drop (`client_transport.py:68`). I would keep it. Worth noting explicitly that the FD ceiling is then a *proxy* with no relationship to account headroom — an arbitrary but bounded pacing value — which is fine as a conservative default but is not what the number means elsewhere in the module.
- **Always-on vs config-gated `io_stats` — question (3).** Always-on matches the raster sink and is the right call; finding (4) is only about the vestigial gate and its comment, not about revisiting the decision.
- **`src/zagg/client.py` at 1,230 lines — question (4).** Pre-existing (1,211 at `66cb111`); not attributed to this PR.

## Housekeeping checked

- **No `docs/specification.md` / conformance-fixture obligation.** §4's normative-spec rule covers the byte-level store contract; the run-parquet telemetry columns are not documented there (no telemetry column name appears anywhere under `docs/`), so adding `n_obs_read` does not trigger it.
- **No dependency added** — `pyproject.toml` is untouched.
- **Both lint failures reproduce on `main` and are correctly left alone**: `N818` at `src/zagg/registry.py:64`, and the `ruff format` diff on `tests/data/benchmark/README.md`. Confirmed independently on this branch.
- No infra, workflow, or `deployment/aws/` file is modified; the point-path handler forwards worker metadata verbatim (`deployment/aws/lambda_handler.py:1844`), so the PR's claim that no deploy change is needed checks out.

---

src/zagg/client.py
espg: 🤖 *from Claude*

Fixed in `8fc63b0` — and mutation-verified, as asked.

Added `TestConcurrencyPreflight::test_probe_body_forwards_fd_bound`, which drives the **real** `_probe_workers` body (no `monkeypatch.setattr(Run, "_probe_workers", ...)`) against a `compute_available_workers` stub that records `k.get("fd_bound")`:

```python
        monkeypatch.setattr(runner, "compute_available_workers", _probe)
        assert run._probe_workers(MagicMock(), 3, fd_bound=False) == 7
        assert run._probe_workers(MagicMock(), 3, fd_bound=True) == 7
        # Not `[None, None]`: the kwarg reaches the probe, and it is the
        # caller's value, not `compute_available_workers`' own default.
        assert seen == [False, True]
```

Asserting on `[False, True]` rather than only `False` is what closes the hole: `seen == [None, None]` catches the kwarg being dropped entirely, and the `False` element catches it being hardcoded.

**Mutation run.** Deleted `fd_bound=fd_bound` at `client.py:664` locally and re-ran the two suites you named:

```
$ pytest tests/test_client.py tests/test_concurrency.py -q
FAILED tests/test_client.py::TestConcurrencyPreflight::test_probe_body_forwards_fd_bound
1 failed, 85 passed
```

One failure, and it is the new test — same 85 that passed green before now sit alongside a red. Line restored; suite back to 86 passing.

Your `_dispatch`-side suggestion is folded too, separately, in `3473fe3` — see the reply on `tests/test_client.py:447`. That one covers the `fd_bound=True` direction end-to-end through `dispatch`; the event direction can't ride `_dispatch` (every `transport="event"` test in the suite goes through an injected client, which skips the probe by design), hence the direct-call test here.

---

tests/test_client.py
espg: 🤖 *from Claude*

Both parts folded in `3473fe3`, taking your option (2) — the `_dispatch` stub now observes the kwarg, so the lines earn their place instead of being deleted.

`_dispatch`'s `_probe` records it alongside what it already records:

```python
            def _probe(requested, lambda_client, cloudwatch_client, function_name, **k):
                calls.append((requested, function_name, k.get("fd_bound")))
                return probe(requested)
```

`test_probe_sizes_the_pool` and `test_probe_denial_falls_back_with_a_warning` now assert `calls == [(3, "process-shard-test", True)]` — so the default (sync) route's forwarded value is pinned end-to-end through the real `_probe_workers` body, in the tests that were already exercising it.

That made the duplicated block redundant for a second reason, so `test_sync_probes_fd_bound` is down to the one line that was ever its content:

```python
    def test_sync_probes_fd_bound(self, catalog, monkeypatch):
        # The sync pool holds one socket per in-flight shard: FD-bounded.
        assert self._probe_fd_bound(catalog, monkeypatch, transport="sync") is True
```

That also settles your last point — no more `_dispatch` and `_probe_fd_bound` sharing one `monkeypatch` and re-patching `boto3.Session` over each other.

On the underlying gap: `_dispatch` can only reach the `True` direction (it has no `transport="event"` route — the event transport is only driven with an injected client, which skips the probe), so the `False` direction is crossed by a direct `_probe_workers` call in `8fc63b0`. Mutation details on the `client.py:664` thread.

---

src/zagg/processing/read.py
espg: 🤖 *from Claude*

Correct, and fixed in `893f062` — I took the first reading (measured-zero everywhere), not the docstring-rewrite alternative: a shard that looked at the granule and decoded nothing is a *measurement*, and the whole reason the column is nullable is to keep it apart from a seam that was never instrumented.

Five `_record_obs_read(io_stats, 0)` calls, exactly where you listed them — `read.py:433`, `:443`, `:471` and `apriori.py:231`, `:242` (the a-priori arm needed `_record_obs_read` added to its local import at `apriori.py:214`). The planned route's three carry one shared comment rather than three:

```python
    # The three short-circuits below return before any base-rate decode, so
    # each stamps a measured zero (issue #374): without it the planned route --
    # the production route for ATL03, where a granule assigned for one beam
    # short-circuits on the other five -- reports ``n_obs_read = None``,
    # indistinguishable from an uninstrumented read seam. That discrimination
    # is the whole point of the nullable column (review finding).
```

I also corrected the two comments that asserted the false property: the `_execute_plan_group` comment you flagged now says the pre-decode short-circuits "never reach this line" and stamp their own zero, and `_planned_read_group`'s `io_stats` docstring paragraph gained the same clause — so neither still reads as if guarding this one line were sufficient.

**Tests.** Yours plus the two array-level exits, since all three are pre-decode and none was covered: `TestPlannedReadGroup::test_planned_short_circuits_read_as_measured_zero` runs `empty_aoi` (your grid, verbatim), `empty_coarse` (zero segments at the spatial-index level) and `no_base_rows` (segments present, link counts summing to zero), asserting `io_stats["obs_read"] == 0` on each. `TestAprioriReadGroup::test_short_circuits_read_as_measured_zero` covers the a-priori pair — no chunk touching the shard, and the beam absent from the boundary parquet.

Both fail on the parent commit and pass on `893f062`:

```
$ git stash push src/zagg/processing/read.py src/zagg/processing/apriori.py
$ pytest tests/test_processing.py tests/test_apriori.py -k "measured_zero or short_circuits" -q
FAILED tests/test_processing.py::TestPlannedReadGroup::test_planned_short_circuits_read_as_measured_zero
FAILED tests/test_apriori.py::TestAprioriReadGroup::test_short_circuits_read_as_measured_zero
2 failed, 1 passed
```

(the one that passes is the pre-existing `test_empty_decode_reads_as_measured_zero`, which pinned the full path only — it stays as your comment describes, now with siblings on the other two routes.)

Agreed on the rollup: `_SUM_OR_NONE` skipping a `None` and summing a `0` agree, so no numeric column moves.

---

src/zagg/index/hierarchical.py
espg: 🤖 *from Claude*

Right on both counts — the stated reason was false and it contradicted the PR body. Dropped the gate in `a977723`, matching `InlineIndex` and the always-on intent, exactly as you wrote it:

```python
        kwargs = {"arrow": arrow, "io_stats": io_stats}
        if granule_url is not None:
            kwargs["granule_url"] = granule_url
```

The comment now states the real situation instead of the fiction — the worker always passes a dict (`worker.py:422`, `:436`), `_record_obs_read` no-ops on `None` for the direct callers, so the gate protected nothing while diverging from the sibling backend.

**Test.** `TestHierarchicalDelegation::test_io_stats_forwarded_unconditionally` pins the contract at the seam, including the case the old gate actually changed — a caller that omits the sink now forwards `None` rather than dropping the kwarg:

```python
        HierarchicalIndex().read_group(object(), "gt1l", {}, 0, None, io_stats=sink)
        HierarchicalIndex().read_group(object(), "gt1l", {}, 0, None)
        assert [k["io_stats"] for k in seen] == [sink, None]
```

Fails with the gate restored (`KeyError: 'io_stats'` on the second call), passes on `a977723`.

I left `granule_url`'s gate alone per your note that it is load-bearing and correctly described, and the direct-call tests in `tests/test_index.py` still pass unchanged (108 passed across `test_index.py` + `test_apriori.py`).

PR-body question (3) is updated in the same round: it no longer asks whether to config-gate `io_stats` — the two backends now agree, and the question is retired rather than left standing as a fork you'd have to answer.

---

src/zagg/processing/worker.py

Both folded in `4177cff` — took your first option (change the new line's verb), since `Read N observations` is the string already in people's CloudWatch filters and the older lines are the ones that shouldn't move.

```python
    metadata["cells_with_data"] = cells_with_data
    metadata["total_obs"] = n_obs_total
    metadata["duration_s"] = duration
    # "Read" already means "kept" in the two lines above (worker.py:673/:676),
    # which is what CloudWatch greps for today, so this one says "Decoded" —
    # one verb, one meaning (review finding, issue #374).
    if obs_read_total is not None:
        logger.info(f"  Decoded {obs_read_total:,} observations pre-filter, kept {n_obs_total:,}")
```

On your demo-run numbers that now reads `Read 31,900,000,000 observations` then `Decoded 100,000,000,000 observations pre-filter, kept 31,900,000,000` — one hit per shard for `Read .* observations`, and the pre-filter count is unambiguous about which number it is.

Second point taken too: the log moved below `metadata["duration_s"]`, so the three `metadata[...]` assignments are contiguous again and the defensive `# stamped above, before the no-data return` comment is gone (the stamp site is pinned by `test_stamped_on_the_no_data_shard` anyway, which is where that claim belongs).

**Test.** `TestObsReadMetadata::test_one_read_log_line_per_shard` asserts the grep property directly rather than the string:

```python
        assert sum("Read " in ln and "observations" in ln for ln in lines) == 1
        assert "  Decoded 10 observations pre-filter, kept 3" in lines
```

Fails on the parent commit with two matches (`Read 3 observations` + `Read 10 observations, kept 3`), passes on `4177cff`.

---

src/zagg/client.py

Documented in `f72e706` — your sentence, plus the trade so a future reader doesn't have to re-derive whether the ramp was noticed or overlooked:

```
            already granted. The window is fired serially by the poller thread
            (:meth:`~zagg.client_transport.StatusPoller._dispatch_up_to_window`),
            so a wider window also lengthens the opening dispatch ramp
            proportionally — roughly 10 ms of invoke round-trip per shard, and
            no LIST runs during it. Deliberate: the ramp grows by seconds while
            the clamp cost whole 900 s waves (issue #375).
```

That is the docstring on `dispatch`'s `max_workers`, where the window semantics already live, rather than on `_probe_workers` — it is a property of the window, not of the probe.

Your two "sound" checks are recorded and not re-litigated. The FD one is worth keeping visible, so I'd rather not bury it in a docstring: **`max_pool_connections=workers` against `RLIMIT_NOFILE` is safe only because `_dispatch_up_to_window` is serial.** If that loop ever goes concurrent it becomes a real EMFILE, and this PR is what makes the coupling load-bearing. Leaving it as a standing note for you rather than acting on it — a guard or a pool-cap change is a design call and beyond `small-fix` scope.

---

espg:
🤖 *from Claude*

Fold round complete for the adversarial self-review of `e67edd8`. All six inline findings folded, one commit each, staging only that finding's files. Every thread has a reply; none resolved.

## Findings → fold commits

| # | finding | fold | files |
|---|---|---|---|
| 1 | `client.py:664` — `fd_bound=fd_bound` had no test; deleting it left the suite green | [`8fc63b0`](https://github.com/englacial/zagg/commit/8fc63b0) | `tests/test_client.py` |
| 2 | `tests/test_client.py:444-447` — four lines copied verbatim, asserting nothing about `fd_bound` | [`3473fe3`](https://github.com/englacial/zagg/commit/3473fe3) | `tests/test_client.py` |
| 3 | "measured-zero, not unmeasured" false on the planned / a-priori routes — five pre-decode exits never touched the sink | [`893f062`](https://github.com/englacial/zagg/commit/893f062) | `read.py`, `apriori.py`, `test_processing.py`, `test_apriori.py` |
| 4 | `hierarchical.py:42-48` — the `io_stats` gate's stated reason was false and contradicted the PR body | [`a977723`](https://github.com/englacial/zagg/commit/a977723) | `hierarchical.py`, `test_index.py` |
| 5 | two `Read N observations` lines per shard with different numbers; log wedged between `metadata[...]` assignments | [`4177cff`](https://github.com/englacial/zagg/commit/4177cff) | `worker.py`, `test_processing.py` |
| 6 | wider event window proportionally lengthens the SERIAL opening dispatch ramp — undocumented | [`f72e706`](https://github.com/englacial/zagg/commit/f72e706) | `client.py` |

Each behavioral fold carries a test in the same commit, and each was **mutation-verified** — fix reverted locally, new test observed to fail, fix restored:

- (1) `fd_bound=fd_bound` deleted → `1 failed, 85 passed` across `test_client.py` + `test_concurrency.py`, the failure being the new `test_probe_body_forwards_fd_bound`. That is the exact suite that was `85 passed` clean in the review.
- (3) `read.py` + `apriori.py` stashed → both new short-circuit tests fail (`KeyError: 'obs_read'`).
- (4) gate restored → `test_io_stats_forwarded_unconditionally` fails on the `io_stats=None` call.
- (5) old log line restored → `test_one_read_log_line_per_shard` fails with two `Read … observations` hits.

(6) is docstring-only, so no test.

Findings (1) and (2) compose, as the review noted: (2) makes `_dispatch`'s probe stub record `k.get("fd_bound")`, which covers the `fd_bound=True` direction end-to-end through `dispatch` in the tests already exercising it. The `False` direction can't ride `_dispatch` — every `transport="event"` test in the suite goes through an injected client, which skips the probe by design — so (1) crosses the seam with a direct `_probe_workers` call instead.

PR body updated in the same round: fold row added to the phases checklist, question (3) rewritten (it no longer asks whether to config-gate `io_stats` — the two backends now agree, so the fork is retired rather than left for you to answer), question (2) carries the reviewer's caveat, question (4) re-counted, and question (5) added for the standing FD note below.

## Left standing (replied on the thread, not acted on)

- **The denied-probe FD fallback stays.** Review assessed the code does what the description says and the reasoning holds. Added the reviewer's caveat to question (2): under `fd_bound=False` the FD ceiling bounds nothing physical, so it is an arbitrary-but-bounded pacing proxy — right shape, meaningless number. A named pacing constant instead is a design call.
- **Always-on `io_stats` — the right call**, and now consistent across both index backends after fold (4).
- **`client.py` at 1,235 lines.** Pre-existing: 1,211 on `main` at 66cb111, i.e. already past the §4 limit before this PR, which adds 24 lines of almost-entirely docstring. §4 says raise, not split — flagged in question (4), not attributable here, and a split is its own PR.
- **`max_pool_connections=workers` vs `RLIMIT_NOFILE`** — safe today only because `_dispatch_up_to_window` is serial, and this PR is what makes that coupling load-bearing. Raised as new question (5) rather than fixed: a guard or a per-transport pool cap is beyond `small-fix`.
- **#372 stays blocked and stays `Refs`.** It needs a human decision between the three options in question (1); no commit in this PR touches it.

The review's verified-sound list was not re-litigated: the lock-free per-granule sink on every path, exactly-one-arm-counts across all four routes, `n_obs_read >= n_obs`, the `TestWorkerSeam` reshape not weakening its gate, no `docs/specification.md` obligation, `SCHEMA_VERSION` correctly not bumped, sync path byte-identical.

## Gate (on `f72e706`)

- `uv run pytest -q` → **3289 passed, 38 skipped**, 0 failures (was 3284 before the fold; +5 new tests).
- `uv run ruff check src tests` → one error, **pre-existing**: `N818` on `src/zagg/registry.py:64`. Reproduces on `main` at 66cb111; not fixed, per the "don't fix unrelated CI failures" rule.
- `uv run ruff format --check src tests` → one file would be reformatted, **pre-existing**: `tests/data/benchmark/README.md`. Same disposition.

Every file this fold touched is clean under both.

## State

Labels unchanged: `implement` + `waiting`. Still a **draft** — the ball is in your court on question (1) (#372) before this is finishable, so it is deliberately not flipped to ready-for-review.

---

espg:
🤖 *from Claude*

**All five questions RULED** (espg, in-session, 2026-08-05):

| question | ruling |
| --- | --- |
| (1) issue #372 unimplementable | **Premise correction + option (2).** The prerequisite HAS landed since this PR's search: `granules_assigned` was introduced by `bacca5c` (now on `origin/main` — `src/zagg/catalog/shardmap.py:543`, `src/zagg/sweep.py:330`); the search here ran against `66cb111`, one commit earlier. What remains missing is the target scripts themselves — `data/build_aoi_shardmap.py` and `data/conus/build_conus_shardmap.py` are untracked working-tree files. espg will commit the `data/` builders; #372 then becomes the one-liner as written. `Refs` stays (nothing closes with this merge). |
| (2) unreadable-account fallback | **As shipped** — keep the fd clamp when the probe is denied. The literal reading fires the whole shard set exactly when headroom is unknown (the 60 s `MaximumEventAgeInSeconds` drop scenario). The arbitrary-but-bounded caveat is recorded; a named pacing constant only if a denied-probe run ever shows real pacing pain. |
| (3) `io_stats` always-on | **Endorsed as shipped** — the gate protected nothing, always-on matches `InlineIndex` and the raster sink; `granule_url`'s load-bearing gate stays. |
| (4) `client.py` at 1,235 | **Overage accepted** (standing espg ruling: line overage is not a present concern). Note for later: `client.py` is not in PR #349's split list, so if it keeps growing it joins the #330 line as its own entry. |
| (5) pool-cap coupling | **Fix it here, now** — one commit incoming on this branch: cap the shared client's `max_pool_connections` at the fd-safe bound regardless of transport. Zero behavioral effect today (sync is already fd-clamped below it; the event dispatch loop is serial with ≤1 connection in flight), and it permanently defuses the latent EMFILE if the dispatch loop ever goes concurrent — the declared pool then matches the actual connection model. |

After the (5) commit lands green, this PR flips ready for review — every question is resolved and it is purely awaiting merge.

espg (commented):
🤖 *from Claude (review)*

The clamp itself is right and it is mutation-sensitive: dropping the `min(...)` back to `max_pool_connections=workers` fails `test_pool_cap_clamped_when_window_exceeds_fd_bound` (`assert 3 == 2`), `fd_safe_max_workers()` is floored at 1 so the cap can never declare a 0-connection pool, the injected-client seam builds no `Config` and is correctly untouched, `_pool_cap` faithfully mirrors `_probe_fd_bound`'s function-local `boto3`/`MagicMock` import style and leaks nothing (all patching via `monkeypatch`), and `ruff check` / `ruff format --check` are clean with `tests/test_client.py` at 59 passed. One finding is worth acting on before this lands: the new tests pin the *declared pool* but nothing in the suite pins that the **fan-out width itself** stays unclamped — I mutated `workers = max(1, min(requested, n))` to `min(requested, n, fd_safe_max_workers())` (which would silently undo this PR's whole `fd_bound=False` win by re-clamping the event pacing window at `RLIMIT_NOFILE`) and `tests/test_client.py` + `tests/test_concurrency.py` still ran **88 passed**. The rest are wording/coverage notes, non-blocking.

tests/test_client.py
```diff
@@ -439,6 +439,126 @@ def _must_not_run(*a, **k):
...
+        assert seen == [False, True]
+
+    def test_pool_cap_clamped_when_window_exceeds_fd_bound(self, catalog, monkeypatch):
+        # The pacing window may clear RLIMIT_NOFILE (fd_bound=False, issue
+        # #375), but the shared client's DECLARED pool never may: today the
+        # serial dispatch loop holds <=1 connection and urllib3 sizes pools
+        # lazily, but the declared cap is what turns into EMFILE if the loop
+        # ever goes concurrent (PR #378, question (5) ruling).
+        monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 2)
+        assert self._pool_cap(catalog, monkeypatch, max_workers=5) == 2
```
espg: 🤖 *from Claude (review)*

**Coverage gap (the one I'd act on): nothing pins that the clamp stops at the *declared pool* and does not reach the fan-out width.** The ruling's "zero behavioral change today" rests entirely on `workers` staying `max(1, min(requested, n))`; if a later edit moved the fd term up one line into `workers`, the event pacing window would go back to being `RLIMIT_NOFILE`-clamped — the exact regression issue #375 fixed — and the suite would not notice. Verified by mutation on a scratch copy of `src/`:

```python
# src/zagg/client.py:764, mutated
workers = max(1, min(requested, n, fd_safe_max_workers()))
# ... with max_pool_connections=workers
```

`pytest tests/test_client.py tests/test_concurrency.py` → **88 passed**. `test_probe_sizes_the_pool` survives because its probe returns 2 (already below the real fd bound), and `test_event_window_is_not_fd_bound` only pins the flag handed to the probe, not the width that comes back out of `dispatch`.

Cheapest fix: assert both numbers at once here — declared pool `== 2` **and** dispatch width still `== 3` — since `_dispatch`'s `_TrackedPool` already captures the executor width for free:

```python
monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 2)
pool, width = self._pool_cap(catalog, monkeypatch, max_workers=5)
assert (pool, width) == (2, 3)  # pool capped; fan-out is NOT
```

tests/test_client.py
```diff
@@ -439,6 +439,126 @@ def _must_not_run(*a, **k):
...
+        # Not `[None, None]`: the kwarg reaches the probe, and it is the
+        # caller's value, not `compute_available_workers`' own default.
+        assert seen == [False, True]
+
+    def test_pool_cap_clamped_when_window_exceeds_fd_bound(self, catalog, monkeypatch):
+        # The pacing window may clear RLIMIT_NOFILE (fd_bound=False, issue
+        # #375), but the shared client's DECLARED pool never may: today the
+        # serial dispatch loop holds <=1 connection and urllib3 sizes pools
+        # lazily, but the declared cap is what turns into EMFILE if the loop
+        # ever goes concurrent (PR #378, question (5) ruling).
```
espg: 🤖 *from Claude (review)*

**The comment describes the event path; the test exercises the sync one.** `_pool_cap(..., max_workers=5)` runs `transport="sync"` (the default) *and* passes an explicit `max_workers`, so `_probe_workers` never runs and `fd_bound=False` is never involved — what actually makes `workers` (3) exceed the patched fd bound (2) here is the probe being skipped, not the pacing window. That is a legitimate way to force the condition, but it leaves the motivating path (`transport="event"` with a probe returning more than `fd_safe_max_workers()`) unexercised at this seam.

Either reword to say what is actually being forced, or parametrize over transport and abort the event run right after the client is built with the same sentinel trick `_probe_fd_bound` uses — the `Config` is recorded at `client.py:770`, before the setup handshake, so raising from a monkeypatched `runner._invoke_lambda_ping` / `runner._invoke_lambda_setup` is enough (no poller or status store needed):

```python
@pytest.mark.parametrize("transport", ["sync", "event"])
```

src/zagg/client.py
```diff
@@ -735,19 +753,28 @@ def dispatch(
...
+                if max_workers is not None
+                else self._probe_workers(session, n, fd_bound=transport == "sync")
+            )
             workers = max(1, min(requested, n))
             # read_timeout must exceed the 900 s function ceiling (same
             # rationale as agg's preflight-built client, issue #148); the
-            # connection pool tracks the fan-out width.
+            # connection pool tracks the fan-out width, but must never exceed
+            # the fd ceiling regardless of pacing window — the event window
+            # may clear RLIMIT_NOFILE (issue #375), a pool of sockets may not.
```
espg, line 768: 🤖 *from Claude (review)*

**Comment accuracy, minor.** botocore never passes `block=True` to urllib3 — `URLLib3Session._get_pool_manager_kwargs` sets only `'maxsize': self._max_pool_connections`, and `urllib3.HTTPConnectionPool.__init__` defaults `block=False` (checked against the installed botocore and urllib3 2.7.0). So `max_pool_connections` bounds the connections the pool **retains**, not how many can be open at once: over-maxsize connections are still created, then discarded on release. This line ("a pool of sockets may not [clear RLIMIT_NOFILE]") reads fine under that meaning, but the test's stronger claim at `tests/test_client.py:486` — "the declared cap is what turns into EMFILE if the loop ever goes concurrent" — overstates it in both directions: an oversized *declared* cap allocates nothing on its own (lazy sizing, as the comment already notes), and a small one would not prevent EMFILE under a concurrent loop. The honest framing is that the pool must not retain more sockets than the fd budget — a retention bound, not an EMFILE guard. Wording only; the cap is still the right thing to declare.

src/zagg/client.py
```diff
@@ -735,19 +753,28 @@ def dispatch(
...
             client = session.client(
                 "lambda",
                 region_name=self.region,
                 config=Config(
                     read_timeout=960,
                     connect_timeout=10,
                     retries={"max_attempts": 0},
-                    max_pool_connections=workers,
+                    max_pool_connections=min(workers, fd_safe_max_workers()),
```
espg: 🤖 *from Claude (review)*

**"Zero behavioral change today" holds for the probe-driven paths, not quite for all of them** — worth scoping the claim in the PR body rather than changing the code. On `transport="sync"` the probe already applies the fd term inside `compute_available_workers`, so `min()` is a no-op; on `transport="event"` today's serial dispatch loop holds ≤1 connection, so it is inert. But the **explicit `max_workers`** path (either transport) skips the probe entirely, so `workers` carries no fd term: an operator running `dispatch(max_workers=300)` under `ulimit -n 256` now gets a declared pool of 224 against a 300-wide thread pool, i.e. urllib3 "Connection pool is full, discarding connection" churn where it previously retained one connection per thread. That is consistent with the ruling ("regardless of transport") and arguably the better behavior — it just is not *no* change.

src/zagg/client.py
```diff
@@ -61,6 +61,7 @@
 from dataclasses import asdict
 from typing import Any, Iterator
 
+from zagg.concurrency import fd_safe_max_workers
```
espg: 🤖 *from Claude (review)*

**Nit, seam consistency.** Everywhere else `client.py` reaches `zagg.concurrency` indirectly through the lazily-imported `runner` module — `runner.compute_available_workers` and `runner._log_concurrency_report` at `client.py:660-667` — and `runner.py:1791` states that convention outright ("referencing the same module seams … so tests patch one set of objects"). This direct top-level import adds a second patch point: `tests/test_concurrency.py` patches `concurrency.fd_safe_max_workers` and `RLIMIT_NOFILE`, and neither reaches this clamp — only `zagg.client.fd_safe_max_workers` does. The import is cheap and cycle-free (`zagg.concurrency` pulls stdlib only), so this is fine as a deliberate choice; flagging it so that it is one.

tests/test_client.py
```diff
@@ -439,6 +439,126 @@ def _must_not_run(*a, **k):
...
+        monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 2)
+        assert self._pool_cap(catalog, monkeypatch, max_workers=5) == 2
+
+    def test_pool_cap_unchanged_below_fd_bound(self, catalog, monkeypatch):
+        monkeypatch.setattr(zagg.client, "fd_safe_max_workers", lambda: 100)
+        assert self._pool_cap(catalog, monkeypatch, max_workers=2) == 2
+
+    @staticmethod
+    def _pool_cap(catalog, monkeypatch, **dispatch_kwargs):
```
espg: 🤖 *from Claude (review)*

**Nit, duplication.** `_pool_cap` is `_dispatch` (line 340) with one line changed: same `MagicMock` session, same `StubLambdaClient`, same `Run.from_config(...)`, same `dispatch(**kwargs).results()` — only the recorded value differs (`config.max_pool_connections` instead of the `ThreadPoolExecutor` width). Folding the config capture into `_dispatch` and returning both would drop ~20 lines and, per the finding on line 489, hand you the fan-out-width assertion the new tests currently lack.

tests/test_client.py
espg: 🤖 *from Claude*

Folded — `89c8009` (on top of the helper dedupe, `8d47425`). Both pool tests now assert the declared pool **and** the fan-out width in one tuple, exactly as suggested:

```python
monkeypatch.setattr(zagg.concurrency, "fd_safe_max_workers", lambda: 2)
width, _, pool = self._dispatch(catalog, monkeypatch, max_workers=5)
assert (pool, width) == (2, 3)
```

`_dispatch` now returns `(pool_width, probe_calls, declared_pool)`, so the `_TrackedPool` width you noted was already free is the width being asserted — no new machinery.

Mutation-verified against your exact mutation: with `workers = max(1, min(requested, n, fd_safe_max_workers()))` at `client.py:764`, `test_pool_cap_clamped_when_window_exceeds_fd_bound` fails (`(2, 2) != (2, 3)`); restored, green. The below-bound test pins `(2, 2)`.

tests/test_client.py
espg: 🤖 *from Claude*

Folded — `8d47425`. `_pool_cap` is gone: the config capture moved into `_dispatch`, whose lambda `side_effect` became a small `_client` function that records `max_pool_connections` when the kwarg carries a `config` (the fan-out client; the probe's `session.client("lambda", …)` carries none, so it is skipped). `_dispatch` returns `(pool_width, probe_calls, declared_pool)` and the three existing call sites unpack a third slot. Net −22 lines, and it hands the width assertion the finding on line 489 asked for (`89c8009`).

tests/test_client.py
espg: 🤖 *from Claude*

Folded as the reword (your first option) — `4b609d2`. The comment now names the mechanism that actually forces the condition, and points at the test that owns the event wiring instead of implying this one covers it:

> Forced here by an explicit `max_workers`, which skips the probe so `workers` carries no fd term at all — the same shape the event transport reaches through the probe when it passes `fd_bound=False` (issue #375; that wiring is pinned by `test_event_window_is_not_fd_bound`, which this test deliberately does not re-derive).

Not parametrized, deliberately: the event variant needs the sentinel abort right after client construction, and an aborted dispatch never reaches the `ThreadPoolExecutor` — so it cannot also carry the fan-out-width assertion from the finding on line 489, which is the coverage that was actually missing. The clamp itself sits in the shared client construction *before* any transport branch (`client.py:774-782`), so a transport-parametrized run would re-execute the same line with the same inputs; what differs by transport is which value `workers` arrives with, and that is already bracketed by `_probe_fd_bound` + `TestFdBound`.

src/zagg/client.py
espg: 🤖 *from Claude*

Folded — `28f9b31`, on the overstating line you named (`tests/test_client.py:486`), which now says retention:

> The cap is a RETENTION bound, not an EMFILE guard: botocore passes only `maxsize` to urllib3 (`block=False`), so an oversized declaration allocates nothing on its own — it is the number of sockets the pool would KEEP once a concurrent dispatch loop opened them.

The source comment here is left as written, since you read it as fine under that meaning and the retention framing is what "a pool of sockets may not [clear `RLIMIT_NOFILE`]" already says. The same correction went into the PR body's question (5) so the claim does not survive there either.

src/zagg/client.py

Folded in the PR body, no code change — question (5) now scopes the claim instead of asserting it flat. It splits the two cases your comment does: inert on both probe-driven paths (sync applies the fd term inside `compute_available_workers`; the serial event dispatch loop holds ≤1 connection), **not** inert on the explicit-`max_workers` path, with your example carried over verbatim — `dispatch(max_workers=300)` under `ulimit -n 256` declaring a 224-connection pool against a 300-wide thread pool, urllib3 discard churn rather than an error. Recorded as consistent with the ruling ("regardless of transport") and arguably better, but a change.

No code touched because the divergence is in the claim, not the behavior: the ruled expression is what shipped.

src/zagg/client.py

Folded — `1f297b5`. The top-level `from zagg.concurrency import fd_safe_max_workers` is gone; `dispatch` now does `from zagg import concurrency, runner` (the same lazy module-seam import it already uses for `runner`) and calls `concurrency.fd_safe_max_workers()`. The second patch point is gone with it: `zagg.concurrency.fd_safe_max_workers` is now the one object that reaches both this clamp and `compute_available_workers`, which is what `tests/test_concurrency.py` was already patching. The two pool tests moved to that target.

Not routed through `runner`: `runner` does not import `fd_safe_max_workers` today (only `compute_available_workers` / `_log_concurrency_report`), so the `runner.` seam would mean adding a re-export that exists solely to be patched. `concurrency.` is the module that owns the ceiling.

Mutation-verified that the new patch point actually reaches the clamp: with the cap reverted to `max_pool_connections=workers`, `test_pool_cap_clamped_when_window_exceeds_fd_bound` fails; restored, green.

PR #379, opened 2026-08-04
overview sweep: cascade coarse levels from the previous overview (issue #376)

### LINKED ISSUE #376: overview sweep: cascade coarse levels from the previous overview; exact-from-leaves becomes a deprecated option
🤖 *from Claude*

## Structural problem (read from `sweep_overview._fold_node`, confirmed against the California o9 store)

The overview fold has block structure — level → node → leaf, fixed `4^10`-cell output slab per node — but two properties break at state scale:

1. **No cascade**: every declared level re-folds from the raw o19 leaves (single-quantization by design), so a 4-level pyramid costs ~4× the full-store digest decode.
2. **Per-node accumulation**: digest fields collect per-output-cell centroid lists across the node's ENTIRE subtree before one k-way fold. California (2,721 leaves, ~462M populated digests across the strata fields) sits **entirely within base cell 3** (verified via `geo2mort` at six corners spanning San Diego to the Oregon line) — so the o1 node's subtree is the whole state, and the coarse-level folds want hundreds of GB of resident centroids. No worker tier fits that; splitting across base cells wouldn't help a single-base AOI at all.

## Ratified direction (espg, 2026-08-04)

Overviews are display artifacts: **no precision guarantee past the first (possibly second) skip level.** Fold level *k* from level *k+2*'s already-materialized overview digests (fold-of-folds):

- input per node becomes its 16 child overview slabs, each cell ≤ δ centroids → **constant, bounded memory per node**, independent of subtree size;
- leaves are read **once** (the finest level only) instead of once per level;
- accuracy loss lands exactly where nothing was promised.

Keep the current exact-from-leaves behavior as a **deprecated opt-in** — e.g. `output.pyramid.overview.fold_source: "cascade" (default) | "leaves"` — and record each materialized level's fold provenance in the pyramid block / spec §4, since fold-of-folds inherits the documented merge order-dependence and a reader deserves to know which regime a level is in. Open sub-decision: whether the second level (o5 on the o9 geometry) also folds from leaves ("first, or possibly second" — espg).

## Sequencing / refs

The demo notebook (`05_california_read.ipynb`) is about to produce the empirical baseline: per-family sweep timings (#354) plus per-level `failed` counts on the current implementation. Related: #344 (overview family), #348 phase E (fold sizing — node-level parallelism remains the orthogonal wall-clock lever; the per-(node, window) fold is already an independent unit), #352/#369 (per-field schedules — orthogonal cost lever), #370 (k-way fold precedent). The sweep's idempotent generation stamps already make repeated invokes a ratchet; cascade makes each click of the ratchet bounded.

espg:
Closes #376

## What this does

Makes the overview sweep **cascade**: a declared level folds from the next *finer* declared level's already-materialized overviews (fold-of-folds) instead of re-folding the whole subtree from the raw leaves. Exact-from-leaves survives as a deprecated opt-in, and every materialized level records which regime produced it.

The structural problem, read off `_fold_node` on `main`: the per-node accumulator is

```python
for name, decoded in cell_digests.items():
    for j, digest in decoded:
        digests[name][j].append(digest)     # the node's ENTIRE subtree, resident
```

before one k-way fold per output cell — so a single-base-cell AOI's coarse nodes want the whole region's centroids resident — and the level loop re-reads every leaf per declared order.

Under cascade, a level-`k` node reads its `4^gap` child overviews at the level below (`gap` = the step to the next finer declared order) **one at a time**, folds each child's slab straight into the parent's *disjoint* output span, and merges at most `4^gap` digests per output cell:

```python
start = _rel_rank(child, node) * span
for name, partial in partials.items():
    slabs[name][start : start + span] = partial   # disjoint spans: assignment, not accumulation
```

Two consequences: per-node resident input is the output slab plus **one** child slab (constant in the subtree size), and the leaves are read **once for the whole pyramid** rather than once per level — pinned by a test that spies on `_fold_node` and asserts only the finest declared order ever reaches it.

The geometry that makes this work: every overview slab in the tree holds the same `4^(cell_order - shard_order)` cells (constant tree depth, spec §4.4), so a child's slab folds `4^gap`-to-one into the `4^(cell_order - shard_order) / 4^gap` cells that child owns in the parent. `fold_dense` handles the exact class unchanged (it is already a `reshape(-1, factor)` reduce); the approximate class re-uses `fold_digests` — the same order-independent k-way merge (#279/#370 precedent), just applied to `4^gap` inputs instead of the subtree.

### Config

```yaml
output:
  pyramid:
    orders: [7, 5, 3, 1]
    exact_levels: 1        # cascade boundary; 1 is the default
    # fold_source: leaves  # DEPRECATED opt-in: exact at every level, unbounded per-node input
```

`fold_source` defaults to `cascade`. `exact_levels` is the **named, testable boundary** the issue's open sub-decision turns on (see question (a)) — how many of the *finest* declared levels keep the exact-from-leaves fold. `fold_source: leaves` warns at config validation and at template time, and refuses to be combined with `exact_levels` (under `leaves` every level is exact, so a boundary would name a distinction that does not exist). An `exact_levels` that covers **every** declared level is that same deprecated regime by another route, so it warns too — at declaration time and at sweep time, the two places that know the derived order schedule.

### Provenance (spec §4.3 / §4.5)

- each overview's `zagg_overview` attrs gain `fold_source` (`"leaves"` / `"cascade"`) and, when cascaded, `fold_from_order`;
- a cascaded overview also records `source_children` — `{"folded": 15, "missing": 1, "unreadable": 0}` — because a cascade folds what is **on disk**: a nonzero `missing`/`unreadable` means the level under-covers its subtree, and a fill cell there is not evidence of emptiness. Without it that distinction lived only in a log line, in a sweep process that has already exited;
- the manifest's sweep-written actuals gain `materialized.fold_sources` — `{"3": "leaves", "1": "cascade"}`, accumulated across sweeps, so a level swept under an earlier declaration keeps the regime it was actually made with. It records only levels that actually **wrote** in a pass; a level that was current (or folded empty) keeps the regime its own earlier sweep recorded, so the map stays an actual rather than a restatement of the plan;
- the declaration itself records `fold_source` / `exact_levels` — what the **next** sweep will do, never a claim about what is on disk;
- `declare_pyramid`'s summary (printed by `python -m zagg.sweep --declare-pyramid`) names the declared regime.

### Two behaviors worth naming explicitly

1. **Skip-if-current keys on the regime.** The exact merge laws are associative, so `count`/`h_min` are *byte-equal* under either source (there is a test asserting exactly that) — the generation stamp and content hash therefore cannot tell a stored overview apart from the other fold. The envelope entry carries `fold_source` **and** `fold_from_order`, and the current-check compares both, so neither flipping the knob nor dropping an intermediate declared order can leave a level reading as `current` while its recorded regime names something that is no longer true. Entries written before this PR carry no key and are read as `leaves`, so **no existing store regenerates on upgrade** (tested). `_content_hash` is deliberately untouched: it is the sweep-internal same-second backstop, and folding the regime into it would have churned every existing overview's digest for nothing.
2. **A cascade folds what is on disk.** A candidate child with no materialized overview leaves its span at fill, where the leaf fold would have folded those leaves. In a normal pass this cannot bite — orders are walked finest-first, so every dirty child is written before its parent reads it — but a partially-swept store (or a window a child has no leaves in) can hit it. It is recorded per node at INFO **and** in the overview's own `source_children` attrs (§4.3), and the store self-heals on the next sweep: the parent's summed `generation.n_leaves` moves once the child exists, so the parent is rewritten rather than read as current (pinned end to end through `run_sweep`).

## Phases

- [x] **Phase 1** — config grammar + validation for `fold_source` / `exact_levels`, recorded in the pyramid block (`f202166`)
- [x] **Phase 2** — the cascade fold path (`_cascade_node` / `_fold_child` / `_fold_sources`), bounded per-node memory, regime-aware skip-if-current (`6a23f30`)
- [x] **Phase 3** — per-level fold provenance in the overview attrs and in `materialized.fold_sources` (`e21e705`)
- [x] **Phase 4** — normative `docs/specification.md` §4.3/§4.4/§4.5 + `docs/hive_layout.md` narrative (`24f743c`)
- [x] **Self-review round 1, folded** — `88f5f32` (record an unmaterialized child overview: the one way a cascaded level can under-cover its subtree), `24691c7` (name the declared regime in the retrofit summary), `734147a` (cover the default 2-order gap: 16 children and the span arithmetic that places them)
- [x] **Self-review round 2, folded** — `c1712e5` (compare `fold_from_order` in skip-if-current, not `fold_source` alone), `7671c85` (`source_children`: the cascade's child coverage, in the artifact and in §4.3/§4.4, with both under-coverage terms counted correctly), `fd752af` (warn when `exact_levels` covers every declared level), `745c676` (`materialized.fold_sources` records the regime that wrote, never the plan), `232c70a` (end-to-end partial-materialization and self-heal coverage; the deprecation staying silent on the default path)

## How it was tested

`uv sync --extra test`, then `uv run pytest` — **3291 passed, 38 skipped** (full suite, ~8.9 min). `uv run ruff check src tests` and `uv run ruff format --check src tests` are clean apart from the two pre-existing failures in question (f).

New in `tests/test_sweep_overview.py::TestCascadeFold` (21 tests):

- `test_coarse_level_folds_from_the_finer_overview` — real leaves swept at two orders: only order 1 reaches `_fold_node` (the leaves-read-once property), the order-0 envelope records `cascade` + `fold_from_order`, and the folded values still carry every observation (digest weight sum);
- `test_two_order_gap_places_each_of_the_16_children` — the *default* schedule's gap on a deeper geometry (shard 3 / cell 6): a leaf under `-3231` must land in parent cell 24, which is the assertion that catches an off-by-one in the child span arithmetic;
- `test_exact_fields_are_byte_equal_under_either_source` — the same leaves swept both ways; `count`/`h_min` agree exactly;
- `test_leaves_source_reads_the_leaves_at_every_level`, `test_exact_levels_two_keeps_the_second_level_on_the_leaves` — the deprecated opt-in and the boundary knob, both asserted through the same spy;
- `test_switching_the_source_regenerates_the_coarse_overview`, `test_dropping_an_intermediate_order_regenerates_the_coarse_overview`, `test_pre_376_envelope_entry_reads_as_the_leaves_fold` — the three halves of the skip-if-current change. The middle one uses an **exact-only** field set, so the summed `n_leaves` and the content hash are identical either way and only `fold_from_order` moves (1 -> 2) when the intermediate order is dropped from the declaration;
- `test_a_partially_swept_store_self_heals_on_the_next_sweep` — two sweeps through `run_sweep`: shard A only, then shard B. The parent at the coarse order is **rewritten** (not current), B's span goes from fill to populated while A's is unchanged, and the parent's summed `generation.n_leaves` moves 1 -> 2 — the mechanism the self-heal actually rests on;
- `test_an_unusable_child_still_leaves_the_parent_covering_the_rest`, `test_under_coverage_is_recorded_in_the_overview_attrs`, `test_an_unreadable_child_is_counted_in_both_terms` — the `source_children` record, the parent still covering the children it could read, and both terms of the count (missing and unreadable, over the candidate children);
- `test_windowed_levels_cascade_per_window` — each coarse window cascades from the same window's child object, the all-time one from the child's `all.zarr`;
- `test_plan_names_the_source_of_every_level`, `test_unusable_fold_declaration_falls_back_loudly`, `test_a_boundary_past_the_last_level_folds_everything_from_the_leaves_loudly` — `_fold_sources` derivation, including the wide-gap fallback and the all-levels-exact warning;
- `test_child_that_is_not_an_overview_is_skipped_loudly`, `test_cascade_without_a_materialized_child_is_empty` — the D11 role gate and the under-coverage record;
- `test_every_level_records_its_fold_in_its_attrs`, `test_materialized_fold_sources_carry_forward`, `test_materialized_fold_sources_record_only_a_level_that_wrote`, `test_cascade_generation_stamp_counts_the_underlying_leaves` — the phase-3 provenance and the §4.5 actuals.

Plus grammar tests in `TestPyramidBlock` (defaults, the deprecation warning and its silence on the default path, the boundary-covers-every-level warning, the declared-off form staying `{"orders": []}`, the unvalidated-retrofit fallbacks, and the `validate_config` refusals).

## Questions for review

**(a) The open sub-decision the issue names: is the second level exact too?** The issue says "no precision guarantee past the first (possibly second) skip level", so I implemented the stated default and made the boundary a knob rather than guessing: `exact_levels: 1`. Concrete options:

- **(a1)** keep `exact_levels: 1` as the shipped default (recommended — it is what the issue states, and (a2) is one config line away for any store that wants it);
- **(a2)** make `exact_levels: 2` the default: the second level is then also single-quantization from the leaves, at the cost of re-reading every leaf a second time (the wall-clock cost the cascade removes) and of that level's nodes accumulating a 16x-larger subtree than the finest level's;
- **(a3)** something conditional (e.g. exact while the node's subtree is under some leaf count) — more machinery, and it makes the regime data-dependent rather than declarable, so §4.5 provenance would have to be reported per node instead of per level.

**(b) Where the knob lives in the config.** The issue writes `output.pyramid.overview.fold_source`, but `output.pyramid` has no `overview` nesting today — `spacing`, `orders`, `all_time`, `summarize` all sit directly under `output.pyramid`, and the `overview` level exists only in the *manifest* block. I put the knob next to its siblings (`output.pyramid.fold_source`), which lands it at exactly `pyramid.overview.fold_source` in the manifest — the path the issue names. Say the word if you meant a literal `output.pyramid.overview:` sub-mapping in the YAML and I will nest it (and `spacing`/`orders`/`all_time` with it, or the grammar goes lopsided).

**(c) `src/zagg/sweep_overview.py` is over the ~1,200-line limit.** It was **1,227 lines on `main` before this PR** (already past the bar independently of this change) and is **1,670** after — 1,614 at the point the second self-review read it, plus 56 lines of fold, all inside the paths those findings named. Per CLAUDE.md §4 and the routine's instruction I am **raising this rather than splitting** — splitting a module is a scope change and not mine to take. For the call: the cascade path is self-contained (`_cascade_node`, `_fold_child`, `_fold_sources`, `_empty_slab`, `_overview_basename` — ~250 lines including docstrings, reached from exactly one call site) and would move out cleanly, as would the declaration half (`build_pyramid_block`, `_fold_plan`, `declare_pyramid` and its validators — ~350 lines that share no state with the sweep half). Note PR #369 raises the same question from 1,361 lines, and PR #349 does *not* target this file.

**(d) The §7 conformance-fixture obligation (CLAUDE.md §4).** This PR changes the attrs/manifest grammar, so `docs/specification.md` §4.3/§4.4/§4.5 is updated here. The **fixture** half is a scope call. Restated below, because the first draft of this question rested on a premise that does not hold:

- both committed fixtures **do** carry a `pyramid` block — the stale pre-#201 `{"orders": [], "aggregation": {}}` at `tests/data/spec/minimal/morton_hive.json:17` and `tests/data/spec/kitchen_sink/morton_hive.json:17`. It is stale, not absent;
- and it is not a block that would have to be invented for: it is produced by `hive.build_manifest` -> `build_pyramid_block` (`src/zagg/hive.py:326`), and the generator's own config (`tools/generate_spec_fixtures.py::_config`) declares `store_layout: hive` with `parent_order: 4` and no `output.pyramid` knob — so a plain re-run of the committed generator emits `spacing: 2`, `orders: [2, 0]`, `all_time`, `fields`, and, on this branch, `fold_source` / `exact_levels`. No second generator, no new fixture directory.

So the keys **do** have somewhere to land. The real argument for not landing them here is the collision: regenerating those two fixtures is a wholesale rewrite (`semantic_hash`, `generated_at`, every leaf's bytes, both `*.expected.json`) that conflicts head-on with PR #369, which adds the first `pyramid/` fixture, for no reader benefit. Options: **(d1)** land as-is and add `fold_source` / `exact_levels` / `materialized.fold_sources` / `source_children` to PR #369's `pyramid/` fixture when the second of the two PRs merges (recommended — I will do it on whichever merges second); **(d2)** regenerate the two existing fixtures here and accept the conflict with #369; **(d3)** add an independent `cascade/` fixture here and accept the overlap.

Flagged separately and **not fixed here** (CLAUDE.md §4: flag, don't fix): the committed `{"orders": [], "aggregation": {}}` contradicts §4.5 as this PR restates it — *"The declared-off form is smaller, and `orders` is the only key a reader may assume"* — because it carries an extra `aggregation` key that no current writer produces. An external reader decoding from spec + fixtures alone sees a store the spec says cannot exist. That is pre-existing debris.

**(e) Both `spec` markers stay at `/1`.** `zagg-overview/1` gains `fold_source` (+ `fold_from_order` and `source_children` on a cascade), `zagg-pyramid/1` gains `fold_source`/`exact_levels`/`materialized.fold_sources`. All are additive, and §4.3/§4.5 now spell out the compatibility rules an old artifact needs: **absent `fold_source` MUST be read as `"leaves"`** (the only fold that existed before #376), and absent `source_children` MUST NOT be read as `missing: 0`. Bumping either marker would make every conformant reader refuse every existing manifest (the §4.5 conformance rule is strict-check-and-fail-loudly), so I did not. **(e1)** stay at `/1` (recommended); **(e2)** bump.

**(f) Pre-existing failures I did not touch** (CLAUDE.md §4: flag, don't fix): `uv run ruff check src tests` is red on `main` with `N818 Exception name UnknownCapability should be named with an Error suffix` (`src/zagg/registry.py:64`), and `uv run ruff format --check src tests` flags a python fence in `tests/data/benchmark/README.md`. The PR lint bot's ruleset (`--select=E,F,W,I`) sees neither.

## Relationship to PR #369 (issue #352, open, unmerged)

Orthogonal cost levers on the same module: #369 narrows **which fields** ride each level, this PR changes **what each level folds from**. I based on `main` and did not build on it. No semantic conflict — #369's per-order `fields` filter and this PR's per-order `(fold_source, source_order)` plan are independent maps over the same `for k in orders` loop, and a field narrowed out of the source level simply is not read by the cascade (a field absent from a child overview contributes nothing, the same schema-evolution path as at the leaves).

Textual conflicts are certain, in three places: the `overview` dict in `build_pyramid_block`, the `for k in orders` body of `sweep_overviews`, and the `_roll_node` signature/call (#369 threads a per-order `fields`, this PR a per-order `plan`). Whichever merges second should expect to resolve those by hand — adjacent-line conflicts, not design ones. One interaction is worth checking at that point: a field scheduled at a coarse order but **narrowed out of the finer source order** has nothing to fold from under cascade, which #369's validator — written against the block schedule, not against the fold plan — does not currently refuse. That is a small check to add to `_field_orders` on merge; it cannot be written here without depending on #369.


espg:
🤖 *from Claude*

**Run status — all four phases landed; the PR is complete and waiting on a decision.**

**Phases** (one commit each, on `claude/376-cascade-overview-fold`, based on `main` at `66cb111`):

| | commit |
|---|---|
| Phase 1 — config grammar + validation (`fold_source`, `exact_levels`), recorded in the pyramid block | `f202166` |
| Phase 2 — the cascade fold path (`_cascade_node`/`_fold_child`/`_fold_sources`), bounded per-node memory, regime-aware skip-if-current | `6a23f30` |
| Phase 3 — per-level fold provenance in the overview attrs and in `materialized.fold_sources` | `e21e705` |
| Phase 4 — normative `docs/specification.md` §4.3/§4.4/§4.5 + `docs/hive_layout.md` | `24f743c` |

**Self-review findings folded** (3 of 3, one commit each):

- `88f5f32` — a cascade folds what is on disk, so a candidate child with no materialized overview leaves its span at fill; that under-coverage is now recorded per node with the missing/total count (INFO, not WARNING: a child with no leaf in *this* window has no overview either, which is ordinary on a windowed store, and the abnormal cases are already counted in `failed`/`empty`).
- `24691c7` — `declare_pyramid`'s summary now names the declared regime, so `--declare-pyramid` prints which fold every future sweep of that store will use.
- `734147a` — the integration tests only exercised a 1-order gap, while the *default* schedule's gap is 2 (16 children). Added a deeper-geometry test whose assertion (a leaf under `-3231` must land in parent cell 24) is what catches an off-by-one in the child span arithmetic.

**Findings left standing:** none — all three were diff-scoped and folded. The open design calls are in the PR body under "Questions for review": (a) the issue's own open sub-decision (`exact_levels` 1 vs 2), (b) the config path the issue writes vs. the grammar that exists, (c) module size, (d) the §7 fixture obligation, (e) leaving both `spec` markers at `/1`.

**Gates:** `uv run pytest` — 3282 passed, 38 skipped (full suite). `uv run ruff check src tests` / `ruff format --check src tests` clean apart from the two failures that are already red on `main` (`N818` in `src/zagg/registry.py`, a python fence in `tests/data/benchmark/README.md`) — flagged, not fixed, per CLAUDE.md §4. CI: ruff and both build jobs green; the test matrix was still running at the time of writing.

**Two deviations from CLAUDE.md, recorded rather than glossed:**

1. The PR stays a **draft** even though all phases are complete and findings are folded — the routine driving this run instructs that explicitly, overriding §2's draft→ready flip. Flipping it is one `gh pr ready` away whenever that is wanted.
2. §2 asks for the adversarial review and the fold to run as two separate fresh-context subagents. No subagent-spawning tool was available in this run, so the self-review was performed in-session against the same checklist (CLAUDE.md conformance, the issue's acceptance criteria, missing tests/edge cases, module size, dependency policy, the spec/fixture obligation) and its findings folded as the three `fold review:` commits above. That is one context reviewing its own diff rather than a fresh one, so it is weaker than the intended loop — worth knowing when reading the review findings.

**Label state:** `implement` + `waiting` — every phase is done and nothing is left to advance; what remains are the decisions in "Questions for review", starting with (a), which the issue itself flagged as open. No new dependency was added, no CI/infra/deploy file was touched, and no tag was created.


---

src/zagg/sweep_overview.py
```diff
@@ -873,6 +999,11 @@ def _roll_node(
         isinstance(existing_entry, dict)
         and existing_entry.get("generation") == fold["generation"]
         and existing_entry.get("content_hash") == fold["content_hash"]
+        # A stored overview folded the other way is NOT current, even when the
+        # bytes agree — the exact folds are byte-equal under either source, so
+        # only the recorded regime distinguishes them (issue #376). Entries
+        # predating #376 carry no key and read as the leaves fold they were.
+        and existing_entry.get("fold_source", "leaves") == fold["fold_source"]
```
espg, line 1029: 🤖 *from Claude (review)*

**The regime check is half a check: `fold_from_order` is not compared, so a cascaded level can keep a stale — and now spec-normative — source order forever.**

```python
and existing_entry.get("fold_source", "leaves") == fold["fold_source"]
```

The reasoning in the PR body ("only the recorded regime distinguishes them") applies just as much to *which level* a cascade folded from, and `fold_from_order` is written into both the envelope entry (`_roll_node`) and the overview attrs (`_fold_provenance`), and §4.3 now documents it as the reader-facing companion of `fold_source`:

> ```json
> "fold_source": "cascade", "fold_from_order": 3
> ```

Concrete sequence that produces a wrong recorded value, on an **exact-only field set** (`count`/`h_min` — the common case, and the one this PR's own `test_exact_fields_are_byte_equal_under_either_source` pins as byte-equal across folds):

1. `orders: [7, 5, 3, 1]`, `exact_levels: 1`, a geometry with `cell_order - shard_order >= 4` (the California store in #376 is depth 10). Sweep → order 3's entry and attrs record `fold_source: "cascade", fold_from_order: 5`.
2. Edit the declaration to `orders: [7, 3, 1]`. `_fold_sources` now plans `3 -> ("cascade", 7)` (gap 4 <= depth, so no fallback).
3. Re-sweep the same leaves. `generation.n_leaves` is unchanged — it is summed from the children's own `n_leaves`, and summing over the order-5 children and over the order-7 children both total the leaves under the node. `max_leaf_timestamp` likewise. `content_hash` is unchanged because the exact laws are associative. `fold_source` is still `"cascade"`.
4. → `current`. The artifact keeps `fold_from_order: 5`, naming a level that is no longer declared and will never be refreshed again.

Same shape in the other direction (`orders: [7, 1]` -> `[7, 4, 1]` leaves order 1 claiming `fold_from_order: 7`), and it is also reachable via the sweep-time `bad` order filter at `sweep_overview.py:747` dropping a declared order.

It is one line — add `existing_entry.get("fold_from_order") == fold.get("fold_from_order")` to the conjunction (both sides are `None` on the leaves path, so the leaves regime is unaffected). A regression test in the shape of `test_switching_the_source_regenerates_the_coarse_overview`, but dropping an intermediate order instead of flipping `fold_source`, would pin it.

---

src/zagg/sweep_overview.py
```diff
@@ -1048,9 +1188,239 @@ def _fold_node(
...
+        # leaves its span at fill until a sweep covers it. Recorded rather
+        # than warned — a candidate child with no leaf in THIS window has no
+        # overview either, which is ordinary on a windowed store; the cases
+        # that are not ordinary (a fold that failed, a node that folded
+        # nothing) are already counted in ``failed``/``empty``.
+        logger.info(
+            f"sweep[overview]: node {node} order {k} window {key!r}: {missing} of "
+            f"{missing + n_sources} candidate child overviews at order {source_order} are "
+            f"not materialized — their cells stay fill until a sweep regenerates them"
+        )
```

**Under-coverage is recorded in a log line, not in the store — so a reader cannot tell an unmaterialized child from genuine emptiness.** Two separate problems here.

**(1) It never reaches an artifact.** Under `fold_source: "leaves"`, a fill cell means "no leaf contributed" and nothing else. Under `cascade` it means *either* that *or* "a child overview was not on disk when this parent was folded" — and the only trace is `logger.info` in a sweep process that has already exited. §4.3, as this PR writes it, tells a reader the cascade "differs only in the values its `approximate` fields carry" (§4.4, `docs/specification.md:576`) — but an under-covered cascade also differs in *which cells are populated at all*, in every class, and that is exactly the distinction the issue's provenance requirement exists to make legible ("a reader deserves to know which regime a level is in").

`_fold_provenance` already assembles the per-overview provenance dict and §4.3 already tells readers to tolerate additional keys, so this is additive on `/1` and lands next to the keys the PR is already adding — e.g.

```python
entry["source_children"] = {"folded": n_sources, "missing": missing, "failed": failed}
```

plus a sentence in §4.3 saying a nonzero `missing`/`failed` means the level under-covers its subtree and a fill cell there is not evidence of emptiness. Without it, `materialized.fold_sources` tells a reader *how* a level was made but not whether it is complete.

**(2) The count itself is wrong when a child is unreadable.** `missing` is only incremented on the `stamp is None` branch at line 1300. A child that fails to open (line 1266) or fails the role/order/shape gate (line 1296) also leaves its span at fill — that is under-coverage too — but it lands in `counts["failed"]` and is excluded from *both* terms of the message:

```
f"{missing} of {missing + n_sources} candidate child overviews at order {source_order} are "
f"not materialized"
```

With 16 candidate children, 1 unstamped, 3 unreadable and 12 folded, this prints "1 of 13", understating the gap and misnaming the denominator ("candidate child overviews" is 16). Track a third counter and include it in both the message and the recorded provenance.

---

src/zagg/sweep_overview.py
```diff
@@ -249,27 +273,84 @@ def build_pyramid_block(config, shard_order: int) -> dict:
...
+    try:
+        exact_levels = DEFAULT_EXACT_LEVELS if raw is None else int(raw)
+    except (TypeError, ValueError):
+        exact_levels = 0
+    if exact_levels < 1:
+        logger.warning(
+            f"pyramid: exact_levels must be an int >= 1 (got {raw!r}); declaring the "
+            f"default {DEFAULT_EXACT_LEVELS}"
+        )
+        exact_levels = DEFAULT_EXACT_LEVELS
```
espg, line 317: 🤖 *from Claude (review)*

**`exact_levels >= len(orders)` silently buys the whole deprecated regime, with none of its warnings and a declaration that says the opposite.**

`exact_levels` is only ever checked against `1`, here and at `_fold_sources` (`max(1, int(...))`, line 1401) and in `_validate_pyramid_fold` (`config.py:2478`). Nothing compares it to the number of declared levels. So:

- config `output.pyramid: {exact_levels: 3}` with `shard_order: 6` → `build_pyramid_block` derives `orders: [4, 2, 0]` and declares `{"fold_source": "cascade", "exact_levels": 3}`;
- `_fold_sources` then returns `{4: ("leaves", None), 2: ("leaves", None), 0: ("leaves", None)}` — every level re-reads the whole subtree, which is precisely the state-scale wall #376 exists to remove;
- no deprecation warning fires (line 318 gates on `fold_source == "leaves"`, and the declared source is `"cascade"`);
- the manifest, and §4.5's reader contract for it, say `"fold_source": "cascade"`.

This is not hypothetical — `tests/test_sweep_overview.py:529` asserts exactly this config validates clean:

```python
validate_config(self._cfg(store_layout="hive", pyramid={"exact_levels": 3}))
```

and `self._cfg()` is the `shard_order=6` / `orders: [4, 2, 0]` geometry used two tests earlier.

`_fold_plan` already has `orders` in hand, so the cheapest honest fix is one more branch here: when `exact_levels >= len(orders)` and `orders` is non-empty, warn that every declared level will fold from the leaves and name the deprecated regime the store is effectively getting. (`_validate_pyramid_fold` cannot do it — it has no `shard_order`, so it cannot know the derived schedule. Worth mirroring the check in `_fold_sources` too, since a hand-edited manifest reaches the sweep without either declaration path.)

---

src/zagg/sweep_overview.py
```diff
@@ -686,7 +785,7 @@ def sweep_overviews(store_root: str, manifest: dict, by_shard: dict, *, store_kw
                 if entry is not None:
                     entries[key] = entry
             if entries:
-                materialized.add(k)
+                materialized[k] = plans[k][0]
```

**`materialized.fold_sources` records the plan, not the actual — and this PR's own §4.5 text promises the opposite.**

```python
if entries:
    materialized[k] = plans[k][0]
```

`entries` is truthy for *carried-forward* entries too, not only written ones: `_roll_node` returns `existing_entry` on both the `current` path and the `fold is None` (empty) path. So a level where every `(node, window)` came back empty — under cascade that is every child overview missing/unreadable, `_cascade_node` returning `None` at line 1325 — still sets `materialized[k] = "cascade"` while every surviving artifact at that level carries `zagg_overview.fold_source: "leaves"` from an earlier sweep. Any write at *another* level then makes `counts["written"]` truthy and the false claim is PUT to the manifest.

That contradicts the paragraph this PR adds at `docs/specification.md:679`:

> — the orders it has written and, per level, the §4.3 regime that wrote it […] Both are **actuals**

The `orders` half is unchanged from `main` and is fine; it is the new `fold_sources` value that is derived from the declaration rather than from anything on disk. Cheapest fix that keeps it an actual: only record `k` when the level actually wrote (thread the per-level write count, or set `materialized[k]` from the `fold_source` of the entries `_roll_node` returned as fresh rather than from `plans[k][0]`).

Related: `tests/test_sweep_overview.py:1040` (`test_materialized_fold_sources_carry_forward`) only exercises the case where the plan and the disk agree, so it does not catch this.

---

tests/test_sweep_overview.py
```diff
@@ -731,6 +804,270 @@ def test_coarser_than_shard_order_accumulates(self, tmp_path):
...
+            COMPOSABLE,
+            CELL_ORDER,
+            SHARD_ORDER,
+            counts,
+            {},
+        )
+        assert fold is None and counts["failed"] == 0
+        # An unmaterialized child is not an error, but it IS under-coverage:
+        # the cascade folds what is on disk, so it is recorded.
+        assert "1 of 1 candidate child overviews at order 1 are not materialized" in caplog.text
```
espg: 🤖 *from Claude (review)*

**The partial-materialization path is the one new behavior with no end-to-end test, and the test that does cover it asserts an implementation detail.**

This test calls `_cascade_node` directly with a hand-built `counts` dict and a literal `["-311"]` node-shard list, then asserts a log substring:

```python
assert "1 of 1 candidate child overviews at order 1 are not materialized" in caplog.text
```

That pins the wording of an INFO message, not a behavior. Nothing here goes through `run_sweep`, so nothing checks what a partially-materialized store actually ends up holding.

The property that makes under-coverage tolerable — the PR body's item (2), *"In a normal pass this cannot bite […] but a partially-swept store (or a window a child has no leaves in) can hit it"* — is the claim that a later sweep repairs it, and it is untested. It is also not self-evident: repair rests entirely on the parent's summed `generation.n_leaves` changing when the previously-missing child appears, which is a non-obvious consequence of `_cascade_node:1307`. A two-sweep regression through `run_sweep` would pin it end to end:

1. sweep with only shard A's leaf present → parent at the coarse order writes with B's span at fill;
2. add shard B's leaf and re-sweep → assert the parent is **rewritten** (not `current`), that B's span is now populated, and that A's span is unchanged.

That single test covers the placement arithmetic under partial coverage, the self-heal, and the generation-stamp derivation at once — and it is the natural home for whatever store-visible under-coverage record comes out of the `_cascade_node` finding above.

Two smaller gaps worth closing in the same class, both cheap:

- no test that a child skipped by the `except` branches (unreadable / failing the role gate) still yields a *written* parent covering the remaining children — `test_child_that_is_not_an_overview_is_skipped_loudly` uses a single child, so it only exercises `n_sources == 0`;
- no test that the deprecation warning does **not** fire on the default path. `_fold_plan` is called on every `build_pyramid_block`, and `hive.py:326` runs it at every manifest bootstrap; a `caplog`-empty assertion for the cascade default is one line and guards against the warning becoming noise later.

---

docs/specification.md
```diff
@@ -621,7 +669,23 @@ overview family under the versioned `pyramid` block:
 - The sweep MAY additionally record materialized-actuals bookkeeping in the
   block; the template-time declaration above is never rewritten by the
   sweep, and the `pyramid` block is excluded from the manifest's frozen
-  append-guard keys.
+  append-guard keys. zagg's sweep writes
+
+  ```json
+  "materialized": {"orders": [3, 1],
+                   "fold_sources": {"3": "leaves", "1": "cascade"},
```
espg, line 698: 🤖 *from Claude (review)* — **STANDING, not a fold item.**

**Question (d)'s premise does not hold as written, even though its conclusion (d1) may still be the right call.** Flagging so the decision gets made on the real tradeoff rather than a stated blocker that is not one.

The PR body says:

> neither committed fixture declares a pyramid at all (both carry the stale pre-#201 block `{"orders": [], "aggregation": {}}` […]), so there is no pyramid grammar in `tests/data/spec/` for these keys to appear in

Both halves are off. The fixtures **do** carry a `pyramid` block (`tests/data/spec/minimal/morton_hive.json:17`, `tests/data/spec/kitchen_sink/morton_hive.json:17`) — it is stale, not absent. And it is not a block a `pyramid/` fixture would have to be invented for: it is produced by `hive.build_manifest` → `build_pyramid_block` (`src/zagg/hive.py:326`), and the generator's own config (`tools/generate_spec_fixtures.py::_config`) declares `store_layout: hive` with `parent_order: 4` and no `output.pyramid` knob — so a plain re-run of the committed generator emits `spacing: 2`, `orders: [2, 0]`, `all_time`, `fields`, and, on this branch, `fold_source` / `exact_levels`. No second generator, no new fixture directory.

Worth noting separately: the committed `{"orders": [], "aggregation": {}}` actively contradicts §4.5 as this PR restates it three paragraphs above — *"The declared-off form is smaller, and `orders` is the only key a reader may assume"* — because it carries an extra `aggregation` key that no current writer produces. An external reader decoding from spec + fixtures alone sees a store the spec says cannot exist. That is pre-existing debris, **not** this PR's to fix (CLAUDE.md §4: flag, don't fix).

So the accurate framing of (d) is: the fixture obligation *could* be met here by regenerating the two existing fixtures, but doing so is a wholesale rewrite (`semantic_hash`, `generated_at`, every leaf's bytes and both `*.expected.json`) that would conflict head-on with PR #369 for no reader benefit — which is a real argument for (d1) and a much better one than "there is nowhere for the keys to land." Deciding between (d1) and regenerating is a scope call, so it stays here rather than becoming a fold item.

---

src/zagg/sweep_overview.py
espg, line 1: 🤖 *from Claude (review)* — **STANDING, not a fold item.**

**Module size: 1,227 → 1,614 lines, past CLAUDE.md §4's ~1,200 ceiling.** Recording it because §4 requires it to be recorded, not because anything should happen in this PR: question (c) already raises it, the raise is the correct move (splitting a module is a scope change and not the fold agent's to take), and the pre-existing 1,227 means the bar was already crossed before this branch. Left standing for a human ruling, alongside the same question from PR #369 at 1,361 lines.

For whoever rules: the seam question (c) proposes is the right one — `_cascade_node` / `_fold_child` / `_fold_sources` / `_empty_slab` / `_overview_basename` is a genuinely self-contained fold path with a narrow interface (it is reached from exactly one call site, `_roll_node:964`), and the declaration half (`build_pyramid_block`, `_fold_plan`, `declare_pyramid`, `_semantic_guard`, `_validate_block_against_store`, `_field_drift`) is ~350 lines that share no state with the sweep half at all. Both cuts follow existing responsibility lines rather than inventing new ones.

No undiscussed dependency was added — the diff introduces no new imports beyond modules already imported elsewhere in this file (`zarr`, `zagg.hive`, `zagg.store`, `zagg.windows`), and `pyproject.toml` is untouched (§4).

---

espg (commented):
🤖 *from Claude (review)*

Adversarial self-review of the cascade fold, fresh context, against issue #376's acceptance criteria and CLAUDE.md. Six inline comments — **four fold items, two standing.**

**What holds up.** I went looking for arithmetic and memory-bound bugs and did not find any.

- The span arithmetic is right for both gaps. `n_cells = 4^(target_order - k)`, `factor = 4^gap`, `span = n_cells // factor`, `start = _rel_rank(child, node) * span`: every overview slab in the tree holds `4^(cell_order - shard_order)` cells regardless of level, so a child's slab and its parent's are the same length and the child's region is exactly `span` contiguous parent cells at nested-morton offset `rel_rank * span`. `_fold_child`'s `values[j*factor : (j+1)*factor]` groups the same way `fold_dense`'s `reshape(-1, factor)` does, and `_rel_rank` is the same placement primitive `_fold_node` already used on `main`. The gap-2 test (`-3231` → parent cell 24 on the shard-3/cell-6 geometry) is a genuine off-by-one catcher, not a tautology.
- The memory bound is real. `_fold_child` rebinds `values` per field and returns only `span`-sized partials, so peak resident is the output slab plus *one field's* child slab — tighter than the docstring's claim, and constant in the subtree where `_fold_node`'s `digests[name][j].append(...)` accumulator is not. The `gap > depth` guard is load-bearing (`span` would be 0 and `fold_dense` would raise) and is covered.
- Leaves-read-once is genuinely pinned rather than asserted in prose — the `_fold_node` spy in `_leaf_orders` is the right instrument, and it is reused for both the `leaves` opt-in and the `exact_levels: 2` boundary.
- The deprecated `leaves` path is byte-identical to `main` — `_roll_node` dispatches to an unchanged `_fold_node`, whose only edit is the `_empty_slab` extraction. `fold_windows` still reaches it. The deprecation warning is gated on `orders` being non-empty so the declared-off form stays silent, and `hive.py:326` is a bootstrap-time call, not a per-worker one, so there is no warning storm.
- Provenance carries the leaf-derived generation correctly: `n_leaves` sums the children's own `generation.n_leaves` and `max_leaf_timestamp` maxes theirs, so staleness stays keyed to leaf commits rather than to overview writes. `granule_count` and `time_range` propagate without double counting.
- Staying at `spec: /1` (question (e1)) is right. Both additions are additive, §4.3 already tells readers to tolerate extra keys, and the "absent `fold_source` MUST be read as `leaves`" rule is the compatibility hinge — pinned by `test_pre_376_envelope_entry_reads_as_the_leaves_fold`, which is the test that matters most for not churning existing stores.

**Fold items.**

1. `sweep_overview.py:1006` — the skip-if-current regime check compares `fold_source` but not `fold_from_order`. Dropping an intermediate declared order leaves a cascaded level reading as `current` with a stale, spec-normative `fold_from_order` naming a level that no longer exists. Byte-equality of the exact class means it is never corrected. One line.
2. `sweep_overview.py:1311` — under-coverage never reaches an artifact (INFO log only), so a reader cannot distinguish "child overview not materialized" from "genuinely empty" — a distinction that does not exist under `leaves`. The `missing` counter also excludes children lost to the `except` branches, so the logged denominator understates the gap.
3. `sweep_overview.py:312` — `exact_levels >= len(orders)` silently yields the deprecated all-leaves regime with no deprecation warning and a manifest that declares `"cascade"`. `tests/test_sweep_overview.py:529` pins exactly that config as valid.
4. `sweep_overview.py:788` — `materialized.fold_sources` records the *plan*, not what was written, which contradicts the "Both are **actuals**" sentence this PR adds at `docs/specification.md:679`.

Plus `tests/test_sweep_overview.py:985` — the partial-materialization path has no end-to-end coverage, and the test that touches it asserts a log substring rather than store state. The PR body's self-heal claim for a partially-swept store is untested and non-obvious.

**Standing (human decision, not the fold agent's).**

5. `docs/specification.md:676` — question (d)'s stated blocker does not hold: the fixtures do carry a `pyramid` block (stale, not absent), and a plain re-run of the committed generator would emit the new keys. (d1) may still be right, but on conflict grounds with #369, not on "there is nowhere for the keys to land."
6. `src/zagg/sweep_overview.py` (file) — 1,614 lines, past §4's ceiling. Already raised as question (c); raising rather than splitting is correct.

On question (a): nothing in the diff argues against **(a1)**. `exact_levels` is a clean, testable boundary, `_fold_sources` makes the regime declarable per level rather than data-dependent, and (a2) is one config line. Worth noting that (a2) costs more than the PR body says — under `exact_levels: 2` the second level's nodes accumulate a 16x subtree through the *unbounded* `_fold_node` accumulator, which is the memory wall #376 exists to remove, not merely a wall-clock cost.

No dependency was added; `pyproject.toml` and everything under `.github/` and `deployment/` are untouched.

---

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Folded in `c1712e5`. The conjunction now compares both halves of the regime:

```python
and existing_entry.get("fold_source", "leaves") == fold["fold_source"]
and existing_entry.get("fold_from_order") == fold.get("fold_from_order")
```

Both sides are `None` on the leaves path — `_fold_node`'s payload carries no `fold_from_order` and `_roll_node` only writes the key when the fold has one — so the leaves regime and every pre-#376 entry are untouched (`test_pre_376_envelope_entry_reads_as_the_leaves_fold` still passes unchanged).

Regression test `TestCascadeFold::test_dropping_an_intermediate_order_regenerates_the_coarse_overview`, built on your sequence: `orders: [2, 1, 0]` on a shard-3/cell-6 geometry with an **exact-only** field set, so the summed `n_leaves` and the content hash are byte-identical either way and re-declaring as `[2, 0]` moves only `fold_from_order` (1 -> 2). I verified it fails on the pre-fold tree — the node reads as `current` and keeps `fold_from_order: 1` — and passes after: `written == 1, current == 2`, with both the envelope entry and the overview attrs reading 2.

---

src/zagg/sweep_overview.py

Both halves folded in `7671c85`.

**(1) It reaches an artifact now.** A cascade records its coverage through `_fold_provenance`, next to the keys this PR was already adding:

```json
"source_children": {"folded": 15, "missing": 1, "unreadable": 0}
```

`docs/specification.md` §4.3 documents it normatively: a nonzero `missing`/`unreadable` means the level **under-covers its subtree**, a fill cell there is not evidence of emptiness, a reader MUST tolerate the key's absence (a `leaves` level, or a pre-#376 artifact) and MUST NOT read absence as `missing: 0`. Additive on `/1` under the existing tolerate-additional-keys rule. §4.4's *"differs only in the values its `approximate` fields carry"* is corrected in the same commit — it now says the under-covered case also differs in which cells are populated at all, in every class, and points at `source_children` as the key that says so.

**(2) The count is fixed.** `unreadable` is incremented on both `except` branches, the candidate list is hoisted so the denominator is the candidate children themselves, and both terms are in the message:

```python
f"{missing + unreadable} of {len(children)} candidate child overviews at order "
f"{source_order} are not usable ({missing} not materialized, {unreadable} unreadable)"
```

Your 16/1/3/12 example now prints `4 of 16 candidate child overviews at order N are not usable (1 not materialized, 3 unreadable)`.

Pinned by `test_an_unreadable_child_is_counted_in_both_terms` (3 candidates: 1 folded, 1 unstamped, 1 failing the role gate -> `{"folded": 1, "missing": 1, "unreadable": 1}` and "2 of 3") and `test_under_coverage_is_recorded_in_the_overview_attrs`, which drives it through `run_sweep` and also asserts the exact-from-leaves level carries **no** such key.

---

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

Folded in `fd752af`, in both places you name.

`_fold_plan` has `orders` in hand, so the declaration path warns when the boundary covers the whole schedule:

```python
elif orders and exact_levels >= len(orders):
    logger.warning(
        f"pyramid: exact_levels {exact_levels} covers all {len(orders)} declared levels "
        f"{orders}, so EVERY level folds from the raw leaves — the DEPRECATED 'leaves' "
        f"regime (issue #376) under a 'cascade' declaration; lower it to keep the cascade"
    )
```

and `_fold_sources` mirrors it, for the hand-edited manifest that reaches the sweep without either declaration path.

Tests: `test_exact_levels_past_the_last_level_warns_it_is_the_leaves_regime` uses your exact repro (`pyramid={"exact_levels": 3}` at `shard_order: 6`, derived `orders: [4, 2, 0]`) and asserts the block still declares `cascade` while the warning names the collapse; `test_a_boundary_past_the_last_level_folds_everything_from_the_leaves_loudly` pins the `_fold_sources` mirror, including that the plan really is all-leaves.

`tests/test_sweep_overview.py:529` is deliberately untouched — `validate_config` still accepts the config, for the reason you give (it has no `shard_order`, so it cannot know the schedule); the warning lives where the schedule is known. `232c70a` adds the complementary assertion that neither warning fires on the shipped default.

---

src/zagg/sweep_overview.py

Folded in `745c676` — **fixing the code**, not the spec sentence. The evidence points that way: the actual *is* determinable at that point (`_roll_node` returns the fresh entry, and that entry carries the `fold_source` that just wrote), and §4.5's "Both are **actuals**" is what makes the map worth reading at all — weakening it would leave readers a key that means "the plan, sometimes."

```python
if entry is not None:
    entries[key] = entry
    if counts["written"] > before:
        materialized[k] = entry["fold_source"]
if entries:
    materialized.setdefault(k, None)
```

The `setdefault(k, None)` keeps the `orders` half exactly as it was on `main` — an entry still means the order is materialized, written this pass or not — while `_update_manifest_pyramid` now skips the `None` when merging `fold_sources`, so a level that wrote nothing keeps whatever regime its own earlier sweep recorded. The docstring says so explicitly.

New test `test_materialized_fold_sources_record_only_a_level_that_wrote` reproduces your scenario end to end: a store swept under `leaves`, re-declared to `cascade` **in place** (so the prior actuals survive the edit), with `_cascade_node` raising so order 0 writes nowhere while order 1 writes. Pre-fold it records `{"1": "leaves", "0": "cascade"}` against artifacts that say `leaves`; post-fold `{"1": "leaves", "0": "leaves"}`, with `materialized.orders` still `[0, 1]` and the order-0 overview's own attrs confirming `leaves`. `test_materialized_fold_sources_carry_forward` is unchanged and still passes.

---

tests/test_sweep_overview.py
espg: 🤖 *from Claude*

All three folded in `232c70a`.

**The two-sweep regression** is `test_a_partially_swept_store_self_heals_on_the_next_sweep`, both passes through `run_sweep`, exactly the shape you specify: shard A's leaf swept first (B's leaf committed and in the coverage MOC, but not in that run's refs, so B has no overview), then B. It asserts

- the parent at the coarse order is **rewritten**, not `current` — `written == 2, current == 0`;
- B's span goes fill -> populated (`counts0[4]`: 0 -> 1) while A's is unchanged (`counts0[0] == 2`);
- and the mechanism the repair actually rests on: the parent's summed `generation.n_leaves` moves 1 -> 2, which is what makes the stamp differ at all;
- plus the `source_children` record from the finding above going `{"folded": 1, "missing": 1}` -> `{"folded": 2, "missing": 0}`.

**The `except`-branch gap** is `test_an_unusable_child_still_leaves_the_parent_covering_the_rest`: two children, one broken by the role gate after a full sweep. Its own level reads as `current` (envelope and D4 stamp untouched), so it is not silently regenerated, and the parent is still **written** — `written == 1, current == 2, failed == 1` — covering the child it could read and recording `{"folded": 1, "missing": 0, "unreadable": 1}`.

**The deprecation-silence assertion** is `test_the_deprecation_stays_silent_on_the_default_path`: `build_pyramid_block` (the `hive.py:326` bootstrap path) plus `_fold_sources` on the default, then `assert "DEPRECATED" not in caplog.text`.

The log-substring assertion in `test_cascade_without_a_materialized_child_is_empty` survives — its wording moved with `7671c85` — but it is no longer the only coverage of that path.

---

docs/specification.md
espg: 🤖 *from Claude*

**Left standing** — the fixture decision is a scope call and stays for a human. What was mine to do is the correction, and question (d) in the PR body is rewritten to state the facts you established. It now says:

- both committed fixtures **do** carry a `pyramid` block — the stale pre-#201 `{"orders": [], "aggregation": {}}` at `tests/data/spec/minimal/morton_hive.json:17` and `tests/data/spec/kitchen_sink/morton_hive.json:17`. Stale, not absent;
- and that block is produced by `hive.build_manifest` -> `build_pyramid_block` (`src/zagg/hive.py:326`) from the generator's own config — `tools/generate_spec_fixtures.py::_config` declares `store_layout: hive` with `parent_order: 4` and no `output.pyramid` knob — so a plain re-run of the committed generator emits `spacing: 2`, `orders: [2, 0]`, `all_time`, `fields`, and, on this branch, `fold_source` / `exact_levels`. No second generator, no new fixture directory. I verified both against the tree.

The recommendation is still **(d1)**, but on your argument rather than the old one: regenerating those two fixtures is a wholesale rewrite (`semantic_hash`, `generated_at`, every leaf's bytes, both `*.expected.json`) that collides head-on with PR #369's `pyramid/` fixture for no reader benefit — not "there is nowhere for the keys to land." The options are relabelled to match: (d2) is now "regenerate the two existing fixtures here", (d3) "add an independent `cascade/` fixture".

The `{"orders": [], "aggregation": {}}` versus §4.5 contradiction is flagged in the same place as pre-existing debris this PR does not touch (CLAUDE.md §4: flag, don't fix).

---

src/zagg/sweep_overview.py
espg: 🤖 *from Claude*

**Left standing**, as recorded — no split here. Splitting a module is a scope change and not the fold's to take, and question (c) already raises it for a ruling.

One number to refresh, since this round moved it: the file is now **1,670** lines. It was 1,227 on `main` before this PR and 1,614 at the commit you reviewed; the five folds added 56 lines, all inside the paths those findings named (the skip-if-current conjunction, `_cascade_node`'s coverage counters, `_fold_plan`/`_fold_sources`' new warning, and `_update_manifest_pyramid`'s docstring). Question (c) is updated with that figure and with the seam you describe — the cascade path reached from exactly one call site, and the ~350-line declaration half that shares no state with the sweep half.

Confirming the dependency note too: no new imports beyond modules already imported in this file, and `pyproject.toml` is untouched by this round as well.

---

espg:
🤖 *from Claude*

**Fold round 2 complete** — the five diff-scoped findings from [the adversarial self-review](https://github.com/englacial/zagg/pull/379#pullrequestreview-4853408257) are folded, one commit each, and the two standing findings are answered on their threads. Every inline thread has a reply; none were resolved.

| # | Finding | Commit | What changed |
|---|---|---|---|
| 1 | [`sweep_overview.py:1006`](https://github.com/englacial/zagg/pull/379#discussion_r3711663819) — skip-if-current ignores `fold_from_order` | `c1712e5` | The current-check compares the pair. Regression test drops an intermediate declared order on an exact-only field set, where `n_leaves` and the content hash are identical and only `fold_from_order` moves. |
| 2 | [`sweep_overview.py:1311`](https://github.com/englacial/zagg/pull/379#discussion_r3711665546) — under-coverage never reaches an artifact; the count is wrong | `7671c85` | New `source_children` `{"folded", "missing", "unreadable"}` in the overview attrs, normative in §4.3 (+ the §4.4 sentence corrected); unreadable children now counted, and the denominator is the candidate children. |
| 3 | [`sweep_overview.py:312`](https://github.com/englacial/zagg/pull/379#discussion_r3711667057) — `exact_levels >= len(orders)` silently buys the deprecated regime | `fd752af` | Warns in `_fold_plan` (the only place that knows the derived schedule) and mirrored in `_fold_sources` for hand-edited manifests. |
| 4 | [`sweep_overview.py:788`](https://github.com/englacial/zagg/pull/379#discussion_r3711668763) — `materialized.fold_sources` records the plan, not the actual | `745c676` | Fixed the **code**, not the §4.5 sentence: only a level that actually wrote records a regime; `materialized.orders` keeps its `main` semantics. |
| 5 | [`test_sweep_overview.py:985`](https://github.com/englacial/zagg/pull/379#discussion_r3711670748) — partial materialization has no end-to-end coverage | `232c70a` | Two-sweep self-heal through `run_sweep` (rewritten not current, spans, and the summed `n_leaves` the repair rests on), the `except`-branch case, and the deprecation staying silent on the default path. |

**Left standing for a human ruling** (both were flagged as standing by the review itself, and both imply a scope decision):

- [the §7 fixture obligation](https://github.com/englacial/zagg/pull/379#discussion_r3711672657) — question (d) in the PR body is **corrected** to state the facts accurately (the fixtures do carry a stale `pyramid` block; the committed generator would emit the new keys on a plain re-run), and the recommendation stays (d1) on the conflict-with-#369 argument rather than the "nowhere to land" one. Landing or regenerating a fixture is the scope call and stays open;
- [module size](https://github.com/englacial/zagg/pull/379#discussion_r3711674211) — 1,670 lines now (1,227 on `main` before this PR, 1,614 at review time; this round added 56). Already question (c); no split taken.

**Gate** (`uv sync --extra test`):

- `uv run pytest -v` — **3291 passed, 38 skipped**, 8m54s (3282/38 before this round; the 9 new tests are the five findings' coverage);
- `uv run ruff check src tests` — one error, **pre-existing on `main`**: `N818` on `src/zagg/registry.py:64`. Not fixed (CLAUDE.md §4: flag, don't fix), and outside the PR lint bot's `--select=E,F,W,I` ruleset;
- `uv run ruff format --check src tests` — one file would be reformatted, **pre-existing on `main`**: a python fence in `tests/data/benchmark/README.md`. Not fixed, same reason. Every file this PR touches is clean under both.

Both are question (f) in the PR body.

**Label state: `implement` + `waiting`.** The PR stays a **draft** — questions (a), (b), (c), (d) and (e) all need a decision, so the ball is in a human's court rather than the draft→ready flip.

---

espg:
🤖 *from Claude*

**All five lettered questions are RULED** (espg, in-session, 2026-08-05) — nothing further blocks review/merge of this PR:

| question | ruling |
| --- | --- |
| (a) `exact_levels` default | **(a1)** — ships as `exact_levels: 1`. Noting for the record: under the pyramid v2 umbrella (#381 point (1)) the exact/approximate boundary becomes the fleet/sweep boundary itself, so this knob is expected to go vestigial; approved as-is precisely to avoid over-investing in it. |
| (b) knob path | **flat `output.pyramid.fold_source` as implemented** — no `overview:` sub-mapping is added to the YAML; the manifest still records it at `pyramid.overview.fold_source`, the path issue #376 named. Same vestigial note as (a) applies. |
| (c) `sweep_overview.py` at 1,670 lines | **overage accepted, no split in this PR** — espg: line overage is not a present concern. New pyramid-v2 code is landing in a new `src/zagg/pyramid.py` (issue #382, in flight), not here. |
| (d) §7 fixture obligation | **(d1)** — the two stale committed fixtures are not regenerated here. Updated landing spot for the new keys' fixture coverage: PR #369 is now closed, so the declaration-side keys (`fold_source`, `exact_levels`, `materialized.fold_sources`) ride the `zagg-pyramid/2` fixture being built under issue #382; until then they are spec-prose-covered per this PR's §4.3/§4.5 text. |
| (e) spec markers | **(e1)** — `zagg-overview/1` and `zagg-pyramid/1` both stay at `/1`; the keys are additive and the absent-key compatibility rules are normative in this PR (absent `fold_source` reads as `"leaves"`; absent `source_children` never reads as `missing: 0`). Not in tension with issue #382's coming `zagg-pyramid/2`: that bump is for the level-geometry change (genuinely breaking for constant-depth readers), not for these keys. |

Sequencing context: PR #380 is blocked behind this PR, and the issue #384 staged sweep keeps this PR's cascade as the retrofit regime — so this is the merge that unblocks the rest of the line.

PR #395, opened 2026-08-05
land the AOI/CONUS shardmap builders (issue #372)

### LINKED ISSUE #372: conus shardmap builder mislabels the considered count as intersecting

espg:
Closes #372

## What / why

The AOI/CONUS shardmap builder scripts have been working-tree-only since issue #202: every artifact they produced (the CONUS/CA/NEON shardmaps, per-shard count tables, cost stats) is committed history, but the code that made them never landed. The PR #378 review round made that untenable — the `granules_assigned` premise for issue #372's one-line fix landed on `main` in `bacca5c` ([ruling comment, row (1)](https://github.com/englacial/zagg/pull/378#issuecomment-5185073447)), while the fix's *target files* were still untracked ([status-correction comment on the issue](https://github.com/englacial/zagg/issues/372#issuecomment-5185079870)). This PR lands the builders and applies the #372 fix on top.

Files landed (repo-relative):

- `data/build_aoi_shardmap.py` — generic AOI shardmap + cost-ceiling builder
- `data/conus/build_conus_shardmap.py` — the CONUS builder (issue #202 leg 4a; #372's target)
- `data/conus/build_conus_polygon.py` — regenerates the CONUS AOI geojson from its documented public source
- `data/conus/conus.geojson` — default `--polygon` input (22 KB)

Two portability fixes on landing: `DEFAULT_CATALOG` in the CONUS builder was a machine-absolute path (`/Users/espg/software/zagg/data/...`), now repo-relative (`data/atl03_v007/atl03_v007_full.parquet`) matching the AOI builder; and the AOI builder's `DEFAULT_CONFIG` pointed at an uncommitted yaml (`tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_k8.yaml`), now the packaged `src/zagg/configs/atl03_tdigest_healpix_hive.yaml` (same benchmark-hive lineage; the script overrides `parent_order` and `worker` from its flags, so the k8 chunking difference never reached the shardmap or the cost ceiling). No flag semantics changed.

## Phases

- [x] Land the builder files with the `DEFAULT_CATALOG` path relativized (`74e4a90`)
- [x] `.gitignore` fence so the heavy `data/` neighbors (the ~556k-granule ATL03 catalog clone, generated shardmap/submap JSONs, per-shard count parquets) can't land accidentally (`b8ea368`)
- [x] The #372 fix: `data/conus/build_conus_shardmap.py` line 254 wrote `sm.metadata["total_granules"]` (records *considered*) under `"granules_intersecting_conus"`, contradicting the correct distinct count at `summary.total_granules_intersecting` in the same JSON. Swapped to `sm.metadata["granules_assigned"]` (recorded by `ShardMap.build` at `src/zagg/catalog/shardmap.py:543` since `bacca5c`). Per the issue's "while there", `data/build_aoi_shardmap.py` now reads the same field instead of hand-computing the identical distinct set (`e2e4bf6`)
- [x] Review folds: bare-Feature GeoJSON handling (`bba05a0`), `--worker-memory` validated against dispatchable sizes (`cca3f21`), CONUS builder reads `granules_assigned` once (`d4687c7`), id-less state features + inert noqa (`1c53319`)
- [x] Final trim (espg ruling, 2026-08-05): drop the byte-duplicate `data/california/california.geojson` — the tracked `src/zagg/data/california.geojson` (resolved by `zagg.data.demo_aoi("california")`) is the single home (`2825e35`); default the AOI builder to the packaged hive config (`e94dcd6`)

## What was deliberately excluded from `data/`

Only the files above. The heavy neighbors in the working tree stay out by design — and the phase-2 fence enforces it:

- `data/atl03_v007/` — the 555,867-granule ATL03 v007 catalog clone (external input, fetched from CMR)
- `shardmap_*.json` / `*_submap*.json` — generated shardmaps (multi-hundred-MB at CONUS scale)
- `*.parquet` under `data/` — generated per-shard count tables / catalog snapshots
- the regression tooling around the counts (not in scope here)
- AOI geojsons that already ship in the package: `src/zagg/data/california.geojson` and `src/zagg/data/neon_aop.geojson` are tracked and registered for `demo_aoi()`, so no CA or NEON copy lands under `data/`

`git check-ignore -v` on each landed file comes back clean; all four fence patterns verified against representative heavy paths.

## How it was tested

- Smoke: `uv run python data/build_aoi_shardmap.py --help`, `.../build_conus_shardmap.py --help`, `.../build_conus_polygon.py --help` all run without the catalog present; the new default config verified to `load_config` cleanly (healpix grid, hive layout).
- `uv run ruff check data` and `uv run ruff format --check data` clean on the new files.
- Gates: `uv sync --extra test`; `uv run pytest -q` (3404 passed post-merge); `uv run ruff check src tests` / `ruff format --check src tests` show only the two known pre-existing findings (N818 at `src/zagg/registry.py:64`, format diff in `tests/data/benchmark/README.md`) — untouched by this PR, flagged not fixed per convention. The known env-dependent `test_lambda_build` failure is local-only.
- The #372 fix itself is exercised only on a real catalog run (the scripts are offline tools, not part of the package/suite); the swap reads the field added in `bacca5c`, whose semantics are covered by the existing `granules_assigned` tests.

## Questions for review

- Standing review threads left by design (see inline replies): the stats-JSON fence asymmetry (`*_stats.json` not fenced while the paired parquet is), the `granules_intersecting_conus` key name (kept for continuity with committed artifacts), and the unpinned `master` source URL in `build_conus_polygon.py` (pinning is a provenance-policy call).


data/california/california.geojson
```diff
@@ -0,0 +1 @@
..., [-120.237181, 33.991713], [-120.248517, 34.001127], [-120.236565, 34.008803], [-120.217616, 34.010541], [-120.201055, 34.004696], [-120.18582, 34.008918], [-120.170963, 34.007414], [-120.158486, 34.015952], [-120.145481, 34.024497], [-120.136098, 34.024726], [-120.096665, 34.017861], [-120.072478, 34.024895], [-120.057875, 34.037785], [-120.044139, 34.037273]]], [[[-120.344582, 34.046492], [-120.335766, 34.048703], [-120.330301, 34.047175], [-120.318326, 34.040055], [-120.311306, 34.034477], [-120.306557, 34.026397], [-120.308301, 34.0182], [-120.329612, 34.018629], [-120.351232, 34.019022], [-120.360439, 34.014757], [-120.373825, 34.017764], [-120.382382, 34.024129], [-120.392774, 34.025966], [-120.406226, 34.031415], [-120.415194, 34.032469], [-120.42377, 34.024946], [-120.438197, 34.02956], [-120.446088, 34.030217], [-120.449203, 34.035866], [-120.43797, 34.038538], [-120.415274, 34.054496], [-120.403406, 34.051457], [-120.394043, 34.053285], [-120.380009, 34.060898], [-120.368979, 34.071311], [-120.369068, 34.076225], [-120.362213, 34.071682], [-120.358966, 34.064972], [-120.356262, 34.058511], [-120.36069, 34.053098], [-120.356018, 34.047081], [-120.344582, 34.046492]]]]}}]}
```
espg: 🤖 *from Claude (review)*

**Byte-identical duplicate of a file already tracked in the package.** `src/zagg/data/california.geojson` is tracked at the same 48,641 bytes and the same md5 (`dcac63ac9ce4a8b95ee85c7ec9a45729`); a `json.load` comparison of the two is `True`, down to the identical `properties` block (Natural Earth 1:10m provenance).

That copy is the one with consumers: `zagg.data.demo_aoi("california")` (`src/zagg/data/__init__.py:22`, issue #328) resolves it from an install so the notebooks run from any working directory, and `demo/01_query.ipynb` / `demo/04_california_o8.ipynb` use exactly that accessor. Nothing in the tree references `data/california/`. Landing a second copy gives one polygon two homes and no rule for which wins when one is regenerated.

This also answers the PR body's second open question: the NEON AOI is *already* landed — `src/zagg/data/neon_aop.geojson` is tracked and registered in `_AOIS`, so neither `data/neon/` nor the root `AOP_NEON.geojson` needs a home.

Suggest dropping this file. If CONUS wants the same treatment (it is the one AOI here with no packaged copy), `src/zagg/data/conus.geojson` + an `_AOIS` entry is the established slot; otherwise `data/conus/conus.geojson` staying next to its builder is fine, but then CA should not be the exception.

.gitignore
```diff
@@ -70,3 +70,11 @@ out.zarr/
...
 demo/wheels/
 demo/atl03_v007_full.parquet
+
+# data/ holds builder scripts + small AOI geojsons only — the heavy neighbors
+# (the local ATL03 catalog clone, generated shardmaps, per-shard count tables)
+# are generated outputs / external inputs and must not land in git.
+data/atl03_v007/
+data/**/shardmap_*.json
+data/**/*_submap*.json
+data/**/*.parquet
```
espg, line 80: 🤖 *from Claude (review)*

**This pattern fences out the CONUS builder's own default output.** Verified:

```
$ git check-ignore -v data/conus/conus_shard_granule_counts.parquet
.gitignore:80:data/**/*.parquet	data/conus/conus_shard_granule_counts.parquet
```

That is the default of `--out-parquet` (`data/conus/build_conus_shardmap.py:179`), and the script's own docstring calls it load-bearing:

> the load-bearing artifacts are the per-shard granule-count table (``conus_shard_granule_counts.parquet``) and the summary/distribution JSON (``conus_shard_stats.json``), which the estimate doc and the (blocked) regression consume

Meanwhile the paired `conus_shard_stats.json` is **not** fenced. So of the two artifacts the script exists to produce, one can be committed and its twin silently cannot — and the docstring now describes something the repo forbids.

Pick one and make it consistent: (a) narrow the fence to what is actually heavy (the catalog clone is already covered by `data/atl03_v007/` on line 77, so this line is aimed at count tables and catalog snapshots — an explicit `data/**/*_shard_granule_counts.parquet` or a `data/**/catalog_*.parquet` says so), (b) move the default `--out-parquet` outside `data/` and update the docstring, or (c) keep the fence and reword the docstring so it does not claim a repo-resident artifact.

One mechanical note while here: `data/**/*.parquet` also matches files directly under `data/` (git's `/**/` matches zero directories) — `git check-ignore` confirms `data/foo.parquet` is caught — so the pattern is broader than the "heavy neighbors" comment above it suggests, and would also swallow any small tracked fixture parquet a future builder wants to land.

data/build_aoi_shardmap.py
```diff
@@ -0,0 +1,320 @@
...
+
+import argparse
+import json
+import time
+from pathlib import Path
+
+import numpy as np
+
+DEFAULT_CATALOG = "data/atl03_v007/atl03_v007_full.parquet"
+DEFAULT_CONFIG = "tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_k8.yaml"
```
espg: 🤖 *from Claude (review)*

**The default `--config` names a file that is not in the repository**, so the invocation this module's own docstring documents (line 20: `--polygon <aoi.geojson> --order 9`) fails at `load_config` in a fresh clone. Tracked o9 configs are:

```
$ git ls-files tests/data/benchmark/configs/ | grep o9
tests/data/benchmark/configs/atl03_tdigest_healpix_o9.yaml
tests/data/benchmark/configs/atl03_tdigest_healpix_o9_aoimask.yaml
tests/data/benchmark/configs/atl03_tdigest_healpix_o9_cached.yaml
tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive.yaml
tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_located.yaml
tests/data/benchmark/configs/atl03_tdigest_healpix_o9_hive_strata.yaml
```

No `_hive_k8`. The PR body raises this under "Questions for review" and leaves it "to avoid changing default semantics" — but the current default has no semantics to preserve outside one machine's working tree, and `--config` is one flag away for anyone who wants the k8 variant. Concrete answer: point `DEFAULT_CONFIG` at `atl03_tdigest_healpix_o9_hive.yaml` (nearest tracked sibling, and the config only supplies the worker/window/store shape the cost preview prices — `--order` overrides `parent_order` at line 226 regardless), or land the k8 config in the same PR.

Same check on the CONUS builder passes: its `DEFAULT_CONFIG` (`atl03_tdigest_healpix_o9.yaml`) is tracked, and the `DEFAULT_CATALOG` relativization noted in the PR body is right — both scripts now agree on `data/atl03_v007/atl03_v007_full.parquet`.

data/build_aoi_shardmap.py
```diff
@@ -0,0 +1,320 @@
...
+            }
+        ),
+        out_dir / f"{stem}_shard_granule_counts.parquet",
+    )
+
+    # The cost ceiling exactly as the run logs it before fan-out (issue #298):
+    # unit accounting via the dispatch path, priced at the configured worker.
+    preview = max_cost_preview(config, catalog=str(shardmap_path))
+
+    props = json.loads(Path(args.polygon).read_text())["features"][0].get("properties", {})
```
espg: 🤖 *from Claude (review)*

**`["features"][0]` contradicts the bare-Feature support `_polygon_parts` advertises.** Line 48 explicitly handles the non-FeatureCollection case:

```python
feats = fc["features"] if fc.get("type") == "FeatureCollection" else [fc]
```

but this line indexes `["features"]` unconditionally. Pass a bare `Feature` (or a raw geometry) as `--polygon` and the run gets all the way through the catalog load, the prefilter, `ShardMap.build`, the leak check, the shardmap JSON write, the parquet write and the cost preview — then dies with `KeyError: 'features'` at stats time. That is minutes-to-hours of CONUS-scale work discarded on a line that only reads decoration (`provenance`, `area_km2*`).

It is also the file's third read of the same GeoJSON (`_polygon_parts` at line 47 already parsed it). Cleanest fix is to have `_polygon_parts` return the properties dict alongside `parts, bbox` — one parse, one branch, no way for the two to disagree. Failing that, mirror line 48's branch here and default to `{}`.

data/build_aoi_shardmap.py
```diff
@@ -0,0 +1,320 @@
...
+    ap = argparse.ArgumentParser(description=__doc__)
+    ap.add_argument("--polygon", required=True, help="AOI GeoJSON (FeatureCollection)")
+    ap.add_argument("--name", required=True, help="Region name (stats + output prefix)")
+    ap.add_argument("--out-dir", required=True, help="Directory for the emitted artifacts")
+    ap.add_argument("--catalog", default=DEFAULT_CATALOG)
+    ap.add_argument("--config", default=DEFAULT_CONFIG)
+    ap.add_argument("--start", default="2018-10-13")
+    ap.add_argument("--end", default="2026-03-15")
+    ap.add_argument("--order", type=int, default=9, help="HEALPix parent_order (shard order)")
+    ap.add_argument("--worker-memory", type=int, default=2048, help="Worker memory (MB)")
```
espg: 🤖 *from Claude (review)*

**`--worker-memory` accepts any int and skips the config-time worker validation**, because `config.worker` is assigned at line 227 *after* `load_config`, and `_validate_worker` only runs inside the loader (`src/zagg/config.py:264`, `690-720`):

```python
if isinstance(memory, bool) or memory not in WORKER_MEMORIES:
    raise ValueError(f"worker.memory must be one of {sorted(WORKER_MEMORIES)} MB ...")
```

`WORKER_MEMORIES` is `{2048, 4096, 8192}` — the pre-provisioned variant sizes, and `config.py:98-100` spells out why an unlisted size matters: the runner resolves `worker:` to a `<base>-<memory>[-disk]` function name, so an unlisted size dispatches to a function that does not exist. Downstream, `_worker_memory_gb` (`src/zagg/runner.py:140-143`) just does `worker["memory"] / 1024.0` with no guard.

Net effect: `--worker-memory 3000` produces a confident, wrong `$X.XX` ceiling, writes it into `max_cost` and echoes the impossible selector under `max_cost.worker` in the committed stats JSON — with the failure only surfacing much later at dispatch. Since the whole point of the `max_cost` block is that it "equals the one the run logs before fan-out" (`max_cost_preview` docstring), it should not be reachable for a size no run can use.

One-liner: `choices=sorted(WORKER_MEMORIES)` on this argument (importing the frozenset from `zagg.config`), or call `zagg.config._validate_worker(config)` right after line 227.

data/conus/build_conus_shardmap.py
```diff
@@ -0,0 +1,281 @@
...
+            "total_shards": sm.metadata["total_shards"],
+            "total_granules_intersecting": None,  # set below from distinct pairs' granules
+            "total_pairs": sm.metadata["total_pairs"],
+            "build_wall_s": round(wall, 1),
+        },
+        "granules_per_shard": dist,
+        "leak_check": leak,
+    }
+    # distinct granules that land in >=1 CONUS shard
+    distinct = {g["id"] for gl in sm.granules for g in gl}
```
espg: 🤖 *from Claude (review)*

**The issue's "while there" dedup was applied to the AOI script but not to this one**, which is the script issue #372 is actually about. After `e2e4bf6` the file derives the same number two different ways: line 254 reads `sm.metadata["granules_assigned"]`, while `summary.total_granules_intersecting` is still filled from this hand-built set (plus the `None` placeholder at line 258 and the `len(distinct)` in the SUMMARY print at line 272).

They are provably equal today — `shardmap.py:543` is literally `len({g["id"] for shard in granules for g in shard})` — so nothing is wrong in the output. But keeping two independent derivations of one field inside one JSON is exactly the drift the issue set out to remove, and it materializes the full CONUS granule-id set (hundreds of thousands of strings) for a count the metadata already carries.

The same one-liner the AOI script got applies here:

```python
n_intersecting = int(sm.metadata["granules_assigned"])
```

used at 254, at 258 (dropping the placeholder and the "set below" comment), and in the print at 272 — which deletes lines 265-267 outright.

data/conus/build_conus_shardmap.py
```diff
@@ -0,0 +1,281 @@
...
+            "child_order": int(grid.child_order),
+            "mortie_moc_order": sm.metadata.get("mortie_order"),
+            "backend": sm.metadata.get("backend"),
+            "footprint": sm.metadata.get("footprint"),
+        },
+        "catalog": {
+            "path": args.catalog,
+            "total_granules": n_total,
+            "granules_after_prefilter": n_kept,
+            "granules_intersecting_conus": sm.metadata["granules_assigned"],
```
espg: 🤖 *from Claude (review)*

**The key name still overstates what the value measures** — worth a decision, since mislabeling is what issue #372 is about.

`granules_assigned` is the distinct granules assigned to at least one shard **of the covered shard set**, and that set is whole HEALPix cells with no polygon clip: a granule that misses CONUS but clips an edge cell whose centre is inside CONUS is counted (the AOI-overhang behaviour of issue #101). The file's own comment at line 265 states it precisely — "distinct granules that land in >=1 CONUS shard" — while the emitted key claims polygon intersection, and the same applies to `granules_intersecting_aoi` / `total_granules_intersecting` in `data/build_aoi_shardmap.py:285,289`.

The swap does not change the value (it was already this set, just computed by hand), so this is not a regression — but if the stats JSON is going to be the reference the estimate doc quotes, `granules_assigned` / `total_granules_assigned` as the key names, or a one-line note in the JSON, would stop the next reader from making the mirror-image of the misread this issue fixed.

data/conus/build_conus_polygon.py
```diff
@@ -0,0 +1,109 @@
...
+import urllib.request
+from pathlib import Path
+
+import pyproj
+import shapely
+from shapely.geometry import mapping, shape
+from shapely.ops import transform, unary_union
+
+SOURCE_URL = (
+    "https://raw.githubusercontent.com/PublicaMundi/MappingAPI/master/data/geojson/us-states.json"
```
espg, line 37: 🤖 *from Claude (review)*

**The regeneration source is an unpinned `master` URL, so "regenerates the CONUS AOI geojson from its documented public source" is not reproducible.** `raw.githubusercontent.com/.../master/data/geojson/us-states.json` returns whatever that branch holds today; if the upstream outline is ever edited, re-running this script silently produces a different `conus.geojson` while the `provenance` string, `source` URL and `issue: 202` stamp in the feature properties stay byte-identical — the artifact would carry no evidence it changed. Since the committed geojson is the polygon reference the whole CONUS cost estimate is built over, that is the one thing the provenance block needs to nail down.

Pin the URL to a commit sha (`raw.githubusercontent.com/PublicaMundi/MappingAPI/<sha>/data/geojson/us-states.json`) and/or record the source's sha256 in `properties` next to `n_source_states`, so a re-run either reproduces the committed file or fails loudly.

Two smaller notes on the same function, neither blocking:
- `f["id"]` at line 53 assumes every feature carries a top-level `id`; with `--states-json` pointed at any other states file that is a bare `KeyError` rather than a message about the expected schema. `f.get("id")` plus a count check against the expected 49 kept would be self-diagnosing (the committed file records `n_source_states: 49`, which does confirm the FIPS filter worked as documented — 48 states + DC, with AK/HI/PR dropped).
- The `# noqa: S310` at line 47 suppresses a `flake8-bandit` rule this repo does not enable — ruff's `select` is `[E, F, W, I, N]` (`pyproject.toml`) — so it is inert today and reads as if a real finding were being silenced (CLAUDE.md §4). Either drop it or keep the parenthetical justification as a plain comment.

espg (commented):
🤖 *from Claude (review)*

Reviewed the three PR commits (`74e4a90`, `b8ea368`, `e2e4bf6`) at `b3b19df`; the `main` merge is not review material.

**The issue #372 swap itself checks out.** `ShardMap.build` computes `granules_assigned` as `len({g["id"] for shard in granules for g in shard})` (`src/zagg/catalog/shardmap.py:543`) — character-for-character the set both scripts hand-built, so the emitted value is unchanged and the CONUS stats JSON no longer contradicts itself. I chased the paths that could make the field stale relative to `granules`: `reproject` recomputes it (`shardmap.py:739`), the sweep leaf writer rewrites it per leaf (`src/zagg/sweep.py:332`), and `merge` deliberately pins `granules_assigned == total_granules` on a folded map (`sweep.py:454-457`). None leaves it stale, so the swap is exact on every path these scripts can reach.

Both geojsons are structurally sound: rings closed, no interior rings, CONUS 5 parts / 836 vertices / bbox `[-124.706553, 25.120779, -66.979601, 49.383625]`, CA 8 parts / 1855 vertices — each matching the `n_parts` / `bbox_lonlat` recorded in its own `properties`.

Eight inline findings; the top three are worth settling before this lands:

1. `data/california/california.geojson` duplicates an already-tracked packaged file byte-for-byte (and moots the PR's open NEON question).
2. The new `data/**/*.parquet` fence ignores the CONUS builder's own default output — the file its docstring calls the load-bearing artifact.
3. The AOI builder's default `--config` names a file that is not in the repository.

Smaller residue, not worth its own thread:

- `.github/scripts/run_full_aoi_benchmark.py:223` points at `data/conus/run_conus_regression.py`, which this PR does not land. Now that `data/conus/` exists in git, that reads as a missing file rather than an untracked one — land it or reword the reference.
- `_polygon_parts` / `_conus_parts` keep only `p.exterior`, so interior rings are silently filled. No effect today (both landed AOIs have zero holes, verified), but the AOI builder advertises "any AOI GeoJSON".
- The bbox prefilter and the leak check do planar longitude math, so an AOI crossing ±180 gets a whole-globe longitude box. Still a superset (nothing is dropped), but the "cheap prefilter" property is lost — worth a docstring caveat.
- `_distribution` and `_leak_check` reduce unguarded (`arr.max()`, `lats.min()`): an AOI with no intersecting granules dies on an empty-array reduction instead of reporting zero shards.
- `_leak_check`'s `margin_deg=1.0` is called "a one-cell margin" in both docstrings; at o9 a cell is ~0.1°, so the guard is ~10 cells loose (and below one cell at o5). Either scale it off `--order` or drop the "one-cell" wording.

Review only — no edits, no thread resolutions, no label changes.

data/california/california.geojson
espg: 🤖 *from Claude*

Left as-is, by design — not mine to decide. The five files in this PR (including `data/california/california.geojson`) were explicitly staged for landing by the repo owner, so where the CA polygon lives is an owner call, not a fold call. The duplication analysis stands unrebutted: `src/zagg/data/california.geojson` is byte-identical, it is the copy `zagg.data.demo_aoi("california")` resolves (`src/zagg/data/__init__.py`), and nothing references `data/california/`. This finding stays standing for review.

Same for the NEON observation — `src/zagg/data/neon_aop.geojson` is indeed already tracked and registered, so that half of the PR-body question is answered here rather than in the diff.

.gitignore
espg: 🤖 *from Claude*

Left as-is, by design. The four `data/**` fence patterns were specified verbatim by the repo owner, and excluding the generated outputs — shardmap JSONs and the per-shard count parquets, the CONUS builder's `--out-parquet` default among them — is the intended behaviour, not an accident of the glob. Generated artifacts stay out of git (CLAUDE.md §3); the builder is the reproducible thing, its outputs are not committed. Narrowing the pattern or moving the default output path would both be changes to that ruling, so neither is folded here.

The mechanical note about `data/**/*.parquet` matching files directly under `data/` is correct and left standing.

One asymmetry worth an explicit call, noted rather than acted on: `conus_shard_stats.json` is *not* fenced while its paired parquet is, so of the two artifacts the docstring calls load-bearing only one is committable. Left for the owner to resolve — either fence the stats JSON too or reword the docstring; no new fence patterns added here.

data/build_aoi_shardmap.py
espg: 🤖 *from Claude*

Left as-is — default-flag semantics are out of scope for a fold. The finding is accurate (no `atl03_tdigest_healpix_o9_hive_k8.yaml` is tracked, so a fresh clone fails at `load_config` on the documented invocation) and it is already the standing open question in the PR body under "Questions for review". Changing `DEFAULT_CONFIG` — or landing the k8 config — is a scope/semantics decision for the owner, so it stays there rather than being decided by the review-fold pass.

Confirmation of the other half is noted: the CONUS builder's `DEFAULT_CONFIG` is tracked and both scripts now agree on the relativized `data/atl03_v007/atl03_v007_full.parquet` catalog default.

data/build_aoi_shardmap.py
espg: 🤖 *from Claude*

Folded in `bba05a0` — took the cleaner of the two options you offered: `_polygon_parts` now returns `(parts, bbox, props)`, so the file is parsed once and the FeatureCollection / bare-Feature / raw-geometry branch exists in exactly one place.

```python
feats = fc["features"] if fc.get("type") == "FeatureCollection" else [fc]
geom = unary_union([shape(f.get("geometry", f)).buffer(0) for f in feats])
...
props = dict(feats[0].get("properties") or {}) if feats else {}
```

The `f.get("geometry", f)` also makes the raw-geometry case work in the union step (it previously would have hit `KeyError: "geometry"` there too), and the late `json.loads(...)["features"][0]` at stats time is gone.

Verified against `src/zagg/data/california.geojson` in all three shapes — FeatureCollection, bare Feature and bare geometry all yield 8 parts and bbox `[-124.409, 32.532, -114.119, 42.0]`; the geometry case yields empty properties, which the stats block already tolerates via `props.get(...)`.

data/build_aoi_shardmap.py
espg: 🤖 *from Claude*

Folded in `cca3f21`. Used the public frozenset rather than `choices=` so the zagg import stays deferred past `parse_args` (importing `zagg.config` at parser-build time would pull the whole chain in just to print `--help`):

```python
from zagg.config import WORKER_MEMORIES, load_config
...
if args.worker_memory not in WORKER_MEMORIES:
    ap.error(
        f"--worker-memory must be one of {sorted(WORKER_MEMORIES)} MB "
        f"(the pre-provisioned variant sizes, issue #235)"
    )
```

It sits immediately after the import block and before the catalog load, so the failure is instant rather than after the build. `--worker-memory 3000` now exits 2 with `error: --worker-memory must be one of [2048, 4096, 8192] MB (the pre-provisioned variant sizes, issue #235)`, and the `--worker-memory` help text names the three sizes. Left `_validate_worker` alone — it is private to the loader, and this keeps the check where the flag is.

data/conus/build_conus_shardmap.py
espg: 🤖 *from Claude*

Folded in `d4687c7`, exactly as prescribed. `n_intersecting = int(sm.metadata["granules_assigned"])` is computed once above the `stats` dict and used in all three places — the `catalog.granules_intersecting_conus` value, the `summary.total_granules_intersecting` value (placeholder and "set below" comment gone), and the SUMMARY print. The hand-built `distinct = {g["id"] for gl in sm.granules for g in gl}` and its follow-up assignment are deleted.

Values are unchanged: `shardmap.py:543` computes `len({g["id"] for shard in granules for g in shard})`, which is the same set the deleted comprehension built. The comment recording that equivalence now sits on the single derivation.

data/conus/build_conus_shardmap.py
espg: 🤖 *from Claude*

Left as-is, by design. `granules_intersecting_conus` / `granules_intersecting_aoi` / `total_granules_intersecting` are kept for continuity with the stats artifacts already committed and quoted elsewhere; renaming emitted JSON keys is outside issue #372, whose scope here was the value swap (hand-built set → `sm.metadata["granules_assigned"]`), which is value-identical. Renaming would silently break any consumer keyed on the current names, so it is an owner call, not a fold.

Your semantics point is correct and unrebutted — the set is "assigned to >=1 covered shard" under whole-cell coverage with no polygon clip (the AOI overhang of issue #101), not strict polygon intersection. The adjacent in-code comment states that precisely, and it now sits on the single derivation after the `d4687c7` dedup. Standing for the owner if the key names should follow.

data/conus/build_conus_polygon.py
espg: 🤖 *from Claude*

Split three ways; the two smaller notes are folded in `1c53319`, the pin is not.

**(a) Unpinned `master` URL — left.** The reasoning is sound (a re-run could produce a different polygon under a byte-identical provenance block), but pinning the source, or adding a source sha256 to `properties`, changes what the committed artifact's provenance asserts and would mean regenerating `conus.geojson` itself. That is a provenance-policy call for the owner, not a fold. Standing.

**(b) `f["id"]` KeyError — folded.** Now `f.get("id") not in NON_CONUS_FIPS`, with a comment noting that an id-less `--states-json` keeps every feature and that `n_source_states` in the output properties is what evidences the filter. Smoke-tested on a synthetic three-feature states dict with one id-less feature: FIPS `02` dropped, id-less kept, no `KeyError`. Did not add a hard count check against 49 — that would reject legitimately different `--states-json` inputs, which is the flag's whole point.

**(c) Inert `# noqa: S310` — folded.** Confirmed inert: ruff's `lint.select` in `pyproject.toml` is `["E", "F", "W", "I", "N"]`, no `S`. Dropped the suppression and kept the justification as a plain comment (`# SOURCE_URL is a fixed https raw-content URL, not caller-supplied.`), per CLAUDE.md §4 on not carrying lint suppressions that silence nothing.

data/california/california.geojson
espg: 🤖 *from Claude*

Folded after all in `2825e35` (espg ruled: trim, then merge): `data/california/california.geojson` is dropped from the PR — the tracked `src/zagg/data/california.geojson` (the copy `demo_aoi("california")` resolves) is the single home. Same reasoning resolves the PR body's NEON question: `src/zagg/data/neon_aop.geojson` is already tracked, so no NEON file lands here either. PR body updated.

data/build_aoi_shardmap.py
espg: 🤖 *from Claude*

Folded in `e94dcd6` (espg ruled the trim): `DEFAULT_CONFIG` now points at the packaged `src/zagg/configs/atl03_tdigest_healpix_hive.yaml` — tracked, same benchmark-hive lineage as the old `o9_hive_k8` yaml (its header says grid/read/aggregation match the benchmark hive arm), and the script overrides `parent_order` (`--order`) and `worker` (`--worker-memory`/`--extra-disk`) anyway, so the k8 chunking difference does not reach the shardmap or the cost ceiling. Verified `load_config` accepts it (healpix grid, hive layout) and `--help` still runs without zagg imports.

espg:
🤖 *from Claude*

Trim complete per the 2026-08-05 ruling: duplicate CA geojson dropped (`2825e35`, packaged `src/zagg/data/california.geojson` is the single home — same reasoning resolves the NEON question, no NEON file lands) and the AOI builder now defaults to the packaged `src/zagg/configs/atl03_tdigest_healpix_hive.yaml` (`e94dcd6`). CI green on `e94dcd6`; the PR is ready for merge.

Issue #402, opened 2026-08-07
Tracker: PR #397 follow-ups — the ten review questions, rulings, and landing sites


espg:
🤖 *from Claude* (filed at espg's direction, 2026-08-07)

Tracking checklist for the ten "Questions for review" from PR #397 (issue #388, skip-if-current + lifecycle touch). Each item carries its ruling, its home, and its state — tick as things land. Rulings recorded at [the leans comment](https://github.com/englacial/zagg/pull/397#issuecomment-5223147293) and the (5)/(6)/(9) [ratification](https://github.com/englacial/zagg/pull/397#issuecomment-5223154774) unless linked otherwise.

## Ruled and landing in PR #397 itself (pre-merge trio, in flight)

- [ ] **(5) `column` → `leaf_column` record/parquet key rename** — ruled (c). Lands as a pre-merge commit on PR #397; tick when pushed + review-folded.
- [ ] **(6) `granule_ids` off the transport → sibling object beside the D20 sidecar** — ruled (c): identity equality stays a `granules_sha256` compare; the id list is read only to name a diff on mismatch. Lands in the same trio.
- [ ] **(9) refusal manifest at the store root when `cells_refused > 0`** — ruled (c)+(a): durable artifact for refusals (diff sourced from (6)'s siblings); pure-skip runs stay row-less. Lands in the same trio.

## Ruled, resolved by documentation — no further work

- [x] **(4) `unrecorded-ids` posture** — ruled (a), count-and-document: a pre-#388 leaf's first rerun rewrites once (counted `cells_unrecorded`), self-healing thereafter. Documented in the `docs/hive_layout.md` operator section; nothing further owed.

## Ruled, follow-up homes exist

- [ ] **(1a) handler plumbing** — six-key raster body fix + sidecar-clobber gate: **#401** (espg-named for `deployment/aws/lambda_handler.py`). Plumbing only.
- [ ] **(1b) fleet gate ARMING** — sequenced strictly after the D19 epoch (below); the arming kwargs are enumerated in #401's out-of-scope note. Tick when armed post-epoch.
- [ ] **(7) `granule_workers` → `DATA_SOURCE_PACKAGING_KEYS`** — ruled (b). Rides the **D19 hash-epoch companion PR** on #384's release train ([plan delta 2](https://github.com/englacial/zagg/issues/384#issuecomment-5223266761)); epoch issue to be filed when #384 is labeled.
[...]

Issue #401, opened 2026-08-07
lambda_handler.py: six-key raster body fix + sidecar-clobber gate (skip-gate plumbing, NOT arming) — deployment/aws/lambda_handler.py


espg:
🤖 *from Claude* (filed at espg's direction, 2026-08-07 — this issue explicitly names `deployment/aws/lambda_handler.py` as its scope, per CLAUDE.md §1's requirement for deploy-tree edits)

## Scope: skip-gate PLUMBING in `deployment/aws/lambda_handler.py` — deliberately NOT arming

PR #397 lands the worker-side skip-if-current seams with default-off kwargs, so the deployed handler's behavior is byte-identical until it opts in. Two plumbing changes are prerequisites for the fleet ever using the feature, and are safe/behavior-preserving to land now (espg ruling: plumbing now, arming only after the D19 hash epoch — see #384's companion-epoch delta):

1. **Six-key raster body fix.** The raster branch rebuilds `body` from a closed key list (`lambda_handler.py:1355-1370`) and drops the seam's stamps. Add: `semantic_hash`, `identity`, `current`, `refused`, `touched_objects`, `touch_failed`. Without the first four, raster fleet sidecars record `null` hashes forever — and once any gate is armed (fleet or a local re-run over a fleet-built raster store), a `null` record can never match, so every raster leaf rewrites forever (`semantic-mismatch`) and `cells_current` stays 0. Without the last two, touch telemetry reads as a touch that never ran. The vector branch needs **nothing** — it already passes the seam's dict through (`lambda_handler.py:1842-1850`). Full recipe with line refs: PR #397 question (1) and the seam note at https://github.com/englacial/zagg/pull/397#discussion_r3738950374.

2. **Sidecar-clobber gate.** In both leaf branches, gate the sidecar/sub-map writes on `not metadata.get("current")` — so a future armed skip cannot overwrite a good sidecar with an `n_obs=0` record. Inert while the seams default off; the reason the seams default off today.

**Explicitly out of scope:** the arming kwargs (`skip_if_current=True`, `allow_contraction=bool(event.get("allow_contraction"))`). Arming waits on the D19 hash epoch ((7)(b)+(8)(c), tracked on #384) so the fallback hash is trustworthy before any fleet skip decision consumes it — a wrong fleet skip is worse than a redundant fleet rewrite.

[...]

PR #397, opened 2026-08-05
leaf skip-if-current: input-identity no-op with lifecycle touch (issue #388)

espg:
Closes #388. Refs #381 (point (11)). ~~Blocked by #391~~ **Unblocked 2026-08-07**: #391 merged into `main` (ad8aa30); base retargeted to `main` and `origin/main` merged into the branch (stacking order #389 → #391 → this PR held).

## What this does

Leaf-level **skip-if-current**: a re-dispatched unit whose planned inputs and semantics match what its leaf already records **no-ops the fold**, reports `current` in the run stats, and **touches** every object under the leaf&#39;s prefix footprint (S3 self-copy, `MetadataDirective=REPLACE`) so bucket lifecycle purges see actively-used products. Identity is the ruled PAIR — the run&#39;s `semantic_hash` (D19, what/how) × the unit&#39;s planned granule-id set (over what) — compared per (shard, window) unit against the leaf&#39;s D20 sidecar; whole-set under `all`/ad-hoc. Both leaf families ship it ([espg ruling on the issue](https://github.com/englacial/zagg/issues/388)): the vector seam (`zagg.hive.process_and_write_hive`) and the raster seam (`zagg.processing.raster.process_and_write_raster_hive`), shared by both backends.

- **Fast path**: one `granules_sha256` compare (+ `semantic_hash` equality). The id-set difference runs only on hash mismatch.
- **Contraction guard** (RULED — option (2), refuse-without-flag; predicate `recorded ∖ planned ≠ ∅`, so the mixed add-and-drop signature trips it): the unit **refuses**, names the missing granule ids, and counts in its own run-stats counter — never folded into failures — unless `allow_contraction` rides the run (an `agg()` kwarg + a worker event field; deliberately NOT a config knob, which would pre-approve contraction store-wide against the ruling&#39;s intent).
- **Expansion / any other mismatch**: wholesale D4 rewrite exactly as today (the #391 column artifact included).
- **Touch**: best-effort, fail-open (a failed touch logs and degrades to today&#39;s behavior, never blocks); covers the unit&#39;s leaf `.zarr` tree, its stats sidecar and D22 sub-map, and the #391 column artifact + its sidecar. Touched-object count rides the run stats. Local stores touch via `os.utime`; S3 via a `boto3` self-copy (already a dependency — obstore&#39;s `copy` has no metadata-directive control, and S3 rejects an identity self-copy without `MetadataDirective=REPLACE`).
[...]

Issue #388, opened 2026-08-04
leaf skip-if-current: input-identity no-op with lifecycle touch (semantic_hash x granule-id set)


espg:
🤖 *from Claude* (child of #381; filed at espg's direction, 2026-08-05)

## Scope

Leaf-level **skip-if-current**: a re-dispatched shard whose planned inputs and semantics match what the leaf already records no-ops the aggregation — and refreshes the leaf's object timestamps so lifecycle purges don't reap actively-used products. Today a re-dispatched shard always re-aggregates and wholesale-rewrites (D4) — correct and deterministic, but wasted compute, and it leaves "still in use" invisible to bucket lifecycle policies.

## Identity is a pair

`semantic_hash` alone is insufficient — it freezes *what/how* (D19), not *over what*:

- **what**: the manifest's `semantic_hash` (already enforced — a rerun into an incompatible manifest refuses);
- **inputs**: the shard's planned **granule-id set**, which only the shardmap/submap carries (leaf sub-maps ride the store per issue #300; the recorded side is the sidecar `granule_ids` identity of the #296/#297/#299 thread).

Skip iff both match. Windowed (D13) stores get this comparison per window, where it is already structurally well-defined; ad-hoc / `all` runs compare the full sets. **Expansion and contraction both fail the comparison** — new orbital cycles rewrite, and a shrunken planned set rewrites too (the shardmap is intent).

## Behavior

Decided **worker-side** (D8-consistent: the worker re-derives its input slice anyway and is the only party that can be trusted on it; a dispatch-side prefilter is a later optimization, and the dispatcher still never writes):

1. planned `(semantic_hash, granule_ids)` == recorded → **no-op the fold**, report `current` in the run stats, and **touch** the leaf;
2. any mismatch → wholesale rewrite exactly as today (column artifact included), generation moves, the sweep ratchet propagates.

**The touch.** S3 has no native timestamp refresh; the mechanism is a self-copy (`CopyObject` onto itself, `MetadataDirective=REPLACE`) — server-side, no data transfer, resets `LastModified`, preserves content (and ETag for non-multipart objects). It must cover **every object under the leaf prefix** (lifecycle rules act per object), so the skip path is LIST + self-copy fan-out — orders cheaper than re-aggregating. Caveats to handle: versioned buckets mint a new version per touch; objects ≥ 5 GB need multipart copy (not expected for leaf objects); the touch is best-effort/fail-open (a failed touch degrades to today's behavior, never blocks the run).

[...]

PR #403, opened 2026-08-08
leaf_column rename, granule-id sibling objects, refusal manifest (issue #388 follow-up trio)

espg:
Refs #388. Refs #402 (items (5)/(6)/(9)). Follows PR #397 (merged as `3f13af2`).

The three schema/artifact changes espg **ruled pre-merge** on the PR #397 thread ([the ratification comment](https://github.com/englacial/zagg/pull/397#issuecomment-5223154774)). PR #397 merged first, so they land here instead — unchanged in content. **Timing still matters**: these were ruled cheap *because the schema is unreleased*, and it still is. The D20 `column` key and the record-borne `granule_ids` exist on `main` only between `3f13af2` and this PR; landing before the next `*.*.*` tag preserves the "nothing published this spelling" premise the rulings rest on.

## Rulings landed

- [x] **(5)(c) — the D20 record/parquet key `column` renames to `leaf_column`** (`e00d7f4`). Every site: the seam metadata stamp (`hive.py`), `telemetry.build_record`'s key set, `_EQ_OR_NONE_KEYS` (both merge arms), `_ROW_SCALARS`/`flatten_record`, the parquet round-trip test, and `tools/generate_spec_fixtures.py`'s read of the seam's metadata. Two reasons, both recorded at the key: `column` is **SQL-reserved** in DuckDB and Trino/Athena — the unquoted `WHERE column IS NOT NULL` is a parse error, so every filter on the run parquet would need `WHERE "column" IS NOT NULL` — and `leaf_column` reads correctly on the pyramid column's own record ("the column this LEAF carries", not "the column this record is"). The spec-fixture *file* key `column` in `tests/data/spec/column.expected.json` is the fixture schema's own name and is deliberately untouched (published; not the D20 record).

- [x] **(6)(c) — `granule_ids` moves off the record/envelope into a sibling object** (`c980446`). The record now carries the catalog **hash** only; the id LIST lives in `granules.json` beside the stats sidecar, on the sidecar's own spec-keyed grammar (`granules_{window}.json`; `{stem}.granules.json` under `morton-hive/3`), derived through one private `_sibling_key` owner so the two names cannot drift.
  - **Identity equality never reads it.** `classify_leaf_identity` takes a `load_recorded_ids` callable and invokes it **only when the recorded and planned `granules_sha256` DISAGREE** — pinned by tests asserting the loader's call count is 0 on the skip path, 0 on the equal-hash/changed-semantic quadrant, and exactly 1 on a hash mismatch. The list exists to *name* the diff a refusal reports. (The narrower gate is review fold `c320eea`; as first written the loader fired whenever the fast path failed, which made a routine config-only rerun pay one sibling GET per shard for a verdict the hashes alone decide.)
  - **The interaction check the ruling asked for holds**: `dedup.shard_status` (one GET per shard across a whole shardmap in `has_run`) and `telemetry.rows_from_status` (one GET per mirrored envelope inside a 900 s worker) read the sidecar/envelope only — neither needs the sibling, and both got *smaller*: the ~4,600-id / ≈550 KB pole payload is off every one of those GETs and off the response envelope entirely.
  - **Self-pairing**: the sibling records the `granules_sha256` of the list it holds, and `dedup.leaf_recorded_ids` accepts it only when that equals the sidecar's. A torn rewrite (either PUT lost, or a stale sibling beside a fresh sidecar) reads as *unrecorded*, not as ids that could name the wrong granules as dropped. Absent sibling on a mismatch → the pre-#388 `unrecorded-ids` path, exactly as ruled for question (4)(a).
  - **Back-compat over the `3f13af2`..this-PR window** (review fold `9d7f65d`): leaves written after PR #397 merged but before this PR carry the id list ON the record with no sibling. `leaf_recorded_ids` falls back to `recorded["granule_ids"]` when the sibling is absent or unusable, so those leaves stay guarded and self-heal on their next rewrite instead of classifying `unrecorded-ids` forever. The sibling wins whenever both exist; the fallback pairs by construction (`build_record` hashed the very ids it embedded).
  - **Written by the leaf SEAMS, not by the dispatchers that write the sidecar** — see "Questions for review" (1) below; this is the one place I did not take the obvious reading of "write sites". The adversarial review confirmed the implementation, the justification and the cost; the numbers are in that question now.
  - Lifecycle: `touch_current_unit`'s footprint gains the sibling (a lifecycle rule reaping it would silently disarm the guard on an otherwise-fresh leaf), with the over-touch pin extended to `granules_2020.json`; the docs' footprint list and the `.github/scripts/bench_objects.py` object model both gain it (see the note under Testing).

- [x] **(9)(c)+(a) — a store-root refusal manifest** (`b462ef2`). A run that ends with `cells_refused > 0` writes ONE small JSON at the store root, `refusals_{ts}_{run_id}.json` — timestamp-first like `stats_*.parquet` and `sweep_stats_*.json`, and outside the `stats_*.parquet` glob the sweep's run-record discovery scans. It carries the run context needed to act (`run_id`, timestamp, the run's `semantic_hash`, zagg version) and one entry per refused unit: `shard_key`, `window`, the classification (`contraction`/`mixed`), `n_missing`, and the **complete** `missing_granules` list — the composition of (6)'s sibling reads. Units sort by (shard, window) so two runs over one refusal set produce comparable objects. **No D20 schema change and no synthesized unit rows** — that was the point of (c) over (b). **Ruled (9)(a)**: a pure-skip run writes nothing here and stays row-less.
  - **D8**: the write rides the same local wrap-up seam as the root `coverage.moc` union and `touch_store_root` — `runner._write_refusals`, with `c71198f`'s rationale mirrored ("the LOCAL dispatcher runs it in-process; this process is also the worker"). The fleet has no once-per-run worker-side seam, so the manifest is **local-backend-only**, exactly like the store-root touch; the docs now say both wait on question (10)'s resolution.
  - The gate's verdict now names the `(shard, window)` unit (`leaf_identity_gate(window=...)`), because on a windowed store the shard alone is ambiguous.
  - Operator surface: the summary gains `refusal_manifest_path` (local summaries only, `None` when nothing refused), the CLI prints it under the refusal remedy line, and the phase-4 docs gain a **The refusal manifest** section. The truncated-log caveat (phase-4 finding 4) now points at the manifest as the durable full list instead of telling operators to triage on the count.
  - Sort caveat, for anyone diffing two manifests: units sort by `str(shard_key)`, i.e. **lexicographically** — `10` precedes `9`. Deterministic and comparable across runs, which is all the ordering is for, but it is not numeric order.

## Review fold

Five findings from the adversarial self-review, one commit each:

- `c320eea` — **the sibling read was not gated on the hash mismatch it is documented to be gated on.** It also fired on the "hashes equal, semantic hash differs" quadrant, whose verdict is provably `semantic-mismatch` regardless of the id list — a routine config-only rerun therefore fanned out one sibling GET per shard (2,721 at CA o9, ~550 KB each on a pole unit). `semantic-mismatch` now short-circuits before any load, pinned at `loads == 0`. Second-order fix: with the sibling absent that quadrant used to degrade to `unrecorded-ids`, so a config-only rerun over a pre-#388 store reported 100% `cells_unrecorded` — the counter that means "the guard was inert" — when the hashes had in fact *proved* no contraction.
- `9d7f65d` — **back-compat read of the record-borne `granule_ids`** for the `3f13af2`..this-PR window (see the (6)(c) bullet above).
- `5a7e3eb` — **`write_refusal_manifest`'s fail-open now covers the unit composition**, not just the PUT. It runs at `_run_local:3212`, *before* the summary and the run-stats parquet, so a `MemoryError` on question (3)'s unbounded case or a `TypeError` on malformed `missing_granules` used to destroy the run record of a run that had already done all its work. `write_granule_ids` already had the whole body inside its guard; these now match.
- `45bb7eb` — **the sibling PUT moved inside the write-phase bracket** on both leaf families. It sat one line after `_write_elapsed +=` (and after `write_s +=` on the raster seam), so `phase_timings["write"]` → `phase_write` under-reported by one PUT per leaf in every run parquet and benchmark record. Pinned by a test that wraps the PUT in a 50 ms sleep and asserts the bracket sees it.
- `7b04101` — **the granule-id sibling got its own `spec: "zagg-granule-ids/1"` marker** instead of borrowing the D20 record's `SCHEMA_VERSION` (see Spec assessment). The PR-body sentence that claimed both new objects already carried their own marker is corrected in the same pass — it was true only of the refusal manifest.

`docs/hive_layout.md` follows the first of these: the "fetched exactly once, only when the hashes disagree" invariant is now true as written, and the pre-#388 paragraph says *granule-hash* mismatch, with an explicit note that a config-only rerun is not one of these and never inflates `cells_unrecorded`.

## Testing

- `uv run pytest -q` on this branch (three ruling commits + five review folds): **3,671 passed / 38 skipped / 1 failed** — the failure is the pre-existing, environment-dependent `test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, flagged not fixed. The other known-flaky one, `test_client_transport.py::TestStatusPoller::test_invoke_fault_burns_an_attempt_and_retries`, failed on one earlier run of the same tree and passed here; it also fails on clean `main`.
- `uv run ruff check src tests`: clean on the diff; the pre-existing `N818` at `src/zagg/registry.py:64` is flagged, not fixed. `uv run ruff format --check src tests`: clean on the diff; the pre-existing `tests/data/benchmark/README.md` reformat is flagged, not fixed.
- New/changed tests, by ruling:
  - **(5)** the parquet round-trip (all-null + mixed, fastparquet *and* pyarrow), the metadata-rides-the-key pin, both merge arms, and the seam's stamp — all renamed in place; `test_column.py`'s 12 seam assertions follow. `metadata["phase_timings"]["column"]` is a phase name, not the record key, and is deliberately unchanged.
  - **(6)** `TestGranuleIdsSibling` (key grammar across all three specs + unknown-spec raise, sibling path, sorted self-hashing round-trip, empty-set case, fail-open write, absent read); `TestLeafRecordedIds` (round-trip, absent, **unpaired-hash rejection**, four malformed bodies, per-window naming); `TestClassifyLeafIdentity` rebuilt around a counting loader (0 calls on the fast path, exactly 1 on mismatch, `None` loader → `unrecorded-ids`); the "pre-#388 leaf" models on both seams now *delete the sibling* rather than nulling a record key, which is what a pre-#388 leaf actually looks like; the lifecycle footprint counts and the S3 exact-key sets gain it; the end-to-end vector rerun asserts the sidecar has no `granule_ids` and the sibling has them.
  - **review fold** `TestClassifyLeafIdentity::test_semantic_mismatch_with_equal_sets_rewrites_not_refuses` (now `loads == 0`) and `::test_semantic_mismatch_without_a_sibling_is_not_unrecorded`; `TestLeafRecordedIds::test_record_borne_ids_are_read_when_no_sibling_exists`, `::test_the_sibling_wins_over_a_record_borne_list`, `::test_the_spec_marker_is_read_leniently`; `TestRefusalManifest::test_composition_is_fail_open_too` (malformed `missing_granules` + a `MemoryError` raiser); `test_hive.py::test_write_phase_covers_the_granule_id_sibling`.
  - **(9)** `TestRefusalManifest` (key shape + traversal guard + not-a-run-record, full per-unit diff with sort order, nothing-refused-writes-nothing, fail-open); end-to-end refusal→manifest on **both** leaf families (`test_hive.py` vector with the exact dropped id, `test_raster_runner.py` raster in the STAC-item id space), the `refusal_manifest_path is None` pin on the `--allow-contraction` rerun, the CLI pointer test, the summary key-set pin, and the windowed unit-identity pin.
- **One file outside `src`/`tests` changed**: `.github/scripts/bench_objects.py`, the benchmark object-count model. The per-leaf model had "TWO node-dir siblings"; with the granule-id list it is three, and without the update the in-tree tests fail (`objects_other == 1`, `total_max` off by one). Flagging it explicitly per CLAUDE.md §1, and **correcting an earlier claim in this body that it is "test-support"** — it is not. Besides `tests/test_benchmark_objects.py`, it is imported by `.github/scripts/run_benchmark.py` and `run_full_aoi_benchmark.py`, executed by `lambda-benchmark.yml` and `lambda-benchmark-fullaoi.yml`, and base-pinned from `main` by `lambda-benchmark-command.yml`. The deploy coupling was checked and is **safe**: the command workflow never redeploys, and the push-to-`main` path rebuilds the layer from the merged tree in the same run, so the object model and the fleet code it models land together. No workflow file is touched.

## Spec assessment

No byte-contract change, consistent with PR #397's own assessment. `docs/specification.md`, `tools/generate_spec_fixtures.py`'s *fixture schema*, and `tests/data/spec/` are unchanged (the fixtures tool's one edit reads the seam's renamed metadata key; the emitted fixture JSON is byte-identical). `SCHEMA_VERSION` stays 1: the D20 record's keys are pre-release and rev in place per the constant's own rule, and this is a rename plus a key removal, both before publication. The two new object families are **operational**, where §4.7's informative scope-out already places the stats/sub-map families: `granules.json` is a sibling of a sidecar the spec does not normatively describe beyond its naming grammar, and the refusal manifest is per-run root telemetry like `stats_*.parquet`/`sweep_stats_*.json`. Both carry a version marker of their own so a future reader has a versioning handle: the manifest `spec: "zagg-refusals/1"` **plus** `schema_version` (it is run-record-class), and the sibling `spec: "zagg-granule-ids/1"` alone (review fold `7b04101` — as first written it borrowed the D20 record's `SCHEMA_VERSION`, welding a schema it deliberately is not). Readers must treat the sibling's marker leniently: the sidecar hash pairing is what decides, and an unknown marker degrades to the ruled `unrecorded-ids` fallback rather than raising.

## Questions for review

1. **The granule-id sibling is written by the leaf SEAMS, not at the dispatcher-side sidecar write sites** — the one place I read the ruling's "write sites" against its most literal sense, deliberately, and the one thing to overrule if you disagree. Reasons: (a) **D8** — a store write belongs worker-side, and the seams already PUT the whole leaf; (b) **fleet coverage with no `deployment/aws/` edit** — PR #397 banked "fleet vector sidecars record the identity half with zero handler changes" via the seam's `semantic_hash` stamp; putting the sibling at the dispatcher sites would have *un*-banked the catalog half, leaving every fleet-written leaf `unrecorded-ids` (guard permanently inert on the fleet) until question (1)'s handler change lands. Via the seam, both leaf families record their input set from this release's deploy onward; (c) **one source for the id space** — the seam writes the very list the gate compares as `planned_ids`, instead of the dispatcher recomputing `_resolve_urls(records, driver)` in lockstep. The cost, stated plainly: the sibling PUT is **unconditional** on a committed leaf, so a caller that never opted into the gate (today's Lambda handler) is no longer byte-identical — it writes one extra small object per leaf. Both seam docstrings say so at the opt-in kwarg.

    **The adversarial review measured this and the placement stands unless you overrule it.** What it confirmed: the PUT is gated on a **committed** leaf (`hive.py:1656` / `processing/raster.py:1412`), so it is one object per **(node, window)**, never per shard-attempt. Object-count delta on `atl03_tdigest_healpix_o9_hive.yaml`: **+9.1% objects/shard** (11 → 12), **+7.7%** on `_located`, **+5.3%** on `_strata`. Request/cost delta at CA o9: **+2,721 PUTs ≈ $0.0136 against $65.30 of compute — 0.02%**. Size: ~110 B/id ⇒ ~22 KB on a typical shard, ~500 KB on a pole shard. Also worth recording: **no test ever pinned the fleet's exact object set**, so PR #397's "byte-identical when unarmed" property was unpinned to begin with — this changes a real behavior, not a guarded one. If you would still rather the fleet stay byte-identical until the handler opts in, this moves to the two `runner.py` sidecar sites in one commit.

    Two facts for the handler recipe (which PR #401 owns), both **pre-existing and out of scope here** under §1 — `deployment/aws/` and the handler are not named by this PR: (a) `lambda_handler.py:1687`/`:1860` pass neither `sidecar_spec` nor `write_sidecar(spec=)`, so the fleet writes **both** the sidecar and the sibling with `spec=None` — on a `morton-hive/3` store both land under the legacy name; (b) on the skip path, an absent sibling costs one extra S3 `head_object` per skipped unit in `touch_current_unit` (**+2,721 HEADs** on an all-skip CA o9 rerun over a pre-#388 store). Neither is fixable in this PR; recording them so the handler change lands with both in hand.
2. **`refusal_manifest_path` is on the two LOCAL summaries only** (always present, `None` when nothing refused), not on the Lambda ones, because there is no fleet writer at all — unlike `run_stats_path`, which both backends produce. That is an asymmetric summary key set; the alternative (always-`None` on the fleet) reads as "no refusals" rather than "not recorded". Left as-is; say the word if you want the key mirrored.
3. **The manifest is unbounded in size by design.** A run where many pole shards refuse could write a large object (~550 KB of ids per refused pole unit). Truncating would defeat the entire reason (9)(c) exists — the docs now tell operators to triage from this object rather than from the truncated log — so it carries the full diff. Worth a size guard, or is "proportional to the refusal set" fine? (Review fold `5a7e3eb` at least made the worst case *degrade*: the unit composition is now inside the same fail-open guard as the PUT, so a `MemoryError` here costs the manifest, not the run record.)
4. **A narrow pre-existing hole in the skip guard, inherited from PR #397 — neither widened nor narrowed by this PR, and deliberately NOT fixed here.** The sidecar PUT happens after the commit stamp, so if a rewrite dies **after the stamp but before the sidecar PUT**, the leaf holds run-2 data under run-1's sidecar. A later run planning run-1's id set then matches both `granules_sha256` and `semantic_hash`, passes `_leaf_is_committed`, and **skips a leaf whose contents do not match what the sidecar says**. Three ways out, all yours to pick: (a) reorder — sidecar before stamp (costs the "sidecar only ever describes a committed leaf" property, and a torn pair the other way then reads as a stale record on a leaf that never landed); (b) carry an id-set digest in the stamp itself, so the commit and the identity claim land in one object and cannot tear apart; (c) accept it as narrow — it needs a crash inside a one-PUT window, and the failure mode is a stale leaf, not corrupt data. My read is (c) is defensible and (b) is the real fix, but this is a D4 change and belongs on its own issue rather than in a follow-up trio.

### DIFF
--- a/src/zagg/dedup.py
+++ b/src/zagg/dedup.py
@@ -183,9 +183,30 @@ def leaf_recorded_ids(leaf_path, recorded, *, spec=None, store_kwargs=None):
     torn rewrite (one of the two PUTs lost, or a stale sibling beside a fresh
     sidecar) reads as no recorded set rather than as ids that could name the
     wrong granules as dropped — or, worse, hide real ones. Fail-open on every
-    other fault (absent, unreadable, malformed): ``None``, which the caller
-    classifies ``unrecorded-ids`` and rewrites.
+    other fault (absent, unreadable, malformed): the RECORD's own
+    ``granule_ids`` if it carries one (the back-compat window below), else
+    ``None``, which the caller classifies ``unrecorded-ids`` and rewrites.
+
+    The sibling is also read leniently on its ``spec`` marker — an absent or
+    unknown one is not a rejection. Only the hash pairing decides, so a future
+    ``zagg-granule-ids/N`` that keeps these two keys stays readable, and an
+    unrecognized one degrades to the ruled ``unrecorded-ids`` fallback rather
+    than to an error.
     """
+    ids = _sibling_ids(leaf_path, recorded, spec, store_kwargs)
+    if ids is not None:
+        return ids
+    # Back-compat (issue #388): leaves written in the window between PR #397's
+    # merge (``3f13af2``) and the commit that moved the list out carry it ON
+    # the record with no sibling at all. Without this they classify
+    # ``unrecorded-ids`` forever; with it those stores self-heal, and the list
+    # pairs by construction (``build_record`` hashed the very ids it embedded).
+    legacy = (recorded or {}).get("granule_ids")
+    return legacy if isinstance(legacy, list) else None
+
+
+def _sibling_ids(leaf_path, recorded, spec, store_kwargs):
+    """The paired granule-id sibling's list, or ``None``; see above."""
     from zagg.telemetry import read_granule_ids
 
     try:
--- a/src/zagg/dedup.py
+++ b/src/zagg/dedup.py
@@ -25,7 +25,10 @@ silently ships wrong data; a wrong "recompute" costs one shard's work.
 :func:`classify_leaf_identity` (issue #388) is the worker-side flavor of the
 same comparison: pure (the caller supplies the recorded sidecar), with the
 id-set reading — equal / expansion / contraction / mixed — that the leaf
-skip-if-current and contraction-guard decisions key on.
+skip-if-current and contraction-guard decisions key on. The recorded id LIST
+it diffs against is NOT in the sidecar: it sits in a sibling object read
+lazily, on hash mismatch only, through :func:`leaf_recorded_ids` — so both
+comparisons above keep their one small GET per shard.
 """
 
 from __future__ import annotations
@@ -166,13 +169,53 @@ def has_run(
 IDENTITY_ACTIONS = ("skip", "rewrite", "refuse")
 
 
-def classify_leaf_identity(recorded, *, semantic_hash, planned_ids) -> dict:
+def leaf_recorded_ids(leaf_path, recorded, *, spec=None, store_kwargs=None):
+    """The leaf's recorded granule-id list, or ``None`` (issue #388).
+
+    Reads the sidecar's ``granules.json`` SIBLING
+    (:func:`zagg.telemetry.read_granule_ids`) — the object the id list moved
+    to so identity EQUALITY stays a ``granules_sha256`` compare against the
+    small sidecar. Call it only on a hash mismatch: this is the one read the
+    split exists to avoid paying per shard.
+
+    ``recorded`` is the sidecar record it must PAIR with: the sibling is
+    accepted only when its own ``granules_sha256`` equals the sidecar's, so a
+    torn rewrite (one of the two PUTs lost, or a stale sibling beside a fresh
+    sidecar) reads as no recorded set rather than as ids that could name the
+    wrong granules as dropped — or, worse, hide real ones. Fail-open on every
+    other fault (absent, unreadable, malformed): ``None``, which the caller
+    classifies ``unrecorded-ids`` and rewrites.
+    """
+    from zagg.telemetry import read_granule_ids
+
+    try:
+        sibling = read_granule_ids(str(leaf_path), spec, **(store_kwargs or {}))
+    except Exception as e:
+        logger.warning(f"granule-id sibling read failed (degrading to rewrite, issue #388): {e}")
+        return None
+    if not isinstance(sibling, dict):
+        return None
+    ids = sibling.get("granule_ids")
+    if not isinstance(ids, list):
+        return None
+    if sibling.get("granules_sha256") != (recorded or {}).get("granules_sha256"):
+        logger.warning(
+            f"granule-id sibling at {leaf_path} does not pair with the stats sidecar "
+            f"(hash mismatch) — treating the recorded id set as unrecorded (issue #388)"
+        )
+        return None
+    return ids
+
+
+def classify_leaf_identity(recorded, *, semantic_hash, planned_ids, load_recorded_ids=None) -> dict:
     """Classify one planned unit against its leaf's recorded identity (#388).
 
     Identity is the ruled PAIR: the run's ``semantic_hash`` (what/how, D19)
     x the unit's planned granule-id set (over what). The recorded side is
-    the leaf's D20 stats sidecar (``semantic_hash``, ``granules_sha256``,
-    and — from issue #388 on — ``granule_ids``).
+    the leaf's D20 stats sidecar (``semantic_hash``, ``granules_sha256``) —
+    and, only when those disagree, the recorded id LIST, which lives in the
+    sidecar's own sibling object and is fetched through
+    ``load_recorded_ids`` (:func:`leaf_recorded_ids`).
 
     Fast path: ``granules_sha256`` equality (one compare — the common rerun
     case). The id-set difference runs ONLY on a mismatch, classifying:
@@ -205,7 +248,8 @@ def classify_leaf_identity(recorded, *, semantic_hash, planned_ids) -> dict:
     ``no-sidecar``/``expansion``: callers must count and surface it apart
     from the ordinary rewrites (the phase-2 run stats, the phase-4 operator
     docs), because the guard cannot protect any leaf written before this
-    release — no fleet sidecar predating issue #388 records ``granule_ids``.
+    release — no leaf written before issue #388 has the granule-id sibling
+    that records its input set.
 
     Parameters
     ----------
@@ -222,6 +266,12 @@ def classify_leaf_identity(recorded, *, semantic_hash, planned_ids) -> dict:
         planned set is UNKNOWN and never refuses (``unknown-planned``); the
         empty list means the unit genuinely plans no inputs, which against a
         non-empty recorded set is a full contraction and does refuse.
+    load_recorded_ids : callable, optional
+        Zero-argument loader for the leaf's RECORDED id list, called at most
+        once and ONLY after the ``granules_sha256`` fast path failed — the
+        laziness is the whole point of keeping the list out of the record
+        (issue #388's ruling). ``None`` (or a loader returning ``None``)
+        means no recorded set is available: ``unrecorded-ids``.
 
     Returns
     -------
@@ -242,13 +292,17 @@ def classify_leaf_identity(recorded, *, semantic_hash, planned_ids) -> dict:
     semantic_match = semantic_hash is not None and recorded.get("semantic_hash") == semantic_hash
     if rec_hash is not None and rec_hash == granules_sha256(planned) and semantic_match:
         return {"action": "skip", "classification": "equal", "missing": []}
-    rec_ids = recorded.get("granule_ids")
+    # The ONE read the fast path exists to avoid: the recorded id list lives
+    # beside the sidecar, not in it (issue #388's ruling on question (6)), and
+    # is fetched only now that the hashes disagree.
+    rec_ids = load_recorded_ids() if load_recorded_ids is not None else None
     if rec_ids is None:
-        # Pre-#388 sidecar (or a cross-leaf rollup, where the field collapses
-        # to None): something mismatched, but there is no recorded set to
-        # diff — undecidable, so today's rewrite. NOT a benign ambiguity: on
-        # a contracted rerun this is the wholesale rewrite the guard exists to
-        # stop, and every leaf written before this release lands here. Its own
+        # No recorded set to diff — a pre-#388 leaf, a lost sibling PUT, or a
+        # sibling that does not pair with the sidecar. Something mismatched,
+        # but the direction is undecidable, so today's rewrite. NOT a benign
+        # ambiguity: on a contracted rerun this is the wholesale rewrite the
+        # guard exists to stop, and every leaf written before this release
+        # lands here. Its own
         # classification for exactly that reason — count it apart from the
         # ordinary rewrites and say so in the operator docs (see the class
         # docstring's one-sided-conservatism note).
@@ -271,4 +325,11 @@ def classify_leaf_identity(recorded, *, semantic_hash, planned_ids) -> dict:
     return {"action": "rewrite", "classification": "expansion", "missing": []}
 
 
-__all__ = ["IDENTITY_ACTIONS", "STATUSES", "classify_leaf_identity", "has_run", "shard_status"]
+__all__ = [
+    "IDENTITY_ACTIONS",
+    "STATUSES",
+    "classify_leaf_identity",
+    "has_run",
+    "leaf_recorded_ids",
+    "shard_status",
+]
--- a/src/zagg/dedup.py
+++ b/src/zagg/dedup.py
@@ -268,7 +268,9 @@ def classify_leaf_identity(recorded, *, semantic_hash, planned_ids, load_recorde
         non-empty recorded set is a full contraction and does refuse.
     load_recorded_ids : callable, optional
         Zero-argument loader for the leaf's RECORDED id list, called at most
-        once and ONLY after the ``granules_sha256`` fast path failed — the
+        once and ONLY when the recorded and planned ``granules_sha256``
+        DISAGREE — never on the skip path, and never on the equal-hash /
+        changed-semantic quadrant, whose verdict the hashes alone decide. The
         laziness is the whole point of keeping the list out of the record
         (issue #388's ruling). ``None`` (or a loader returning ``None``)
         means no recorded set is available: ``unrecorded-ids``.
@@ -290,8 +292,20 @@ def classify_leaf_identity(recorded, *, semantic_hash, planned_ids, load_recorde
     planned = [str(g) for g in planned_ids]
     rec_hash = recorded.get("granules_sha256")
     semantic_match = semantic_hash is not None and recorded.get("semantic_hash") == semantic_hash
-    if rec_hash is not None and rec_hash == granules_sha256(planned) and semantic_match:
+    hashes_match = rec_hash is not None and rec_hash == granules_sha256(planned)
+    if hashes_match and semantic_match:
         return {"action": "skip", "classification": "equal", "missing": []}
+    if hashes_match:
+        # Equal hashes mean an identical sorted id multiset, so the diff is
+        # provably empty and the ONLY reachable verdict is semantic-mismatch:
+        # decide it here rather than paying the sibling GET for a read that
+        # cannot change the answer. This is the ordinary config-only rerun
+        # (same inputs, changed semantic core) — at CA o9 that would be 2,721
+        # needless GETs of up to ~550 KB each, which is exactly the per-shard
+        # cost the id list was split out of the sidecar to avoid. It also
+        # keeps this quadrant OFF ``unrecorded-ids`` when the sibling is
+        # absent: the guard is not inert here, it has proved no contraction.
+        return {"action": "rewrite", "classification": "semantic-mismatch", "missing": []}
     # The ONE read the fast path exists to avoid: the recorded id list lives
     # beside the sidecar, not in it (issue #388's ruling on question (6)), and
     # is fetched only now that the hashes disagree.
--- a/src/zagg/hive.py
+++ b/src/zagg/hive.py
@@ -1231,7 +1231,7 @@ def leaf_identity_gate(
     way, and a contraction over debris is still a contraction the operator
     should be told about.
     """
-    from zagg.dedup import classify_leaf_identity
+    from zagg.dedup import classify_leaf_identity, leaf_recorded_ids
     from zagg.telemetry import read_sidecar
 
     t0 = time.time()
@@ -1254,7 +1254,15 @@ def leaf_identity_gate(
         )
         recorded = None
     identity = classify_leaf_identity(
-        recorded, semantic_hash=semantic_hash, planned_ids=planned_ids
+        recorded,
+        semantic_hash=semantic_hash,
+        planned_ids=planned_ids,
+        # LAZY by construction (issue #388's ruling on question (6)): the
+        # recorded id list is a sibling object, GET only when the
+        # granules_sha256 fast path failed and the diff has to be named.
+        load_recorded_ids=lambda: leaf_recorded_ids(
+            leaf_path, recorded, spec=sidecar_spec, store_kwargs=store_kwargs
+        ),
     )
     if identity["action"] == "skip":
         drift = None
@@ -1367,8 +1375,11 @@ def process_and_write_hive(
     sidecar returns a ``{"current": True}`` metadata dict and writes nothing;
     a contraction (``recorded ∖ planned ≠ ∅``, the ruled predicate) returns
     ``{"refused": True, "missing_granules": [...]}`` unless
-    ``allow_contraction`` rides the call. Default ``False`` keeps this seam
-    byte-identical for callers that have not opted in (the Lambda handler).
+    ``allow_contraction`` rides the call. Default ``False`` keeps the GATE
+    inert for callers that have not opted in (the Lambda handler); the
+    granule-id sibling below is written either way, because it is what a
+    LATER run's guard diffs against — a leaf must record its input set
+    whether or not the run that wrote it had the gate armed.
     ``semantic_hash`` is the RUN config's D19 digest when the caller holds it
     (the local runner); ``None`` falls back to hashing this worker's own
     ``config`` — the same digest except under the per-cell
@@ -1380,7 +1391,11 @@ def process_and_write_hive(
     ``telemetry.build_record``'s validated metadata fallback.
     ``sidecar_spec`` is the manifest spec in effect, keying the sidecar name
     (``telemetry.sidecar_key``); the gate reads with it and degrades to
-    rewrite when the sidecar is absent under that name.
+    rewrite when the sidecar is absent under that name. It also keys the
+    granule-id SIBLING this seam writes after the commit stamp
+    (``telemetry.write_granule_ids``, issue #388) — the recorded id set the
+    contraction guard later diffs, kept out of the sidecar and the response
+    envelope so an identity check stays one small GET.
     """
     from zagg.processing import (
         process_shard,
@@ -1640,6 +1655,15 @@ def process_and_write_hive(
             time_range=time_range,
         )
         _write_elapsed += time.time() - _t0
+        # The recorded granule-id list, as this leaf's own sibling object
+        # (issue #388): AFTER the stamp, so it never certifies a leaf that
+        # did not land, and written here rather than at the caller's sidecar
+        # PUT because ``granule_urls`` is the very list the identity gate
+        # compares — one source for the recorded id space, on both backends.
+        # Fail-open inside (telemetry class, D9).
+        from zagg.telemetry import write_granule_ids
+
+        write_granule_ids(leaf_path, granule_urls, spec=sidecar_spec, **store_kwargs)
     # Write-phase split (issue #249): read/index/aggregate come from
     # ``process_shard``; ``write`` is the leaf write-out above (template +
     # dense chunks + ragged + coverage sidecar + stamp). Same gate as the flat
--- a/src/zagg/hive.py
+++ b/src/zagg/hive.py
@@ -1629,7 +1629,8 @@ def process_and_write_hive(
     # sidecar is PUT just before it (issue #200 phase 2), and both inherit
     # its debris semantics: a torn worker's coverage never becomes visible.
     # The leaf write order is pinned: dense (streamed, or one object each when
-    # sharded) -> ragged (one object, issue #209) -> coverage sidecar -> stamp.
+    # sharded) -> ragged (one object, issue #209) -> coverage sidecar -> stamp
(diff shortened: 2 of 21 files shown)


src/zagg/dedup.py
```diff
@@ -242,13 +292,17 @@ def classify_leaf_identity(recorded, *, semantic_hash, planned_ids) -> dict:
     semantic_match = semantic_hash is not None and recorded.get("semantic_hash") == semantic_hash
     if rec_hash is not None and rec_hash == granules_sha256(planned) and semantic_match:
         return {"action": "skip", "classification": "equal", "missing": []}
-    rec_ids = recorded.get("granule_ids")
+    # The ONE read the fast path exists to avoid: the recorded id list lives
+    # beside the sidecar, not in it (issue #388's ruling on question (6)), and
+    # is fetched only now that the hashes disagree.
+    rec_ids = load_recorded_ids() if load_recorded_ids is not None else None
```
espg, line 333: 🤖 *from Claude (review)*

[medium] **The sibling read is not gated on the hash mismatch it is documented to be gated on — a semantic-only rerun fans out one large GET per shard that provably cannot change the verdict.**

The loader fires whenever the fast path fails, and the fast path is `rec_hash == granules_sha256(planned)` **AND** `semantic_match`:

```python
if rec_hash is not None and rec_hash == granules_sha256(planned) and semantic_match:
    return {"action": "skip", "classification": "equal", "missing": []}
rec_ids = load_recorded_ids() if load_recorded_ids is not None else None
```

So the "hashes equal, semantic hash differs" quadrant — the routine *same inputs, changed config* rerun, which is the ordinary reason to rerun a store with the gate armed (`runner.py`: `skip_if_current=not overwrite`, i.e. on by default locally) — pays the sibling read. Verified on this branch:

```
classify_leaf_identity(rec, semantic_hash='b'*64, planned_ids=ids, load_recorded_ids=load)
-> {'action': 'rewrite', 'classification': 'semantic-mismatch', 'missing': []}   loads = 1
```

Equal hashes mean an identical sorted id multiset (`granules_sha256` hashes `"\n".join(sorted(ids))`), so `missing` is necessarily empty and `set(planned) == recorded_set` necessarily holds — the only reachable outcome is `semantic-mismatch`, which is returnable without reading anything. At CA o9 that is 2,721 needless GETs; on a pole unit each one is the ~550 KB object the split exists to keep off the per-shard path. `docs/hive_layout.md` states the invariant as "fetched exactly once, only when the hashes disagree" — the code's trigger is broader than the doc's.

Second-order, same line: when the sibling is absent (a pre-#388 leaf, a lost PUT, a torn pair) the same quadrant returns

```
-> {'action': 'rewrite', 'classification': 'unrecorded-ids', 'missing': []}   loads = 1
```

`cells_unrecorded` is the counter `_identity_counts` documents as "the guard was INERT for them" and the phase-4 docs tell operators to treat specially — but with the hashes equal the guard is not inert, it has *proved* no contraction. Every config-only rerun over a pre-#388 store now reports 100% `cells_unrecorded`, which is the one signal that is supposed to shrink as leaves get rewritten under this release.

Both fall out of one guard: call the loader only when `rec_hash != granules_sha256(planned)`.

src/zagg/telemetry.py
```diff
@@ -474,16 +490,130 @@ def run_parquet_key(run_id: str, timestamp: str | None = None) -> str:
...
+
+    Fail-open (D9 telemetry class, the sweep run record's posture): a failed
+    PUT logs and returns ``None`` — the run's exit status and its
+    ``cells_refused`` count are unaffected. Callers must be store-writers in
+    their own right (D8): the local dispatcher is also the worker, which is
+    why this rides the same wrap-up seam as the root ``coverage.moc``.
+    """
+    import logging
+
+    units = []
```

[medium] **`write_refusal_manifest`'s fail-open covers only the PUT — the unit composition sits outside the `try`, so the unbounded-size case question (3) raises kills the run instead of degrading.**

Everything from `units = []` down through `body = {...}` runs before `try:`, and `_write_refusals` does not wrap the call either:

```python
    return write_refusal_manifest(
        store_path,
        [m for m in metas if isinstance(m, dict) and m.get("refused")],
        ...
    )
```

The docstring three lines up promises the opposite: *"Fail-open (D9 telemetry class, the sweep run record's posture): a failed PUT logs and returns `None` — the run's exit status and its `cells_refused` count are unaffected."* Contrast `write_granule_ids` in this same commit's sibling helper, which puts the whole body (`sorted(str(g) ...)`, `granules_sha256`, `json.dumps`) inside its try.

Why it matters concretely: in `_run_local` the call is at line 3212, **before** the `summary` dict is built and before the run-stats parquet write — the same ordering `RasterStrategy` documents as deliberate ("Run-level stats parquet … BEFORE the all-failed raise below so the failure evidence persists at the store root"). A raise in the composition therefore destroys the run summary *and* the `stats_*.parquet` of a run that has already done all of its work.

The realistic raisers are exactly the ones question (3) is about:

- `MemoryError` building `units` for a full-fleet refusal — 2,721 CA o9 units at pole-density recorded sets (~4,600 ids each) is ~1.25e7 strings and a >1 GB JSON body. A `MemoryError` inside `json.dumps` is caught and degrades; the identical failure one statement earlier is not.
- `TypeError` from `for g in (meta.get("missing_granules") or [])` on a malformed meta — a shape the function explicitly anticipates one line above with `if not isinstance(meta, dict): continue`, so the defensive posture is half-applied.

Moving the composition inside the `try` (or wrapping the call in `_write_refusals`) makes the docstring true and turns question (3)'s worst case into "no manifest" rather than "no run record".

src/zagg/hive.py
```diff
@@ -1640,6 +1655,15 @@ def _write_chunk(block_index, carrier, ragged):
...
         _write_elapsed += time.time() - _t0
+        # The recorded granule-id list, as this leaf's own sibling object
+        # (issue #388): AFTER the stamp, so it never certifies a leaf that
+        # did not land, and written here rather than at the caller's sidecar
+        # PUT because ``granule_urls`` is the very list the identity gate
+        # compares — one source for the recorded id space, on both backends.
+        # Fail-open inside (telemetry class, D9).
+        from zagg.telemetry import write_granule_ids
+
+        write_granule_ids(leaf_path, granule_urls, spec=sidecar_spec, **store_kwargs)
```
espg, line 1674: 🤖 *from Claude (review)*

[low] **The sibling PUT lands outside the write-phase bracket, so `phase_timings["write"]` stops accounting for every write the seam performs.**

```python
        _write_elapsed += time.time() - _t0
        # The recorded granule-id list, as this leaf's own sibling object ...
        from zagg.telemetry import write_granule_ids

        write_granule_ids(leaf_path, granule_urls, spec=sidecar_spec, **store_kwargs)
```

The accumulator closes one line before the PUT. Same shape on the raster seam (`processing/raster.py`: `write_s += time.time() - _t0` immediately precedes its `write_granule_ids(...)`), so both leaf families are affected. Everything else the bracket covers — the `coverage.moc` sidecar PUT and `stamp_commit` — is inside it, and the docstring's pinned order ("dense … -> ragged -> coverage sidecar -> stamp") is now one object longer than the bracket.

Consequence is measurement, not correctness: `phase_timings` flattens to `phase_write` on the D20 record (`telemetry.flatten_record`) and lands in every run parquet and every benchmark record, so from this release on the write bucket silently under-reports by one PUT per leaf and no longer decomposes the seam's wall. Reusing `_t0` and moving the `_write_elapsed +=` after the call costs nothing and keeps the bracket honest.

src/zagg/telemetry.py
```diff
@@ -695,6 +733,82 @@ def write_sidecar(leaf_path: str, record: dict, spec: str | None = None, **store
...
+        from zagg.store import open_object_store
+
+        prefix, _, name = leaf_path.rstrip("/").rpartition("/")
+        ids = sorted(str(g) for g in granule_ids) if granule_ids else []
+        obstore.put(
+            open_object_store(prefix, **store_kwargs),
+            granule_ids_key(name, spec),
+            json.dumps(
+                {
+                    "schema_version": SCHEMA_VERSION,
```

[low] **The granule-id sibling carries no `spec` marker of its own and borrows the D20 *record's* `SCHEMA_VERSION` — the PR body's "Both carry a `schema_version`/`spec` marker of their own so a future reader has a versioning handle" is only true of the refusal manifest.**

```python
json.dumps(
    {
        "schema_version": SCHEMA_VERSION,
        "granules_sha256": granules_sha256(ids),
        "granule_ids": ids,
    }
).encode(),
```

`SCHEMA_VERSION` is declared as *"Version stamped into every record (issue #297). The bump-on-key-change rule applies once the schema is released; while unreleased, pre-release key changes (e.g. the D19 rev 2 / D20 record) rev in place and this stays 1."* — it versions the run **record**, which this object is deliberately *not* (that separation is the whole point of ruling (6)(c)). Welding them means a future D20 rev bumps the sibling's declared version without the sibling changing, and a sibling format change cannot be versioned without revving the record.

The refusal manifest in this same PR got the shape right — `REFUSAL_SPEC = "zagg-refusals/1"` **plus** `schema_version` — and the `_sibling_key` refactor was justified precisely as "so the two siblings are named by ONE rule and cannot drift"; the version marker is the other half of that identity and it drifted. Note also that `dedup.leaf_recorded_ids` never reads either field, so today this is a write-only marker owned by the wrong schema: a reader that one day *does* branch on it would branch on D20's number.

A `GRANULE_IDS_SPEC = "zagg-granule-ids/1"` beside `GRANULE_IDS_NAME`, emitted alongside the existing key, matches the manifest's own precedent and costs one line.

src/zagg/dedup.py
espg: 🤖 *from Claude*

Fixed in `c320eea` — the loader is now gated on the granule hashes DISAGREEING, and the equal-hash/changed-semantic quadrant returns `semantic-mismatch` before any read:

```python
hashes_match = rec_hash is not None and rec_hash == granules_sha256(planned)
if hashes_match and semantic_match:
    return {"action": "skip", "classification": "equal", "missing": []}
if hashes_match:
    # Equal hashes mean an identical sorted id multiset, so the diff is
    # provably empty and the ONLY reachable verdict is semantic-mismatch ...
    return {"action": "rewrite", "classification": "semantic-mismatch", "missing": []}
rec_ids = load_recorded_ids() if load_recorded_ids is not None else None
```

Both halves are pinned:

- `test_semantic_mismatch_with_equal_sets_rewrites_not_refuses` now asserts `self.loads == 0` — the 2,721 needless CA o9 GETs are gone.
- New `test_semantic_mismatch_without_a_sibling_is_not_unrecorded`: the same quadrant with the sibling absent returns `semantic-mismatch`, not `unrecorded-ids`, so a config-only rerun over a pre-#388 store no longer reports 100% `cells_unrecorded`.

`id-multiset-drift` still needs the read (hashes differ, sets equal) and is unaffected — `test_duplicate_drift_with_equal_sets_rewrites` still passes.

Docs followed: the `load_recorded_ids` parameter docstring now says "ONLY when the recorded and planned `granules_sha256` DISAGREE", and `docs/hive_layout.md`'s pre-#388 paragraph now says *granule-hash* mismatch, with an explicit note that a config-only rerun is not one of these and never inflates `cells_unrecorded`. The "fetched exactly once, only when the hashes disagree" invariant is now true as written.

src/zagg/telemetry.py

Fixed in `5a7e3eb` — the whole body of `write_refusal_manifest` (unit composition, sort, `body` dict, key, PUT) is now inside the one `try`, matching `write_granule_ids`'s posture in the same PR. The `if not units: return None` early return moved inside with it, so the nothing-refused case is unchanged.

The docstring now states the widened guarantee and why the ordering makes it matter:

> The COMPOSITION is inside the same guard as the PUT (`write_granule_ids`'s posture), because this runs before the summary and the run-stats parquet: a `MemoryError` on an unbounded refusal set, or a malformed `missing_granules`, must cost the manifest, never the run record of a run that already did all its work.

New `TestRefusalManifest::test_composition_is_fail_open_too` pins both raisers you named, at the exact statements:

- `missing_granules=7` → the `TypeError` out of `for g in (meta.get("missing_granules") or [])` → `None`.
- a truthy list subclass whose `__iter__` raises `MemoryError` → `None`.

Left `_write_refusals` unwrapped: after this its only work is the `identity["cells_refused"]` lookup and the import, and wrapping it too would hide a genuine programming error at the seam without covering anything the helper now doesn't.

src/zagg/hive.py
espg: 🤖 *from Claude*

Fixed in `45bb7eb`, on both leaf families — the accumulator now closes AFTER the sibling PUT, reusing the same `_t0`:

```python
        write_granule_ids(leaf_path, granule_urls, spec=sidecar_spec, **store_kwargs)
        _write_elapsed += time.time() - _t0
```

and the raster seam's `write_s += time.time() - _t0` moved below its `write_granule_ids(...)` identically.

The three pinned-order comments that stopped at the stamp now name the sibling, so the docstring and the bracket agree again: `hive.py`'s "dense … -> ragged -> coverage sidecar -> stamp -> granule-id sibling (issue #388; after the stamp, inside the bracket)", and both raster sites ("dense slabs -> coverage sidecar -> stamp -> granule-id sibling"; the write-out gloss is now "template + slabs + sidecar + stamp + sibling").

Pinned by a new `test_write_phase_covers_the_granule_id_sibling`: it wraps `telemetry.write_granule_ids` in a 50 ms sleep and asserts `meta["phase_timings"]["write"] >= 0.05`, so a future move back outside the bracket fails rather than silently under-reporting `phase_write` in every run parquet.

src/zagg/telemetry.py

Fixed in `7b04101` — `GRANULE_IDS_SPEC = "zagg-granule-ids/1"` now sits beside `GRANULE_IDS_NAME`, and the object carries it:

```python
{
    "spec": GRANULE_IDS_SPEC,
    "granules_sha256": granules_sha256(ids),
    "granule_ids": ids,
}
```

One judgement call worth flagging: I **replaced** `schema_version` rather than emitting the marker alongside it. Keeping `SCHEMA_VERSION` would have preserved the exact defect you identified — a D20 rev would still bump the sibling's declared version — and the sibling has no second version axis to carry, so `spec` is the whole handle. Say the word if you want both keys for symmetry with the refusal manifest; it is one line. (The manifest keeps both because it *is* run-record-class root telemetry.)

Lenient read, as asked: `leaf_recorded_ids` never branches on the marker — the hash pairing decides — and its docstring now says so explicitly, so an unknown `zagg-granule-ids/N` that keeps these two keys stays readable and an unrecognized one degrades to the ruled `unrecorded-ids` fallback rather than raising. Pinned by `TestLeafRecordedIds::test_the_spec_marker_is_read_leniently` (marker rewritten to `zagg-granule-ids/99` and to `None`, both still round-trip the ids); `test_write_read_roundtrip_sorted_and_self_hashing` pins `spec == "zagg-granule-ids/1"` and the absence of `schema_version`.

The PR body's Spec-assessment sentence is corrected in the same pass — it now says the manifest carries `spec` + `schema_version` and the sibling carries its own `spec`, instead of claiming both carry a `schema_version`/`spec` pair.

espg:
🤖 *from Claude*

**Question triage — three of the four are defaults, not decisions. Merging this PR accepts them.**

Reviewing the four questions against what is actually at stake, only one warranted espg's time, and it is now filed separately so it does not gate this merge:

- **(1) seam vs dispatcher placement — RECOMMENDED: keep, merging accepts.** This is not an open question any more; it has data. The adversarial review verified the implementation, the justification, and the cost: gated on a committed leaf, one object per (node, window), **+9.1% objects/shard** (11→12 on the o9 hive config), **+2,721 PUTs ≈ $0.0136 against $65.30 compute = 0.02%**, ~110 B/id. It also established that no test ever pinned the fleet's exact object set, so #397's "byte-identical when unarmed" property was unpinned to begin with. The alternative leaves the contraction guard permanently inert on the fleet.
- **(2) `refusal_manifest_path` on local summaries only — RECOMMENDED: leave as-is.** A coin flip with a stated rationale (an always-`None` fleet key reads as "no refusals" rather than "not recorded"). Mirroring it is one line whenever a fleet writer exists.
- **(3) manifest unbounded — RECOMMENDED: leave as-is.** Proportional-to-the-refusal-set is the point of (9)(c); truncating would defeat it, and fold `5a7e3eb` already made the worst case degrade safely (composition moved inside the fail-open guard, so a `MemoryError` costs the manifest, not the run record). Revisit if a real run produces an unwieldy object.
- **(4) the torn-rewrite hole — SPLIT OUT to #407.** This is the one genuine item: pre-existing (inherited from #397, neither widened nor narrowed here), a correctness question, and a D4 write-discipline change that touches the spec — so it belongs in its own PR with a fixture pass, not in a three-ruling follow-up trio.

**So: this PR is ready to merge.** All three rulings implemented, review folded (5 findings), CI green on `7b04101`, `MERGEABLE`. Nothing above blocks it; (1)–(3) are accepted by merging, and (4) lives on #407.

Tracked in #402.

PR #412, opened 2026-08-08
state satisfied mortie minimums in arrival voice, guard the rest

espg:
No originating issue — directed by espg in conversation on 2026-08-08, so the rationale lives here.

## Why

zagg's prose cites mortie release numbers in ~16 places, in three distinct voices. Only one of them is a live obligation, and conflating them makes the docs read as if the package carried more unenforced version contracts than it does:

1. **The package floor** — `mortie>=0.9.3` in `[project.dependencies]`. Single source: `deployment/aws/build_layer.sh` derives `MORTIE_SPEC` from it via `tomllib` (issue #322), and it was removed from `CLAUDE.md` §7 in 6988e9e6 for the same reason. Not touched here.
2. **Per-feature minimums with no runtime enforcement** — historically true, permanently satisfied because the floor already exceeds them. Written in requirement voice (`mortie >= 0.8.4`) they read as live obligations a reader might have to satisfy; they are not. **These get the voice fix.**
3. **Genuine provenance** — "resolved in mortie 0.9.0", "mortie 0.8.5, espg/mortie#100". Immutable statements about when something landed, correct as written. Left alone.

## Phases

- [x] Phase 1 — restate the four unenforced per-feature minimums in arrival voice (prose only, no behavior change)
- [x] Phase 2 — drift guard so the prose can never promise more than the floor guarantees
- [x] Phase 3 — adversarial self-review folded (6 of 7 findings; the 7th is declined and stands for review)

## Phase 1 — the four sites

Requirement voice → arrival voice, matching each site's surrounding prose rather than one template:

**`docs/morton_arrow.md:11`** (table cell — terse, so the parenthetical stays terse):

```diff
-| arro3 carrier (worker) | `arro3.core.Table` | `morton_index` Arrow **extension type** over the PyCapsule C Data Interface (mortie ≥ 0.8.4) |
+| arro3 carrier (worker) | `arro3.core.Table` | `morton_index` Arrow **extension type** over the PyCapsule C Data Interface (added in mortie 0.8.4) |
```

**`docs/ragged_layout.md:144`** (mid-sentence, beside the "frozen for mortie 1.x" provenance that follows it):

```diff
-`mortie.rank_to_xy` / `xy_to_rank` (mortie ≥ 0.9.3), whose contract is
+`mortie.rank_to_xy` / `xy_to_rank` (added in mortie 0.9.3), whose contract is
```

**`src/zagg/grids/morton.py:7`** (module docstring; rewrapped so `mortie 0.8.4` stays on one line):

```diff
-Interface (:func:`morton_to_arrow` / :func:`morton_from_arrow`; mortie >= 0.8.4,
-issue #135) — typed on both the pandas and arro3 surfaces, no pyarrow on the
-worker path. On disk it is stored as plain ``uint64`` — Zarr stores numpy
-dtypes, and the extension metadata lives at the interchange layer only — and
-reconstructed as a ``MortonIndexArray`` on read.
+Interface (:func:`morton_to_arrow` / :func:`morton_from_arrow`; added in
+mortie 0.8.4, issue #135) — typed on both the pandas and arro3 surfaces, no
+pyarrow on the worker path. On disk it is stored as plain ``uint64`` — Zarr
+stores numpy dtypes, and the extension metadata lives at the interchange layer
+only — and reconstructed as a ``MortonIndexArray`` on read.
```

**`src/zagg/grids/morton.py:202`** (`morton_to_arrow` docstring):

```diff
-    Interface (``MortonIndexArray.__arrow_c_array__``; mortie >= 0.8.4) — no
+    Interface (``MortonIndexArray.__arrow_c_array__``; added in mortie 0.8.4) — no
```

No code changed anywhere in phase 1 — the only non-citation edit is the rewrap of the three lines following the `morton.py` module-docstring citation, which keeps the paragraph at its existing ~78-column width.

### Deliberately not touched: the `aoi_mask` six

Six further sites cite `mortie >= 0.8.3` — `docs/aoi_mask.md:82,102,197,204` and `src/zagg/grids/aoi.py:26,88`. Requirement voice is *correct* there, because that one is a **real, enforced runtime gate**. `src/zagg/grids/aoi.py:42` defines the constant and `_assert_mortie_version` raises on a stale environment (`src/zagg/grids/aoi.py:66-80`):

```python
MIN_MORTIE_VERSION = "0.8.3"
...
    if resolved < Version(MIN_MORTIE_VERSION):
        raise RuntimeError(
            f"aoi_mask requires mortie >= {MIN_MORTIE_VERSION} (the order-29 MOC "
            f"coverage cap, espg/mortie#59 + #70, plus the public WKB/WKT cover "
            f"entry points, espg/mortie#89); resolved {raw}. Upgrade mortie "
            "or disable output.aoi_mask."
        )
```

Rewriting those six to arrival voice would make the docs *less* accurate — a user really can hit that `RuntimeError`. They stay, and phase 2 pins them to the constant instead.

Also untouched, as category (3): `src/zagg/grids/healpix.py:37,355`, `src/zagg/grids/morton.py:122`, `src/zagg/readers/_layout.py:7`, `docs/design/sparse_coverage.md:932`, and the `child_order 19` config-YAML comments — all already arrival/provenance voice.

## Phase 2 — the drift guard

`tests/test_mortie_versions.py`, following the `pyproject.toml`-reading convention already in `tests/test_lambda_build.py::TestLayerExtraParity` (`tomllib.loads((REPO_ROOT / "pyproject.toml").read_text())`) and the floor-vs-runtime check in `tests/test_rectilinear.py::TestTransformerThreadSafety`.

**Two invariants:**

1. Every mortie version quoted under `docs/` or `src/` is **≤ the `pyproject.toml` floor**. This catches the one failure mode that matters: prose promising a feature the dependency does not guarantee, so a user hits `AttributeError` where the docs said they wouldn't. The floor is parsed from `[project.dependencies]` with `packaging.requirements.Requirement`, and every comparison goes through `packaging.version.Version` — never a string compare.
2. `MIN_MORTIE_VERSION` is **≤ the floor**, and every **requirement-voice** version quoted in the `aoi_mask` family (`docs/aoi_mask.md`, `src/zagg/grids/aoi.py`) **equals** it. That constant is the source of truth for the enforced gate, and those six sites are exactly the ones that cannot be de-versioned — so they are where a guard earns its keep. Arrival voice in those same two files ("the MOC cap shipped in mortie 0.8.2") is history rather than a restatement of the gate, so it is held to the floor like any other citation, not to the constant.

**The regex** — `mortie[\s(:,_-]*(?:version|release)?\s*(>=|==|~=|>|≥)?\s*v?(\d+\.\d+(?:\.\d+)*)`, case-insensitive — handles `>=`, `≥`, `==`/`~=`/`>`, bare `mortie 0.8.1`, `mortie (0.9.0)` / `mortie: 0.9.0` / `mortie-0.9.0`, `mortie version 0.9.0`, sentence-initial `Mortie 0.9.0`, and backticked/unbackticked variants (backticks fall outside the match). Whitespace spans newlines so a citation wrapped across a line break is not a blind spot. It does **not** match `espg/mortie#89` (`#` is outside the separator class), `frozen for mortie 1.x` (no second numeric component), or upper bounds like `mortie < 1.0` (an upper bound promises no API). Group 1 — the comparison operator, present only in requirement voice — is what scopes invariant (2).

The floor line in `pyproject.toml` is never read back as a citation because `.toml` is not in `SCAN_SUFFIXES` *and* the file sits outside both `SCAN_ROOTS`; the guard asserts both mechanisms rather than a path check that no scanned suffix could satisfy anyway.

**What it finds — 16 sites, floor `0.9.3`, of which 6 are requirement voice:**

```
docs/aoi_mask.md:82                            0.8.3  requirement
docs/aoi_mask.md:102                           0.8.3  requirement
docs/aoi_mask.md:197                           0.8.3  requirement
docs/aoi_mask.md:204                           0.8.3  requirement
docs/design/sparse_coverage.md:932             0.9.0  arrival
docs/morton_arrow.md:11                        0.8.4  arrival
docs/ragged_layout.md:144                      0.9.3  arrival
src/zagg/configs/atl03_tdigest_healpix.yaml:13 0.8.1  arrival
src/zagg/grids/aoi.py:26                       0.8.3  requirement
src/zagg/grids/aoi.py:88                       0.8.3  requirement
src/zagg/grids/healpix.py:37                   0.8.1  arrival
src/zagg/grids/healpix.py:355                  0.8.5  arrival
src/zagg/grids/morton.py:8                     0.8.4  arrival
src/zagg/grids/morton.py:122                   0.9.1  arrival
src/zagg/grids/morton.py:202                   0.8.4  arrival
src/zagg/readers/_layout.py:7                  0.9.3  arrival
```

The 6 requirement-voice sites are exactly the `aoi_mask` family, all `0.8.3` == `MIN_MORTIE_VERSION`; the other ten are all ≤ `0.9.3`. Two of them (`morton.py:122`, `readers/_layout.py:7`) sit inside sentences where the word `mortie` and the number land on different lines, so a line-scoped grep misses them — the newline-tolerant whitespace is what surfaces them.

### Proving the guard fires

A guard that passes vacuously is worse than none, so non-vacuity is asserted rather than assumed:

- `test_scan_finds_the_versioned_sites` asserts the scan returns ≥ 12 sites — a regex that silently matched nothing would otherwise let every other test in the module pass forever. Deliberately a count and not a per-file list: naming the files would ratchet them in place, and *deleting* an unenforced per-feature minimum is the very edit phase 1 argues for.
- `test_a_citation_above_the_floor_is_caught` feeds a synthetic `mortie >= 99.0.0` through the *same* `_cited_versions` / `_above` helpers the tree scan uses and asserts it is flagged.
- `test_ordering_is_pep440_not_lexicographic` pins the exact case a string compare gets backwards: `"0.10.0" < "0.9.5"` as text, `Version("0.10.0") > Version("0.9.5")` as versions, and `_above` agreeing with the latter.
- `test_issue_references_are_not_versions`, `test_every_citation_spelling_is_recognized`, and `test_the_spellings_that_used_to_pass_silently_are_caught` pin the regex's negative and positive cases — including `mortie==0.10.0`, sentence-initial `Mortie`, and the wrapped-across-a-newline spelling.
- `test_a_wrapped_citation_reports_the_version_line` pins that a site is reported at the line carrying the *number*, not the line carrying the word `mortie`, so a failure never sends the reader to a line with no version on it.
- `test_requirement_voice_is_told_from_arrival_voice` pins the operator split that invariant (2) rides on.
- `test_the_floor_line_is_not_scanned_as_a_citation` shows the regex *would* match `"mortie>=0.9.3"` in `pyproject.toml`, then asserts the two mechanisms that actually exclude it (`.toml` absent from `SCAN_SUFFIXES`, and the file outside `SCAN_ROOTS`).

Beyond the unit tests, both invariants were mutation-tested against the real tree, twice — once as first written and once after the fold:

```
# docs/ragged_layout.md:144 → mortie 9.9.9
AssertionError: prose quotes a mortie version above the pyproject floor 0.9.3:
[('docs/ragged_layout.md', 144, <Version('9.9.9')>)]. Either raise the floor in
[project.dependencies] (an espg-authorized dependency change) or restate the prose
against a shipped release.

# src/zagg/grids/aoi.py:26 → mortie >= 0.8.9
AssertionError: aoi_mask prose disagrees with MIN_MORTIE_VERSION (0.8.3):
[('src/zagg/grids/aoi.py', 26, <Version('0.8.9')>)]
```

Both mutations were reverted; `git status` is clean of them.

## Phase 3 — adversarial self-review, folded

A fresh-context review agent posted 7 inline findings (§2). Six were folded, one commit each; one is declined and stands for review:

| Finding | Disposition |
| --- | --- |
| Line numbers off by one for citations stitched across a newline | Fixed — report the version group's own line (`0f9d8798`) |
| Regex false negatives: no `re.IGNORECASE`, only `>=`/`≥`, so `mortie==0.10.0` passed silently | Fixed — widened operator + separator sets, with regression cases both ways (`ac5d864f`) |
| Floor parse: first-match `next()`, uncanonicalized name compare, wildcard pin raising `InvalidVersion` | Fixed — collect every mortie requirement and assert exactly one, `canonicalize_name`, wildcard kept out of `Version` (`01ca19f1`) |
| The per-file presence list ratchets citations in place, against phase 1's own thesis | Fixed — count-based anti-vacuity assertion kept, file list demoted to a comment (`6f989ed9`) |
| `aoi.py:35,37,54` state versions in detached voice, invisible to the scan, so a `MIN_MORTIE_VERSION` bump leaves them stale — and file-scoped equality *penalized* fixing them | Fixed — equality scoped to requirement voice, claim narrowed, residual blind spot named in the module docstring (`4cf0c51c`) |
| `test_the_floor_line_is_not_scanned_as_a_citation` was a tautology, and misattributed the mechanism to `SCAN_ROOTS` | Fixed — asserts `SCAN_SUFFIXES` and a roots check that can actually fail (`01bb10b2`) |
| Regex false *positives*: `mortie 1.0` or "gated on mortie 0.10.0" would redden correct prose | **Declined, standing for review.** A sentence naming a release above the floor is the signal this guard exists to raise, and both proposed remedies (an opt-out marker, or excluding `docs/design/`) narrow the directive rather than the false-positive set. Worth overriding if the design docs' forward-looking voice turns out to trip it in practice. |

## How it was tested

`uv run ruff check src tests`, `uv run ruff format --check src tests`, `uv run pytest -q` in a worktree synced with the catalog/analysis/viz/benchmark/test extras + dev group (mortie 0.9.5).

Phase 1 is prose-only, so the existing suite is its regression check. Baseline with the new file excluded: **3699 passed, 10 skipped, 1 failed**. With it: **3711 passed, 10 skipped, 1 failed** — exactly the 12 tests this PR adds. The one failure is `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, pre-existing, local-only, and green on CI.

## Questions for review

- **Scan scope — the known blind spot.** A version is recognized only where the release number is lexically attached to the word `mortie`. Three bare literals inside `src/zagg/grids/aoi.py` (`:35` "shipped in 0.8.2", `:37` "ship in 0.8.3", `:54` "the same 0.8.3 release" — the last inside `_assert_mortie_version`'s own docstring) are therefore invisible to the guard, so a `MIN_MORTIE_VERSION` bump still needs a human eye on that block. Widening to catch them means matching every bare `x.y.z` in prose that also carries `zarr>=3.1.5`, `pyproj 3.7`, `numpy 2.2.6`, and Python 3.12 — I judged the false-positive cost worse than the gap, and wrote it into the module docstring as a scope note rather than hiding it. Say the word if you'd rather have the noisier version.
- **False positives are possible, by design.** `mortie 1.0` matches and would fail the guard; so would a forward-looking "gated on mortie 0.10.0" of the kind `docs/design/sparse_coverage.md:392,905` currently writes without a number. This is the declined review finding above. My reasoning is that such a sentence is exactly what should get a human's attention, but the counter-argument (design docs are narrative, not promises) is real — excluding `docs/design/` from `SCAN_ROOTS` is a one-line change if you prefer it.
- **`tests/` is not scanned.** `tests/test_grids.py:158,165` and `tests/test_morton_arrow.py:3` cite mortie versions too. They document test intent rather than a user-facing promise, so they are out of scope here — say the word if they should be in.

Pre-existing and not fixed here (unrelated to this change): `N818` at `src/zagg/registry.py:64`, and the `ruff format --check` diff on `tests/data/benchmark/README.md`.

### DIFF
--- a/tests/test_mortie_versions.py
+++ b/tests/test_mortie_versions.py
@@ -17,12 +17,19 @@ an API the dependency floor does not guarantee, which reaches a user as an
 ``AttributeError`` where the docs said otherwise. So every mortie version
 quoted under ``docs/`` or ``src/`` must be at or below the floor, and the
 ``aoi_mask`` family — the sites that document (3) and therefore cannot be
-de-versioned — must quote the enforced constant exactly.
+de-versioned — must quote the enforced constant exactly wherever they state it
+as a *requirement* (``mortie >= 0.8.3``). Arrival voice in those same files
+("shipped in 0.8.2") is history, not a restatement of the gate, so it is held
+to the floor like any other citation and not to the constant.
 
-Scope note: a citation is recognized only where the release number is
-lexically attached to the word ``mortie``. A bare "shipped in 0.8.2" further
-along the same sentence (``src/zagg/grids/aoi.py``) is not matched, because no
-regex separates it from the other version numbers prose carries.
+Scope note — two blind spots, both deliberate and both needing a human eye on a
+``MIN_MORTIE_VERSION`` bump. A citation is recognized only where the release
+number is lexically attached to the word ``mortie``: bare literals further along
+a sentence ("espg/mortie#59 + #70, shipped in 0.8.2", ``src/zagg/grids/aoi.py``)
+are invisible here, because no regex separates them from the other version
+numbers prose carries. And whether a wrapped citation is seen depends on the
+words between ``mortie`` and the number, so reflowing a paragraph can move one
+in or out of scope.
 """
 
 import re
@@ -61,20 +68,26 @@ AOI_MASK_FILES = ("docs/aoi_mask.md", "src/zagg/grids/aoi.py")
 # version — ``#`` is outside the separator class), "frozen for mortie 1.x" (no
 # second numeric component), and upper bounds like ``mortie < 1.0``, which
 # promise no API and so are not citations this guard is about.
+#
+# Group 1 is the comparison operator, present only in **requirement** voice
+# ("zagg needs at least this"); arrival voice ("added in mortie 0.8.4") leaves
+# it ``None``. That distinction is what scopes the ``aoi_mask`` equality rule.
 _CITATION = re.compile(
-    r"mortie[\s(:,_-]*(?:version|release)?\s*(?:>=|==|~=|>|≥)?\s*v?(\d+\.\d+(?:\.\d+)*)",
+    r"mortie[\s(:,_-]*(?:version|release)?\s*(>=|==|~=|>|≥)?\s*v?(\d+\.\d+(?:\.\d+)*)",
     re.IGNORECASE,
 )
 
 
-def _cited_versions(text):
+def _cited_versions(text, requirement_voice_only=False):
     """Every mortie release ``text`` names, as ``(Version, line number)``."""
     return [
-        (Version(m.group(1)), text.count("\n", 0, m.start(1)) + 1) for m in _CITATION.finditer(text)
+        (Version(m.group(2)), text.count("\n", 0, m.start(2)) + 1)
+        for m in _CITATION.finditer(text)
+        if m.group(1) or not requirement_voice_only
     ]
 
 
-def _scan():
+def _scan(requirement_voice_only=False):
     """Every mortie citation under the scan roots, as ``(path, line, Version)``."""
     sites = []
     for root in SCAN_ROOTS:
@@ -82,7 +95,8 @@ def _scan():
             if path.suffix not in SCAN_SUFFIXES or not path.is_file():
                 continue
             rel = path.relative_to(REPO_ROOT).as_posix()
-            for version, line in _cited_versions(path.read_text(encoding="utf-8")):
+            text = path.read_text(encoding="utf-8")
+            for version, line in _cited_versions(text, requirement_voice_only):
                 sites.append((rel, line, version))
     return sites
 
@@ -148,7 +162,10 @@ class TestQuotedVersionsAgainstFloor:
     def test_aoi_mask_prose_matches_the_runtime_gate(self):
         # These six sites document a gate that raises (``aoi.py`` lines 66-80),
         # so they keep requirement voice — and must quote the constant itself.
-        family = [s for s in _scan() if s[0] in AOI_MASK_FILES]
+        # Requirement voice only: the same two files also carry arrival-voice
+        # history ("shipped in 0.8.2"), which is factually correct at a version
+        # *below* the gate and must not be dragged into an equality rule.
+        family = [s for s in _scan(requirement_voice_only=True) if s[0] in AOI_MASK_FILES]
         assert len(family) >= 6, f"aoi_mask citations vanished from the scan: {family}"
         wrong = [s for s in family if s[2] != Version(MIN_MORTIE_VERSION)]
         assert not wrong, (
@@ -217,6 +234,13 @@ class TestGuardFires:
             "tagged mortie v0.10.0",
         ):
             assert [str(v) for v, _ in _cited_versions(spelling)] == ["0.10.0"], spelling
+
+    def test_requirement_voice_is_told_from_arrival_voice(self):
+        # The split the ``aoi_mask`` equality rule rides on: only a citation
+        # carrying a comparison operator states an obligation.
+        text = "aoi_mask needs mortie >= 0.8.3; the MOC cap shipped in mortie 0.8.2"
+        assert [str(v) for v, _ in _cited_versions(text)] == ["0.8.3", "0.8.2"]
+        assert [str(v) for v, _ in _cited_versions(text, requirement_voice_only=True)] == ["0.8.3"]
 
     def test_the_floor_line_is_not_scanned_as_a_citation(self):
         # The regex would happily match ``mortie>=0.9.3`` in pyproject.toml, so
--- a/tests/test_mortie_versions.py
+++ b/tests/test_mortie_versions.py
@@ -1,0 +1,179 @@
+"""Drift guard: no mortie version zagg's prose quotes may outrun the floor.
+
+zagg names mortie release numbers in three different voices, and only one of
+them is a live obligation:
+
+1. the **package floor** — ``mortie>=…`` in ``[project.dependencies]``, the
+   single source (``deployment/aws/build_layer.sh`` derives ``MORTIE_SPEC``
+   from it, issue #322);
+2. **arrival notes** in narrative docs and docstrings ("added in mortie
+   0.8.4") — historically true, permanently satisfied because the floor
+   already exceeds them, and enforced by nothing at runtime;
+3. the one **runtime gate**, :data:`zagg.grids.aoi.MIN_MORTIE_VERSION`, which
+   ``_assert_mortie_version`` turns into a ``RuntimeError``.
+
+The failure mode worth guarding is (2) drifting *above* (1): prose promising
+an API the dependency floor does not guarantee, which reaches a user as an
+``AttributeError`` where the docs said otherwise. So every mortie version
+quoted under ``docs/`` or ``src/`` must be at or below the floor, and the
+``aoi_mask`` family — the sites that document (3) and therefore cannot be
+de-versioned — must quote the enforced constant exactly.
+
+Scope note: a citation is recognized only where the release number is
+lexically attached to the word ``mortie``. A bare "shipped in 0.8.2" further
+along the same sentence (``src/zagg/grids/aoi.py``) is not matched, because no
+regex separates it from the other version numbers prose carries.
+"""
+
+import re
+import tomllib
+from pathlib import Path
+
+from packaging.requirements import Requirement
+from packaging.version import Version
+
+from zagg.grids.aoi import MIN_MORTIE_VERSION
+
+REPO_ROOT = Path(__file__).parent.parent
+
+# Prose scanned for citations: narrative docs plus the shipped package
+# (docstrings, comments, bundled config YAML). ``pyproject.toml`` is
+# deliberately outside both roots — it is the floor these citations are
+# measured against, not one of them.
+SCAN_ROOTS = (REPO_ROOT / "docs", REPO_ROOT / "src")
+SCAN_SUFFIXES = {".md", ".py", ".yaml", ".yml"}
+
+# The ``aoi_mask`` family: the six sites documenting the runtime gate.
+AOI_MASK_FILES = ("docs/aoi_mask.md", "src/zagg/grids/aoi.py")
+
+# ``mortie`` followed by a release number, in every voice the tree uses:
+# ``mortie >= 0.8.3``, ``mortie ≥ 0.8.4``, ``mortie>=0.8.3``, bare
+# ``mortie 0.8.1``, and the backticked variants (the backticks fall outside
+# the match). Whitespace spans newlines so a wrapped citation is not a blind
+# spot. ``espg/mortie#89`` does not match: an issue reference is not a
+# version, and neither is "frozen for mortie 1.x".
+_CITATION = re.compile(r"mortie\s*(?:>=|≥)?\s*v?(\d+\.\d+(?:\.\d+)*)")
+
+
+def _cited_versions(text):
+    """Every mortie release ``text`` names, as ``(Version, line number)``."""
+    return [
+        (Version(m.group(1)), text.count("\n", 0, m.start()) + 1) for m in _CITATION.finditer(text)
+    ]
+
+
+def _scan():
+    """Every mortie citation under the scan roots, as ``(path, line, Version)``."""
+    sites = []
+    for root in SCAN_ROOTS:
+        for path in sorted(root.rglob("*")):
+            if path.suffix not in SCAN_SUFFIXES or not path.is_file():
+                continue
+            rel = path.relative_to(REPO_ROOT).as_posix()
+            for version, line in _cited_versions(path.read_text(encoding="utf-8")):
+                sites.append((rel, line, version))
+    return sites
+
+
+def _above(sites, floor):
+    """The comparison the guard makes — PEP 440 ordering, never string compare."""
+    return [s for s in sites if s[2] > floor]
+
+
+def _mortie_floor():
+    """The ``>=`` floor from ``[project.dependencies]``."""
+    deps = tomllib.loads((REPO_ROOT / "pyproject.toml").read_text(encoding="utf-8"))
+    reqs = [Requirement(d) for d in deps["project"]["dependencies"]]
+    spec = next(r.specifier for r in reqs if r.name == "mortie")
+    lower = [Version(s.version) for s in spec if s.operator in (">=", "==", "~=")]
+    assert len(lower) == 1, f"expected one mortie lower bound in pyproject, got {lower}"
+    return lower[0]
+
+
+class TestQuotedVersionsAgainstFloor:
+    """Nothing zagg's prose promises may exceed what the floor guarantees."""
+
+    def test_scan_finds_the_versioned_sites(self):
+        # A regex that silently matched nothing would let every other test in
+        # this module pass forever, so pin that the scan has real work to do.
+        sites = _scan()
+        assert len(sites) >= 12, f"mortie citation scan went blind: {sites}"
+        files = {path for path, _, _ in sites}
+        for expected in (
+            "docs/aoi_mask.md",
+            "docs/morton_arrow.md",
+            "docs/ragged_layout.md",
+            "src/zagg/grids/aoi.py",
+            "src/zagg/grids/morton.py",
+            "src/zagg/grids/healpix.py",
+        ):
+            assert expected in files, f"{expected} no longer reached by the scan"
+
+    def test_no_citation_outruns_the_floor(self):
+        floor = _mortie_floor()
+        over = _above(_scan(), floor)
+        assert not over, (
+            f"prose quotes a mortie version above the pyproject floor {floor}: {over}. "
+            "Either raise the floor in [project.dependencies] (an espg-authorized "
+            "dependency change) or restate the prose against a shipped release."
+        )
+
+    def test_runtime_gate_is_at_or_below_the_floor(self):
+        floor = _mortie_floor()
+        assert Version(MIN_MORTIE_VERSION) <= floor, (
+            f"zagg.grids.aoi.MIN_MORTIE_VERSION ({MIN_MORTIE_VERSION}) is above the "
+            f"pyproject floor ({floor}) — aoi_mask would raise on a conforming install"
+        )
+
+    def test_aoi_mask_prose_matches_the_runtime_gate(self):
+        # These six sites document a gate that raises (``aoi.py`` lines 66-80),
+        # so they keep requirement voice — and must quote the constant itself.
+        family = [s for s in _scan() if s[0] in AOI_MASK_FILES]
+        assert len(family) >= 6, f"aoi_mask citations vanished from the scan: {family}"
+        wrong = [s for s in family if s[2] != Version(MIN_MORTIE_VERSION)]
+        assert not wrong, (
+            f"aoi_mask prose disagrees with MIN_MORTIE_VERSION ({MIN_MORTIE_VERSION}): {wrong}"
+        )
+
+
+class TestGuardFires:
+    """A guard that cannot fail is worse than no guard — prove each part bites."""
+
+    def test_a_citation_above_the_floor_is_caught(self):
+        floor = _mortie_floor()
+        text = "the ragged reader rides `mortie >= 99.0.0`, whose contract is normative"
+        sites = [("docs/synthetic.md", line, v) for v, line in _cited_versions(text)]
+        assert [str(s[2]) for s in sites] == ["99.0.0"]
+        assert _above(sites, floor) == sites
+
+    def test_ordering_is_pep440_not_lexicographic(self):
+        # The case a string compare gets backwards: "0.10.0" < "0.9.5" as text.
+        assert "0.10.0" < "0.9.5"
+        assert Version("0.10.0") > Version("0.9.5")
+        sites = [("docs/synthetic.md", 1, Version("0.10.0"))]
+        assert _above(sites, Version("0.9.5")) == sites
+        assert _above(sites, Version("0.10.0")) == []
+
+    def test_issue_references_are_not_versions(self):
+        assert _cited_versions("(espg/mortie#89, espg/mortie#100) and mortie 1.x") == []
+
+    def test_every_citation_spelling_is_recognized(self):
+        text = (
+            "`mortie >= 0.8.3` and mortie ≥ 0.8.4 and `mortie>=0.9.3` and "
+            "mortie 0.8.1 and added in mortie\n0.9.0"  # wrapped by a line break
+        )
+        assert [str(v) for v, _ in _cited_versions(text)] == [
+            "0.8.3",
+            "0.8.4",
+            "0.9.3",
+            "0.8.1",
+            "0.9.0",
+        ]
+
+    def test_the_floor_line_is_not_scanned_as_a_citation(self):
+        # The regex would happily match ``mortie>=0.9.3`` in pyproject.toml, so
+        # the exclusion has to come from the scan roots — not from luck.
+        floor = _mortie_floor()
+        assert [v for v, _ in _cited_versions(f'    "mortie>={floor}",')] == [floor]
+        scanned = {path for path, _, _ in _scan()}
+        assert not any(path.endswith("pyproject.toml") for path in scanned)
--- a/tests/test_mortie_versions.py
+++ b/tests/test_mortie_versions.py
@@ -120,18 +120,14 @@ class TestQuotedVersionsAgainstFloor:
     def test_scan_finds_the_versioned_sites(self):
         # A regex that silently matched nothing would let every other test in
         # this module pass forever, so pin that the scan has real work to do.
+        # Only the count, deliberately: naming the files would ratchet them in
+        # place, and *deleting* an unenforced per-feature minimum is the very
+        # edit this PR argues for. The actual is 16, across docs/aoi_mask.md,
+        # docs/design/sparse_coverage.md, docs/morton_arrow.md,
+        # docs/ragged_layout.md, src/zagg/configs/atl03_tdigest_healpix.yaml,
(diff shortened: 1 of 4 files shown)


tests/test_mortie_versions.py
```diff
@@ -0,0 +1,179 @@
...
+# The ``aoi_mask`` family: the six sites documenting the runtime gate.
+AOI_MASK_FILES = ("docs/aoi_mask.md", "src/zagg/grids/aoi.py")
+
+# ``mortie`` followed by a release number, in every voice the tree uses:
+# ``mortie >= 0.8.3``, ``mortie ≥ 0.8.4``, ``mortie>=0.8.3``, bare
+# ``mortie 0.8.1``, and the backticked variants (the backticks fall outside
+# the match). Whitespace spans newlines so a wrapped citation is not a blind
+# spot. ``espg/mortie#89`` does not match: an issue reference is not a
+# version, and neither is "frozen for mortie 1.x".
+_CITATION = re.compile(r"mortie\s*(?:>=|≥)?\s*v?(\d+\.\d+(?:\.\d+)*)")
```
espg: 🤖 *from Claude (review)*

**False positives this pattern will produce on ordinary future prose.**

1. **Milestone / freeze references.** `mortie 1.x` doesn't match, but `mortie 1.0` does — and `1.0 > 0.9.3`, so the guard fails. The tree already writes "frozen for mortie 1.x" three times (`docs/ragged_layout.md:145`, `src/zagg/readers/_layout.py:8`, `pyproject.toml:41`), and `docs/design/sparse_coverage.md:28` reads "[mortie #48](…) v1.0 discussion". One word of rewording — `mortie v1.0 discussion` — gives `['1.0']` and a red CI on a sentence that is entirely correct.
2. **Forward references.** This repo's design docs routinely say a feature is *gated on* an unreleased mortie: "Mixed-order reader support is gated on mortie#116" (`docs/design/sparse_coverage.md:392`, and again at `:905`). The moment that is restated with a release number — "gated on mortie 0.10.0" — the guard fires on a statement whose accuracy *depends* on the version being above the floor.
3. **Line wrapping decides coverage.** `\s*` crossing newlines means whether a citation is seen depends on where the paragraph happens to wrap: `_CITATION.findall('the mortie\n\n0.9.9 line') == ['0.9.9']`, while the same words on one line with an intervening verb are invisible. Reflowing a docstring (exactly what commit c1f6ba56 did to `morton.py`) can silently move a citation into or out of scope.

In (1) and (2) the assertion message at L116-118 offers only two remedies — raise the floor (an espg-authorized dependency change, CLAUDE.md §4) or "restate the prose against a shipped release" — and neither applies; the only escape is rewording a correct sentence. Worth an opt-out marker, or excluding `docs/design/` (historical + forward-looking narrative) from the scan roots.

espg: 🤖 *from Claude (review)*

**False negatives — the comment above claims "every voice the tree uses", but the pattern is narrower, and every gap here is a silent pass, which is the one failure mode a drift guard cannot afford:**

```
'Mortie 0.10.0 adds X'                    -> []   # no re.IGNORECASE
'the layer pins mortie==0.10.0'           -> []   # only >= and ≥; ==, ~=, > all miss
'mortie ~= 0.10.0'                        -> []
'mortie (0.10.0)'                         -> []   # yet '(mortie 0.8.5, …)' matches
'mortie: 0.10.0'                          -> []
'mortie, 0.10.0'                          -> []
'mortie-0.10.0'                           -> []
'mortie version 0.10.0'                   -> []
'requires at least mortie release 0.10.0' -> []
```

`mortie==X` deserves a specific mention: it is how a pin gets written up, and the `lambda` extra is nothing but `==` pins (`pyproject.toml:80-95`), so "the layer pins `mortie==0.10.0`" in a doc sails straight through. Sentence-initial `Mortie` is the other one likely to show up in narrative prose.

`re.IGNORECASE` plus widening the separator (something like `mortie[\s(:,-]*(?:[><~=!]=|≥)?\s*v?`) closes most of it without touching the `espg/mortie#89` exclusion, which is carried by `#` not being a digit.

tests/test_mortie_versions.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def _cited_versions(text):
+    """Every mortie release ``text`` names, as ``(Version, line number)``."""
+    return [
+        (Version(m.group(1)), text.count("\n", 0, m.start()) + 1) for m in _CITATION.finditer(text)
```
espg: 🤖 *from Claude (review)*

The reported line is where the word `mortie` sits, not where the version sits, so every citation the newline-spanning `\s*` stitches together is reported one line early. Both of the wrapped-sentence catches the PR body highlights are off by one:

- `src/zagg/readers/_layout.py:6` is `…the pure bit deinterleave that mortie` — `0.9.3` is on line 7.
- `src/zagg/grids/morton.py:121` ends `…``decimal_to_word``, mortie` — `0.9.1+` is on line 122.

So the PR-body table lists two sites at lines that contain no version, and a real failure would do the same. `text.count("\n", 0, m.start(1)) + 1` names the version's own line.

tests/test_mortie_versions.py
```diff
@@ -0,0 +1,179 @@
...
+def _above(sites, floor):
+    """The comparison the guard makes — PEP 440 ordering, never string compare."""
+    return [s for s in sites if s[2] > floor]
+
+
+def _mortie_floor():
+    """The ``>=`` floor from ``[project.dependencies]``."""
+    deps = tomllib.loads((REPO_ROOT / "pyproject.toml").read_text(encoding="utf-8"))
+    reqs = [Requirement(d) for d in deps["project"]["dependencies"]]
+    spec = next(r.specifier for r in reqs if r.name == "mortie")
```
espg: 🤖 *from Claude (review)*

Three gaps in the floor parse:

- **First match wins, silently.** `next(...)` takes the first `mortie` requirement. A marker-split pair (`"mortie>=0.9.3; python_version<'3.14'"` + `"mortie>=1.0; python_version>='3.14'"`) would be measured against the first entry alone and half the matrix goes unguarded. Collecting every mortie requirement and asserting exactly one fails loudly instead — and matches the `len(lower) == 1` style already used one line down.
- **Name compare is not normalized.** `r.name == "mortie"` is case- and separator-exact; a legal `Mortie>=0.9.3` raises a bare `StopIteration` out of `next()` with no diagnostic at all. `packaging.utils.canonicalize_name` is the standard fix.
- **`Version(s.version)` assumes a concrete release.** L88 explicitly admits `==`, but a wildcard pin (`mortie==0.9.*`) raises `InvalidVersion` here rather than reaching the assertion message on L89.

Nit: `deps` on L85 is bound to the whole parsed document, not the dependency list.

tests/test_mortie_versions.py
```diff
@@ -0,0 +1,179 @@
...
+        files = {path for path, _, _ in sites}
+        for expected in (
+            "docs/aoi_mask.md",
+            "docs/morton_arrow.md",
+            "docs/ragged_layout.md",
+            "src/zagg/grids/aoi.py",
+            "src/zagg/grids/morton.py",
+            "src/zagg/grids/healpix.py",
+        ):
+            assert expected in files, f"{expected} no longer reached by the scan"
```
espg: 🤖 *from Claude (review)*

`len(sites) >= 12` on L100 is the real anti-vacuity guard; this per-file list is a ratchet, and it points against the direction phase 1 is moving.

Phase 1's thesis is that unenforced per-feature minimums are noise. The natural follow-through is deleting some of them, and two of these files hang by a single citation each in exactly that category: `docs/morton_arrow.md` has one (the table cell at L11, rewritten by this PR), and both `src/zagg/grids/healpix.py` hits (`:37`, `:355`) are incidental provenance in comments. Removing one is a *correct* edit that fails here with `"… no longer reached by the scan"` — a message that reads as regex blindness and sends the next reader hunting a bug in `_CITATION` that isn't there.

Suggest dropping the file list (or moving it to a comment) and keeping the count — raise it if you want it tighter; the actual is 16.

tests/test_mortie_versions.py
```diff
@@ -0,0 +1,179 @@
...
+            f"zagg.grids.aoi.MIN_MORTIE_VERSION ({MIN_MORTIE_VERSION}) is above the "
+            f"pyproject floor ({floor}) — aoi_mask would raise on a conforming install"
+        )
+
+    def test_aoi_mask_prose_matches_the_runtime_gate(self):
+        # These six sites document a gate that raises (``aoi.py`` lines 66-80),
+        # so they keep requirement voice — and must quote the constant itself.
+        family = [s for s in _scan() if s[0] in AOI_MASK_FILES]
+        assert len(family) >= 6, f"aoi_mask citations vanished from the scan: {family}"
+        wrong = [s for s in family if s[2] != Version(MIN_MORTIE_VERSION)]
```
espg, line 170: 🤖 *from Claude (review)*

The class docstring and the PR body both say the `aoi_mask` family "must quote the enforced constant exactly", but the scan reaches only 6 of the ~9 version statements in these two files — and the three it misses are the ones closest to the gate:

- `src/zagg/grids/aoi.py:35` — "espg/mortie#59 + #70, **shipped in 0.8.2**"
- `src/zagg/grids/aoi.py:37` — "``from_geometry``, espg/mortie#89) **ship in 0.8.3**"
- `src/zagg/grids/aoi.py:54` — inside `_assert_mortie_version`'s own docstring: "which **ship in the same 0.8.3 release**"

So bumping `MIN_MORTIE_VERSION` leaves three stale numbers inside the docstring of the function that enforces it, and this test stays green — which is the exact drift the invariant is written to catch, at the exact site it claims to pin.

The file-scoped equality rule also discourages fixing it: rewrite L37 to attached voice (`ship in mortie 0.8.3`) and it passes, but rewrite L35 the same way (`mortie 0.8.2 shipped the order-29 cap`) and the test fails on a factually correct historical note. Two ways out — scope the equality rule to *requirement* voice (`>=` / `≥` citations only) so arrival-voice history in the same file is allowed, or add a targeted assertion that every version literal in `aoi.py`'s gate block equals `MIN_MORTIE_VERSION`. Either way the class docstring's claim should be narrowed to what the scan can actually see.

tests/test_mortie_versions.py
```diff
@@ -0,0 +1,179 @@
...
+            "0.8.3",
+            "0.8.4",
+            "0.9.3",
+            "0.8.1",
+            "0.9.0",
+        ]
+
+    def test_the_floor_line_is_not_scanned_as_a_citation(self):
+        # The regex would happily match ``mortie>=0.9.3`` in pyproject.toml, so
+        # the exclusion has to come from the scan roots — not from luck.
```
espg: 🤖 *from Claude (review)*

This does not show what the comment says, and the PR body repeats the same wrong mechanism ("the exclusion demonstrably comes from the scan roots and not from luck").

`pyproject.toml` is excluded by **`SCAN_SUFFIXES`** (L44 — no `.toml`), not by `SCAN_ROOTS`. Add `REPO_ROOT` to `SCAN_ROOTS` and this test still passes; the L179 assertion cannot fail for *any* value of `SCAN_ROOTS`, only if someone adds `".toml"` to the suffix set — and no `.md`/`.py`/`.yaml` file can ever end in `pyproject.toml`, so the `endswith` check is a tautology on both counts.

If the intent is "the floor line is never read back as a citation", assert the mechanism that enforces it: `assert ".toml" not in SCAN_SUFFIXES` alongside the L177 demonstration that the regex would otherwise match it.

tests/test_mortie_versions.py
espg: 🤖 *from Claude*

**Declined — left standing.** No code change; the rationale, so it can be argued with:

- (1) and (2) are the guard working, not misfiring. A sentence in `docs/` or `src/` that names a mortie release *above* the floor is precisely the thing a human should look at before it ships — a forward reference ("gated on mortie 0.10.0") is a promise the installed dependency does not keep, and the reader of that doc has no way to tell it apart from an arrival note. Red CI is the intended cost of writing one, and the resolution is the same either way: raise the floor, or restate it in a form that isn't a version promise (`mortie#116`, `mortie 1.x` — both already excluded).
- The two proposed remedies both narrow the directive rather than the false-positive set. An opt-out marker is a silencer the next author reaches for by reflex, and excluding `docs/design/` drops 1 of the 16 live sites and the whole directory the sparse-coverage work is documented in — the directive is "anywhere in `docs/` or `src/`", and carving out the largest narrative subtree is an espg-sized scope call, not a fold-sized one.
- Point (3) (wrap-dependent coverage) is real and unfixable by regex — it is a false *negative*, so it never reddens CI on correct prose. Rather than argue it away it is now written into the module docstring's scope note alongside the unattached-literal blind spot (4cf0c51c), so the next reader knows a `MIN_MORTIE_VERSION` bump still wants a human sweep.

If you want the escape hatch, the cheap version is a one-line allowance in the assertion message pointing at a `# noqa`-style marker — say the word and it lands in a follow-up rather than being invented here.

espg: 🤖 *from Claude*

**Accepted** — fixed in ac5d864f. A silent pass is the one failure this guard cannot afford, so the pattern is now:

```python
_CITATION = re.compile(
    r"mortie[\s(:,_-]*(?:version|release)?\s*(>=|==|~=|>|≥)?\s*v?(\d+\.\d+(?:\.\d+)*)",
    re.IGNORECASE,
)
```

Every spelling in your list now matches — `Mortie 0.10.0`, `mortie==0.10.0`, `~=`, `>`, `(0.10.0)`, `: `, `, `, `-`, `version`/`release`, `v0.10.0` — pinned as a loop in a new `test_the_spellings_that_used_to_pass_silently_are_caught`.

Two deliberate non-inclusions, both now regression-tested in `test_issue_references_are_not_versions`:

- `<` / `<=` stay out. `mortie < 1.0` is an upper bound: it promises no API, so measuring it against the floor would fail on a correct sentence — a false positive introduced while chasing false negatives.
- `espg/mortie#89`, `mortie#116` and `mortie 1.x` still don't match (`#` is outside the separator class; `1.x` has no second numeric component).

Re-ran the scan after widening: the same 16 sites, no new matches anywhere in `docs/` or `src/` — the widening is pure coverage, not new surface.

tests/test_mortie_versions.py
espg: 🤖 *from Claude*

**Accepted** — plain bug, fixed in 0f9d8798. `m.start()` → `m.start(1)` (now `m.start(2)`, after the operator capture group landed in 4cf0c51c), plus a regression test:

```python
assert _cited_versions("added in mortie\n0.9.0") == [(Version("0.9.0"), 2)]
```

Both wrapped sites move as you predicted: `src/zagg/readers/_layout.py:6 → :7` and `src/zagg/grids/morton.py:121 → :122`. The PR body's table is being corrected to match.

tests/test_mortie_versions.py
espg: 🤖 *from Claude*

**Accepted, all three** — fixed in 01ca19f1.

```python
mortie = [r for r in reqs if canonicalize_name(r.name) == "mortie"]
assert len(mortie) == 1, f"expected exactly one mortie requirement in pyproject, got {mortie}"
lower = [
    Version(s.version)
    for s in mortie[0].specifier
    if s.operator in (">=", "==", "~=") and "*" not in s.version
]
```

- First-match: `next()` is gone; every mortie requirement is collected and `len(...) == 1` asserted, so a marker-split pair fails loudly instead of guarding half the matrix.
- Name compare: `packaging.utils.canonicalize_name`, so `Mortie>=0.9.3` is found rather than raising a bare `StopIteration`.
- Wildcard pin: took it. `"*" not in s.version` keeps `mortie==0.9.*` out of `Version()`, so it reaches the assertion below — which now names the offending requirement (`got [] from mortie==0.9.*`) instead of surfacing `InvalidVersion` from a line with no context.
- Nit taken: `deps` → `pyproject`.

tests/test_mortie_versions.py
espg: 🤖 *from Claude*

**Accepted** — fixed in 6f989ed9. The per-file loop is gone; `len(sites) >= 12` (the anti-vacuity assertion the directive actually asked for) stays, and the file list moved into a comment that says why it is a comment:

> Only the count, deliberately: naming the files would ratchet them in place, and *deleting* an unenforced per-feature minimum is the very edit this PR argues for. The actual is 16, across …

You're right that it pointed against the thesis — `docs/morton_arrow.md` and both `healpix.py` hits are exactly the citations phase 1 says are noise, and failing a correct deletion with "no longer reached by the scan" would have sent the next reader hunting a regex bug that isn't there. Left the threshold at 12 rather than tightening to 16 for the same reason.

tests/test_mortie_versions.py
espg: 🤖 *from Claude*

**Accepted — your option (a)** — fixed in 4cf0c51c. The equality rule is now scoped to *requirement* voice, so arrival-voice history in the same file is held to the floor like any other citation and not to the constant:

```python
family = [s for s in _scan(requirement_voice_only=True) if s[0] in AOI_MASK_FILES]
```

The pattern captures the comparison operator as group 1 — present only when the citation states an obligation, `None` for "added in mortie 0.8.4" — and `test_requirement_voice_is_told_from_arrival_voice` pins the split. All six `aoi_mask` sites carry `>=`, so the family is unchanged at 6/6; the six sites themselves are untouched, since they document a gate that really raises.

That also removes the trap you identified: `aoi.py:35`'s "shipped in 0.8.2" can now be rewritten to attached voice (`mortie 0.8.2 shipped the order-29 cap`) without failing on a factually correct historical note.

The module docstring's claim is narrowed to what the scan can see, and the residual blind spot is written down rather than implied — the three unattached literals at `aoi.py:35,37,54` still need a human sweep on a `MIN_MORTIE_VERSION` bump, and the scope note now says so by name (along with the wrap-sensitivity from the sibling thread). A targeted "every literal in the gate block equals the constant" assertion was the alternative and was rejected for the reason you gave: it reddens on `:35`, which is correct history at a version below the gate.

tests/test_mortie_versions.py
espg: 🤖 *from Claude*

**Accepted** — fixed in 01bb10b2. You're right that `endswith("pyproject.toml")` could not fail for any scanned suffix. The test now asserts both mechanisms that do the excluding, keeping the demonstration that the regex would otherwise match the floor line:

```python
assert [v for v, _ in _cited_versions(f'    "mortie>={floor}",')] == [floor]
assert ".toml" not in SCAN_SUFFIXES
pyproject = REPO_ROOT / "pyproject.toml"
assert not any(pyproject.is_relative_to(root) for root in SCAN_ROOTS)
```

Kept the roots check too — as `is_relative_to` rather than a string suffix, so it now bites if `REPO_ROOT` is ever added to `SCAN_ROOTS` — but the suffix assertion is the one carrying the guarantee, and the comment says so. The PR body's "comes from the scan roots" claim is being corrected alongside.

Issue #399, opened 2026-08-07
raster hive under a /2 pyramid declaration: leaf columns are never written (rule PR #391 options (a)/(b))


espg:
🤖 **Addendum 2026-09-10 (relabel manifest, espg-approved):** ruling stands — option (b), mirror-base-aggregation. This issue is now the raster-overview umbrella at `plan`: the next unit of work is a plan answering these questions from the design comment, not more discussion.

- [ ] **(1) Time-axis alignment across children** — options: (i) stage workers union child time axes and reindex; (ii) raster columns fold time away; (iii) raster `/2` **requires** `output.windowing` so each `(leaf, window)` artifact aligns by construction. Lean: (iii).
- [ ] **(2) Where does a band's fold law get declared?** — (i) per-band `overview_fold: nn|sum|mean|mode` under `data_source.bands`, default `nn`; (ii) synthesize all-`nn`, no new key; (iii) mirror `aggregation.variables`. Lean: (i).
- [ ] **(3) Is `nn` in the shared fold vocabulary?** — joins `EXACT_MERGE_LAWS` with a subsample arm, or raster gets its own kernel; either way `nn` needs a defined child-selection rule identical leaf-side and sweep-side.
- [ ] **(4) Module home** — raster.py/sweep_overview.py/sweep_stage.py are all at/past the line cap; lean: one new `src/zagg/raster_overview.py` with both kernels.
- [ ] **(5) Fixture input** for a raster `column/` conformance artifact — committed vs synthetic COG.

---

🤖 *from Claude* (filed at espg's direction, 2026-08-07, from the PR #391 review)

## The asymmetry

`runner.py:866-875` builds the raster hive manifest through the same `hive.build_manifest` → `sweep_overview.build_pyramid_block` as aggregation stores, so a raster hive config carrying `output.pyramid.overviews` records a full `zagg-pyramid/2` block (`_validate_pyramid` only requires `store_layout: hive`). But `processing/raster.py::process_and_write_raster_hive` — the function both dispatchers call for raster units — has no column hook, so after PR #391 a `/2` declaration means "leaf columns exist" for aggregation stores and "no columns" for raster stores.

Per #381 point (11) this is a *legal* recorded state — the declaration records intent, materialization is discovered from the artifacts, and spec §4.6 spells out the reader posture ("read the columns, not the block"). This issue exists so the asymmetry is **named and owned** rather than silent, per the standing "raster is never scoped out of store/schema changes" ruling.

## The decision (from the review thread)

[...]

PR #389, opened 2026-08-04
pyramid v2: (node, cells) level grammar and zagg-pyramid/2 (issue #382)

espg:
Closes #382. Refs #381 (design record, points (2), (4), (5), (11)).

## What

The `zagg-pyramid/2` overviews grammar, in its final ruled (collapsed) form: `output.pyramid.overviews` declares **leaf cell resolutions only** — a scalar (sugar for one) or a strictly descending list of ints, each strictly between `parent_order` and `child_order`; omitted, the default is the grid's resolved chunk order; `pyramid: false` stays the off switch. Everything above the shard is **law, not declaration** — the fixed every-order ladder: with `d = base − parent_order` (`base` the coarsest leaf resolution), every order from `parent_order − 1` down to 0 inclusive carries one member at `order + d`. The manifest records the **fully expanded** `(node, cells)` list at block level (`pyramid.overviews`) so readers never re-derive; slab length per member is `4^(cells − node)`; there is no `partial/` grammar, no gather declaration, and no member promotion (the ladder makes every within-footprint member spec-guaranteed). A `/2` store is inherently a multiresolution statistical grid — every HEALPix order from 0 through the base, plus native — with the ruled accuracy doctrine: exact-class fields exactly correct at every order, approximate-class fields analysis-grade at their recorded generation. Grammar, validation, expansion, declaration, and the spec revision only — **no writer or sweep materialization** (that is #383/#384): a `/2` declaration is loudly declared-but-not-yet-sweepable, and `/1` stores sweep exactly as before.

New grammar code lives in a new module `src/zagg/pyramid.py` rather than growing `sweep_overview.py` (already at 1,227 lines on `main`, past the §4 ceiling).

## Approach

- **Normalization** (`normalize_overviews`): leaf resolutions — scalar sugar, int/bool shape refusals by name, and a pointed refusal for the deleted `{node, cells}` pair spelling.
- **Validation** (`validate_overviews`): strictly descending; each resolution strictly between `parent_order` and `child_order`. Loud refusals, never silent widening — the PR #369 precedent.
- **Expansion** (`expand_overviews`): the leaf entry plus the fixed every-order ladder to node 0 — what the manifest records; `default_overviews` is the omitted-knob case (`[chunk_order]`), validated before expansion so the two degenerate geometries (K == 1, chunk == child) refuse loudly. For the 19/13/9 reference geometry the default is (9,[13]) (8,[12]) … (0,[4]).
[...]

PR #416, opened 2026-08-09
pyramid v2: staged dense sweep on the partition machinery (issue #384)

### LINKED ISSUE #384: pyramid v2: staged dense sweep on the partition machinery
🤖 *from Claude* (child of #381, point (6); filed at espg's direction)

## Scope

The staged dense sweep that materializes everything above the shard node from leaf columns, built on PR #380's partition machinery. Under #381's residency law, all of it is bounded-memory gathers and resident merges — no raw-leaf reads.

## Shape

A sweep stage is a set of `(node, depth)` work units; PR #380's 2^n partitions are the stage-0 degenerate case (same prefix-ownership predicate, same disjointness proofs — one orchestration schema, not two).

Under the ruled every-order ladder (option D, 2026-08-05) the default cadence is **3-order tuples** — orchestration only, never store shape (grouping changes no bytes):

- **tuple 1, dispatch at o6** — one worker per dirty o6 node: reads its 64 leaf columns; emits the (8,·), (7,·), (6,·) slabs — **all pure gathers** of leaf members (every level at node ≥ 5 is a gather; the ladder's first true k-way merge appears at node 4) — plus its own column members.
- **tuple 2, dispatch at o3** — same shape over the o6 columns: (5,·) gathers, (4,·)/(3,·) resident merges (merges-from-raw 2).
- **tuple 3, dispatch at the 12 base cells** — (2,·), (1,·), (0,·) from the o3 columns (generations ≤ 3 at the root); this tuple also owns the finisher-inherited duties (root `coverage.moc`, manifest RMW).

Parents not ≡ 0 mod 3 get a ragged last tuple; the tuple width is a dispatch knob defaulting to 3 (fan-in `4^3 = 64` columns per worker).

Inter-stage barriers are **scheduling preferences, not correctness**: the #379 ratchet (fold what is on disk, record `source_children`, rewrite when the summed generation moves) makes early or partial stage runs self-healing. The dispatcher chains stages in the immediate post-fleet path (D8: orchestrate, never write); the deferred path is the existing sweep ratchet.

PR #380's finisher deferrals — root `coverage.moc` refresh, manifest `pyramid.materialized` RMW — belong to the last stage, unchanged. Per-field streaming keeps a stage worker's resident set bounded at any subtree depth.

PR #379's cascade survives as the **retrofit regime** for stores whose leaves predate columns (and for finer-than-base level appends, which are a leaf-reading partitioned sweep with a wholesale column rewrite).

## Scope argument (the sweep API)

A sweep takes **no work-spec** — discovery is store-side (root `coverage.moc` + generation stamps), and the ratchet makes first-sweep and update the same operation. It takes an optional **scope**: canonically a set of node prefixes (a MOC — the same ownership predicate as the partitions), with a **shardmap accepted as sugar** (its keys coarsen to ancestor prefixes). The post-fleet immediate path auto-scopes to the fleet's shardmap footprint (the dispatcher has it in hand); unscoped sweeps the whole store; scope composes with partitions by intersection. Stage workers read **all** children on disk within scope, so an update adjacent to prior data (the NEON per-domain case) folds old neighbors in automatically, and nodes whose summed generation did not move no-op on skip-if-current.

## Acceptance

- disjointness pinned against the real writers per stage (PR #380's spy pattern);
- a partially-run stage under-covers loudly (`source_children`) and self-heals on re-sweep, end to end through `run_sweep`;
- merges-from-raw recorded per level and never exceeding 2 under the default schedule;
- per-partition/stage rows in the sweep run record (#354 shape).

Depends on: the grammar and leaf-column children of #381; PRs #379 and #380 merged. Refs: #381, #377, #376, #354.

espg:
Closes #384. Refs #381 (umbrella), #383/PR #391 (leaf columns), PR #380 (partition machinery), PR #389 (`/2` grammar), PR #397 (lifecycle seam + run-record shape).

Implements the staged dense sweep for `zagg-pyramid/2` stores per the implementation plan on #384 as amended by the two plan deltas and governed by the four rulings recorded there ([merge-source law / finisher / stale-MOC discovery](https://github.com/englacial/zagg/issues/384#issuecomment-5230157470), [sweep-admission lease + concurrency matrix](https://github.com/englacial/zagg/issues/384)):

- **merge-source law (Rule A)**: every stage worker reads exactly its `4^width` immediate child columns; its own column relays the subtree's gen-1 leaf node-order partials as a pure gather; every merge at every level consumes only the relayed gen-1 tier — never a merged tier — so the merge tree is grouping-independent and `tuple_width=1` / `tuple_width=3` builds are byte-identical (the single test that pins the law).
- **finisher-worker**: the root singletons (root `coverage.moc` refresh, manifest RMW with per-entry actuals, lease release) belong to one designated finisher fired after the root tuple, never to the 12 base cells.
- **stale-MOC discovery**: unscoped discovery is listing-based (run records); the root `coverage.moc` is an accelerator only.
- **sweep-admission lease**: atomic per-store intent object (conditional PUT `If-None-Match: *`), run id + scope + heartbeat + TTL; live intent refuses naming the runner; expired heartbeat is claimable; stamps carry the run id and a foreign-fresh stamp aborts loudly (the residual-race backstop). Control-plane framing: no data object is ever locked.

## Phases

- [x] Phase 1 — stage planner (`src/zagg/sweep_stage.py`, pure functions): tuple grouping (`tuple_width` knob, dispatch at nodes ≡ 0 mod width, ragged finest tuple), derived gather/merge classification (gather iff `cells >= shard_order`), column-member derivation (relay + coarser tuples' gatherables), scope normalization (node-prefix MOC; shardmap keys as sugar), scope ∩ partition composition via mortie's scalar `moc_and` (TODO markers reference espg/mortie#173 / PR 174 for the batch `mocs_and` swap at mortie 0.9.6 — no unreleased-mortie dependency).
- [x] Phase 2 — stage worker: per dirty node at the dispatch order, per-field streaming reads of the `4^width` child columns with optimistic stamp validation (pre/post-read check, re-read on movement), level slabs + the worker's own column (relay as pure gather), `/2` per-artifact attrs (`zagg-overview/2`: regime, merges-from-raw, `source_children`), stamps carrying the run id, skip-if-current keyed on summed child generations, foreign-fresh-stamp abort, soft barriers (partial tuples under-cover loudly, self-heal).
- [x] Phase 3 — designated finisher-worker: root `coverage.moc` refresh sequenced before any scoped fan-out reads it, manifest RMW writing per-entry actuals into `pyramid.overviews` entries (composing with, not duplicating, the PR #397 root-touch — the RMW refreshes the manifest's `LastModified`; `aggregation.yaml` gets the lifecycle touch), the `/2` default flip in `build_pyramid_block` (+ worker-gate symmetry + fixture regen), lease release as the final act.
- [x] Phase 4 — dispatch orchestration: `run_stage_sweep` (in-process), the sweep-admission lease (`src/zagg/sweep_lease.py`), `partitions=` composition, CLI backstop (`python -m zagg.sweep --stages`), post-fleet chaining opt-in (`output.sweep: "stages"`), per-stage + lease intent/completion rows extending the #397 run-record shape, listing-based unscoped discovery.
- [x] Phase 5 — spec + docs: §4.5 per-entry actuals grammar (additive), §4.4 `/2` per-artifact attrs + stamp `run_id` (additive), the lease object's byte shape, the retrofit-regime statement (#379's cascade remains the `/1`/pre-column path), the concurrency matrix (fleet∥fleet / fleet∥sweep / sweep∥sweep) + control-plane framing in the operator docs, the sweep-written ancestor-overview aging note (owned here per plan delta 2), conformance fixtures through the generator, raster stores column-less by construction with a clean seam for #399.

## Module layout

The run machinery split at its natural seams to respect the §4 module cap: `src/zagg/sweep_stage.py` (planner + one stage worker's fold kernels and writers, ~1,185 lines), `src/zagg/sweep_stages.py` (the run: scope planning, pass driver, designated finisher, orchestration, record, chaining, ~585 lines), `src/zagg/sweep_lease.py` (admission, ~220 lines). After the review folds pushed the worker module past 1,200, the four pure scope/partition-composition helpers moved to the run module (their only consumer) — disclosed here rather than raised as a blocking question since both files sit comfortably under the cap; veto welcome. Tests: `tests/test_sweep_stage.py` (planner/worker/finisher/orchestration battery), `tests/test_sweep_lease.py`.

## Acceptance battery (lands with phases 2–4)

Disjointness pinned per tuple via the #380 spy pattern; byte-identity `tuple_width=1` vs `=3`; gather levels carry gen-1 content untouched; per-entry merges-from-raw == 2 for every upfront merge level (gathers record 1; 3 is append-later cascade territory only); partial-tuple under-coverage loud + self-healing end-to-end; scoped NEON-adjacent update (ancestor prefixes only, old neighbors fold in, clean nodes no-op); stale-root-MOC discovery; lease contention (refusal names the runner), expired-heartbeat claim completing a partial prior run, foreign-stamp abort, stamp-validation re-read, fleet-append-during-live-sweep; `/2` fully sweepable with the declared-not-yet-sweepable gate retired and `/1` unchanged.

## Key mechanics for review

- **Byte-identity test** (`TestMergeSourceLaw::test_byte_identity_across_tuple_widths`): builds the same fixture store twice, sweeps one at `tuple_width=1` and one at `tuple_width=3`, and compares every ladder artifact's `morton`/exact/digest array contents element-for-element (digest payload bytes compared exactly). It bites because a width-1 build merges through TWO relay hops (stage columns at orders 2 and 1) while width-3 merges straight from the leaf columns — under rejected reading (B) the digest bytes would differ; under Rule A they cannot. A companion spy test (`test_merge_consumes_only_the_gen1_tier`) asserts stage columns are only ever read at the relay resolution.
- **Lease acceptance**: contention (second sweep refused naming the runner), expired-heartbeat claim that completes a partial prior run end-to-end (run A sweeps one base cell scoped, dies holding the lease; run C claims — `claimed_from: A` — and finishes the rest), foreign-fresh-stamp abort via an injected zombie stamp, stamp-validation re-read (a reader hook rewrites the leaf column mid-read; the pre/post stamp check catches it and the merge consumes the coherent post-rewrite column), and fleet-append-during-live-sweep (missing candidate column recorded as under-coverage, healed by the next pass).

The fresh-context review posted six inline findings; all six are folded with a reply + fix sha on each thread: per-node lease heartbeat (7a994e37), missing-member tolerance + real `unreadable` counting (f5de4dd2), foreign-fresh guard re-run at the validation re-read (f5de4dd2 + 7d5536ce), per-artifact actuals assignment summed once — no partition inflation (bebfc70d), claim-race pre-delete re-verify (fd068e5b), and windowed/all-time arm coverage (663eafd9). One scoped-actuals semantic (a scoped run's `actuals` record that run's observation, named by `run_id`) is left standing as designed and flagged for espg on the thread.

## How tested

`uv run pytest -v` (full suite), `uv run ruff check src tests`, `uv run ruff format --check src tests` — per-phase results recorded in the phase commits' CI runs. Baseline note: `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` fails on this machine on a clean `main` checkout (docker/lambda build environment; pre-existing, unrelated — flagged, not fixed).

## Questions for review

1. **(a) family `materialized` on `/2` stores** — implemented lean: with per-entry actuals landing in `pyramid.overviews` entries, the staged sweep does NOT write the family dict's order-keyed `materialized` on `/2` stores (one source of truth per fact; the `/1`-era inventory is preserved verbatim where it exists). Flagging per the plan.
2. **(b) stage transport** — in-process orchestration only in this PR; the Lambda `RequestResponse` forward and the `deployment/aws/` handler change stay espg's call (also §1: this PR does not touch `deployment/aws/`). The worker-event `stage` block is therefore not yet emitted anywhere; raised as a question, not committed.
3. **(c) post-fleet chaining default** — implemented lean: opt-in via `output.sweep: "stages"`; the default stays the families sweep. Flip-to-default is espg's call once operationally proven.
4. **`/2` default-flip breadth** — the ruled "one line" in `build_pyramid_block` required two symmetries to be correct: the worker column gate (`leaf_column_plan`) must derive the same default resolution (the grid's resolved chunk order), or a default-flipped store would declare leaf levels no worker writes; and grids with no strictly-interior chunk order (K == 1) have no valid `/2` default, so they keep the `/1` fallback. Both spelled out in phase 3; please confirm the K == 1 posture.
5. **Byte-identity scope** — a `tuple_width=1` build materializes stage columns at every order (each tuple needs its child relay), where `tuple_width=3` materializes them at dispatch orders only. The byte-identity test therefore pins every ladder artifact (the reader-facing product, which is what #381 point (6) governs); column *placement* is cadence, documented as orchestration-never-contract in spec §4.6. Flagged for the record.
6. **Skip-gate resolution** — the ruled skip key is the summed child generation pair (`n_leaves` + max stamp timestamp, 1 s resolution). A child rewritten within the same second with the same leaf count reads current; the `/1` content-hash backstop cannot apply without folding (which is the cost the skip exists to avoid). Inherited tradeoff, flagged.
7. **`bench_objects` model** — the CI-owned `.github/scripts/bench_objects.py` object-count model has no term for the leaf column the default now writes; `tests/test_benchmark_objects.py::test_hive_sharded_store_matches_model` pins its audited posture with an explicit `pyramid: false` and the model's column term is left for espg (this PR does not touch `.github/`).
8. **Fixture regen drift** — regenerating `tests/data/spec/` also surfaced `granules.json` sidecars (issue #388 machinery) absent from the committed fixtures: pre-existing generator drift, included here since the generator is the source of truth.

- `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` fails on this machine on a clean `main` checkout (lambda build environment) — unrelated.
- `ruff check` flags a pre-existing `N818` on `zagg.registry.UnknownCapability` (the CI lint bot runs without `N`); this PR's own exception names carry the `Error` suffix.


src/zagg/sweep_stages.py
```diff
@@ -0,0 +1,486 @@
...
+            part = sweep_stage_pass(
+                store_root,
+                manifest,
+                by_shard,
+                scope=part_scope,
+                tuple_width=tuple_width,
+                run_id=run_id,
+                run_started=run_started,
+                store_kwargs=store_kwargs,
+                on_stage=lambda row: heartbeat_lease(store_root, lease, store_kwargs=store_kwargs),
```
espg, line 490: 🤖 *from Claude (review)*

**The lease heartbeat fires once per TUPLE, so on any real store the holder's lease expires while it is still writing — the chimera the lease exists to prevent becomes the common case, not a residual race.**

`on_stage` is only invoked from the tuple loop (`sweep_stages.py:154-155`), so between two heartbeats a whole tuple runs. On the o9/d=4 reference store at the default `tuple_width=3`, the finest tuple (`dispatch=6`) is `12·4^6 = 49,152` dispatch nodes, each opening 64 child columns and folding three levels. That is hours against `DEFAULT_TTL_S = 900` (`sweep_lease.py:49`). The lease is therefore expired — and claimable — for essentially the entire first tuple. `run_finisher` also runs with no heartbeat at all (`sweep_stages.py:404`).

Confirmed against this branch (`lease_ttl_s=2`, one dispatch node made to take 3 s):

```
mid-run sibling admission -> ('SIBLING', 'stage-20260809T080155Z-bd7cb9')
run raised: SweepRefusedError sweep lease ... is no longer held by run
  'stage-20260809T080155Z-...' (found 'SIBLING') — another run claimed an
  expired heartbeat; aborting
```

The sibling was admitted with `claimed_from` naming the live holder, and the holder only noticed at the *next tuple boundary* — after both had been writing stage artifacts and columns. The `_foreign_fresh` backstop (`sweep_stage.py:289`) does not close this: it only fires when a run *reads* an artifact the other already stamped, so the interleaved PUTs that produce a chimera column happen first.

Repro: `run_stage_sweep(root, leaves, lease_ttl_s=2)` with any `stage_node` that sleeps past the TTL, calling `acquire_lease(root, run_id="SIBLING")` from inside.

The heartbeat needs a cadence bounded by wall time, not by tuple count — e.g. beat from the per-node loop in `sweep_stage_pass` (or from `stage_node`'s completion) whenever `ttl_s/3` has elapsed since the last beat, and beat once more before `run_finisher`. The per-tuple `on_stage` row is the wrong clock.

src/zagg/sweep_stage.py
```diff
@@ -0,0 +1,1186 @@
...
+        """One group array, stamp-validated; ``None`` for an absent field."""
+        import zarr
+
+        for _attempt in range(3):
+            before = self.stamp
+            try:
+                group = zarr.open_group(self._store, path=str(res), mode="r", zarr_format=3)
+                values = group[name][:]
+            except KeyError:
+                return None  # schema evolution: the field postdates this column
```
espg: 🤖 *from Claude (review)*

**A single column short one resolution group — or one transient chunk read error — aborts the entire staged sweep, in direct contradiction of the module's soft-barrier posture.**

`_ColumnReader.read` only guards `KeyError`, which covers a missing *array inside an existing group*. A missing **resolution group** raises `zarr.errors.GroupNotFoundError`, and nothing between here and `run_stage_sweep` catches it: `_gather_slabs` / `_merge_slabs` / `_stage_fold` have no `try`, and `stage_node`'s `try` wraps only `_write_stage_overview` (line 1011) and `write_stage_column` (line 1098) — never the fold.

Confirmed against this branch (fixture store from `tests/test_sweep_stage.py`, one leaf column's `3/` group removed):

```
RAISED GroupNotFoundError No group found in store
  LocalStore('.../1/1/1/1/all.pyramid.zarr') at path '3'
```

The whole pass dies; nothing is written, no `failed` count, no under-coverage record.

Reachable ways to get there:

1. **Deepening the declaration.** `overviews: [4]` on an o3 store gives leaf columns carrying `{4,3}`; re-declaring `overviews: [6]` makes the sweep gather at cells 5 (`column_resolutions` would now be `{6,5,4,3}`). Every leaf column written before the change is stamped, committed, and short the `5/` group — the first sweep after the re-declaration takes the whole run down instead of under-covering the stale leaves and healing as the fleet rewrites them.
2. **Any transient object-store fault on a chunk GET** during the fold. At 49k dispatch nodes × 64 columns × fields, one 5xx is a certainty per run.

This is also inconsistent within the module: `_readers_for` (line 419) deliberately counts an unreadable column as `failed` and yields `None` so the fold proceeds — the fold's own reads get none of that treatment. Suggest widening the guard to the group-absent case and wrapping the per-`(target, level)` fold in the same `try` that already protects the write, counting `failed` + `source_children.unreadable` (which is currently hardcoded to `0` everywhere).

src/zagg/sweep_stages.py
```diff
@@ -0,0 +1,486 @@
...
+                level_actuals=level_actuals,
```
espg, line 492: 🤖 *from Claude (review)*

**`level_actuals` is shared across partition passes, so `partitions=N` multiplies every level's `source_children` by N in the manifest.**

The same dict is threaded into every partition's `sweep_stage_pass` (line 392), and `_accumulate_actuals` (`sweep_stage.py:1185-1186`) **adds** `source_children` for *both* the write path and the skip path (`sweep_stage.py:984-992`). Coarse dispatch nodes intersect every partition MOC — the test at `tests/test_sweep_stage.py:846-847` says so explicitly ("later partitions re-visit them and read current") — so each re-visit accumulates the *same* artifact's counts again.

Confirmed against this branch, identical fixture store swept twice:

```
partitions=None: {"2": {"folded": 4, ...}, "1": {"folded": 4, ...}, "0": {"folded": 4, ...}}
partitions=4   : {"2": {"folded": 16,...}, "1": {"folded": 16,...}, "0": {"folded": 16,...}}
```

4 leaves, 4 children folded — the `partitions=4` run records 16. `run_finisher` writes that straight into `pyramid.overviews[k].actuals.source_children`, which is the #381 point (7) provenance record readers are told to trust. Nothing in the battery pins it: `test_partitions_compose_under_one_lease` never inspects actuals, and `test_per_entry_actuals_land_in_the_manifest` only asserts `missing == 0`.

Same root cause, second symptom: a **scoped** run (including the default `stage_sweep_after_run` chaining, which scopes to the fleet's footprint) accumulates only its own subtree, then `run_finisher` overwrites the previous whole-store actuals with it. After one small append the manifest claims level 0 folded 2 children when the store has thousands.

Both want the accumulator keyed by artifact rather than summed blindly — e.g. `level_actuals[k]["source_children"]` accumulated per `(target, key)` seen once, and reset (or merged, not added) per partition pass. Worth a test: `partitions=4` and `partitions=None` over the same fixture must produce identical `actuals`.

src/zagg/sweep_stage.py
```diff
@@ -0,0 +1,1186 @@
...
+            after, attrs = self._root()
+            if _same_stamp(before, after):
+                return values
+            logger.info(f"stage sweep: column {self.path} moved mid-read; re-reading")
+            self.stamp, self.attrs = after, attrs
```
espg, line 343: 🤖 *from Claude (review)*

**The stamp-validation re-read drops the foreign-fresh check, so the one race the backstop was ruled for slips through it.**

`_foreign_fresh` is evaluated exactly once, in `_ColumnReader.__init__` (line 328), against the stamp read at construction. When the pre/post check here detects movement, the reader adopts the new stamp (`self.stamp, self.attrs = after, attrs`) and retries **without re-testing it**. So:

- fleet rewrites the column mid-read → correct, re-read, fold the coherent column (the ruled fleet ∥ sweep case);
- a **foreign sweep** rewrites the column mid-read → also just re-read, and the merge silently consumes another run's stage output.

The second case is precisely the residual the ruling names ("TTL-expiry clock skew, zombie workers from a crashed run", issue #384 §7 / the phase-2 addition (a)), and given the heartbeat cadence problem I've flagged on `sweep_stages.py:391` it is not hypothetical: the zombie is the *previous* holder, still writing after a sibling claimed its expired lease.

`test_stamp_validation_rereads_a_moved_column` (`tests/test_sweep_stage.py:519`) only exercises the fleet arm — it rewrites the leaf with `_write_leaf`, which stamps no `run_id`, so `_foreign_fresh` would return `False` even if it were called.

One line: re-run the `_foreign_fresh(after, ...)` test after adopting the moved stamp and raise `ForeignSweepError` there, same as construction. That also needs the reader to keep `run_id`/`run_started` as instance state (currently they are constructor locals).

tests/test_sweep_stage.py
```diff
@@ -0,0 +1,943 @@
...
+        "semantic_hash": "t",
+        "cell_order": 5,
+        "shard_order": 3,
+        "split_schedule": [1, 1, 1],
+        "path_grouping": 1,
+        "pyramid": {
+            "spec": PYRAMID_SPEC_V2,
+            "overviews": expand_overviews([4], parent_order=3),
+            "overview": {
+                "all_time": False,
```
espg, line 267: 🤖 *from Claude (review)*

**Test gap: the entire windowed / all-time arm of the stage worker is unexercised — `all_time` is never `True` in any test in the repo.**

Every staged-sweep test builds its store through `_stage_store`, whose manifest has no `temporal` block, so `sweep_stage_pass` computes `windowed = False` and `_window_work` yields the single `(SCHEDULE_NONE_TOKEN, [None])` item. `all_time` is consequently `False` on every `stage_node` call in the suite (`grep -rn "temporal" tests/test_sweep_stage.py tests/test_sweep_lease.py` → nothing). What that leaves untested:

- `_stage_fold`'s `res_src = r if (all_time and regime == STAGE_GATHER) else shard_order` branch (line 599) — the arm that decides whether the all-time fold stays on the gen-1 tier, i.e. the merge-source law's one non-obvious case;
- `_merge_slabs`' multi-window path: `windows = next(...)` (line 501), the exact arm's `combine_dense` accumulation across windows (line 521), and the digest arm folding every window's payloads into one `pending` cell (line 536);
- the `if dispatch == 0 or (windowed and all_time): return` guard (line 1067) — the rule that stops an all-time column from relaying merged content;
- per-window stage columns (`window=fold_windows[0]`, line 1109) and the per-window `_readers_for` fan-out.

I ran a windowed variant of the fixture by hand (o3/d=1, two windows, `all_time: True`) and it does behave correctly — `1/2020.zarr` = 816, `1/2021.zarr` = 2448, `1/all.zarr` = 3264, root regime `stage-merge`, `merges_from_raw: 2` — so this is a coverage gap, not a live defect. But it is a large, ruling-governed branch with zero pinning, and the acceptance battery in the PR body makes no claim about it either way. A windowed fixture asserting (a) the all-time artifact equals the cross-window fold of the per-window ones, (b) no all-time stage column is written, and (c) byte-identity across `tuple_width` **with** windows, would close it.

src/zagg/sweep_lease.py
```diff
@@ -0,0 +1,212 @@
...
+            f"(acquired {existing.get('acquired_at')}, last heartbeat "
+            f"{existing.get('heartbeat_at')}, ttl {existing.get('ttl_s')}s) — sweeps "
+            f"serialize per store; wait for it, or claim after the heartbeat expires"
+        )
+    # Expired (or debris): claim. delete + conditional create + verify — the
+    # pair is not atomic, so read back and confirm ownership; a lost race
+    # refuses exactly like a live lease.
+    prior = (existing or {}).get("run_id")
+    try:
+        obstore.delete(store, LEASE_NAME)
```
espg, line 159: 🤖 *from Claude (review)*

**The crash-recovery claim deletes the intent unconditionally, so a second claimant can delete a claim that has already been verified — admitting two holders at once.**

The docstring says the verify "closes most of the window", but the delete is what opens the uncloseable part. Interleaving, both runners seeing the same expired intent:

```
A: delete(expired)  A: put(create) -> ok   A: verify -> A   A: returns, starts sweeping
B: delete(A's FRESH intent)  B: put(create) -> ok   B: verify -> B   B: returns, starts sweeping
```

A verified before B's delete and never looks again until its first heartbeat — one tuple later (see my note on `sweep_stages.py:391`). Both are writing stage columns into the same store in the meantime. Note this is *not* the same as the acknowledged "lost the claim race" path: the loser there refuses, whereas here the earlier claimant wins the verify and is then silently dispossessed.

Cheap tightening: re-read immediately before the delete and skip the delete if the intent is no longer the expired one we read (`heartbeat_at` moved or `run_id` changed), then fall through to the ordinary live-lease refusal. That makes the sequence `read → confirm-still-expired → delete → create → verify`, which turns the double-admission into a refusal for whichever runner arrives second. It does not make the pair atomic, but it removes the case where a *fresh* lease is deleted.

src/zagg/sweep_stages.py
espg: 🤖 *from Claude*

Folded in 7a994e37: `sweep_stage_pass` gains an `on_node` seam fired after every dispatch node, and `run_stage_sweep` rides it with a wall-clock-throttled beat (`ttl_s / 3` since the last beat), keeping the per-tuple beat as well and beating once more immediately before the finisher's RMW. Pinned by `TestHeartbeatCadence` (the seam fires once per node; the orchestrator beats inside a tuple under a 1 s TTL).

src/zagg/sweep_stage.py
espg: 🤖 *from Claude*

Folded in f5de4dd2, two halves: (1) `_ColumnReader.read` now treats an absent GROUP like an absent field — `(KeyError, FileNotFoundError, GroupNotFoundError)` reads `None`, so a deepened declaration (or a transient member fault) under-fills and completes instead of aborting (`TestSoftBarrierReadFaults::test_missing_member_group_fills_never_aborts`); (2) `source_children.unreadable` is real now — `_root` raises on corrupt root metadata (clean absence still reads absent), `_readers_for` marks such children with an `UNREADABLE` sentinel, and both fold kernels count it separately from `missing` (`test_corrupt_column_counts_unreadable_not_missing`).

src/zagg/sweep_stages.py
espg: 🤖 *from Claude*

Folded in bebfc70d: `_accumulate_actuals` is keyed per `(artifact node, window)` and ASSIGNS rows; `aggregate_actuals` sums them exactly once at the end, so partition re-visits of shared coarse ancestors (and current-skip re-visits generally) cannot inflate the counts. The partitions test now pins the manifest actuals byte-for-byte against the unpartitioned expectation (`folded: 4` at node 0). On the scoped-run half: a scoped run's `actuals` deliberately record that run's observation — `run_id` names the writing run, and the spec text frames `source_children` as "the run's per-artifact coverage counts" — so a scoped update overwriting the previous whole-store row with its own (skip-included: current entries also contribute their recorded rows, which carry the prior coverage forward for every artifact the run visits) is the designed posture. If a cross-run accumulation semantic is wanted instead, that is an espg call — flagged in the PR body.

src/zagg/sweep_stage.py
espg: 🤖 *from Claude*

Folded in f5de4dd2 (guard) + 7d5536ce (regression test): the reader keeps its run identity, `_foreign_guard` is re-run on every post-read stamp before a re-read is accepted, and `TestForeignStampAtReread` drives the exact scenario — a mid-read rewrite stamped by a foreign run — to the loud abort.

tests/test_sweep_stage.py
espg: 🤖 *from Claude*

Folded in 663eafd9: `TestWindowedStageSweep` builds a two-window `morton-hive/2` store with `all_time: true` and pins the per-window artifacts (gather regime, gen-1), the all-time fold as `stage-merge` at exactly 2 merges from raw with the exact-class cross-window sum, and the absence of any all-time stage column at width 1 (per-window columns carry the gen-1 tier).

src/zagg/sweep_lease.py
espg: 🤖 *from Claude*

Folded in fd068e5b: the claim path re-reads immediately before the delete and refuses on ANY movement since the expiry read, so a sibling's already-verified fresh claim can no longer be deleted. The remaining GET→DELETE window is the documented residual the foreign-fresh-stamp abort backstops. Pinned by `test_claim_refuses_if_the_intent_moved_before_the_delete` (the racing sibling's intent survives intact).

espg:
🤖 *from Claude*

Three of the "Questions for review" now have follow-up homes (espg-directed): question 6 (same-second skip-gate blind spot) → #417, carrying the cheap fix this PR accidentally enabled — the skip key comparing `(timestamp, run_id)` since stamps already carry `run_id`; question 7 (`bench_objects.py` column term) → #418, which doubles as the explicit `.github` change authorization; questions 2 + 3 (Lambda stage transport, chaining default flip) → #419, holding both operational graduations with their trigger conditions. Questions 1, 4, and 5 remain review-time confirmations on this thread; question 8 is resolved by this PR itself (the regenerated fixtures include the sidecars).

espg:
🤖 *from Claude*

**espg rulings recorded (2026-08-09), closing this PR's review questions:** (1) **confirmed** — family `materialized` is not written on `/2` stores; per-entry actuals are the sole record. (4) **confirmed** — the K==1 posture stands: grids with no strictly-interior chunk order keep the `/1` fallback (the cascade-from-raw regime), preserving pyramid capability for the one shape `/2` cannot express. (5) **blessed** — the byte-identity scoping stands as documented: ladder artifacts (reader-facing product) are byte-invariant under any tuple width; stage-column placement is orchestration-never-contract per spec §4.6. Questions 2/3/6/7/8 were previously routed to #419, #417, #418, and this PR's own fixture regen.

Issue #438, opened 2026-08-16
Re-pin benchmark shardmap fixtures under the authalic convention (mortie 0.9.8)


espg:
🤖 *from Claude*

espg directed (in-session, 2026-08-16): purge all pre-authalic shardmaps locally and update/overwrite the ones stored in git, so no cross-catalog morton-convention mixing can occur after mortie 0.9.8 (authalic default, espg/mortie#186/#188).

Local purge is done (24 untracked pre-authalic shardmap/submap JSONs removed; `demo/outputs/` keeps only the five rebuilt today under 0.9.8). Remaining tracked artifacts to re-pin on a `claude/` branch:

- `tests/data/benchmark/shardmaps/sm_healpix_{o9,o10,o11,o9_88s,o10_88s}.json` — all `backend: mortie`, morton-keyed, pinned under geodetic-as-spherical; their drift guard (`test_pinned_shardmap_no_drift`) rebuilds live and now diverges under 0.9.8 by design.
- the per-entry `shard_key` morton words in `tests/data/benchmark/targets.json` (densest-shard pins, incl. the nested o10-inside-o9 88S pin).

Out of scope: `sm_rect_{3km,6km}.json` (rectilinear, no morton convention); committed catalog parquets (plain stac-geoparquet, convention-free); the untracked `targets_*.json` fleet manifests (espg regenerates at the next fleet run); `tests/data/spec/` fixtures (regenerate with the zagg-side authalic spec-marker PR, refs #392).

PR #440, opened 2026-08-16
Stored-index query path: intersect before materializing granule records

### LINKED ISSUE #439: Stored-index query path: intersect before materializing granule records
🤖 *from Claude*

Measured while reflowing the query demo to polygon-direct (no bbox prefilter), against the full 555,867-granule ATL03 clone with mortie 0.9.8 and an o9 `footprint_cells` column:

- `ShardMap.build(full_idx, grid9, region=ca_parts)` (stored-index fast path, engaged and verified): **38.1 s**
- `Catalog.granule_records()` alone over the same table: **24.9 s** — shapely-parses every row's WKB into record dicts *before* the intersection runs
- the same build on a 4,354-row bbox-cut subset: **0.32 s**

So the polygon-direct query is record-materialization-bound, not morton-bound: the batched `mocs_and` sees arrays that already exist (`values`/`offsets` from the column), but `build` materializes all 555,867 granule records first and only then intersects, discarding ~99.6% of them (2,357 assigned). The stored-index plan (`_footprint_cells_plan`) could invert this: run the intersection on the stored arrays first, then materialize records only for rows with nonempty intersection — the polygon-direct query should land near the bbox-cut cost (~sub-second + the intersection), making the local morton query path strictly dominate a bbox prefilter.

Care points: `granule_records` also screens empty/non-polygonal rows (the plan's id-alignment relies on that screen's row semantics), and `metadata["total_granules"]` currently records records-considered — the inversion should keep both stable or document the change.

Refs #396 (the column + fast path), #400 (batch coverage); measured numbers from the demo notebook reflow session, 2026-08-16.

espg:
Closes #439.

## What

The stored-index fast path (`_footprint_cells_plan` → `_intersect_footprint_cells`, issue #396) now runs its morton intersection **before** any granule record exists, and decodes records only for the rows the intersection kept.

`ShardMap.build` used to open with an unconditional `records = catalog.granule_records()` (`shardmap.py:1007` at `ed8a31ac`). Over the 555,867-granule ATL03 clone that is ~18 s of shapely WKB parsing plus `to_pylist` over the nested `assets` column — to feed an intersection that assigns 2,357 granules and throws the other 99.6% away. The stored `footprint_cells` column is already table-row-ordered, so nothing in the intersection needed the records at all.

## Approach

**1. The plan stops taking records** (`_footprint_cells_plan(catalog, grid, chosen, footprint, mortie_order)`).

It used to derive `rows` (record index → table row) by looking each record's id up in an id→row dict built from the whole column:

```python
row_of = {gid: i for i, gid in enumerate(catalog.table.column("id").to_pylist())}
...
rows = np.fromiter((row_of[r["id"]] for r in records), dtype=np.int64, count=len(records))
```

`granule_records` emits rows in table order, skipping only rows whose geometry is empty or non-polygonal, so `rows` is exactly `np.flatnonzero(screen)` — the same map, derivable without a record. The plan now returns `(values, offsets, order, rows, considered)`, where `considered` is the screen's popcount.

**2. The screen is shared, not re-implemented** — new `Catalog.granule_row_mask()` (`sources.py`), lifted verbatim out of `index_footprints` (the batched `shapely.from_wkb` + `is_empty`/`get_type_id` predicate at `sources.py:601-608` on `ed8a31ac`), and now called by both. Same predicate as `granule_records`'s row-wise `geom.is_empty or geom.geom_type not in ("Polygon", "MultiPolygon")`, so a screened row cannot be assigned and cannot shift its neighbours. It costs **0.47 s** over the clone against `granule_records`'s **18.3 s** (measured below). `index_footprints`'s behaviour is unchanged — the shapely objects still die before the cover runs, they just die with the helper's frame instead of an explicit `del geoms`.

One input is *not* the same predicate, and the screen refuses it rather than diverging (`84c47078`, review fold): a **null** geometry. `shapely.from_wkb` maps null → `None`, whose vectorised `get_type_id` is `-1` (would screen silently) while the row-wise `None.is_empty` raises `AttributeError`. Silently screening it would let an indexed build drop a granule — and drop it from `total_granules` — where the geometry build still crashed, so `granule_row_mask` now raises a `ValueError` naming the count and the first offending row. Both paths refuse, as both did before this PR. `index_footprints` inherits the refusal (it shares the screen), so it raises where it would previously have given the null row an empty MOC.

**3. Duplicate-id refusal stays table-wide.** The ids no longer carry the alignment, but a repeat still means the column cannot be trusted to belong to the rows beside it, so the check is kept as a catalog-integrity check — including for duplicates among rows this build's AOI would have discarded. Same trigger; it is now `pyarrow.compute.count_distinct(..., mode="all")` instead of materializing the dict (`mode="all"` so a null id counts as its own value, matching the dict it replaces). The **message** is reworded (`3573fad5`, review fold) because the old one named the deleted mechanism — it now says the column is *matched to rows by position*, so a repeated id means the table was concatenated or re-keyed and the stored footprints can no longer be shown to belong to the rows they sit beside. The same rewrite lands on `_footprint_cells_plan`'s `Raises` section and `Catalog.footprint_cells`' `Returns` text, both of which still asserted id alignment.

**4. `build` defers, then renumbers** — new `_hit_records(catalog, rows, shard_to_idx)` takes the hit rows out of the table (`Catalog(catalog.table.take(hit_rows), ...)`), decodes that slice, and renumbers `shard_to_idx` onto it. Record *shape* is unchanged because it is the same `granule_records()` on the same rows. It carries a defensive raise if the screen and `granule_records` ever disagree on a row count, because a silent mismatch there would shift every granule onto its neighbour's entry.

**5. Provenance is byte-stable.** `metadata["total_granules"]` still records the records **considered** (the whole post-screen catalog), now from the screen's popcount rather than `len(records)`. `granules_assigned`, `total_shards`, `total_pairs`, `footprint_cells`, `mortie_order` are all untouched. `build_wall_s` keeps `main`'s boundary: the timer opens **before** `_footprint_cells_plan`, so the plan (which is where the record → row alignment lives, an id → row dict on `main` and the `granule_row_mask` screen here) is inside it exactly as it was, and only record decode is outside. An earlier revision of this branch opened the timer after the plan, which made the number incomparable with the `build_wall_s` values already recorded in existing manifests; corrected in `ce0bd717` (review fold) and re-measured below.

**6. Nothing else moves.** The live-geometry path, every other backend, `footprint="beams"`, a pinned `mortie_order`, a spherely build, and a catalog with no column all still call `granule_records()` eagerly, up front, exactly as before. The multipolygon superset semantics are unchanged (the column covers every part; `granule_records` reads the largest part's ring) — pinned by a test.

## Phases

- [x] **Phase 1** (`47669b4b`) — records-free `_footprint_cells_plan`, `Catalog.granule_row_mask()`, `_hit_records`, deferred materialization in `build`. ~~existing plan callsites updated~~ — **not true as written**: `bench/shardmap_batch_vs_serial.py:246` was missed and still passed `records` plus a 4-tuple unpack, so the bench `cells` arm raised `TypeError`. Nothing under `tests/` covers `bench/`, so CI was green regardless. Fixed in `3819764e` (review fold) and verified by running the arm.
- [x] **Phase 2** (`244b6691`) — `TestDeferredRecords`: byte-identical serialization vs the eager order, provenance stability, screen/duplicate/order/multipolygon traps.
- [x] **Review fold** (`3819764e`, `ce0bd717`, `3573fad5`, `84c47078`) — bench callsite; `build_wall_s` timer boundary restored to `main`'s; the three doc sites + refusal message rewritten to positional alignment; null-geometry predicate parity (both paths refuse) with regression tests in `tests/test_shardmap.py` and `tests/test_sources.py`.

## Measured

`demo/atl03_v007_full.parquet` (555,867 rows) → `index_footprints(9)` → `ShardMap.build(idx, HealpixGrid(9, child_order=19, chunk_inner=13), region=load_polygon(demo_aoi("california")))`. mortie 0.9.8, macOS arm64, min of 3 after one warm-up, same process, same catalog object.

| | `main` (`ed8a31ac`) | this branch | |
|---|---|---|---|
| `ShardMap.build` wall | **19.69 s** | **1.79 s** | **11.0× faster** |
| `metadata["build_wall_s"]` (plan + intersection) | 1.52 s | 1.57 s | like-for-like |
| `total_granules` | 555,867 | 555,867 | identical |
| `granules_assigned` | 2,357 | 2,357 | identical |
| `total_shards` | 2,726 | 2,726 | identical |
| `total_pairs` | 191,219 | 191,219 | identical |
| assigned granule ids | — | — | **identical** (2,357, set-equal) |
| `(shard_key, granule-id tuple)` pairs | — | — | **identical** (all 2,726, order included) |

Component costs over the same clone, which is where the 10.8× comes from:

| | |
|---|---|
| `Catalog.granule_records()` (whole catalog) | 18.33 s |
| `Catalog.granule_row_mask()` (the screen that replaces it) | 0.50 s |
| `_hit_records` decode (2,357 rows) | the remainder, ~0.15 s |

The residual 1.79 s is now the plan + intersection (1.57 s, the screen included) plus the hit-row decode — i.e. the build is morton-bound again, which is what #439 asked for. The issue quotes 38.1 s / 29.5 s for the same build; this run measures the `main` side at 19.7 s on a warm page cache, so read the ratio rather than the absolute.

**Re-measured after the review fold** (`ce0bd717`), same script/process/catalog object as above, since restoring the timer boundary moves `build_wall_s`:

```
granule_row_mask 0.502 s  kept 555867 / 555867
build wall (s)   2.504  1.796  1.792  1.968   -> min-of-3 1.79
build_wall_s     2.248  1.571  1.572  1.735   -> min-of-3 1.57
total_granules 555867  granules_assigned 2357  total_shards 2726  total_pairs 191219
```

The total wall is unchanged (1.83 → 1.79 s is run-to-run noise); `build_wall_s` rises 1.21 → **1.57 s**, which is the 0.50 s screen plus the plan's `count_distinct` re-entering the timer where `main` had them. Against `main`'s 1.52 s that is now a comparable number rather than a boundary artifact. All four provenance counts are identical to the pre-fold run.

## How tested

- `uv run pytest -q` — **3869 passed, 38 skipped, 3 failed**, all three pre-existing and unrelated (see below). Post-fold count; `tests/test_shardmap.py tests/test_sources.py` alone are 148 passed, 1 skipped.
- `ruff check` / `ruff format --check` — clean on every touched file (`src/zagg/catalog/shardmap.py`, `src/zagg/catalog/sources.py`, `tests/test_shardmap.py`, `tests/test_sources.py`, `bench/shardmap_batch_vs_serial.py`). One pre-existing `N818` in `src/zagg/registry.py` is untouched and predates the branch (CI's lint bot runs `--select=E,F,W,I`, which excludes it). `tests/data/benchmark/README.md` is reported unformatted on `main` too; left alone.
- `uv run python bench/shardmap_batch_vs_serial.py --measure neon --path cells --order 9` — the `cells` arm runs to completion on the committed NEON fixture (2,089 granules, 4 shards, 314 pairs), which is what the fixed callsite needed to prove.
- New `TestDeferredRecords` in `tests/test_shardmap.py`:
  - `test_serializes_identically_to_the_eager_order` — builds twice on the same catalog/grid/AOI, once with `_hit_records` monkeypatched to `_eager_hit_records` (the pre-#439 shape: every record decoded, indices untouched), and compares the full `to_json` payload key-for-key with only `build_wall_s` (a stopwatch) popped. Fixture has a screened row mid-catalog and an AOI narrower than the catalog, so `rows` is non-identity and most records are unassigned.
  - `test_total_granules_still_counts_the_records_considered` — fast vs geometry build agree on `total_granules` and `granules_assigned` with a screened row present.
  - `test_screened_row_inside_the_aoi_is_never_assigned` — a `Point` and an empty `Polygon` **inside** the AOI cover; neither may be assigned, and neither may be counted.
  - `test_duplicate_ids_outside_the_aoi_still_refuse` — both copies of the repeated id placed far outside the region, so a hits-only check would sail through.
  - `test_order_refusal_precedes_any_record_decode` — the coarse-column refusal is unchanged, and (new) is now reached with `granule_records` monkeypatched to `pytest.fail`, so the refused build decodes nothing.
  - `test_multipolygon_superset_survives_the_inversion` — the disclosed superset still holds and the extra shards still carry the granule's own record (`s3` href asserted).
  - `test_null_geometry_refuses_on_both_paths` (review fold) — row 0's WKB nulled: `index_footprints(11)` refuses, `ShardMap.build` on the **indexed** catalog refuses (it previously built a manifest silently omitting the granule), and `ShardMap.build` on the **unindexed** one still raises.
- `tests/test_sources.py::TestFootprintCells::test_null_geometry_raises_rather_than_screening` (review fold) — the unit-level pin next to the contrast case `test_non_polygonal_rows_get_an_empty_moc`: `granule_row_mask()` raises naming the count and first row, `index_footprints` raises, `granule_records()` raises.
- Mutation-checked the new pins: forcing `rows` to the identity fails 3 of them (plus the existing `test_batched_cells_equal_the_scalar_loop`), and perturbing `_hit_records`'s renumbering order fails the serialization test.

Pre-existing failures, **not** fixed here:
- `tests/test_content_hash.py::TestWorkerWiring::test_matches_pinned_fixture_hashes` and `::TestStreamingLeafStaging::test_unsharded_leaf_reproduces_pinned_hashes_from_staged_values` — this env resolves **mortie 0.9.8**, so these are the pinned-hash failures belonging to #438 / `claude/438-authalic-repin`.
- `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` — environmental.
- The `TestStatusPoller` flake did not reproduce in this run.

## Questions for review

1. **`shardmap.py` is 1,452 lines** (1,369 on `main`, so +83 — comments and docstrings, plus the 24-line `_hit_records`). It was already past the 1,200 raise trigger before this branch, so nothing here crosses it, but flagging rather than splitting the module unilaterally. Want a follow-up issue to carve the fast path (`_intersect_footprint_cells` / `_footprint_cells_plan` / `_hit_records`, ~250 lines) into `catalog/footprint_cells.py`, or leave it?
2. **`Catalog.granule_row_mask()` is new public API.** It could have stayed a module-private helper in `sources.py` imported across into `shardmap.py`, but the mask is a property of the catalog's rows and two callers now want it. Fine as a method, or prefer `_granule_row_mask`?
3. **`build_wall_s` excludes `_hit_records`** (record decode has always been outside that timer, and the plan is now back inside it per `ce0bd717`). That does mean the fast path's reported wall does not account for ~0.15 s of real build time it performs after the intersection. Keep the convention, or fold the decode in too?
4. **The duplicate-id check is now `pc.count_distinct(mode="all")`** rather than a materialized dict. Same number; `mode="all"` was chosen so a null id counts as its own value the way the dict key did. Worth pinning a null-**id** case in a test, or is a null `id` already impossible upstream? (A null **geometry** is now pinned, and refuses — see (2) in Approach.)
5. The measured `main` baseline here (19.7 s) is well under the 29.5–38.1 s in the issue — same script, warm page cache, catalog loaded once and re-used across builds. Flagging in case the difference matters for how the number gets cited.
6. **The null-geometry refusal makes `index_footprints` stricter than it was on `main`**, which is the one behaviour change outside the fast path: `main` gave a null-geometry row an empty MOC and indexed happily, and only `granule_records` later crashed on it. Refusing at the screen is what makes the two build paths agree (the alternative — screening on both — would need `granule_records` loosened, which is a scope change). Flagging in case any real catalog carries null geometries; none in the tree does.

## CI

`ruff`, `check`, `build`, `build / build-x86_64`, `build / build-arm64` all pass. `test (3.12)` and `test (3.13)` fail on exactly two tests — `tests/test_content_hash.py::TestWorkerWiring::test_matches_pinned_fixture_hashes` and `::TestStreamingLeafStaging::test_unsharded_leaf_reproduces_pinned_hashes_from_staged_values` — with the identical hash mismatch (`...2b91e46843c7` vs the pinned `...2f77bcafab31`) that `main`'s own most recent Tests run shows at `ed8a31ac` ([run 31976725212](https://github.com/englacial/zagg/actions/runs/31976725212)). Pre-existing, belongs to #438; nothing else in the suite fails.


espg:
🤖 *from Claude*

Status: both phases landed (`47669b4b`, `244b6691`), full local suite run, PR left in **draft** — no review cycle run from this session by design, the parent session handles review/fold.

CI: `ruff` / `check` / `build` / `build-x86_64` / `build-arm64` green. `test (3.12)` and `test (3.13)` red on exactly two tests, `tests/test_content_hash.py::TestWorkerWiring::test_matches_pinned_fixture_hashes` and `::TestStreamingLeafStaging::test_unsharded_leaf_reproduces_pinned_hashes_from_staged_values` — same mismatch (`...2b91e46843c7` vs pinned `...2f77bcafab31`) that `main`'s own Tests run at `ed8a31ac` shows ([run 31976725212](https://github.com/englacial/zagg/actions/runs/31976725212)). Pre-existing, tracked by #438; deliberately not touched here.

Headline: `ShardMap.build` over the 555,867-granule ATL03 clone against the California AOI goes **19.69 s → 1.83 s** (10.8×) with byte-identical output — all 2,357 assigned ids and all 2,726 `(shard, granule-list)` pairs equal, `total_granules` / `granules_assigned` / `total_shards` / `total_pairs` unchanged. Details and the component breakdown are in the PR body.

Five items under **Questions for review**, the load-bearing one being (1): `shardmap.py` is now 1,452 lines (1,369 on `main`). It was already past the 1,200 raise trigger before this branch, so this PR does not cross it, but flagging rather than splitting the module unilaterally.

src/zagg/catalog/shardmap.py
```diff
@@ -475,7 +475,7 @@ def _intersect_footprint_cells(rows, values, offsets, grid, all_shards) -> Dict[
     return _regroup_hits(hit_shards, hit_owners)
 
 
-def _footprint_cells_plan(catalog, records, grid, chosen, footprint, mortie_order):
+def _footprint_cells_plan(catalog, grid, chosen, footprint, mortie_order):
```
espg, line 478: 🤖 *from Claude (review)*

**The signature change breaks an in-tree caller that was not updated: `bench/shardmap_batch_vs_serial.py:246` still passes `records`, so the `cells` arm of that benchmark now dies with `TypeError` before it measures anything.**

The PR body says "existing plan callsites updated", but the bench harness was missed:

```python
# bench/shardmap_batch_vs_serial.py:245-248
def fn(records, grid, all_shards, order):  # unused args: the shared arm signature
    plan = shardmap._footprint_cells_plan(cat, records, grid, "mortie", "swath", None)
    values, offsets, _order, rows = plan
    return shardmap._intersect_footprint_cells(rows, values, offsets, grid, all_shards)
```

Confirmed in the worktree at `244b6691`:

```
$ uv run python -c "import inspect; from zagg.catalog import shardmap; \
    print(inspect.signature(shardmap._footprint_cells_plan)); \
    shardmap._footprint_cells_plan(1,2,3,\"mortie\",\"swath\",None)"
(catalog, grid, chosen, footprint, mortie_order)
TypeError: _footprint_cells_plan() takes 5 positional arguments but 6 were given
```

Two edits needed, not one: drop the `records` argument, and widen the unpack to the new 5-tuple (`values, offsets, _order, rows, _considered`) — the second would `ValueError` even if the call were fixed. The surrounding comment there ("resolving the plan (which includes the id -> table-row alignment `build` pays every time)") is also now describing a mechanism this PR removed; the plan pays the shapely row screen instead, which is the term that arm should be timing after #439.

No test covers `bench/`, so this fails silently until someone runs the batch-vs-serial harness — which is exactly the harness the numbers in this PR body would be re-checked with.

src/zagg/catalog/shardmap.py
```diff
@@ -1005,14 +1076,22 @@ def build(
...
+        plan = _footprint_cells_plan(catalog, grid, chosen, footprint, mortie_order)
+        # Records are the fast path's whole remaining cost -- shapely-parsing
+        # every row's WKB and ``to_pylist``-ing the nested assets column ran ~25 s
+        # of a 29.5 s clone-scale build, to feed an intersection that discards
+        # 99.6% of them (issue #439). So decode them AFTER the intersection
+        # there, for its hit rows only; every other path needs them all up front.
+        records = None if plan is not None else catalog.granule_records()
+        considered = plan[4] if plan is not None else len(records)
+
+        t0 = time.perf_counter()
```
espg, line 1108: 🤖 *from Claude (review)*

**`build_wall_s` did not keep the pre-existing convention: `t0` moved to *after* `_footprint_cells_plan`, so the fast path's record→row alignment left the timer too — not just the decode. The "1.52 → 1.21 s, unchanged in kind" row in the PR body is partly a timer-boundary artifact.**

On `main` (`ed8a31ac:src/zagg/catalog/shardmap.py`, lines 1008-1013) the timer opened *before* the plan:

```python
        t0 = time.perf_counter()
        # Fast path (issue #396): ...
        plan = _footprint_cells_plan(catalog, records, grid, chosen, footprint, mortie_order)
        if plan is not None:
            values, offsets, mortie_order, rows = plan
            shard_to_idx = _intersect_footprint_cells(rows, values, offsets, grid, all_shards)
```

so `build_wall_s` on `main` included the plan's whole-column `to_pylist()` + id→row dict — the alignment work. On this branch the plan runs at line 1083 and `t0` opens at 1092, so the alignment (now `granule_row_mask()`: a full-catalog `shapely.from_wkb` over every row, the PR's own 0.47 s at clone scale) is *outside*. The comment at 1116-1117 ("`build_wall_s` has always meant the intersection's wall, with record decode outside it") is accurate about the **decode** and inaccurate about the **plan**.

Measured in the worktree at `244b6691` (6,000-row synthetic catalog, o11 column, `HealpixGrid(11, 17, "fullsphere")`, SERC AOI):

```
rows=6000  granule_row_mask=0.002s  _footprint_cells_plan=0.002s
build_wall_s=0.003  (ShardMap.build total 0.027s)
```

i.e. the excluded plan is ~40% the size of the number `build_wall_s` now reports. At clone scale the PR's own figures say the same thing: 0.47 s of screen sits outside a 1.21 s reported wall, and ~0.3 s of the 1.52→1.21 drop is the `main` plan cost leaving the timer rather than work getting cheaper.

That matters twice: `build_wall_s` is written into every manifest (provenance), and Q3 in the PR body asks about the timer while only disclosing the `_hit_records` side. Either fold the plan back inside `t0` (which restores comparability with `main` and with `build_wall_s` values already recorded in existing manifests), or say in the PR body — and in the 1116 comment — that the fast path's wall now excludes its alignment step as well as its decode.

src/zagg/catalog/shardmap.py
```diff
@@ -526,18 +542,64 @@ def _footprint_cells_plan(catalog, records, grid, chosen, footprint, mortie_orde
...
+    # ``mode="all"`` counts a null id as its own value, matching the id ->
+    # row dict this replaces (which the pre-#439 plan built from the whole
+    # column, unique-or-not, purely to align records back to rows).
+    distinct = pc.count_distinct(catalog.table.column("id"), mode="all").as_py()
+    if distinct != catalog.table.num_rows:
         raise ValueError(
-            f"catalog has duplicate granule ids ({catalog.table.num_rows - len(row_of)} "
+            f"catalog has duplicate granule ids ({catalog.table.num_rows - distinct} "
             f"repeats over {catalog.table.num_rows} rows); the footprint_cells column is "
             f"aligned to records by id, so a repeat would hand records another granule's "
```

**The duplicate-id refusal now tells the operator a mechanism that this PR deleted — "the footprint_cells column is aligned to records by id" is false as of `47669b4b`, and the same claim is left standing in two docstrings.**

Alignment is now positional (`np.flatnonzero(catalog.granule_row_mask())`, line 565); no id lookup exists anywhere on this path. The PR body treats "same message" as the conservative choice, but the message is not neutral text — it states *why* the build refused, and an operator reading it will go looking for an id-keyed join that is gone. Worse, the stated remedy reads as optional ("De-duplicate the catalog") for a cause that no longer applies, when the real reason is the one the new docstring gives: a repeated id means the column cannot be trusted to belong to the rows it sits beside.

Three places assert the removed mechanism at `244b6691`:

- this message, `shardmap.py:555`
- `shardmap.py:520-522`, the `Raises` section of the very function that changed — `"The column is aligned to records by id, so a repeated id makes the lookup last-wins and hands every earlier record with that id another granule's footprint"`. The paragraph added below it ("The check stays **table-wide**...") is bolted onto a premise the same commit falsified; a reader hits the contradiction inside one docstring.
- `sources.py:675`, `Catalog.footprint_cells`' Returns text: `"(not every ``granule_records`` entry -- ``build`` aligns the two by granule id)"`. That is the public docstring of the accessor the fast path reads, and it is now wrong about the only thing it says on alignment.

The existing test only pins `"duplicate granule ids .*1 repeats over 3 rows"`, so none of this is caught. Suggest rewording all three to the row-position framing the code actually implements (the message can keep its shape and trigger; only the causal clause needs to change).

src/zagg/catalog/sources.py
```diff
@@ -513,6 +513,39 @@ def filter_bbox(self, boxes) -> "Catalog":
...
+        Time is small (0.02 s over the 35,639-granule 88S catalog, ~0.3 s over
+        the 555,867-row ATL03 clone) but memory is not: the ``to_numpy`` WKB
+        copy plus the live shapely objects put the clone's RSS ~1 GB over the
+        parquet read. It is a peak, not a leak -- the objects die with the call.
+        """
+        import shapely
+
+        geoms = shapely.from_wkb(self.table.column("geometry").to_numpy(zero_copy_only=False))
+        # geom_type ids 3 and 6 are Polygon and MultiPolygon.
+        return ~shapely.is_empty(geoms) & np.isin(shapely.get_type_id(geoms), (3, 6))
```
espg: 🤖 *from Claude (review)*

**The vectorized screen is not the same predicate as `granule_records`'s row-wise check on one input: a **null** geometry. The vector form drops it silently; the row-wise form raises `AttributeError`. Post-#439 that turns a hard failure on the fast path into a silently-vanished granule.**

`shapely.from_wkb` maps `None` to `None` in both forms, but the two then diverge — verified on shapely 2.1.2 / pyarrow 25.0.1 in the worktree at `244b6691`:

```
vec  from_wkb([None, <POINT>])  -> [None <POINT (0 0)>]
     is_empty -> [False False]   get_type_id -> [-1  0]        # mask row -> False
row  from_wkb(None) -> None; None.is_empty -> AttributeError
```

so `granule_row_mask` screens the row out while `granule_records`' `if geom.is_empty or ...` (`sources.py:751`) never gets to run its predicate at all. Differential check, same worktree, two-row catalog with row 0's geometry nulled:

```
mask:                       [False  True]
granule_records():          AttributeError NoneType object has no attribute is_empty
ShardMap.build(indexed):    ok — total_granules 1, granules_assigned 1     # was AttributeError on main
ShardMap.build(unindexed):  AttributeError NoneType object has no attribute is_empty
```

Before this PR both branches crashed, because `build` opened with `catalog.granule_records()`. Now an indexed catalog with a null-geometry row builds a manifest that silently omits it (and omits it from `total_granules`), while the same catalog on the geometry path still raises — the two paths disagree about whether the catalog is usable. That is exactly the class of divergence the `_hit_records` length guard was added to catch, and it slips past it: the row is never taken, so the counts agree.

Cheap and in the spirit of the rest of the change: make the screen's null handling explicit, e.g.

```python
geoms = shapely.from_wkb(self.table.column("geometry").to_numpy(zero_copy_only=False))
# get_type_id is -1 for a null WKB; the row-wise loop in ``granule_records``
# raises on those, so exclude them here rather than diverge silently.
```

and either state in the docstring that a null geometry is screened (making the fast path deliberately more permissive), or refuse it so both paths agree. A one-row null-geometry test would pin whichever you choose. Note the `Notes` on `index_footprints` and the "Same predicate as the row-wise loop" claim in this docstring both need the caveat either way.

src/zagg/catalog/shardmap.py
espg: 🤖 *from Claude*

Fixed in `3819764e`. Both edits made — dropped `records` and widened the unpack:

```python
plan = shardmap._footprint_cells_plan(cat, grid, "mortie", "swath", None)
values, offsets, _order, rows, _considered = plan
```

The surrounding comment now names the term this arm actually times after #439 (`granule_row_mask`'s whole-table screen, in place of the id lookup), rather than the id → row dict that no longer exists.

Swept the rest of `bench/shardmap_batch_vs_serial.py` for other stale plan assumptions: `_load_case` and `_intersect_serial` touch neither the plan nor `granule_row_mask`, and the only other `footprint_cells` references are prose about `index_footprints` being one-time, which is still true. Nothing else needed changing.

Verified by running the arm rather than by inspection — `neon` is a committed fixture, so it runs in any checkout:

```
$ uv run python bench/shardmap_batch_vs_serial.py --measure neon --path cells --order 9
{"granules": 2089, "shards_aoi": 4, "shards_hit": 4, "pairs": 314, "digest": -4289743750552133645,
 "wall_s": 0.01, "index_s": 0.16, "rss_load_mb": 193.1, ...}
```

The PR body's "existing plan callsites updated" claim is corrected there too.

src/zagg/catalog/shardmap.py
espg: 🤖 *from Claude*

Folded the plan back inside the timer in `ce0bd717` — `main`'s boundary restored, so `build_wall_s` stays comparable with the values already written into existing manifests.

Two segments rather than one span, because the geometry path's `granule_records()` now sits between the plan and the intersection and has always been outside:

```python
t0 = time.perf_counter()
plan = _footprint_cells_plan(catalog, grid, chosen, footprint, mortie_order)
plan_wall = time.perf_counter() - t0
records = None if plan is not None else catalog.granule_records()
...
t0 = time.perf_counter()
...
wall = plan_wall + (time.perf_counter() - t0)
```

`_hit_records` stays outside, which is the part of the 1116 comment that was accurate; the comment now says so explicitly instead of claiming the timer ever meant "the intersection only".

Re-measured on `demo/atl03_v007_full.parquet` (555,867 rows, `index_footprints(9)`, `HealpixGrid(9, child_order=19, chunk_inner=13)`, California AOI, min of 3 after one warm-up, same process/catalog object as the PR-body run):

```
granule_row_mask 0.502 s  kept 555867
build wall (s)   2.504  1.796  1.792  1.968   -> min-of-3 1.79
build_wall_s     2.248  1.571  1.572  1.735   -> min-of-3 1.57
total_granules 555867  granules_assigned 2357  total_shards 2726  total_pairs 191219
```

So exactly as predicted: the `ShardMap.build` wall is unchanged (1.83 → 1.79 s, run-to-run noise) and `build_wall_s` rises 1.21 → **1.57 s**, ~the 0.5 s screen plus the plan's `count_distinct`. Against `main`'s 1.52 s that is now a like-for-like number. PR-body timing table and Q3 updated.

src/zagg/catalog/shardmap.py

All three rewritten to the positional truth in `3573fad5`.

1. The refusal message (`shardmap.py`) keeps its shape and trigger, and replaces the causal clause:

> catalog has duplicate granule ids (N repeats over M rows); the footprint_cells column is **matched to rows by position**, so a repeated id means the table was concatenated or re-keyed and the stored footprints can no longer be shown to belong to the rows they sit beside -- building from them could put one granule's footprint on another. De-duplicate the catalog, or drop the column to build from geometry.

2. `_footprint_cells_plan`'s `Raises` section no longer asserts the id lookup, so it stops contradicting the paragraph below it: alignment is positional, a repeat cannot misdirect a lookup (there is none), and the refusal is the catalog-integrity statement the next paragraph already described as table-wide.

3. `Catalog.footprint_cells`' `Returns` now reads "``build`` aligns the two by row position, via ``np.flatnonzero(granule_row_mask())``, since the screened rows carry a zero-length run rather than being absent" — which also explains *why* the column is per table row, the thing that sentence was there to say.

Pinned so the wording cannot drift back: `test_duplicate_ids_outside_the_aoi_still_refuse` now matches `...matched to rows by position` and asserts `"by id" not in str(exc.value)`.

src/zagg/catalog/sources.py
espg: 🤖 *from Claude*

Took the refuse-so-both-paths-agree option in `84c47078`. `granule_row_mask` now raises on a null WKB rather than screening it:

```python
type_ids = shapely.get_type_id(geoms)
null = type_ids == -1
if null.any():
    rows = np.flatnonzero(null)
    raise ValueError(
        f"catalog has {rows.size} row(s) with a null geometry (first at row {int(rows[0])}); "
        f"a granule with no footprint can be neither covered nor assigned. "
        f"Drop those rows from the catalog before indexing or building."
    )
```

Deliberately not the other direction: making the row-wise loop tolerant would change what `granule_records` accepts, which is outside this PR.

Docs updated on both sites you flagged — `granule_row_mask` gains a `Raises` section stating the divergence and why it refuses instead of screening, and its "same predicate" sentence now reads "…and refusing on the one input where 'same predicate' would otherwise be false". `index_footprints`' `Notes` records that a null geometry is the one row it refuses rather than skipping, i.e. that it now raises where it would once have indexed.

Regression tests, both directions:

- `tests/test_shardmap.py::TestDeferredRecords::test_null_geometry_refuses_on_both_paths` — a two-row catalog with row 0 nulled: `index_footprints(11)` raises, `ShardMap.build` on the **indexed** catalog raises `ValueError` (previously: built a manifest silently omitting the granule *and* omitting it from `total_granules`), and `ShardMap.build` on the **unindexed** one still raises. New helper `_with_null_geometry(cat, row)` builds the fixture.
- `tests/test_sources.py::TestFootprintCells::test_null_geometry_raises_rather_than_screening` — the unit-level pin: `granule_row_mask()` raises naming the count and first row, `index_footprints` raises, `granule_records()` raises. Sits next to `test_non_polygonal_rows_get_an_empty_moc`, which is the contrast case (screened, not refused).

PR #450, opened 2026-08-17
small fixes 2026-08-17: NaN-safe dispatch payload, LP DAAC credentials, GEDI filter set

### LINKED ISSUE #448: Dispatch payload is not NaN-safe: normalize float NaN fill_value to the string form
🤖 *from Claude*

First-run finding from the GEDI smoke on the 0.45.0 fleet (2026-08-17, the exact verification run PR #432 flagged): **the dispatch payload is not NaN-safe**. `gedi01b_waveform_healpix_hive.yaml` declares `fill_value: .nan` on five per-shot companions (noise_mean, noise_stddev, rx_energy, elevation_bin0, elevation_lastbin — the invalid-when-pooled sentinel). YAML parses `.nan` to float NaN; `json.dumps` emits literal `NaN`; Lambda's strict JSON parser refuses:

```
InvalidRequestContentException: Could not parse payload into json: Non-standard token 'NaN'
  (zagg/runner.py:5225, _invoke_lambda_ping -> lambda_client.invoke)
```

No prior config tripped this — the ATL03 templates use integer/zero fills. The string `"NaN"` is already the grammar's native form: `_is_nan_fill` (config.py:2843) accepts it, and the healpix template's own DEFAULT is the string (`grids/healpix.py:674`, `meta.get("fill_value", "NaN")`), matching zarr v3's JSON fill encoding. Verified: patching the five floats to `"NaN"` passes `validate_config` and serializes under `json.dumps(allow_nan=False)` — the demo notebook now carries that client-side normalization as a workaround (no redeploy needed; the deployed template path speaks string-NaN natively).

Fix (small):
1. `gedi01b_waveform_healpix_hive.yaml`: `.nan` → `"NaN"` (the native form).
2. Config load normalizes float-NaN fill_value → `"NaN"` so no future config can regress (and the semantic hash sees one canonical form).
3. Guard test: every packaged config's dispatch event serializes under `json.dumps(allow_nan=False)`.

Refs #425 / PR #432 (the template), #434 (E2E validation — this is its first catch).

### LINKED ISSUE #449: GEDI template ships without an LPDAAC credentials provider — reads 403 as NSIDC
🤖 *from Claude*

Second first-run finding from the GEDI smoke (0.45.0 fleet, 2026-08-17; sibling of #448): **all four shards failed with every granule read raising `TypeError: memoryview: a bytes-like object is required, not 'NoneType'`**, reported as "No data after filtering". Root cause is credentials, not the reader: `gedi01b_waveform_healpix_hive.yaml` declares no `data_source.credentials_provider`, so `_resolve_source_credentials` (runner.py:2649) fell back to the **NSIDC** default — and NSIDC temporary S3 creds cannot read LPDAAC's `lp-prod-protected`. The 403'd body surfaces as `None` → `memoryview(None)` per granule, swallowed into the per-granule error path, so the shard reports the generic no-data message with the TypeError only as the example.

Fix (small):
1. `auth.py`: register an `lpdaac` built-in credential provider — `get_daac_s3_credentials("LPDAAC")` already works (verified client-side: earthaccess returns the key/secret/token triple).
2. `gedi01b_waveform_healpix_hive.yaml`: declare `credentials_provider: lpdaac`.
3. Consider: the worker's per-granule catch that renders a credentials failure as "No data after filtering" cost this diagnosis a round trip — an auth/403 failure class deserves its own loud message (may be its own follow-up if the seam is deep).

No-redeploy workaround verified and now in the demo notebook: `Run.from_config(..., source_credentials=get_daac_s3_credentials("LPDAAC"))` — providers run orchestrator-side by design.

Refs #425 / PR #432 (the template), #434 (E2E validation — second catch), #448 (first catch).

espg:
Closes #448
Closes #449

Both closed issues are first-run findings from the GEDI smoke on the 0.45.0 fleet (2026-08-17) — template boundary conditions no prior config exercised — so they bundle per CLAUDE.md §5. A third GEDI-template item (the L2A filter set) was added to the bundle by espg in-session.

## Checklist

- [x] **#448 — dispatch payload is not NaN-safe.** `gedi01b_waveform_healpix_hive.yaml`'s five `.nan` fills → `"NaN"`, plus a load-time canonicalization so no future config can regress, plus a guard test over every packaged config.
- [x] **#449 — GEDI template ships without an LP DAAC credentials provider.** `lpdaac` built-in provider + the template declaration, plus (point 3) a distinct, loud failure class for denied reads.
- [x] **Drop the two L2A retrieval-fitness filters from the GEDI template** (espg ruling in-session 2026-08-17; Refs #425 / issue #422 — not a separate issue).

## #448 — NaN-safe dispatch payload

**What.** YAML `.nan` parses to a float NaN; `json.dumps` emits the non-standard token `NaN`; Lambda's parser refuses the whole payload (`InvalidRequestContentException`, `runner.py:5225`).

**Approach.** The string `"NaN"` is already the grammar's native form — it is the healpix/rect template default (`meta.get("fill_value", "NaN")`, `grids/healpix.py:674`), what `_is_nan_fill` compares against, and what the `aggregate`/`spill`/`streaming` fill checks test with `== "NaN"` (so the float form was *already* falling out of those branches). Two layers:

1. The template's five per-shot companions now declare `"NaN"`.
2. `config.load_config_from_dict` canonicalizes any float-NaN `fill_value`, at any depth, to `"NaN"`. That is the single funnel every config passes through — `load_config`, `default_config`, `zagg.client`, `client_transport`, and the worker's `load_config_from_dict(event["config"])` — so authored `.nan` stays legal but never reaches a wire payload. The walk returns the node **unchanged (same object)** when there is nothing to rewrite, so the nothing-to-do case allocates nothing and callers keep their dict identity (pinned by a test).

3. **Validate-time backstop** (review fold, `0a0baec2`). Canonicalization is scoped to `fill_value` on purpose — rewriting every float NaN in the tree would silently mangle authored values — so a NaN *elsewhere* (a filter `value`, a variable `attrs` entry) still reproduced #448. `validate_config` now calls `_validate_json_floats`, which walks `asdict(cfg)` and refuses any non-finite float, naming the path: `config value at data_source.filters[0].value is a non-finite float (NaN/Infinity), which strict JSON refuses -- the Lambda dispatch payload would be rejected (InvalidRequestContentException, issue #448)…`. `json.dumps(asdict(cfg), allow_nan=False)` sits behind the walk as the backstop; the walk runs first because it also catches `np.float32("nan")`, which `json.dumps` reports as an unserializable *type* rather than an out-of-range float. A `TypeError` from that call (a non-JSON *type*, e.g. a datetime) is deliberately swallowed — a different fault from the value question this check owns. The check runs before `validate_config`'s pipeline-kind branch, so temporal/event/raster configs are covered too. Same fold also fixed two walk sub-notes: tuples now traverse (tuple type preserved on rewrite), and `_is_float_nan` accepts `np.floating` so `np.float32("nan")` is caught.

**Tests** (`tests/test_config.py`): `.nan` from YAML and from a dict both normalize; a normalized config passes `json.dumps(asdict(cfg), allow_nan=False)`; `.nan` and `"NaN"` hash identically post-normalization; the string reaches both consumers (`_is_nan_fill` true, and the emitted healpix template array's `fill_value` is NaN); a NaN-free config is not copied. Guard: **every** packaged config (enumerated from `zagg.configs`, not hard-coded) survives `json.dumps(asdict(default_config(name)), allow_nan=False)` — that is the exact shape `runner._dispatch_lambda` builds and `_invoke_lambda_ping` / `_invoke_lambda` serialize.

## #449 — `lpdaac` credentials provider

**What.** `_resolve_source_credentials` (`runner.py:2649`) fell back to the NSIDC default because the template declared no provider; NSIDC temporary creds cannot read LP DAAC's `lp-prod-protected`, so every granule 403'd.

**Approach.**

1. `auth.get_lpdaac_s3_credentials()` = `get_daac_s3_credentials("LPDAAC")`, registered as `lpdaac` exactly mirroring the `nsidc`/`gesdisc` registrations.
2. The template declares `data_source.credentials_provider: lpdaac` with the bucket and the per-DAAC rule in a comment.
3. Docstring/doc touch-ups where the built-in list is enumerated (`_resolve_source_credentials`, `docs/temporal_events.md`).

**Tests** (`tests/test_auth.py`): the provider requests the `LPDAAC` DAAC; it is registered and resolves by name; and the packaged GEDI template both declares it *and* resolves through `runner._resolve_source_credentials` to LP DAAC creds — with `earthaccess.login` patched, so no network.

### Failure class for denied reads (issue #449 point 3) — included

The seam turned out to be **shallow and singular**: every per-granule and per-group failure already funnels through one closure, `_record_read_error` in `src/zagg/processing/worker.py`, and one message-composing branch ~40 lines later. Nothing in the reader internals had to change, so this is in-PR rather than a follow-up.

- New module-level `is_auth_failure(exc)` classifies the two **unambiguous** signatures: a botocore-shaped error (duck-typed on its `response` mapping, so **no botocore import** is added to the worker) carrying HTTP 401/403 or a denial code, and a specific token list in the message text. A bare `403` deliberately does **not** match — it would hit granule names.
- The **empty-body** signature `TypeError: memoryview: a bytes-like object is required, not 'NoneType'` (what a 403'd byte-range read looks like to the reader, which never sees the status code) is a **hint, not a classification** (review fold, `bf26298f`). `h5coro.s3driver.S3Driver.read` returns the same `None` from a bare `except Exception` as from `NoSuchKey`, so a missing object, a `SlowDown`/503, a timeout and a reset all raise it identically — and `spill.py`'s `memoryview(arr).cast("B")` raises it byte-for-byte from an unrelated cause. It lives in its own predicate `is_empty_body_failure`, never sets the auth class or `auth_errors`, and never flips the log to ERROR; instead the generic no-data message gains `N read(s) got an EMPTY body — likely a denied read (check data_source.credentials_provider names the DAAC hosting this product), a missing object, or throttling`, credentials first. The diagnostic value without the false certainty.
- `auth_errors` is counted alongside `read_errors`/`granule_errors`, **declared in `ProcessingMetadata`** (`src/zagg/schema.py`, review fold `618bf9c4` — it clears the `typeddict-unknown-key` mypy error and matches the rationale-comment style of its siblings), and lands in the result payload **even when the shard still produced data** (a partially-denied read is a partially-wrong product).
- **Run-summary warning for the partial case** (espg ruling in-session 2026-08-17, `624baecc`). An all-denied shard already announces itself through the worker's own error and `cells_error`; the silent case is the partial one — some granules read, some 403'd, so the shard returns data, no error, and a leaf missing exactly what the credentials could not reach. `runner._warn_partial_auth_denials` emits one dispatch-side WARNING naming the shard count, the summed auth-shaped failures, and the provider check; it is wired into both spatial summaries (local, from the result metas; lambda, from the worker envelopes).
- On the no-data path the message becomes its own class: `Access denied reading source granules (N auth-shaped failures; …): check data_source.credentials_provider names the DAAC hosting this product; e.g. …`, logged at ERROR. The fault is the config, not the data, and rerunning the shard just burns another invoke.
- Benign-error accounting is untouched: `dispatch.BENIGN_ERRORS` matches by **exact equality**, so the parenthesized no-data variant was never benign either. This changes the wording, not the pass/fail classification.

One existing test moved with the behavior: `TestGranuleScopeReadErrors::test_serial_granule_failure_is_counted_and_exemplared` raises `RuntimeError("h5coro open failed: expired token")` — literally the credential shape its own comment names — so it now asserts the auth-class message. The generic no-data wording stays covered by the non-auth cases in that class and by a new explicit test.

## GEDI template: drop the L2A retrieval-fitness filters (espg ruling, Refs #425 / issue #422)

`data_source.filters` keeps the two **validity** gates — `geolocation/degrade == 0` (geolocation validity) and `rx_assess/rx_clipbin_count <= 0` (saturation) — and drops the two **retrieval-fitness** gates, `quality_flag == 1` and `sensitivity >= 0.9`:

- They gate L2A *ground-retrieval* fitness, not waveform validity, and `quality_flag` already contains a sensitivity test internally, so the AND is brutally tight: measured on a real SERC granule, 1.61% of shots pass `quality_flag` and **0 of 434** in-shard shots survive the joint stack.
- zagg's own ratified clip policy *is* the signal criterion (issue #422: the declared false-positive operating point against each shot's noise model), so a second, differently-derived signal gate double-counts.
- Sensitivity gating biases the flux store against low-SNR / dense-canopy shots, asymmetrically vs the ATL03 products.
- Sensitivity stays per-shot recoverable downstream through the L2A join on `shot_number` — the asset join itself is untouched.

The rationale lives as a comment block in place of the removed filters, and the `assets.l2a` comment was corrected in the review fold (`73f6c9c0`) — after the drop the only consumed L2A dataset is the saturation counter, not "per-shot quality datasets". Tests updated in the same commit: `TestGediTemplate::test_template_reads_and_aggregates_the_fixture` now expects shots 101/102/104 (102 was excluded by `quality_flag`, 104 by `sensitivity: 0.50`; only 103 still goes, on `degrade`), with the per-shot companion assertions aggregating shot 101's rows since those companions are defined for a single-shot cell; `test_sibling_asset_filters_are_not_mapped` drops its now-vacuous `quality_flag` clause; and a new `test_filters_are_validity_gates_only` pins the exact remaining filter list so the gates cannot drift back in.

This also changes the GEDI template's semantic hash — same acceptance as the `credentials_provider` note below: no GEDI store exists yet.

## How it was tested

- `uv run pytest -q` → **3,996 passed, 38 skipped, 1 failed** (post-fold). The one failure is the known environment-only `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds` (needs the Docker/Lambda build toolchain); it fails identically on `origin/main` here.
- `ruff check src tests` / `ruff format --check src tests`: clean for every file this PR touches. Two failures exist on `origin/main` untouched by this PR and were verified pre-existing by stashing: `N818` on `src/zagg/registry.py:64` (`UnknownCapability`; CI's lint bot selects `E,F,W,I` only, so it does not fire there) and a `ruff format` diff in `tests/data/benchmark/README.md`.
- No fleet run: both fixes are client-side. The provider and the yaml are orchestrator-side (providers run orchestrator-side by design), and the NaN normalization runs at load — so **neither needs a redeploy**. The worker-side failure-class message does need the next release to be visible on the fleet; until then a denied run still shows the old wording from the deployed worker, with the same non-benign status.

## Questions for review

1. **`credentials_provider` is not in `DATA_SOURCE_PACKAGING_KEYS`, so declaring it moves the semantic hash.** It selects *how* source bytes are fetched, never *what* is computed, which reads as packaging (D19) — the same argument as `reader`/`driver`/`read_plan`. But it is not on that list today, so adding `credentials_provider: lpdaac` to the GEDI template **changes that template's semantic identity**. That is harmless right now (no GEDI store exists yet), which is exactly why it is worth deciding now rather than after the first build. I pinned today's behavior in `tests/test_semantics.py::test_credentials_provider_is_not_in_the_packaging_list` with a comment saying it is a pin, not an endorsement — if it is ruled packaging, that test flips to the equality assertion and the key joins the tuple. Not changed unilaterally: it would rehash any existing store whose config declares a provider (the temporal/MERRA-2 path can, per `docs/temporal_events.md`).
2. **Token list tuning for `is_auth_failure`.** The text-matching arm is deliberately conservative (no bare `403`). If there is a known LP DAAC / NSIDC denial wording it misses, it is a one-line addition to `_AUTH_ERROR_TOKENS`.
3. ~~**Should a shard that produced data but recorded `auth_errors` be louder than a payload counter?**~~ **Ruled yes** (espg, in-session 2026-08-17) and implemented in `624baecc` — see the run-summary warning above. The warning is dispatch-side (`runner._warn_partial_auth_denials`), wired into both the local and lambda spatial summaries, with unit tests plus one end-to-end test per backend.


src/zagg/processing/worker.py
```diff
@@ -65,6 +65,61 @@
...
+        meta = response.get("ResponseMetadata")
+        if isinstance(meta, dict) and meta.get("HTTPStatusCode") in (401, 403):
+            return True
+        error = response.get("Error")
+        if isinstance(error, dict):
+            code = str(error.get("Code", "")).lower()
+            if code in ("403", "401") or any(t in code for t in _AUTH_ERROR_TOKENS):
+                return True
+    text = str(exc).lower()
+    if isinstance(exc, TypeError) and "memoryview" in text and "nonetype" in text:
```

**The `memoryview`/`NoneType` arm is not a 403 signature — it is h5coro's signature for *every* failed S3 read, so missing granules, timeouts and throttling now get diagnosed as a credentials fault.**

`h5coro.s3driver.S3Driver.read` (installed 1.0.5, `.venv/lib/python3.12/site-packages/h5coro/s3driver.py:110-127`) returns `None` on two branches, and only one of them is auth-shaped:

```python
        try:
            response = self.client.get_object(...)
            return response["Body"].read()
        except self.client.exceptions.NoSuchKey:
            logger.error(f"S3 key {self.key} does not exist in bucket {self.bucket_name}")
            return None
        except Exception as e:
            logger.error(f"Error reading {self.key} from {self.bucket_name}: {e}")
            return None
```

The bare `except Exception` swallows the status code for *anything* — 404 `NoSuchKey`, `SlowDown`/503 throttling, read timeouts, connection resets, a bad endpoint — and hands the reader the same `None`, which surfaces as the same `TypeError: memoryview: a bytes-like object is required, not 'NoneType'`. `is_auth_failure` cannot tell those apart from the 403 case, so a shard that hit a throttling storm or a genuinely-absent granule is now logged at ERROR with `Access denied reading source granules ... check data_source.credentials_provider names the DAAC hosting this product`. That is exactly the misdiagnosis-costs-a-round-trip failure mode this change exists to remove, pointed the other way — and it is *louder* than the message it replaced, so it is more likely to be believed.

There is also a non-S3 producer of the identical string inside this repo: `_record_read_error` is called on fold-side failures too (`src/zagg/processing/worker.py:663`, the streaming/spill fold `except`), and `zagg/processing/spill.py:164` does `view = memoryview(arr).cast("B")` — a `None` array there raises the byte-for-byte same `TypeError` from a cause that has nothing to do with credentials.

Two ways out, both cheap: (1) drop the arm and rely on the botocore/`response` arm plus `_AUTH_ERROR_TOKENS` — h5coro logs the real boto exception at ERROR before returning `None`, so the operator has the code in CloudWatch either way; or (2) keep it but *downgrade* it — e.g. treat the None-body signature as "unclassified read failure" that adds a hint line to the generic message (`this can also be a denied read — check data_source.credentials_provider`) rather than asserting the auth class and switching the log level to ERROR. Either way the docstring at `worker.py:96-99` should say the signature is ambiguous, not that it *is* what a 403 looks like.

src/zagg/config.py
```diff
@@ -175,9 +175,57 @@ def load_config(path: str) -> PipelineConfig:
...
+
+
 def load_config_from_dict(d: dict) -> PipelineConfig:
     """Build a PipelineConfig from a plain dict (e.g. Lambda JSON payload).
 
+    Float-NaN ``fill_value`` declarations are canonicalized to the string
+    ``"NaN"`` here (issue #448) -- this is the single funnel every config
+    passes through (``load_config``, ``default_config``, the client, and the
+    Lambda worker's ``event["config"]``), so no config reaches a dispatch
+    payload carrying a value ``json.dumps(..., allow_nan=False)`` refuses.
```

**"no config reaches a dispatch payload carrying a value `json.dumps(..., allow_nan=False)` refuses" is false as written — the walk only canonicalizes keys literally named `fill_value`, so a float NaN anywhere else still reproduces #448 verbatim.**

`_has_nan_fill` / `_normalize_nan_fills` both gate on `k == "fill_value"`, which is the right *scope* (rewriting every float NaN in the tree would silently mangle authored values), but the docstring promises the general guarantee. Verified on this branch:

```
$ python -c "... load_config_from_dict(cfg_with_filter_value_nan) ...; json.dumps(asdict(cfg), allow_nan=False)"
1 filter value NaN: RAISES -> Out of range float values are not JSON compliant: nan
2 attrs NaN     : RAISES -> Out of range float values are not JSON compliant: nan
```

(case 1 is `data_source.filters: [{dataset: /x, op: le, value: .nan}]`; case 2 is a variable's `attrs: {scale: .nan}` — both survive normalization untouched and both die at `json.dumps`.) `TestPackagedConfigsAreDispatchable` does not cover this either, since it only enumerates the packaged tree, and the packaged tree has no such value.

Two clean options. Narrow the claim to what is enforced (`float-NaN **fill_value** declarations are canonicalized`) — cheap, and honest. Or, better, make the guarantee real at the same seam it is claimed at: a single `json.dumps(asdict(cfg), allow_nan=False)` in `validate_config` turns *any* residual non-JSON float into a loud, located config error at load time instead of an opaque `InvalidRequestContentException` one dispatch later — which is the actual user-facing win #448 is after, and it subsumes the packaged-config guard test.

Two smaller scope notes on the same walk, both only reachable from Python-built dicts (never from YAML/JSON), so probably fine to leave — but worth a line if you narrow the docstring anyway: `_has_nan_fill` recurses into `dict` and `list` only, so a `tuple` of dicts is skipped (verified: `({"fill_value": nan},)` passes through unchanged); and `_is_float_nan`'s `isinstance(value, float)` catches `np.float64` (a `float` subclass) but not `np.float32` (verified: `np.float32("nan")` survives as-is).

src/zagg/processing/worker.py
```diff
@@ -606,6 +669,10 @@ def _iter_granule_reads():
         metadata["read_errors"] = read_errors
     if granule_errors:
         metadata["granule_errors"] = granule_errors
+    if auth_errors:
+        # Counted even when the shard still produced data (a partially-denied
+        # read is a partially-wrong product), so the run summary can see it.
+        metadata["auth_errors"] = auth_errors
```
espg, line 702: 🤖 *from Claude (review)*

**`auth_errors` joins the wire payload without a declaration in `ProcessingMetadata`, unlike every other error counter it sits beside — and it adds a `typeddict-unknown-key` to mypy.**

`metadata` is annotated `ProcessingMetadata` (`worker.py:342`), and `src/zagg/schema.py:18-67` declares each sibling counter explicitly with a `NotRequired` and a rationale comment — `read_errors` (schema.py:44), `granule_errors` (schema.py:49), `read_error_exemplars` (schema.py:54). `auth_errors` gets neither, so the one place a consumer would go looking for the result-payload contract does not mention the field this PR adds to it.

Confirmed with the pinned checker (`.pre-commit-config.yaml` rev `v1.19.1`):

```
$ uvx mypy@1.19.1 src/zagg/processing/worker.py --ignore-missing-imports
src/zagg/processing/worker.py:675: error: TypedDict "ProcessingMetadata" has no key "auth_errors"  [typeddict-unknown-key]
```

Not a CI regression — mypy is already red on `main` here (109 errors, 18 files), and `worker.py:806` shows the same error for `time_range`, so there is precedent for the omission. But the fix is three lines in `schema.py` and it is what §4's match-the-surrounding-code asks for; the `time_range` precedent is a bug to not repeat rather than a pattern to follow.

src/zagg/processing/worker.py

Folded in `bf26298f` (worker.py, tests/test_processing.py) — option (2), downgraded to a hint.

Rationale for the downgrade rather than the drop: the diagnostic that motivated the arm is real (an operator staring at `TypeError: memoryview ...` has no idea credentials are even a candidate), but as you show, the signature cannot carry the certainty the message asserted. So the *hint* survives and the *classification* does not.

What changed:

- `is_auth_failure` now matches only what is unambiguous — the botocore-shaped 401/403/denial code, and `_AUTH_ERROR_TOKENS` in the exception text. The memoryview/None-body arm is gone from it.
- New sibling `is_empty_body_failure`, documented as ambiguous by construction, naming both producers you found (h5coro's bare `except Exception`, and `spill.py:164`'s `memoryview(arr).cast("B")` on a `None` array).
- `_record_read_error` counts the two separately; only `auth_errors` reaches the payload, so a 404/throttle/timeout-shaped `None` body no longer counts as auth and no longer flips the log to ERROR.
- The generic no-data message carries the hint, credentials first:

```
No data after filtering (2 granule reads raised; 2 read(s) got an EMPTY body — likely a
denied read (check data_source.credentials_provider names the DAAC hosting this product),
a missing object, or throttling; e.g. TypeError: memoryview: ...)
```

Tests: `test_memoryview_none_signature_is_a_hint_not_the_class` pins that a None-body failure does NOT count as auth and DOES still get the hint (with `credentials_provider` ordered before `missing object`); `test_empty_body_hint_reaches_the_log` pins WARNING, not ERROR; `test_classifier_units` now asserts `not is_auth_failure(none_body)` / `is_empty_body_failure(none_body)`. The partial-denial test switched its fixture to a definitive `AccessDenied`.

src/zagg/config.py

Folded in `0a0baec2` (src/zagg/config.py, tests/test_config.py) — your second option: the claim is now real rather than narrowed.

- New `_validate_json_floats(config)`, called at the **top** of `validate_config` (before the pipeline-kind branch and its early returns, so temporal/event/raster configs are covered too). It walks `asdict(cfg)` for a non-finite float, and raises naming the path:

```
config value at data_source.filters[0].value is a non-finite float (NaN/Infinity), which
strict JSON refuses -- the Lambda dispatch payload would be rejected
(InvalidRequestContentException, issue #448). Use the string "NaN" for a fill_value
declaration; elsewhere, drop the value or spell it as a string.
```

  The `json.dumps(asdict(cfg), allow_nan=False)` call is kept behind the walk as the backstop. The walk runs first because it also catches `np.float32("nan")`, which `json.dumps` reports as an unserializable *type*, not an out-of-range float. A `TypeError` from that call (a non-JSON *type*, e.g. a datetime) is deliberately swallowed — a different fault from the value question this check owns, and refusing it here would newly reject configs that never dispatch.
- `fill_value` canonicalization is untouched: it is the semantic normalization (one wire form, one semantic hash); this is only the backstop. `test_canonicalized_fill_value_still_validates` pins that a YAML `.nan` fill_value sails through.
- Both sub-notes fixed in the same commit: `_has_nan_fill`/`_normalize_nan_fills` traverse tuples (preserving the tuple type on rewrite), and `_is_float_nan` is now `isinstance(value, (float, np.floating))` so `np.float32("nan")` is caught.
- Docstring on `load_config_from_dict` rewritten: it states the `fill_value` scope, says why the scope is deliberate, and points at `_validate_json_floats` for the general guarantee.

Tests (`TestNonFiniteFloatsAreRefusedAtValidation`, 9 cases): your two reproductions (filter `value`, variable `attrs`) refused by path, plus Infinity, `np.float32` NaN, the temporal early-return path, the non-JSON-type pass-through, and the baseline-still-validates control; plus two cases in `TestNanFillCanonicalization` for the tuple and `np.float32` sub-notes.

src/zagg/processing/worker.py
espg: 🤖 *from Claude*

Folded in `618bf9c4` (src/zagg/schema.py) — `auth_errors: NotRequired[int]` now sits with its siblings, right after `granule_errors`, with the rationale line the neighbours carry: what it counts (definite matches only), why it is split out (the diagnosis inverts — config fault, so a retry just burns an invoke), that it is present only when non-zero *including* on data-producing shards, and that the ambiguous empty-body shape is deliberately not counted (see the reply on the classifier thread).

Verified with the pinned checker:

```
$ uvx mypy@1.19.1 src/zagg/processing/worker.py --ignore-missing-imports | grep typeddict-unknown
src/zagg/processing/worker.py:849: error: TypedDict "ProcessingMetadata" has no key "time_range"
```

`auth_errors` is gone from that list. `time_range` is left standing — pre-existing on `main` and out of this PR's scope; agreed it is a bug not to repeat rather than a pattern to follow.

PR #454, opened 2026-08-17
vlen route: derive the base extent from the link, expand at planned rate (issue #452)

### LINKED ISSUE #452: vlen route derives base extent as Σcount — strided products (GEDI L1B) silently read zero
🤖 *from Claude*

Root cause of the GEDI zero-read (the 0.45.0 fleet smoke: every shard `total_obs_read: 0`, store empty — the #434-class finding espg has been driving down all day). Diagnosed by driving the real worker entry (`zagg.processing.read._read_group` → `_vlen_read_group`) locally over h5coro's HTTPDriver against a real L1B+L2A pair.

**The bug: `read_vlen.py:309` derives the flat base extent as `n_base = Σ rx_sample_count` (the issue #43 contiguity assumption). GEDI L1B is STRIDED, not contiguous** — `rxwaveform` allocates a fixed 1,420-sample window per shot (`rx_sample_start_index` steps by exactly 1,420) while `rx_sample_count` is the valid count (61–1,420, typically ~700). Verified on three granules across the mission: `dims == n_shots × 1420` exactly; `Σcount` understates the array by 44–56%. Consequence in `read_plan.py:196-200`: every AOI run's `base_end` is clamped to the understated `n_base`, `base_end <= base_start`, ALL runs silently dropped → `parent_runs == []` → `obs_read 0`, no exception. The shipped synthetic fixture is perfectly packed (`WAVE len == Σcount`), which is exactly why this passed CI.

**Fix 1 (the extent):** replace Σcount with the link's true extent — `max((ibeg - index_base + cnt)[cnt > 0])` — best as a shared `link_base_extent()` helper beside `_broadcast_segment_to_base` (read.py:96); it reduces to Σcount for genuinely contiguous products (ATL03 no-op). Update the two prose blocks (read_vlen.py:98-108, 300-312) asserting `parent_idx == -1` rows are unreachable — under striding they are routine (143k of 310k planned rows on the test shard) and correctly dropped.

**Fix 2 (REQUIRED companion, or the fleet OOMs the moment fix 1 lands):** with a correct `n_base`, the vlen route still expands per-record values at FULL base rate — `_read_segment_broadcasts` (read_vlen.py:480-482) builds one length-n_base array per shot-level variable (6 companions × 195.8M = **7.0 GB per group**) and `_expand_mask_to_base` (:449-451, :458) allocates 195.8 MB per cross-level filter. Measured locally: **7.95 GB RSS for ONE beam group** vs `worker.memory: 2048` × `granule_workers: 4`. The gather map already exists at planned rate — broadcasts and masks should index through it (`seg_values[parent_safe]`; searchsorted-planned parent index for non-coordinate levels), per the module's own #43 OOM note.

**Verification evidence** (one beam, one granule, shard 5347294481781620745): 217 in-shard shots → 1 parent run → 310,427 planned rows (167,074 real + 143,353 slack, correctly dropped) → **165,506 rows in 217 distinct o18 cells**; gather alignment byte-checked against the raw product (origin-1 handling correct). Read amplification 1,420/~762 ≈ 1.86× (slack fetched then dropped) — note for the cost model. Regression fixture: add a STRIDED twin of tests/test_read_vlen.py's packed fixture (e.g. STARTS=[1,7,13,19], same COUNTS, WAVE len 24) asserting parity with the packed rows and no gap leakage.

**Worker-side both — needs a 0.45.1 release + fleet redeploy; no client-side workaround exists** (dropping `read_plan.spatial_index` hits the full-read arm's extent check instead: `record link does not tile the declared base extent`).

Repro/regression harness at `data/gedi_debug/gedi_drive_fixed.py` (drives the real read path over HTTPS; `--nofix` reproduces the zero; `--drop-quality` mirrors the fleet payload).

Refs #425 / PR #432 (the route), #43 (the contiguity assumption), #434 (the E2E that caught it), #448/#449/PR #450 (the sibling first-run findings).

espg:
Closes #452. Implements the [plan comment](https://github.com/englacial/zagg/issues/452#issuecomment-5311277217) (espg-directed, in-session 2026-08-17) — root cause diagnosed and verified there, not re-litigated here.

Two fixes, one PR: the extent derivation (fix 1) is what makes GEDI read anything at all; planned-rate expansion (fix 2) is its mandatory companion, since a correct extent with base-rate expansion trades the silent zero for a 7.95 GB worker OOM.

## What changed

**The extent.** `link_base_extent(index_beg, count, index_base)` — `max((ibeg - index_base + cnt)[cnt > 0])`, 0 for an all-empty link — now derives `n_base` on both planned routes (`read_vlen._vlen_read_group`, `read._planned_read_group`). GEDI L1B is strided: `rx_sample_start_index` steps by a fixed 1,420-sample window while `rx_sample_count` is the valid-sample count, so `Σcount` understates `rxwaveform` by ~46% and `read_plan.plan_read` clamped every AOI run to `base_end <= base_start`.

**Planned-rate expansion.** On the **planned** arm the vlen route no longer materializes anything of length `n_base`:

| expansion | before | after (planned arm) |
| --- | --- | --- |
| record-level variable broadcasts | `_broadcast_segment_to_base` → length-`n_base` array per companion (6 × 195.8M ≈ 7.0 GB/group) | gathered at planned rows (~11 MB) |
| cross-level structured filters | `_expand_mask_to_base` → 195.8 MB per filter | gathered per planned row (~310 KB) |
| sibling-asset (L2A) join mask | `_expand_mask_to_base` + slice | `record_mask[parent_safe] & valid` — the gather map the route already holds |

The **full** arm keeps the base-rate forms it always had (`_broadcast_segment_to_base`, `_expand_mask_to_base`) and never builds a row vector: it decodes every base row anyway, so there is no length-`n_base` intermediate to avoid, and routing it through the planned-rate helpers would *cost* — 8 B/row for the identity `arange` plus 33 B/row of `searchsorted` temporaries where the paint pays 1-4 B/row. That distinction was got wrong in phase 2 and corrected in the review fold; both arms are now pinned (see below).

Which owner map a consumer uses:

- **coordinates-level** consumers (the common case — all six of the shipped template's companions, and any filter on `coordinates.level`) reuse the gather map the route already holds, on both arms. One derivation, one truth.
- a **genuinely different** link derives its own: one `searchsorted` over that level's `index_beg` (`_link_parent_at_rows`) on the planned arm, the base-rate paint on the full one.

Every new helper is defined as exactly `<base-rate form>(...)[base_rows]` — fill semantics included (NaN for float dtypes on an unowned row, zero otherwise), validation scope included — and pinned that way in tests; the base-rate arrays were only ever consumed at the planned rows.

**Link disjointness is now enforced.** Non-overlapping records were the link grammar's standing precondition, assumed everywhere and checked nowhere — the paint maps *resolve* an overlap by paint order while the searchsorted derivation resolves it last-start-wins, so on a malformed product the two disagreed per row and silently emitted mismatched columns. `_validate_link_disjoint` refuses it where each level's link is first assembled (one stable argsort, at record rate).

**Prose.** The two blocks asserting `parent_idx == -1` rows are unreachable (read_vlen.py:98-108, 300-312) are rewritten: under striding those rows are routine (143,353 of 310,427 planned rows on the reference shard) and their drop is load-bearing, not defensive — GEDI zero-fills the slack in the product, so a leaked gap row would aggregate as a real zero-amplitude sample.

**One thing beyond the plan** (flagged below): the full-read arm now trims base-rate arrays to the link extent. A strided product's flat array is *longer* than the extent (tail padding past the last record's `count`), which would otherwise mismatch the gather map's length; a dataset *shorter* than the extent now raises with the gather map's wording instead of silently mis-shaping.

## Phases

- [x] **Phase 1** (93a14d69) — `link_base_extent` beside `_broadcast_segment_to_base`; used at read_vlen.py:309 and read.py:450; prose rewrite; full-arm extent trim. ATL03 no-op pinned: helper == `Σcount` on contiguous/origin-1/issue-#116-empty links, and the planned read's hyperslices are byte-identical (`TestLinkBaseExtent`).
- [x] **Phase 2** (359780f2) — `_link_parent_at_rows` / `_expand_mask_at_rows` / `_gather_segment_at_rows`; `_read_segment_broadcasts(..., base_rows=)`; vlen route rewired (`_to_planned` gone; `global_idx` is the arm switch, `None` on the full arm). Equivalence pinned against the base-rate twins (`TestPlannedRateExpansion`).
- [x] **Phase 3** (81adc27e) — strided twin of the packed fixture (`STARTS=[1,7,13,19]`, `COUNTS=[5,3,1,4]`, `WAVE` len 24, **slack zero-filled** as in the product): row-parity with the packed case, no slack leakage, planned/full parity, slack NaN posture under linspace synthesis, and the bug itself pinned — with `Σcount` as the extent the planned read returns `None` with `obs_read == 0`.
- [x] **Phase 4** (b52b7a8d, corrected in the fold) — memory pin: four shots strided across 3,000,004 base rows with 5 planned rows; a numpy proxy over `read`/`read_vlen`'s module-global `np` records every allocation those two modules make. `TestVlenMemoryPosture` now drives **both** arms — the planned arm asserts no allocation exceeds 64 elements, the full arm asserts no at-rows helper is called, no allocation exceeds `n_base`, and the base-rate allocation count stays bounded. Both verified to fail on their respective reverts (planned: `max = 3000004`; full: the six at-rows calls listed by name).
- [x] **Review fold** — six adversarial-review findings, one commit each: 6a544887 full-arm memory posture, 0f5b4477 link-disjointness validation, fd101fbc at-rows mask refusal scope, 50370083 planned-arm truncation guard, 3c4af324 third-level link coverage, d55ecb63 coordinates-level owner-map reuse.

## Evidence

Real granule, shard `5347294481781620745`, `GEDI01_B_2019121080625_O02166_02_T05019_02_005_01_V002`, BEAM0000, fleet payload (`--drop-quality`), driven through the real `_read_group` over h5coro's HTTPDriver:

| | before (main, `--nofix`) | after (this branch, no wrappers) |
| --- | --- | --- |
| `n_base` | 105,099,512 (`Σcount`) | 195,813,021 (link extent) |
| `parent_runs` | 0 | 1 — slice `(134026700, 134337127)` |
| planned rows | 0 | 310,427 (167,074 with a record + 143,353 slack) |
| `io_stats.obs_read` | 0 | 310,427 |
| rows returned | **0** | **165,506** |
| distinct o18 cells | — | **217** (= the 217 in-shard shots) |
| slack-sourced zero samples in output | — | **0** |
| peak RSS, one beam group | — | **382.8 MB** (vs 7.95 GB measured with base-rate expansion) |

Read amplification from fetching the strided slack: 310,427 / 167,074 ≈ 1.86×, as recorded on the issue for the cost model.

**Re-run post-fold**, same granule / shard / beam group, production code path with no wrappers: identical `n_base = 195,813,021`, 310,427 planned rows (167,074 with a record + 143,353 slack), **165,506 rows out**, 217 distinct o18 cells, identical column values — and `tracemalloc` peak 167.3 → 165.8 MB. All eight beam groups run clean at 102-124 MB peak each.

**Full-arm memory, before/after the fold** (`tracemalloc` peak, contiguous fixture so both revisions derive the same `n_base = 14,000,000`; full arm, one cross-level filter + two companions):

| | peak |
| --- | --- |
| phase 4 as pushed (b52b7a8d) | 2,352.1 MB |
| after the fold (d55ecb63) | **2,240.1 MB** |

The 112 MB difference is the identity `arange(n_base)` phase 2 introduced on the full arm; 2,240.1 MB is the pre-PR baseline to the tenth of a MB. Per-call, at `n_base = 19,998,560`: `_broadcast_segment_to_base` 80.0 MB vs `_gather_segment_at_rows` 660.7 MB, `_expand_mask_to_base` 20.0 MB vs `_expand_mask_at_rows` 660.6 MB — which is why the full arm keeps the paint.

## How tested

- `pytest -q` in the worktree: **4,003 passed, 38 skipped, 1 failed** — the failure is the known environmental `tests/test_lambda_build.py::TestFunctionBuild::test_function_build_succeeds`, unrelated and pre-existing.
- `ruff check` / `ruff format --check` clean on every file this PR touches. (Two pre-existing findings elsewhere on `main` under local ruff 0.16.3 — `src/zagg/registry.py:64` N818 and a reformat of `tests/data/benchmark/README.md` — are left alone per §4.)
- The real-granule drive above (network, run locally — never against the fleet), adapted from `data/gedi_debug/gedi_drive_fixed.py` with the fix wrappers removed since the production source now carries the fix. The unfixed baseline is that same harness on `main` with `--nofix`; the post-fold re-run above was A/B'd against a detached worktree at the pre-fold tip (b52b7a8d) with the same harness, and the two agree row for row.

## Questions for review

1. **Full-arm trim.** Trimming full reads to the link extent (and raising when a dataset is *shorter* than it) is beyond the plan's letter, but the full-read fallback is reachable on GEDI via `full_read_threshold` and would otherwise raise a shape mismatch on any strided product whose last window is padded. Keep as is, or would you rather the full arm read the dataset length and use `min(len(dataset), extent)`?
2. **Dense route left at base rate.** `_execute_plan_group` still expands cross-level masks to `n_base` and slices (read.py:651, :685). It is correct and cheap for ATL03 (bool/`float` at photon rate), and the plan scoped fix 2 to the vlen route, so I left it — the new helpers make it a two-line change if you want the allocation gone there too. Follow-up issue?
3. **`_expand_mask_to_base` kept.** The plan suggested renaming it if it no longer expands to base; the dense routes still call it at base rate, so it keeps its name and the planned-rate twin is `_expand_mask_at_rows`. Say the word if you'd rather have one function with a `base_rows=None` arm (the shape `_read_segment_broadcasts` took).
4. **Harness not ported into `tests/`.** The plan mentioned porting the real-granule case as a network-marked test; it needs EDL credentials plus a specific granule and shardmap, so I left it as the local harness and recorded the numbers above instead. Want it in the suite behind a marker + skip?
5. **Validation weakened by construction — partly answered in the fold.** With `n_base` derived from the coordinates link, the gather maps' "record does not tile the declared base extent" check can no longer fire for that link. The fold put two refusals back: 50370083 raises the same domain-worded error on the **planned** arm when a dataset comes back shorter than the plan's rows (it was an opaque `IndexError` from the column assembly before), and 0f5b4477 adds an overlap refusal the grammar never had at all. Net, the route catches strictly more than it did on `main` — but the question stands as a design point: is a link-derived extent the right posture, or should a product be able to declare its base extent and have the link checked against it?
6. **`read.py` is now 1,354 lines** (1,113 before; +241 for the planned-rate helpers, `link_base_extent`, and the fold's `_validate_link_disjoint` / `_gather_segment_at_parents`). That crosses the §4 raise trigger, so I am raising it rather than splitting anything: the new helpers are deliberately co-located with the paint-and-slice forms they mirror, since the pairs have to stay equivalent. Options — (1) leave as is for this PR and split in a follow-up, (2) move the four link helpers into a small `processing/links.py` that both `read.py` and `read_vlen.py` import, or (3) move them into `read_vlen.py` (695 lines) since only the vlen route uses the planned-rate arms. My preference is (2), as a separate PR so this one stays reviewable as a bug fix.


src/zagg/processing/read_vlen.py
```diff
@@ -357,9 +372,20 @@ def _read_full(paths: list[str]) -> dict:
...
+        """Global base index of each row this read decoded.
+
+        The plan's rows, or the identity map on the full arm (which is O(n_base)
+        by necessity anyway). Built lazily and once: a read with no cross-level
+        filter and no segment-level variable never needs it, and on the full arm
+        it is a sample-rate array.
+        """
+        nonlocal rows_cache
+        if rows_cache is None:
+            rows_cache = global_idx if global_idx is not None else np.arange(n_base, dtype=np.int64)
```

**The full arm's memory posture regressed by ~8-33x, and phase 4 does not pin it.** `_to_planned` was the *identity* on the full arm (`global_idx is None` → `return arr`), so the base-rate paint-and-slice cost exactly one array. `_base_rows()` replaces that identity with `arange(n_base)` and then feeds it to helpers whose temporaries are 5 int64 arrays wide.

Measured in this worktree, 2,000 records x count 700 on a 1,420 stride (`n_base = 19,998,560`), `tracemalloc` peak per call:

| | before (`9538d244`) | after (`b52b7a8d`, `base_rows = arange(n_base)`) |
| --- | --- | --- |
| one segment companion | `_broadcast_segment_to_base` **80.0 MB** | `_gather_segment_at_rows` **660.7 MB** (8.3x) |
| one cross-level filter | `_expand_mask_to_base` **20.0 MB** | `_expand_mask_at_rows` **660.6 MB** (33x) |
| `_base_rows()` itself | — (identity, 0 B) | **160.0 MB** (8 B/row) |

End-to-end A/B on a *contiguous* fixture (so both revisions derive the same `n_base = 14,000,000`), full arm, one cross-level filter + two companions: peak **2,240.1 MB → 2,352.1 MB**; an `np.arange` spy on `read_vlen`'s module globals confirms the extra 112 MB is `_base_rows()` allocating the full 14,000,000.

Scaled to this PR's own reference beam group (`n_base = 195,813,021`): `_base_rows()` alone is **1.57 GB** against `worker.memory: 2048`, and one companion gather peaks at ~6.5 GB — larger than the 7.0 GB base-rate figure the PR exists to remove. Q1 states the full arm is reachable on GEDI via `full_read_threshold`, and `TestPlannedVlenMemoryPosture` only drives the planned arm, so nothing in the suite sees this.

The fix keeps the identity where it already held — `base_rows=_base_rows() if global_idx is not None else None` for `_read_segment_broadcasts` (the kwarg already has the `None` arm), and `_expand_mask_to_base(..., n_base)` when `global_idx is None` for the cross-level loop.

src/zagg/processing/read.py
```diff
@@ -166,6 +166,135 @@ def _broadcast_segment_to_base(
...
+
+    The planned-rate twin of the paint-and-slice loops above: instead of
+    building a length-``n_base`` array and selecting the read's rows out of it,
+    each requested row's record is located directly, with one ``searchsorted``
+    over the link's starts. On the vlen route ``n_base`` is sample rate (~2·10^8
+    per GEDI beam group) while ``base_rows`` is the plan's rows (~3·10^5), so
+    the difference is a 2 GB worker surviving or not (the issue #43 OOM
+    posture, applied to the expansion side).
+
+    Records are assumed not to overlap — the link grammar's contract, which the
```
espg: 🤖 *from Claude (review)*

**"which the gather maps validate" is not true — neither gather map checks overlap, and on the input this line names the two owner derivations disagree silently.** `expand_link_indices` (read_vlen.py:118-140) and `_planned_gather_map` (read_vlen.py:167-179) validate only `beg >= 0` and `beg + cnt > n_base`; overlap is *resolved* by paint order, never rejected.

Run in this worktree, `index_beg=[0, 5]`, `count=[10, 2]`, `index_base=0`:

```
expand_link_indices parent: [ 0  0  0  0  0  1  1  0  0  0]   # no raise
_planned_gather_map parent: [ 0  0  0  0  0  1  1  0  0  0]   # no raise
_link_parent_at_rows      : [ 0  0  0  0  0  1  1 -1 -1 -1]
```

So inside one `_vlen_read_group` call, rows 7-9 keep record 0's coordinates and `synthesize_linspace` value (from the paint map) while the companion gather NaNs them and any record-level filter drops them — mismatched columns, no error. The `Identical to <base-rate form>[base_rows]` claims at :240 and :267 fail on the same input:

```
_broadcast_segment_to_base : [1. 1. 1. 1. 1. 2. 2. 1. 1. 1.]
_gather_segment_at_rows    : [1. 1. 1. 1. 1. 2. 2. nan nan nan]
_expand_mask_to_base   (mask [True, False]) : [1 1 1 1 1 1 1 1 1 1]
_expand_mask_at_rows                        : [1 1 1 1 1 0 0 0 0 0]
```

The derivation itself is right under the stated contract — I fuzzed 3,000 random *non-overlapping* links (unsorted starts, interleaved gaps, `count == 0` #116 empties, `index_base` 0 and 1) against both base-rate twins and got **0 divergences**. It is only the validation claim that is wrong. Either drop it ("assumed, unvalidated") or make it true — the cheapest check is one line in `link_base_extent`'s neighbourhood: sort the non-empty records by `beg` and assert `beg[order][1:] >= (beg + cnt)[order][:-1]`.

src/zagg/processing/read.py
```diff
@@ -166,6 +166,135 @@ def _broadcast_segment_to_base(
...
+def _expand_mask_at_rows(
+    coarse_mask: np.ndarray,
+    index_beg_arr,
+    count_arr,
+    index_base: int,
+    base_rows: np.ndarray,
+) -> np.ndarray:
+    """:func:`_expand_mask_to_base` restricted to ``base_rows`` (issue #452).
+
+    Identical to ``_expand_mask_to_base(...)[base_rows]`` — a row owned by no
```
espg, line 289: 🤖 *from Claude (review)*

**Not identical: the validation scope widened, so this can hard-fail a granule `_expand_mask_to_base` read fine.** In the base-rate form the `beg < 0` check sits *after* the predicate guard (read.py:74-76 — `for p, keep in enumerate(coarse_mask): if not keep: continue`), so a record the predicate rejects is never validated. `_expand_mask_at_rows` validates every non-empty record regardless of its verdict (`bad = (cnt > 0) & (beg < 0)`).

Run in this worktree — origin-1 link, record 0 non-empty with a malformed origin-0 start, filtered *out* by the mask:

```python
ibeg = np.array([0, 1]); cnt = np.array([2, 3]); mask = np.array([False, True])
_expand_mask_to_base(mask, ibeg, cnt, 1, 4)          -> [1 1 1 0]
_expand_mask_at_rows(mask, ibeg, cnt, 1, np.arange(4))
    ValueError: index_beg_arr[0]=0 is less than index_base=1
```

Reachability is low (it needs `count > 0` together with `index_beg < index_base`, which is not the #116 shape — that is `count == 0` *and* `ph_index_beg == 0`), but the direction is fail-hard on the production route and the docstring asserts equivalence. Either match the base-rate scope (`bad = (cnt > 0) & (beg < 0) & np.asarray(coarse_mask, dtype=bool)`) or say in the docstring that the check is deliberately stricter than the form it replaces.

src/zagg/processing/read_vlen.py
```diff
@@ -419,7 +445,23 @@ def _base_read_fn(path, hyperslice=None):
...
+        # strided product pads the tail of its last record's window (GEDI:
+        # 1,420 samples allocated, ``count`` valid). Those rows belong to no
+        # record, so trim them here and keep every base-rate array the same
+        # length as the gather map (issue #452). A dataset SHORTER than the
+        # extent is a real inconsistency (the link over-runs the array) and is
+        # reported with the gather map's wording.
+        arrays_by_path = {}
+        for path, arr in _read_full(base_paths).items():
+            arr = np.asarray(arr)
+            if len(arr) < n_base:
```
espg, line 471: 🤖 *from Claude (review)*

**The short-dataset guard is on the full arm only; the planned arm — the production GEDI route — surfaces the same corruption as an opaque `IndexError`.** Q5 is right that the "does not tile the declared base extent" check can no longer fire for the coordinates link. This block restores a domain-worded refusal for that, but only where `plan is None`.

The exact truncation `test_dataset_shorter_than_the_link_extent_raises` pins (`rxwaveform` cut to 20 rows, link reaches 21), driven through the planned arm instead (`read_plan={"spatial_index": "shots", "pad": 0}`, same strided fixture):

```
shard 1 / _LatGrid      -> IndexError: boolean index did not match indexed array along axis 0;
                           size of axis is 8 but size of corresponding boolean axis is 10
shard 1 / _LastShotGrid -> IndexError: ... size of axis is 2 but size of corresponding boolean axis is 4
```

It does raise, so this is diagnosability plus the asymmetry, not silent bad data — but on the fleet that traceback lands in a worker log with no hint that the granule's link over-runs its sample array, which is exactly the class of failure #452 spent a day isolating. The planned arm can't see the dataset length, but it can check the shape it does control, next to `arrays_by_path = _read_paths_pooled(...)` at :442:

```python
if len(arr) != len(global_idx):  # the gather map's length
    raise ValueError(f"dataset {path!r} returned {len(arr)} of {len(global_idx)} planned rows; "
                     f"the record link does not tile the declared base extent")
```

The strided suite would then cover both arms with the same wording.

tests/test_read_vlen.py
```diff
@@ -743,3 +743,307 @@ def test_template_reads_and_aggregates_the_fixture(self, cfg):
...
+        monkeypatch.setattr(rv, "link_base_extent", lambda ibeg, cnt, base: int(np.sum(cnt)))
+        assert _read_group(h5, "BEAM0000", ds, 1, grid, io_stats=io_stats) is None
+        assert io_stats["obs_read"] == 0
+        monkeypatch.undo()
+        df = _read_group(_FakeH5(_l1b_arrays_strided()), "BEAM0000", ds, 1, grid)
+        assert df["shot_number"].tolist() == [104] * 4
+
+    def test_record_level_filter_expands_at_planned_rate(self):
+        # degrade == 0 at the shots level drops shot 103's single sample; the
+        # cross-level expansion now gathers per planned row (searchsorted over
```
espg, line 1053: 🤖 *from Claude (review)*

**Coverage gap: the "searchsorted over the level's link" path is never exercised on a link that differs from the coordinates link.** This comment names the one genuinely new mechanism in phase 2 — a non-coordinate level deriving its own owner map — but the fixture makes it a no-op test of that property.

Every vlen data source in this file declares exactly two levels, `samples` (base) and `shots` (coordinates); `grep -n '"level":' tests/test_read_vlen.py` returns 129, 135, 302, 470, 881, 1015 and every one of them is `shots`. So `_expand_mask_at_rows` is only ever called with `ibeg_arr`/`cnt_arr`/`index_base` identical to the link the gather map already built `parent_planned` from, and the searchsorted derivation is compared against itself. A second level with a different `index_base`, a different stride, or a coarser count would pass this suite regardless.

Given the striding fixture already exists, the cheap addition is a third level in `_vlen_ds` — say `blocks`, one record per two shots, `index_base: 0`, starts `[0, 12]`, counts `[9, 10]` over the same 24-row `WAVE_STRIDED` — with a `{"level": "blocks", ...}` filter asserting the same rows drop as the equivalent `shots` filter. That is the only case in the PR where a wrong `index_base` on a second link would show up.

src/zagg/processing/read_vlen.py
```diff
@@ -477,8 +529,16 @@ def _base_read_fn(path, hyperslice=None):
...
 
-    seg_broadcasts = _read_segment_broadcasts(
-        h5obj, group, data_source, levels, n_base, read_fn=read_fn
+    # Record-level variable broadcasts, gathered straight onto this read's rows
+    # (issue #452): the base-rate form is one sample-rate array per companion
+    # variable — ~7 GB for GEDI's six against a 2 GB worker. ``_base_rows`` is
+    # only built when there is something to broadcast.
+    seg_broadcasts = (
+        _read_segment_broadcasts(
+            h5obj, group, data_source, levels, n_base, read_fn=read_fn, base_rows=_base_rows()
```

**The companions re-derive by searchsorted the owner map the route is already holding — six times per beam group, by a different algorithm than the one that produced the coordinates.** The sibling arm at :508 does this right (`record_mask[parent_safe] & valid`, reusing the gather map). This call does not: `_read_segment_broadcasts` loops variables and each one lands in `_gather_segment_at_rows` → `_link_parent_at_rows` over that level's link.

Instrumented on the strided fixture with the shipped template's shot-level `variables` block (`gedi01b_waveform_healpix_hive.yaml:56-63`, six companions, all on `shots` — which *is* `coordinates.level`):

```
template shot-level companions: 6
_link_parent_at_rows calls: 6 | distinct argument tuples: 1
```

Six identical searchsorted derivations over 310,427 planned rows per beam group, times eight groups per granule, for a map `parent_planned` already contains. The cost is small next to what the PR removed, but the second concern is not: `parent_planned` is built by paint order (`_planned_gather_map`, later record shadows earlier) while `_link_parent_at_rows` is built by last-start-wins, so the route now carries two independently-derived owner maps that have to agree — and they do not agree on overlapping records (see the comment on read.py:212, rows 7-9 of `ibeg=[0,5], cnt=[10,2]`). Passing `parent_planned` straight through when `f["level"] == coord_key` (and for the segment levels that equal it) collapses both problems.

src/zagg/processing/read_vlen.py

Fixed in 6a544887 — the full arm keeps the base-rate forms, and `_base_rows()` is gone entirely (there is no lazy row vector any more; `global_idx` is the arm switch, `None` on the full arm).

- `_read_segment_broadcasts(..., base_rows=global_idx)` — `None` on the full arm, so `_broadcast_segment_to_base` runs exactly as it did at `9538d244`.
- the cross-level loop branches on `global_idx is not None`: `_expand_mask_at_rows` on the planned arm, `_expand_mask_to_base(..., n_base)` on the full one.

**Your table reproduces exactly**, same fixture (2,000-record shape scaled to your `n_base`, count 700 on a 1,420 stride, `n_base = 19,998,560`, `tracemalloc` peak per call):

| | measured here |
| --- | --- |
| `_broadcast_segment_to_base` | **80.0 MB** |
| `_gather_segment_at_rows` | **660.7 MB** |
| `_expand_mask_to_base` | **20.0 MB** |
| `_expand_mask_at_rows` | **660.6 MB** |
| `arange(n_base)` (`_base_rows`) | **160.0 MB** |

End-to-end A/B on the contiguous fixture (`n_base = 14,000,000`, full arm, one cross-level filter + two companions), peak:

```
before this fold (b52b7a8d)  2352.1 MB
after  this fold (d55ecb63)  2240.1 MB
```

— i.e. back to your `9538d244` baseline to the tenth of a MB, the 112 MB `_base_rows()` allocation removed.

**The pin now drives both arms.** `TestPlannedVlenMemoryPosture` is renamed `TestVlenMemoryPosture` and gains `test_full_read_keeps_the_base_rate_forms`, which drives the same 3,000,004-row fixture through the full arm and asserts (a) zero calls to `_link_parent_at_rows` / `_expand_mask_at_rows` / `_gather_segment_at_rows`, (b) no allocation wider than `n_base`, (c) a bounded count of base-rate allocations. It pins dispatch rather than a byte count deliberately: end-to-end the full arm's peak is dominated by the base data itself (2,240 vs 2,352 MB is only 5%), so a threshold there would be flaky where the per-call ratio (8-33x) is not.

Both arms verified by revert-in-process on the folded tree:

```
full arm, _base_rows() + at-rows helpers restored:
  AssertionError: the full arm used the at-rows helpers:
  ['_expand_mask_at_rows', '_link_parent_at_rows', '_gather_segment_at_rows',
   '_link_parent_at_rows', '_gather_segment_at_rows', '_link_parent_at_rows']

planned arm, base-rate expansion restored (build at n_base, slice by global_idx):
  AssertionError: a base-rate allocation survived: [5, 3000004, 3000004]
```

One refinement lands on top in d55ecb63 (the comment on read_vlen.py:538): companions on the *coordinates* level now gather off the map the route already holds, on **both** arms. That preserves this finding's property — no row vector, no searchsorted at base rate — and is a shade cheaper than the paint it replaces (one fancy-index allocation, no per-record Python loop). Companions on any other level still take the arm conditional above.

Real granule, unchanged: `GEDI01_B_2019121080625_O02166_02_T05019_02_005_01_V002` / BEAM0000 / shard `5347294481781620745` gives identical `n_base = 195,813,021`, 310,427 planned rows, 165,506 output rows and identical column values before and after the fold; peak 167.3 → 165.8 MB.

src/zagg/processing/read.py
espg: 🤖 *from Claude*

Made it true rather than reworded — 0f5b4477.

New `_validate_link_disjoint(index_beg_arr, count_arr, index_base, level=None)` in `read.py`: one stable argsort over the non-empty records' starts, then `beg_sorted[1:] >= (beg + cnt)[order][:-1]`, raising with the level and both offending records named. It runs where each level's link is **first assembled**, so it is once per level per group, at record rate:

- `_vlen_read_group`, the coordinates link, next to `link_base_extent`
- `_vlen_read_group`, each cross-level filter's link (the non-coordinates branch)
- `_read_segment_broadcasts`, each segment level's link

`_link_parent_at_rows`' docstring now points at that instead of claiming the gather maps do it:

> Records must not overlap — the link grammar's contract, enforced by `_validate_link_disjoint` where each level's link is assembled, since this derivation (last start at or before the row wins) and the paint maps' (later record shadows earlier) agree only on disjoint records.

Your construction, pinned in `TestLinkDisjointValidation` — including your own divergence table as the *reason* the check exists, so the two derivations' disagreement is documented in the suite rather than only in this thread:

```
ibeg=[0, 5], cnt=[10, 2]
  expand_link_indices  -> [0 0 0 0 0 1 1  0  0  0]
  _link_parent_at_rows -> [0 0 0 0 0 1 1 -1 -1 -1]
  _validate_link_disjoint -> ValueError: link records overlap: record 0 covers
      base rows [0:10] but record 1 starts at 5; the link grammar requires
      disjoint record ranges
```

with `level='shots'` the message reads `link records on level 'shots' overlap: ...`. Also pinned: the shipped packed and strided fixtures pass, an unsorted link with gaps and an issue #116 empty passes, records that *touch* (`[0,5) [5,8) [8,10)` — the ATL03 shape) pass, and the full vlen route refuses an overlapping coordinates link end to end.

Two scoping notes:

1. I did **not** add it to the non-vlen photon route's own `_expand_mask_to_base` call sites (read.py:798, read.py:1154). Those only ever use the paint form, so there is no second derivation to disagree with, and widening the refusal onto the ATL03 production route is a behavior change beyond this finding. `_read_segment_broadcasts` does cover ATL03's segment links, and the full suite is green (4,003 passed).
2. d55ecb63 then removes most of the exposure anyway — coordinates-level consumers no longer derive a second map at all (your comment on read_vlen.py:538). The validation still earns its place for genuinely different links, and it is what lets the docstring's precondition be stated as fact.

src/zagg/processing/read.py
espg: 🤖 *from Claude*

Took the first option — matched the base-rate scope, so the equivalence claim stands as written. fd101fbc.

```python
keep = np.asarray(coarse_mask, dtype=bool)
...
bad = keep & (cnt > 0) & (beg < 0)
```

and the gather at the end reuses `keep` rather than re-casting `coarse_mask`. The docstring now says *why* the keep-mask is part of the test, so nobody tightens it back:

> Equivalence includes the *scope* of the `beg < 0` refusal: the base-rate form checks it inside a loop the predicate guard has already `continue`d past, so a record the predicate rejects is never validated. Widening that to every non-empty record would fail-hard on a granule the production route reads fine.

Your construction is the test (`TestExpandMaskAtRowsEquivalence`):

```
ibeg=[0, 1], cnt=[2, 3], index_base=1, mask=[False, True]
  _expand_mask_to_base  -> [ True  True  True False]
  _expand_mask_at_rows  -> [ True  True  True False]   (was: ValueError)
```

plus two guards so the narrowing does not become a hole: with the same record's verdict flipped to keep, **both** forms still raise `index_beg_arr[0]=0 is less than index_base=1`; and a parametrized equivalence sweep over the shipped strided fixture (all-keep, all-drop, alternating) asserts the two forms agree row for row.

Worth noting the reachability you flagged shrank further in 0f5b4477 and d55ecb63: a coordinates-level filter no longer reaches `_expand_mask_at_rows` at all (it reuses the gather map), so this path now only runs for a genuinely different link — but the equivalence claim is the docstring's, and it is now true unconditionally.

src/zagg/processing/read_vlen.py
espg: 🤖 *from Claude*

Added, essentially as you wrote it — 50370083. Right after `arrays_by_path = _read_paths_pooled(...)` on the planned arm:

```python
for path, arr in arrays_by_path.items():
    if len(arr) != len(global_idx):
        raise ValueError(
            f"dataset {path!r} returned {len(arr)} of {len(global_idx)} planned "
            f"rows; the record link does not tile the declared base extent"
        )
```

Driving your exact truncation (`rxwaveform` cut to 20 rows, link reaches 21) through the planned arm now gives:

```
ValueError: dataset '/BEAM0000/rxwaveform' returned 2 of 4 planned rows;
            the record link does not tile the declared base extent
```

instead of `IndexError: boolean index did not match indexed array along axis 0`.

`test_dataset_shorter_than_the_link_extent_raises` is parametrized `["full", "planned"]` and both arms match on the shared `does not tile the declared base extent` wording, so the phrase a fleet log gets grepped for is identical whichever arm the worker took.

One deviation from your sketch: the check loops every path rather than testing one, since the base read pulls variables *and* base-level filter datasets and a truncation can sit in either.

tests/test_read_vlen.py
espg: 🤖 *from Claude*

Added in 3c4af324 — a third level `blocks` over the existing 24-row strided base, with a link that is genuinely its own.

I made it harder than the `blocks` you sketched on one axis, because a level that merely *aggregates* shots still can't distinguish a right `index_base` from a wrong one at the boundaries that matter. The block edges cut **inside** a shot's samples:

```
row     0 .......................... 18 | 19 | 20 21 | 22 23
block   0 ...........................   | -- | 1 .... | -----
shot    101 .. | 102 .. | 103 |    | 104 ......... |

BLOCK_STARTS = [0, 20]   ORIGIN-0 (shots is origin-1)
BLOCK_COUNTS = [19, 2]   row 19 is owned by no block
BLOCK_FLAG   = [0, 1]
```

Keeping block 1 keeps base rows 20-21 only — the last two samples of shot 104 and nothing else, which no `shots`-level verdict can produce. Shard 1 / pad 0 plans rows 12..21, so the hand-computed expectation is:

```
_expand_mask_at_rows(flag, [0,20], [19,2], 0, arange(12,22))
  rows 12-18 -> block 0 (flag 0)      -> False
  row  19    -> no block              -> False
  rows 20-21 -> block 1 (flag 1)      -> True
  == [False]*8 + [True, True]
```

`TestThirdLevelLink` pins five things: that expectation by hand; equality with `_expand_mask_to_base(...)[planned_rows]`; the end-to-end planned read returning `shot_number == [104, 104]` and `rxwaveform == WAVE[11:13]`; planned/full parity on the new level; and — the property you actually named — that taking the level's origin from the coordinates link instead of its own **changes the answer** (`[False]*7 + [True, True, False]`, the window slid by a row). A fifth test asserts no `shots`-level filter reproduces those rows, so the fixture can't quietly decay back into a no-op.

The fixture is a local variant (`_vlen_ds_blocks` / `_l1b_arrays_blocks`) rather than a change to `_vlen_ds`, to keep the ~30 existing tests on the shared two-level source untouched.

`blocks` also earns its keep in d55ecb63: it is the case that proves the new coordinates-level reuse doesn't swallow a level whose link really differs — that test asserts exactly one `_link_parent_at_rows` call, with the blocks link's arguments.

src/zagg/processing/read_vlen.py

Done in d55ecb63, and I took the parenthetical too — the filter case as well as the segment levels.

Two changes:

- `_read_segment_broadcasts` gains `parents_by_level: dict[str, tuple[parent_safe, valid]]`. A level present there skips **both** the link read and the owner derivation and gathers straight off the supplied map, via a new `_gather_segment_at_parents` (the gather half of `_gather_segment_at_rows`, which now calls it too — one implementation of the fill semantics). `_vlen_read_group` passes `{coord_key: (parent_safe, valid)}`.
- the cross-level filter loop branches on `f["level"] == coord_key` and uses `coarse_fmask[parent_safe] & valid` — literally the sibling arm's line at :508. That branch also stops reading the level's `index_beg`/`count`, since they are the coordinates link already in hand.

Call counts pinned in `TestOwnerMapReuse`, instrumented the way you did (arguments recorded per call, so "distinct tuple" is visible):

```
shipped template, six shot-level companions + the shots-level degrade filter
  before:  7 calls / 1 distinct argument tuple   (your 6 + the filter)
  after:   0 calls          <- both arms, full and planned

blocks fixture (a genuinely different link)
  before:  3 calls
  after:   1 call, args = ((0, 20), (19, 2), 0, 10)   the blocks link
```

Verified by revert-in-process on the folded tree; the pre-fix run prints all seven calls carrying the identical `shots` link.

On the second concern — agreed, and that was the more important half. There is now one owner map for the coordinates level on the whole route, the paint-order one, and nothing re-derives it by last-start-wins. Where a second derivation genuinely is needed (a different link), 0f5b4477 refuses an overlapping link at assembly so the two rules cannot disagree there either.

Two notes on scope:

1. The reuse applies on the **full** arm as well, not only the planned one. `parent_planned` is the full-rate paint map there, so gathering off it is exactly `_broadcast_segment_to_base`'s answer for one fancy-index allocation instead of a per-record Python loop — same allocation width, no `arange`, no searchsorted. It does mean the full arm stops calling `_broadcast_segment_to_base` for coordinates-level companions; the memory property your read_vlen.py:387 comment asked for is preserved and pinned (see that thread).
2. `_gather_segment_at_parents` takes `(parent_safe, valid)` rather than the raw `parent` array on purpose — the route already holds both, and having it re-derive `np.where(valid, parent, 0)` would cost 8 B/row (1.57 GB on your reference beam group) to save a tuple.

Issue #460, opened 2026-08-17
inline backend: chunk-map B-tree walk strands 4 MiB h5coro cache lines per node (GEDI vlen shards OOM at any memory tier)


espg:
🤖 *from Claude*

## Symptom

Every GEDI first-flux-store attempt on `serc_gedi_flux.zarr` OOM'd — at **2 GB and 8 GB alike**, invariant to `streaming: spill`, `buffer_granules`, and worker count, in 115–870 s (fleet request IDs `dcb5809c…`, `06c2bc44…`, run `e3e06894` stats parquet). The read *plan* is healthy: `plan_read` selects ~310k samples/beam (`full=False`, one contiguous run, 217 shard shots) on the SERC o9 shards.

## Root cause

Two facts compose:

1. **`build_chunk_map` walks the chunk B-tree through h5coro's cached reader.** `_walk_chunk_btree` (`src/zagg/index/inline.py:140`) reads every node via `ds.readField`, which lands in `H5Coro.ioRequest` with `caching=True` — each touch fetches and **retains a 4 MiB cache line** (`CACHE_LINE_SIZE_DEFAULT = 0x400000`, h5coro `h5coro.py:52`) in a cache that is only released at granule close (`worker.py`, the issue #66 per-granule release).
2. **GEDI L1B B-tree nodes are scattered across a ~7 GB file.** `BEAM0000/rxwaveform` is `dims=(195,813,740), chunk=(14,200)` → **13,790 chunks**; the nodes rarely share a 4 MiB line, so the walk strands ~8 KB of node data per resident 4 MiB line.

Measured (local repro, `data/gedi_debug/gedi_drive_backend.py`, real `InlineIndex.read_group` over HTTPS): **one beam group = +531 MB retained RSS** (180 → 711 MB) for a read that yields 165,506 rows — byte-identical to the hierarchical path, which does the same group in seconds at trivial memory. Scaled: ~531 MB × 8 beams ≈ **4.2 GB per granule**, held until granule close, × `granule_workers: 4` in flight ≈ **17 GB** on an 8 GB Lambda. The OOM was structural before the first granule closed.

ATL03 never trips this: ~10× fewer chunks per dataset and denser files keep the same walk cheap — which is why the defect shipped invisible.

## Why the fleet ran inline at all

[...]

PR #461, opened 2026-08-17
small fixes 2026-08-17: read-plan 1-element runs, GEDI template right-sizing

espg:
Refs #460, Refs #425

Two espg-directed small fixes from today's GEDI flux runs (serial baseline run `8344cde5` and optimized probe `5d081b68`, both 4/4 shards green; prior failure rounds: #448 / #449 / #452 / #460).

## Phases

- [x] Phase 1 — `execute_read_plan`: survive genuine 0-d parts, keep failed (`None`) reads loud (issue #425) — reviewed, findings folded
- [x] Phase 2 — right-size `src/zagg/configs/gedi01b_waveform_healpix_hive.yaml` to the measured envelope + pin test (issue #460) — reviewed, findings folded

## Phase 1: 0-d parts made `np.concatenate` drop group reads

`execute_read_plan` (`src/zagg/read_plan.py`) built its parts with `np.asarray(read_fn(...), dtype)` and `np.concatenate`d them; a 0-d part raised `ValueError: zero-dimensional arrays cannot be concatenated`. On today's fleet run the error path was warn-and-continue, so **26 group reads were silently dropped** (e.g. `Error reading track BEAM0101: zero-dimensional arrays cannot be concatenated`, request `d9b1a833`, log group `/aws/lambda/process-shard-8192-disk`).

The adversarial review sharpened the mechanism (see the inline threads): h5coro returns shape `(1,)` for a single-element hyperslice — the actual 0-d source is a **failed h5coro read**, which comes back `None` and coerces to a 0-d NaN under `np.asarray(..., float)`. The fleet hit this via reads running off the link extent (the striding bug fixed for the local path in #452). The concatenate raise was accidentally load-bearing: `tests/test_index.py`'s chunk-boundary guard relied on it to keep a `None` read from becoming silent NaN data.

Shipped behavior, pinned by tests written failing-first:

- `None` from `read_fn` now raises an explicit `ValueError` on **both** exits (per-chunk and `full_read`) — a failed read stays loud instead of fabricating a NaN row.
- Genuine 0-d numeric parts (defensive, for scalar-returning `read_fn` wrappers) are promoted with `np.atleast_1d` and concatenate correctly; multi-element and 2-D (column-selector) parts are unchanged.

## Phase 2: GEDI template right-sizing

The template shipped the pre-measurement guesses (pooled aggregation, `granule_workers: 4`, 2048 MB, no index pin, a stale "hierarchical is the default" comment — absent `index` resolves to **inline** since #170, `src/zagg/index/__init__.py:235`). Measured envelope, optimized probe run `5d081b68` (identical output to the serial baseline: 23,353,274 obs / 27,727 cells on the reference shard):

| knob | before | after | evidence |
|---|---|---|---|
| `data_source.shard_workers` | `granule_workers: 4` (legacy key) | `4` (canonical key) | 300 s/shard at the 4096-disk tier vs 587–757 s serial on 8192 (`8344cde5`) |
| `data_source.index` | absent (= inline since #170) | `{backend: hierarchical}` pinned | inline's chunk-map B-tree walk strands 4 MiB h5coro cache lines; re-benchmark after #460 lands |
| `aggregation.streaming` | absent (pooled) | `{mode: spill, buffer_granules: 8}` | 50–82 M rows/shard pre-clip (21–33 M kept) — the pooled comment reasoned post-clip and was wrong four rounds running; spill wrote 1.6–2.5 GB /tmp, `spill_blocks_closed: 0` |
| `worker` | `memory: 2048` | `memory: 4096` + `extra_disk: true` | worker max_memory 2,464 MB / container HWM 2,472 MB (60% of tier); $0.069/run vs $0.286 at 8192/serial |

Fallback ladder (in the template comment, split by failure mode per review): a memory pinch drops `shard_workers` 4 → 2 before raising the tier; a `SpillOverflowError` needs the 8192-disk tier (worker count is inert for spill volume — the single-block threshold at 4096-disk is ~2,765 MB vs 1.6–2.5 GB measured spill, and these reducers cannot cross-block fold). `TestGediTemplate::test_measured_envelope` pins all four knobs.

**Clamp bug found by the phase-2 review (folded in `349f7e38`):** this template is the first shipped config on the canonical `shard_workers` key, and the issue #184 per-cell clamp (`runner.py` `_clamped_data_source`) wrote its clamped value back under the *legacy* `granule_workers` key while the worker prefers the canonical one — so the clamp was silently ignored for canonical-key configs. The write-back now targets whichever key the source dict carries; regression test in `tests/test_runner.py`.

**Interaction note:** a separate in-flight kernel PR adds temporal/toc keys to this same template file. This PR touches only the keys above; whichever lands second rebases mechanically.

## How tested

- `tests/test_read_plan.py`: regression tests written first and confirmed failing (0-d part → `ValueError` at the concatenate); now cover single-element runs, mixed runs, `None` on both exits, and 2-D parts. `tests/test_index.py` chunk-boundary guard green again.
- `tests/test_read_vlen.py::TestGediTemplate::test_measured_envelope` pins the four template knobs; template loads through `default_config` + `validate_config`.
- Local: `ruff check` / `ruff format --check` on all touched files, full `pytest` 4,16x passed. Known pre-existing local-only issues, not touched per §4: `test_lambda_build` (shells to `deployment/aws/build_function.sh`), `test_client_transport` poller wall-clock flake, a ruff N818 on `src/zagg/registry.py` and a format diff in `tests/data/benchmark/README.md` that appear only under the local ruff version (both files untouched by this PR).

## Questions for review

- **Rerun of affected stores:** every store written while the phase-1 bug was live silently dropped any group read whose h5coro read failed. The first flux store `serc_gedi_flux.zarr` is ~2% short (26 dropped group reads). Rerun the 4 shards to backfill, or accept the first store as a shakedown artifact and only rerun forward?
- **Per-part planned-extent check (review finding 1, declined in fold):** a length check of each part against the plan preempts the truncated-dataset contract pinned in `8cb820f5` (#452) — the fold left it standing rather than rewording that committed contract. Follow-up: add the missing length check to `read.py`'s planned arm and decide which arm owns the "does not tile the declared base extent" wording.

https://claude.ai/code/session_014qr6qZXy6dQRh8XGejzQKd

### DIFF
--- a/src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
+++ b/src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
@@ -22,8 +22,10 @@ data_source:
                                  # workers=4 on the 4096-disk tier runs 300 s/shard
                                  # at 60% memory (vs 587-757 s serial on 8192,
                                  # run 8344cde5) for identical output. If a denser
-                                 # AOI pinches 4096 at workers=4, drop
-                                 # shard_workers to 2 BEFORE raising the tier.
+                                 # AOI pinches 4096 on MEMORY at workers=4, drop
+                                 # shard_workers to 2 before raising the tier --
+                                 # but not for a SpillOverflowError, which this
+                                 # knob cannot move (see worker.extra_disk).
   reader: h5coro
   driver: s3
   # DAAC S3 credentials are per-DAAC: GEDI is LP DAAC's lp-prod-protected,
@@ -229,7 +231,36 @@ worker:
                                  # max_memory 2,464 MB / container HWM 2,472 MB
                                  # (60% of tier), $0.069/run vs $0.286 at
                                  # 8192/serial (run 8344cde5), identical output.
-                                 # Fallback ladder: drop shard_workers to 2
-                                 # before raising the tier.
-  extra_disk: true               # the -disk twin: spill needs up to 2.5 GB /tmp
-                                 # per shard (aggregation.streaming.mode: spill)
+                                 # Fallback ladder, BY FAILURE MODE:
+                                 #   memory pinch (OOM / container HWM near the
+                                 #     tier) -> shard_workers 4 -> 2, then raise
+                                 #     the tier; shard_workers is what bounds
+                                 #     concurrent in-memory read buffers.
+                                 #   SpillOverflowError (block close, see below)
+                                 #     -> the 8192-disk tier (10240 MB /tmp,
+                                 #     4608 MB threshold) or a finer
+                                 #     parent_order. shard_workers is INERT for
+                                 #     this one: it does not change total
+                                 #     spilled bytes, only in-flight buffers.
+  extra_disk: true               # the -disk twin sizes /tmp at memory + 2048 MB,
+                                 # so 6144 MB here. The spill block-close
+                                 # threshold is min(0.2 x memory x K, 0.45 x free
+                                 # /tmp); K = 4^(12-9) = 64 puts the memory term
+                                 # at ~52 GB, so /tmp always dominates -> ~2765
+                                 # MB. Measured spill is 1.6-2.5 GB/shard (run
+                                 # 8344cde5), a ~1.16x margin on the worst shard,
+                                 # and f_bavail is read live so anything else in
+                                 # /tmp eats into it. Clearing the threshold for
+                                 # a 2.5 GB block needs ~5.6 GB free /tmp, not
+                                 # 2.5 GB: /tmp holds the closing block beside
+                                 # the filling one, which is what the 45% is for.
+                                 # This is a cliff, not a slope, because the
+                                 # config is EXACT-SINGLE-BLOCK-ONLY --
+                                 # validate_spill_fold rejects it
+                                 # (build_waveform_digest and the
+                                 # single_shot_value companions have no
+                                 # cross-block fold law), so a threshold
+                                 # crossing raises SpillOverflowError instead of
+                                 # degrading. spill_blocks_closed: 0 is a
+                                 # PRECONDITION of these results, not merely an
+                                 # observation of them.
--- a/src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
+++ b/src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
@@ -18,7 +18,12 @@
 # would make `op: le, value: 0` admit or reject every shot), is what to check
 # there first.
 data_source:
-  granule_workers: 4
+  shard_workers: 4               # measured optimized envelope (run 5d081b68):
+                                 # workers=4 on the 4096-disk tier runs 300 s/shard
+                                 # at 60% memory (vs 587-757 s serial on 8192,
+                                 # run 8344cde5) for identical output. If a denser
+                                 # AOI pinches 4096 at workers=4, drop
+                                 # shard_workers to 2 BEFORE raising the tier.
   reader: h5coro
   driver: s3
   # DAAC S3 credentials are per-DAAC: GEDI is LP DAAC's lp-prod-protected,
@@ -26,8 +31,12 @@ data_source:
   # body, so every granule read raises memoryview(None) and the shard reports
   # "No data after filtering" (issue #449, first-fleet-run failure).
   credentials_provider: lpdaac
-  # No index block: the hierarchical route is the pinned baseline for the
-  # first GEDI builds; the vlen route also runs behind index: inline.
+  index:
+    backend: hierarchical      # pinned: an ABSENT block resolves to inline since
+                               # issue #170, and inline's chunk-map B-tree walk
+                               # strands 4 MiB h5coro cache lines on GEDI (issue
+                               # #460). Re-benchmark inline on GEDI after #460
+                               # lands before flipping.
   groups: [BEAM0000, BEAM0001, BEAM0010, BEAM0011, BEAM0101, BEAM0110, BEAM1000, BEAM1011]
   coordinates:
     # The base (sample) level carries no native coordinates: lat/lon expand
@@ -108,9 +117,17 @@ data_source:
     pad: 1
 
 aggregation:
-  # Pooled (no streaming block): o9 shards are light at GEDI density, and the
-  # flux field's composability is `none` initially (issue #422 -- no pyramid,
-  # no spill whitelist growth). Promote alongside overviews when needed.
+  # o9 shards are NOT light at GEDI density: the first successful run (run
+  # 8344cde5, issue #460) read 50-82 M rows/shard pre-clip (21-33 M kept) --
+  # the pooled-mode assumption here reasoned about post-clip density and was
+  # wrong four failure rounds running (#448/#449/#452/#460). Spill wrote
+  # 1.6-2.5 GB /tmp per shard with spill_blocks_closed: 0 (exact single-block
+  # regime, pooled-identical results). Pairs with worker.extra_disk below.
+  streaming:
+    mode: spill
+    buffer_granules: 8           # optimized probe (run 5d081b68): 8 with
+                                 # shard_workers 4 held container HWM at 2,472 MB
+                                 # (60% of the 4096 tier), identical output
   coordinates:
     morton:
       dtype: uint64
@@ -206,4 +223,13 @@ output:
   pyramid: false                 # composability none initially (issue #422)
 
 worker:
-  memory: 2048                   # pooled path, no spill: the light variant
+  memory: 4096                   # optimized probe (run 5d081b68, 4/4 green):
+                                 # process-shard-4096-disk at shard_workers 4 /
+                                 # buffer_granules 8 -> 300 s/shard, worker
+                                 # max_memory 2,464 MB / container HWM 2,472 MB
+                                 # (60% of tier), $0.069/run vs $0.286 at
+                                 # 8192/serial (run 8344cde5), identical output.
+                                 # Fallback ladder: drop shard_workers to 2
+                                 # before raising the tier.
+  extra_disk: true               # the -disk twin: spill needs up to 2.5 GB /tmp
+                                 # per shard (aggregation.streaming.mode: spill)
--- a/src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
+++ b/src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
@@ -121,8 +121,11 @@ data_source:
 aggregation:
   # o9 shards are NOT light at GEDI density: the first successful run (run
   # 8344cde5, issue #460) read 50-82 M rows/shard pre-clip (21-33 M kept) --
-  # the pooled-mode assumption here reasoned about post-clip density and was
-  # wrong four failure rounds running (#448/#449/#452/#460). Spill wrote
+  # the pooled-mode assumption here reasoned about post-clip density and
+  # undersized the envelope. The other GEDI failure rounds had unrelated
+  # causes (issue #449 DAAC credentials, issue #452 read-plan striding, issue
+  # #460 inline cache-line residency -- the index.backend: hierarchical pin
+  # above is what unblocked that one, not this streaming block). Spill wrote
   # 1.6-2.5 GB /tmp per shard with spill_blocks_closed: 0 (exact single-block
   # regime, pooled-identical results). Pairs with worker.extra_disk below.
   streaming:
--- a/src/zagg/read_plan.py
+++ b/src/zagg/read_plan.py
@@ -239,15 +239,44 @@ def execute_read_plan(plan: ReadPlan, read_fn, dataset_path: str, dtype) -> np.n
     np.ndarray
         Concatenated data for all runs in the plan, or an empty array if the
         plan matched nothing. If ``plan.full_read``, returns the full dataset.
+
+    Raises
+    ------
+    ValueError
+        If ``read_fn`` returns ``None`` -- h5coro's read-failure signal.
     """
     if plan.full_read:
-        return np.asarray(read_fn(dataset_path, hyperslice=None), dtype=dtype)
+        return np.asarray(
+            _require_read(read_fn(dataset_path, hyperslice=None), dataset_path, None), dtype=dtype
+        )
     if not plan.parent_runs:
         return np.empty(0, dtype=dtype)
-    # h5coro's readDatasets returns a 0-d array for a single-element
-    # hyperslice; np.atleast_1d keeps concatenate from raising on it.
+    # h5coro returns shape ``(1,)`` -- NOT a 0-d array -- for a single-element
+    # hyperslice: ``h5dataset.py`` reshapes only when ``ndims > 1``. So the
+    # ``np.atleast_1d`` below is cheap defense against a ``read_fn`` wrapper
+    # that hands back a scalar, not a fix for the reader itself.
+    #
+    # What actually produces a 0-d array is a FAILED read: h5coro warns and
+    # substitutes a null dataset, and ``np.asarray(None, dtype)`` is 0-d
+    # (``array(nan)`` for a float dtype). ``np.concatenate`` used to reject
+    # that as a side effect; ``_require_read`` now rejects it on purpose, so
+    # widening can never fabricate a NaN row and silently drop the group. A
+    # read that comes back SHORT is caught one level up, by the consumer that
+    # knows the declared extent (``read_vlen._read_group``, issue #452).
     parts = [
-        np.atleast_1d(np.asarray(read_fn(dataset_path, hyperslice=chunk), dtype=dtype))
+        np.atleast_1d(
+            np.asarray(
+                _require_read(read_fn(dataset_path, hyperslice=chunk), dataset_path, chunk),
+                dtype=dtype,
+            )
+        )
         for chunk in plan.chunk_lists
     ]
     return np.concatenate(parts)
+
+
+def _require_read(raw, dataset_path, hyperslice):
+    """Reject a ``None`` read (h5coro's failure signal) before dtype coercion."""
+    if raw is None:
+        raise ValueError(f"read returned None for {dataset_path} (hyperslice={hyperslice})")
+    return raw
--- a/src/zagg/read_plan.py
+++ b/src/zagg/read_plan.py
@@ -244,8 +244,10 @@ def execute_read_plan(plan: ReadPlan, read_fn, dataset_path: str, dtype) -> np.n
         return np.asarray(read_fn(dataset_path, hyperslice=None), dtype=dtype)
     if not plan.parent_runs:
         return np.empty(0, dtype=dtype)
+    # h5coro's readDatasets returns a 0-d array for a single-element
+    # hyperslice; np.atleast_1d keeps concatenate from raising on it.
     parts = [
-        np.asarray(read_fn(dataset_path, hyperslice=chunk), dtype=dtype)
+        np.atleast_1d(np.asarray(read_fn(dataset_path, hyperslice=chunk), dtype=dtype))
         for chunk in plan.chunk_lists
     ]
     return np.concatenate(parts)
--- a/tests/test_read_plan.py
+++ b/tests/test_read_plan.py
@@ -250,6 +250,41 @@ class TestExecuteReadPlan:
         expected = np.concatenate([data[5:10], data[20:25]])
         np.testing.assert_array_equal(out, expected)
 
+    def test_single_element_run_zero_dim_part(self):
+        # h5coro's readDatasets returns a 0-d array (numpy scalar) for a
+        # single-element hyperslice; np.concatenate raises on 0-d inputs
+        # ("zero-dimensional arrays cannot be concatenated"). Regression for
+        # the 2026-08-17 GEDI fleet run, which dropped 26 group reads this way.
+        data = np.arange(100.0, dtype=np.float32)
+        plan = self._make_plan([(42, 43)])
+
+        def read_fn(path, hyperslice=None):
+            lo, hi = hyperslice[0]
+            if hi - lo == 1:
+                return data[lo]  # 0-d numpy scalar, as h5coro returns
+            return data[lo:hi]
+
+        out = execute_read_plan(plan, read_fn, "/h", np.float32)
+        assert out.ndim == 1
+        np.testing.assert_array_equal(out, data[42:43])
+
+    def test_mixed_single_and_multi_element_runs(self):
+        # A single-element run alongside normal runs: order and values of the
+        # multi-element parts must be unchanged.
+        data = np.arange(100.0, dtype=np.float32)
+        plan = self._make_plan([(5, 10), (42, 43), (20, 25)])
+
+        def read_fn(path, hyperslice=None):
+            lo, hi = hyperslice[0]
+            if hi - lo == 1:
+                return data[lo]  # 0-d numpy scalar
+            return data[lo:hi]
+
+        out = execute_read_plan(plan, read_fn, "/h", np.float32)
+        expected = np.concatenate([data[5:10], data[42:43], data[20:25]])
+        assert out.dtype == np.float32
+        np.testing.assert_array_equal(out, expected)
+
     def test_full_read_plan_calls_without_hyperslice(self):
         data = np.arange(50.0, dtype=np.float32)
         plan = ReadPlan(
--- a/tests/test_read_plan.py
+++ b/tests/test_read_plan.py
@@ -251,17 +251,19 @@ class TestExecuteReadPlan:
         np.testing.assert_array_equal(out, expected)
 
     def test_single_element_run_zero_dim_part(self):
-        # h5coro's readDatasets returns a 0-d array (numpy scalar) for a
-        # single-element hyperslice; np.concatenate raises on 0-d inputs
-        # ("zero-dimensional arrays cannot be concatenated"). Regression for
-        # the 2026-08-17 GEDI fleet run, which dropped 26 group reads this way.
+        # h5coro itself returns shape (1,) for a single-element hyperslice
+        # (h5dataset.py reshapes only when ndims > 1), so this fake stands in
+        # for a scalar-returning ``read_fn`` WRAPPER, not for h5coro: a bridge
+        # that indexes rather than slices hands back a 0-d numpy scalar, and
+        # np.concatenate rejects 0-d inputs ("zero-dimensional arrays cannot be
+        # concatenated"). Pin that atleast_1d absorbs it.
         data = np.arange(100.0, dtype=np.float32)
         plan = self._make_plan([(42, 43)])
 
         def read_fn(path, hyperslice=None):
             lo, hi = hyperslice[0]
             if hi - lo == 1:
-                return data[lo]  # 0-d numpy scalar, as h5coro returns
+                return data[lo]  # 0-d numpy scalar from a wrapper that indexes
             return data[lo:hi]
 
         out = execute_read_plan(plan, read_fn, "/h", np.float32)
@@ -269,7 +271,8 @@ class TestExecuteReadPlan:
         np.testing.assert_array_equal(out, data[42:43])
 
     def test_mixed_single_and_multi_element_runs(self):
-        # A single-element run alongside normal runs: order and values of the
+        # A scalar-returning single-element run (see above -- a wrapper's shape,
+        # not h5coro's) alongside normal runs: order and values of the
         # multi-element parts must be unchanged.
         data = np.arange(100.0, dtype=np.float32)
         plan = self._make_plan([(5, 10), (42, 43), (20, 25)])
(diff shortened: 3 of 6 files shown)


src/zagg/read_plan.py
```diff
@@ -244,8 +244,10 @@ def execute_read_plan(plan: ReadPlan, read_fn, dataset_path: str, dtype) -> np.n
         return np.asarray(read_fn(dataset_path, hyperslice=None), dtype=dtype)
     if not plan.parent_runs:
         return np.empty(0, dtype=dtype)
+    # h5coro's readDatasets returns a 0-d array for a single-element
+    # hyperslice; np.atleast_1d keeps concatenate from raising on it.
     parts = [
-        np.asarray(read_fn(dataset_path, hyperslice=chunk), dtype=dtype)
+        np.atleast_1d(np.asarray(read_fn(dataset_path, hyperslice=chunk), dtype=dtype))
```
espg: 🤖 *from Claude (review)*

**The stated premise does not hold for the pinned reader, and this widening masks the read failure that actually produced the fleet error.**

h5coro 1.0.5 (pinned exactly for the layer, `pyproject.toml:109`) never returns a 0-d array for a single-element hyperslice. `H5Dataset` derives `shape` from the hyperslice extents and then does `values = numpy.frombuffer(buffer, dtype=datatype, count=elements)`, reshaping **only** when `ndims > 1` (`h5dataset.py:310-315`) — so a 1-element slice of a 1-d dataset comes back shape `(1,)`. Checked against the committed fixture (`tests/data/index/atl03_mini.h5`, `FileDriver`, this worktree's `.venv`):

```
h_ph  [(5, 6)]              -> ndarray (1,)
h_ph  [(0, 1)]              -> ndarray (1,)
signal_conf_ph [(5, 6)]     -> ndarray (1, 5)
h_ph  [(0, 1_000_000_000)]  -> None    # WARNING H5Coro encountered error reading gt1l/heights/h_ph:
                                       #         invalid hyperslice ...
                                       # np.asarray(None, dtype=np.float32) -> array(nan)  <- 0-d
h_ph  [(5, 5)]              -> None    # WARNING empty read: typeSize=4, numElements=0
```

What *does* produce a 0-d array here is a **failed or empty read**: `h5promise.datasetThread` catches `RuntimeError` (`FatalError` subclasses it) and substitutes a null dataset whose `.values is None` (`h5promise.py:39-44`); an out-of-range hyperslice is exactly such a failure (`h5dataset.py:205-208`), and an empty one returns early leaving `values = None` (`h5dataset.py:218-220`). `np.asarray(None, dtype)` is 0-d — `array(nan)` for a float dtype, `array(None, dtype=object)` when `dtype is None`, which is what `read.py:848-851` and `read_vlen.py:442-444` pass for variables and filter datasets.

So `ValueError: zero-dimensional arrays cannot be concatenated` on `BEAM0101` most plausibly means the planned hyperslice ran off the declared end of `rxwaveform` — the striding hazard of issue #452, which `read_vlen.py:447-459` assumes "comes back short". It does not come back short with h5coro: it errors and nulls.

With `atleast_1d` that same read now yields **one fabricated row** instead of raising:

- `read_vlen` planned arm: re-surfaces as the misleading `dataset ... returned 1 of N planned rows; the record link does not tile the declared base extent` (`read_vlen.py:454-459`) — an h5coro read failure reported as a link-tiling bug.
- `read.py` planned arm: no length check exists. A nulled coordinate read gives `len(lats) == 1`, `grid.assign` on NaN fails the shard mask, and `_planned_read_group` returns `None` (`read.py:829-838`) — the group is dropped *silently*, which is precisely the failure mode this phase set out to remove.

Suggested direction — reject rather than widen:

```python
raw = read_fn(dataset_path, hyperslice=chunk)
if raw is None:
    raise OSError(f"{dataset_path}: h5coro returned no data for hyperslice {chunk}")
part = np.asarray(raw, dtype=dtype)
```

plus a per-part length check against the planned extent (`plan.base_slices[i][1] - base_slices[i][0]`) — information this function already holds, and it catches short reads too. If a genuine 0-d return from some other reader is confirmed, guard it narrowly (`part.ndim == 0 and part.size == 1`) *after* the `None` check, and cite that reader in the comment.

Two follow-ons for the PR body if this holds: (1) the comment on lines 247-248 and the matching PR-body/commit claims about `readDatasets` need rewording to the mechanism actually observed — please link the traceback or log line that showed the 0-d; (2) the "Rerun of affected stores" question changes shape: if those 26 groups failed on out-of-range reads rather than 1-element runs, a rerun on this commit loses the same rows again, only without the error.

src/zagg/read_plan.py
```diff
@@ -244,8 +244,10 @@ def execute_read_plan(plan: ReadPlan, read_fn, dataset_path: str, dtype) -> np.n
         return np.asarray(read_fn(dataset_path, hyperslice=None), dtype=dtype)
```
espg: 🤖 *from Claude (review)*

The `full_read` arm keeps the un-normalized `np.asarray(...)`, so whichever policy the planned arm settles on, this branch now disagrees with it.

The same null-dataset source applies here: `read_fn(dataset_path, hyperslice=None)` is `h5obj.readDatasets([path])[path]` (`read.py:790-791`, `read_vlen.py:438-439`), which is `None` when h5coro's per-dataset thread catches the read error (`h5promise.py:39-44`). `np.asarray(None, dtype=np.float64)` is 0-d, and unlike the planned arm nothing downstream normalizes it — the first consumer is `_record_obs_read(io_stats, len(lats))` at `read.py:829`, which raises `TypeError: len() of unsized object`: an opaque worker-log failure for what is really "h5coro could not read this dataset".

Whatever lands for the planned parts (raise on `None`/0-d per the other finding, or `atleast_1d`) should cover this return too, so `execute_read_plan`'s documented contract — "Concatenated data for all runs in the plan" — is one shape law across all three exits rather than per-branch.

src/zagg/read_plan.py
```diff
@@ -244,8 +244,10 @@ def execute_read_plan(plan: ReadPlan, read_fn, dataset_path: str, dtype) -> np.n
...
     if not plan.parent_runs:
         return np.empty(0, dtype=dtype)
+    # h5coro's readDatasets returns a 0-d array for a single-element
+    # hyperslice; np.atleast_1d keeps concatenate from raising on it.
     parts = [
-        np.asarray(read_fn(dataset_path, hyperslice=chunk), dtype=dtype)
+        np.atleast_1d(np.asarray(read_fn(dataset_path, hyperslice=chunk), dtype=dtype))
         for chunk in plan.chunk_lists
     ]
     return np.concatenate(parts)
```
espg, line 275: 🤖 *from Claude (review)*

Fixing it at the concatenate leaves two of the three planned `read_fn` bridges uncovered, because they index the h5coro return **before** `execute_read_plan` ever sees it:

- `src/zagg/index/inline.py:810-811` — `arr = h5obj.readDatasets([...])[path]` then `parts.append(arr[s - lo :])`
- `src/zagg/processing/apriori.py:183-184` — same shape: `arr = h5obj.readDatasets([...])[path]` then `return arr[start - r0 :]`

On the null/empty read described in the other finding those raise inside `read_fn` — `TypeError: 'NoneType' object is not subscriptable` (`values is None`) or, for any reader that really does hand back a 0-d array, `IndexError: too many indices for array: array is 0-dimensional` — so `np.atleast_1d` at the concatenate cannot help them. That matters for GEDI specifically: `read_vlen.py:433-440` passes the backend's `read_fn` straight through when one is supplied, so the same plan run under `index: {backend: inline}` fails in a different place with a different message than under the hierarchical bridge this phase measured.

Worth pulling the normalization/validation into one helper in `read_plan.py` that all three bridges call (or validating inside `execute_read_plan` only after the bridges stop pre-slicing), so the contract lives in one place. Flagging rather than asking for scope creep: if you'd rather keep phase 1 to the concatenate, a one-line note here plus a follow-up issue would close it.

tests/test_read_plan.py
```diff
@@ -250,6 +250,41 @@ def read_fn(path, hyperslice=None):
...
+        # single-element hyperslice; np.concatenate raises on 0-d inputs
+        # ("zero-dimensional arrays cannot be concatenated"). Regression for
+        # the 2026-08-17 GEDI fleet run, which dropped 26 group reads this way.
+        data = np.arange(100.0, dtype=np.float32)
+        plan = self._make_plan([(42, 43)])
+
+        def read_fn(path, hyperslice=None):
+            lo, hi = hyperslice[0]
+            if hi - lo == 1:
+                return data[lo]  # 0-d numpy scalar, as h5coro returns
```
espg: 🤖 *from Claude (review)*

Both new tests pin a mechanism the pinned reader does not produce, and leave the one that does untested.

`data[lo]` is a `np.float32` scalar, so `np.asarray(..., dtype)` makes it 0-d and the test does fail before the fix — but real h5coro 1.0.5 returns shape `(1,)` for `hyperslice=[(42, 43)]` (`h5dataset.py:310-315`; verified on `tests/data/index/atl03_mini.h5` — see the finding on `src/zagg/read_plan.py`). So the fake's `if hi - lo == 1: return data[lo]` branch encodes a reader behavior that does not exist, and the comment "as h5coro returns" is wrong as written.

Two additions would make the suite pin real behavior:

1. **A `read_fn` returning `None`** — the actual 0-d source (`h5promise.py:39-44` substitutes a null dataset on any read error; `np.asarray(None, dtype=np.float32)` is `array(nan)`). Today that case silently becomes one fabricated NaN/`None` row; a test asserting whatever policy you land on (raise, per the other finding) is the regression guard that matters here.
2. **A 2-D part** — 2-D datasets do flow through `execute_read_plan`: `_variable_specs`/`_select_column` (`read.py:482-509`) and `_predicate_mask` (`read.py:512-526`) take a `column` selector into an N-D array (`signal_conf_ph`), and h5coro returns `(1, 5)` for a single-row hyperslice. `np.atleast_1d` is a no-op there, which is correct — but nothing in the suite pins it, so swapping it for a `reshape(-1)`/`ravel`-flavored normalization later would silently flatten `(n, 5)` to `(5n,)` and only surface as a confusing "'column' set but array is 1-D" from `_select_column`. A test with parts of shape `(k, 5)` asserting the output is `(n, 5)` locks that in cheaply.

src/zagg/read_plan.py
espg: 🤖 *from Claude*

Folded the premise correction and the `None` guard in bcf4db5f; **declined the per-part length check** — it collides with a contract pinned one commit before this PR. Detail below.

**Taken:** the empirical table is right, and the comment/commit premise was wrong. `read_plan.py` now rejects the real 0-d source instead of widening it:

```python
def _require_read(raw, dataset_path, hyperslice):
    """Reject a ``None`` read (h5coro's failure signal) before dtype coercion."""
    if raw is None:
        raise ValueError(f"read returned None for {dataset_path} (hyperslice={hyperslice})")
    return raw
```

applied on **both** exits (full-read and per-chunk), and the comment now says what you measured: h5coro returns `(1,)` for a single-element hyperslice (`h5dataset.py` reshapes only when `ndims > 1`), `None` is the failure signal, and `np.atleast_1d` survives only as cheap defense against a `read_fn` **wrapper** that hands back a scalar. `ValueError` rather than `OSError` to match the sibling wording in `read_vlen._read_group`; `tests/test_index.py::TestInlineReadGroup::test_chunk_boundary_start_succeeds_where_plain_hyperslice_fails` takes `pytest.raises(Exception)`, so the raise it depends on is preserved either way.

**Declined — the per-part extent check.** I implemented it (`expected = chunk[0][1] - chunk[0][0]` vs `len(part)` after `atleast_1d`, correct for 2-D since `chunk[0]` is the dim-0 range) and it broke a committed test: `tests/test_read_vlen.py::TestStridedRecords::test_dataset_shorter_than_the_link_extent_raises[planned]`, landed in 8cb820f5 for issue #452, which pins that *both* arms refuse a truncated dataset "in the same domain words":

```
Expected regex: 'does not tile the declared base extent'
Actual message: 'read returned 2 of 4 planned rows for /BEAM0000/rxwaveform (hyperslice=[(18, 22)])'
```

The check is strictly earlier and more precise, so it preempts that check rather than complementing it — retargeting the expectation is a deliberate revision of an issue-#452 decision, not a fold, and `tests/test_read_vlen.py` also carries unrelated in-flight edits in this worktree that must not enter a fold commit. Two things also lower its marginal value: with h5coro 1.0.5 an over-running hyperslice returns `None` (now caught), so the short-read shape is reachable only from fakes or another reader; and on `read.py`'s planned arm a *short* read is loud-but-opaque (mismatched lengths → boolean-index `IndexError`), not silent — the silent drop was the `None`→NaN path, which is closed.

Standing for @espg as a follow-up if wanted, in this order: add the missing per-part or per-path length check to `read.py`'s planned arm (the arm with no check at all), and decide which arm owns the wording for a truncated dataset — retargeting `tests/test_read_vlen.py:1185` accordingly.

**Follow-ons you flagged:** the PR body's `readDatasets` claim is corrected to the observed mechanism. The rerun question stands unanswered here and I've left it for @espg — if those 26 groups died on out-of-range reads, this branch now fails them loudly instead of dropping them, which changes what a rerun buys.

src/zagg/read_plan.py
espg: 🤖 *from Claude*

Fixed in bcf4db5f — folded into the same commit as the other read_plan finding, since it is one shape law and splitting it would have left an intermediate commit where the two arms disagreed.

The `full_read` arm now goes through the same guard as the planned parts:

```python
    if plan.full_read:
        return np.asarray(
            _require_read(read_fn(dataset_path, hyperslice=None), dataset_path, None), dtype=dtype
        )
```

so a nulled full read raises `ValueError: read returned None for <path> (hyperslice=None)` instead of reaching `_record_obs_read(io_stats, len(lats))` as a 0-d array and dying on `TypeError: len() of unsized object`. `test_none_full_read_raises` in `tests/test_read_plan.py` pins it; the docstring gained a `Raises` section so the contract is stated once for all three exits.

Note the shape law is now "reject `None`" only — the per-part *length* half of the sibling finding is declined for the reason spelled out there (it preempts the issue-#452 cross-arm wording pinned in `tests/test_read_vlen.py`, 8cb820f5).

src/zagg/read_plan.py
espg: 🤖 *from Claude*

Left by design — no code change, and I'm keeping the phase to the concatenate as you offered.

Both bridges you name fail **loudly** on the null read, which is the property that matters here: `TypeError: 'NoneType' object is not subscriptable` at `src/zagg/index/inline.py:810-811` and `src/zagg/processing/apriori.py:183-184`. Neither can fabricate a row or drop a group silently, so they are worse messages, not a correctness hole — unlike the `read.py` planned arm, where the widened NaN row did disappear quietly and is what bcf4db5f closes.

Against acting now: pulling the normalization into a shared helper means editing `inline.py` and `apriori.py`, both outside the small-fix bundle this branch carries, and the `read_fn` passthrough at `read_vlen.py:433-440` means a change there moves the GEDI inline arm's behavior — that deserves its own PR with its own fixture coverage rather than a fold commit.

One thing worth recording for whoever picks it up: I verified each bridge already returns exactly the requested extent (`inline.py` trims with `arr[s - lo:]`, `apriori.py` with `arr[start - r0:]`), so a single shared guard could validate all three uniformly once they stop pre-slicing.

Standing for @espg — happy to file the follow-up issue on his word, but per the repo conventions I don't open issues off a review-bot comment.

tests/test_read_plan.py
espg: 🤖 *from Claude*

Fixed in f3acd28a (the `None` test you asked for landed one commit earlier, bcf4db5f).

**Comments corrected.** The `# 0-d numpy scalar, as h5coro returns` lines were wrong and are gone. The tests stay, re-scoped to what they actually pin — a scalar-returning `read_fn` **wrapper**, which is the only thing `atleast_1d` now defends against:

```python
        # h5coro itself returns shape (1,) for a single-element hyperslice
        # (h5dataset.py reshapes only when ndims > 1), so this fake stands in
        # for a scalar-returning ``read_fn`` WRAPPER, not for h5coro: a bridge
        # that indexes rather than slices hands back a 0-d numpy scalar, and
        # np.concatenate rejects 0-d inputs ...
```

**(1) `read_fn` returning `None`** — `test_none_read_still_raises` and `test_none_full_read_raises` (bcf4db5f) pin the policy you recommended: raise, on both exits. Not a widened NaN row.

**(2) 2-D part** — `test_two_dim_parts_keep_their_columns`, built exactly as you described, including the single-row `(1, 5)` part:

```python
        data = np.arange(100.0, dtype=np.float32).reshape(20, 5)
        plan = self._make_plan([(2, 5), (12, 13), (17, 20)])
        ...
        out = execute_read_plan(plan, read_fn, "/h", np.float32)
        assert out.shape == (7, 5)
        expected = np.concatenate([data[2:5], data[12:13], data[17:20]])
        np.testing.assert_array_equal(out, expected)
```

Row order and values are asserted against the ground-truth concatenate, so a later `reshape(-1)`/`ravel`-flavored normalization flattening `(n, 5)` to `(5n,)` fails here rather than downstream in `_select_column`.

The one thing not folded is the per-part **length** check from the sibling finding — declined there, with the collision against the issue-#452 wording pinned in `tests/test_read_vlen.py` (8cb820f5) spelled out. `tests/test_read_plan.py` and `tests/test_index.py` are green (122 passed), and the full suite is green apart from two unrelated flakes noted for @espg.

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
```diff
@@ -18,16 +18,25 @@
 # would make `op: le, value: 0` admit or reject every shot), is what to check
 # there first.
 data_source:
-  granule_workers: 4
+  shard_workers: 4               # measured optimized envelope (run 5d081b68):
```
espg, line 21: 🤖 *from Claude (review)*

**Renaming to the canonical key silently disables the issue #184 per-cell clamp.**

`_clamped_data_source` writes the **legacy** key, but `_granule_workers` prefers the **canonical** one, so with `shard_workers` present the clamp is a no-op:

```
# src/zagg/runner.py:2761
return {**data_source, "granule_workers": clamped}

# src/zagg/processing/worker.py:181
w = data_source.get("shard_workers", data_source.get("granule_workers", 4))
```

Demonstrated on this branch:

```
>>> _clamped_data_source({"shard_workers": 4}, 1)
{"shard_workers": 4, "granule_workers": 1}     # both keys present
>>> _granule_workers(_)
4                                              # canonical wins -> clamp ignored
>>> _granule_workers(_clamped_data_source({"granule_workers": 4}, 1))
1                                              # legacy key: clamp honored
```

So before this commit the GEDI template got `min(4, n_granules)` per cell; after it, every 1- or 2-granule cell spins a 4-wide pool. `_clamped_data_source` is also the only thing that keeps unclamped cell payloads byte-identical (`runner.py:3063`, `runner.py:3675`), and the run-config `semantic_hash` is threaded specifically so the clamp cannot perturb leaf comparison — so the clamp going quiet is a real behavior change, not cosmetic.

`atl03_tdigest_healpix_hive.yaml:8` still uses `granule_workers`, which is why this has stayed invisible: **this is the first shipped config to use the canonical key**, so it is the first to hit the bug.

Suggested fix (one line + a test): make `_clamped_data_source` write back the key it read, e.g.

```python
key = "shard_workers" if "shard_workers" in data_source else "granule_workers"
return {**data_source, key: clamped}
```

and add a `tests/test_runner.py` case asserting `_granule_workers(_clamped_data_source({"shard_workers": 4}, 1)) == 1`. Alternatively hold the rename until the clamp is fixed. Either way the pin at `tests/test_read_vlen.py:817` (`"granule_workers" not in cfg.data_source`) cements the broken path, so it should land together with the fix.

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
```diff
@@ -206,4 +223,13 @@ output:
   pyramid: false                 # composability none initially (issue #422)
 
 worker:
-  memory: 2048                   # pooled path, no spill: the light variant
+  memory: 4096                   # optimized probe (run 5d081b68, 4/4 green):
```
espg, line 231: 🤖 *from Claude (review)*

**The tier drop halves the spill block-close threshold, and this config cannot survive a block close — so the margin against a hard failure goes from ~1.9x to ~1.16x.**

Two facts compose.

(1) **This config is outside `validate_spill_fold`.** Verified against the template on this branch:

```
$ python -c "... validate_spill_fold(load_config(...))"
spill blocks cannot cross-block fold this config: field 'rx_flux': ragged function
'zagg.stats.waveform.build_waveform_digest' has no fold law (only ...build_tdigest ...);
field 'shot_count': scalar function 'zagg.stats.waveform.shot_count' has no cross-block
fold (only 'len'/'count'/...pack_composition'); field 'shot_number': ... ; field 'noise_mean': ...
```

`SpillAggregator` accepts it (`spill.py:537` swallows the ValueError, exact in the single-block regime) but `_close_block` then **raises `SpillOverflowError`** at the first threshold crossing (`spill.py:743`). The pooled path this replaces had no such cliff — it just used more memory. So `mode: spill` here is not a strictly-safer swap: it trades a soft memory pressure for a hard shard failure once total spill exceeds one block.

(2) **The threshold is set by free `/tmp`, which the tier halves.** `_default_block_bytes` (`spill.py:464`) is `min(0.2 * memory * K, 0.45 * free_tmp)`. With `parent_order: 9` / `chunk_inner: 12`, K = `4**(12-9)` = 64, so the memory term is ~52 GB and **`/tmp` always dominates**. `-disk` sizes `/tmp` at memory + 2048 MB (`deployment/aws/template.yaml:180-182` `WorkerDiskTmp`):

| tier | `/tmp` | block threshold (0.45x) | worst measured spill | headroom |
|---|---|---|---|---|
| 8192-disk (baseline `8344cde5`) | 10240 MB | **4608 MB** | 2.5 GB (2384 MiB) | 1.93x |
| 4096-disk (this commit) | 6144 MB | **2765 MB** | 2.5 GB (2384 MiB) | **1.16x** |

So the shipped config runs at ~86% of a hard-failure threshold on the worst shard already measured, on a denser AOI it fails outright rather than degrading, and `f_bavail` is measured live so anything else in `/tmp` eats the remaining 16%.

(3) **The stated fallback ladder does not address this failure mode.** `shard_workers` changes nothing about total spilled bytes — the shard reads the same granules and writes the same rows either way; it only bounds concurrent in-memory read buffers. `SpillOverflowError`'s own remedy list is *"a bigger memory tier, a `-disk` variant with more ephemeral storage, or a finer `parent_order`"* — i.e. exactly the thing the comment says to try last. As written, an operator following the ladder on a block-close failure will drop to `shard_workers: 2`, see the identical `SpillOverflowError`, and have burned a fleet round.

Concretely, three things worth doing:

- Reword the ladder to split the two failure modes: **OOM / container HWM** -> `shard_workers` 4 -> 2; **`SpillOverflowError`** -> raise the tier or go finer on `parent_order` (`shard_workers` is inert for it).
- Record in the comment that this config is *exact-single-block-only* (no cross-block fold law for `build_waveform_digest` / the `single_shot_value` companions), so the `spill_blocks_closed: 0` telemetry is a **precondition**, not just an observation.
- Note the `/tmp` requirement correctly on line 234: a 2.5 GB single block needs ~5.6 GB free `/tmp` for the threshold to sit above it, not 2.5 GB — the 45% factor exists because `/tmp` holds the closing block beside the filling one.

Whether to also keep 8192 is an @espg call given the 4x cost delta; my read is the 4096 tier is right for the measured AOI but should be labeled as spending most of the block-close margin to get there.

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
```diff
@@ -108,9 +117,17 @@ data_source:
     pad: 1
 
 aggregation:
-  # Pooled (no streaming block): o9 shards are light at GEDI density, and the
-  # flux field's composability is `none` initially (issue #422 -- no pyramid,
-  # no spill whitelist growth). Promote alongside overviews when needed.
+  # o9 shards are NOT light at GEDI density: the first successful run (run
+  # 8344cde5, issue #460) read 50-82 M rows/shard pre-clip (21-33 M kept) --
+  # the pooled-mode assumption here reasoned about post-clip density and was
+  # wrong four failure rounds running (#448/#449/#452/#460). Spill wrote
```
espg: 🤖 *from Claude (review)*

**The "wrong four failure rounds running" attribution misstates three of the four root causes — including the one it cites for its own evidence.**

The comment blames the pooled-mode assumption for #448 / #449 / #452 / #460, but:

- **#449** was DAAC credentials — this very file says so 90 lines up (`gedi01b_waveform_healpix_hive.yaml:29-33`: *"the 403 comes back as an empty body ... the shard reports 'No data after filtering' (issue #449, first-fleet-run failure)"*). Pooling was not involved.
- **#452** was the read-plan striding / gather bug (cf. `8cb820f5`, `b6fafb92`, `eb9a543b` on this branch) — a correctness defect, not a memory-envelope one.
- **#460** is the *inline* backend stranding 4 MiB h5coro cache lines during the chunk-map B-tree walk, and its own diagnosis states the OOM was **"invariant to `streaming: spill`, `buffer_granules`, and worker count"** — i.e. issue #460 explicitly rules out pooled-vs-spill as its cause. What unblocked GEDI was the `index: {backend: hierarchical}` pin on line 35, not this streaming block.

That leaves #448 as the only round the pooled assumption could plausibly own. The PR body has this right (*"prior failure rounds: #448 / #449 / #452 / #460"* — a list of rounds, not of causes); the template comment converts a chronology into a causal claim.

This matters beyond tidiness: the next operator reading line 120-125 will conclude spill mode is what made GEDI work, and so will not suspect the hierarchical pin (or the inline eviction fix tracked on #460) when re-benchmarking. Suggest narrowing to what the evidence supports, e.g. *"the pooled-mode assumption reasoned about post-clip density and undersized the envelope (run `8344cde5` measured 50-82 M rows/shard pre-clip); the other GEDI failure rounds (#449 credentials, #452 read-plan striding, #460 inline cache-line residency) had unrelated causes."*

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
espg: 🤖 *from Claude*

Real bug, folded as suggested in `349f7e38`.

`_clamped_data_source` now writes back the key the source dict actually carries:

```python
key = "shard_workers" if "shard_workers" in data_source else "granule_workers"
return {**data_source, key: clamped}
```

so the merged dict never carries both names and `_granule_workers` (`worker.py:181`) resolves the clamped value. Your reproducer now gives `_granule_workers(_clamped_data_source({"shard_workers": 4}, 1)) == 1`.

Regression coverage lands in the same commit as `TestClampedDataSource::test_canonical_key_written_back` (`tests/test_runner.py`), mirroring the existing #184 cases: the canonical-key clamp at `n_granules` 1 and 2, `None` at/above K, no in-place mutation, plus the legacy-key path asserted to keep the legacy name (no canonical key injected). The `tests/test_read_vlen.py:819` pin (`"granule_workers" not in cfg.data_source`) is now correct rather than cementing the broken path — with the fix, a clamped GEDI payload stays on the canonical name only.

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
espg: 🤖 *from Claude*

Folded the three comment corrections in `a48d3725`; the tier itself stays at 4096-disk by espg direction (the optimized shape from run `5d081b68`, 4/4 green at exactly this memory/disk/workers/buffer combination), so this is a documentation fold, not a shape change. The cliff you found is now written down where an operator hits it.

What changed in `worker.extra_disk`:

- The headroom math, explicitly: `/tmp` is memory + 2048 MB = 6144 MB here; the threshold is `min(0.2 x memory x K, 0.45 x free /tmp)` with K = `4**(12-9)` = 64, so the ~52 GB memory term never binds and `/tmp` gives ~2765 MB against 1.6-2.5 GB measured spill — a ~1.16x margin on the worst shard, with `f_bavail` read live so other `/tmp` residents eat into it.
- The `/tmp` figure corrected per your line-234 point: clearing the threshold for a 2.5 GB block needs ~5.6 GB free, not 2.5 GB, because `/tmp` holds the closing block beside the filling one (that is what the 45% is for).
- `spill_blocks_closed: 0` is now labeled a **precondition**: the config is exact-single-block-only (`validate_spill_fold` rejects `build_waveform_digest` and the `single_shot_value` companions), so a crossing raises `SpillOverflowError` rather than degrading.

The fallback ladder is now split by failure mode in `worker.memory`: memory pinch -> `shard_workers` 4 -> 2 -> raise the tier; `SpillOverflowError` -> the 8192-disk tier (10240 MB `/tmp`, 4608 MB threshold) or a finer `parent_order`, with `shard_workers` marked **inert** for that mode since it moves in-flight buffers, not total spilled bytes. The `data_source.shard_workers` comment is scoped to match, so the drop-to-2 advice there no longer reads as a remedy for spill volume.

Whether to keep an 8192 variant alongside is left standing for espg, as you framed it.

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
espg: 🤖 *from Claude*

Agreed and reworded in `f18a1e80`, close to your suggested text. The `aggregation` comment now reads:

```yaml
  # o9 shards are NOT light at GEDI density: the first successful run (run
  # 8344cde5, issue #460) read 50-82 M rows/shard pre-clip (21-33 M kept) --
  # the pooled-mode assumption here reasoned about post-clip density and
  # undersized the envelope. The other GEDI failure rounds had unrelated
  # causes (issue #449 DAAC credentials, issue #452 read-plan striding, issue
  # #460 inline cache-line residency -- the index.backend: hierarchical pin
  # above is what unblocked that one, not this streaming block).
```

The density numbers are unchanged (50-82 M pre-clip, 21-33 M kept), and the "four failure rounds running" causal framing is gone — each round now names its own cause, with the hierarchical pin credited for issue #460 so the next operator re-benchmarking will not mistake spill mode for what made GEDI work.

Checked `tests/test_read_vlen.py` for the same claim: the pin comment at line 823 states only the density and the spill-vs-pooled choice ("50-82 M rows/shard pre-clip: spill, not pooled"), with no causal attribution to the failure rounds, so it needed no edit.

PR #462, opened 2026-08-17
inline backend: evict the chunk-map walk's h5coro cache lines (issue #460)

### LINKED ISSUE #460: inline backend: chunk-map B-tree walk strands 4 MiB h5coro cache lines per node (GEDI vlen shards OOM at any memory tier)

espg:
Closes #460

## What

The inline backend's chunk-map B-tree walk reads every node through h5coro's cached `ioRequest`, and `H5Coro.cache` retains a full 4 MiB cache line per touched byte range until granule close (`h5coro/h5coro.py` `ioRequest`, `CACHE_LINE_SIZE_DEFAULT = 0x400000`). On sparse files the nodes rarely share a line, so the walk strands ~a line per node: GEDI L1B `rxwaveform` (13,790 chunks) measured **+531 MB retained RSS per beam group**, ~4.2 GB/granule, OOM at every Lambda memory tier (the issue's diagnosis).

Fix per the [espg-directed plan](https://github.com/englacial/zagg/issues/460#issuecomment-5313837312): **snapshot-and-evict around the walk, zagg-side only** — snapshot `h5obj.cache` keys before the walk, drop every key the walk added once the `ChunkMap` is assembled. Pre-existing lines (the shared front-of-file metadata block) always survive; an evicted line a later read does want is simply re-fetched (B-tree nodes are metadata, not data, so in practice none is). After the review fold, eviction happens at two kinds of points, both single-threaded on the handle: immediately in `build_chunk_map` for its serial callers, and — for the lazy builds `read_fn` performs inside the pooled read fan-out — deferred to the end of `read_group`, after the thread pool has joined (evicting mid-pool races h5coro's unlocked cached reads; see the review thread).

## Phases

- [x] Phase 1 — snapshot-and-evict in `build_chunk_map` (`src/zagg/index/inline.py`): `_cache_keys` / `_evict_new_cache_lines` helpers, eviction scoped to the walk (the `H5Dataset` object-header parse's lines stay cached; measured bounded at ~tens of MB on GEDI, see evidence). Unit tests pin: walk-added lines evicted, pre-existing lines survive, map identical to an unevicted walk, rebuild re-fetches exactly the walk's lines.
- [x] Phase 2 — same eviction for the write-back full-coverage walk (`full_granule_maps` / `_iter_datasets`, the issue #190 path), per-dataset so the transient peak stays one walk deep.
- [x] Phase 3 — DEBUG-level lines-touched counter per map build + regression test asserting `len(h5obj.cache)` stays flat across N successive `build_chunk_map` calls.
- [x] Review fold, cycle 1 — all four findings folded (see the inline threads): live-dict snapshot fix, deferred eviction past the pooled fan-out + concurrency regression test, `cache_locks` pruned alongside, retained-parse-lines claim grounded with the GEDI measurement.
- [x] Review fold, cycle 2 — all four second-cycle findings folded: deferred-drain coverage on the planned and vlen routes (both tests fail with the drain neutralized), per-build DEBUG attribution on the deferred branch, honest wording for the concurrent sweep's scope, walk-line pins instead of count bounds.

## Acceptance evidence (real GEDI L1B over HTTPS)

Harness: `data/gedi_debug/gedi_drive_backend.py 0 0 BEAM0000` (real `InlineIndex.read_group`, EDL-authenticated HTTPS, granule `GEDI01_B_2019121080625_O02166_02_T05019_02_005_01_V002`), run unmodified for the before arm and as a scratch copy pointed at this branch's `src` for the after arm, augmented only to also report current RSS (`ps -o rss=`) and the post-group cache census. Identical output both arms: **rows=165506, distinct leaf cells=217**, `obs_read=310427`.

| metric (one beam group) | before (main) | after (this branch, post-fold) |
| --- | --- | --- |
| resident `h5obj.cache` lines after the group | **90 lines / 377 MB** (held to granule close) | **11 lines / 46 MB** |
| resident `cache_locks` entries | 90 | 11 (pruned with the lines) |
| peak RSS delta (`ru_maxrss`) | +525 MB (issue recorded +531) | +533 MB (the group's walk lines, freed when the pool joins) |
| current RSS after the group | 705 MB | 596 MB |

Notes: `ru_maxrss` is a high-water mark, so the transient walk (~90 scattered lines fetched once) keeps the *peak* similar by design — the plan's "transient peak becomes one walk's lines" (pre-fold, with immediate per-dataset eviction, the peak measured +420; the review fold's deferred eviction holds the group's walk lines until the read pool joins, trading that back for thread-safety). The retained-side win is the cache census: 377 MB → 46 MB per beam group, which is what accumulated ×8 beams ×`granule_workers` into the fleet OOM. The macOS current-RSS numbers understate the recovery because the allocator keeps freed pages resident (a probe run that cleared the entire cache *and* dropped the result dataframe recovered 0 MB of current RSS on macOS); on the Lambda fleet (glibc) freed 4 MiB blocks are returned via munmap.

## How it was tested

- `TestChunkMapCacheEviction` + full-coverage-walk eviction test in `tests/test_index.py`: real h5coro `H5Coro` over the committed fixture opened with **512-byte cache lines** (at the default 4 MiB the whole 86 KB fixture is one line, so walk-added vs pre-existing keys would be indistinguishable) plus a counting `FileDriver` subclass. Pins: retained key set == exactly the parse's lines and none of the walk's; map identical to an unevicted walk; rebuild's extra driver reads == exactly the walk's line count; `len(h5obj.cache)` (and `cache_locks`) flat across repeated builds; concurrency regression test driving `read_fn` lazy builds + `readDatasets` on one handle across threads.
- `pytest` full suite green locally except the known-environmental `test_lambda_build` failure (pre-existing, unrelated — Docker-dependent); `ruff check` / `ruff format --check` clean (see below), CI green on every push.

## Questions for review

- `ruff check src tests` reports one pre-existing finding on clean `origin/main`: `N818` on `UnknownCapability` (`src/zagg/registry.py:64`) — untouched here, flagged not fixed (per §4). The PR lint bot runs `--select=E,F,W,I` (no `N`), which is why it never surfaced on PRs.
- The upstream h5coro `caching=False` metadata-read mode (the structural fix) remains a follow-up to raise after this lands, per the plan.
- Standing from review cycle 2 (left by design, see the thread): under concurrent lazy builds the deferred drain can sweep another dataset's object-header lines (bounded re-fetch cost); making exact parse-line retention a hard invariant via a floor subtraction was declined as extra machinery — it becomes moot with the upstream `caching=False` mode.

### DIFF
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -184,12 +184,45 @@ def _walk_chunk_btree(ds) -> list[tuple[tuple[int, ...], int, int, int]]:
     return entries
 
 
+def _cache_keys(h5obj) -> set:
+    """Snapshot of ``h5obj.cache``'s keys (empty for cache-less handles)."""
+    return set(getattr(h5obj, "cache", None) or ())
+
+
+def _evict_new_cache_lines(h5obj, before: set) -> int:
+    """Drop every h5coro cache line added since ``before`` was snapshotted.
+
+    The snapshot-and-evict half of issue #460: ``H5Coro.cache`` retains a full
+    ``cacheLineSize`` line (4 MiB default) for every byte range a cached
+    ``ioRequest`` touches, and releases nothing until granule close. A chunk
+    B-tree walk reads each node exactly once, but on a sparse file the nodes
+    rarely share a line (GEDI L1B ``rxwaveform``: 13,790 chunks → +531 MB
+    retained per beam group, ~4.2 GB per granule), so the walk's lines are
+    pure dead weight afterwards. Evicting them caps residency at one walk's
+    transient peak; a later read that does want an evicted line just
+    re-fetches it (B-tree nodes are not data, so in practice none does).
+    Only keys absent from ``before`` are dropped — pre-existing lines (the
+    front-of-file metadata block every parse shares) always survive. Returns
+    the number of lines evicted.
+    """
+    cache = getattr(h5obj, "cache", None)
+    if not cache:
+        return 0
+    added = [k for k in cache if k not in before]
+    for k in added:
+        cache.pop(k, None)
+    return len(added)
+
+
 def build_chunk_map(h5obj, path: str) -> ChunkMap:
     """Build a :class:`ChunkMap` for one dataset by walking its metadata.
 
     Metadata-only: no chunk is ever read or decompressed. A contiguous-layout
     dataset yields a single pseudo-chunk covering the full first axis
     (mirroring h5py's ``get_offset()`` treatment in the bench extractor).
+    Cache lines the B-tree walk pulls into ``h5obj.cache`` are evicted before
+    returning (issue #460) — the object-header lines the ``H5Dataset`` parse
+    touched stay cached, since every dataset in the group shares them.
 
     Raises ``KeyError`` for an absent path (h5coro's ``metaOnly`` traversal
     never raises on its own — it just leaves default metadata) and
@@ -200,7 +233,11 @@ def build_chunk_map(h5obj, path: str) -> ChunkMap:
     ds = H5Dataset(h5obj, path, earlyExit=True, metaOnly=True, enableAttributes=False)
     if ds.meta.typeSize == 0:
         raise KeyError(path)
-    return _chunk_map_from_dataset(h5obj, ds, path)
+    before = _cache_keys(h5obj)
+    try:
+        return _chunk_map_from_dataset(h5obj, ds, path)
+    finally:
+        _evict_new_cache_lines(h5obj, before)
 
 
 def _chunk_map_from_dataset(h5obj, ds, path: str) -> ChunkMap:
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -387,6 +387,13 @@ def full_granule_maps(h5obj, existing: dict[str, ChunkMap]) -> dict[str, ChunkMa
     for path, ds in _iter_datasets(h5obj):
         if path in maps:
             continue
+        # Per-dataset snapshot-and-evict, same as build_chunk_map (issue
+        # #460): this walk maps EVERY dataset in the granule, so unevicted it
+        # accumulates worst of all. Evicting after each dataset keeps the
+        # transient peak one walk deep; the enumeration's own object-header
+        # lines are in each snapshot (the generator parses the next node
+        # before the loop body runs) and stay cached for the datasets after.
+        before = _cache_keys(h5obj)
         try:
             maps[path] = _chunk_map_from_dataset(h5obj, ds, path)
         except ValueError:
@@ -397,6 +404,8 @@ def full_granule_maps(h5obj, existing: dict[str, ChunkMap]) -> dict[str, ChunkMa
             # there): drop just this dataset from the manifest instead of
             # sinking the whole granule's write-back.
             logger.warning(f"  write-back: no chunk map for {path} ({exc}); skipping")
+        finally:
+            _evict_new_cache_lines(h5obj, before)
     return maps
 
 
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -300,17 +300,25 @@ def build_chunk_map(h5obj, path: str, *, evict_into: set | None = None) -> Chunk
     ever touched across the group, against the hundreds of MB the unevicted
     walks stranded. The absent-path ``KeyError`` below returns before the
     snapshot on purpose: nothing walk-touched exists yet, and its parse lines
-    are retained by the same choice.
+    are retained by the same choice. That retention is exact only on the
+    immediate path (single-threaded); the deferred path below drops some parse
+    lines too, so the GEDI figure above — measured through it at the default
+    ``read_workers`` — is if anything the lower residency of the two.
 
     ``evict_into`` defers that eviction for a caller whose builds run
     concurrently on one handle: pass a set and the walk's keys are *added* to
     it (a GIL-atomic ``set.update``) instead of popped, for the caller to
     evict once its fan-out has joined — see :func:`_evict_deferred_lines`.
-    Keys other threads' reads happen to add inside the same window are
-    over-attributed to the walk, which is harmless: they are still this
-    group's lines and are dropped only after the pool joins. The default
-    (``None``) evicts immediately, which is what every single-threaded caller
-    wants (``_prebuild_group_maps``, the full-coverage walk, direct use).
+    The window is a plain before/after key diff, so lines *other threads* add
+    inside it are swept in as well — measurably, including other datasets'
+    object-header parse lines (review finding on PR #462: 23 of 40 concurrent
+    rounds on the test fixture). Those lines re-fetch on next touch (one small
+    ranged read each), and that bounded cost is the deliberate price of a
+    race-free drain: the alternative — subtracting a floor of every build's
+    ``before`` snapshot — is extra machinery to protect what is a memory *win*.
+    The default (``None``) evicts immediately, which is what every
+    single-threaded caller wants (``_prebuild_group_maps``, the full-coverage
+    walk, direct use).
 
     Raises ``KeyError`` for an absent path (h5coro's ``metaOnly`` traversal
     never raises on its own — it just leaves default metadata) and
@@ -435,7 +443,12 @@ def _iter_datasets(h5obj):
     whole enumeration is served from the front-of-file metadata block h5coro
     already cached — one ranged GET, no chunk data (issue #190; the full walk
     stays inside the single ~4 MiB cache line the config's read already paid
-    for, measured on real ATL03). A node whose ``typeSize == 0`` is a group
+    for, measured on real ATL03). Since issue #460 that "already cached" is a
+    best case rather than a guarantee: each group's deferred drain may have
+    swept out object-header lines the read had cached (see ``evict_into`` in
+    :func:`build_chunk_map`), so on a sparse file this walk re-fetches some of
+    them — bounded extra ranged reads, still no chunk data. A node whose
+    ``typeSize == 0`` is a group
     (or a typeless link) and is descended; anything else is a dataset yielded
     with the ``H5Dataset`` its classification already parsed, so the caller
     can build its chunk map without re-parsing the object header. Each node's
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -290,8 +290,17 @@ def build_chunk_map(h5obj, path: str, *, evict_into: set | None = None) -> Chunk
     dataset yields a single pseudo-chunk covering the full first axis
     (mirroring h5py's ``get_offset()`` treatment in the bench extractor).
     Cache lines the B-tree walk pulls into ``h5obj.cache`` are evicted before
-    returning (issue #460) — the object-header lines the ``H5Dataset`` parse
-    touched stay cached, since every dataset in the group shares them.
+    returning (issue #460); the lines the ``H5Dataset`` parse itself touched
+    are deliberately kept (the snapshot is taken after the parse). Parse-line
+    residency does grow per *distinct* dataset rather than being fully shared,
+    but it is bounded by the group's object-header / metadata footprint, not by
+    the chunk count: measured on a real GEDI L1B beam group (PR #462 evidence),
+    a whole group's read left **11 lines / 46 MB** resident at the 4 MiB
+    default — the data path's own lines included — out of **90 distinct lines**
+    ever touched across the group, against the hundreds of MB the unevicted
+    walks stranded. The absent-path ``KeyError`` below returns before the
+    snapshot on purpose: nothing walk-touched exists yet, and its parse lines
+    are retained by the same choice.
 
     ``evict_into`` defers that eviction for a caller whose builds run
     concurrently on one handle: pass a set and the walk's keys are *added* to
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -185,7 +185,12 @@ def _walk_chunk_btree(ds) -> list[tuple[tuple[int, ...], int, int, int]]:
 
 
 def _cache_keys(h5obj) -> set:
-    """Snapshot of ``h5obj.cache``'s keys (empty for cache-less handles)."""
+    """Snapshot of ``h5obj.cache``'s keys (empty for cache-less handles).
+
+    ``set(dict)`` is a single C-level copy, so it cannot observe a concurrent
+    insertion mid-iteration the way a Python-level comprehension can (review
+    finding on PR #462).
+    """
     return set(getattr(h5obj, "cache", None) or ())
 
 
@@ -204,11 +209,20 @@ def _evict_new_cache_lines(h5obj, before: set) -> int:
     Only keys absent from ``before`` are dropped — pre-existing lines (the
     front-of-file metadata block every parse shares) always survive. Returns
     the number of lines evicted.
+
+    Structurally non-raising, which matters because callers run it from a
+    ``finally`` where a raise would replace a successful ``ChunkMap`` return
+    with a housekeeping error (review finding on PR #462): ``list(cache)``
+    snapshots the keys in one C-level copy instead of iterating the live dict
+    (a Python-level comprehension can hit ``RuntimeError: dictionary changed
+    size during iteration``), and ``pop(k, None)`` cannot ``KeyError`` on a
+    line another thread already dropped. Both operations complete without
+    releasing the GIL, so no blanket ``except`` is needed here.
     """
     cache = getattr(h5obj, "cache", None)
     if not cache:
         return 0
-    added = [k for k in cache if k not in before]
+    added = [k for k in list(cache) if k not in before]
     for k in added:
         cache.pop(k, None)
     return len(added)
--- a/src/zagg/index/inline.py
+++ b/src/zagg/index/inline.py
@@ -228,7 +228,36 @@ def _evict_new_cache_lines(h5obj, before: set) -> int:
     return len(added)
 
 
-def build_chunk_map(h5obj, path: str) -> ChunkMap:
+def _evict_deferred_lines(h5obj, deferred: set) -> int:
+    """Drop the cache lines a pooled ``read_fn`` fan-out deferred (issue #460).
+
+    :func:`build_chunk_map`'s immediate eviction is only safe on a handle no
+    other thread is reading: ``read_fn``'s lazy map builds run inside
+    ``_read_paths_pooled``'s thread pool (``data_source.read_workers``,
+    default 8, one shared ``h5obj`` — both the planned and the vlen route fan
+    out that way), and h5coro's cached ``ioRequest`` reads ``self.cache``
+    *outside* ``cache_locks``, so popping a line another thread is mid-read
+    raises there (review finding on PR #462, reproduced as a ``KeyError`` in
+    ``ioRequest`` plus silently-degraded ``H5Promise`` parses). Those builds
+    therefore only *record* their walk lines, and ``read_group`` evicts them
+    through this helper in a ``finally`` — after the route returned, so the
+    ``ThreadPoolExecutor`` context has joined and the handle is back to the
+    single worker thread that owns this granule (the ``_pending`` docstring's
+    "each granule is read by a single worker thread" invariant). Returns the
+    number of lines evicted; same non-raising discipline as
+    :func:`_evict_new_cache_lines`.
+    """
+    cache = getattr(h5obj, "cache", None)
+    if not cache or not deferred:
+        return 0
+    evicted = sum(cache.pop(k, None) is not None for k in list(deferred))
+    deferred.clear()
+    if evicted:
+        logger.debug(f"  chunk maps: evicted {evicted} deferred cache line(s) after the fan-out")
+    return evicted
+
+
+def build_chunk_map(h5obj, path: str, *, evict_into: set | None = None) -> ChunkMap:
     """Build a :class:`ChunkMap` for one dataset by walking its metadata.
 
     Metadata-only: no chunk is ever read or decompressed. A contiguous-layout
@@ -237,6 +266,16 @@ def build_chunk_map(h5obj, path: str) -> ChunkMap:
     Cache lines the B-tree walk pulls into ``h5obj.cache`` are evicted before
     returning (issue #460) — the object-header lines the ``H5Dataset`` parse
     touched stay cached, since every dataset in the group shares them.
+
+    ``evict_into`` defers that eviction for a caller whose builds run
+    concurrently on one handle: pass a set and the walk's keys are *added* to
+    it (a GIL-atomic ``set.update``) instead of popped, for the caller to
+    evict once its fan-out has joined — see :func:`_evict_deferred_lines`.
+    Keys other threads' reads happen to add inside the same window are
+    over-attributed to the walk, which is harmless: they are still this
+    group's lines and are dropped only after the pool joins. The default
+    (``None``) evicts immediately, which is what every single-threaded caller
+    wants (``_prebuild_group_maps``, the full-coverage walk, direct use).
 
     Raises ``KeyError`` for an absent path (h5coro's ``metaOnly`` traversal
     never raises on its own — it just leaves default metadata) and
@@ -251,9 +290,14 @@ def build_chunk_map(h5obj, path: str) -> ChunkMap:
     try:
         return _chunk_map_from_dataset(h5obj, ds, path)
     finally:
-        evicted = _evict_new_cache_lines(h5obj, before)
-        if evicted:
-            logger.debug(f"  chunk map {path}: evicted {evicted} cache line(s) the walk touched")
+        if evict_into is not None:
+            evict_into.update(_cache_keys(h5obj) - before)
+        else:
+            evicted = _evict_new_cache_lines(h5obj, before)
+            if evicted:
+                logger.debug(
+                    f"  chunk map {path}: evicted {evicted} cache line(s) the walk touched"
+                )
 
 
 def _chunk_map_from_dataset(h5obj, ds, path: str) -> ChunkMap:
@@ -728,17 +772,23 @@ class InlineIndex(VirtualIndex):
 
             if self.write_back:
                 self._prebuild_group_maps(h5obj, group, data_source)
-            return _vlen_read_group(
-                h5obj,
-                group,
-                data_source,
-                shard_key,
-                grid,
-                arrow=arrow,
-                read_fn=self._chunk_aligned_read_fn(h5obj, planned=True),
-                io_stats=io_stats,
-                siblings=siblings,
-            )
+            # read_fn is hoisted out of the call so the finally can reach its
+            # deferred cache lines (issue #460 — see _evict_deferred_lines).
+            read_fn = self._chunk_aligned_read_fn(h5obj, planned=True)
+            try:
+                return _vlen_read_group(
+                    h5obj,
+                    group,
+                    data_source,
+                    shard_key,
+                    grid,
+                    arrow=arrow,
+                    read_fn=read_fn,
+                    io_stats=io_stats,
+                    siblings=siblings,
+                )
+            finally:
+                _evict_deferred_lines(h5obj, read_fn.deferred_lines)
 
         rp = data_source.get("read_plan")
         # Two routes, one addressing seam (issue #170 phase 2): sources with a
@@ -758,8 +808,19 @@ class InlineIndex(VirtualIndex):
             # group's datasets to the granule manifest.
             self._prebuild_group_maps(h5obj, group, data_source)
         read_fn = self._chunk_aligned_read_fn(h5obj, planned=planned)
-        if planned:
-            return _planned_read_group(
+        try:
+            if planned:
+                return _planned_read_group(
+                    h5obj,
+                    group,
+                    data_source,
+                    shard_key,
+                    grid,
+                    arrow=arrow,
+                    read_fn=read_fn,
+                    io_stats=io_stats,
+                )
+            return _read_group_full(
                 h5obj,
                 group,
                 data_source,
@@ -769,16 +830,10 @@ class InlineIndex(VirtualIndex):
                 read_fn=read_fn,
                 io_stats=io_stats,
             )
-        return _read_group_full(
-            h5obj,
-            group,
-            data_source,
-            shard_key,
-            grid,
-            arrow=arrow,
-            read_fn=read_fn,
-            io_stats=io_stats,
-        )
+        finally:
+            # The route's pooled fan-out has joined here, so the lines its
+            # lazy chunk-map builds deferred can be dropped (issue #460).
+            _evict_deferred_lines(h5obj, read_fn.deferred_lines)
 
     def _chunk_aligned_read_fn(self, h5obj, *, planned=True):
         """Build the addressing seam: planned ranges, compiled decode.
@@ -807,7 +862,10 @@ class InlineIndex(VirtualIndex):
         THIS granule's pending sub-dict instead (``_pending_for`` — keyed per
         granule so concurrent in-flight granules never interleave, issue
         #180), accumulating the granule's maps across groups for
-        ``finish_granule`` to persist. Each dataset gets
+        ``finish_granule`` to persist. Those lazy builds defer their cache-line
+        eviction onto ``read_fn.deferred_lines`` (issue #460) because they run
+        across the read pool's threads; ``read_group`` drains it once the route
+        has joined. Each dataset gets
         its own single-dataset in-memory ``Index`` (~ms), so a spec hidefix
         rejects degrades that dataset alone.
 
@@ -839,6 +897,13 @@ class InlineIndex(VirtualIndex):
         # and degrade innocent paths with it (review finding, PR #173).
         indices: dict = {}
         direct: set[str] = set()  # datasets pinned to the h5coro decoder
+        # Cache lines the lazy builds below touched, evicted by ``read_group``
+        # once this callable's fan-out has joined (issue #460): these builds
+        # run on ``read_workers`` threads sharing one h5obj, where popping a
+        # line races h5coro's unlocked cached ``ioRequest`` — see
+        # :func:`_evict_deferred_lines`. Exposed on the callable so every
+        # route can drain it in a ``finally``.
+        deferred: set = set()
 
         def _vidx_for(path):
             vidx = indices.get(path)
@@ -888,7 +953,7 @@ class InlineIndex(VirtualIndex):
                     # boundary workaround is needed here.
                     return h5obj.readDatasets([path])[path]
                 try:
-                    cm = maps[path] = build_chunk_map(h5obj, path)
+                    cm = maps[path] = build_chunk_map(h5obj, path, evict_into=deferred)
                 except Exception:
                     if planned and hyperslice is not None:
                         # The planned route required the map before #170 too:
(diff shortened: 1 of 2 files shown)


src/zagg/index/inline.py
```diff
@@ -200,7 +233,11 @@ def build_chunk_map(h5obj, path: str) -> ChunkMap:
     ds = H5Dataset(h5obj, path, earlyExit=True, metaOnly=True, enableAttributes=False)
     if ds.meta.typeSize == 0:
         raise KeyError(path)
-    return _chunk_map_from_dataset(h5obj, ds, path)
+    before = _cache_keys(h5obj)
+    try:
+        return _chunk_map_from_dataset(h5obj, ds, path)
+    finally:
+        _evict_new_cache_lines(h5obj, before)
```
espg: 🤖 *from Claude (review)*

**Blocker: this eviction is not thread-safe, and I reproduced a hard failure on the shipped fixture.**

`read_fn` builds unmapped chunk maps lazily (`inline.py:862`) and it is called **concurrently on one `h5obj`**: `_read_paths_pooled` fans the coordinate and variable/filter reads across `_read_workers(data_source)` threads, default **8** (`src/zagg/processing/read.py:811`, `:1175`; `src/zagg/processing/read_vlen.py:287` for the GEDI vlen route that motivated issue #460). So one thread's `finally` pops cache lines another thread is actively using, and h5coro's cached `ioRequest` reads the dict **outside** `cache_locks`:

```python
if cache_line not in self.cache:
    with self.cache_locks[cache_line]:
        if cache_line not in self.cache:
            self.cache[cache_line] = memoryview(self.driver.read(...))
data_view = self.cache[cache_line]                        # unguarded
data_blocks += self.cache[cache_line][start:stop],        # h5coro.py:277
```

Repro on `tests/data/index/atl03_mini.h5` (`cacheLineSize=512`, `sys.setswitchinterval(1e-7)`, 4 threads calling `build_chunk_map` + 4 calling `h5obj.readDatasets` on the same handle, 6 s):

```
KeyError: 27648
  File ".../h5coro/h5coro.py", line 277, in ioRequest
    data_blocks += self.cache[cache_line][start_index:stop_index],
WARNING H5Coro encountered error reading gt2l/heights/h_ph: invalid data layout: None
```

The identical script with `_evict_new_cache_lines` stubbed to a no-op: **0 errors, 0 warnings.**

Two failure modes, both bad:
1. `KeyError` escapes `build_chunk_map` -- `read_fn`'s bare `except Exception` silently degrades that dataset to `direct`, or re-raises and sinks the whole group on the planned route.
2. A concurrent `H5Dataset` parse breaks mid-flight and `H5Promise` **swallows** it (`h5promise.py:43`, the `invalid data layout: None` warning above), so a dataset reads back missing/degraded with only a log line. That is a silent-wrong-data path, not merely a crash.

Every phase-1 test is single-threaded, so CI cannot see any of this.

Concrete options:
1. Take eviction out of the fan-out: build a group's maps serially *before* any pooled read and evict there. `_prebuild_group_maps` already is exactly that pass -- it is just gated on `write_back`. With no eviction happening inside the pool, nothing races.
2. Never touch the shared dict: read the B-tree nodes with `caching=False` (the plan's rejected option (a)). It needs a small h5coro `readField(..., caching=False)` plumb, but it is race-free by construction and is the "upstream later" note brought forward.
3. Keeping eviction in-pool would require a lock covering **every** cached `ioRequest` on the handle, which zagg cannot install.

Whichever way it lands, please add a regression test that drives concurrent `build_chunk_map` + `readDatasets` on one handle -- the repro above is ~20 lines and fails reliably today.

src/zagg/index/inline.py
```diff
@@ -184,12 +184,45 @@ def walk(addr: int) -> None:
...
+    transient peak; a later read that does want an evicted line just
+    re-fetches it (B-tree nodes are not data, so in practice none does).
+    Only keys absent from ``before`` are dropped — pre-existing lines (the
+    front-of-file metadata block every parse shares) always survive. Returns
+    the number of lines evicted.
+    """
+    cache = getattr(h5obj, "cache", None)
+    if not cache:
+        return 0
+    added = [k for k in cache if k not in before]
```
espg: 🤖 *from Claude (review)*

`added = [k for k in cache if k not in before]` iterates the live dict in a **Python-level** loop, so a concurrent cached `ioRequest` inserting a line mid-iteration raises. In the same repro as the comment on line 240, this line raised **322 times in 8 s**:

```
  File "src/zagg/index/inline.py", line 211, in _evict_new_cache_lines
    added = [k for k in cache if k not in before]
RuntimeError: dictionary changed size during iteration
```

Two things to fix here independent of how the concurrency question is settled:

1. Snapshot before iterating -- `added = [k for k in list(cache) if k not in before]` (`list(dict)` completes in C without a GIL release, unlike the comprehension).
2. This raise happens inside `build_chunk_map`'s `finally`, so it **replaces the successful return** -- a fully built `ChunkMap` is thrown away by a housekeeping error. Cache hygiene should never be able to fail a read: either make `_evict_new_cache_lines` non-raising, or wrap the `finally` body.

`_cache_keys` at line 189 uses `set(...)` over the dict, which is a C-level loop and did not reproduce, but the same snapshot discipline is worth applying for symmetry.

src/zagg/index/inline.py
```diff
@@ -200,7 +233,11 @@ def build_chunk_map(h5obj, path: str) -> ChunkMap:
     ds = H5Dataset(h5obj, path, earlyExit=True, metaOnly=True, enableAttributes=False)
     if ds.meta.typeSize == 0:
         raise KeyError(path)
-    return _chunk_map_from_dataset(h5obj, ds, path)
+    before = _cache_keys(h5obj)
```
espg, line 332: 🤖 *from Claude (review)*

The snapshot sits **after** the `H5Dataset` parse, so every dataset's object-header lines are permanently retained -- the plan specified the snapshot "before `_walk_chunk_btree`", and the docstring at lines 223-225 justifies the deviation with "the object-header lines the `H5Dataset` parse touched stay cached, since every dataset in the group shares them."

That sharing claim does not hold as a general property. Mapping all 23 datasets of `tests/data/index/atl03_mini.h5` at `cacheLineSize=512` grows `len(h5obj.cache)` from **1 to 39**, with 16 of the 23 builds adding lines that are never evicted -- i.e. residency still grows per *distinct dataset*, roughly 1.6 lines each. At 512 B that is nothing; the question is whether GEDI L1B beam-group headers co-locate inside a 4 MiB line, and that is exactly the unverified premise the acceptance criterion ("~tens of MB" per beam group) rests on. A GEDI beam group has dozens of datasets; if their headers do not share lines, the residual is hundreds of MB and phase 1 does not reach its target.

Two ways to settle it:
- Snapshot at function entry (before the `H5Dataset(...)` construction) so the call evicts everything it added, matching the plan's wording. Cost is re-fetching a header line per dataset, which is one small ranged read, not a walk.
- Or keep it as-is and record the measured per-dataset header residency from the `data/gedi_debug/gedi_drive_backend.py` run in the PR body, so the docstring claim is grounded.

Two smaller points on the same hunk:
- The `ds.meta.typeSize == 0` / `raise KeyError(path)` arm returns before `before` exists, so an absent-path probe strands its parse lines with no eviction at all. `read_fn` catches that `KeyError` and continues, so it is a reachable path.
- No test covers the many-distinct-datasets shape, which is the GEDI shape and the only one where the residual shows. The existing tests all map a single path (or repeat the same paths), so they pass either way.

src/zagg/index/inline.py
```diff
@@ -184,12 +184,45 @@ def walk(addr: int) -> None:
...
+    Only keys absent from ``before`` are dropped — pre-existing lines (the
+    front-of-file metadata block every parse shares) always survive. Returns
+    the number of lines evicted.
+    """
+    cache = getattr(h5obj, "cache", None)
+    if not cache:
+        return 0
+    added = [k for k in cache if k not in before]
+    for k in added:
+        cache.pop(k, None)
```
espg: 🤖 *from Claude (review)*

Low severity / your call: the eviction drops `h5obj.cache` entries but leaves `h5obj.cache_locks` untouched. It is a `defaultdict(threading.Lock)` keyed by the same aligned offsets (`h5coro.py:125`), so every walk-touched line permanently adds a `Lock` + dict slot that survives to granule close. Measured flat across repeated builds of the same paths (the lock already exists), so the growth is per *distinct line ever touched* -- which for `rxwaveform`'s 13,790 B-tree nodes is ~13.8k locks per beam group, ~110k per granule, order 15-20 MB retained, times `granule_workers`.

That is ~0.4% of the 4.2 GB being fixed, so it is not a blocker -- but the docstring claims residency is capped at "one walk's transient peak", which is not literally true while the lock table keeps a per-node entry. It also matters for the phase-3 guard: an assertion on `len(h5obj.cache)` staying flat will pass while `len(h5obj.cache_locks)` grows.

Note that popping the lock is itself racy (two threads could then hold different `Lock` objects for the same line and both fetch it), so this is only safe once the exclusivity question on line 240 is resolved -- flagging it so the choice is deliberate rather than implicit.

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Fixed in b360cafe.

- `_evict_new_cache_lines` now iterates `list(cache)` (one C-level snapshot) instead of the live dict, so the `RuntimeError: dictionary changed size during iteration` you reproduced cannot occur; `pop(k, None)` was already there and cannot `KeyError` on a line another thread dropped.
- Rather than wrapping the `finally` in a blanket `except`, the helper is now documented as **structurally non-raising**: snapshot + pop-with-default both complete without releasing the GIL, so there is no path by which cache hygiene can replace a successful `ChunkMap` return with a housekeeping error. The docstring says exactly that, so a future edit that introduces a Python-level loop there is visibly breaking the contract.
- `_cache_keys` keeps `set(dict)` (already a C-level copy) and now records why in its docstring, per your symmetry note.

The separate question of *when* eviction is allowed to touch the shared dict at all is folded in 7d6b41b2 (reply on the line-240 thread).

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Fixed in 7d6b41b2 — confirmed blocker, and your diagnosis is exactly right. The fix is **deferred eviction** rather than either of options (1) or (2):

- (1) (prebuild serially before the pooled read) was rejected: `_prebuild_group_maps` walks the *selection* datasets too, which is precisely the issue #185 regression (~41% at `granule_workers: 1`) that gating it on `write_back` bought back. Making the race go away by re-walking selection datasets on every default-config read trades one bug for a measured slowdown.
- (2) (`readField(..., caching=False)`) is the right end state and stays the upstream follow-up — it needs an h5coro change, out of scope here.

What landed:

- `build_chunk_map(h5obj, path, *, evict_into: set | None = None)`. Default (`None`) evicts immediately in the `finally` exactly as before — correct for every single-threaded caller (`_prebuild_group_maps`, the full-coverage walk, tests, direct use). With a set passed, nothing is popped; the walk adds `_cache_keys(h5obj) - before` to it (`set.update`, GIL-atomic). Lines other threads add inside that window are over-attributed to the walk, which is harmless — they are still this group's lines and are dropped only after the pool joins.
- `_chunk_aligned_read_fn` creates one `deferred: set`, passes `evict_into=deferred` at the single lazy build site inside `read_fn`, and exposes it as `read_fn.deferred_lines`.
- `read_group` now dispatches **every** route inside `try: return <route>(...) finally: _evict_deferred_lines(h5obj, read_fn.deferred_lines)` (the vlen arm's `read_fn` is hoisted out of the call expression so the `finally` can reach it). At that point the route has returned, so `_read_paths_pooled`'s `with ThreadPoolExecutor(...)` context has joined, and the handle is back to the single thread that owns this granule — the `_pending` docstring's "each granule is read by a single worker thread" invariant. Count logged at DEBUG; the immediate path keeps its own DEBUG line.

Regression test drives the real seam, not a stub: `TestChunkMapCacheEviction::test_pooled_read_fn_defers_eviction_under_concurrency` builds a `read_fn` via `InlineIndex()._chunk_aligned_read_fn(h5obj, planned=True)` on `_open_fixture_small_lines()` and runs 4 threads calling `read_fn(path, hyperslice=[(300, 1500)])` on the chunked `gt1l/heights` datasets against 4 threads calling `h5obj.readDatasets([...])` on the `gt2l` ones, `sys.setswitchinterval(1e-7)` (restored in a `finally`), bounded at ~2 s with a fresh `read_fn` per round so the lazy builds keep happening. It asserts no exceptions, byte-equal contents against a reference handle, and that a final single-threaded `_evict_deferred_lines` leaves none of the deferred keys in `h5obj.cache`.

It reproduces your failure when the deferral is removed — reverting just the `evict_into=deferred` argument fails it in well under a second:

```
AssertionError: assert not [('/gt1l/heights/h_ph', 'KeyError(17920)'), ('/gt2l/heights/h_ph', 'KeyError(59392)')]
```

Full `tests/test_index.py tests/test_processing.py tests/test_read_vlen.py` green (444 passed).

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Taken — fixed in 2d8c683a, and it is now deliberate rather than implicit.

Your exclusivity caveat is what made this safe to do: after 7d6b41b2 every eviction point is single-threaded on the handle (immediate path callers — `_prebuild_group_maps`, the full-coverage walk, direct use — are serial; the deferred path runs only after the read pool has joined). So the two-locks-for-one-line hazard you flagged cannot arise, and both tables are pruned together.

- New `_drop_cache_lines(h5obj, keys)` pops from `h5obj.cache` **and** `h5obj.cache_locks` (`getattr` guard for stub handles, `.pop(k, None)` since it is a `defaultdict`), and is the single place both `_evict_new_cache_lines` and `_evict_deferred_lines` drop lines. Its docstring records why single-threadedness is the precondition, quoting your ~13.8k-locks-per-beam-group / order 15-20 MB numbers.
- The docstring claim you called out ("one walk's transient peak") is now literally true again, since the lock table no longer keeps a per-node entry.
- The phase-3 guard is extended: `test_cache_stays_flat_across_repeated_builds` now pins `len(h5obj.cache_locks)` flat as well. Worth noting your own observation defeats the count assertion on its own — a rebuild re-fetches the same offsets, so an unpruned lock table also stays flat across identical builds — so the test additionally asserts that no walk line lingers in `cache_locks` (`not walk_lines & set(h5obj.cache_locks)`, plus `set(cache_locks) <= set(cache)`). Disabling the prune fails that assertion; the count assertion alone still passes.

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Settled empirically, and the measurement says the snapshot stays where it is. Docstring + test updated in fcc797cd.

**Measured on the real GEDI L1B granule** (BEAM0000, a `data/gedi_debug/gedi_drive_backend.py` derivative, at the 4 MiB default line size, after this PR's fix):

- post-group resident cache: **11 lines / 46 MB** — and that includes the data path's own lines, not just headers;
- **90 distinct lines** ever touched across the whole group (`len(cache_locks)`);
- rows **165,506** across **217 cells**, byte-identical to the reference.

So GEDI beam-group object headers *do* co-locate within a handful of 4 MiB lines, and the residual is tens of MB — the acceptance premise holds without moving the snapshot. Moving it to function entry would force a header re-fetch per dataset to buy nothing (the fixture's 1.6-lines-per-dataset growth is a 512-B-line artifact, which is why it shows there and not on GEDI).

What changed:

- The overclaiming sentence in `build_chunk_map` ("every dataset in the group shares them") is gone. It now states that parse-line residency grows per *distinct* dataset but is bounded by the group's object-header/metadata footprint rather than by chunk count, with the numbers above cited as PR evidence.
- The `KeyError` arm stays as-is, and the docstring now says why: it returns before the snapshot because nothing walk-touched exists yet, and its parse lines are retained by the same deliberate choice as every other build's.
- On the missing many-distinct-datasets coverage: `test_full_coverage_walk_evicts_its_cache_lines` (phase 2) already maps all 23 fixture datasets at 512-B lines, so it is the right test to extend rather than adding a new one. It now asserts `len(maps) > 20`, that residency stays within parse-line growth (`len(cache) <= 2 * len(maps)`), and that the unevicted reference arm busts that bound (`len(ref.cache) > 2 * len(maps)`) — i.e. the assertion discriminates on the shape you were worried about. Measured on the fixture: 39 lines retained for 22 datasets evicted vs 49 unevicted.

src/zagg/index/inline.py
```diff
@@ -703,16 +865,10 @@ def read_group(
...
-            shard_key,
-            grid,
-            arrow=arrow,
-            read_fn=read_fn,
-            io_stats=io_stats,
-        )
+        finally:
+            # The route's pooled fan-out has joined here, so the lines its
+            # lazy chunk-map builds deferred can be dropped (issue #460).
+            _evict_deferred_lines(h5obj, read_fn.deferred_lines)
```
espg, line 896: 🤖 *from Claude (review)*

**The three `read_group` drains — including the vlen hoist this fold added — have no test coverage. Reverting them keeps the suite green and restores the full issue #460 leak.**

I neutralized just the drain (session plugin, `zagg.index.inline._evict_deferred_lines = lambda h5obj, deferred: 0`) and ran the two suites that touch this code:

```
FAILED tests/test_index.py::TestChunkMapCacheEviction::test_pooled_read_fn_defers_eviction_under_concurrency
1 failed, 176 passed, 2 warnings in 2.29s
```

The single failure is the test that calls `_evict_deferred_lines` *itself* (`tests/test_index.py:759`). Nothing exercises `InlineIndex.read_group`s `finally` blocks, so the deferral is a pure regression trap: with `evict_into=deferred` in place (`inline.py:991`) and no drain, **every** walk line is now retained for the whole granule — strictly worse than before this PR, where phase 1 at least evicted immediately. That is the exact failure mode a future refactor of this dispatch would introduce silently.

The vlen arm (`inline.py:826`) is worse off still: no test anywhere constructs `InlineIndex` and calls `read_group` with a `coordinates.level` config — `tests/test_read_vlen.py::TestInlineWriteBackPrebuild` only calls `_prebuild_group_maps` directly, and `_read_group` in that module never routes through the inline backend. So the hoist that motivated the whole fold (GEDI L1B, the product in issue #460s acceptance criterion) is entirely unexercised.

Concrete: add one test per route asserting on the observable — against `_open_fixture_small_lines()` capture `walk_lines` the way `_unevicted_reference` does, call `InlineIndex().read_group(...)` on the planned route, and assert `not walk_lines & set(h5obj.cache)` after it returns. A `monkeypatch` spy on `_evict_deferred_lines` recording `(called, len(deferred))` per route would cover the vlen arm cheaply if driving `_vlen_read_group` end-to-end with a real handle is too heavy.

src/zagg/index/inline.py
```diff
@@ -200,7 +321,18 @@ def build_chunk_map(h5obj, path: str) -> ChunkMap:
     ds = H5Dataset(h5obj, path, earlyExit=True, metaOnly=True, enableAttributes=False)
     if ds.meta.typeSize == 0:
         raise KeyError(path)
-    return _chunk_map_from_dataset(h5obj, ds, path)
+    before = _cache_keys(h5obj)
+    try:
+        return _chunk_map_from_dataset(h5obj, ds, path)
+    finally:
+        if evict_into is not None:
+            evict_into.update(_cache_keys(h5obj) - before)
```
espg: 🤖 *from Claude (review)*

**Phase 3s DEBUG counter is dead on every default-config read.** The `evict_into` branch records lines and logs nothing, so the per-build counter 9a627425 added only fires from the `else` branch — and the only callers that reach it are the two write-back paths (`_prebuild_group_maps`, gated on `self.write_back` at `inline.py:840`; `full_granule_maps`, reached from `finish_granule` only when `write_back_coverage == "full"`) plus direct/test use. With `write_back` off — the shipped default, and the config the GEDI acceptance run in the plan uses — **no** per-build line is emitted at all.

What you get instead is one aggregate per group from `_evict_deferred_lines` (`inline.py:282`):

```
  chunk maps: evicted 137 deferred cache line(s) after the fan-out
```

No dataset attribution, and the count is the over-attributed union (see the separate note on line 310), so it is not a per-dataset footprint. That defeats the stated purpose — the plans phase 3 asked for "a lines-touched counter logged at DEBUG per map build", and the new test comment says it is "so a pathological granule is visible in logs". The pathology issue #460 is about is exactly one dataset (`rxwaveform`, ~13,790 B-tree nodes) inside a group of dozens; the aggregate cannot single it out.

`tests/test_index.py:768` (`test_lines_touched_logged_at_debug`) hides this because it calls `build_chunk_map` directly, i.e. the immediate branch — the one production never takes.

Concrete: log in the deferred branch too, from the same `finally`:

```python
if evict_into is not None:
    added = _cache_keys(h5obj) - before
    evict_into.update(added)
    if added:
        logger.debug(f"  chunk map {path}: deferred {len(added)} cache line(s) the walk touched")
```

and extend the test to assert the message on a `read_fn`-built map, not just a direct `build_chunk_map`.

src/zagg/index/inline.py
```diff
@@ -184,12 +184,133 @@ def walk(addr: int) -> None:
...
+    walks stranded. The absent-path ``KeyError`` below returns before the
+    snapshot on purpose: nothing walk-touched exists yet, and its parse lines
+    are retained by the same choice.
+
+    ``evict_into`` defers that eviction for a caller whose builds run
+    concurrently on one handle: pass a set and the walk's keys are *added* to
+    it (a GIL-atomic ``set.update``) instead of popped, for the caller to
+    evict once its fan-out has joined — see :func:`_evict_deferred_lines`.
+    Keys other threads' reads happen to add inside the same window are
+    over-attributed to the walk, which is harmless: they are still this
```
espg: 🤖 *from Claude (review)*

**"Over-attributed ... which is harmless" is understated: the deferred set routinely sweeps up other datasets *object-header* lines — the very lines the snapshot-after-parse placement exists to retain.** So the deliberate design decided in the fold on line 324 only holds at `read_workers: 1`; at the default 8 it is silently reversed for whichever headers happen to be parsed inside another threads walk window.

Measured on `tests/data/index/atl03_mini.h5` at the fixtures 512-B lines, 4 threads calling `read_fn(p, hyperslice=[(300, 1500)])` on 4 distinct `/gt1l/heights` datasets, 40 rounds with `switchinterval=1e-7`:

```
rounds with parse lines swept into deferred: 23 /40
example over-attributed parse lines: [7168, 7680]
```

(`parse lines` = the keys a lone `H5Dataset(..., metaOnly=True)` parse of each dataset adds on a fresh handle.) Correctness is fine — a re-fetch is always available — but two documented claims stop being true:

1. This docstrings own "the lines the `H5Dataset` parse itself touched are deliberately kept" (line 293) and the grounding you added at 294-303 describe the immediate path. The **11 lines / 46 MB** GEDI figure was measured through the deferred path, so it characterizes `read_workers=8` sweeping headers out, not the retention policy the paragraph above it describes. As written a reader will conclude header residency is bounded *by design*; it is bounded partly *by the race*.
2. `_iter_datasets` (line 424, unchanged) still says the enumeration "is served from the front-of-file metadata block h5coro already cached — one ranged GET". `finish_granule` runs after every groups drain, so on a sparse file the header lines the read paid for may now be gone and the write-back walk re-fetches them.

Cheapest fix is honest wording rather than code: say in the `evict_into` paragraph that the deferred set intentionally also drops lines other threads added during the window — including their object-header lines — so deferred-path residency is *lower* than the immediate paths and the tradeoff is a re-fetch of those headers on the next group / in `finish_granule`s walk, and note that the 11-line figure is a `read_workers=8` measurement. If you would rather keep the invariant, subtract the union of the builds `before` snapshots at drain time (`deferred -= floor`, where `floor` is `_cache_keys(h5obj)` captured in `_chunk_aligned_read_fn` plus each builds `before`) — but I would not add that machinery for what is a memory *win*.

tests/test_index.py
```diff
@@ -1511,6 +1705,43 @@ def flaky(h5obj, ds, path):
...
+        assert set(h5obj.cache) < set(ref.cache)
+        # ...and this is the many-distinct-datasets shape the GEDI residual
+        # lives in (review finding on PR #462): with the whole granule mapped,
+        # residency tracks the parses' object-header lines (a couple per
+        # dataset at the fixture's 512-B lines) and NOT the walks' per-B-tree-
+        # node lines. On GEDI L1B at the 4 MiB default those headers co-locate:
+        # 11 lines / 46 MB resident per beam group, 90 distinct lines touched.
+        assert len(maps) > 20  # every dataset in the granule, not just one
+        assert len(h5obj.cache) <= 2 * len(maps)  # parse-line growth only...
+        assert len(ref.cache) > 2 * len(maps)  # ...which the unevicted arm busts
```
espg: 🤖 *from Claude (review)*

**The `2 * len(maps)` bound is 5 lines from flipping in both directions, and the narrative it claims is not what makes it pass.** Reproduced the exact sequence:

```
maps 22   cache 39   bound 44
ref cache 49
```

So `39 <= 44` and `49 > 44` — five lines of margin each way on a magic scale factor (`2`) that is a guess at parse lines per dataset. Worse, the discriminating assertion does not discriminate on what the comment says: the unevicted arm strands only **10** extra lines across ~21 walks (49 − 39), i.e. the fixtures B-tree node lines are largely *shared* between datasets. `len(ref.cache) > 2 * len(maps)` is therefore satisfied by the parse-line baseline, not by "the walks per-B-tree-node lines" — flip the fixtures chunk count or gain a dataset and it inverts without any behavior change.

The real discrimination is already carried by `assert set(h5obj.cache) < set(ref.cache)` on line 1729, which is scale-free.

Concrete: drop both `2 * len(maps)` lines and assert on the delta instead, which is exactly the property under test:

```python
stranded = set(ref.cache) - set(h5obj.cache)
assert stranded                                  # eviction removed real lines
_, walk_lines, _ = TestChunkMapCacheEviction()._unevicted_reference()
assert walk_lines <= stranded or not walk_lines & set(h5obj.cache)
```

`assert len(maps) > 20` is worth keeping — that one really does pin the many-distinct-datasets shape from the line-324 thread.

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Fixed in 7fabad1d — you were right that reverting the drains kept the suite green. Two new tests in `TestChunkMapCacheEviction`, both sub-second, both failing by construction when the drain is neutralized.

**Planned route** — `test_read_group_drains_deferred_lines_on_the_planned_route`: two `_open_fixture_small_lines()` handles run the identical `InlineIndex().read_group(h5obj, "gt1l", _fixture_data_source(), 1, _LeafSetGrid(_UNALIGNED_LEAVES))`, with `monkeypatch.setattr(inline_mod, "_evict_deferred_lines", lambda h5obj, deferred: 0)` on the reference arm only. Asserts `pd.testing.assert_frame_equal(df, df_ref)` (the drain changes no output) and `set(h5obj.cache) < set(ref.cache)` — the scale-free discriminator you pointed at, no count arithmetic.

**Vlen route** — `test_read_group_drains_deferred_lines_on_the_vlen_route`: takes your cheaper suggestion for the arm itself. `monkeypatch.setattr(read_vlen_mod, "_vlen_read_group", stub)` where the stub does one `read_fn("/gt1l/heights/h_ph", hyperslice=[(0, 10)])` (forcing a lazy build, hence deferred lines), captures `set(read_fn.deferred_lines)`, and returns `None`. The data_source is `_fixture_data_source()` with `coordinates` carrying `level: "segments"`, which is what selects the arm. Then it asserts against `_unevicted_reference()` walk lines both directions: `seen["deferred"] >= walk_lines` (deferred, not evicted) and `not walk_lines & set(h5obj.cache)` (the `finally` drained them). So the hoist itself is under test, not just a spy on the call.

Verified they discriminate — with `_evict_deferred_lines` short-circuited to `return 0` in the source:

```
E       assert not ({18432} & {0, 512, 1024, 1536, 2048, 2560, ...})
FAILED tests/test_index.py::TestChunkMapCacheEviction::test_read_group_drains_deferred_lines_on_the_planned_route
FAILED tests/test_index.py::TestChunkMapCacheEviction::test_read_group_drains_deferred_lines_on_the_vlen_route
2 failed, 97 deselected in 0.38s
```

The third drain (the vlen `finally` and the shared planned/full `finally`) is the same two code paths — `_read_group_full` dispatches through the identical `finally` as `_planned_read_group`, so the planned test covers that block.

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Fixed in 36c315b9 — the counter was indeed dead on every default-config read.

The deferred branch now logs per build, from the same `finally`:

```python
if evict_into is not None:
    added = _cache_keys(h5obj) - before
    evict_into.update(added)
    if added:
        logger.debug(
            f"  chunk map {path}: walk touched {len(added)} cache line(s); "
            "deferred for post-read eviction"
        )
```

Wording says "walk touched … deferred for post-read eviction" rather than "evicted", so the two branches stay distinguishable in a log; the group-level aggregate from `_evict_deferred_lines` is unchanged and now reads as the drain total for lines the per-build records already attributed. The comment above it records why the branch needs its own line (with `write_back` off every build lands here, and the aggregate cannot single out one `rxwaveform`-shaped dataset inside a group of dozens).

Test: new `test_deferred_lines_touched_logged_at_debug`, and per your last sentence it drives a `read_fn`-built map rather than a direct `build_chunk_map` — `InlineIndex()._chunk_aligned_read_fn(h5obj, planned=True)` then `read_fn(self.PATH, hyperslice=[(300, 1500)])`, asserting a record carrying the path plus "deferred". The existing `test_lines_touched_logged_at_debug` stays as the immediate-branch case.

src/zagg/index/inline.py
espg: 🤖 *from Claude*

Taken as wording, no behavior change — 8f9beebd. Your 23/40 measurement is cited in the docstring so the scope is on the record.

The `evict_into` paragraph now says it straight: the window is a plain before/after key diff, so lines *other threads* add inside it are swept in as well — measurably including other datasets’ object-header parse lines (23 of 40 concurrent rounds on the fixture, your figure) — those lines re-fetch on next touch, one small ranged read each, and that bounded cost is the deliberate price of a race-free drain.

The retention paragraph above it no longer reads as an unconditional policy either: it now notes the snapshot-after-parse retention is exact only on the immediate (single-threaded) path, and that the **11 lines / 46 MB** GEDI figure — measured through the deferred path at the default `read_workers` — is if anything the *lower* residency of the two. That is your point (1) without moving the number.

`_iter_datasets` amended per your point (2): "already cached" is now flagged as a best case rather than a guarantee, since each group’s deferred drain may have swept out object-header lines the read had cached, so on a sparse file the write-back walk re-fetches some of them — bounded extra ranged reads, still no chunk data.

**Declined: the `deferred -= floor` subtraction**, on your own reasoning. It would need a floor threaded through `_chunk_aligned_read_fn` plus every build’s `before` snapshot, to protect a memory *win* against a bounded re-fetch — machinery on the hot dispatch path buying a strictly worse residency. Left standing for @espg if the exact-retention invariant is wanted as a hard property later; option (2) (`readField(..., caching=False)` upstream in h5coro) would make the whole question moot and is already the follow-up noted on the line-240 thread.

tests/test_index.py
espg: 🤖 *from Claude*

Fixed in fae74d49 — both `2 * len(maps)` lines are gone, and nothing replaced them with different arithmetic.

What the test asserts now:

```python
assert len(maps) > 20  # every dataset in the granule, not just one
_, walk_lines, _ = TestChunkMapCacheEviction()._unevicted_reference()
assert walk_lines <= set(ref.cache)      # the unevicted arm strands them...
assert not walk_lines & set(h5obj.cache) # ...and eviction dropped them
```

plus the scale-free `assert set(h5obj.cache) < set(ref.cache)` that was already carrying the real discrimination on line 1729. The new pair is a two-directional pin on actual walk-added keys — present in the unevicted arm, absent from the evicted one — so it does not care how many datasets the fixture grows or how many of their B-tree node lines happen to be shared. `assert len(maps) > 20` kept, as you suggested, since it is what pins the many-distinct-datasets shape from the line-324 thread.

I went with the walk-line disjointness rather than your `stranded` variant because `stranded = set(ref.cache) - set(h5obj.cache)` is non-empty by construction once the strict subset holds, so it adds no independent signal; asserting on the reference walk lines does.

Verified it still discriminates — with `_evict_new_cache_lines` short-circuited to `return 0`, the test fails (`Both sets are equal`, then the walk lines would be resident too). Full run: `tests/test_index.py tests/test_processing.py tests/test_read_vlen.py` → 447 passed.

Issue #465, opened 2026-08-17
GEDI DEM-gate follow-ups: DEM-void shot policy; per-shot L2A stored companions


espg:
🤖 *from Claude*

Tracker for two items deliberately deferred from #464's DEM-referenced geolocation-validity gate (espg ruling in-session 2026-08-17: "agree ×3; note 2 and 3 as a follow-up issue"). Unlabeled by design — return when wanted.

## 1. DEM-void shots (water) are dropped by the gate

The #464 filter — `abs((elevation_bin0 + elevation_lastbin)/2 - digital_elevation_model) <= 200` — uses plain expression semantics, so a shot whose TanDEM-X `digital_elevation_model` is void/NaN fails the gate and drops. Ruled acceptable for now (unverifiable placement ⇒ untrusted; water shots are low-value for a vegetation flux store), but it systematically excludes open-water and DEM-void terrain. If a future product wants those shots (bathymetric-adjacent work, inland water), options: a NaN-passes variant of the gate (`or isnan(dem)`), a per-config knob, or an alternative placement check for void areas (e.g. geoid ± tolerance over water).

## 2. Per-shot L2A quantities as stored companions

#464's asset-sourced shot variables make L2A datasets first-class columns, but the ruling keeps the gate filter-only — nothing new is stored. The machinery now supports carrying e.g. `digital_elevation_model` or `sensitivity` into the store as per-shot companions (the same way `noise_mean` / `rx_energy` ride today), which would end the "re-join L2A externally for sensitivity" workaround the GEDI template comment concedes. Needs: a ruling on which quantities earn store bytes, dtype/fill declarations, and (for sensitivity) care that storing it is not read as endorsing it as a filter (the retrieval-fitness exclusion ruling stands).

Refs #464 (the gate + machinery), #442 (non-record-rate asset joins — the wider generalization), #422 (filter taxonomy rulings).

PR #466, opened 2026-08-17
Asset-sourced shot-level variables + GEDI DEM geolocation-validity gate (issue #464)

### LINKED ISSUE #464: GEDI: epoch-poisoned absolute geolocation passes degrade==0 — DEM-referenced validity gate via asset-sourced variables
🤖 *from Claude*

## Symptom

The first GEDI flux store (`serc_gedi_flux.zarr`, 2026-08-17) has elevation ranges spanning kilometers over sea-level SERC: block tensor fits at median 64 m/bin, worst 128 m/bin (8,192 m span), offsets to −1,103 m — the top range bins carry the junk and set the fit width, gutting vertical resolution store-wide (espg observation, paired-notebook first light).

## Diagnosis: epoch-poisoned absolute geolocation passing `degrade == 0`

Shot-rate probe over the shard's granules (BEAM0000, in-shard shots, all `degrade == 0`):

| epoch | `elevation_bin0` median | `elevation_lastbin` median | verdict |
|---|---|---|---|
| 2020-024 | **+29 m** | −96 m | correct (SERC ellipsoid ≈ −35 m ground) |
| 2022-011 | **+44 m** | −61 m | correct |
| 2019-121 (first weeks of science) | **+827 m** (tail to 2,001) | +721 m | bogus ≈ +860 m offset |
| 2023-031 | **+5,761 m** (spread ±8 m) | +5,656 m | bogus constant ≈ +5.8 km offset |

Window *spans* are sane everywhere (105–213 m = 700–1,420 × 0.15 m), so the linspace synthesis is fine — the absolute placement is broken in specific mission epochs, with the L1B `degrade` flag blind to it. Cells mixing epochs force `fit="degrade_resolution"` to cover −1,100…+8,100 m; cells sampled only by a bad granule are pure junk, so read-side quantile trims cannot recover the store.

## Fix: DEM-referenced geolocation-validity gate at write

GEDI's own geolocation sanity practice: **`|elev_lowestmode − digital_elevation_model| ≤ ~150 m`**, both datasets on the already-joined L2A sibling. DEM-referenced, so it works at any real elevation (no brittle epoch date lists, no static ranges that break in mountains). This is a *validity* gate (geolocation sanity) — distinct from the deliberately-excluded L2A retrieval-fitness gates (`quality_flag`, `sensitivity`; espg ruling 2026-08-17, refs #425/#422), which stay out.

## Machinery gap and design

Not expressible in today's yaml: expression filters see only primary-granule columns (`read_vlen.py:604` — `variables` + segment broadcasts), and asset data flows only through `_asset_keep_mask`'s single-dataset-vs-constant predicates (`read_vlen.py:634`). Two designs:

- **(A) asset-expression filters**: an expression branch inside `_asset_keep_mask`. Smallest diff; filter-only; adds a grammar form that (B) would make redundant.
- **(B) asset-sourced shot-level variables** *(recommended, espg-reviewed in-session)*: `variables` entries may declare `asset:`; the shot-number join is factored out of `_asset_keep_mask` and the joined record-rate values feed the existing segment-broadcast machinery. The DEM gate is then the **stock expression filter** over two ordinary columns — no new filter grammar. Scoped to record-rate joins (the L2A pairing); non-record-rate stays with #442. Bonus surface: L2A quantities (sensitivity, DEM) become first-class columns — usable in filters now, storable as per-shot companions later (the template's own comment concedes users must externally re-join L2A for sensitivity today). One policy to specify: a granule with a missing sibling asset yields NaN-with-warning for asset variables (asset *filters* keep their fail-closed semantics).

## Scope

1. Design (B) in `read_vlen.py`/config validation + tests (synthetic mini-granule pins the join + broadcast + expression path).
2. GEDI template: declare `elev_lowestmode` + `digital_elevation_model` as l2a-sourced shot variables; add the expression filter `abs(elev_lowestmode - digital_elevation_model) <= 150`; comment citing this issue.
3. After next release/deploy: flux store rebuild (~$0.07) picks this up together with the #454-adjacent zero-dim read fix.

Refs #425 (vlen/asset grammar), #422 (filter rulings), #442 (non-record-rate asset joins — deliberately not taken here), #460 (sibling read-path issue from the same store's diagnosis).

espg:
Closes #464

Asset-sourced shot-level variables (design B, espg-reviewed on the issue) plus the GEDI DEM-referenced geolocation-validity gate.

## What / approach

The first GEDI flux store showed epoch-poisoned absolute geolocation passing `degrade == 0` (issue #464 table: 2019-121 ≈ +860 m, 2023-031 ≈ +5.8 km offsets), gutting vertical resolution store-wide. The fix is a DEM-referenced validity gate at write, which today's grammar cannot express: expression filters see only primary-granule columns, and sibling-asset (L2A) data flows only through `_sibling_record_mask`'s single-dataset predicates.

Design (B): a `levels.<coordinates.level>.variables` entry may be a mapping `{asset: <name>, path: <template>}`. The dataset is read from the named sibling's handle, joined per record on the declared `assets.<name>.join` key (the shot-number join, factored out of `_sibling_record_mask` into `_sibling_joined_values`), and broadcast to base rate through the existing coordinates-level gather (`_gather_segment_at_parents`) on both the planned and full arms. The DEM gate is then a stock expression filter over ordinary columns — no new filter grammar.

**espg rulings (2026-08-17, in-session):**
- The template filter is the **window-midpoint form**: `abs((elevation_bin0 + elevation_lastbin)/2 - digital_elevation_model) <= 200` — geolocation validity without entangling retrieval quality. Only `digital_elevation_model` is declared as an l2a asset variable (no `elev_lowestmode`).
- DEM-void/NaN shots **drop** by plain expression semantics (no special-casing). This also drops DEM-void water shots; revisiting that inclusion is tracked in issue #465.
- **Filter-only**: the DEM is not stored as a per-shot companion; stored L2A companions are also tracked in issue #465.

**Missing-sibling policy:** asset *variables* yield NaN with a single warning per granule; asset *filters* keep their existing fail-closed semantics (unchanged). Composition in the SHIPPED template (review finding, corrected): the same `l2a` asset also carries the saturation *filter*, and filters fail closed first — a granule with a missing L2A sibling raises on the filter (`no open sibling handle`) and errors out group by group before the NaN arm is ever consulted. The NaN-degrade path is live only for variable-only assets (pinned by test).

## Phases

- [x] Phase 1 — factor the shot-number join out of `_sibling_record_mask` into `_sibling_joined_values` (behavior byte-identical; the 80 existing vlen tests pass unchanged)
- [x] Phase 2 — asset-sourced shot-level variables: config validation (`_validate_overviews` asset form, `_validate_vlen_source` cross-checks) + read path (join → existing coordinates-level gather, both arms, NaN-with-warning missing-sibling policy)
- [x] Phase 3 — GEDI template: `digital_elevation_model` as an l2a asset variable + the midpoint expression gate, citing #464/#465
- [x] Phase 4 — tests: join+broadcast pins on both arms, mixed primary+asset expression drop, missing-sibling NaN + single warning, fail-closed asset filters unchanged, config validation errors, template-driven epoch-drop fixture

## Review/fold log

- Phase-1 adversarial review: 4 findings, all folded (`fc265a53` empty-sibling placeholder contract, `4b233852` duplicate-key refusal, `126d5244` join dedup + direct-unit tests, `0d146645` one join per asset serving filters and variables).

- Phases-2-4 adversarial review: 5 findings — 4 folded (`70090e84` pairless-granule pin + docs, `ace26e92` NaN-minimal dtype promotion, `fec56690` articulate the asset form's 1-D scope, `703da877` refuse a sibling dataset off the join key's rate), 1 declined by design (record-rate pushdown of the gate — real, pre-existing route shape, left standing for a follow-up issue).

## Acceptance evidence

Real-granule drives over HTTPS (issue #464's own granules) posted below: the gate drops 98.6% of shots on the 2019-121 epoch granule and zeroes 6 of 8 beams on 2023-031 (74.7%), while the 2020-024 control loses 1.0%. See the evidence comment for the table and driver caveats.

## Template interaction note

Two other in-flight PRs touch `src/zagg/configs/gedi01b_waveform_healpix_hive.yaml` (a template-sizing small-fix on `worker`/`streaming`/`index` keys, and the temporal-kernels PR on `temporal`/`toc` keys). This PR touches ONLY the `data_source.levels.shots.variables` and `data_source.filters` keys, so the three should merge cleanly in any order.

## Testing

`ruff check --select=E,F,W,I --ignore=E501 src tests` (the CI lint select), `ruff format --check`, and `pytest -v` locally; per-phase results in the phase comments.

## Questions for review

- `ruff check src tests` with the repo's full pyproject select fails on a pre-existing `N818` in `src/zagg/registry.py:64` (`UnknownCapability`) — present on `origin/main`, invisible to the CI lint bot (`--select=E,F,W,I`). Flagged per §4, not fixed here.


src/zagg/processing/read_vlen.py
```diff
@@ -664,32 +664,61 @@ def _sibling_record_mask(
...
+
+def _sibling_joined_values(
+    sibling,
+    group: str,
+    asset: str,
+    cfg: dict,
+    n_records: int,
+    dataset_paths: list[str],
+    read_full,
+) -> tuple[dict[str, np.ndarray], np.ndarray]:
```
espg, line 829: 🤖 *from Claude (review)*

The helper returns gathered *values* but throws the join itself away (`order` / `sib_idx`), so a second consumer on the same asset must redo the whole join. Phase 2 does exactly that: asset-sourced variables need their own `_sibling_joined_values` call, which means a second uncached `read_full(["/{group}/shot_number"])` byte-range read on the primary, a second `sibling.readDatasets` that re-reads the L2A key dataset, and a second `argsort` + `searchsorted` — per group, per granule.

This is not hypothetical for the shipped config: `src/zagg/configs/gedi01b_waveform_healpix_hive.yaml` already declares an `l2a` asset *filter* (`/{group}/rx_assess/rx_clipbin_count` under `data_source.filters`), and #464 phase 2 adds `l2a`-sourced *variables* on that same asset — so every GEDI worker pays the double join. `_read_full` in this module has no cache, so it is real I/O, not a dict hit.

Suggestion: make one join serve every consumer. Either (a) also return `sib_idx` (a small `(sib_idx, matched)` join record) so a second consumer can gather extra sibling datasets off it, or (b) hoist path collection into the caller — build the per-asset union of filter datasets plus asset-variable datasets and make a single call per asset. (b) keeps the current return shape and is the smaller change.

src/zagg/processing/read_vlen.py
```diff
@@ -664,32 +664,61 @@ def _sibling_record_mask(
...
+    sorted_keys = right_keys[order]
+    pos = np.searchsorted(sorted_keys, left_keys, side="left")
+    pos_clip = np.minimum(pos, len(sorted_keys) - 1) if len(sorted_keys) else pos
+    matched = (
+        (sorted_keys[pos_clip] == left_keys)
+        if len(sorted_keys)
+        else np.zeros(n_records, dtype=bool)
+    )
+    sib_idx = order[pos_clip] if len(sorted_keys) else np.zeros(n_records, dtype=np.int64)
+    values_by_path = {p: np.asarray(sib_data[p])[sib_idx] for p in dict.fromkeys(dataset_paths)}
```
espg, line 906: 🤖 *from Claude (review)*

The docstring's contract — "An unmatched row's value is a clamped-gather placeholder, so every consumer must apply its own missing-row policy against ``matched``" — does not hold when the sibling asset has **zero** rows: this gather raises `IndexError: index 0 is out of bounds for axis 0 with size 0`. The `if len(sorted_keys) else` guards just above (lines 716, 717-721, 722) exist precisely to keep that case alive (all-False `matched`, zeros `sib_idx`), and the unconditional gather here defeats them.

Reproduced against the test fixture (an `_l2a_arrays()` variant with length-0 `shot_number` + `quality_flag`, one `quality_flag == 1` asset filter): `IndexError`, with neither the asset nor the dataset named. It is pre-existing — the removed `flag = np.asarray(sib_data[...])[sib_idx]` inside the filter loop raised identically — so not a phase-1 regression. But the refactor promotes it to a *documented* contract that phase 2's NaN-on-missing policy will be built on, and a zero-row sibling is exactly the missing-data shape that policy exists for.

Suggestion: give the empty case an explicit branch rather than documenting a placeholder it cannot produce — e.g. when `len(sorted_keys) == 0`, build each value as `np.zeros((n_records, *arr.shape[1:]), dtype=arr.dtype)` alongside the all-False `matched`; or raise a `ValueError` naming the asset. Either way, say which in the docstring.

src/zagg/processing/read_vlen.py
```diff
@@ -664,32 +664,61 @@ def _sibling_record_mask(
...
+    sib_paths = [right_path] + list(dataset_paths)
+    sib_data = sibling.readDatasets(list(dict.fromkeys(sib_paths)))
+    right_keys = np.asarray(sib_data[right_path])
+    if len(left_keys) != n_records:
+        raise ValueError(
+            f"assets.{asset}.join.left has {len(left_keys)} records; the "
+            f"coordinates level has {n_records}"
+        )
+    # Key-value join: position of each left key in the sibling's rows.
+    order = np.argsort(right_keys, kind="stable")
```
espg, line 888: 🤖 *from Claude (review)*

Nothing checks that the right-side keys are unique, yet the docstring now asserts a property the code does not enforce: "row ``i`` of each gathered array is the sibling's value for primary record ``i``". With duplicate `right_keys`, `argsort(kind="stable")` + `searchsorted(side="left")` silently selects the *first* duplicate in original file order — an arbitrary row, not "the" row.

Under phase 1 that arbitrariness can only move a filter decision. Phase 2 turns the same gather into a **stored data value** (`elev_lowestmode`, `digital_elevation_model`), so a duplicated `shot_number` in an L2A granule would silently write one of N candidate elevations into the store with nothing recording the ambiguity — the same class of silent-junk problem #464 was opened to fix.

Suggestion: enforce it, e.g. `if len(sorted_keys) > 1 and np.any(sorted_keys[1:] == sorted_keys[:-1]): raise ValueError(f"assets.{asset}.join.right has duplicate keys")` — one vectorized comparison on an already-sorted record-rate array, negligible cost. Failing that, make the docstring honest ("on duplicate right keys the first row in stable key order wins"). Enforcing is preferable given what phase 2 does with these values.

src/zagg/processing/read_vlen.py
```diff
@@ -664,32 +664,61 @@ def _sibling_record_mask(
...
-        # Key-value join: position of each left key in the sibling's rows.
-        order = np.argsort(right_keys, kind="stable")
-        sorted_keys = right_keys[order]
-        pos = np.searchsorted(sorted_keys, left_keys, side="left")
-        pos_clip = np.minimum(pos, len(sorted_keys) - 1) if len(sorted_keys) else pos
-        matched = (
-            (sorted_keys[pos_clip] == left_keys)
-            if len(sorted_keys)
-            else np.zeros(n_records, dtype=bool)
+        dataset_paths = [f["dataset"].format(group=group) for f in filters]
```

The PR's stated requirement is byte-identical `_sibling_record_mask` behavior, but the refactor does change one shape and no test pins it. The gather moved from *per filter* — the removed `for f in filters:` body ran `np.asarray(sib_data[path])[sib_idx]` once per filter, so two filters naming the same dataset gathered it twice — to *once per unique path* via `dict.fromkeys(dataset_paths)` on line 723. `TestSiblingAssetJoin` only ever uses two *distinct* sibling datasets (`quality_flag` + `sensitivity`), so the dedup path is unexercised, and this commit adds no tests at all.

Suggestion: one cheap test — two predicates on the *same* sibling dataset (e.g. `quality_flag ge 1` and `quality_flag le 1`) asserting the same rows as `test_join_filters_by_shot_number`. While there, a direct unit on `_sibling_joined_values` would pin the contract phase 2 is about to depend on (`matched` False for unmatched shot 103; the returned dict keyed by unique path), which no test touches today even indirectly.

src/zagg/processing/read_vlen.py
espg: 🤖 *from Claude*

Fixed in 0d146645 — took option (b), hoisted into the route.

`_vlen_read_group` now builds the per-asset union of paths (each predicate's `dataset` plus each asset variable's `path`) and calls the new `_sibling_joins` once per asset; both arms read off that result. `_sibling_record_mask` lost its I/O and now takes `(group, asset_filters, n_records, joins)`, and the variable arm takes `asset_joins.get(asset)`. So the shipped GEDI config pays one primary key read, one `sibling.readDatasets` and one argsort per group per granule, not two.

The two missing-sibling policies are preserved and now sit in one place: an asset carrying a **filter** raises the same `no open sibling handle` ValueError (fail-closed), a **variable-only** asset is simply absent from the returned dict and its columns NaN with one `_warn_missing_sibling_once` per granule. Absence-means-no-handle rather than a `None` value keeps `_sibling_record_mask`'s unpack type-clean without an unreachable guard.

Tests: `test_one_join_per_asset_serves_both_arms` (filter + variable on the same `l2a` asset; spies on `_sibling_joined_values` and on the sibling handle's `readDatasets`, asserting exactly one call each and that the single call's path list is the union), plus `test_variable_only_asset_without_sibling_nans_and_warns` pinning the one-warning degrade against the filter arm's raise.

src/zagg/processing/read_vlen.py
espg: 🤖 *from Claude*

Fixed in fc265a53 — made the contract true rather than document the raise.

`_sibling_joined_values` now branches on `len(right_keys) == 0` and returns the placeholder directly: `matched` all-False, and per path a `np.zeros((n_records, *arr.shape[1:]), dtype=arr.dtype)` (shape-preserving beyond axis 0). The scattered `if len(sorted_keys) else` guards are gone — the empty case never reaches the argsort/searchsorted/gather at all, so there is one branch instead of three half-guards. Zeros are safe for both consumers by construction: filters AND with `matched`, variables NaN `~matched`. The docstring now states the empty sibling as the contract's degenerate case.

Tests: `TestSiblingAssetJoin::test_empty_sibling_drops_every_record` (zero-row sibling + an asset filter — every record fails closed, group returns `None`, no `IndexError`), `test_empty_sibling_variable_is_all_nan` (same sibling on the variable arm — 13 rows kept, column all-NaN), and the direct unit `TestSiblingJoinedValues::test_empty_sibling_returns_placeholder_arrays` (shape `(4,)`, dtype `uint8` preserved).

src/zagg/processing/read_vlen.py
espg: 🤖 *from Claude*

Fixed in 4b233852 — enforced, per your preference.

One vectorized pass on the already-sorted right keys:

```python
dup = sorted_keys[1:] == sorted_keys[:-1]
if dup.any():
    i = int(np.argmax(dup))
    raise ValueError(
        f"assets.{asset}.join.right {right_path!r} has duplicate key "
        f"{sorted_keys[i]}; the record join requires unique sibling keys"
    )
```

Record rate, and it names both the asset and the offending key. This intentionally tightens the filter path too — a duplicate-key sibling is a malformed product either way, and the phase-1/phase-2 arms now share one join, so one check covers both. Docstring updated to say the duplicate raises instead of resolving by sort order.

Tests: `TestSiblingAssetJoin::test_duplicate_sibling_keys_raise` (two L2A rows for shot 101, through `_read_group`) and the direct unit `TestSiblingJoinedValues::test_duplicate_right_keys_raise`, both matching on `duplicate key <N>`.

src/zagg/processing/read_vlen.py

Fixed in 126d5244 — both halves.

(a) The dedup shape, through `_read_group`: `TestSiblingAssetJoin::test_two_predicates_on_one_dataset_anded` puts `quality_flag ge 1` and `quality_flag le 1` on the same sibling dataset and asserts the same result as `test_join_filters_by_shot_number` (shots 101 + 104, 9 rows) — so the once-per-unique-path gather still ANDs both verdicts.

(b) A direct unit class on the join, `TestSiblingJoinedValues` (fixtures `_FakeH5` / `_l1b_arrays` / `_l2a_arrays` as elsewhere in the file):

- `test_matched_vector_and_gather_order` — `matched == [T, T, F, T]`, and the gathered values are the L2A file order reordered to primary record order, not sliced.
- `test_unmatched_row_is_a_placeholder_only` — shot 103's row carries the clamped neighbor's value with `matched[2]` False, pinning why a consumer must test `matched`.
- `test_duplicate_paths_collapse_to_one_key` — the returned dict is keyed by unique path.
- `test_record_count_mismatch_raises` — `join.left has 4 records`.
- plus the empty-sibling and duplicate-key units from the two commits above.

The asset-variable arm phase 2 added also had no tests, so this commit and 0d146645 add three: unmatched record NaN vs. matched value, the empty-sibling all-NaN column, and the missing-handle one-warning degrade.

espg:
🤖 *from Claude*

**Acceptance evidence: the gate on real granules (issue #464's own table), driven over HTTPS through this branch's read path.**

Driver: `_read_group` with the shipped template config on the SERC o9 shardmap (`demo/outputs/shardmap_gedi_serc_o9.json`), h5coro HTTPDriver, per-granule L2A sibling handles, gate OFF (expression filter removed) vs ON — all 8 beam groups, in-shard planned reads. Shot counts are distinct `shot_number`s surviving all filters.

| granule | epoch (issue #464 verdict) | shots gate OFF | shots gate ON | dropped |
|---|---|---:|---:|---:|
| GEDI01_B_2019121080625_O02166 | 2019-121 (bogus ≈ +860 m) | 1,602 | 23 | **98.6%** |
| GEDI01_B_2023031234128_O23435 | 2023-031 (bogus ≈ +5.8 km) | 1,638 | 415 | **74.7%** (6 of 8 beams zeroed) |
| GEDI01_B_2020024042519_O06322 | 2020-024 (correct) | 1,629 | 1,612 | 1.0% |

Row totals (per-sample): 2019-121 gate off 1,367,418 → on 16,193; 2023-031 off 1,149,361 → on 290,915; 2020-024 off 1,341,540 → on 1,325,430.

Per-beam on the 2023-031 granule the gate's verdict is beam-resolved: BEAM0000/0001/0101/0110/1000/1011 lose every shot (the constant ≈ +5.8 km offset), while BEAM0010/0011 keep 203/212 of 216/217 — the epoch poisoning evidently did not hit all beams equally, which is exactly the behavior a DEM-referenced per-shot gate handles and an epoch blocklist would not.

Two driver-local caveats (production code untouched; the `data/gedi_debug/gedi_drive_fixed.py` wrap-the-fix precedent):
1. A 1-element planned run returns 0-d / `None` through `execute_read_plan` on main — the PR #461 defense (`np.atleast_1d` + None-check) was applied as a wrapper in the driver.
2. On these granules h5coro's HTTPS driver deterministically fails one large `rxwaveform` hyperslice per beam with `invalid location to read chunk` (pre-existing reader issue, independent of this PR — the fleet reads via S3 + inline index). The drive therefore omitted the `rxwaveform` column; every gate-relevant dataset (record coordinates, link arrays, `degrade`, elevation endpoints, and the L2A join: `rx_clipbin_count` + `digital_elevation_model`) was read for real, and shot/row survival counts are unaffected by the waveform payload.

src/zagg/processing/read_vlen.py
```diff
@@ -579,6 +626,35 @@ def _base_read_fn(path, hyperslice=None):
         else {}
     )
 
+    # Asset-sourced record-level variables (issue #464): read from the sibling
+    # handle via the record-key join, NaN'd where the record has no sibling
+    # row, then broadcast through the same coordinates-level gather map the
+    # companions above reuse — identical on both arms, since ``parent_safe`` /
+    # ``valid`` are already at this read's rate. A granule with no open
+    # sibling handle degrades to all-NaN columns with one warning per granule
+    # (asset FILTERS above keep their fail-closed semantics).
```

This comment's missing-sibling arm is unreachable for the shipped GEDI template, and the PR body's "composition" note states the opposite outcome.

`_sibling_joins` **raises** for any asset that carries a *filter* (`read_vlen.py:754-759`), and the template hangs both the `rx_clipbin_count` filter and the DEM variable off the same `l2a` asset. So for a granule whose shardmap entry carries no L2A URL, `worker.py` hands the read `siblings=None` (`worker.py:559-563` only sets the kwarg when `siblings` is non-empty), the join raises, and all eight groups fail through the group-error path (`_record_read_error` → `read_errors`, granule contributes nothing) — no all-NaN DEM column and no warning is ever emitted. The PR body's "with the template's DEM gate declared, a granule with a missing L2A sibling produces an all-NaN DEM column and the expression filter then drops every shot" describes a path only a *variable-only* asset can take, which no shipped config has.

Suggest correcting the PR body and adding a clause here: for the shipped template an unpaired granule is a hard group-read failure (fail-closed via the clipbin filter), and the NaN+warn arm is grammar generality. It matters for #465: adding a stored L2A companion does not loosen this, and *dropping* the clipbin filter later would silently flip an unpaired granule from "hard error" to "every row dropped by the gate".

src/zagg/processing/read_vlen.py
```diff
@@ -579,6 +626,35 @@ def _base_read_fn(path, hyperslice=None):
...
+                name: np.full(len(coarse_lats), np.nan, dtype=np.float64) for name in name_paths
+            }
+        else:
+            values_by_path, matched = join
+            record_vals = {}
+            for name, path in name_paths.items():
+                # Always float64: the unmatched-row NaN (and the missing-
+                # sibling arm above) need a float column either way.
+                vals = _select_column(np.asarray(values_by_path[path]), None, name)
+                vals = vals.astype(np.float64)
```

The unconditional `float64` upcast costs more than the comment above claims it needs. NaN-filling requires a *float* dtype, not float64 — and GEDI's `digital_elevation_model` is float32, so this doubles the widest kind of array this route allocates. It is also an asymmetry with the plain companion arm, which keeps its source dtype (`_gather_segment_at_parents` NaN-fills only when the dtype is already floating).

The cost is not confined to the filter, either: `asset_cols` is gathered to read rate, sliced into `data_dict`, returned, and then concatenated and permuted with every other column by `_concat_and_group` / `_group_columns` into the shard's pooled arrays. The template's "filter-only, not stored" is true of the *store* only — every surviving row carries the DEM through pooling and the per-cell split. Against this file's own scale figures (~1.2 GB per base-rate companion on a full-arm GEDI beam group, 2 GB worker) float32-vs-float64 is a few hundred MB of headroom for a column nothing reads after the gate.

Suggest `vals = vals.astype(np.promote_types(vals.dtype, np.float32))` (float32 stays float32, float64 stays float64, uint64/int64 still land on float64, so no precision surprise) and rewording the comment to "a float dtype". Dropping unreferenced asset columns after the expression pass would recover the rest, but that needs aggregation knowledge the read path does not have — a #465 note rather than this PR.

src/zagg/config.py
```diff
@@ -2383,10 +2417,28 @@ def _validate_overviews(data_source: dict) -> None:
...
                         f"levels.{name}.variables: variable names must be non-empty strings"
                     )
-                if not isinstance(tmpl, str) or not tmpl:
+                if isinstance(tmpl, dict):
+                    # Asset-sourced record-level variable (issue #464): the
+                    # dataset is read from the named sibling asset's handle and
+                    # joined per record on the declared ``assets.<name>.join``
+                    # key. Shape only here; cross-checks against ``assets`` and
+                    # ``coordinates.level`` live in ``_validate_vlen_source``.
+                    unknown = set(tmpl) - {"asset", "path"}
```
espg, line 2426: 🤖 *from Claude (review)*

Rejecting `column` in the asset form leaves a dead end whose error message recommends the very key this validator refuses. `_vlen_read_group` calls `_select_column(..., None, name)` (`read_vlen.py:651`), so an asset variable pointed at a 2-D sibling dataset raises `variable 'rh': N-D dataset requires an integer 'column'` — and there is no way to supply one.

The 2-D case is the next consumer, not a hypothetical: L2A's per-shot `rh` is `(n_shots, 101)`, and #465 is precisely about storing L2A companions. Either accept `column` here and thread it through (the base-level form already means the same thing; `_select_column` just needs `entry.get("column")` instead of the hardcoded `None`), or keep the restriction and say so in the error the reader actually hits (e.g. "asset variables read 1-D sibling datasets"). Cheap either way — the part worth fixing is the two messages contradicting each other.

src/zagg/processing/read_vlen.py
```diff
@@ -627,69 +708,188 @@ def _base_read_fn(path, hyperslice=None):
...
+        i = int(np.argmax(dup))
+        raise ValueError(
+            f"assets.{asset}.join.right {right_path!r} has duplicate key "
+            f"{sorted_keys[i]}; the record join requires unique sibling keys"
+        )
+    pos = np.searchsorted(sorted_keys, left_keys, side="left")
+    pos_clip = np.minimum(pos, len(sorted_keys) - 1)
+    matched = sorted_keys[pos_clip] == left_keys
+    sib_idx = order[pos_clip]
+    values_by_path = {p: np.asarray(sib_data[p])[sib_idx] for p in dict.fromkeys(dataset_paths)}
```
espg, line 906: 🤖 *from Claude (review)*

`sib_idx` is bounded by `len(right_keys)`, not by the length of each gathered dataset, so a sibling product whose value dataset is shorter than its key dataset surfaces as a bare `IndexError: index 3 is out of bounds for axis 0 with size 3` in a worker log naming neither the asset nor the dataset. Every other malformed-input case on this route is refused in domain words: the planned arm's short-hyperslice check ("the record link does not tile the declared base extent"), the full arm's twin, `assets.{asset}.join.left has N records`, and the duplicate-key refusal three lines up.

Phase 2 raises the stakes: a length mismatch on the filter arm only mis-verdicts a record, while the same mismatch on the variable arm feeds a *column*. A one-line per-path check in the same style would close it, e.g. refuse `len(sib_data[p]) != len(right_keys)` with `assets.{asset}: dataset {p!r} has N rows against M join keys`.

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
```diff
@@ -103,6 +110,16 @@ data_source:
...
+    # DEM-referenced geolocation-validity gate (issue #464): epoch-poisoned
+    # absolute geolocation passes degrade == 0 (2019-121 ≈ +860 m, 2023-031 ≈
+    # +5.8 km constant offsets — the first flux store's kilometer-span cells),
+    # so gate the waveform window's MIDPOINT against the L2A DEM (espg ruling
+    # 2026-08-17: midpoint form, not elev_lowestmode — tests geolocation
+    # without entangling retrieval quality). A third VALIDITY gate, distinct
+    # from the excluded retrieval-fitness gates above. DEM-void (NaN) shots
+    # drop by plain expression semantics — including DEM-void water shots;
+    # revisiting that inclusion is tracked in issue #465.
+    - expression: "abs((elevation_bin0 + elevation_lastbin)/2 - digital_elevation_model) <= 200"
```
espg, line 133: 🤖 *from Claude (review)*

Non-blocking, for a follow-up rather than this PR — the shipped behavior here is correct. As an `expression` filter this gate can only run at sample rate, after the waveforms are decoded and every column assembled (`read_vlen.py:684-702`), so the granules it exists to reject pay full price. The 2023-031 case in #464 is a *constant* ≈5.8 km offset — every shot in the granule fails — yet the shard still plans, fetches and gathers all of that granule's `rxwaveform` rows before dropping 100% of them, and `_record_obs_read` counts them. The #464 scope-3 rebuild therefore pays the full read of the poisoned epochs it is meant to exclude.

Worth noting that no filter prunes the plan today: the record-level structured `degrade` gate above is evaluated *after* the base read too, so this is a pre-existing shape rather than something phase 3 introduced. But the hook already exists — `plan_read(..., coarse_mask=coarse_mask)` takes a per-record mask — so a record-rate verdict (all three inputs here are shot-level, making the verdict constant per record) could be ANDed into `coarse_mask` before planning and cut the poisoned granules to near-zero read. That needs a record-level `expression` form plus filter-before-plan ordering, i.e. its own issue; filing it beside #465 would keep the cost visible.

src/zagg/processing/read_vlen.py

Correct, and the behavior is intentional — folded as documentation + a pin, no semantics change (70090e84).

`l2a` carries both the `rx_clipbin_count` filter and the DEM variable, and `_sibling_joins` raises for a filter asset before the variable arm is ever reached, so on the shipped template a pairless granule fails the group and the NaN arm never fires. That's the locked policy split: asset FILTERS fail closed (an unfilterable record must not pass a quality gate silently), asset VARIABLES NaN-degrade with one warning per granule. The NaN arm serves a config whose asset carries variables only.

What changed:

- the block comment at the asset-variable arm now says which arm wins when one asset carries both, names the template as exactly that shape, and points at the pin;
- `_warn_missing_sibling_once` gained a line saying only a variable-only asset reaches it;
- new pin `TestGediTemplate::test_pairless_granule_fails_the_group` — packaged template config, no `siblings`, asserts `l2a` is the filter asset and that the read raises `no open sibling handle` (i.e. the granule fails via the filter). The NaN arm keeps its own coverage in `test_variable_only_asset_without_sibling_nans_and_warns`.

The PR body's composition note is being corrected separately by the coordinating agent.

src/zagg/processing/read_vlen.py

Fixed in ace26e92 — the column now promotes only as far as the NaN requires:

```python
vals = vals.astype(np.promote_types(vals.dtype, np.float32))
```

A float32 DEM stays float32 (no doubled paint on every asset column); an integer flag lands in a float type able to hold its values (uint8 → float32, uint64 → float64). `astype` still copies, so the `vals[~matched] = np.nan` write never reaches the join's shared `values_by_path` — one join per asset now serves several consumers.

Tests added in `TestAssetVariables`:

- `test_float32_sibling_stays_float32` — fixture DEM is float32, column dtype is float32, shot 103 (no L2A row) still NaN;
- `test_integer_sibling_promotes_to_float` — uint8 `quality_flag` as an asset variable → float32 column, unmatched row NaN, matched rows 1.0/0.0;
- `test_nan_fill_never_mutates_the_shared_join` — spies the join's returned arrays and asserts the joined DEM is unmutated after the read (verified it fails if the `astype` is given `copy=False`).

The missing-sibling arm keeps its `np.float64` all-NaN fill: with no handle there is no dtype to preserve.

src/zagg/config.py
espg: 🤖 *from Claude*

Not adding `column` support here — this PR's asset form is scoped to record-rate scalars, and richer L2A companions (`rh` percentiles and friends, which are exactly the 2-D case) are issue #465's vehicle. What is fixed is the misleading failure: the message now names the scope at both layers (fec56690).

Read path — the asset-variable arm no longer routes through `_select_column(..., None, name)` (whose message says "N-D dataset requires an integer 'column'", pointing at a key this form rejects):

```python
vals = np.asarray(values_by_path[path])
if vals.ndim > 1:
    raise ValueError(
        f"asset variable {name!r}: dataset {path!r} is "
        f"{vals.ndim}-D; the asset form takes no 'column' "
        f"selector yet (1-D datasets only; refs issue #465)"
    )
```

Config path — the unknown-key error keeps its existing wording (a test pins it) and appends a hint when the offending key is `column`: `the asset form reads 1-D datasets only, with no 'column' selector yet (refs issue #465)`.

Tests: `TestAssetVariables::test_two_dimensional_sibling_dataset_raises_articulately` drives a `(3, 101)` `rh` sibling dataset through `_read_group`, and `TestAssetVariableValidation::test_column_key_says_why_it_is_unsupported` covers the config hint.

src/zagg/processing/read_vlen.py
espg: 🤖 *from Claude*

Fixed in 703da877 — `_sibling_joined_values` now validates every gathered dataset against the join key's rate before the `[sib_idx]` gather:

```python
for p in dict.fromkeys(dataset_paths):
    n_rows = len(np.asarray(sib_data[p]))
    if n_rows != len(right_keys):
        raise ValueError(
            f"assets.{asset} dataset {p!r} has {n_rows} rows; the join key "
            f"{right_path!r} has {len(right_keys)}"
        )
```

Placed above the empty-sibling early return, so it also catches a zero-row key beside a populated dataset — that is a malformed product, not the degenerate case the placeholder contract covers. Docstring updated alongside the duplicate-key paragraph.

Tests: `TestSiblingJoinedValues` gains `test_dataset_at_a_different_rate_raises` (short), `test_longer_dataset_raises_too`, and `test_populated_dataset_beside_an_empty_key_raises`; `TestAssetVariables::test_short_sibling_dataset_raises_through_the_route` covers it through `_read_group` with the asset and dataset named instead of a bare `IndexError`.

src/zagg/configs/gedi01b_waveform_healpix_hive.yaml
espg: 🤖 *from Claude*

Agreed it's real, and left standing by design — no code change in this PR.

Quantified: the gate is an expression filter over assembled columns, so it prunes rows only after the plan has fetched them. On the 2023-031 analogue the whole granule's DEM offset is ~5.8 km, so every in-shard shot fails and the read still decodes ~1.1M waveform samples to drop all of them. That is the pre-existing shape of the route rather than something this PR introduces: structured record-level filters don't prune `plan_read`'s coarse mask either, so a record-rate pushdown is one change covering both, not a DEM-gate special case.

Two reasons to keep it out of here: it would widen this PR past the asset-variables scope it was labeled for, and opening a follow-up issue is a side-effecting action reserved for espg (CLAUDE.md §6). So this stays as a standing finding for espg to spin into its own issue — the natural framing being "push record-rate filters (structured + asset + DEM gate) into `plan_read`'s coarse mask".

